PR #31, opened 2026-03-27
Externalize blackboard as the first real plugin and add protobuf-based plugin runtime

i386:
This PR introduces plugins to `mesh-llm` for the first time.

This PR adds the first end-to-end plugin runtime, the first built-in plugin running through it, and the first Rust SDK for building additional plugins.

### A real plugin runtime in `mesh-llm`

`mesh-llm` can now launch, supervise, and communicate with plugins as first-class runtime components.

### The first plugin in the system

`blackboard` now runs through that new plugin runtime, making it the first real plugin supported by `mesh-llm`.

### A broad MCP surface for plugins

The new plugin system is not limited to a narrow tools-only interface. Plugins can now participate in a much broader MCP surface, including:

- tools
- prompts
- resources and resource templates
- resource subscriptions
- completion
- logging level control
- tasks
- plugin channel messaging
- bulk transfer flows
- mesh event delivery

### A shared Rust plugin crate

This PR also introduces a shared Rust crate for plugin authors, so new plugins can be built against a real SDK instead of reimplementing transport, lifecycle, and MCP plumbing themselves.

### A full example plugin

The example plugin now demonstrates the full MCP surface currently exposed by the host bridge, giving us a concrete reference for how future plugins should be built.

## Why it matters

This PR changes plugins from a concept into an actual supported capability in `mesh-llm`:

- plugins now exist in the product for the first time
- there is a real runtime for hosting them
- there is a reusable Rust SDK for building more of them
- the initial plugin surface already covers a broad MCP capability set

## Validation

- `cargo check` from `mesh-llm/`
- `cargo test pii_scrub -- --nocapture`

### DIFF
--- a/mesh-llm/src/plugins/blackboard/mod.rs
+++ b/mesh-llm/src/plugins/blackboard/mod.rs
@@ -7,7 +7,7 @@ pub mod mcp;
 
 use anyhow::Result;
 use mesh_llm_plugin::{
-    json_schema_tool, plugin_server_info, MeshVisibility, PluginMetadata, PluginRuntime,
+    json_schema_tool, plugin_server_info, PluginMetadata, PluginRuntime, PluginStartupPolicy,
     SimplePlugin, ToolRouter,
 };
 use schemars::JsonSchema;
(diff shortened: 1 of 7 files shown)


mesh-llm/src/plugins/blackboard/mod.rs
```diff
@@ -7,6 +7,8 @@ use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 use tokio::sync::Mutex;
 
+pub const BLACKBOARD_CHANNEL: &str = "blackboard.v1";
```
michaelneale, line 18: WDYT about putting this in a src/plugins dir to show case cc @i386 ?

i386: Yep great idea. I'll do that.

i386: done

PR #36, opened 2026-03-27
Add tagged cross-platform GitHub release workflow

i386:
## Summary

Adds a tagged GitHub release workflow for `mesh-llm`, moves release packaging into repo-owned scripts and `just` targets, standardizes version bumps across the repo, and documents and automates installation around flavor-specific release bundles.

## What This PR Does

- Adds a GitHub Actions release workflow that runs on pushed `v*` tags.
- Builds and publishes release artifacts for:
  - Linux CPU
  - Linux CUDA
  - Linux AMD ROCm
  - Linux Vulkan
  - macOS Metal
- Publishes the GitHub release automatically from the pushed tag and uploads the generated assets.
- Keeps normal CI separate from tagged release publishing.
- Adds repo-owned release build and packaging paths so the release logic is reproducible outside workflow YAML.
- Adds and wires release helpers in `just`:
  - `just release-version <version>`
  - `just release-build`
  - `just release-build-cuda`
  - `just release-build-amd`
  - `just release-build-vulkan`
  - `just release-bundle <version>`
  - `just release-bundle-cuda <version>`
  - `just release-bundle-amd <version>`
  - `just release-bundle-vulkan <version>`
- Updates version references consistently across:
  - `mesh-llm/src/main.rs`
  - `mesh-llm/Cargo.toml`
  - `mesh-llm/plugin/Cargo.toml`
  - `mesh-llm/src/plugins/example/Cargo.toml`
- Adds an interactive `install.sh` that:
  - probes the machine and recommends a flavor
  - prompts the user to confirm or override it
  - supports non-interactive installs via `MESH_LLM_INSTALL_FLAVOR`
  - removes stale old-style and differently flavored llama binaries before installing the new bundle
- Updates README, RELEASE, and install/update guidance to use the installer flow.

## Release Matrix And Bundle Layout

- Linux CPU remains the generic Linux release asset.
- Linux CUDA is built in the NVIDIA CUDA container path.
- Linux AMD ROCm is built in the ROCm path.
- Linux Vulkan is built on Ubuntu with the Vulkan SDK/dev packages installed.
- macOS remains part of the tagged release flow.
- The publish job waits for the release artifact jobs and assembles the GitHub release from uploaded artifacts.
- Bundles now install flavor-specific llama.cpp binaries so multiple flavors can coexist cleanly:
  - macOS: `rpc-server-metal`, `llama-server-metal`
  - Linux CPU: `rpc-server-cpu`, `llama-server-cpu`
  - Linux CUDA: `rpc-server-cuda`, `llama-server-cuda`
  - Linux ROCm: `rpc-server-rocm`, `llama-server-rocm`
  - Linux Vulkan: `rpc-server-vulkan`, `llama-server-vulkan`
- `mesh-llm --llama-flavor <flavor>` can be used to force a specific bundled pair when multiple flavors are installed in one directory.

## Linux Build Tooling

- `scripts/build-linux.sh` is backend-aware and can build CUDA, ROCm, or Vulkan.
- Linux backend selection can be auto-detected or forced explicitly through `just build backend=...`.
- Added ROCm arch detection for the Linux AMD release path.
- Added llama device detection and then moved the runtime toward using the selected binary pair instead of guessing a backend from host probes alone.
- Launch and packaging paths now resolve flavor-specific llama.cpp binaries consistently.

## Installer Flow

- The recommended install path is now:
  - `curl -fsSL https://raw.githubusercontent.com/michaelneale/mesh-llm/main/install.sh | bash`
- The installer chooses a recommended flavor based on local probes and asks the user what to install.
- The console update notification now points users to the same installer command instead of the old tarball extraction one-liner.

## CUDA Notes

- The Linux CUDA workflow uses `nvidia/cuda:12.8.0-devel-ubuntu22.04`.
- CUDA arch defaults include `sm_120` in addition to the earlier architecture set.
- CUDA release packaging includes the dependency fixes needed for the current container build.

## Release Process After This PR

1. Bump the version consistently:
   - `just release-version v0.X.Y`
2. Build and verify locally if needed:
   - `just build`
   - `just bundle`
3. Commit the release changes.
4. Create and push the tag from the release branch:
   - `git tag v0.X.Y`
   - `git push origin <branch> --tags`
5. GitHub Actions picks up the `v*` tag and:
   - builds the Linux CPU, Linux CUDA, Linux AMD ROCm, Linux Vulkan, and macOS bundles
   - uploads the release assets
   - creates the GitHub release automatically

## Local Repro

- Standard tagged release artifacts can be reproduced with the repo-owned release scripts and `just` targets used by CI.
- Linux CUDA can be reproduced with `just release-build-cuda` and `just release-bundle-cuda v0.X.Y`.
- Linux AMD ROCm can be reproduced with `just release-build-amd` and `just release-bundle-amd v0.X.Y`.
- Linux Vulkan can be reproduced with `just release-build-vulkan` and `just release-bundle-vulkan v0.X.Y`.
- The local portable bundle from `just bundle` now follows the same naming convention as the release bundles for the current platform.

## Notes

- The generic Linux release asset is CPU-only; CUDA, ROCm, and Vulkan are shipped as separate Linux artifacts.
- The unversioned macOS `mesh-bundle.tar.gz` asset still exists for direct archive installs, but the documented install path now goes through `install.sh`.


i386:
@michaelneale the release pipeline works for all targets now

Proof: https://github.com/i386/mesh-llm/actions/runs/23675178362

PR #38, opened 2026-03-27
Hide client VRAM in mesh overview

i386:
showing `0` for vram in Mesh Overview is dumb for client nodes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 # Contributing
 
+Join the [#mesh-llm channel on the Goose Discord](https://discord.gg/goose-oss) for discussion and questions.
+
 This file covers local build and development workflows for this repository.
 
 ## Prerequisites
--- a/Justfile
+++ b/Justfile
@@ -22,34 +22,7 @@ build backend="" cuda_arch="" rocm_arch="":
 
 # Build on macOS Apple Silicon (Metal + RPC)
 build-mac:
-    #!/usr/bin/env bash
-    set -euo pipefail
-    if [ ! -d "{{ llama_dir }}" ]; then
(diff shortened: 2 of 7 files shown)


mesh-llm/ui/src/App.tsx
```diff
@@ -2475,7 +2481,7 @@ function DashboardPage({
                       <TableCell>{peer.statusLabel}</TableCell>
                       <TableCell className="max-w-[180px] truncate">{peer.modelLabel}</TableCell>
                       <TableCell className="text-right">{peer.latencyLabel}</TableCell>
-                      <TableCell className="text-right">{peer.vram_gb.toFixed(1)} GB</TableCell>
+                      <TableCell className="text-right">{peer.displayVramGb.toFixed(1)} GB</TableCell>
```
sheikhlimon: Would it make sense to add a conditional in the table cell to show "n/a" for clients instead?

```suggestion
                      <TableCell className="text-right">{peer.role === 'Client' ? 'n/a' : `${peer.displayVramGb.toFixed(1)} GB`}</TableCell>
```

mesh-llm/ui/src/App.tsx
sheikhlimon: Small nit on the latency display - would it be possible to show an em dash (—) instead of n/a while it’s still waiting for the first RTT?

n/a reads a bit like an error or missing data, whereas `—` makes it clearer that the value just hasn’t been measured yet.

Not a big deal, but could be a nice clarity improvement
```ts
  if (value == null || !Number.isFinite(Number (value))) return '—';
```

sheikhlimon (approved):
LGTM overall, left a small nit

mesh-llm/ui/src/App.tsx
i386: I like that more - agree

PR #43, opened 2026-03-27
Auto-detect CPU-only build when no NVIDIA GPU present

sheikhlimon:
### Summary
- `detect-cuda-arch.sh` checks for NVIDIA GPU presence before attempting arch detection
- If no GPU found, outputs "cpu" and falls back to CPU-only build automatically
- `just build` now works on systems without NVIDIA GPU

### Test plan
- Run `just build` on a system without NVIDIA GPU → builds CPU-only without error


i386:
I don't believe this PR is needed once we merge https://github.com/michaelneale/mesh-llm/issues/36

PR #46, opened 2026-03-28
fix: rpc-server startup reliability — crash detection, retries, better errors

michaelneale:
## Problem

On Linux, users are seeing:
```
Error: rpc-server failed to start on port 43879 within 15s
```

The old code had three issues:
1. **No crash detection** — if rpc-server exited immediately (port conflict, missing lib, bad device), it would poll the port for the full 15s before reporting a generic error
2. **TOCTOU port race** — `find_free_port()` drops the listener before rpc-server binds, so the port can be stolen in between. Silent failure, no retry.
3. **15s too tight for Linux CUDA** — CUDA init (PTX JIT, driver init) can take 15-30s+ on first run

## Fix

| Change | Effect |
|---|---|
| `try_wait()` each poll iteration | Crash/port-conflict detected immediately, error includes exit status + log tail |
| Log tail in all error paths | Shows `/tmp/mesh-llm-rpc-{port}.log` contents — no more guessing |
| 3 retries with fresh port | Handles TOCTOU port race automatically |
| Timeout 15s → 60s | Gives CUDA init time on slow Linux systems |
| Progress logging every 10s | Visible feedback during slow startups |

## Tests

- 11 new unit tests for launch helpers (port allocation, log tail, crash detection, retry logic)
- Added parallel macOS CI job

## CI

Added `macos` job running in parallel with `linux` — same build + test + smoke test + fmt + clippy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,12 +6,18 @@ on:
   pull_request:
     branches: [main]
 
+env:
+  # Tiny model for integration tests (~138MB Q8_0, has chat template)
+  MODEL_URL: https://huggingface.co/unsloth/SmolLM2-135M-Instruct-GGUF/resolve/main/SmolLM2-135M-Instruct-Q8_0.gguf
+  MODEL_FILE: SmolLM2-135M-Instruct-Q8_0.gguf
+
 jobs:
   linux:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


michaelneale:
Splitting — CI changes moved to a dedicated branch, rpc-server changes shelved.

PR #55, opened 2026-03-28
Add model search, download APIs, and configurable model storage

i386:
## What this lets users do

- list curated recommended models with `mesh-llm models`
- search Hugging Face GGUFs with `mesh-llm search <query>`
- inspect an exact model ref before downloading with `mesh-llm show <ref>`
- download an exact curated or Hugging Face model ref with `mesh-llm download <ref>`
- download split Hugging Face GGUFs by giving the first part URL or ref
- configure one or more model directories, including external drives and NAS paths
- provide a Hugging Face token for authenticated search and download
- start a background model download over HTTP and then poll progress or cancel it

## CLI changes

- `mesh-llm models`
- `mesh-llm search <query>`
- `mesh-llm show <ref>`
- `mesh-llm download <ref>`
- `mesh-llm provenance repair --source huggingface`

`download` is now exact-only. Discovery happens through `models`, `search`, and `show`.

## API changes

New management API endpoints on port `3131`:

- `GET /api/models`
- `GET /api/models/search?q=...`
- `GET /api/models/show?ref=...`
- `POST /api/models/download`
- `GET /api/models/download?ref=...`
- `DELETE /api/models/download?ref=...`

`POST /api/models/download` starts or resumes a background job and returns immediately. `GET` reports progress. `DELETE` cancels the job.

## Metadata and Provenance

This PR now writes a sidecar metadata file next to downloaded models: `<model>.manifest.json`.

That metadata matters because model filename and `_name_or_path` are not strong identity signals. Different exports, revisions, or fine-tunes can share similar names, and different runtimes can package the same checkpoint in different formats.

The sidecar gives us a canonical provenance record for each local model, including:

- Hugging Face source repo
- source revision when known
- source file or remote path when known
- resolved download URL
- canonical internal identity and display name
- local metadata such as download time and file size

That lets the rest of mesh-llm answer questions like “is this the same model in GGUF and MLX?” using provenance first instead of name heuristics.

If the sidecar is missing, `mesh-llm provenance repair --source huggingface` can scan local models, reconcile them against Hugging Face or curated metadata, and rebuild sidecars when there is a strong match. Runtime identity stays strict: the repair command is the manual path for backfilling metadata rather than silently guessing from filenames during normal operation.

## Implementation notes

- moves curated mesh metadata into `mesh-llm/models/metadata.toml` and embeds it at build time
- routes CLI and API model operations through one shared `models` module
- keeps curated metadata as the mesh-specific overlay while using Hugging Face for search
- preserves split GGUF download handling from #50 in the refactored model path
- stores canonical model provenance in a shared internal struct and serializes it to `.manifest.json` sidecars
- writes provenance sidecars for curated downloads, exact Hugging Face downloads, and direct URL downloads
- adds provenance repair for existing local model files and Hugging Face-style model directories

## Validation

- `just build`
- `cargo test -p mesh-llm`

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,6 +427,66 @@ mesh-llm --model ~/my-models/custom-model.gguf
 
 Downloads have resume support, so interrupted transfers pick up where they left off.
 
+## Manifest sidecars
+
+Downloaded models get a sibling manifest file with the suffix `.manifest.json`.
+
+- `Qwen3-8B-Q4_K_M.gguf` writes `Qwen3-8B-Q4_K_M.gguf.manifest.json`
+- `Qwen3-30B-A3B-UD-Q4_K_XL-00001-of-00002.gguf` writes `Qwen3-30B-A3B-UD-Q4_K_XL-00001-of-00002.gguf.manifest.json`
+- directory-style model bundles write a sibling manifest next to the directory, for example `MyModel/` writes `MyModel.manifest.json`
+
+Current manifest schema:
+
+```json
+{
+  "version": 1,
+  "source": {
+    "provider": "huggingface",
+    "repo": "Qwen/Qwen3-8B-GGUF",
+    "revision": "abc123",
+    "file": "Qwen3-8B-Q4_K_M.gguf",
+    "resolved_url": "https://huggingface.co/Qwen/Qwen3-8B-GGUF/resolve/abc123/Qwen3-8B-Q4_K_M.gguf"
+  },
+  "identity": {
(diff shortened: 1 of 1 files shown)


mesh-llm/models/metadata.toml
```diff
@@ -0,0 +1,358 @@
+[[model]]
```
ndizazzo, line 1: This is coming from `build.rs` right? Perhaps it could live on GitHub as a job + artifact that gets refreshed more often?

I only mention it because one of the biggest issues I had working on a similar set of curated models, was keeping up with the sheer number of updates. I would constantly be slammed manually testing and compiling lists of models every time a new one dropped.

PR #84, opened 2026-03-30
Normalize HTTP proxy request framing and connection handling

### LINKED ISSUE #66: Replicated-model routing never advances past the first target
## Summary
Replicated-model routing never advances past the first host because `ModelTargets::get()` indexes with `counter`, but that counter does not advance across requests.

## Problem
`ModelTargets` is cloned per request in the API proxy path. Because the internal selection state restarts at `0`, replicated models always route to target index `0` instead of round-robin balancing across the full target set.

This weakens resilience under load and reduces effective fail-soft behavior when the first host is degraded but still reachable.

## Expected behavior
Requests for replicated models should advance selection across the available targets, and that progression should survive `ModelTargets` clones used during request handling.

## Acceptance criteria
- Replicated models route across all configured targets instead of always using the first one.
- Cloned `ModelTargets` instances share the same progression state.
- Regression tests cover multi-host selection and clone behavior.

## Related work
- Draft fix: PR #62

### LINKED ISSUE #67: Proxy routing should not depend on a single peeked buffer
## Summary
The HTTP proxy currently makes routing decisions from a single `TcpStream::peek()` buffer.

## Problem
The proxy extracts `model`, session hints, and JSON body data from one peeked slice before selecting a target. TCP does not preserve request boundaries, so valid requests can arrive fragmented across multiple reads. Larger bodies, partial headers, chunked transfer encoding, and `Expect: 100-continue` can all produce incomplete data in that first snapshot.

When parsing from the peeked bytes fails or fields are absent in the initial slice, the proxy can fall back to generic host selection. That risks routing a request to the wrong model, losing MoE stickiness, and polluting demand tracking.

## Expected behavior
Routing should be based on a buffered, correctly framed HTTP request rather than packet timing. Incomplete requests should continue buffering until the proxy can safely decide how to route them.

## Acceptance criteria
- Routing no longer depends on a single `peek()` call.
- Fragmented POST bodies route correctly.
- Bodies larger than 32 KiB route correctly.
- Chunked requests and `Expect: 100-continue` are handled without misrouting.
- Regression tests cover each of the above cases.

## Related work
- Draft fix: PR #63

### LINKED ISSUE #68: Pipeline proxy should not reject valid fragmented requests
## Summary
Pipeline mode can reject valid fragmented requests because it parses JSON only from the bytes captured during the initial proxy peek.

## Problem
Unlike the plain raw-proxy path, pipeline mode does not first reconstruct a complete HTTP request before inspecting the body. If the client sends a valid request whose body arrives across multiple packets or reads, pipeline mode can treat that as invalid JSON and return `400`.

That makes the most complex request path less robust than the normal proxy path.

## Expected behavior
Pipeline mode should only operate on a fully buffered, parsed request body. If pipeline setup cannot safely proceed, the proxy should fall back to the direct raw proxy path rather than rejecting a valid request.

## Acceptance criteria
- Pipeline mode uses the buffered request path instead of a partial peek buffer.
- Valid fragmented requests no longer fail with `400 invalid JSON body`.
- Unsupported or failed pipeline setup falls back to direct proxying.
- Regression tests cover fragmented requests and fallback behavior.

## Related work
- Draft fix: PR #64
- Depends on the buffered request handling in PR #63

### LINKED ISSUE #69: Management API request parsing should not depend on a single read
## Summary
The management API still parses requests from a single read buffer instead of a framed HTTP request.

## Problem
`mesh-llm/src/api.rs` reads from the client socket once and then splits/parses the resulting string to route and handle requests. That has the same partial-read fragility the main inference proxy had before PR #63.

Valid POST requests to management endpoints can be fragmented across multiple reads, especially for larger bodies or slower clients. A single-read parser can misparse or reject those requests depending on timing.

## Expected behavior
Management API request handling should read and parse complete HTTP requests instead of assuming one read contains the full request.

## Acceptance criteria
- Management API request parsing no longer depends on a single socket read.
- Fragmented POST bodies are handled correctly.
- Larger request bodies are handled correctly.
- Regression tests cover fragmented request parsing for relevant management endpoints.

### LINKED ISSUE #70: Proxy should not emit duplicate 100 Continue responses
## Summary
The main proxy should not risk sending duplicate `100 Continue` interim responses when forwarding requests with `Expect: 100-continue`.

## Problem
The buffered proxy path now handles `Expect: 100-continue` locally so it can continue reading the request body. But if the original request is then replayed upstream with the `Expect` header still intact, the upstream may also emit its own `100 Continue`.

That can lead to duplicate interim responses or otherwise incorrect handshake semantics for clients and upstream servers.

## Expected behavior
The proxy should either:
- fully terminate the `Expect: 100-continue` handshake locally and strip `Expect` before replaying upstream, or
- leave the handshake entirely to the upstream.

It should not do both.

## Acceptance criteria
- Requests using `Expect: 100-continue` do not produce duplicate interim responses.
- Proxy/upstream handshake ownership is explicit and tested.
- Regression tests cover the chosen behavior.

## Related work
- Buffered request handling landed in PR #63

### LINKED ISSUE #71: Define and harden proxy behavior for HTTP/1.1 connection reuse
## Summary
HTTP/1.1 connection reuse and pipelined multi-request connections still need explicit handling in the proxy.

## Problem
The current buffered proxy path routes based on the first request seen on a connection and then hands the socket off to a selected upstream. If a client reuses the connection or pipelines multiple requests, extra request bytes may already be buffered and get forwarded to the same upstream without per-request routing decisions.

That behavior is unsafe if different requests on the same client connection should route to different model targets.

## Expected behavior
The proxy should either:
- explicitly support persistent multi-request connections with a request-by-request routing loop, or
- explicitly constrain/disable that behavior so the routing contract is clear.

## Acceptance criteria
- The proxy's intended behavior for HTTP/1.1 keep-alive and pipelined requests is documented.
- Multi-request connections are either handled correctly per request or explicitly rejected/terminated.
- Tests cover at least one reused-connection or pipelined-request scenario.

### LINKED ISSUE #72: Add end-to-end proxy robustness integration tests
## Summary
The proxy robustness work needs end-to-end integration tests with real upstream servers, not only unit tests.

## Problem
Current regression coverage focuses on parsing and helper behavior. That is useful, but it can miss protocol bugs at the actual proxy boundary, especially around replaying buffered bytes, `Expect: 100-continue`, chunked bodies, fallback behavior, and what the upstream server actually receives.

## Expected behavior
Add integration-style tests that start a fake upstream HTTP server, send real requests through the proxy entrypoint, and assert both proxy behavior and upstream-observed requests.

## Acceptance criteria
- Tests exercise fragmented POST bodies through the real proxy listener.
- Tests exercise chunked requests through the real proxy listener.
- Tests exercise `Expect: 100-continue` through the real proxy listener.
- Tests verify that the upstream receives the correct request bytes/headers/body.
- Tests cover pipeline fallback behavior through the real proxy path.

## Related work
- PR #63 and PR #64 added unit-level regression coverage

i386:
## Summary
This PR hardens the routed HTTP inference path so requests are fully framed before routing, forwarded with a strict single-request connection contract, and exercised end-to-end across streaming and fallback scenarios.

## Behavior Before
The proxy behavior was inconsistent across several edge cases:
- routing decisions could depend on partial socket reads instead of a fully framed request
- fragmented and chunked request bodies were not covered end-to-end through the real proxy listener
- `Expect: 100-continue` handling was not owned cleanly at the proxy boundary
- pipeline-planner fallback paths were not validated through the real listener path
- follow-up bytes on reused or pipelined HTTP/1.1 connections could be carried past the initial routing decision instead of being explicitly constrained
- hand-rolled header parsing had to replicate RFC framing behavior itself
- chunked uploads could be rejected based on wire framing overhead rather than decoded body size
- replicated-model target selection did not reliably advance across available targets
- management API parsing depended on a single socket read
- tunnel timeout diagnostics truncated sub-second first-byte timeouts in logs/tests

## Behavior After
This PR normalizes that behavior by:
- buffering and framing one complete HTTP request before routing
- parsing headers with `httparse` instead of manual split/find iteration
- honoring RFC framing precedence where `Transfer-Encoding` wins over `Content-Length`
- preserving request bodies correctly for fragmented and chunked uploads
- responding to `Expect: 100-continue` locally and stripping the upstream `Expect` header
- forcing routed upstream requests to `Connection: close`
- enforcing a single-request-per-connection contract for routed inference traffic so follow-up bytes are not replayed to an already-selected upstream
- bounding chunked uploads with a separate wire-size guard while keeping the decoded-body limit correct
- falling back cleanly from pipeline planning to direct proxying when pipeline setup fails
- sharing round-robin state correctly across replicated target selections
- parsing management API requests from buffered reads instead of assuming one recv contains the full request
- reporting first-byte timeout durations with sub-second precision

## Additional Coverage
The branch adds end-to-end proxy tests that start fake upstream servers and assert upstream-observed requests and client-visible behavior for:
- fragmented POST bodies
- chunked request bodies
- chunked requests whose wire framing exceeds the decoded body size
- `Expect: 100-continue`
- pipeline fallback through the real proxy path
- pipelined follow-up bytes not being forwarded upstream
- direct streaming responses arriving incrementally through the routed proxy
- pipeline streaming responses arriving incrementally after planner pre-plan injection
- client disconnects mid-stream not hanging the proxy path
- slow but healthy follow-up response chunks after the first byte timeout window
- `/api/events` SSE initial payload plus pushed updates

## Documentation
The routed HTTP/1.1 connection contract is documented in:
- `mesh-llm/docs/DESIGN.md`
- `mesh-llm/docs/TESTING.md`

## Validation
- `cargo test -p mesh-llm proxy::tests::test_read_http_request_ -- --nocapture`
- `cargo test -p mesh-llm test_api_proxy_integration_ -- --nocapture`
- `cargo test -p mesh-llm relay_response_ -- --nocapture`
- `cargo test -p mesh-llm test_api_events_sends_initial_payload_and_updates -- --nocapture`

## Supersedes
- #60
- #62
- #63
- #64

Closes #66
Closes #67
Closes #68
Closes #69
Closes #70
Closes #71
Closes #72

### DIFF
--- a/mesh-llm/src/proxy.rs
+++ b/mesh-llm/src/proxy.rs
@@ -519,13 +519,9 @@ pub async fn route_to_target(
                         let _ = send_503(tcp_stream).await;
                         return false;
                     }
-                    if let Err(e) = upstream.shutdown().await {
-                        tracing::warn!(
-                            "API proxy: failed to half-close request stream to local llama-server on {port}: {e}"
-                        );
-                        let _ = send_503(tcp_stream).await;
-                        return false;
-                    }
+                    // Don't half-close the write side — llama-server interprets
+                    // a TCP FIN as a client disconnect and stops streaming.
+                    // The Connection: close header already signals one-shot.
                     if let Err(e) = tokio::io::copy(&mut upstream, &mut tcp_stream).await {
                         tracing::debug!("API proxy (local) ended: {e}");
                     }


michaelneale:
hrm "buffering and framing one complete HTTP request before routing" - slightly concerned with large images/streams, and if there are any side effects with different clients and different agents. The idea was to make it as transparent as possible to the host then 

"sharing round-robin state correctly across replicated target selections" - not sure I follow?

michaelneale:
have some tweaks coming

PR #85, opened 2026-03-30
feat(protocol): cut over control-plane streams to protobuf v1

ndizazzo:
This PR hard-cuts the QUIC control-plane from ad-hoc JSON/raw-bytes to a versioned protobuf schema (meshllm.node.v1) on ALPN mesh-llm/1. There is no JSON fallback — nodes that don't speak mesh-llm/1 are rejected at connection time.

Talking with @i386, we discussed converting the communication protocol over to protobuf so we could version additions / modifications 

### What changed

**New proto schema** — `mesh-llm/proto/node.proto`

Package `meshllm.node.v1`. All control-plane message types are now defined in one place:

| Message | Stream | Notes |
|---|---|---|
| `GossipFrame` / `PeerAnnouncement` | 0x01 | includes `gen` field for generation gating |
| `TunnelMap` | 0x03 | — |
| `RouteTableRequest` / `RouteTable` | 0x05 | request-only from passive/client nodes |
| `PeerDown` | 0x06 | requires reachability confirmation before removing |
| `PeerLeaving` | 0x07 | removes only the sending peer (anti-forgery) |

Endpoint IDs are raw `bytes` throughout. `CompactModelMetadata` + `ExpertsSummary` carry GGUF-derived model info in gossip so peers share full model visibility without additional round-trips.

**Wire framing**

Every control frame uses a fixed 5-byte header: `[u8 type][u32 LE length][proto bytes]`. Max frame size is 8 MB (`MAX_CONTROL_FRAME_BYTES`). `NODE_PROTOCOL_GENERATION = 1`.

**Admission gate (quarantine-until-gossip)**

A connecting peer is held in quarantine until a valid `GossipFrame` with `gen == 1` is received. Wrong-generation and malformed frames are dropped before any peer state is mutated. This prevents stale or corrupted peers from polluting routing tables.

**Streams 0x02 and 0x04 are unchanged** — raw TCP relay pass-through, no framing applied.

---

### What is explicitly out of scope

- `nostr.rs` — zero diff, Nostr discovery is JSON over HTTP and stays that way
- Invite/join token flow — still serde_json + base64url, no proto
- `model_key` derivation — unchanged (file-stem based)
- Plugin/blackboard streams — not touched
- Fly.io console app / relay — not touched

---

### Runtime characteristics (measured)

- Joiner admitted: **~2 s** after first valid gossip on `mesh-llm/1`
- Dead-peer cleanup: **~41 s** via reconnect → gossip → probe cycle
- Passive/client nodes receive protobuf route tables and can serve `/v1/models` without being promoted to peers

---

### Test coverage

245 tests pass (`cargo test -p mesh-llm`). New unit tests cover:

- Legacy-ALPN rejection (gen=0 frames rejected)
- Malformed payload rejection (invalid proto bytes)
- Wrong-generation rejection
- Gossip propagation and peer promotion
- Passive-client route-table behavior (request-only, never admitted as peer)
- Dead-peer cleanup on kill

---

### Docs

- `message_protocol.md` (new) — canonical reference for all stream types, framing layout, admission model, and out-of-scope areas
- `mesh-llm/docs/DESIGN.md` — added Control-Plane Protocol section, updated QUIC stream table with Format column, updated gossip fields and peer management sections
- `mesh-llm/docs/TESTING.md` — added protobuf v1 test scenarios, ALPN rejection testing instructions, updated dead-peer cleanup timeline (~41 s)

---

### Upgrade / compatibility note

This is a **breaking wire change**. Any node still on `mesh-llm/0` will be rejected and must upgrade. Mixed meshes are not supported.

### DIFF
--- a/message_protocol.md
+++ b/message_protocol.md
@@ -1,0 +1,205 @@
+# mesh-llm Message Protocol
+
+This document describes the wire protocol for control-plane communication between mesh-llm nodes. All control-plane traffic runs over QUIC using the `meshllm.node.v1` protobuf schema.
+
+## ALPN
+
+Control-plane connections use ALPN `mesh-llm/1`.
+
+Peers that connect on the legacy ALPN `mesh-llm/0` are rejected at admission. Mixed meshes containing both `/0` and `/1` nodes are unsupported. **No JSON fallback** — the hard cut-over means JSON control-plane fallback does not exist.
+
+## Stream Types
+
+Each QUIC connection carries multiple logical streams, distinguished by a 1-byte prefix:
+
+| Byte | Name | Direction | Format |
+|------|------|-----------|--------|
+| 0x01 | GOSSIP | bidirectional | protobuf `GossipFrame` |
+| 0x02 | TUNNEL_RPC | bidirectional | raw TCP relay (not protobuf) |
+| 0x03 | TUNNEL_MAP | send | protobuf `TunnelMap` |
+| 0x04 | TUNNEL_HTTP | bidirectional | raw TCP relay (not protobuf) |
+| 0x05 | ROUTE_REQUEST | bidirectional | protobuf `RouteTableRequest` / `RouteTable` |
+| 0x06 | PEER_DOWN | send | protobuf `PeerDown` |
+| 0x07 | PEER_LEAVING | send | protobuf `PeerLeaving` |
+
+Streams 0x02 and 0x04 are raw TCP relay tunnels. They carry llama.cpp RPC and HTTP traffic respectively and are not subject to protobuf framing or generation validation.
+
+## Framing
+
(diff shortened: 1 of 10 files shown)


message_protocol.md
```diff
@@ -0,0 +1,205 @@
+# mesh-llm Message Protocol
+
```
i386, line 2: move this to `docs/` and call it PROTOCOL.md

i386 (changes requested):
Looks good, very workable going forward. 

We do need to figure out how to do backwards compatibility as mentioned in https://github.com/michaelneale/mesh-llm/pull/85#issuecomment-4152275817 

I would remove any json payloads from plugins too.

i386:
@ndizazzo this should do the trick https://github.com/michaelneale/mesh-llm/pull/92 😎 - if it looks good, merge it into your branch

PR #92, opened 2026-03-30
feature(protocol): add support for transition to protobuf

i386:
Talking with @i386, we discussed converting the communication protocol over to protobuf so we could version additions / modifications 

### What changed

**New proto schema** — `mesh-llm/proto/node.proto`

Package `meshllm.node.v1`. All control-plane message types are now defined in one place:

| Message | Stream | Notes |
|---|---|---|
| `GossipFrame` / `PeerAnnouncement` | 0x01 | includes `gen` field for generation gating |
| `TunnelMap` | 0x03 | — |
| `RouteTableRequest` / `RouteTable` | 0x05 | request-only from passive/client nodes |
| `PeerDown` | 0x06 | requires reachability confirmation before removing |
| `PeerLeaving` | 0x07 | removes only the sending peer (anti-forgery) |

Endpoint IDs are raw `bytes` throughout. `CompactModelMetadata` + `ExpertsSummary` carry GGUF-derived model info in gossip so peers share full model visibility without additional round-trips.

**Wire framing**

Every control frame uses a fixed 5-byte header: `[u8 type][u32 LE length][proto bytes]`. Max frame size is 8 MB (`MAX_CONTROL_FRAME_BYTES`). `NODE_PROTOCOL_GENERATION = 1`.

**Admission gate (quarantine-until-gossip)**

A connecting peer is held in quarantine until a valid `GossipFrame` with `gen == 1` is received. Wrong-generation and malformed frames are dropped before any peer state is mutated. This prevents stale or corrupted peers from polluting routing tables.

**Streams 0x02 and 0x04 are unchanged** — raw TCP relay pass-through, no framing applied.

---

### What is explicitly out of scope

- `nostr.rs` — zero diff, Nostr discovery is JSON over HTTP and stays that way
- Invite/join token flow — still serde_json + base64url, no proto
- `model_key` derivation — unchanged (file-stem based)
- Plugin/blackboard streams — not touched
- Fly.io console app / relay — not touched

---

### Runtime characteristics (measured)

- Joiner admitted: **~2 s** after first valid gossip on `mesh-llm/1`
- Dead-peer cleanup: **~41 s** via reconnect → gossip → probe cycle
- Passive/client nodes receive protobuf route tables and can serve `/v1/models` without being promoted to peers

---

### Test coverage

245 tests pass (`cargo test -p mesh-llm`). New unit tests cover:

- Legacy-ALPN rejection (gen=0 frames rejected)
- Malformed payload rejection (invalid proto bytes)
- Wrong-generation rejection
- Gossip propagation and peer promotion
- Passive-client route-table behavior (request-only, never admitted as peer)
- Dead-peer cleanup on kill

---

### Docs

- `message_protocol.md` (new) — canonical reference for all stream types, framing layout, admission model, and out-of-scope areas
- `mesh-llm/docs/DESIGN.md` — added Control-Plane Protocol section, updated QUIC stream table with Format column, updated gossip fields and peer management sections
- `mesh-llm/docs/TESTING.md` — added protobuf v1 test scenarios, ALPN rejection testing instructions, updated dead-peer cleanup timeline (~41 s)

---

## Followup What changed
- restored backward compatibility for legacy `mesh-llm/0` control-plane payloads while keeping protobuf as the preferred `mesh-llm/1` path
- accepted both ALPNs on the listener and taught outbound connects to prefer `/1` with `/0` fallback
- switched gossip, tunnel-map, route-table, peer-down, and peer-leaving handling to choose protobuf vs legacy JSON/raw per negotiated connection
- updated protocol docs to describe mixed-version behavior instead of a hard cut-over

## Why
PR #85 hard-cut the control plane to protobuf and rejected older JSON/raw peers at connection time. That violated the repo's protocol-compatibility requirement and would strand pre-protobuf nodes during rollout.

## Impact
- post-protobuf nodes can interoperate with pre-protobuf nodes over negotiated `mesh-llm/0`
- `mesh-llm/1` still enforces protobuf framing and generation checks
- mixed meshes remain limited to the legacy field set on `/0`, which is the expected compatibility mode

## Validation
- `cargo test -p mesh-llm`
- added a real mixed-version async test that stands up a post-protobuf node plus a legacy `/0` peer and verifies gossip plus route-table exchange over a live QUIC connection

### DIFF
--- a/mesh-llm/src/mesh.rs
+++ b/mesh-llm/src/mesh.rs
@@ -1,7 +1,8 @@
 //! Mesh membership via iroh QUIC connections.
 //!
-//! Single ALPN, single connection per peer. Bi-streams multiplexed by
-//! first byte: 0x01 = gossip, 0x02 = tunnel (RPC), 0x03 = tunnel map, 0x04 = tunnel (HTTP).
+//! Multiple ALPNs are accepted for backward compatibility, but control traffic still uses
+//! one QUIC connection per peer. Bi-streams are multiplexed by first byte:
+//! 0x01 = gossip, 0x02 = tunnel (RPC), 0x03 = tunnel map, 0x04 = tunnel (HTTP).
 
 use anyhow::Result;
 use base64::Engine;
@@ -12,6 +13,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::Arc;
 
+use crate::protocol::*;
 use tokio::sync::{watch, Mutex};
 
 /// Demand signal for a model — tracks interest via API requests and --model declarations.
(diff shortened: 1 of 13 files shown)


ndizazzo:
Taking this one instead

mesh-llm/src/mesh.rs
```diff
@@ -258,6 +310,113 @@ pub fn scan_local_models() -> Vec<String> {
     names
 }
 
+/// Scan model directories for GGUF files and return a map of stem name to file size in bytes.
+pub fn scan_local_model_sizes() -> HashMap<String, u64> {
```
i386, line 314: All of these guessing games go away with https://github.com/michaelneale/mesh-llm/pull/102 @michaelneale @ndizazzo

i386 (commented):
Can we get an integration test? pull a v0 client from GitHub releases, fire it up as a node and run the v2 code - check that it all works etc.

ndizazzo (approved):

PR #94, opened 2026-03-30
[codex] Stabilize streaming proxy integration tests

i386:
## Summary

Stabilize the streaming proxy integration tests by removing scheduler-dependent timing assumptions.

- add a gated streaming upstream test helper that emits the first SSE chunk immediately and only releases the second chunk when the test explicitly allows it
- update the direct streaming proxy integration test to verify the first chunk arrives before the second without relying on a 40ms race
- update the pipeline streaming proxy integration test to use the same deterministic harness

## Root cause

The macOS CI failure was in `tests::test_api_proxy_integration_streaming_response_arrives_incrementally`. The test used a `tokio::time::timeout(Duration::from_millis(40), ...)` probe to assert that no additional bytes arrived immediately after the first SSE chunk. On the hosted macOS runner that assumption is brittle because scheduler jitter and socket buffering can allow the delayed follow-up read to complete within that window even when proxy behavior is still correct.

## Impact

This keeps coverage on the streaming contract that matters: the proxy forwards the first chunk promptly and does not require buffering the whole response before sending data downstream. It also removes a flaky macOS-only failure mode from the unit-test job.

## Validation

- `npm ci && npm run build` in `mesh-llm/ui`
- `cargo test -p mesh-llm streaming_response_arrives_incrementally --release -- --nocapture`
- `cargo test --release`

### DIFF
--- a/mesh-llm/src/main.rs
+++ b/mesh-llm/src/main.rs
@@ -3203,6 +3203,73 @@ mod tests {
         (port, request_rx, handle)
     }
 
+    async fn spawn_gated_streaming_upstream(
+        content_type: &str,
+        first_chunk: Vec<u8>,
+        second_chunk: Vec<u8>,
+    ) -> (
+        u16,
+        oneshot::Receiver<Vec<u8>>,
+        oneshot::Sender<()>,
+        tokio::task::JoinHandle<()>,
+    ) {
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let port = listener.local_addr().unwrap().port();
+        let content_type = content_type.to_string();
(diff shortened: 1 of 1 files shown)


mesh-llm/src/main.rs
```diff
@@ -3203,6 +3203,73 @@ mod tests {
         (port, request_rx, handle)
     }
 
+    async fn spawn_gated_streaming_upstream(
```
michaelneale, line 3206: this doesn't seem to be used - only in tests? (or is that the idea - is testing infra? wonder if we should have main_test.rs if big enough now?)

i386: probably should split main.rs up - was planning to do that. I think it is a test func.

PR #101, opened 2026-03-31
[codex] Cache the Vite UI build

i386:
## Summary
- add a shared UI build helper that skips `vite build` when `mesh-llm/ui/dist` is already current
- avoid rerunning `npm ci` unless `node_modules` is missing or the package manifests changed
- wire the cache check into the macOS, Linux, Linux AMD, release, and Windows build scripts

## Why
Local `just build` paths were rebuilding the UI unconditionally even when the Vite output was already up to date, which added unnecessary frontend work to backend-focused rebuilds.

## Validation
- `bash -n scripts/build-ui.sh scripts/build-linux.sh scripts/build-linux-amd.sh scripts/build-release.sh`
- `zsh -n scripts/build-mac.sh`
- ran `scripts/build-ui.sh mesh-llm/ui` twice; first run built the UI, second run skipped with `dist` up to date
- Windows PowerShell syntax was not validated in this environment because `pwsh` is unavailable

### DIFF
--- a/scripts/build-windows.ps1
+++ b/scripts/build-windows.ps1
@@ -102,6 +102,85 @@ function Invoke-NativeCommand {
     }
 }
 
+function Test-UiBuildRequired {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$UiDirectory
+    )
+
+    $distDir = Join-Path $UiDirectory "dist"
+    if (-not (Test-Path $distDir)) {
+        return $true
+    }
+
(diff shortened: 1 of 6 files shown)


scripts/build-windows.ps1
```diff
@@ -102,6 +102,85 @@ function Invoke-NativeCommand {
     }
 }
 
+function Test-UiBuildRequired {
```
ndizazzo, line 105: its hilarious how much less readable powershell is compared to bash

i386: I don't know how windows users can live with themselves.

PR #106, opened 2026-03-31
fix: re-elect when peer RTT drops below split threshold (relay→direct)

michaelneale:
## Problem

When a peer first connects, the initial gossip goes via relay (RTT ~2600ms), exceeding the 80ms split threshold. The election skips the peer, split mode never activates. Once the direct path is established (RTT ~15ms), no re-election fires because `update_peer_rtt` didn't trigger `peer_change` events.

This means **split mode never works on first connection** — the peer is always skipped due to relay RTT, and the mesh falls back to solo serving or load balancing.

## Fix

Three changes to `mesh.rs`:

1. **`update_peer_rtt`**: fire `peer_change_tx` (triggering re-election) when RTT drops from above 80ms to below — the peer is now reachable for split
2. **`connect_to_peer`**: schedule a 5-second delayed RTT recheck via connection path info, catching the relay→direct transition
3. **`initiate_gossip_inner`**: also check connection path info after gossip round-trip, since the path RTT may already be lower than the gossip RTT

## Testing

- **Unit tests**: 3 new tests for the RTT re-election trigger (threshold crossing triggers, same-side doesn't, unknown peer doesn't panic)
- **312 tests pass** (full suite)
- **Live tested**: two-node Metal split (M4 Max 64GB + M4 16GB) with Qwen2.5-0.5B — tensor split 0.20,0.80 confirmed via `--rpc` and `--tensor-split` in llama-server process args

Log showing the fix in action:
```
📡 Peer 7337c73098 RTT: 15ms (direct) [path info]
📡 Peer 7337c73098 RTT improved (2907ms → 15ms) — re-electing for split
...
  ✓ Adding 7337c73098 — 12.9GB VRAM, RTT 15ms
  Tensor split: 0.20,0.80 (2 node(s), 64GB total)
✅ llama-server ready on internal port 61262
```

## Note

There is a separate issue with `iroh-quinn 0.16.1` panicking on connection drain (`drained connections always have an error`), which can abort mesh-llm during aggressive tunnel teardowns in split mode. That's tracked separately — upgrading to iroh 0.97.0 should fix it.

### DIFF
--- a/mesh-llm/src/main.rs
+++ b/mesh-llm/src/main.rs
@@ -307,7 +307,8 @@ async fn main() -> Result<()> {
             tracing_subscriber::EnvFilter::from_default_env()
                 .add_directive("mesh_inference=info".parse()?)
                 .add_directive("nostr_relay_pool=off".parse()?)
-                .add_directive("nostr_sdk=warn".parse()?),
+                .add_directive("nostr_sdk=warn".parse()?)
+                .add_directive("noq_proto::connection=warn".parse()?),
         )
         .with_writer(std::io::stderr)
         .init();


michaelneale:
Thanks James — two things here:

**1. `RETIRE_CONNECTION_ID for unknown path` errors**

This is a known noq-proto 0.16.0 bug (noq is iroh 0.97's quinn fork). It fires during connection reconnection/path migration when a `RETIRE_CONNECTION_ID` frame arrives for `PathId(0)` that's already been cleaned up. It's an error log, not a crash — cosmetic noise. We can suppress it with a tracing filter, or wait for noq-proto to fix it upstream. I'll add a `noq_proto=warn` tracing filter to suppress these.

**2. llama-server 600s timeout on build.local**

That looks like MoE shard loading taking too long, not related to the RTT changes in this PR. The RTT gate only applies to tensor-split mode (`--rpc` / `--tensor-split`), not MoE — MoE nodes run independent llama-server processes with their own shards and don't go through the RTT-gated worker selection path at all.

Is the Qwen3-Coder-Next model new / first time loading that shard? A cold shard build for a large model can be slow. After the retry it found a cached shard which is expected.

**Split itself** — can you confirm split is actually working for the 2-node case? Check `ps -eo args | grep llama-server | grep -v grep` on the host node — should show `--rpc` and `--tensor-split` in the args if tensor splitting activated.

michaelneale:
Actually, correction on my previous comment — let me be more precise about what's happening.

The MoE vs tensor-split decision in `election_loop` (line 294) is:

```rust
let need_moe_split = force_split || !model_fits_locally;
```

This is **per-node based on local VRAM**, not mesh-wide consensus. So:
- studio54 (big VRAM) → model fits locally → skips MoE → tensor-split election → serves full model solo
- build.local (small VRAM) → model doesn't fit → MoE election loop → tries to shard

They're in different election modes for the same model. This is a **pre-existing bug on main** — same code exists before this PR. Both nodes should agree on MoE vs tensor-split, but the decision is local.

This isn't caused by the RTT changes. Should I file a separate issue for the MoE/tensor-split consensus bug?

PR #131, opened 2026-04-02
router: context-aware host selection and retry on context overflow

IvGolovach:
## Problem

Routing does not account for a host's effective runtime context window. A long request can be sent to a smaller-context host and fail with a `400 Bad Request` even when another host in the mesh can handle it.

## Fix

- advertise the effective runtime `-c` chosen at launch via `ServedModelDescriptor.topology.context_length`
- update served descriptors as local inference processes start and stop so gossip reflects live capacity
- estimate request budget from the buffered JSON request plus completion budget
- prefer targets whose advertised context can fit the request
- probe the upstream HTTP response before committing bytes to the client and retry the next candidate only for context-overflow-style `400` responses
- keep generic `400` behavior unchanged and avoid masking real validation errors on single-target routes
- avoid learning sticky affinity toward a target that rejected the request for context overflow

## Testing

- `cargo test -q`
- Result: `354 passed; 0 failed`

## Notes

- `ModelTopology.context_length` is optional, so nodes that do not advertise it continue to route as before.


michaelneale:
hi @IvGolovach some feedback addressed here: https://github.com/michaelneale/mesh-llm/pull/141 - if that is clean and you are ok, we can merge that one in? (I didn't want to presumptively push to your fork branch without permission) - but if it looks ok, would be great to get it in for a release

PR #166, opened 2026-04-04
feat: auto-enable row split for local multi-GPU (tensor parallelism)

michaelneale:
## What

When a machine has multiple local GPUs (e.g. DGX with 8× H100, or a desktop with 2× RTX 4090), automatically pass `--split-mode row` to llama-server instead of the default layer split.

## Why

Row split (true tensor parallelism) shards weight **matrices** across GPUs — all GPUs compute on every token. This is faster than layer/pipeline split, where GPUs take turns processing layers sequentially.

llama.cpp already supports `--split-mode row` on CUDA and SYCL backends, but mesh-llm never passed the flag, so it always defaulted to `--split-mode layer` even on multi-GPU machines where row split would be faster.

## How

- Added `SplitMode` enum (`Layer` / `Row`) to `launch.rs`
- Added `split_mode: Option<SplitMode>` to `ModelLaunchSpec`
- Added `local_multi_gpu_split_mode()` helper in `election.rs` that checks:
  1. Binary flavor is CUDA or ROCm (backends that implement `ggml_backend_split_buffer_type`)
  2. `gpu_count > 1` (detected via `HardwareSurvey`)
- Applied to all solo launch paths (no RPC workers): normal solo, MoE solo, MoE shard
- RPC (network) splits are **not** affected — they continue to use pipeline parallelism since the RPC backend does not support row split

## Scope

- 3 files changed, 66 insertions
- No new CLI flags — detection is automatic
- No changes to network split, MoE split logic, or API
- All 366 existing tests pass

### DIFF
--- a/mesh-llm/src/inference/election.rs
+++ b/mesh-llm/src/inference/election.rs
@@ -18,19 +18,35 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use tokio::sync::watch;
 
+/// Returns `true` when `flavor` and `gpu_count` together call for row-split
+/// tensor parallelism.
+///
+/// Row split requires a backend that implements `ggml_backend_split_buffer_type`
+/// (CUDA and ROCm).  When no flavor is specified the binary may still be a CUDA
+/// or ROCm build discovered automatically, so `None` is treated as potentially
+/// supported; if the binary turns out to be CPU/Metal/Vulkan, llama.cpp falls
+/// back safely.
+fn should_use_row_split(flavor: Option<BinaryFlavor>, gpu_count: u8) -> bool {
+    let backend_supported = matches!(
+        flavor,
+        Some(BinaryFlavor::Cuda) | Some(BinaryFlavor::Rocm) | None
+    );
+    backend_supported && gpu_count > 1
+}
+
 /// Returns `Some(SplitMode::Row)` when the local machine has multiple GPUs and
 /// the llama.cpp backend supports row-level tensor parallelism (CUDA, ROCm).
 ///
 /// Row split shards weight matrices across local GPUs so all GPUs are active on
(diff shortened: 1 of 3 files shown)


mesh-llm/src/inference/election.rs
```diff
@@ -9,13 +9,38 @@ use crate::inference::{launch, moe};
...
+/// This does NOT work over RPC (network) — only for GPUs on the same machine.
+fn local_multi_gpu_split_mode(flavor: Option<BinaryFlavor>) -> Option<SplitMode> {
+    let supports_row = matches!(flavor, Some(BinaryFlavor::Cuda) | Some(BinaryFlavor::Rocm));
+    if !supports_row {
+        return None;
+    }
+    let hw = hardware::query(&[hardware::Metric::GpuCount]);
+    if hw.gpu_count > 1 {
+        tracing::info!(
+            "Local multi-GPU detected ({} GPUs) — using row split for tensor parallelism",
```
ndizazzo, line 51: Is there some world where we do addition checks for homogeneous GPUs? I did some testing a few weeks ago with my 5090 and a 3080, and decided I didn't want to mix models across them because of the performance loss across the PCI bus (also, different memory bus widths)

michaelneale: yeah I wasn't sure about this - do you have any opinions/thoughts before we get this in?

ndizazzo:
@michaelneale verified, sorry for the delay! Holiday up here in 🇨🇦 

Screenshot 2026-04-04 at 10 31 01 PM

image

mesh-llm/src/inference/election.rs
ndizazzo: Perhaps later when we have more of the stuff from #168 is place - that gives us PCIe device enumeration info (we can check for contiguous PCIe device numbers) info and could likely do more with correlating devices with the same name / model

PR #168, opened 2026-04-04
feat(config,system): add versioned mesh.toml schema, PR #130-aligned types, and GpuFacts consolidation

ndizazzo:
This PR adds support for the configuration types and scaffolds everything I had originally set up in #130 

It deliberately does not include any real functionality.


i386:
@ndizazzo yeah can do when I get home :)

PR #169, opened 2026-04-04
Build the new plugin system for MCP, HTTP, and inference endpoints

i386:
## Summary

`mesh-llm` can now build bundled and external plugins around one manifest-driven plugin system instead of adding more plugin-specific host code.

This branch makes the surface-first plugin DSL real for MCP, HTTP, inference, mesh channels, and mesh event subscriptions, ports the bundled plugins onto that model, and keeps host-private RPC separate from the surfaced plugin contract.

## What This Enables

After this branch, `mesh-llm` can:

- discover plugin-declared MCP tools, prompts, resources, and resource templates from plugin manifests
- mount plugin-declared HTTP bindings from the host instead of hand-writing plugin routes in core
- register external inference endpoints and route `/v1/*` traffic to them
- register external MCP endpoints and aggregate them into the host MCP server
- negotiate side streams for streamed stapled HTTP instead of forcing everything through the plugin control socket
- deliver mesh channel traffic and mesh events only when a plugin explicitly declares them
- keep host-private services such as blobstore on an explicit internal RPC path instead of pretending they are normal surfaced plugins

## What Changed

This branch now includes:

- manifest-driven MCP discovery, HTTP stapling, endpoint registration, and streamed side-channel plumbing
- a concrete `plugin!` DSL for surfaced plugins with:
  - `provides`
  - `mcp`
  - `http`
  - `inference`
  - `mesh`
  - `events`
  - `startup_policy`
  - `health`
  - `on_initialized`
  - `on_channel_message`
  - `on_mesh_event`
- host-side allowlisting of mesh channels and mesh events from plugin manifest declarations
- blackboard ported to the declarative plugin surface
- lemonade ported to the declarative plugin surface
- blobstore moved onto an explicit internal RPC plugin builder for host-private methods
- removal of the standalone example plugin crate instead of extending the DSL around demo-only runtime hooks
- updated plugin docs so `PLUGINS.md` matches the implemented DSL rather than a future-only sketch

## Architecture

The intended split on this branch is now:

- surfaced plugins use `plugin!` and declare MCP, HTTP, inference, mesh channels, and mesh event subscriptions
- lifecycle and runtime hooks stay local to the plugin definition
- host-private services use a lower-level internal RPC builder
- host-owned MCP and HTTP projections stay in core rather than being reimplemented per plugin

That keeps the normal plugin model small and explicit while preserving an escape hatch for internal plumbing.

## Protocol

This branch adds and starts consuming protocol support for:

- `PluginManifest`
- MCP manifest entries
- HTTP binding manifests
- endpoint manifests
- mesh channel manifests
- mesh event subscription manifests
- `OpenStreamRequest` / `OpenStreamResponse`
- `CancelStreamNotification`
- `CloseStreamNotification`
- `StreamError`

`PROTOCOL_VERSION` is bumped from 1 to 2 to reflect the addition of `InvokeService` — a new wire message that old SDK v1 plugins cannot handle. This ensures old plugins are rejected at handshake time rather than silently mishandling the new message type.

Protocol changes here are additive beyond the version bump. Older paths are still supported where migration is not complete.

## Validation

Validated locally with:

```bash
CARGO_BUILD_JOBS=1 just build
CARGO_BUILD_JOBS=1 cargo test --release --no-run
```

These checks covered the compile paths that were failing in GitHub Actions for PR `#169`.

## Follow-up

Recent fixes on top of the main plugin work:

- fix blobstore to use the current capability invocation path instead of the removed `call_tool_by_capability(...)` API
- mark the nostr key-rotation test serially to avoid a `$HOME` race with other tests touching global process state

### DIFF
--- a/mesh-llm/src/api/routes/plugins.rs
+++ b/mesh-llm/src/api/routes/plugins.rs
@@ -150,7 +150,7 @@ async fn handle_manifest(
             respond_error(stream, 404, "Plugin did not publish a manifest").await?;
         }
         Err(e) => {
-            respond_error(stream, 404, &e.to_string()).await?;
+            respond_error(stream, 500, &e.to_string()).await?;
         }
     }
     Ok(())
@@ -213,7 +213,7 @@ async fn handle_stapled_http(
             return Ok(());
         }
         Err(err) => {
-            respond_error(stream, 404, &err.to_string()).await?;
+            respond_error(stream, 500, &err.to_string()).await?;
             return Ok(());
         }
     };
@@ -369,9 +369,15 @@ fn rewrite_http_request_path(raw_request: &[u8], path: &str) -> anyhow::Result<V
 
     let method = req.method.unwrap_or("GET");
     let version = req.version.unwrap_or(1);
+    let original_path = req.path.unwrap_or("/");
+    let query = original_path
+        .find('?')
+        .map(|i| &original_path[i..])
+        .unwrap_or("");
     let mut rebuilt = format!(
-        "{method} {} HTTP/1.{version}\r\n",
-        normalized_http_path(path)
+        "{method} {}{} HTTP/1.{version}\r\n",
+        normalized_http_path(path),
+        query
     );
 
     for header in req.headers.iter() {
@@ -497,10 +503,18 @@ mod tests {
         let raw = b"POST /api/plugins/demo/http/feed?x=1 HTTP/1.1\r\nHost: localhost\r\nContent-Length: 7\r\nConnection: keep-alive\r\n\r\n{\"a\":1}";
         let rewritten = rewrite_http_request_path(raw, "/feed").unwrap();
         let text = String::from_utf8(rewritten).unwrap();
-        assert!(text.starts_with("POST /feed HTTP/1.1\r\n"));
+        assert!(text.starts_with("POST /feed?x=1 HTTP/1.1\r\n"));
         assert!(text.contains("Host: localhost\r\n"));
         assert!(text.contains("Connection: close\r\n"));
         assert!(text.ends_with("\r\n\r\n{\"a\":1}"));
+    }
+
+    #[test]
+    fn rewrite_http_request_path_without_query_string() {
+        let raw = b"GET /api/plugins/demo/http/items HTTP/1.1\r\nHost: localhost\r\n\r\n";
+        let rewritten = rewrite_http_request_path(raw, "/items").unwrap();
+        let text = String::from_utf8(rewritten).unwrap();
+        assert!(text.starts_with("GET /items HTTP/1.1\r\n"));
     }
 
     #[test]
--- a/mesh-llm/src/plugin/transport.rs
+++ b/mesh-llm/src/plugin/transport.rs
@@ -14,6 +14,8 @@ pub(crate) enum LocalStream {
     Unix(tokio::net::UnixStream),
     #[cfg(windows)]
     PipeServer(tokio::net::windows::named_pipe::NamedPipeServer),
+    #[cfg(windows)]
+    PipeClient(tokio::net::windows::named_pipe::NamedPipeClient),
 }
 
 pub(crate) enum LocalListener {
@@ -178,6 +180,8 @@ impl LocalStream {
             LocalStream::Unix(stream) => stream.write_all(bytes).await?,
             #[cfg(windows)]
             LocalStream::PipeServer(stream) => stream.write_all(bytes).await?,
+            #[cfg(windows)]
+            LocalStream::PipeClient(stream) => stream.write_all(bytes).await?,
         }
         Ok(())
     }
@@ -188,6 +192,8 @@ impl LocalStream {
             LocalStream::Unix(stream) => stream.shutdown().await?,
             #[cfg(windows)]
             LocalStream::PipeServer(stream) => stream.shutdown().await?,
+            #[cfg(windows)]
+            LocalStream::PipeClient(stream) => stream.shutdown().await?,
         }
         Ok(())
     }
@@ -198,6 +204,8 @@ impl LocalStream {
             LocalStream::Unix(stream) => stream.read(bytes).await?,
             #[cfg(windows)]
             LocalStream::PipeServer(stream) => stream.read(bytes).await?,
+            #[cfg(windows)]
+            LocalStream::PipeClient(stream) => stream.read(bytes).await?,
         };
         Ok(read)
     }
@@ -210,6 +218,10 @@ impl LocalStream {
             }
             #[cfg(windows)]
             LocalStream::PipeServer(stream) => {
+                let _ = stream.read_exact(bytes).await?;
+            }
+            #[cfg(windows)]
+            LocalStream::PipeClient(stream) => {
                 let _ = stream.read_exact(bytes).await?;
             }
         }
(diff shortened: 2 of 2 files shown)


mesh-llm/src/runtime/mod.rs
```diff
@@ -836,8 +835,7 @@ async fn run_auto(
     .await?;
     node.start_accepting();
     let token = node.invite_token();
-    node.set_blackboard_name(blackboard_display_name(&cli, &node))
-        .await;
+    node.set_display_name(node_display_name(&cli, &node)).await;
```
i386, line 838: @copilot  I don't think we really want blackboard to set a name of a node. Node identity should come from a node configuration not arbitrarily set by a plug-in.

mesh-llm/src/plugins/lemonade/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+
+fn lemonade_manifest(base_url: &str) -> mesh_llm_plugin::proto::PluginManifest {
+    mesh_llm_plugin::plugin_manifest![
+        capability("endpoint:inference"),
+        capability("endpoint:inference/openai_compatible"),
+        openai_http_inference_endpoint("lemonade", base_url).managed_by_plugin(false),
+    ]
+}
+
+fn build_lemonade_plugin(name: String) -> SimplePlugin {
```
i386: @copilot simple plug-in doesn't sound like the template solution that we described in plugins.MD. Why?

mesh-llm/src/protocol/mod.rs
```diff
@@ -38,9 +38,8 @@ pub const STREAM_TUNNEL_HTTP: u8 = 0x04;
 pub(crate) const STREAM_ROUTE_REQUEST: u8 = 0x05;
 pub(crate) const STREAM_PEER_DOWN: u8 = 0x06;
 pub(crate) const STREAM_PEER_LEAVING: u8 = 0x07;
-pub(crate) const STREAM_BLACKBOARD: u8 = 0x08;
-pub(crate) const STREAM_PLUGIN_CHANNEL: u8 = 0x09;
-pub(crate) const STREAM_PLUGIN_BULK_TRANSFER: u8 = 0x0a;
+pub(crate) const STREAM_PLUGIN_CHANNEL: u8 = 0x08;
+pub(crate) const STREAM_PLUGIN_BULK_TRANSFER: u8 = 0x09;
```
i386, line 42: @copilot Why do we even have a difference between a channel and a bulk transfer channel? I thought that we were just going to have channels that the plug-in can open and create etc. To do whatever it needed to do and we didn't really care because they had custom channels and a decent framing strategy. It seems like this is a hangover from the previous design?

mesh-llm/src/plugin/blobstore.rs
```diff
@@ -1,13 +1,20 @@
-use super::{PluginManager, BLOBSTORE_PLUGIN_ID};
```
i386, line 1: @copilot this needs to be its own module not stuffed into the plug-in module

PLUGINS_PLAN.md
```diff
@@ -0,0 +1,329 @@
...
+
+Build real plugins that exercise the design.
+
+The first plugin-hosted inference migration should be the current llama backend.
+
+After that, build an MLX endpoint provider plugin.
+
+After that, build at least one external MCP endpoint plugin.
+
+These plugins should validate:
```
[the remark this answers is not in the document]
i386: @copilot we no longer don't care about this at all

mesh-llm/src/api/routes/plugins.rs
```diff
@@ -89,3 +190,482 @@ async fn handle_call(
...
+        metadata_json: Some(
+            serde_json::json!({
+                "binding_id": binding.binding_id,
+                "method": method_name(binding.method),
+                "path": binding.path,
+            })
+            .to_string(),
+        ),
+        expected_bytes: Some(forwarded_request.len() as u64),
+        idle_timeout_ms: Some(30_000),
```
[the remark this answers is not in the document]
i386: @copilot I agree this needs to be fixed

mesh-llm/src/plugin/transport.rs
```diff
@@ -223,6 +243,32 @@ pub(crate) async fn bind_local_listener(instance_id: &str, name: &str) -> Result
...
+            tokio::net::UnixStream::connect(endpoint)
+                .await
+                .with_context(|| format!("Failed to connect side stream socket {endpoint}"))?,
+        )),
+        #[cfg(windows)]
+        proto::StreamTransportKind::StreamNamedPipe => Ok(LocalStream::PipeServer(
+            tokio::net::windows::named_pipe::ServerOptions::new()
+                .create(endpoint)
+                .with_context(|| format!("Failed to create side stream pipe {endpoint}"))?,
+        )),
```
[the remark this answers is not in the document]
i386: @copilot that sounds reasonable. Can you fix that?

mesh-llm/src/plugin/mcp.rs
```diff
@@ -12,28 +12,307 @@ use rmcp::{
...
+                    let client = tokio::net::windows::named_pipe::ClientOptions::new()
+                        .open(name)
+                        .with_context(|| {
+                            format!(
+                                "Connect named pipe MCP endpoint '{}:{}' at '{}'",
+                                endpoint.plugin_name, endpoint.endpoint_id, name
+                            )
+                        })?;
+                    ().serve(client).await.map_err(anyhow::Error::from)
+                }
```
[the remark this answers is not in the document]
i386: @copilot if you say so, let's fix that

mesh-llm/src/protocol/mod.rs
i386: @copilot is this code still here?

mesh-llm/plugin/src/runtime.rs
```diff
@@ -304,6 +354,63 @@ pub trait Plugin: Send {
         Ok(None)
     }
 
+    async fn invoke_service(
```
ndizazzo, line 357: @copilot does this InvokeService change mean that we should be bumping the PROTOCOL_VERSION so old clients don't fail silently?

mesh-llm/src/api/routes/plugins.rs
```diff
@@ -89,3 +190,497 @@ async fn handle_call(
...
+}
+
+async fn handle_streamed_http_binding(
+    client_stream: &mut TcpStream,
+    plugin_manager: &crate::plugin::PluginManager,
+    plugin_name: &str,
+    binding: &crate::plugin::proto::HttpBindingManifest,
+    raw_request: &[u8],
+) -> anyhow::Result<()> {
+    let forwarded_request = rewrite_http_request_path(raw_request, &binding.path)?;
```
ndizazzo, line 297: @i386 this silently doesn't include any params that might be coming in with the HTTP request

mesh-llm/src/plugin/runtime.rs
ndizazzo, line 231: @copilot what will happen here if a plugin defines duplicate tool/operation names in it's manifest, or use colliding HTTP bindings?

i386:
@copilot apply changes based on the comments in [this thread](https://github.com/michaelneale/mesh-llm/pull/169#pullrequestreview-4058934569)

PR #175, opened 2026-04-04
Resolve managed models from the Hugging Face cache only

i386:
## Summary

Users now get one consistent managed-model store: mesh-llm resolves managed models from the Hugging Face cache instead of the legacy flat `~/.models` directory.

Custom GGUF launches also gain an explicit `--mmproj` override, so vision models loaded with `--gguf` can provide their projector file directly instead of relying on catalog metadata or sibling-file auto-detection.

When no explicit `--mmproj` is provided, mesh-llm now resolves projector sidecars only from the selected GGUF's exact local directory using a multi-tier heuristic:

1. **Model-name-aware match** — prefer projectors whose filename contains the model's base name
2. **Quantization match** — when multiple named candidates exist, prefer the one whose filename contains the same quantization as the model (e.g. `Q4_K_M`), matching LM Studio's battle-tested heuristic
3. **Precision-variant fallback** — when all remaining candidates are the same projector at different precisions (`mmproj-BF16.gguf`, `mmproj-F16.gguf`, `mmproj-F32.gguf`), deterministically prefer `BF16`, then `F16`, then `F32`
4. **Return None** if still genuinely ambiguous, prompting the user to pass `--mmproj` explicitly

Quantization matching also applies when there are no model-name-aware candidates, so a directory containing generic siblings like `mmproj-Q4_K_M.gguf` and `mmproj-Q8_0.gguf` will automatically pick the right one.

The local smoke scripts also accept an optional mmproj path so multimodal local and compatibility smokes can launch a vision GGUF with an explicit projector.

## Why

The old `~/.models` fallback was still affecting runtime resolution, installed-model scans, and helper scripts even though the Hugging Face cache had become the intended managed store. That made model location rules harder to reason about and kept stale compatibility behavior alive in user-facing flows.

At the same time, users serving custom local vision GGUFs still needed a direct way to supply an `mmproj` file when auto-discovery was not enough.

Recent model layouts also exposed an additional ambiguity: some Hugging Face vision repos ship multiple projector siblings with different quantizations in the same snapshot directory. Quantization-aware selection resolves this common case deterministically without requiring an explicit `--mmproj`.

## User Impact

- Managed models are resolved from the Hugging Face cache rather than `~/.models`.
- `mesh-llm models installed` and related helper scripts now reflect that same managed-model location.
- Custom vision GGUFs can be launched with `--gguf /path/to/model.gguf --mmproj /path/to/mmproj.gguf`.
- Existing catalog-based and sibling-sidecar mmproj discovery still works when no explicit override is provided.
- Sibling-sidecar auto-discovery stays confined to the chosen GGUF's exact directory instead of scanning the wider cache.
- When multiple mmproj candidates share the model's name prefix, mesh-llm now prefers the one matching the model's quantization (e.g. `Q4_K_M`).
- When a snapshot only contains generic precision variants of the same projector, mesh-llm prefers `BF16`, then `F16`, then `F32`.
- Local smoke and SDK compatibility smoke scripts can now pass an optional mmproj path through to mesh-llm.

## Validation

- `cargo test -p mesh-llm models::local::tests::`
- `cargo test -p mesh-llm mmproj_path --lib`
- `bash -n scripts/ci-smoke-test.sh`
- `bash -n scripts/ci-compat-smoke.sh`

### DIFF
--- a/mesh-llm/src/models/local.rs
+++ b/mesh-llm/src/models/local.rs
@@ -407,8 +407,8 @@ fn strip_quant_suffix(stem: &str) -> &str {
 fn extract_quant_suffix(stem: &str) -> Option<String> {
     let stripped = strip_quant_suffix(stem);
     if stripped.len() < stem.len() {
-        // +1 to skip the '-' separator between base and quant
-        Some(stem[stripped.len() + 1..].to_string())
+        // +1 to skip the '-' separator; use .get() for safe UTF-8 slicing
+        stem.get(stripped.len() + 1..).map(|s| s.to_string())
     } else {
         None
     }
@@ -417,7 +417,7 @@ fn extract_quant_suffix(stem: &str) -> Option<String> {
 /// Return the sole candidate from `candidates` whose lowercased filename
 /// contains `quant`, or `None` if zero or multiple candidates match.
 fn pick_quant_match(candidates: &[PathBuf], quant: &str) -> Option<PathBuf> {
-    let matches: Vec<_> = candidates
+    let mut matches: Vec<_> = candidates
         .iter()
         .filter(|path| {
             path.file_stem()
@@ -428,7 +428,7 @@ fn pick_quant_match(candidates: &[PathBuf], quant: &str) -> Option<PathBuf> {
         .cloned()
         .collect();
     if matches.len() == 1 {
-        Some(matches.into_iter().next().unwrap())
+        matches.pop()
     } else {
         None
     }


mesh-llm/src/models/local.rs
```diff
@@ -480,6 +401,66 @@ fn strip_quant_suffix(stem: &str) -> &str {
...
+    let rank = match precision {
+        "bf16" => 0,
+        "f16" => 1,
+        "f32" => 2,
+        _ => return None,
+    };
+    Some((base, rank))
+}
+
+fn choose_mmproj_candidate(candidates: &[PathBuf]) -> Option<PathBuf> {
```
ndizazzo, line 474: This is out of scope, and I'm just thinking out loud... I wonder if we could do this for draft detection too?

ndizazzo (approved):
LGTM! No outstanding comments 👍🏻

i386:
@copilot let's do the quantization matching change here.

Issue #179, opened 2026-04-04
Native OS keychain backends for owner keystore


Bortlesboat:
Following up on @michaelneale's note in #156 ("I wonder if can use other keystores - in goose we use the native mac one etc") and the scope note in #178 — happy to pick this up.

Plan: wrap `keyring-rs` (macOS Keychain, Windows Credential Manager, libsecret/Secret Service on Linux, keyutils fallback) behind a small trait so the auth flow stays coherent.

One design question before I start: how should the keychain relate to the existing file keystore?

**Option A — keychain caches the passphrase, file keystore unchanged.** `auth init --keychain` generates a random passphrase, stores it in the OS keychain, and writes the existing encrypted JSON with that passphrase. `load_keystore` reads the passphrase from the keychain instead of prompting. Preserves the "Copy this keystore to other trusted nodes" flow that `auth init`'s next-steps output advertises — the file stays the portable identity, the keychain just unlocks it. Small change surface (essentially a new passphrase source in `cli/commands/auth.rs`), and the file stays encrypted at rest as defense-in-depth.

**Option B — keychain stores the secret keys directly, no file.** `auth init --backend keychain` puts the Ed25519/X25519 secrets straight in the keychain; metadata either lives in a small JSON or gets derived on load. Closer to what Goose does for API provider secrets, but breaks the copy-to-other-nodes workflow unless we add an `auth export` command, and it's a bigger divergence in the keystore code path.

Leaning Option A — the `ssh-agent` mental model (identity file on disk + unlock secret cached in OS keyring) fits the existing architecture, and respects the portability goal the CLI output advertises. Option B is defensible if you want to lean harder into the Goose pattern.

Either way: fallback to the passphrase keystore when no native backend is available (headless Linux, Secret Service not running, etc.) with a warning rather than a hard failure.

Happy to open the PR against main once we settle on direction.

Issue #135, opened 2026-04-02
Design owner identity propagation and direct verification


i386:
## Summary
This issue covers the next slice after the owner keystore and crypto foundation work: propagating owner identity through peer announcements and verifying it on direct peer communication.

It is a follow-on from:
- #132 - rollout tracker for splitting PR #130
- #134 - owner keystore and Rust signed+encrypted message primitives

Related issues:
- https://github.com/michaelneale/mesh-llm/issues/132
- https://github.com/michaelneale/mesh-llm/issues/134

## Goals
- Bind a running node identity to an owner identity.
- Propagate owner identity across mesh peer announcements.
- Verify direct ownership claims cryptographically.
- Distinguish direct verified ownership from transitive display-only ownership.
- Expose owner metadata in status for later read-only and authorization work.

## Non-goals
- No config signing or config authorization in this issue.
- No encrypted control-message usage in this issue.
- No config editing permissions in this issue.
- No console mutation flows in this issue.
- No group- or role-based authorization beyond same-owner identity metadata.

### Local node owner state
When mesh-llm starts:
- if a keystore is loaded successfully, the node has:
  - owner_id
  - owner signing public key
  - owner signing secret key available in memory for attestation signing
- if no keystore is present, the node runs in an unowned state

### Direct owner attestation
Each owned node should advertise a signed owner attestation that binds:
- the node network identity
- the owner identity
- the owner signing public key
- a versioned domain separator

Suggested attestation payload inputs:
- mesh-llm-owner-attestation-v1
- node identity or public id
- owner id
- owner signing public key

The attestation is signed with the owner signing secret key.

### Verification rules
If a peer receives a direct self-announcement from node P:
- verify the attestation against P and the claimed owner signing public key
- if valid:
  - mark owner_verified = true
  - mark owner_transitive = false
- if invalid:
  - discard the ownership claim
  - keep the peer connected unless another protocol error requires disconnect

If a peer learns owner metadata transitively through relay or gossip:
- preserve it for display if useful
- but mark:
  - owner_verified = false
  - owner_transitive = true
- never upgrade a transitive claim into trusted ownership without a direct verified attestation

### Trust rule
Matching owner ids alone are not enough.
Only a direct verified attestation creates trusted same-owner semantics.

## Protocol shape
This should be additive and backward-compatible.

Suggested fields in peer announcement or equivalent protocol struct:
- owner_id
- owner_sign_public_key
- owner_attestation
[...]

Issue #132, opened 2026-04-02
Track splitting PR #130 into smaller rollout PRs


i386:
## Summary
PR #130 bundles several distinct features into one large vertical slice: owner-key auth, protocol/status changes, config infrastructure, runtime scan plumbing, and a substantial config UI.

This issue tracks breaking that work into a staged sequence of smaller PRs that can land independently and be refined as we go.

## Reference
- Original PR: https://github.com/michaelneale/mesh-llm/pull/130

## Goals
- Reduce review surface area
- Land value incrementally
- Isolate protocol-sensitive changes
- Separate backend primitives from console UX
- Keep room to refine ordering and scope as implementation proceeds

### Completed
- [x] **1. Bugfix and cleanup** — independent fixes from #130 such as HF cache filename matching, remote detach cleanup, stale field cleanup, and unrelated TS/test fixes.  
  Done in #162.
- [x] **2. Owner-key primitives and auth CLI** — owner-key derivation, fingerprint utilities, `mesh-llm auth init`, `mesh-llm auth status`, and startup `--owner-key` wiring.  
  Done in #156.
- [x] **3. Owner fingerprint propagation and verification** — gossip/status fields for owner fingerprint and attestation, direct verification rules, and `/api/status` exposure.  
  Minimal read-only console/status surface. Protocol-sensitive; mixed-version behavior should be tested explicitly.  
  Done in #194.
- [x] **4. Authored config schema and local runtime consumption** — versioned mesh config, per-node assignments, split metadata, local runtime sync, and config file docs.  
  File/config semantics only.  
  Done in #193.
- [x] **5. Authorized config writes** — signed config checkpoint flow, conflict handling, validation errors, and broadcast/apply endpoints.  
  Save path exists, but can remain TOML-first initially.  
  Done in #194.

### Remaining
- [x] **6. Local model scan and status enrichment**
  - Scope: richer model scan metadata, model sizes, reserved VRAM reporting, and improved status payloads
  - UX: identify and supplement **existing** UI with any new data, but no net-new config editing or display yet
  - Note: zero communication protocol changes

- [x] **8. UI / UX implementation**
  - Augment the existing node sidebar where appropriate
  - Configuration UI stages:
    - [x] model catalog modal
    - [x] Generic UI playground page     
    - [x] TOML editor
    - [x] GPU blocks
    - [x] model info blocks
    - [x] VRAM / context visualization     
    - [x] assignment / unassignment
    - [x] node-level model placement

[...]

PR #180, opened 2026-04-05
crypto: native OS keychain backend for owner keystore unlock

Bortlesboat:
Implements the design proposed in #179 (Option A). Adds a `--keychain` flag to `mesh-llm auth init` that stores a random unlock passphrase in the OS-native credential store (macOS Keychain, Windows Credential Manager, Linux Secret Service via `keyring-rs` v3) instead of prompting.

The file keystore from #156 is unchanged — this is just a new source for the unlock passphrase. No format changes, no crypto changes, no migration path needed.

## What's new

- `mesh-llm auth init --keychain` generates a random 256-bit passphrase, stores it in the OS keychain, and uses it to encrypt the existing JSON keystore via the existing Argon2id + ChaCha20Poly1305 path.
- `mesh-llm auth status` auto-detects when a keystore can be unlocked via the keychain and reports the unlock source explicitly (plain / keychain / no-entry / backend-unavailable).
- `--keychain` and `--no-passphrase` are clap `conflicts_with` — can't mix.
- Each keystore path gets its own keychain slot: account name is `owner-keystore:<sha256(absolute_path)[:16]>`. Lowercased on Windows for NTFS case-insensitivity, preserved on Unix. Uses `std::path::absolute` rather than `canonicalize` so the account name is stable whether the file exists yet (init) or already exists (status/load).

## Rollback safety

`auth init --force --keychain` on an existing keystore overwrites any prior keychain entry at that slot. If the subsequent `save_keystore` call fails, the prior entry is restored (or a fresh entry is deleted), so a failed force-init can't strand an existing keystore with a lost unlock secret.

## Portability note (correct by design)

A keychain-unlocked keystore is bound to the machine whose keychain holds the passphrase. Next-steps text in `auth init --keychain` reflects this accurately — it tells the user the `(service, account)` pair to retrieve from, and suggests using a manual passphrase init if the same identity needs to unlock on multiple nodes without extracting the random secret.

## Tests

8 new tests:
- `init_with_keychain_then_load_round_trip` — end-to-end: init → verify keystore encrypted → verify keychain entry → load via keychain → cleanup.
- `force_keychain_save_failure_restores_previous_secret` — seeds a keychain entry, triggers a save failure, verifies the prior secret is restored.
- `fresh_keychain_save_failure_leaves_no_orphan` — save failure with no prior entry leaves the keychain clean.
- `account_name_is_case_insensitive_on_windows` / `account_name_is_case_sensitive_on_unix` — platform-appropriate path hashing.
- Plus 4 tests on the `crypto::keychain` wrapper (set/get/delete/overwrite) hitting the real backend.

All keychain-touching tests use `#[serial]` (from the existing `serial_test` dev-dep) since Windows Credential Manager doesn't handle concurrent access from the same process cleanly. 33 crypto+auth tests pass, zero regressions against the existing suite.

## Runtime scope

`load_keystore` is currently only called from `auth.rs` — the runtime doesn't consume the owner keystore yet. When it does (slice 3 in #132's plan, covered by #135), that loader can use the same `load_with_keychain` helper this PR introduces.

## Try it

```
mesh-llm auth init --keychain
# ... generates random passphrase, stashes in keychain, creates encrypted keystore
mesh-llm auth status
# Keystore:        valid (unlocked from OS keychain)
```

### DIFF
--- a/mesh-llm/src/cli/commands/auth.rs
+++ b/mesh-llm/src/cli/commands/auth.rs
@@ -1,15 +1,22 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use anyhow::{bail, Result};
+use base64::Engine as _;
+use sha2::{Digest, Sha256};
 use zeroize::Zeroizing;
 
 use crate::crypto::{
     default_keystore_path, keystore_exists, keystore_metadata, load_keystore, save_keystore,
-    OwnerKeypair,
+    OwnerKeypair, DEFAULT_OWNER_ACCOUNT, KEYCHAIN_SERVICE,
 };
 
 /// Run `mesh-llm auth init`.
-pub(crate) fn run_init(owner_key: Option<PathBuf>, force: bool, no_passphrase: bool) -> Result<()> {
+pub(crate) fn run_init(
+    owner_key: Option<PathBuf>,
+    force: bool,
+    no_passphrase: bool,
+    keychain: bool,
+) -> Result<()> {
     let path = match owner_key {
         Some(p) => p,
         None => default_keystore_path()?,
@@ -22,21 +29,42 @@ pub(crate) fn run_init(owner_key: Option<PathBuf>, force: bool, no_passphrase: b
         );
     }
 
-    let passphrase = if no_passphrase {
-        None
+    let (passphrase, source): (Option<Zeroizing<String>>, PassphraseSource) = if no_passphrase {
+        (None, PassphraseSource::None)
+    } else if keychain {
+        if !crate::crypto::keychain_available() {
+            bail!(
+                "No OS keychain backend is available on this host.\n\
+                 Retry without --keychain to set a passphrase, or with --no-passphrase \
+                 to store keys unencrypted."
+            );
+        }
+        let account = keychain_account_for_path(&path);
+        // Capture any existing keychain entry so we can roll back cleanly if
+        // the keystore write fails after we overwrite it.
+        let previous = crate::crypto::keychain_get(KEYCHAIN_SERVICE, &account)
+            .ok()
+            .flatten()
+            .map(Zeroizing::new);
+        let generated = Zeroizing::new(generate_random_passphrase());
+        crate::crypto::keychain_set(KEYCHAIN_SERVICE, &account, generated.as_str())?;
+        (
+            Some(generated),
+            PassphraseSource::Keychain { account, previous },
+        )
     } else {
         let pass = Zeroizing::new(rpassword::prompt_password_stderr(
             "Enter passphrase (empty for none): ",
         )?);
         if pass.is_empty() {
-            None
+            (None, PassphraseSource::None)
         } else {
             let confirm =
                 Zeroizing::new(rpassword::prompt_password_stderr("Confirm passphrase: ")?);
             if pass.as_str() != confirm.as_str() {
                 bail!("Passphrases do not match.");
             }
-            Some(pass)
+            (Some(pass), PassphraseSource::Prompt)
         }
     };
     let encrypted = passphrase.is_some();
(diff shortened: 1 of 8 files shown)


mesh-llm/src/cli/commands/auth.rs
```diff
@@ -22,21 +29,42 @@ pub(crate) fn run_init(owner_key: Option<PathBuf>, force: bool, no_passphrase: b
...
+                 Retry without --keychain to set a passphrase, or with --no-passphrase \
+                 to store keys unencrypted."
+            );
+        }
+        let account = keychain_account_for_path(&path);
+        // Capture any existing keychain entry so we can roll back cleanly if
+        // the keystore write fails after we overwrite it.
+        let previous = crate::crypto::keychain_get(KEYCHAIN_SERVICE, &account)
+            .ok()
+            .flatten()
```
ndizazzo, line 47: I think this may end up with a hard to catch error - if `keychain_get` fails, the `ok().flatten()` covers up that issue and it gets treated as "no previous entry". Then, rolling back below would cause loss of the old secret instead of restoring it.

mesh-llm/src/cli/commands/auth.rs
```diff
@@ -100,13 +165,298 @@ pub(crate) fn run_status(owner_key: Option<PathBuf>) -> Result<()> {
...
+fn keychain_account_for_path(path: &Path) -> String {
+    let absolute = std::path::absolute(path).unwrap_or_else(|_| path.to_path_buf());
+    let path_str = absolute.to_string_lossy();
+    #[cfg(windows)]
+    let normalized = path_str.to_lowercase();
+    #[cfg(not(windows))]
+    let normalized = path_str.into_owned();
+    let hash = Sha256::digest(normalized.as_bytes());
+    format!("{}:{}", DEFAULT_OWNER_ACCOUNT, &hex::encode(hash)[..16])
+}
```
ndizazzo, line 276: The PR description mentions runtime eventually wanting the same behavior... would it make sense to move `load_with_keychain` / `keychain_account_for_path` into `crypto/` now? 

This feels more like reusable domain logic than CLI orchestration, and it would keep auth.rs closerto the other command handlers.

mesh-llm/src/cli/commands/auth.rs
```diff
@@ -22,21 +29,42 @@ pub(crate) fn run_init(owner_key: Option<PathBuf>, force: bool, no_passphrase: b
...
+        }
+        let account = keychain_account_for_path(&path);
+        // Capture any existing keychain entry so we can roll back cleanly if
+        // the keystore write fails after we overwrite it.
+        let previous = crate::crypto::keychain_get(KEYCHAIN_SERVICE, &account)
+            .ok()
+            .flatten()
+            .map(Zeroizing::new);
+        let generated = Zeroizing::new(generate_random_passphrase());
+        crate::crypto::keychain_set(KEYCHAIN_SERVICE, &account, generated.as_str())?;
```
ndizazzo, line 50: This updates the keychain before `save_keystore(...)` returns successfully, so a crash/kill in between can leave the old keystore on disk while the keychain points at a different secret.

michaelneale:
yeah I don't have strong opionions - keychain for goose sometimes has been annoying with its popup for auth - would be my only concern (ie user input). feels like we never properly solved that on macos from a UX PoV - @Bortlesboat and no worries, so much is happening so fast that any input can help really, nice to see this attention to detail. 

I think strong identity if we start looking at trust and sharing of agents as well becomes super important (and obviously for people who are building privacy first meshes of their own!)

i386:
@Bortlesboat @ndizazzo thanks for your work on this. I am preparing a new branch based on the comments and feedback here, so we can get it in the release. https://github.com/michaelneale/mesh-llm/pull/187

Issue #184, opened 2026-04-05
Roll out config as local node config first, then reconcile mesh-wide desired state


i386:
## Summary

The current config work is still mixing two different problems:

1. local node startup configuration
2. mesh-wide desired state and editor reconciliation

This issue should stay focused on the first problem first.

The rollout should land explicit local `serve` / `client` commands together with GPU identity and inspection before expanding local config. Mesh-wide desired state can come later through editor-side reconciliation of per-node config rather than by making every node own a full-mesh TOML on disk.

## Problem

The main issue with the earlier `mesh.toml` direction is that it puts the authored file at the wrong level too early.

Operators mostly want to say:

- what should this node do on startup?
- should it serve or act as a client?
- what models should it load locally?
- what GPU should a model target once pinning exists?

They do not usually want every machine to carry a full mesh document describing every other machine.

That creates two problems immediately:

- the user-facing file is shaped around mesh-global state instead of local startup behavior
- it creates a distributed state problem before we have conflict-resolution, versioning, or reconciliation semantics

There is also a second issue: parts of the schema got ahead of the runtime surface. Things like `placement_mode` and `gpu_index` appeared before the CLI and runtime had a real local serve model that could own those concepts.

### Phase 1: CLI migration and GPU inspection (Competed https://github.com/michaelneale/mesh-llm/pull/185)

Add explicit local entrypoints and a local GPU inspection surface:

```bash
mesh-llm serve ...
mesh-llm client ...
mesh-llm gpus
```

Goals:

- make serving and client behavior explicit
- keep top-level `--model` / `--gguf` / `--client` working temporarily as compatibility shims
- warn on legacy top-level usage and print the preferred replacement command
- reject invalid mixed legacy usage such as top-level `--client` plus serving flags
- expose friendly local GPU inspection backed by stable GPU identity
- keep the new CLI surfaces visually consistent with existing mesh-llm output by using emoji-prefixed status lines

Examples:

```bash
mesh-llm --auto --model Qwen3-8B-Q4_K_M
# warns, then behaves like:
mesh-llm serve --auto --model Qwen3-8B-Q4_K_M

mesh-llm --auto --client
# warns, then behaves like:
mesh-llm client --auto
```

Example warning text:

```text
⚠️ top-level serving flags now map to `mesh-llm serve`.
  Please use: mesh-llm serve ...
```

```text
⚠️ top-level `--client` now maps to `mesh-llm client`.
  Please use: mesh-llm client ...
```

[...]

PR #193, opened 2026-04-06
feat(protocol): add distributed config protocol with ownership

ndizazzo:
Adds the wire protocol for mesh-wide node configuration, implementing the subscribe/push streams described in issue #184's Phase 4. This is part of the decomposed rollout of #130.

## What users can now do

Node operators using the same owner key across machines will have their nodes recognized as siblings in the mesh. The protocol infrastructure for reading and writing per-node config over the mesh is now in place, ready for the console and API layer to build on.

## What changed

**Proto schema** — two new control streams on the QUIC mesh connection:

- Stream `0x0b` — config subscribe: a node opens this stream to an owned peer and receives a config snapshot followed by live update notifications as the peer's config changes
- Stream `0x0c` — config push: a node sends a signed config update to an owned peer; the peer verifies the Ed25519 signature, checks the CAS revision, and applies or rejects the change

Six new proto message types (`ConfigSubscribe`, `ConfigSnapshotResponse`, `ConfigUpdateNotification`, `ConfigPush`, `ConfigPushResponse`, `NodeConfigSnapshot`) and three new `PeerAnnouncement` fields (`owner_id` on field 28; fields 29/30 reserved).

**Ownership and signing** — config pushes are signed with the node's Ed25519 owner key. Receivers verify the signature and the `owner_id` fingerprint before applying any change. A node will only serve config snapshots to a subscriber whose `owner_id` matches its own. Unowned nodes (no local owner key) reject both subscribe and push requests.

**CAS semantics** — pushes carry an `expected_revision`. A push is rejected with a `RevisionConflict` response if the node's current revision doesn't match, preventing lost updates when two editors push concurrently.

**Config persistence** (`ConfigState`) — atomic TOML write + revision sidecar file (named `<config>.revision` to avoid collisions when multiple configs share a directory), so config survives restarts and the revision is stable across process boundaries.

**`owner_id` in gossip** — each node advertises its ownership fingerprint in gossip so that consumers (the future console/API layer) can identify which peers belong to the same operator without probing every node.

**`config_revision`/`config_hash` kept off gossip** — these fields belong on the subscribe stream, which already delivers them authoritatively alongside the full config snapshot. Adding them to gossip would be dead weight with no consumer today; proto field numbers 29/30 are reserved for the future if a gossip-level "config unchanged" optimization is ever needed.

**Live vs. staged apply** — `ConfigPushResponse` carries `applied_live` and `saved_to_disk` flags. Currently all pushes write to disk (`saved_to_disk = true`) and `applied_live` is always `false` — changes take effect on next restart. The `applied_live` field is the designed hook for a future dynamic path where model load/unload is triggered immediately without a node restart.

**Protocol validation hardening** — `ConfigSnapshotResponse` and `ConfigPushResponse` validation now correctly allows error-shaped responses (with `error` set and empty structural fields) to pass, so callers can surface rejection reasons. `subscribe_to_config` surfaces `snapshot.error` as a proper `Err` rather than a generic validation failure.

## Protocol compatibility

- Peers running older builds that do not send `owner_id` are accepted with `owner_id = None`; they are treated as unowned nodes and are not eligible for config subscribe/push
- Proto unknown fields are ignored on receipt per standard protobuf rules
- `PeerAnnouncementV0` (JSON gossip path) carries `owner_id` with `serde(default)` so older peers can participate without it

## Architecture

The subscribe and push handlers live in `mesh/mod.rs` alongside the existing stream dispatch. Config state management is isolated in `runtime/config_state.rs`. Protocol types and conversion functions are in `protocol/`. No CLI, API endpoint, or UI changes are included in this PR.

## Validation

- 585 tests pass (13 new config protocol tests added)
- Integration tests cover: subscribe happy path, wrong-owner rejection, unowned-node rejection, valid signed push, revision conflict, bad signature rejection, and live update notification delivery
- Unit tests cover: Ed25519 sign/verify roundtrip on canonical payload, signature payload stability, protocol validation edge cases for error-shaped frames
- 2-node mesh tested on local macOS (Apple Silicon) + remote Linux/CUDA host
- Peer discovery, model routing, and inference unaffected

### DIFF
--- a/mesh-llm/proto/node.proto
+++ b/mesh-llm/proto/node.proto
@@ -37,7 +37,6 @@ message PeerAnnouncement {
   repeated ServedModelIdentity served_model_identities = 26;
   repeated ModelRuntimeDescriptor served_model_runtime = 27;
    optional string owner_id = 28;           // sha256(owner_signing_public_key), 64-char hex
-   reserved 29, 30;  // config_revision and config_hash — moved to subscribe stream only
 }
 
 message ServedModelDescriptor {
@@ -274,7 +273,5 @@ message ConfigPushResponse {
   uint64 current_revision = 3;
   bytes config_hash = 4;
   optional string error = 5;
-  reserved 6, 7;
-  reserved "saved_to_disk", "applied_live";
   ConfigApplyMode apply_mode = 8;
 }
--- a/mesh-llm/src/mesh/mod.rs
+++ b/mesh-llm/src/mesh/mod.rs
@@ -1683,11 +1683,11 @@ impl Node {
                 Arc::new(tx)
             },
             cached_owner_id: {
-                use crate::crypto::{default_keystore_path, load_keystore};
+                use crate::crypto::{default_keystore_path, keystore_metadata};
                 default_keystore_path()
                     .ok()
-                    .and_then(|p| load_keystore(&p, None).ok())
-                    .map(|kp| kp.owner_id())
+                    .and_then(|p| keystore_metadata(&p).ok())
+                    .map(|info| info.owner_id)
             },
         };
 
--- a/mesh-llm/src/protocol/mod.rs
+++ b/mesh-llm/src/protocol/mod.rs
@@ -1604,9 +1604,10 @@ mod tests {
 
     #[test]
     fn config_sync_v0_announcement_roundtrip() {
-        // Prove PeerAnnouncementV0 serde handles new fields gracefully:
-        // - owner_id and config_revision ARE serialized (use #[serde(default)])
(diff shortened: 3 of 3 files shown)


i386:
@copilot do new configurations being applied to nodes automatically restart the nodes? is that an option we can make e.g. stage (wait until restart) or apply changes (automatic) as we can load/unload models now. I sort of feel like this should be dynamic?

i386:
> @copilot do new configurations being applied to nodes automatically restart the nodes? is that an option we can make e.g. stage (wait until restart) or apply changes (automatic) as we can load/unload models now. I sort of feel like this should be dynamic?

@copilot please answer asap

PR #194, opened 2026-04-06
Enforce owner-attested config sync with trust policy and reliability fixes

i386:
Nodes now enforce owner-attested config sync with explicit trust policy controls, so only trusted peers can subscribe/push config and config updates stay consistent under retries and disconnects.

## What Changed

- enforce owner-attested authorization for distributed config subscribe/push flows
- remove legacy `owner_id` transport fields from config protocol messages (keeping wire-compat reserves)
- verify attestation signatures on load and regenerate invalid local attestations on startup
- add trust controls and CLI plumbing:
  - `--owner-required`
  - `--trust-policy off|prefer-owned|require-owned|allowlist`
  - `--trust-owner <owner_id>`
- propagate ownership/trust state into gossip and `/api/status`
- bound peer-rejection logging/state growth to avoid noisy/unbounded rejection tracking
- fix config sidecar write semantics so failed sidecar writes do not incorrectly advance write-hash state
- keep config-subscribe notification streams alive correctly (avoid premature unsubscribe on request-half EOF)
- remove duplicate config command ownership flags and use global CLI ownership options
- update TODO tracking for trust-policy follow-up

## Protocol

- config protocol messages no longer carry `owner_id` fields; authorization is based on owner attestation and trust policy
- compatibility-safe reserved fields are kept where required for wire evolution

## Validation

- `cargo test -p mesh-llm`
- targeted config subscribe/push authorization and notification tests
- protocol frame validation/conversion tests


i386:
@ndizazzo haven't tested yet

mesh-llm/proto/node.proto
```diff
@@ -36,7 +36,21 @@ message PeerAnnouncement {
   repeated ServedModelDescriptor served_model_descriptors = 25;
   repeated ServedModelIdentity served_model_identities = 26;
   repeated ModelRuntimeDescriptor served_model_runtime = 27;
-   optional string owner_id = 28;           // sha256(owner_signing_public_key), 64-char hex
+  optional SignedNodeOwnership owner_attestation = 28;
+  reserved 29, 30;  // config_revision and config_hash — moved to subscribe stream only
```
ndizazzo: My changes in #193 blew these away - we're clear to just get rid of them

mesh-llm/proto/node.proto
```diff
@@ -36,7 +36,21 @@ message PeerAnnouncement {
   repeated ServedModelDescriptor served_model_descriptors = 25;
   repeated ServedModelIdentity served_model_identities = 26;
   repeated ModelRuntimeDescriptor served_model_runtime = 27;
-   optional string owner_id = 28;           // sha256(owner_signing_public_key), 64-char hex
+  optional SignedNodeOwnership owner_attestation = 28;
```
ndizazzo, line 39: My `owner_id` stuff is still in main from #193 and should be replaced, I think nothing will be using it so we're clear to replace with the `SignedNodeOwnership`

mesh-llm/src/crypto/mod.rs
```diff
@@ -17,3 +18,10 @@ pub use self::keystore::{
     default_keystore_path, keystore_exists, keystore_metadata, load_keystore, save_keystore,
     KeystoreInfo,
 };
+pub use self::ownership::{
+    certificate_needs_renewal, default_node_ownership_path, default_trust_store_path,
```
ndizazzo, line 23: @i386 I'm not seeing where `certificate_needs_renewal` is called in this PR, but I'm assuming this will be used to decide when to re-mint the certificate for us when expired / invalid?

mesh-llm/src/cli/commands/mod.rs
```diff
@@ -100,7 +100,101 @@ pub(crate) async fn dispatch(cli: &Cli) -> Result<bool> {
                 no_passphrase,
                 keychain,
             } => auth::run_init(owner_key.clone(), *force, *no_passphrase, *keychain),
-            AuthCommand::Status { owner_key } => auth::run_status(owner_key.clone()),
+            AuthCommand::Status {
```
ndizazzo, line 102: I think we also need the flag plumbing for:

```
--owner-required
--trust-owner <owner_id>
--trust-policy off|prefer-owned|require-owned|allowlist
```

mesh-llm/src/mesh/mod.rs
```diff
@@ -4491,6 +4665,26 @@ impl Node {
 
     async fn add_peer(&self, id: EndpointId, addr: EndpointAddr, ann: &PeerAnnouncement) {
```
ndizazzo, line 4699: Just reading through this... verification failure rejects the peer, which is 👍🏻 

With peers running older builds, every gossip tick is going to re-classify them as unverified though.

We might want to implement more of a "last logged at" for this per peer with a longer interval, or even just logging only on status transitions rather than every tick. 

It's cheap to add and saves us from a noisy debugging experience the moment this ships. It probably won't be that big of a deal since it's early still, but I'd rather have something to bound the log output for being nice our future debugging selves

mesh-llm/src/crypto/ownership.rs
```diff
@@ -0,0 +1,598 @@
...
+
+pub fn default_node_ownership_path() -> Result<PathBuf, CryptoError> {
+    Ok(mesh_dir()?.join("node-ownership.json"))
+}
+
+pub fn default_trust_store_path() -> Result<PathBuf, CryptoError> {
+    Ok(mesh_dir()?.join("trusted-owners.json"))
+}
+
+pub fn load_node_ownership(path: &Path) -> Result<SignedNodeOwnership, CryptoError> {
```
ndizazzo, line 153: For this `save_node_ownership` and `load_node_ownership`, shouldn't we always sign on startup if the owner key is available? Otherwise we could potentially run into the issue of stale ownership on key rotation

Bortlesboat (commented):
Reviewed the crypto verification path and the config persistence layer.

mesh-llm/src/crypto/ownership.rs
```diff
@@ -0,0 +1,598 @@
...
+    let sig_bytes: [u8; 64] = match sig_bytes.try_into() {
+        Ok(value) => value,
+        Err(_) => {
+            summary.status = OwnershipStatus::InvalidSignature;
+            return summary;
+        }
+    };
+    let signature = ed25519_dalek::Signature::from_bytes(&sig_bytes);
+    if verifying_key.verify(&canonical, &signature).is_err() {
+        summary.status = OwnershipStatus::InvalidSignature;
```
Bortlesboat, line 374: This uses `verify` while the config push handler in `mesh/mod.rs` uses `verify_strict`. `verify_strict` rejects signatures with small-order public key components (cofactor malleability) — for a trust boundary like owner attestation it'd be safer to use `verify_strict` here too, so both paths apply the same acceptance criteria.

Minor since the signing key is self-generated and not externally sourced, but worth aligning for defense-in-depth.

mesh-llm/src/runtime/config_state.rs
```diff
@@ -0,0 +1,473 @@
...
+        .create(true)
+        .truncate(true)
+        .open(&tmp)?;
+    file.write_all(contents)?;
+    file.sync_all()?;
+    drop(file);
+    #[cfg(windows)]
+    if target.exists() {
+        std::fs::remove_file(target)?;
+    }
```
Bortlesboat, line 101: On Windows there's a gap between `remove_file` and `rename` where a crash leaves no config file at all. `std::fs::rename` can't atomically replace on Windows via the std API alone.

`MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` is the Windows equivalent of atomic rename-over, or the `tempfile` crate's `persist()` handles this cross-platform. Not urgent if Windows is a secondary target, but worth a `// TODO` at minimum so it doesn't get forgotten.

mesh-llm/src/crypto/ownership.rs
i386: @copilot please make this change

mesh-llm/src/runtime/config_state.rs
i386: @copilot improve based on this comment

mesh-llm/src/cli/commands/mod.rs
i386: @copilot can you do this please

mesh-llm/src/mesh/mod.rs
```diff
@@ -4109,7 +4682,35 @@ impl Node {
...
+            id.as_bytes(),
+            &trust_store,
+            self.trust_policy,
+            current_time_unix_ms(),
+        );
+        if !policy_accepts_peer(self.trust_policy, &owner_summary) {
+            let mut state = self.state.lock().await;
+            let last_status = state.policy_rejected_peers.get(&id).cloned();
+            if last_status.as_ref() != Some(&owner_summary.status) {
+                tracing::warn!(
```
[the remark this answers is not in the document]
i386: @copilot please fix this

mesh-llm/src/cli/mod.rs
```diff
@@ -30,6 +60,145 @@ pub(crate) enum AuthCommand {
         /// Path to the keystore file.
         #[arg(long)]
         owner_key: Option<PathBuf>,
+        /// Path to the node identity file (default: ~/.mesh-llm/key).
```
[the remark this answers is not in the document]
i386: @copilot fix by relying own parent flag

mesh-llm/src/cli/mod.rs
```diff
@@ -30,6 +60,145 @@ pub(crate) enum AuthCommand {
...
+        #[arg(long)]
+        node_key: Option<PathBuf>,
+        /// Output path for the signed node certificate.
+        #[arg(long)]
+        out: Option<PathBuf>,
+        /// Optional label attached to the signed node certificate.
+        #[arg(long)]
+        node_label: Option<String>,
+        /// Optional hostname hint attached to the certificate.
```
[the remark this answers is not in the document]
i386: @copilot remove inheritance

ndizazzo (approved):

ndizazzo:
I hate that copilot renames PRs. Is there any way to turn that off @michaelneale? It's totally stomping the meaning of PRs by doing that

mesh-llm/src/mesh/mod.rs
```diff
@@ -3770,13 +3896,460 @@ impl Node {
...
+                error: None,
+            }
+        };
+        write_len_prefixed(&mut send, &snapshot.encode_to_vec()).await?;
+
+        let mut rev_rx = self.config_revision_tx.subscribe();
+        loop {
+            if rev_rx.changed().await.is_err() {
+                break;
+            }
```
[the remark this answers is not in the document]
i386: Fixed in a564aa9d. `handle_config_subscribe` now uses `tokio::select!` to wait on both config revision updates and inbound stream activity/closure, so disconnected subscribers do not leave parked tasks.

mesh-llm/src/crypto/ownership.rs
i386: Addressed. Verification now uses verify_strict on this path too, aligned with config push verification.

mesh-llm/src/runtime/config_state.rs
i386: Addressed via the split apply-result path for revision-sidecar failures.

mesh-llm/src/mesh/mod.rs
```diff
@@ -1618,6 +1706,42 @@ impl Node {
...
+                    .and_then(|config| config.node_label.clone()),
+                hostname.clone(),
+            )?),
+            None => None,
+        };
+        let owner_summary = verify_node_ownership(
+            owner_attestation.as_ref(),
+            endpoint.id().as_bytes(),
+            &trust_store,
+            trust_policy,
```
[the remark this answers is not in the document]
i386: Implemented. Local owner summary verification uses TrustPolicy Off; trust policy gating is for peer admission.

mesh-llm/src/mesh/mod.rs
```diff
@@ -9311,150 +9541,52 @@ pub fn clear_public_identity() {
...
+        return Ok(key);
+    }
+
+    let dir = key_path
+        .parent()
+        .ok_or_else(|| anyhow::anyhow!("Invalid node key path {}", key_path.display()))?;
+    let key = SecretKey::generate(&mut rand::rng());
+    tokio::fs::create_dir_all(&dir).await?;
+    tokio::fs::write(&key_path, hex::encode(key.to_bytes())).await?;
+    tracing::info!("Generated new key, saved to {}", key_path.display());
```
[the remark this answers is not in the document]
i386: Implemented in commit 2f68f6a6. This path now uses hardened atomic key persistence and regular file checks.

mesh-llm/proto/node.proto
```diff
@@ -226,13 +239,15 @@ message NodePluginEntry {
 message ConfigSubscribe {
   uint32 gen = 1;                         // must equal NODE_PROTOCOL_GENERATION
   bytes subscriber_id = 2;               // 32 bytes — subscribing node's endpoint id
-  string owner_id = 3;                   // filter: only configs from nodes with this owner_id
+  reserved 3;
+  reserved "owner_id";                   // removed: authorization now uses attested peer identity
```
ndizazzo: @i386 Nothing in the protocol was using `owner_id` previously, by remove it, I think we can literally erase it from existence instead of keeping it as reserved?

mesh-llm/src/protocol/convert.rs
```diff
@@ -5,6 +5,69 @@ use crate::protocol::NODE_PROTOCOL_GENERATION;
...
+            tracing::warn!(
+                "dropping local owner attestation from gossip: invalid signature hex: {err}"
+            );
+            return None;
+        }
+    };
+    Some(crate::proto::node::SignedNodeOwnership {
+        version: attestation.claim.version,
+        cert_id: attestation.claim.cert_id.clone(),
+        owner_id: attestation.claim.owner_id.clone(),
```
ndizazzo, line 41: 👋🏻 if we nuke this reservation, do we still need to populate it from the attestation?

mesh-llm/src/runtime/config_state.rs
```diff
@@ -4,29 +4,21 @@ use std::path::{Path, PathBuf};
...
-    Noop,
-}
-
 #[derive(Debug)]
 pub(crate) enum ApplyResult {
     Applied {
         revision: u64,
         hash: [u8; 32],
-        apply_mode: ConfigApplyMode,
+        saved_to_disk: bool,
```
ndizazzo: I thought we got rid of this in favour of the enumeration? Did this regress? 

`ConfigApplyMode` was stronger

ndizazzo (changes requested):
I noticed some things - mainly stuff we chatted about previously on owner_id and `save_to_disk` not being a good way to track state, and moving towards a `ConfigApplyMode` instead

ndizazzo (approved):

i386:
@copilot apply changes based on the comments in [this thread](https://github.com/michaelneale/mesh-llm/pull/194#pullrequestreview-4087093736)

PR #195, opened 2026-04-06
Make model migration safer and easier to debug

i386:
Users can now see exactly why `mesh-llm models migrate --apply` adopts, skips, or re-downloads legacy GGUF files, including partial split sets and ambiguous duplicate basenames under `~/.models`.

## What changed

- detect partial split GGUF sets up front and report missing parts before downloads begin
- summarize each migrated model group with adopted files, downloaded files, ambiguous files, and split coverage
- stop silently choosing between duplicate legacy files with the same basename; warn and skip automatic adoption for ambiguous assets
- use explicit Hugging Face LFS SHA-256 metadata as the checksum source for adoption
- if the current upstream revision no longer matches a local file, search recent repo history for a historical checksum match before re-downloading
- warn when a model mixes flat `~/.models` files with nested legacy subdirectories

## Why

Migration was difficult to trust for very large split models because a single mismatched part could trigger a large re-download without saying why. It also silently collapsed duplicate basenames to one path, which made behavior brittle when users had both flat and nested legacy copies of the same file.

## Root Cause

The old migration flow keyed legacy candidates only by basename and verified files against the current upstream revision without enough diagnostic output. That meant duplicate local paths were resolved implicitly, and legitimate older local files could be treated as mismatches with no clear explanation.

## Validation

- `cargo test -p mesh-llm maintenance --lib`
- `cargo check -p mesh-llm`

### DIFF
--- a/mesh-llm/src/models/maintenance.rs
+++ b/mesh-llm/src/models/maintenance.rs
@@ -978,7 +978,11 @@ fn fetch_recent_repo_commits(
     let endpoint = hf_endpoint();
     let token = hf_token_override();
     let url = commits_api_url(&endpoint, repo_id, revision, limit);
-    let client = reqwest::blocking::Client::new();
+    let client = reqwest::blocking::Client::builder()
+        .connect_timeout(std::time::Duration::from_secs(30))
+        .timeout(std::time::Duration::from_secs(60))
+        .build()
+        .context("Build HTTP client")?;
     let mut request = client.get(url);
     if let Some(token) = token {
         request = request.bearer_auth(token);
@@ -1025,10 +1029,11 @@ fn link_or_copy_file(src: &Path, dst: &Path) -> Result<()> {
     }
 }
 
+static SPLIT_GGUF_RE: std::sync::LazyLock<regex_lite::Regex> =
+    std::sync::LazyLock::new(|| regex_lite::Regex::new(r"-\d{5}-of-\d{5}\.gguf$").unwrap());
+
 fn is_split_gguf_file(file: &str) -> bool {
-    regex_lite::Regex::new(r"-\d{5}-of-\d{5}\.gguf$")
-        .unwrap()
-        .is_match(file)
+    SPLIT_GGUF_RE.is_match(file)
 }
 
 fn expected_split_gguf_files(model: &catalog::CatalogModel) -> Vec<String> {


mesh-llm/src/models/maintenance.rs
```diff
@@ -475,73 +778,214 @@ fn adopt_legacy_asset_into_hf_cache(
...
+    repo_id: &str,
+    revision: &str,
+    limit: usize,
+) -> Result<Vec<RepoCommit>> {
+    let endpoint = hf_endpoint();
+    let token = hf_token_override();
+    let url = format!(
+        "{}/api/models/{}/commits/{}?limit={limit}",
+        endpoint,
+        urlencoding::encode(repo_id),
```
ndizazzo: Wouldn't `owner/model-name`  encode a %2F here? maybe we should add a test to ensure this case doesn't happen

ndizazzo (approved):

mesh-llm/src/models/maintenance.rs
i386: @copilot could you check against huggingface python for encoding rules

i386: Great catch @ndizazzo

mesh-llm/src/models/maintenance.rs
```diff
@@ -361,11 +461,57 @@ fn migrate_catalog_model(
...
-    let legacy_files: BTreeMap<String, &PathBuf> = entries
+    let mut adopted_files = Vec::new();
+    let mut historical_files = Vec::new();
+    let mut downloaded_files = Vec::new();
+    let mut ambiguous_files = Vec::new();
+    let mut split_downloads = Vec::new();
+    let legacy_root = legacy_models_dir();
+    let legacy_files = collect_legacy_candidates(entries, &legacy_root);
+    let duplicate_legacy_files = duplicate_legacy_files(&legacy_files);
+    let expected_split_files = expected_split_gguf_files(model);
```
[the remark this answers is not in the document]
i386: @copilot apply changes based on [this feedback](https://github.com/michaelneale/mesh-llm/pull/195#discussion_r3039273610)

i386:
@copilot apply changes based on the comments in [this thread](https://github.com/michaelneale/mesh-llm/pull/195#pullrequestreview-4064692679)

PR #202, opened 2026-04-07
fix: prevent detail panel from being pushed when in fullscreen mode

ndizazzo:
I was playing around with the UI visualization and noticed a problem when selecting a node when in fullscreen mode. 

The fix is pretty straightforward, It's to prevent pushing the detail panel into view when the react-flow node visualization is already fullscreen.


i386 (approved):
LGTM

PR #209, opened 2026-04-07
ci: reorganize mesh-llm builds for 50% speedups using warm build caches

ndizazzo:
## Summary

- **Removes a duplicate ~7 minute release build** from the Linux and macOS CI jobs. They used to run `cargo test --release` and then `cargo build --release -p mesh-llm --bin mesh-llm` back-to-back. The second step is redundant: `cargo test --release` already compiles `target/release/mesh-llm` as part of building the workspace, and every downstream integration test (`ci-smoke-test.sh`, `ci-split-test.sh`, `ci-moe-mesh-test.sh`, `ci-client-auto-test.sh`, CLI smoke) already consumes that same binary.
- **Moves `cargo fmt --check` to an early-fail gate**, before unit tests and the llama.cpp build. Formatting issues now fail the job in seconds instead of ~15 minutes in. The `|| true` suffix is also removed so the check actually blocks.
- **Fixes pre-existing rustfmt violations** that had been silently masked by `|| true`, in a separate style-only commit. Surfacing these is the point of making the fmt check blocking.

## Why the duplicate build is safe to remove

`cargo test --release` at the workspace root compiles every `[[bin]]` target of every workspace member in release mode (it has to, in order to build and link the test harnesses against the same crate). After that step, `target/release/mesh-llm` exists and is byte-for-byte the binary the subsequent tests were already using. The old `cargo build --release -p mesh-llm --bin mesh-llm` step was a full second-compile pass that accomplished nothing new.

## Rust caching — findings (no change in this PR)

The user asked whether there is an sccache-equivalent for Rust that we could adopt. Current state:

- `Swatinem/rust-cache@v2` **is** wired up and caches the Rust `target/` dir + registry. This is the primary Rust cache today.
- `mozilla-actions/sccache-action@v0.0.9` is installed and `SCCACHE_GHA_ENABLED: "true"` is set, but `RUSTC_WRAPPER: sccache` is **not** set, and `CARGO_INCREMENTAL` is not pinned. So sccache currently only accelerates the llama.cpp C/C++ build via `-DCMAKE_C_COMPILER_LAUNCHER=sccache`. It does **not** cache any Rust compilation.

Enabling sccache for Rust is possible (`RUSTC_WRAPPER: sccache` + `CARGO_INCREMENTAL: 0`) but interacts non-trivially with Swatinem/rust-cache — they cache at different layers and combining them can either compound or fight each other depending on the workload. I did not change this in this PR because (a) it is risky and should be measured, and (b) this PR's goal is to eliminate an obvious 7 minute duplicate. Worth a follow-up PR that A/B tests the two caching strategies.

## Validation

Locally on this branch:

\`\`\`
\$ cargo fmt --manifest-path mesh-llm/Cargo.toml -- --check
(clean)
\`\`\`

The four files touched in \`83907bf\` are pure whitespace/line-wrapping changes produced by \`cargo fmt\` and contain no behavioral changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,20 +94,34 @@ jobs:
 
       - uses: mozilla-actions/sccache-action@v0.0.9
 
+      # Workspace `. -> target` because the cargo workspace is at repo root
+      # (`Cargo.toml` members = ["mesh-llm", ...]) and cargo writes to
+      # `./target/`, NOT `mesh-llm/target/`. Using `workspaces: mesh-llm`
+      # silently caches an empty dir and gives zero PR cache hits — we
+      # measured a 4-minute regression from this exact misconfiguration.
       - uses: Swatinem/rust-cache@v2
         with:
-          workspaces: mesh-llm
+          workspaces: . -> target
           prefix-key: ${{ env.CACHE_NAMESPACE }}-rust-${{ hashFiles('.github/cache-version.txt') }}
 
-      - name: Unit tests
-        run: cargo test --release
-
-      - name: Build mesh-llm
-        run: cargo build --release -p mesh-llm --bin mesh-llm
-
+      # Early-fail gate: runs before expensive compile/test work.
       - name: Check formatting
         working-directory: mesh-llm
-        run: cargo fmt -- --check || true
+        run: cargo fmt -- --check
+
+      # Build in dev/debug profile, NOT release. Two reasons:
+      #  1. Build + Unit tests steps share one target/ subdir and one
+      #     incremental cache — zero duplicated codegen.
+      #  2. debug has `incremental = true`; release has `incremental = false`.
+      #     Warm rebuild after a lib edit: release ~43s, debug ~4s (M4 Pro).
+      # mesh-llm is a thin orchestrator around llama-server; the hot loop is
+      # inside llama-server C++, so debug vs release binary perf is
+      # negligible for smoke tests. Integration steps below use target/debug.
+      - name: Build mesh-llm binary (debug)
+        run: cargo build -p mesh-llm --bin mesh-llm
+
+      - name: Unit tests
+        run: cargo test --lib --tests
 
       - name: Clippy
         working-directory: mesh-llm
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -94,20 +94,34 @@ jobs:
...
+      #  2. debug has `incremental = true`; release has `incremental = false`.
+      #     Warm rebuild after a lib edit: release ~43s, debug ~4s (M4 Pro).
+      # mesh-llm is a thin orchestrator around llama-server; the hot loop is
+      # inside llama-server C++, so debug vs release binary perf is
+      # negligible for smoke tests. Integration steps below use target/debug.
+      - name: Build mesh-llm binary (debug)
+        run: cargo build -p mesh-llm --bin mesh-llm
+
+      - name: Unit tests
+        run: cargo test -p mesh-llm --lib --tests
```
Bortlesboat, line 125: This drops the `mesh-llm/plugin` package tests — `manifest.rs` and `runtime.rs` both have test blocks that were previously covered by the workspace-wide `cargo test --release`. Worth adding `-p mesh-llm-plugin` or dropping the `-p` flag.

Bortlesboat (commented):
The cache workspace fix and format-check changes are correct. One thing to check: the  crate has tests in  and  that were previously covered by the workspace-wide . The new  flag drops those.

michaelneale:
@ndizazzo any risks - in the past I have been bitten by testing debug vs release but as long as release bin is tested at least once?

.github/workflows/ci.yml
ndizazzo: Good catch!

ndizazzo:
> @ndizazzo any risks - in the past I have been bitten by testing debug vs release but as long as release bin is tested at least once?

@michaelneale Low risk, but not zero. PR CI intentionally uses debug for speed + incremental, and I think we're covered via the release workflow, where release-profile binaries are built and validated. 

As long as we still exercise the release binary in the release path at least once, I’m fairly comfortable with this change. 

I'm also thinking that our docker builds can run on main with a release path, so that's another angle we cover (once #220 gets finalized)

ndizazzo:
@michaelneale making an executive decision to get this in. I was blocked on MLX I didn't want to conflict with that work, but we're losing the efficiency this brings. I'll pull changes into the MLX branch and update for compatibility.

PR #216, opened 2026-04-08
Harden GGUF parser, identity key I/O, and symlink safety

michaelneale:
Addresses five security review findings from the GGUF/identity hardening PR.

## Changes

- **UTF-8 rejection in GGUF strings** — `read_gguf_string` now uses `String::from_utf8` instead of `from_utf8_lossy`; invalid UTF-8 returns `InvalidData` rather than silently substituting replacement characters that could bypass key matching.

- **Non-blocking key I/O** — `load_or_create_key_at` wraps all filesystem work (dir creation, permission hardening, file read, atomic write) in `tokio::task::spawn_blocking` to avoid stalling the Tokio executor.

- **Symlink attack prevention** — `ensure_private_identity_file` and `ensure_private_nostr_key_file` now use `symlink_metadata` and reject non-regular files before adjusting permissions or reading key material, preventing `chmod` through a symlink to an unintended target.

- **Generic error messages** — Symlink rejection errors no longer include the exact file path to avoid leaking filesystem layout in security-critical paths.

- **Regression tests** — Symlink rejection is covered by new tests in both the mesh identity and Nostr key loading paths.


i386:
@copilot apply changes based on the comments in [this thread](https://github.com/michaelneale/mesh-llm/pull/216#pullrequestreview-4073434176)

ndizazzo (approved):
LGTM!

PR #217, opened 2026-04-08
runtime: introduce scoped instance runtime with safe process lifecycle management

### LINKED ISSUE #174: Scope child-process cleanup to the owning mesh-llm instance and surface mixed-version state
## Problem

`mesh-llm` currently treats inference child-process cleanup as machine-global in some places, while ownership is really per `mesh-llm` instance.

Two related concerns came up during local testing and code review:

1. Startup cleanup is broader than the instance that launched the children.
2. The current API/UI does not make mixed-version or multi-process situations obvious.

This matters if a machine has:

- multiple `mesh-llm` worktrees
- multiple installed versions
- one manually started `llama-server`
- one stale/orphaned child from a previous crash
- more than one local `mesh-llm` process serving different things

Today, a fresh non-client startup calls global cleanup for `llama-server`, and orphan cleanup for `rpc-server` is handled separately. That means we are mixing two concepts:

- `orphan cleanup`: remove children whose owner is gone
- `global cleanup`: kill anything with a matching process name

Those are not equivalent.

## Current behavior

From code inspection:

- startup calls `kill_llama_server()` and `kill_orphan_rpc_servers()`
- `kill_llama_server()` kills by process name
- `kill_orphan_rpc_servers()` is narrower and only targets PPID-1 orphaned `rpc-server` processes
- peer version is gossiped internally, but not surfaced clearly in `/api/status` peer payloads
- local runtime processes are visible through `/api/runtime/processes`, but that only helps once you are attached to a particular local console instance

In practice, this creates operational ambiguity:

- if two local `mesh-llm` instances exist, one startup may interfere with the other
- if multiple versions are running in one mesh, operators cannot easily see that from the main status API
- if an orphaned `llama-server` exists, cleanup policy is broader than necessary

## Repro / observations

I was not able to reproduce a leak on clean shutdown:

- `Ctrl-C` cleanly removed parent + `rpc-server` + `llama-server`
- `SIGTERM` to the parent also cleanly removed parent + children

So this is **not** a confirmed clean-shutdown leak bug.

However, the underlying ownership model is still weak because cleanup is not consistently scoped. The remaining risk is primarily around:

- hard-killed parent processes
- stale orphaned children from abnormal exits
- multiple local `mesh-llm` instances or versions on the same host

## Desired outcome

Make cleanup and visibility instance-aware:

- only kill child processes that are provably owned by the relevant `mesh-llm` instance
- clean up orphaned `llama-server` processes with the same level of care already applied to `rpc-server`
- make it obvious when multiple local processes or multiple mesh versions are present

### Option A: Keep global name-based cleanup

Example:

- kill all `llama-server`
- kill all `rpc-server`

Strengths:

- simplest to implement
- easy to reason about in single-instance development
- masks many stale-process problems quickly

Weaknesses:

- unsafe when multiple `mesh-llm` instances or versions share a machine
- can kill manually started `llama-server` processes
- does not model ownership, only naming
- gets riskier as more local workflows are added

### Option B: Scoped orphan cleanup by ownership markers

When spawning `rpc-server` / `llama-server`, tag them with ownership metadata such as:

- parent pid
- owning `mesh-llm` executable path
- runtime/state directory identifier
- managed-by flag

On startup, only reap processes that:

- are marked as managed by `mesh-llm`
- belong to this instance scope
- are orphaned or otherwise stale

Strengths:

- preserves isolation between worktrees / versions / installs
- avoids killing unrelated manual processes
- directly models ownership
- safest long-term default

Weaknesses:

- more implementation complexity
- process inspection is platform-specific
- env-based ownership discovery may be awkward on some platforms

### Option C: Scoped cleanup via pidfiles in the runtime/state dir

When `mesh-llm` launches a child, write pidfiles under its runtime dir. On restart:

- read pidfiles
- verify pid still matches expected process/command
- kill only those stale children
- remove stale pidfiles

Strengths:

- concrete ownership record
- simpler operational story than broad process scans
- works well with per-instance runtime directories

Weaknesses:

- pid reuse requires validation beyond pid alone
- stale pidfiles need careful handling
- still benefits from process metadata checks for safety

### Option D: Hybrid approach (recommended)

Combine scoped ownership with pidfiles:

1. tag managed children on spawn
2. write pidfiles in the owning runtime dir
3. replace global `kill_llama_server()` startup cleanup with scoped orphan cleanup
4. add equivalent orphan cleanup for `llama-server`
5. keep explicit direct shutdown for owned children during normal exit

Strengths:

- strongest safety story
- works for both normal shutdown and abnormal orphan recovery
- supports multiple versions/worktrees on one machine
- avoids collateral damage

Weaknesses:

- more moving parts than the current approach
- requires careful cross-platform implementation

## Visibility improvements

Independent of cleanup, operators should be able to see when the system is in a potentially confusing state.

Suggested additions:

- include peer `version` in `/api/status` peer payloads and console UI
- optionally show local parent/child process ownership in `/api/runtime/processes`
- surface a warning when multiple distinct `mesh-llm` versions are present in one mesh
- surface a warning when multiple local `mesh-llm` instances appear to share the same machine/runtime namespace

## Recommendation

Implement **Option D (hybrid scoped cleanup)**.

That gives us:

- safe orphan cleanup for both `rpc-server` and `llama-server`
- no machine-global process-name kill on startup
- support for multiple local versions/worktrees
- clearer observability for mixed-version meshes

## Open questions

- What should define ownership scope: executable path, runtime dir, mesh id, or some combination?
- Do we want to support multiple local `mesh-llm` instances on one host as a first-class workflow, or merely avoid breaking it?
- Should we treat mixed-version meshes as supported, tolerated, or warned-about?

ndizazzo:
## What changed

Previously, mesh-llm managed its child processes (llama-server, rpc-server) using global pkill -f calls and name-based helpers. This was unsafe: it could terminate unrelated processes with similar names, leave orphaned children after crashes, and provided no reliable way to determine which instances were actually alive.

This PR replaces that system with a per-instance runtime directory model using flock(2) for liveness, atomic JSON pidfiles, and scoped orphan reaping.

## How it works

Each mesh-llm process creates a runtime directory at:
`~/.mesh-llm/runtime/{pid}/`

**Contents:**

- `lock` — an flock(2) advisory lock held for the process lifetime (released automatically on crash or SIGKILL)
- `owner.json` — instance metadata (pid, version, api_port, started_at)
- `pidfiles/` — one JSON pidfile per child (llama-server.json, rpc-server-{port}.json)
- `logs/` — stdout/stderr per child

**On startup:**

- Each instance scans the runtime root for directories with no active flock (dead owners)
- It reaps orphaned children listed in those pidfiles
- This is cooperative: live instances never interfere with each other

**Signal delivery:**

- Validates /proc/{pid}/comm (Linux) or `ps -p {pid} -o comm=` (macOS)
- Verifies process start time within ±2s tolerance
- Uses `is_safe_kill_target(pid)` to prevent unsafe signals (e.g. kill(0), kill(-1))

## Key design decisions

**flock over polling**

- Kernel-enforced, race-free liveness
- No heartbeat, no stale PID windows
- Automatically released on crash

**RAII pidfile guards (PidfileGuard)**

- Atomic writes (.tmp → rename)
- Cleanup on Drop
- Surviving children are reaped on next startup

**Pure reap logic (reap module)**

- No filesystem or I/O dependencies
- Returns typed ReapDecision
- Fully unit-testable (skip-self, skip-locked, etc.)

**Unified child handles**

- RpcServerHandle mirrors LlamaServerHandle
- Tracks PID, port, and expected_exit
- Prevents noisy logs during intentional shutdown
- Writes/removes pidfiles automatically

**Runtime root override** (Priority):

1. MESH_LLM_RUNTIME_ROOT
2. $XDG_RUNTIME_DIR/mesh-llm/runtime
3. ~/.mesh-llm/runtime

Used in tests to isolate state.

## Legacy removal

Removed entirely:

- `kill_llama_server()`, `kill_orphan_rpc_servers()`
- `terminate_process_by_name()`, `force_kill_process_by_name()`, `kill_process_by_name()`
- `is_process_running()` (name-based)

**Regression test:** `no_pkill_f_in_source_tree` ensures these patterns cannot reappear

## API & UI

`/api/status` now includes:

- `version`
- `local_instances` — all mesh-llm instances on the same machine

**Console updates:**

- Peer version shown in node table and sidebar
- Warnings for:
  - mixed-version meshes
  - multiple local instances

## Known limitations

- **NFS home directories:** flock is unreliable → Override MESH_LLM_RUNTIME_ROOT to a local path
- **Windows:** orphan detection is best-effort (flock is a no-op)
- **Symlinked paths:** May appear as separate dirs but share correct liveness behavior

## Testing notes

- All existing API tests pass (45 tests)
- New unit tests cover:
  - pidfile round-trip
  - liveness detection
  - reap decision matrix
  - scanner snapshot mapping
  - status self-entry safety net

**Manual verification:**

1. `kill -9` a running mesh-llm
2. Restart
3. Confirm orphaned llama-server and rpc-server processes are reaped

Closes #174


i386:
@copilot should we align the plugin process management to this same strategy for consistency?

i386 (approved):
LGTM - might be worth following up with this suggestion https://github.com/michaelneale/mesh-llm/pull/217#issuecomment-4227636336

i386:
@copilot apply changes based on the comments in [this thread](https://github.com/michaelneale/mesh-llm/pull/217#pullrequestreview-4087606081)

PR #220, opened 2026-04-08
docker: multi-stage images for most architectures, CI, and fly rewrite

### LINKED ISSUE #115: feature: dockerized setup for K3S / container environments
## Purpose

Provide a set of Docker images that cover every runtime variant of mesh-llm (CPU, CUDA, ROCm, Vulkan, and client-only), produced by multi-stage Dockerfiles that share build stages wherever possible. The Docker build path must reuse the same compiler-cache, layer-cache, and sccache tooling already employed by `just build` and the GitHub Actions workflows, keeping parity between native and containerised outputs with minimal additional CI time.

---

### UI (React web console — `mesh-llm/ui/`)

- Add a dedicated `ui-builder` stage in every Dockerfile that clones the UI source, runs `npm ci`, and `npm run build`, producing the `dist/` artefact consumed by the Rust build.
- The UI stage uses `node:24-alpine` to match the Node 24 requirement used in existing CI.
- The UI build output is consumed by the `rust-builder` stage via `COPY --from=ui-builder`.
- No changes to the UI source code are required; the stage mirrors the existing `just build` / `npm ci && npm run build` step.

### Rust Binary (`mesh-llm/src/`)

- Add a `rust-deps` stage using `cargo-chef` to pre-cook dependencies independently of application source. This allows Docker layer caching to skip the full `cargo build --release` when only `src/` files change.
- The `rust-deps` stage is shared across all image variants (CPU, CUDA, ROCm, Vulkan, client-only) because the Rust crate has no backend-specific compilation flags.
- The `rust-builder` stage copies the pre-cooked deps from `rust-deps` and the UI artefact from `ui-builder`, then runs `cargo build --release --locked -p mesh-llm`.
- Wire BuildKit cache mounts (`--mount=type=cache,target=/usr/local/cargo/registry` and `target=/src/mesh-llm/target`) to accelerate incremental builds without inflating image layer size.
- The final runtime stages copy only `/src/mesh-llm/target/release/mesh-llm`.

### llama.cpp Build (`llama.cpp/` fork)

Define one build stage per backend. All stages share a common cmake base configuration; only the `-DGGML_*` flags differ.

#### Shared cmake flags across all backends
```
-DGGML_RPC=ON
-DBUILD_SHARED_LIBS=OFF
-DLLAMA_OPENSSL=OFF
-DCMAKE_BUILD_TYPE=Release
```

Inline `git clone -b rebase-upstream-master --depth 1 https://github.com/michaelneale/llama.cpp.git` inside each llama builder stage. The `.dockerignore` already excludes `llama.cpp/`, so the clone must happen at build time. Use BuildKit's `--mount=type=cache` on the cmake build directory to reuse compilation artefacts across local rebuilds.

#### `llama-builder-cpu`
- Base image: `debian:bookworm-slim` with `cmake ninja-build build-essential`
- Flags: `-DGGML_CUDA=OFF -DGGML_HIP=OFF -DGGML_VULKAN=OFF -DGGML_METAL=OFF`
- Produces: `rpc-server`, `llama-server`, `llama-moe-split`

#### `llama-builder-cuda`
- Base image: `nvidia/cuda:12.8.0-devel-ubuntu22.04` (matches existing CI container)
- Additional flags: `-DGGML_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES="75;80;86;89;90;120"`
- Wire `CMAKE_CUDA_COMPILER_LAUNCHER=sccache` when sccache is present (same logic as `build-linux.sh`)
- Produces: `rpc-server`, `llama-server`, `llama-moe-split`

#### `llama-builder-rocm`
- Base image: `rocm/dev-ubuntu-24.04:7.0-complete` (matches existing CI container)
- Additional flags: `-DGGML_HIP=ON -DAMDGPU_TARGETS="gfx90a;gfx942;gfx1100;gfx1101;gfx1102;gfx1200;gfx1201"`
- Set `HIPCXX` and `HIP_PATH` from `hipconfig` as the existing build script does
- Produces: `rpc-server`, `llama-server`, `llama-moe-split`

#### `llama-builder-vulkan`
- Base image: `debian:bookworm-slim` with `libvulkan-dev glslc`
- Additional flags: `-DGGML_VULKAN=ON`
- Produces: `rpc-server`, `llama-server`, `llama-moe-split`

### Runtime Images

Each runtime image is a minimal `debian:bookworm-slim` layer with only `ca-certificates` and `libgomp1` installed — no toolchain.

| Image tag | Llama backend stage | Includes llama binaries? |
|---|---|---|
| `mesh-llm:client` | — | No |
| `mesh-llm:cpu` | `llama-builder-cpu` | Yes |
| `mesh-llm:cuda` | `llama-builder-cuda` | Yes |
| `mesh-llm:rocm` | `llama-builder-rocm` | Yes |
| `mesh-llm:vulkan` | `llama-builder-vulkan` | Yes |

The `cuda` runtime image uses `nvidia/cuda:12.8.0-runtime-ubuntu22.04` (not the devel base) to avoid shipping compiler tools.
The `rocm` runtime image uses `rocm/dev-ubuntu-24.04:7.0` runtime layer where available, otherwise `debian:bookworm-slim` with minimal ROCm runtime libs (`libamdhip64`, `libhsa-runtime64`).

All images set `EXPOSE 3131 9337` and use the same entrypoint pattern as the existing `fly/Dockerfile`.

**Entrypoint design** — a single `/entrypoint.sh` copies across from `fly/Dockerfile`, accepting `APP_MODE=console|api|worker`:

```sh
case "$APP_MODE" in
  api)     exec mesh-llm --client --auto --port 9337 --listen-all ;;
  console) exec mesh-llm --client --auto --port 9337 --console 3131 --listen-all ;;
  worker)  exec mesh-llm --auto --port 9337 --bin-dir /usr/local/lib/mesh-llm/bin ;;
  *)       exec mesh-llm "$@" ;;
esac
```

llama.cpp binaries for the `worker` and full-node images are copied to `/usr/local/lib/mesh-llm/bin/` with flavour-suffixed names (e.g. `rpc-server-cuda`) matching the convention used by `just bundle`.

### Dockerfile Layout

```
docker/
  Dockerfile.client           # mesh-llm binary only (supersedes fly/Dockerfile)
  Dockerfile.cpu              # CPU full node
  Dockerfile.cuda             # CUDA full node
  Dockerfile.rocm             # ROCm full node
  Dockerfile.vulkan           # Vulkan full node
  entrypoint.sh               # Shared entrypoint (all images)
```

The existing `fly/Dockerfile` is left in place but updated to delegate to `docker/Dockerfile.client` via a `FROM` reference or a symlink note in its header comment, so `fly deploy` continues to work unchanged.

### Build System (Justfile)

Add the following `just` recipes, keeping existing recipes intact:

```justfile
# Build the client-only Docker image (no GPU, no llama.cpp)
docker-build-client tag="mesh-llm:client":
    docker build -f docker/Dockerfile.client -t {{ tag }} .

# Build the CPU full-node Docker image
docker-build-cpu tag="mesh-llm:cpu":
    docker build -f docker/Dockerfile.cpu -t {{ tag }} .

# Build the CUDA full-node Docker image
docker-build-cuda tag="mesh-llm:cuda" cuda_arch="75;80;86;89;90;120":
    docker build -f docker/Dockerfile.cuda \
        --build-arg CUDA_ARCH="{{ cuda_arch }}" \
        -t {{ tag }} .

# Build the ROCm full-node Docker image
docker-build-rocm tag="mesh-llm:rocm" rocm_arch="gfx90a;gfx942;gfx1100;gfx1101;gfx1102;gfx1200;gfx1201":
    docker build -f docker/Dockerfile.rocm \
        --build-arg ROCM_ARCH="{{ rocm_arch }}" \
        -t {{ tag }} .

# Build the Vulkan full-node Docker image
docker-build-vulkan tag="mesh-llm:vulkan":
    docker build -f docker/Dockerfile.vulkan -t {{ tag }} .

# Run the client console image locally
docker-run-client tag="mesh-llm:client":
    docker run --rm -p 3131:3131 -p 9337:9337 -e APP_MODE=console {{ tag }}

# Run a CPU worker node locally (requires model volume mount)
docker-run-cpu models=home_dir/".models" tag="mesh-llm:cpu":
    docker run --rm -p 9337:9337 \
        -v {{ models }}:/root/.models \
        -e APP_MODE=worker {{ tag }}
```

`DOCKER_BUILDKIT=1` is assumed (required for `--mount=type=cache`). Recipes can export it inline or document the requirement in a comment header.

### CI / GitHub Actions

Add a new workflow file `.github/workflows/docker.yml`:

**Triggers:**
- `push` to `main` (builds and pushes `:latest` and `:sha-<short>` tags)
- `push` of `v*` tags (builds and pushes `:vX.Y.Z` and `:latest`)
- `pull_request` (builds only, no push)
- `workflow_dispatch`

**Registry:** `ghcr.io/${{ github.repository_owner }}/mesh-llm`

**Multi-arch target platforms:**

| Image | `linux/amd64` | `linux/arm64` | Notes |
|---|---|---|---|
| `client` | ✅ | ✅ | Pure Rust + networking; ARM64 runs on Graviton, Ampere, Raspberry Pi 5, Jetson (client mode) |
| `cpu` | ✅ | ✅ | llama.cpp CPU backend builds cleanly on ARM64 |
| `vulkan` | ✅ | ✅ | Vulkan is available on ARM64 (Mali, Adreno, etc.) |
| `cuda` | ✅ | ❌ | NVIDIA's `cuda:*-devel` images are `amd64`-only; Jetson uses a separate JetPack image — out of scope |
| `rocm` | ✅ | ❌ | ROCm has no ARM64 Linux support |

QEMU emulation is explicitly rejected for C++ and Rust compilation — it is orders of magnitude too slow. Use native runners for each platform.

**CI job structure for multi-arch (native-runner merge pattern):**

Each multi-arch image variant is built in two parallel jobs — one per platform — then merged into a single multi-arch manifest. This keeps build times close to single-arch totals.

```
docker-client-amd64   (ubuntu-latest)        ─┐
docker-client-arm64   (ubuntu-24.04-arm)     ─┤─ docker-client-merge (imagetools create)
                                              │
docker-cpu-amd64      (ubuntu-latest)        ─┐
docker-cpu-arm64      (ubuntu-24.04-arm)     ─┤─ docker-cpu-merge
                                              │
docker-vulkan-amd64   (ubuntu-latest)        ─┐
docker-vulkan-arm64   (ubuntu-24.04-arm)     ─┤─ docker-vulkan-merge
                                              │
docker-cuda           (ubuntu-latest)         ─── (single-arch, no merge needed)
docker-rocm           (ubuntu-latest)         ─── (single-arch, no merge needed)
```

Each per-platform job uses `build-push-action` to push a platform-specific digest (e.g. `ghcr.io/.../mesh-llm:sha-abc123-amd64`). The merge job calls `docker buildx imagetools create` to combine them into the canonical tag. This avoids the complexity of a shared BuildKit builder across runners.

**Jobs summary** (`fail-fast: false`):

| Job name | Dockerfile | Runner | Platforms |
|---|---|---|---|
| `docker-client-amd64` | `Dockerfile.client` | `ubuntu-latest` | `linux/amd64` |
| `docker-client-arm64` | `Dockerfile.client` | `ubuntu-24.04-arm` | `linux/arm64` |
| `docker-client-merge` | — | `ubuntu-latest` | manifest merge |
| `docker-cpu-amd64` | `Dockerfile.cpu` | `ubuntu-latest` | `linux/amd64` |
| `docker-cpu-arm64` | `Dockerfile.cpu` | `ubuntu-24.04-arm` | `linux/arm64` |
| `docker-cpu-merge` | — | `ubuntu-latest` | manifest merge |
| `docker-vulkan-amd64` | `Dockerfile.vulkan` | `ubuntu-latest` | `linux/amd64` |
| `docker-vulkan-arm64` | `Dockerfile.vulkan` | `ubuntu-24.04-arm` | `linux/arm64` |
| `docker-vulkan-merge` | — | `ubuntu-latest` | manifest merge |
| `docker-cuda` | `Dockerfile.cuda` | `ubuntu-latest` | `linux/amd64` |
| `docker-rocm` | `Dockerfile.rocm` | `ubuntu-latest` | `linux/amd64` |

**Actions used:**
- `docker/setup-buildx-action@v3` — enables BuildKit on every job
- `docker/login-action@v3` — logs in to `ghcr.io` using `GITHUB_TOKEN`
- `docker/metadata-action@v5` — generates image tags from git ref and SHA
- `docker/build-push-action@v6` — builds and (on non-PR) pushes platform-specific digests; uses `cache-from`/`cache-to` with GHA cache backend
- `docker/bake-action` (optional alternative) — if the job count becomes unwieldy, a Bake file can replace individual job definitions while keeping the same native-runner split

**Cache strategy in CI:**
- Each per-platform job uses `cache-from: type=gha,scope=docker-<variant>-<arch>` and `cache-to: type=gha,mode=max,scope=docker-<variant>-<arch>` (e.g. `scope=docker-client-arm64`) so amd64 and arm64 caches are stored independently and don't evict each other.
- The `ui-builder` and `rust-deps` stages are shared across all variants scoped by arch. A `scope=docker-shared-amd64` / `scope=docker-shared-arm64` exported by the `docker-client-*` jobs and imported by `docker-cpu-*` and `docker-vulkan-*` of the same arch avoids redundant re-builds of those stages.
- No sccache action needed inside Docker builds; BuildKit cache mounts on `~/.cargo/registry` and the cmake build dir provide equivalent benefits without requiring an external sccache server.

**Integration with the existing release workflow:**
- The existing `release.yml` continues to produce native tarballs unchanged.
- Docker images are published by `docker.yml` on tag push, running in parallel with `release.yml`.
- No dependency between the two workflows.

---

## Acceptance Criteria

- [ ] `docker build -f docker/Dockerfile.client .` completes from a clean context and produces a runnable `mesh-llm:client` image that responds to `docker run mesh-llm:client --version`.
- [ ] `docker build -f docker/Dockerfile.cpu .` completes and the produced image passes `docker run mesh-llm:cpu --version && docker run mesh-llm:cpu --help`.
- [ ] `docker build -f docker/Dockerfile.cuda .` completes (compile-only; a GPU is not required to build) on a standard `ubuntu-latest` runner with no pre-installed CUDA toolkit, using the `nvidia/cuda:12.8.0-devel-ubuntu22.04` base image.
- [ ] `docker build -f docker/Dockerfile.rocm .` completes on a standard `ubuntu-latest` runner, using the `rocm/dev-ubuntu-24.04:7.0-complete` base image.
- [ ] `docker build -f docker/Dockerfile.vulkan .` completes on a standard `ubuntu-latest` runner.
- [ ] A second `docker build` run on an unchanged source tree for any variant completes in under 60 seconds due to layer and BuildKit cache hits.
- [ ] All five `just docker-build-*` recipes work from the repo root and produce correctly tagged local images for the host architecture.
- [ ] `ghcr.io/.../mesh-llm:client`, `:cpu`, and `:vulkan` are published as multi-arch manifests covering `linux/amd64` and `linux/arm64`; `docker manifest inspect` confirms both platform digests are present.
- [ ] `ghcr.io/.../mesh-llm:cuda` and `:rocm` are published as single-arch (`linux/amd64`) manifests.
- [ ] The `linux/arm64` `client` image runs correctly on an ARM64 host (e.g. Graviton EC2, Raspberry Pi 5, or `--platform linux/arm64` emulation as a smoke check) and responds to `--version`.
- [ ] No QEMU or cross-compilation is used for any image build; all images are built on native runners for their respective architectures.
- [ ] The `.github/workflows/docker.yml` workflow runs to completion on a PR without pushing any images, and pushes tagged images on merge to `main`.
- [ ] The existing `fly deploy --config fly/console/fly.toml --dockerfile fly/Dockerfile` command continues to work unchanged.
- [ ] The existing `just build`, `just release-build`, and all release workflow jobs remain unaffected by the new Dockerfiles.
- [ ] Images produced by `Dockerfile.cpu`, `Dockerfile.cuda`, `Dockerfile.rocm`, and `Dockerfile.vulkan` contain `rpc-server-<flavor>`, `llama-server-<flavor>`, and `llama-moe-split` under `/usr/local/lib/mesh-llm/bin/` with the correct flavour suffixes matching the `just bundle` convention.
- [ ] The `APP_MODE=worker` entrypoint starts mesh-llm with `--bin-dir /usr/local/lib/mesh-llm/bin` and resolves the correct flavoured binaries at runtime.

---

## Notes

- **macOS / Metal — no pass-through possible**: Docker on macOS runs Linux containers inside a Linux VM (Apple Virtualization Framework). Metal is implemented entirely in Apple's macOS userspace frameworks (`Metal.framework`) and kernel drivers; Apple ships no Linux Metal driver and provides no container pass-through mechanism for Metal Compute. What the VM exposes to containers is a `virtio-gpu` *display* device only — not a Metal compute device. This is architecturally different from NVIDIA CUDA, where NVIDIA ships a Linux kernel driver and `nvidia-container-toolkit` to expose GPU devices into containers. No `Dockerfile.metal` is defined. Mac users run GPU nodes natively via `just build`.

  **Hybrid option (macOS GPU node + containerised mesh-llm)**: It is possible to run `llama-server` and `rpc-server` as native macOS processes (with full Metal access) while running `mesh-llm` itself inside a container. `mesh-llm` accepts `--bin-dir` to locate llama binaries, and if the native binaries are bind-mounted into the container (e.g. `-v $(which llama-server-metal):/usr/local/lib/mesh-llm/bin/llama-server-metal:ro`) the mesh binary can launch them as host processes via the host PID namespace (`--pid=host`). This is complex, breaks container isolation, and provides no practical benefit over native `just build` for most Mac users. It is documented here for completeness but is out of scope for this feature.
- **GPU access at runtime**: CUDA images require `--gpus all` (NVIDIA Container Toolkit) at `docker run` time. ROCm images require `--device=/dev/kfd --device=/dev/dri`. These flags are documented in each Dockerfile header comment but are not enforced by the entrypoint.
- **Multi-arch builds**: `linux/amd64` + `linux/arm64` is a first-class goal for `client`, `cpu`, and `vulkan` images. CUDA and ROCm remain `amd64`-only because neither NVIDIA's CUDA devel images nor ROCm support `arm64` in the standard way (Jetson uses a separate JetPack-based image; that is out of scope). Native GitHub Actions ARM runners (`ubuntu-24.04-arm`) are used — QEMU cross-compilation is explicitly rejected for C++ and Rust builds due to speed. The native-runner merge pattern (build each arch as a separate job, combine with `imagetools create`) adds parallelism rather than wall-clock time.
- **llama.cpp clone in Dockerfile**: The `.dockerignore` currently excludes `llama.cpp/`, so each llama builder stage must `git clone` the fork during the build. The depth-1 clone is fast (~seconds) but requires network access. BuildKit cache mounts on the cmake build directory keep subsequent rebuilds fast without re-cloning.
- **`cargo-chef`**: If adding `cargo-chef` to the `rust-deps` stage proves impractical (e.g. requiring a pinned toolchain image), the fallback is a simpler two-step: copy `Cargo.toml` + `Cargo.lock` only, run `cargo build --release` on a stub `lib.rs`, then overwrite with real source. This is less clean but avoids an extra tool dependency.
- **`fly/Dockerfile` compatibility**: The Fly deploy pipeline uses `fly/Dockerfile` with the build context rooted at the repo root. The new `docker/Dockerfile.client` duplicates that functionality with improved caching. The `fly/Dockerfile` header should be updated to note `docker/Dockerfile.client` as the canonical version, but it should not be deleted until Fly deploy is confirmed to work with the new path.
- **Caching scope isolation in CI**: If the `rust-deps` and `ui-builder` stages are shared across Docker jobs in the same workflow run, they must use the same BuildKit GHA cache scope name. Prefer a dedicated `scope=docker-shared` exported by whichever job runs first (`docker-client`) and imported by the others.
- **Windows**: Windows Docker builds are out of scope _for now_. Windows users build natively via `just build backend=...`.
- **Model volumes**: Docker images do not bundle models. The `docker-run-cpu` recipe mounts `~/.models` from the host. This matches the native `mesh-llm` behaviour of reading from `~/.models`.

ndizazzo:
## Summary

This PR adds five multi-stage Docker images (client, cpu, vulkan, cuda, rocm), a CI workflow that publishes them as native multi-arch manifests, and fixes a latent bug in `fly/Dockerfile` that would break any cleanroom deploy.

## What you can do now

```bash
# Smallest, most portable — runs the web console
docker run --rm -p 3131:3131 -p 9337:9337 \
  -e APP_MODE=console ghcr.io/<owner>/mesh-llm:latest

# Full worker nodes with bundled llama.cpp
docker run --rm -p 9337:9337 -v ~/.models:/root/.models \
  -e APP_MODE=worker ghcr.io/<owner>/mesh-llm:cpu

# GPU workers
docker run --rm --gpus all -p 9337:9337 \
  -v ~/.models:/root/.models -e APP_MODE=worker \
  ghcr.io/<owner>/mesh-llm:cuda

docker run --rm --device=/dev/kfd --device=/dev/dri -p 9337:9337 \
  -v ~/.models:/root/.models -e APP_MODE=worker \
  ghcr.io/<owner>/mesh-llm:rocm
```

Or build locally with the new `just` recipes:

```bash
just docker-build-client
just docker-build-cpu
just docker-build-vulkan
just docker-build-cuda   # override arch: just docker-build-cuda "mesh-llm:cuda" "75;80"
just docker-build-rocm
```

## Images published

| Tag | Variant | Arch | Size |
|---|---|---|---|
| `:latest`, `:client` | client-only (no llama.cpp, no GPU) | amd64 + arm64 | ~54 MB |
| `:cpu` | full node, llama.cpp CPU | amd64 + arm64 | ~58 MB |
| `:vulkan` | full node, Vulkan backend | amd64 + arm64 | ~75 MB |
| `:cuda` | full node, CUDA 12.8.0 | amd64 | ~2.7 GB |
| `:rocm` | full node, ROCm 7.0 | amd64 | ~1.4 GB |

All images ship a shared entrypoint with three modes selected via `APP_MODE`:

- `console` (default) — client node, API on `:9337`, web console on `:3131`
- `worker` — full node with bundled llama binaries at `/usr/local/lib/mesh-llm/bin/`
- *(empty)* — passthrough, args go straight to `mesh-llm`

The UI is embedded in the binary (`include_dir!`) and always co-deployed with the runtime — there is no UI-less mode.

## Fly.io deploys now actually work from clean state

`fly/Dockerfile` previously used a single `FROM rust:latest` stage with no UI build step. It relied on `mesh-llm/ui/dist/` already existing in the build context, so any deploy from a fresh checkout would hit the `include_dir!` compile-time macro and fail.

This PR rewrites it as a 4-stage multi-stage build structurally identical to `docker/Dockerfile.client`, with `node:24-alpine` building the UI and `debian:bookworm-slim` for the runtime. `fly.toml` is untouched — the runtime contract (`EXPOSE 3131 9337`, `ENV APP_MODE=console`, `ENTRYPOINT ["/entrypoint.sh"]`) is preserved, so `fly deploy --config fly/console/fly.toml --dockerfile fly/Dockerfile` works unchanged. Cleanroom worktree builds are part of the test evidence.

## Architecture

**Multi-stage pattern (all Dockerfiles).** Four or five stages: `ui-builder` (node:24-alpine) → `rust-deps` (cargo-chef 0.1.68 with full workspace manifests + `mesh-llm/proto/`) → `rust-builder` (overwrites stub `ui/dist` with real build before `cargo build --release --locked`) → optional `llama-builder-<flavor>` (clones `michaelneale/llama.cpp@upstream-latest`, shared cmake flags `GGML_RPC=ON BUILD_SHARED_LIBS=OFF LLAMA_OPENSSL=OFF CMAKE_BUILD_TYPE=Release`) → `runtime`.

Llama binaries land at `/usr/local/lib/mesh-llm/bin/{rpc-server,llama-server}-<flavor>` with an unsuffixed `llama-moe-split` shared across flavors. The llama.cpp commit SHA is baked into `/usr/local/share/mesh-llm/llama-revision.txt` and an OCI label.

**CUDA correctness.** `docker/Dockerfile.cuda` ships `-DGGML_CUDA_FA_ALL_QUANTS=ON` — this is a correctness requirement, not a perf tuning knob. Without it, asymmetric K/V quantization paths hit `BEST_FATTN_KERNEL_NONE` and crash rpc-server (see [ggml-org/llama.cpp#20866](https://github.com/ggml-org/llama.cpp/issues/20866)). The CI workflow grep-verifies `GGML_CUDA_FA_ALL_QUANTS:BOOL=ON` appears in the CMake cache on every CUDA build, so the flag cannot silently drop.

**Spec compliance harness.** `scripts/verify-docker-spec.sh` enforces 13 invariants across all Dockerfiles and the workflow (ui-builder before cargo build, `libdbus-1-dev` for keyring sync-secret-service, no `just bundle`, no `protobuf-compiler` apt package since protoc is vendored via `protoc_bin_vendored`, no QEMU, flavored binary naming, runtime libs `ca-certificates libgomp1 libdbus-1-3`, etc.). Current state: **52 PASSED, 0 FAILED**. `docker/SPEC_COMPLIANCE.md` tracks each item with file:line references and a verification log.

**CI.** `.github/workflows/docker.yml` defines 11 jobs: client/cpu/vulkan × (amd64, arm64, merge) + cuda + rocm amd64-only. ARM64 uses native `ubuntu-24.04-arm` runners — no QEMU anywhere. Merge jobs stitch per-arch digests with `docker buildx imagetools create`. GHA cache scopes are split so a Rust-only change invalidates only the llama layer of one variant, and CUDA/ROCm caches are isolated from debian-based variants (different base images). `release.yml` is untouched — no coupling in either direction.

## Validation

- `bash scripts/verify-docker-spec.sh` → **52 PASSED / 0 FAILED / 0 SKIPPED**
- All 5 images built and smoke-tested locally: `--version` returns `mesh-llm 0.58.0`, flavored binaries present at expected paths, `ldd` reports zero unresolved libs on the client image, all images under their size budgets (250 MB / 600 MB / 900 MB / 4.5 GB / 12 GB respectively).
- `fly/Dockerfile` cleanroom build verified via `git worktree add` with `ui/dist` pre-deleted.
- `just docker-build-*` recipes dry-runned with arch overrides.
- Entrypoint tested in all three modes.

## Files changed

```
.dockerignore                    +14
.github/workflows/docker.yml     new (590 lines)
Justfile                         +34 (7 new recipes, append-only)
docker/Dockerfile.client         new
docker/Dockerfile.cpu            new
docker/Dockerfile.cuda           new
docker/Dockerfile.rocm           new
docker/Dockerfile.vulkan         new
docker/entrypoint.sh             new
docker/SPEC_COMPLIANCE.md        new
fly/Dockerfile                   rewritten (+66/-33)
scripts/verify-docker-spec.sh    new
```

No changes to `mesh-llm/src/`, `Cargo.toml`, `Cargo.lock`, `release.yml`, or `fly.toml`.

## Notes for reviewers

- Stages 1–3 of all Dockerfiles are intentionally near-identical copies rather than shared via a common base image — Docker doesn't support cross-file stage inheritance, and keeping them self-contained avoids drift between the fly deploy and the published `:client` image. `docker/SPEC_COMPLIANCE.md` calls this out.
- `cargo-chef` is pinned to `0.1.68 --locked` and the `cook` step now also uses `--locked` for reproducibility.
- The Vulkan image installs Khronos Vulkan-Headers v1.3.296 on top of Debian Bookworm's `libvulkan-dev` (1.3.239) because current `llama.cpp@upstream-latest` needs `VK_EXT_layer_settings` from 1.3.261+. The runtime loader from Debian is kept; only headers are upgraded.
- ROCm runtime uses the plain `rocm/dev-ubuntu-24.04:7.0` base (not `-complete`) to keep the image reasonable — `-complete` is ~20 GB.
- Images run as root to match the existing `fly/Dockerfile` convention; K3S/pod users can override via `securityContext.runAsNonRoot`.

Closes #115

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,17 @@ macos/
 iroh-relay/
 evals/
 *.tar.gz
+# Node.js build artefacts (UI)
+**/node_modules/
+**/dist/
+# Local workspace state
+.sisyphus/
+.opencode/
+.claude/
+# Release bundle outputs
+*.bundle.tar.gz
+# Editor/IDE
+.vscode/
+.idea/
+# OS cruft
+.DS_Store
--- a/.github/workflows/docker-precheck.yml
+++ b/.github/workflows/docker-precheck.yml
@@ -1,0 +1,218 @@
+name: Docker precheck
+
+on:
+  workflow_call:
+    inputs:
+      job_name:
+        required: true
+        type: string
+      dockerfile_path:
+        required: true
+        type: string
+      validate_shared_core:
+        required: false
+        type: boolean
+        default: false
+      validate_llama_builder:
+        required: false
+        type: boolean
+        default: false
+      validate_cuda_extra:
+        required: false
+        type: boolean
+        default: false
+      validate_workflow_no_qemu:
+        required: false
+        type: boolean
+        default: false
+      validate_fly_ui_builder:
+        required: false
+        type: boolean
+        default: false
+      validate_entrypoint_modes:
+        required: false
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+jobs:
+  precheck:
+    name: ${{ inputs.job_name }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Validate target Dockerfile exists
+        shell: bash
+        run: |
+          set -euo pipefail
+          test -f "${{ inputs.dockerfile_path }}"
+
+      - name: Shared Dockerfile copies UI build before cargo build
+        if: ${{ inputs.validate_shared_core }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          ui_line=$(grep -n "COPY --from=ui-builder" "${{ inputs.dockerfile_path }}" | head -1 | cut -d: -f1 || true)
+          build_line=$(grep -n "cargo build" "${{ inputs.dockerfile_path }}" | head -1 | cut -d: -f1 || true)
+          if [[ -z "$ui_line" || -z "$build_line" ]]; then
+            echo "FAIL: missing COPY --from=ui-builder or cargo build in ${{ inputs.dockerfile_path }}" >&2
+            exit 1
+          fi
+          if (( ui_line >= build_line )); then
+            echo "FAIL: ui-builder COPY (line $ui_line) must precede cargo build (line $build_line) in ${{ inputs.dockerfile_path }}" >&2
+            exit 1
+          fi
+          echo "PASS: ui-builder COPY (line $ui_line) precedes cargo build (line $build_line)"
+
+      - name: Cargo chef copies full workspace manifests
+        if: ${{ inputs.validate_shared_core }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          file="${{ inputs.dockerfile_path }}"
+          grep -q "COPY Cargo.toml" "$file"
+          grep -q "COPY Cargo.lock" "$file"
+          grep -q "COPY mesh-llm/Cargo.toml" "$file"
+          grep -q "COPY mesh-llm/build.rs" "$file"
+          grep -q "COPY mesh-llm/plugin/Cargo.toml" "$file"
+          grep -q "COPY mesh-llm/plugin/build.rs" "$file"
+          grep -q "COPY mesh-llm/proto/" "$file"
+          if grep -qE "echo.*fn main|stub.*lib" "$file"; then
+            echo "FAIL: stub lib.rs shortcut detected in $file" >&2
+            exit 1
+          fi
+          echo "PASS: full workspace manifests present with corrected proto path"
+
+      - name: Rust builder installs libdbus-1-dev
+        if: ${{ inputs.validate_shared_core }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          grep -q "libdbus-1-dev" "${{ inputs.dockerfile_path }}"
+          echo "PASS: libdbus-1-dev present"
+
+      - name: Llama builder uses upstream-latest branch
+        if: ${{ inputs.validate_llama_builder }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          file="${{ inputs.dockerfile_path }}"
+          grep -q "upstream-latest" "$file"
+          if grep -q "rebase-upstream-master" "$file"; then
+            echo "FAIL: legacy rebase-upstream-master branch referenced in $file" >&2
+            exit 1
+          fi
+          echo "PASS: uses upstream-latest branch"
+
+      - name: CUDA build enables FA all quants
+        if: ${{ inputs.validate_cuda_extra }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          grep -q "DGGML_CUDA_FA_ALL_QUANTS=ON" "${{ inputs.dockerfile_path }}"
+          echo "PASS: GGML_CUDA_FA_ALL_QUANTS=ON present"
+
+      - name: Llama builder uses correct flavored binary names
+        if: ${{ inputs.validate_llama_builder }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          file="${{ inputs.dockerfile_path }}"
+          grep -q "/usr/local/lib/mesh-llm/bin/rpc-server-" "$file"
+          grep -q "/usr/local/lib/mesh-llm/bin/llama-server-" "$file"
+          grep -q "/usr/local/lib/mesh-llm/bin/llama-moe-split" "$file"
+          if grep -q "llama-moe-split-" "$file"; then
+            echo "FAIL: llama-moe-split must not be flavor-suffixed in $file" >&2
+            exit 1
+          fi
+          echo "PASS: flavored binary naming correct"
+
+      - name: Shared Dockerfiles do not invoke just bundle
+        if: ${{ inputs.validate_shared_core }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          if grep -q "just bundle" "${{ inputs.dockerfile_path }}"; then
+            echo "FAIL: just bundle present in ${{ inputs.dockerfile_path }}" >&2
+            exit 1
+          fi
+          echo "PASS: no just bundle"
+
+      - name: Shared Dockerfiles do not install protobuf-compiler
+        if: ${{ inputs.validate_shared_core }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          if grep -q "protobuf-compiler" "${{ inputs.dockerfile_path }}"; then
+            echo "FAIL: protobuf-compiler present in ${{ inputs.dockerfile_path }}" >&2
+            exit 1
+          fi
+          echo "PASS: no protobuf-compiler"
+
+      - name: Docker workflow avoids QEMU-based runners
+        if: ${{ inputs.validate_workflow_no_qemu }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          file=".github/workflows/docker.yml"
+          if grep -qiE '(^|[^[:alnum:]_])qemu([^[:alnum:]_]|$)' "$file"; then
+            echo "FAIL: QEMU tooling reference found in $file" >&2
+            exit 1
+          fi
+          if grep -q "setup-qemu-action" "$file"; then
+            echo "FAIL: setup-qemu-action found in $file" >&2
+            exit 1
+          fi
+          echo "PASS: no QEMU tooling references"
+
+      - name: Fly Dockerfile includes UI builder stage
+        if: ${{ inputs.validate_fly_ui_builder }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          file="${{ inputs.dockerfile_path }}"
+          grep -q "AS ui-builder" "$file"
+          grep -q "COPY --from=ui-builder" "$file"
+          echo "PASS: ui-builder stage present"
+
+      - name: Llama builder uses shared CMake flags
+        if: ${{ inputs.validate_llama_builder }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          file="${{ inputs.dockerfile_path }}"
+          grep -q "DGGML_RPC=ON" "$file"
+          grep -q "DBUILD_SHARED_LIBS=OFF" "$file"
+          grep -q "DLLAMA_OPENSSL=OFF" "$file"
+          grep -q "DCMAKE_BUILD_TYPE=Release" "$file"
+          echo "PASS: shared cmake flags present"
+
+      - name: Runtime apt libraries are present
+        if: ${{ inputs.validate_shared_core }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          file="${{ inputs.dockerfile_path }}"
+          grep -q "ca-certificates" "$file"
+          grep -q "libgomp1" "$file"
+          grep -q "libdbus-1-3" "$file"
+          echo "PASS: runtime apt libs present"
+
+      - name: Entrypoint supports console worker and default modes only
+        if: ${{ inputs.validate_entrypoint_modes }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          file="docker/entrypoint.sh"
+          grep -qE '^\s*console\)' "$file"
+          grep -qE '^\s*worker\)' "$file"
+          grep -qE '^\s*\*\)' "$file"
+          if grep -qE '^\s*api\)' "$file"; then
+            echo "FAIL: api mode present in $file" >&2
+            exit 1
+          fi
+          echo "PASS: console/worker/default only"
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -192,8 +192,8 @@ jobs:
             type=ref,event=branch
             type=semver,pattern={{version}}
             type=sha,prefix=sha-,format=short
-            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
-            type=raw,value=client,enable=${{ github.ref == 'refs/heads/main' }}
+            type=raw,value=latest
+            type=raw,value=client
       - name: Create and push multi-arch manifest
         run: |
           SHORT_SHA=$(echo "${{ github.sha }}" | head -c 7)
@@ -319,7 +319,7 @@ jobs:
             type=ref,event=branch,suffix=-cpu
             type=semver,pattern={{version}}-cpu
             type=sha,prefix=sha-,format=short,suffix=-cpu
-            type=raw,value=cpu,enable=${{ github.ref == 'refs/heads/main' }}
+            type=raw,value=cpu
       - name: Create and push multi-arch manifest
         run: |
           SHORT_SHA=$(echo "${{ github.sha }}" | head -c 7)
@@ -445,7 +445,7 @@ jobs:
             type=ref,event=branch,suffix=-vulkan
             type=semver,pattern={{version}}-vulkan
             type=sha,prefix=sha-,format=short,suffix=-vulkan
-            type=raw,value=vulkan,enable=${{ github.ref == 'refs/heads/main' }}
+            type=raw,value=vulkan
       - name: Create and push multi-arch manifest
         run: |
           SHORT_SHA=$(echo "${{ github.sha }}" | head -c 7)
@@ -487,7 +487,7 @@ jobs:
             type=ref,event=branch,suffix=-cuda
             type=semver,pattern={{version}}-cuda
             type=sha,prefix=sha-,format=short,suffix=-cuda
-            type=raw,value=cuda,enable=${{ github.ref == 'refs/heads/main' }}
+            type=raw,value=cuda
       - name: Build
         uses: docker/build-push-action@v6
         with:
@@ -535,7 +535,7 @@ jobs:
             type=ref,event=branch,suffix=-rocm
             type=semver,pattern={{version}}-rocm
             type=sha,prefix=sha-,format=short,suffix=-rocm
-            type=raw,value=rocm,enable=${{ github.ref == 'refs/heads/main' }}
+            type=raw,value=rocm
       - name: Build
         uses: docker/build-push-action@v6
         with:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,555 @@
+name: docker
+
+on:
+  push:
+    tags: ['v*']
+  pull_request:
+    paths:
+      - '.dockerignore'
+      - 'Justfile'
+      - 'docker/**'
+      - 'fly/Dockerfile'
+      - 'mesh-llm/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - '.github/workflows/docker.yml'
+      - '.github/workflows/docker-precheck.yml'
+concurrency:
+  group: docker-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+env:
+  REGISTRY: ghcr.io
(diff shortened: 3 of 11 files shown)


i386 (approved):
LGTM - I would just double check some of the copilot comments before merging.

PR #222, opened 2026-04-08
display: rename misleading VRAM labels to capacity

Bortlesboat:
Fixes #219.

`vram_bytes` isn't raw GPU VRAM — it's `gpu_vram + 0.75 * (system_ram - gpu_vram)`, which accounts for llama.cpp's ability to offload layers to system RAM. The formula is intentional and correct, but the label was wrong. A machine with 17GB GPU VRAM and 96GB system RAM reports ~78GB, which is accurate as a capacity number but confusing when it says "78.2GB VRAM."

Changed "VRAM" to "capacity" in all user-visible log and display strings that pass `vram_bytes`. Internal variable names and the `--max-vram` flag are untouched.

### DIFF
--- a/mesh-llm/src/api/mod.rs
+++ b/mesh-llm/src/api/mod.rs
@@ -114,14 +114,14 @@ fn fit_hint_for_machine(size_gb: f64, my_vram_gb: f64) -> (String, String) {
     if size_gb <= 0.0 || my_vram_gb <= 0.0 {
         return (
             "Unknown".into(),
-            "No local VRAM signal is available for this machine yet.".into(),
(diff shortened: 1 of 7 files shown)


ndizazzo:
@Bortlesboat seems reasonable - we also display aggregate "VRAM" badges for hosts in the mesh. Can you verify if that's updated too?

Not the ones that are for SoC and GPU though, those should stay as VRAM I think. If you want to see those, use `--enumerate-hosts`

michaelneale (approved):
thanks, yeah it did bug me that it was called that. I am ok with capacity if you are @ndizazzo

PR #226, opened 2026-04-09
client: only surface models that can actually serve requests

IvGolovach:
## Summary
Passive and public mesh clients now only surface models that can actually accept HTTP inference requests. `/api/models`, `/v1/models`, and the console stay aligned when worker-only legacy peers or placeholder targets are present.

## Why
Passive clients could previously advertise models that were only present on worker-only peers or only resolved to `InferenceTarget::None`. That made the console look ready even though the next request could still fail with a `503`.

## What Changed
- treat only `Host` peers as HTTP-callable inference targets for user-facing availability
- compute warm status, node counts, active nodes, and mesh VRAM in `/api/models` from HTTP-callable routes only
- filter `/v1/models` and passive auto-routing candidates so models backed only by `InferenceTarget::None` are not advertised as available
- keep the chat composer disabled until `/api/models` reports a warm model instead of falling back to `status.model_name`

## Example
```bash
mesh-llm --client --auto
curl -s http://localhost:9337/v1/models
curl -s http://localhost:3131/api/models
```

With this change, those surfaces only report models that a passive client can actually route to successfully.

## Architecture
The mesh catalog remains broader than immediate routability: cold catalog entries still appear in `/api/models`, but only HTTP-callable hosts contribute warm state and route statistics. This keeps discovery intact while making client readiness and model availability honest.

## Validation
- `cargo fmt --all`
- `CARGO_HOME=/tmp/mesh-llm-cargo-home CARGO_TARGET_DIR=/tmp/mesh-llm-passive-visible cargo test -p mesh-llm test_http_route_stats_only_count_http_callable_legacy_hosts`
- `CARGO_HOME=/tmp/mesh-llm-cargo-home CARGO_TARGET_DIR=/tmp/mesh-llm-passive-visible cargo test -p mesh-llm worker_only_legacy_models_are_excluded_from_http_routes`
- `CARGO_HOME=/tmp/mesh-llm-cargo-home CARGO_TARGET_DIR=/tmp/mesh-llm-passive-visible cargo test -p mesh-llm test_callable_models_excludes_none_only_targets`
- `npm test -- App.test.tsx`
- `npm run typecheck`
- `npm run build`
- `git diff --check`

## UI
The console now waits for a warm model from `/api/models` before enabling chat on passive clients.

### DIFF
--- a/mesh-llm/src/api/mod.rs
+++ b/mesh-llm/src/api/mod.rs
@@ -89,6 +89,59 @@ fn likely_reasoning_model(name: &str, description: Option<&str>) -> bool {
         .any(|needle| haystack.contains(needle))
 }
 
+#[derive(Debug, Default, PartialEq)]
+struct HttpRouteStats {
+    node_count: usize,
+    active_nodes: Vec<String>,
+    mesh_vram_gb: f64,
+}
+
+fn http_route_stats(
+    model_name: &str,
+    peers: &[mesh::PeerInfo],
+    my_hosted_models: &[String],
+    my_hostname: Option<&str>,
+    my_vram_gb: f64,
+) -> HttpRouteStats {
+    let mut active_nodes = Vec::new();
+    let mut node_count = 0usize;
+    let mut mesh_vram_gb = 0.0;
+
+    if my_hosted_models.iter().any(|hosted| hosted == model_name) {
+        node_count += 1;
+        mesh_vram_gb += my_vram_gb;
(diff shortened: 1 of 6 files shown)


ndizazzo:
@michaelneale you were poking this recently - whatcha think?

i386 (approved):
@IvGolovach this has annoyed me greatly. Thank you for fixing it.

PR #229, opened 2026-04-09
mesh: refresh degraded relay-backed peer connections

IvGolovach:
## Summary
Long-lived mesh connections now recover from relay-path decay without requiring a manual rejoin. When a peer stays stuck on a degraded relay path, mesh-llm refreshes that connection and re-gossips it before it becomes effectively isolated.

## Why
`ROADMAP.md` calls out a real stability problem here: some relay-backed connections degrade over hours and eventually strand otherwise healthy peers. Today we only react after a connection is already dead, or after a client loses every peer and falls back to Nostr re-discovery.

This change adds an intermediate recovery step for the common case where the peer is still alive but its selected path has decayed.

## What Changed
- added a relay health monitor on the long-lived runtime paths (`run_auto` and plugin MCP mode)
- sampled each connected peer's selected iroh path and tracked whether it is using a direct path or a relay path
- refreshed at most one peer connection per monitor cycle when the selected path is still relay-backed and the relay RTT is clearly degraded
- refreshed at most one peer connection per monitor cycle when the selected path has remained relay-only for too long
- skipped relay refresh when requests are in flight, when there is no current home relay, or when the reconnect cooldown is still active
- re-gossiped the refreshed connection before keeping it as the live replacement
- hardened connection replacement with `Connection::stable_id()` so a closing stale dispatcher cannot remove or reconnect over a newer replacement connection

## Architecture
- stays inside the existing heartbeat and reconnect architecture instead of introducing a new protocol
- targets the specific failure mode described in the roadmap: long-lived relay decay, not total mesh loss
- avoids churn by refreshing at most one peer per cycle and by respecting inflight traffic and cooldowns
- keeps full peer-loss recovery unchanged: heartbeat failure handling and Nostr re-discovery still own the hard-down case

## Protocol
Not applicable: runtime-only change. No wire format or gossip schema changes.

## Validation
- `cargo fmt --all -- mesh-llm/src/mesh/mod.rs mesh-llm/src/runtime/mod.rs`
- `cargo test -p mesh-llm relay_health_ -- --nocapture`
- `cargo test -p mesh-llm stale_dispatcher_cannot_remove_replacement_connection -- --nocapture`
- `cargo test -p mesh-llm reconnect_gossip_failure_removes_zombie_peer -- --nocapture`
- `cargo test -p mesh-llm test_rtt_ -- --nocapture`
- `cargo test -p mesh-llm -- --skip network::nostr::integration_tests::publish_discover_round_trip`
- Result: `635 passed; 0 failed; 1 filtered`

## Known Residual Risks
- this does not yet add relay health telemetry to the UI
- this does not replace full mesh recovery; it only improves recovery for degraded-but-still-live relay-backed peer connections


i386 (changes requested):
@IvGolovach this is a very good improvement 👍 

I can LGTM and merge if you can take a look at the copilot comments first - one seems to have caught a legit race.

i386:
@IvGolovach are you in the discord? drop by and say hi 👋

michaelneale (approved):

PR #232, opened 2026-04-09
split serving: recover dense models on surviving workers

IvGolovach:
## Summary

Dense tensor-split models now recover onto surviving workers after a split peer drops, instead of treating broad worker-set churn as the launch plan. The host re-plans the exact survivor set it can actually launch with and only advertises a split target when that exact set is ready.

## Why

Today the dense split path computes a broad candidate worker set in election, but launch re-selects workers again later. That mismatch can make restarts depend on spare peers that were never part of the actual launch, and it can overestimate readiness by counting capacity or tunnels that the selected split will not actually use.

This change makes dense split fail-down deterministic and survivor-aware on top of the process-subtree lifecycle work in #217.

## Diff scope

- added an explicit dense launch plan in `inference/election.rs` and reused it for restart gating, capacity checks, and launch-time worker selection
- re-plan dense split survivors by exact eligible worker IDs after peer loss, ordered by RTT and only as large as needed to satisfy the model footprint
- stop restart churn from unselected spare-worker changes by comparing the running plan to the exact planned survivor set
- require a complete RPC port map for the selected survivors before launching a split host, rather than proceeding on approximate peer counts
- keep insufficient-capacity cases in a waiting state instead of launching a broken split target

## Test proof

- `cd mesh-llm/ui && npm ci`
- `cargo test -p mesh-llm dense_launch_plan_ -- --nocapture` -> `4 passed; 0 failed; 664 filtered out`
- `cargo test -p mesh-llm selected_worker_ids_require_complete_rpc_port_map -- --nocapture` -> `1 passed; 0 failed; 667 filtered out`
- `cargo test -p mesh-llm -- --skip network::nostr::integration_tests::publish_discover_round_trip --skip runtime::instance::tests::validate_self_process_start_time_is_recent` -> `666 passed; 0 failed; 2 filtered out`

## Verification-pack proof

Not applicable - no CI, workflow, infra, governance, replay, or migration logic changed.

## Migration notes

Not applicable - no protocol or schema migration.

## CI context confirmation

Not applicable - no CI workflow or required context names changed.

## Rollback plan

`git revert <merge_commit_sha>`

No DB downgrade required.

## Known residual risks

- this improves dense split survivor recovery by restarting onto the exact remaining worker set; it does not add live in-process reshards without restart
- if the remaining eligible workers do not have enough capacity, or their selected RPC tunnels are not ready yet, the host intentionally waits instead of advertising a broken split target


ndizazzo:
Hey @IvGolovach - there is a PR already where I'm resolving the GPU workflows. Please hold on that stuff so we don't cross streams - it's ok for this PR to be red for now while that's sorted and we can rebase from there

IvGolovach:
> Hey @IvGolovach - there is a PR already where I'm resolving the GPU workflows. Please hold on that stuff so we don't cross streams - it's ok for this PR to be red for now while that's sorted and we can rebase from there

Thanks for the heads-up, that makes sense.

I’ll leave the GPU workflow changes out of this PR for now and wait for your workflow PR to land, then I’ll rebase on top of it. Happy to avoid crossing the streams, that usually ends badly :)

ndizazzo:
@IvGolovach should be good to go now - if you update your fork and rebase the changes, they should be picked up by your branch and build faster

ndizazzo:
I've been watching your CI jobs - it looks like they're still using the old non-warm cached versions. Just calling out that once changes are taken from main, you'll see 8-9 minute warm builds instead of 1 hour long full CUDA builds :)

IvGolovach:
> I've been watching your CI jobs - it looks like they're still using the old non-warm cached versions. Just calling out that once changes are taken from main, you'll see 8-9 minute warm builds instead of 1 hour long full CUDA builds :)

Thanks! That makes sense.

It looks like the base branch still hasn’t picked up the warm-cache changes from `main` yet, so that explains why the CUDA jobs are still taking the long path. I’ve already cleaned the workflow-specific changes out of this PR, and once those updates are in the base branch I’ll rebase again so #232 picks them up as well :)

Really appreciate you flagging it 👍

ndizazzo:
> It looks like the base branch still hasn’t picked up the warm-cache changes

OO I'll take care of that now, I didn't check this PR's base

i386:
Re-ran the Github workflow - looks like their container system was having a moment.

i386 (approved):

PR #233, opened 2026-04-10
feature: add embedded SDK support

### LINKED ISSUE #188: Add native embedded SDKs for Swift and Kotlin/Java using a shared Rust core
## Summary

Add a native embedded SDK strategy for `mesh-llm` built around a shared Rust core, with first-class Swift and Kotlin/Java bindings.

The goal is to let native apps join public or private meshes and make inference requests directly, without depending on a running local `mesh-llm` node, localhost HTTP, a browser SDK, or a CLI sidecar.

This issue covers both:

- embedded **client mode** for `iOS`, `macOS`, and `Android`
- embedded **host mode** for `macOS` apps that want to serve models locally and optionally contribute them to the mesh

## User-visible outcome

After this work, native apps should be able to:

- join a private mesh by invite token
- discover and auto-join public meshes
- list available models
- send `chat` / `responses` requests
- stream tokens back into the app
- do all of the above without starting `:9337`, `:3131`, or a sidecar process

On macOS, apps should also be able to optionally embed model serving in-process.

## Non-goals

- browser SDK
- localhost HTTP wrapper around the existing CLI
- shipping a native SDK that is just a thin launcher for `mesh-llm --client`
- iOS or Android local model serving in the first phase

### Apple

The Swift SDK must:

- support the latest major `iOS` release only
- support the latest major `macOS` release only
- pass App Store review
- use public APIs only
- avoid private frameworks, helper daemons, localhost control servers, and any hidden sidecar requirement

If `macOS` host mode is shipped in an App Store-safe app profile, it should run **in-process**, not by spawning helper binaries.

### Android

The Android SDK must:

- support the latest major Android release only
- join public/private meshes in-process
- avoid depending on a localhost sidecar

### 1. `mesh-client-core`

A Rust crate that implements the passive embedded mesh client only.

Responsibilities:

- public mesh discovery
- private join via invite token
- QUIC connection/session lifecycle
- route-table fetch and model inventory
- request tunneling for `chat` and `responses`
- token streaming
- reconnect and failover
- object upload / multimodal parity later

This crate should be app-library-safe:

- no CLI assumptions
- no subprocess management
- no localhost listeners
- no web console
- no plugin host
- no local model launching

### 2. `mesh-host-core`

A Rust crate for embedded host/serve behavior, scoped to `macOS` first.

Responsibilities:

- local model loading
- embedded serving lifecycle
- joining the mesh as a serving node
- optional contribution of local models to the mesh

This should be treated separately from `mesh-client-core` so mobile-safe client behavior stays narrow and easy to embed.

### 3. `mesh-ffi`

A stable C ABI over the Rust cores.

Responsibilities:

- opaque handles for client/host instances
- callback registration
- event and stream delivery
- explicit memory ownership rules
- stable error mapping

Do not expose Rust internals directly across the language boundary.

### 4. `bindings/swift`

A Swift wrapper packaged as a Swift Package and/or `XCFramework`.

Responsibilities:

- `MeshClient` for `iOS` + `macOS`
- `MeshHost` for `macOS` only
- `AsyncThrowingStream` or equivalent evented API
- public-API-only Apple integration

### 5. `bindings/jni`

A host-runnable JNI wrapper over `mesh-ffi`.

Responsibilities:

- native library loading
- JNI marshalling
- plain JVM testability on `macOS` and `Linux`

### 6. `bindings/kotlin`

A thin Kotlin layer over JNI.

Responsibilities:

- ergonomic Kotlin API
- `Flow` / suspend wrappers
- Android-facing surface with minimal framework-specific logic

## SDK surface

Keep the public native SDK intentionally small.

### Client API

- `joinPrivate(token)`
- `discoverPublic(filter)`
- `autoJoin(filter)`
- `listModels()`
- `chat(request)`
- `responses(request)`
- `disconnect()`
- `status()`

### Host API (`macOS` only)

- `startHost(config)`
- `stopHost()`
- `loadModel(spec)`
- `unloadModel(name)`
- `hostStatus()`

### Event model

Expose an evented/streaming interface rather than blocking callbacks only.

Events should include at least:

- `connecting`
- `joined`
- `modelsUpdated`
- `tokenDelta`
- `completed`
- `failed`
- `disconnected`

Mapping:

- Swift: `AsyncThrowingStream`
- Kotlin: `Flow`

## Extraction strategy from the current repo

Use the existing passive client, routing, and discovery behavior as the source material, but do **not** ship those current modules wholesale into the SDK without refactoring.

Relevant areas today include:

- `mesh-llm/docs/DESIGN.md`
- `mesh-llm/src/runtime/mod.rs`
- `mesh-llm/src/runtime/proxy.rs`
- `mesh-llm/src/network/proxy.rs`
- `mesh-llm/src/network/nostr.rs`
- `mesh-llm/src/api/` (only for understanding current request surfaces, not for embedding the local HTTP layer)

### Move into `mesh-client-core`

- discovery and scoring logic for public meshes
- invite-token private join flow
- passive/client connection lifecycle
- route-table fetch logic
- model listing logic derived from the client path
- tunneled inference request/response handling
- streaming decode and cancellation
- reconnect/failover behavior

### Keep out of `mesh-client-core`

- CLI parsing and command dispatch
- listeners on `:9337` and `:3131`
- embedded console assets/UI
- plugin hosting and MCP bridge
- local model launch/runtime orchestration
- subprocess lifecycle management
- desktop service/install assumptions

### Move into `mesh-host-core`

- host participation logic needed for an embedded macOS app
- local model load/unload
- embedded serving lifecycle
- mesh advertisement of locally served models

### Rewrite / separate carefully

Anything currently shaped around launching helper binaries or binding localhost ports should be considered desktop-runtime behavior, not SDK behavior.

## Protocol / compatibility requirement

The embedded SDK client should be treated as a first-class protocol client.

That means:

- preserve compatibility with normal `mesh-llm` nodes
- avoid accidental coupling to CLI-only behavior
- keep the SDK path compatible with the supported client/passive mesh behavior
- if protocol changes become necessary, document compatibility impact explicitly

## Packaging strategy

Suggested repo/workspace shape:

```text
/mesh-llm
/mesh-client-core
/mesh-host-core
/mesh-ffi
/bindings/swift
/bindings/jni
/bindings/kotlin
/tests/ffi-smoke
```

If churn needs to be minimized, extraction can happen incrementally inside the existing workspace first, then moved into separate crates once the boundaries are stable.

## Smoke tests

Smoke tests are required from the start. These should primarily validate binding and lifecycle integrity, not deep model behavior.

### Rust core tests

- private join flow
- public discovery / auto-join flow
- route-table fetch
- model listing
- one short streamed request
- cancellation
- reconnect/failover
- error mapping

### FFI smoke tests

- create/free client handle
- register callback
- join private test mesh
- list models
- start one short stream
- receive delta + complete
- cancel without crash/leak
- repeated create/destroy loop

### Swift smoke tests

Must run on `macOS`.

- load framework
- create client
- join private test mesh
- list models
- stream one short completion
- destroy cleanly

When host mode exists on `macOS`:

- create host
- start embedded serving
- serve one request
- stop cleanly

### JNI/Kotlin smoke tests

Must run on `macOS` and `Linux`.

- load JNI/native library
- create client
- join private test mesh
- list models
- stream one short completion
- cancel one stream
- destroy cleanly

These tests should be host-runnable without requiring an Android emulator as the primary smoke path.

### Required

- `macOS`
  - Rust core tests
  - FFI smoke tests
  - Swift smoke tests
  - JNI/Kotlin smoke tests

- `Linux`
  - Rust core tests
  - FFI smoke tests
  - JNI/Kotlin smoke tests

### Later / secondary

- iOS simulator package/build verification
- Android packaging/AAR verification
- Android device/emulator integration coverage

The primary binding smoke path should remain runnable on `macOS` and `Linux` hosts.

### Phase 1: scope and boundaries

- confirm SDK is native embedded only
- confirm no browser SDK path
- confirm `iOS`/`Android` are client-only initially
- confirm `macOS` can support host mode later

### Phase 2: extract `mesh-client-core`

- isolate passive client behavior
- strip listener/CLI/runtime assumptions
- expose a Rust-native client API

### Phase 3: add `mesh-ffi`

- design stable handle-based ABI
- add FFI smoke tests
- lock down event and memory ownership model

### Phase 4: first native binding

- ship Swift first or Kotlin first, but only one at first
- prove join/list/stream lifecycle end to end

### Phase 5: second native binding

- add the other language binding
- keep API concepts aligned across platforms

### Phase 6: macOS host mode

- add `mesh-host-core`
- make embedded local serving work in a macOS app profile
- keep App Store-safe requirements in view

### Phase 7: protocol hardening

- treat embedded SDK clients as a permanent supported client type
- document and test compatibility implications

## Suggested acceptance criteria for the first real milestone

- a Swift app on the latest `macOS` can create an embedded client, join a private mesh, list models, and stream one completion without localhost ports or a sidecar
- a Kotlin/JNI test on `macOS` and `Linux` can do the same
- the same Rust core supports public discovery + auto-join
- smoke tests run in CI on `macOS` and `Linux`
- no browser path is required for any of the above

## Open questions

- whether `mesh-host-core` should embed llama.cpp directly or introduce another in-process serving boundary for App Store-safe macOS
- whether the first native binding should be Swift or Kotlin
- whether to keep the new crates in the existing workspace immediately or do a staged internal extraction first

ndizazzo:
This branch adds an embedded Mesh SDK stack for app integrations.

After this change, app runtimes can embed a Mesh client directly instead of treating Mesh as a localhost-only sidecar. The branch introduces a public Rust SDK surface, a UniFFI layer for native bindings, and native Swift and Kotlin packages built on top of the same client core.

This branch also adds CI coverage for the embedded client path, so SDK consumers are validated against a real served model rather than only through the CLI and server workflows.

Closes #188

## Architecture
The branch defines the embedded SDK layering explicitly:
- `mesh-client` is the internal client implementation crate.
- `mesh-api` is the public Rust SDK for app integrations such as Jan.
- `mesh-api-ffi` is the UniFFI/native bindings layer.
- Swift and Kotlin bindings sit on top of `mesh-api-ffi`.

That gives the repo one shared client core with separate product-facing surfaces for Rust and native mobile/desktop consumers.

## Protocol
This branch is not intended to introduce a breaking mesh protocol change.
- Existing protocol compatibility coverage remains in place.
- The embedded SDK work is about packaging and client integration surfaces, not changing mesh wire semantics.

## Validation
Validated locally:
- `cargo test -p mesh-llm --test protocol_compat_v0_client`
- `cargo test -p mesh-llm --test protocol_convert_matrix`
- `scripts/ci-native-sdk-smoke.sh target/debug/mesh-llm llama.cpp/build/bin ~/.models/SmolLM2-135M-Instruct-Q8_0.gguf`
- `scripts/ci-kotlin-sdk-smoke.sh target/debug/mesh-llm llama.cpp/build/bin ~/.models/SmolLM2-135M-Instruct-Q8_0.gguf`
- `scripts/ci-swift-sdk-smoke.sh target/debug/mesh-llm llama.cpp/build/bin ~/.models/SmolLM2-135M-Instruct-Q8_0.gguf`

CI notes:
- `native_sdk_smoke` and `kotlin_sdk_smoke` run after `inference_smoke_tests`.
- The legacy `macos` lane is documented as a Linux cross-build lane and does not require a macOS runner.
- `swift_sdk_smoke` runs on `macos-latest` after `inference_smoke_tests`.


i386:
@copilot apply changes based on the comments in [this thread](https://github.com/Mesh-LLM/mesh-llm/pull/233#pullrequestreview-4095777539)

i386:
@copilot apply changes based on the comments in [this thread](https://github.com/Mesh-LLM/mesh-llm/pull/233#pullrequestreview-4096196482)

i386:
@copilot resolve the merge conflicts in this pull request

i386:
@copilot resolve the merge conflicts in this pull request

ndizazzo:
So far it looks like we have a good portion of this already working, but we haven't reached the embedded SDK boundary yet. The overall structure of the branch looks right as far as I can tell:

* the crate split is in place  
* Rust, FFI, and native layers are wired up  
* the public API surface is mostly defined  
* there's enough end to end scaffolding to make the intended direction clear  

On the Rust side, everything I checked builds and tests cleanly. `cargo build -p mesh-llm --bin mesh-llm` passed, along with the targeted protocol tests, mesh-client tests, and mesh-api-ffi tests.

The Swift path is also usable locally. The XCFramework builds, a Swift exemplar project builds and runs, and the example and CI flow look like they can exercise a real request when the fixture environment is set up as expected.

The live smoke fixture works in bridged mode as well. CI and the native smoke setup can start mesh-llm, discover a model, send a request, and stream a completion through the SDK stack. As it stands, this branch already functions as a prototype native client for interacting with a local Mesh-LLM API backed runtime, and that part is solid.

---

### Remaining workstreams

1. **Make `mesh-client` the real runtime**  
   Remove the dependency on `MESH_CLIENT_API_BASE` for the core happy path. `join`, `list_models`, `chat`, `responses`, `cancel`, `status`, `disconnect`, and `reconnect` should operate on real embedded transport and session state, not placeholder flags or fixture-only HTTP bridging.

2. **Restore shared core fidelity with `mesh-llm`**  
   Reconcile descriptor inference, capability inference, protocol definitions, and lifecycle semantics so the extracted crates aren't simplified forks. The embedded core should preserve the same behavior the host relies on.

3. **Make Swift and Kotlin honest wrappers over the Rust core**  
   Swift shouldn't silently fall back to a fake implementation for normal package consumers. Kotlin should ship a complete generated binding surface that matches the Rust FFI exactly, with an install and build path that works in a real external app.

4. **Make CI validate the shipped product boundary**  
   SDK changes should always trigger SDK validation. CI should prove the published consumer path, use the correct runtime dependencies, and test real embedded behavior instead of only a repo local smoke fixture.

michaelneale:
Reviewed this — direction looks good and it's mostly additive, so low risk to existing mesh behavior. Two things worth raising before merge:

## Proto duplication looks unintentional

`mesh-llm/proto/node.proto` and `mesh-client/proto/node.proto` both declare `package meshllm.node.v1;`, share most message names (`GossipFrame`, `PeerAnnouncement`, `ServedModelDescriptor`, `NodeConfigSnapshot`, `ConfigPush`, …), and both `build.rs` files generate into the same Rust module path. So they're the same schema, just duplicated.

The two copies have already drifted:
- `mesh-client`'s copy is missing `HardwareInfo`, `GpuInfo`, `SignedNodeOwnership`, `ConfiguredModelRef` — the attested-identity + hardware-gossip work that's on main.
- Field 28 on `NodeAnnouncement` disagrees: `mesh-llm` has `SignedNodeOwnership owner_attestation = 28`, `mesh-client` has `string owner_id = 28`. Same wire tag, different types — peers built from one decode that field as garbage against peers built from the other.

Fix is structural rather than a resync: point both crates at a single `.proto` (either shared at workspace root, or one crate's `build.rs` reads the other's file). Makes it impossible to drift again.

## Kotlin AAR ships stub bindings, not the real ones

`sdk/kotlin/src/main/kotlin/uniffi/mesh_ffi/MeshFfi.kt` is 67 lines of interface declarations with no implementation. The real generated UniFFI bindings (2111 lines) live under `sdk/kotlin/example/example-jvm/`. The packaged AAR picks up the stub, so consumers installing the artifact would get `MeshClientHandleInterface` with nothing implementing it and no JNI loader wired up.

`kotlin_sdk_smoke` passes because it exercises the example app, which has its own copy of the real bindings — not the library AAR consumers would actually depend on. Worth confirming which artifact is intended to ship.

Everything else (forked router/nostr/affinity copies in `mesh-client` vs `mesh-llm`, dep allowlist drift, a couple of FFI `unwrap`s) is follow-up material, not blockers.

PR #242, opened 2026-04-10
Fix inconsistent peer views: PeerDown verification and transitive peer stability

michaelneale:
Nodes in the public mesh sometimes disagree about which peers are alive. A peer that's healthy and reachable from some nodes gets removed by others, causing inconsistent routing and repeated "reported dead" churn.

Three root causes addressed:

### 1. PeerDown blindly trusted when no connection exists

When node A reports peer C dead to node B, and B has no stored QUIC connection to C (common for relay-only and transitive peers), B was auto-confirming the death without any verification. Now B:
- Checks if C was seen recently via inbound gossip — if so, ignores the report
- If B knows C's address but has no connection, attempts to reach C (5s timeout) before confirming
- Only trusts the reporter for completely unknown peers

**Before:**
```
⚠️  Peer 3f14379340 reported dead by 7259e58b87, confirmed, removing  ← no verification
```

**After:**
```
ℹ️  Peer 3f14379340 reported dead by 7259e58b87 but seen recently (inbound alive), ignoring
ℹ️  Peer 3f14379340 reported dead by 7259e58b87 but we reached them, keeping
```

### 2. Transitive peers flicker every 6 minutes

Peers learned through gossip (not directly connected) had `last_seen` set only on initial add and never refreshed. After `PEER_STALE_SECS * 2` (6 min) they'd get pruned, then re-added on the next gossip cycle — flickering forever.

Now `last_seen` is refreshed whenever a bridge peer mentions the transitive peer. Since `collect_announcements` already filters out stale peers before gossiping, presence in gossip is proof-of-life from the bridge's perspective.

### 3. PeerDown cascading past heartbeat leniency

The heartbeat gives peers 2 failures before removal (`💛 unreachable (1/2), will retry`), but an external PeerDown arriving during this grace period would bypass the leniency and immediately remove the peer. Now the PeerDown handler checks `last_seen` before acting, which covers the case where the peer is alive via inbound gossip even if outbound fails.

## Protocol

No wire format changes. These are local-only policy changes to how PeerDown claims are evaluated and how transitive peer freshness is tracked. Mixed-version meshes are unaffected — older nodes will continue with the previous behavior while newer nodes are more conservative about removals.


ndizazzo (commented):
Copilot picked up a lot of the potential issues with this. I think what the comments are getting at are that we should probably split PeerInfo.last_seen into two concepts: 

1) one field for directly observed liveness and one for transitive mention freshness, then use the direct timestamp for PeerDown silencing and for collect_announcements() / pruning decisions so hearsay don't count alive peers. 
2) Also make the connect_mesh recovery path re-check state.connections under the lock before inserting and creating dispatch_streams, and a test proving a transitively learned peer stays visible while mentions continue, but expires once mentions stop.

michaelneale:
need to hand test this more

mesh-llm/src/mesh/mod.rs
```diff
@@ -1152,8 +1152,15 @@ pub struct PeerInfo {
     /// Models this node has requested the mesh to serve
```
i386, line 1152: @michaelneale can you get an agent to split this file its horrifically long now

michaelneale: yeah taking a look

PR #248, opened 2026-04-11
feat(hardware): add hardware information enrichment

### LINKED ISSUE #136: Add hardware runtime status enrichment without changing model metadata sources
## Summary
This issue narrows the earlier "local model scan and status enrichment" idea down to hardware runtime status only.

It is a follow-on from #132, but intentionally excludes any new model metadata or model-size source of truth in /api/status.

Related tracker: https://github.com/michaelneale/mesh-llm/issues/132

## Why this is narrowed
The broader proposal in PR #130 bundled two different ideas together:
- richer hardware/runtime reporting
- direct node-native model scan payloads exposed through status

We want to keep the first and reject the second.

Reason:
- main already has a cache/inventory-based model metadata and model-size pipeline
- introducing raw `model_scans` and `model_sizes` in `/api/status` would create a second overlapping source of truth
- later UI/config work should continue to use the existing inventory/cache-backed model metadata path rather than pivoting to a new status-fed model scan path

So this issue keeps only the hardware/runtime reporting improvements.

## Goals
- Improve per-GPU hardware visibility in runtime/status surfaces.
- Surface reserved VRAM so reported capacity better reflects reality.
- Preserve the current model metadata architecture.
- Keep this as a low-risk read-only backend change.

## Non-goals
- No `model_scans` in `/api/status`.
- No new `model_sizes` status field as a primary data source.
- No replacement of the current cache/inventory-backed model metadata flow.
- No config editing work.
- No auth or authorization changes.

### Hardware detection
Extend hardware detection to retain per-GPU reserved memory where available.

For NVIDIA systems this can come from:
- `nvidia-smi --query-gpu=memory.total,memory.reserved --format=csv,noheader,nounits`

For other platforms, use sensible defaults or omit reserved values when unavailable.

### Status payload
Enrich GPU entries in runtime status with:
- `name`
- `vram_bytes`
- `reserved_bytes`
- optional `bandwidth_gbps`

This remains strictly hardware/runtime-oriented.

### UI effect
Allows console and future config surfaces to show a more honest view of usable GPU capacity without changing how model metadata is sourced.

## Suggested implementation areas
Likely code areas:
- `mesh-llm/src/hardware.rs`
- `mesh-llm/src/api.rs`
- tests for GPU parsing and JSON serialization
- docs describing `reserved_bytes` and hardware reporting semantics

## Suggested tests
- parse reserved VRAM from NVIDIA output correctly
- reserved VRAM defaults cleanly when unavailable
- per-GPU ordering is preserved even with partial/unparsable values
- `/api/status` serializes `reserved_bytes` correctly
- older consumers tolerate the additive field

## Acceptance criteria
- Hardware detection can populate per-GPU reserved VRAM where supported.
- `/api/status` includes `reserved_bytes` in GPU entries.
- The change is additive and read-only.
- No new raw model-scan or model-size status source is introduced.
- Docs/tests explain the narrowed scope clearly.

ndizazzo:
## Summary

- Add end-to-end GPU TFLOPS support so benchmark outputs, cached fingerprints, node runtime state, mesh gossip, protobuf transport, and `/api/status` all carry `compute_tflops_fp32` and `compute_tflops_fp16` consistently.
- Update all four platform fingerprint binaries to emit TFLOPS data when available, keep missing values omitted rather than serialized as `null`, and preserve backward compatibility for older fingerprint and gossip data.
- Finish the branch with the repo-side cleanup required to satisfy final verification, including clippy, fmt, tests, and full build validation.

### Benchmarks and hardware

- Added TFLOPS fields to Rust benchmark and hardware structs.
- Added Apple Silicon lookup-based TFLOPS fallback and NVIDIA derivation in Rust.
- Updated Swift, CUDA, HIP, and Intel benchmark binaries to emit TFLOPS JSON fields.
- Preserved cache compatibility with older fingerprint JSON that lacks TFLOPS fields.

### API, runtime, and mesh propagation

- Exposed per-GPU TFLOPS in `/api/status` with `skip_serializing_if` behavior so absent values stay omitted.
- Wired benchmark results into node runtime state so bandwidth and TFLOPS are stored together from the same benchmark run.
- Added TFLOPS fields to mesh gossip structs and v0 JSON compatibility paths.
- Added protobuf fields and conversion plumbing so TFLOPS and memory bandwidth survive v1 transport round-trips.

### Validation and cleanup

- Added/updated tests for benchmark parsing, API serialization, gossip/proto round-trips, and runtime result propagation.
- Cleaned up repo-wide clippy findings required for final verification on this branch.

## Notes

- CUDA, HIP, and Intel benchmark source changes were validated by code review in this environment, but their native compiler toolchains were not installed locally.
- Final verification approved the branch under the user-authorized scope that included the code-quality cleanup needed to pass F2.

Closes #136


i386:
@ndizazzo lmk when the changes we discussed are up, so I can give this a big +1 and you can merge :)

michaelneale:
Nice - I am guessing them being options means this works fine in mixed mode scenarios where no one yet has the same versions?

i386:
This is really close! 

While you're here, could you make a JSON formatter for `mesh-llm gpus --json` and `mesh-llm gpus benchmark --json` ?

i386:
I think we should store the fingerprint in `$HOME/.cache/mesh-llm/benchmark-fingerprint.json` rather than `$HOME/.mesh-llm/benchmark-fingerprint.json`. We are using `$HOME/cache` for other non-persistent information.

mesh-llm/benchmarks/membench-fingerprint-intel.cpp
```diff
@@ -34,6 +34,18 @@ static double ratedBandwidthGBps(const sycl::device& dev) {
     }
```
i386, line 34: any reason why these files are not under `mesh-llm`?

mesh-llm/docs/DESIGN.md
```diff
@@ -233,7 +233,7 @@ trait Collector {
 | `DefaultCollector` | Linux AMD | `/sys/class/drm`, `rocm-smi` |
 | `TegraCollector` | Jetson / Tegra | sysfs + `tegrastats` |
 
-`survey()` calls all applicable collectors and returns a `HardwareSurvey` with `gpu_name`, `gpu_vram` (per-GPU bytes), `vram_bytes` (total), `hostname`, and `is_soc`.
+`survey()` calls all applicable collectors and returns a `HardwareSurvey` with `gpu_name`, `gpu_vram` (per-GPU bytes), `gpu_reserved` (per-GPU reserved or unavailable bytes when known), `vram_bytes` (total), `hostname`, `is_soc`, and per-device `GpuFacts` entries. Benchmark-derived memory-bandwidth and compute-throughput hints are attached later when cached or freshly measured results are available.
```
i386: Somewhat related to this PR but will the default privacy still be to not show hardware info? I feel like this should be visible until a user configures it off in `config.toml`

Image

i386 (changes requested):
LGTM, minor comments left on PR.

mesh-llm/benchmarks/membench-fingerprint-intel.cpp
ndizazzo: Nope! Not other than that's where the swift memory benchmark file started, so the rest got homed in the same spot. I can move it there since it makes sense

mesh-llm/docs/DESIGN.md
ndizazzo: @i386 Yeah, it was one of the very first things I did - I wanted to see the hostname of a node but didn't want anyone else to see my machine's hostname, so I had it disabled by default for privacy. 

Generally I prefer private by default... the idea was you may be able to identify a node based on its hardware fingerprint and that might be undesirable. But it's starting to make more and more sense to just expose this stuff, since it has an implication on the performance of the mesh.

i386: @ndizazzo agree on mesh performance. and if someones bugging about privacy they can just disable.

ndizazzo: Let's invert this logic in a follow-up

i386: @copilot create an issue for this

PR #260, opened 2026-04-13
feat: add fail-closed pinned GPU assignment for local config

ndizazzo:
## Summary
- Validate two-node mixed-version mesh behavior between the local macbook and Linux CUDA host with 2 GPUs.
- Confirm that a current local node (`0.59.0`) can join an older Linux host (`0.58.0`) as a standby worker with real VRAM, route inference successfully, and preserve expected `/api/status` fields.
- Upgrade the Linux host to the pushed `feature/pinned-gpu-config` branch, rebuild/restart both nodes, and verify the same two-node flow on matching current versions.

### Mixed-version validation
- Built current local branch on macbook with:
  - `just build`
  - `just bundle`
- Built an older Linux checkout from a pinned `v0.58.0` clone.
- Started Linux on `0.58.0` as host and joined from macbook on `0.59.0`.
- Verified `/api/status` on both sides:
  - macbook reported:
    - `version: 0.59.0`
    - `node_status: Standby`
    - `is_host: false`
    - `is_client: false`
    - nonzero `my_vram_gb`
    - peer `version: 0.58.0`
  - Linux reported:
    - `version: 0.58.0`
    - `node_status: Serving`
    - `is_host: true`
    - `is_client: false`
    - peer count `1`
- Verified `/v1/models` from the mesh API.
- Verified inference from both sides with `/v1/chat/completions`; both returned `OK`.

### Post-upgrade validation
- Pushed branch:
  - `origin/feature/pinned-gpu-config`
- Reset Linux to the pushed branch and reran the repo’s e2e workflow:
  - `./.sisyphus/e2e-test/mesh-test.sh --model Qwen2.5-3B`
- Script completed successfully, including:
  - local rebuild
  - linux sync + rebuild
  - process cleanup
  - port checks
  - owner-key verification
  - invite/join flow
  - peer discovery
  - inference check
- Re-verified `/api/status` after upgrade:
  - both sides reported `version: 0.59.0`
  - shared `mesh_id`
  - peer count `1`
  - linux remained `Host`
  - macbook remained `Standby`
  - `is_client: false` on both
  - `peers[*].version` populated symmetrically as `0.59.0`
- Re-verified `/v1/chat/completions` from both sides; both returned `OK`.

## Notes
- During mixed-version testing, Linux's older build reported the newer peer with `version: null`, while the upgraded current-current run reported peer versions symmetrically. This was the main observed field difference before and after upgrade.


mesh-llm/proto/node.proto
```diff
@@ -231,6 +231,7 @@ message NodeModelEntry {
   string model = 1;
```
i386, line 237: hmm @ndizazzo I think these need to be the model identity struct not strings

ndizazzo: I'm not so sure about this - it mixes two concepts that we have: `authored configuration intent for the node` vs `completed and resolved model entity`. `NodeModelEntry` is only really meant for what a node operator configures, not a concrete resolution at this point yet.

If we were to do anything, it would be to declare something like:

```proto
message ConfiguredModelRef {
  string declared_ref = 1;
  optional string source_kind = 2;
  optional string revision = 3;
}

message NodeModelEntry {
  ConfiguredModelRef model = 1;
  optional ConfiguredModelRef mmproj = 2;
  optional uint32 ctx_size = 3;
  optional string gpu_id = 4;
}
```

But I would only really do it when a string isn't serving us well enough...

ndizazzo: @copilot what do you think about this?

i386 (approved):
LGTM

ndizazzo:
@i386  hold up - do you think we should do `ConfiguredModelRef` now? Proto changes 👎🏻 if the eventual shape should be to have a Ref message

michaelneale (approved):
if you can fix up tests.rs in mesh/ dir - so they are not duplicated/orphaned, then good to go.

PR #266, opened 2026-04-13
Fix tunnel listener fd leak on peer departure

michaelneale:
Tunnel listeners and their accept-loop tasks now get torn down when peers disconnect, fixing a file descriptor leak that caused `too many open files` on Linux nodes with peer churn.

## What changed

`watch_peers()` created a `TcpListener` and spawned an accept-loop task for each peer, but never cleaned them up when the peer left. The `tunnel_ports` HashMap only grew — entries for departed peers stayed forever, and their tasks sat parked in `accept()` holding an open fd each.

On Linux with the default `ulimit -n 1024`, mesh nodes with enough peer churn (e.g. `--auto` with Nostr rediscovery) would exhaust file descriptors.

The fix tracks a `JoinHandle` alongside each tunnel port. On every peer-change notification, the live peer set is diffed against the tunnel map. Tasks for departed peers are aborted (dropping their `TcpListener` and reclaiming the fd), and stale entries are removed.

## Validation

- `cargo check` clean, no warnings
- Existing tunnel relay tests pass
- No new dependencies

### DIFF
--- a/mesh-llm/src/network/tunnel.rs
+++ b/mesh-llm/src/network/tunnel.rs
@@ -21,6 +21,14 @@ use std::time::Duration;
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::Mutex;
+use tokio::task::JoinHandle;
+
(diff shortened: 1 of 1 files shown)


i386 (approved):
LGTM

PR #269, opened 2026-04-13
Harden Nostr discovery: validate tokens, candidate fallback

michaelneale:
Simpler alternative to #268.

## What changed

Malformed Nostr listings (bad invite tokens) used to poison auto-join — the client would pick the top-ranked mesh, fail to decode its token, and give up. Now:

1. **`discover()` validates invite tokens before listings enter the candidate pool** — bad tokens are logged and skipped
2. **All candidates are queued**, not just the #1 pick — if the top mesh is unreachable, the client falls back to the next candidate
3. **`Node::decode_invite_token()`** extracted so tokens can be validated without connecting

## What's different from #268

- No typed `InviteTokenError` / `thiserror` — uses `anyhow::context()` (simpler, same result)
- No field-length validation (name, region, model list size) — those are nice-to-have but not the fix
- No `invalid_invite_token_message` branching at every join callsite — once `discover()` rejects bad tokens, downstream join loops won't hit them
- ~90 lines changed vs ~300

## Files

- `mesh/mod.rs` — `decode_invite_token()` + use it in `join()`
- `network/nostr.rs` — validate tokens in `discover()`, fix integration test to use valid tokens
- `runtime/mod.rs` — queue all candidates in client `--auto` path and MCP join path

### DIFF
--- a/mesh-llm/src/mesh/mod.rs
+++ b/mesh-llm/src/mesh/mod.rs
@@ -1763,9 +1763,17 @@ impl Node {
         });
     }
 
+    /// Decode an invite token into an [`EndpointAddr`] without connecting.
+    /// Returns `Err` if the token is not valid base64 or not valid JSON.
+    pub fn decode_invite_token(invite_token: &str) -> Result<EndpointAddr> {
+        let json = base64::engine::general_purpose::URL_SAFE_NO_PAD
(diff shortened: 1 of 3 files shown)


i386 (approved):

i386:
LGTM

PR #271, opened 2026-04-13
feat: TTFT inference performance tracker

michaelneale:
## What

Nodes now locally track **time-to-first-token (TTFT)** for every inference request they proxy, per model and per target host.

This gives visibility into which models/hosts are fast vs slow, and provides the data foundation for smarter routing — preferring fast hosts and avoiding slow ones.

### How it works

- Timing is captured in the proxy's `route_local_attempt()` and `route_remote_attempt()` — the interval between sending the request and receiving the first response bytes
- Only successful (2xx) responses are recorded
- Each (model, target) pair keeps a ring buffer of the last 8 samples
- Reported TTFT is the **best of the last 3** samples — this filters out transient spikes from cold KV cache, contention, or network blips
- Entries expire after 30 minutes with no updates; hard cap of 512 entries

### API

Exposed on `/api/status` as `inference_perf`:
```json
{
  "inference_perf": {
    "models": [
      {
        "model": "Qwen3-8B-Q4_K_M",
        "best_ttft_ms": 200,
        "best_target": "local:8080",
        "total_samples": 5,
        "targets": [
          { "target": "local:8080", "best_ttft_ms": 200, "samples": 3 },
          { "target": "a1b2c3d4e5", "best_ttft_ms": 1200, "samples": 2 }
        ]
      }
    ]
  }
}
```

### What's next (not in this PR)

- **Feed TTFT into routing**: `pick_model_classified()` already accepts a `tok_s` param (always 0.0 today). Use observed TTFT to prefer faster models for auto-routing.
- **Target ordering**: When multiple hosts serve the same model, prefer lower-TTFT hosts over hash-based selection.
- **Background probes**: Send tiny synthetic requests to new hosts on discovery to get a TTFT baseline without waiting for real traffic.

## Scope

Purely local, no protocol changes, no gossip additions. Safe for mixed-version meshes — older nodes just won't have the `inference_perf` field in their status.

### DIFF
--- a/mesh-llm/src/api/mod.rs
+++ b/mesh-llm/src/api/mod.rs
@@ -37,7 +37,7 @@ use self::status::{
 };
 use crate::inference::election;
 use crate::mesh;
-use crate::network::{affinity, nostr, proxy};
+use crate::network::{affinity, nostr, perf, proxy};
 use crate::plugin;
 use std::sync::Arc;
 use tokio::net::{TcpListener, TcpStream};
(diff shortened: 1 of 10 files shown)


ndizazzo:
@michaelneale love this - I think this is something we want to add to our list of things to discuss, around what data can we emit around the mesh, so eventually we can strategically collect for health + metrics + etc.

michaelneale:
Closing as this needs more desighn from linked issue - branch can stay as a draft.

PR #283, opened 2026-04-14
Fix plugin channel message flooding causing OOM

michaelneale:
## What changed

Mesh nodes stay stable at ~70MB instead of growing to multi-GB and crashing. The Fly console and local clients no longer OOM.

### The problem

Plugin channel messages (used by the blackboard shared-text feature) were being amplified exponentially across the mesh. A handful of original sync messages generated 30,000+ receives per 10 seconds, spawning 130,000+ tokio tasks, each opening a QUIC bidi stream. Memory grew at ~50MB/s.

### Root cause

The message dedup window used a fixed 4096-entry count cap. At high message rates, entries were evicted in ~1 second. When the same message arrived via an alternate path a few seconds later, it was no longer in the dedup set and got re-broadcast, creating a self-sustaining amplification loop.

### Fixes

1. **Time-based dedup** — message IDs are now retained for 120 seconds (with a 100K hard cap). This ensures delayed duplicates from alternate paths are always caught.

2. **Stop re-flooding plugin messages** — broadcast messages are delivered locally only (the originator already sent to all direct connections). Targeted messages are forwarded only to the specific target peer, not flood-broadcast to all connections.

3. **Clients skip connecting to other clients** — in gossip peer discovery, client nodes no longer initiate connections to other clients. Clients only need connections to hosts/workers serving models.

### Results

| Metric | Before | After |
|--------|--------|-------|
| RSS after 3 min | 2.3 GB (crash) | 69 MB (stable) |
| Plugin msgs received/10s | 30,000+ | 10,000 (dedup+drop) |
| Spawned broadcast tasks/10s | 137,000 | 0 |
| Rebroadcasts/10s | 22,000 | 300 (targeted only) |
| Connections (client, 21 peers) | 16 | 9 |

Compatible with older nodes — fixed nodes simply stop contributing to the storm. As more nodes upgrade, the residual flood from old nodes will diminish.

### DIFF
--- a/mesh-llm/src/mesh/gossip.rs
+++ b/mesh-llm/src/mesh/gossip.rs
@@ -253,17 +253,25 @@ impl Node {
         }
 
         if discover_peers {
-            for (addr, _) in &their_announcements {
+            let my_role = self.role.lock().await.clone();
+            for (addr, ann) in &their_announcements {
                 let peer_id = addr.id;
-                if peer_id != self.endpoint.id() {
-                    let has_conn = self.state.lock().await.connections.contains_key(&peer_id);
-                    if !has_conn {
-                        if let Err(e) = Box::pin(self.connect_to_peer(addr.clone())).await {
-                            tracing::debug!(
-                                "Could not connect to discovered peer {}: {e}",
-                                peer_id.fmt_short()
-                            );
-                        }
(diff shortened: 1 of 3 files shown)


ndizazzo (approved):
🫠

i386 (approved):
Nasty... good find.

PR #290, opened 2026-04-15
Improve topology visualization layout and zoom behavior

i386:
## Summary
Users can now explore the topology view with more stable, readable node placement and a single dark visual treatment.

- remove the light-theme variant from the new topology visualization
- keep zoom centered on the pointer for wheel zoom and centered in the viewport for +/- controls
- keep background effects, halo, grid, and label aligned with the current node while panning and zooming
- place nodes with stable random positioning, bias radial distance from `latencyMs`, and reduce overlap at the default zoom level

image

## Validation
- `just ui-dev`
- `just build` was started, but the full run was still in the first-time `llama.cpp` native build path in this worktree when work finished

### DIFF
--- a/mesh-llm/ui/src/features/dashboard/components/topology/MeshTopologyDiagram.tsx
+++ b/mesh-llm/ui/src/features/dashboard/components/topology/MeshTopologyDiagram.tsx
@@ -46,8 +46,6 @@ type ScreenNode = RenderNode & {
   px: number;
   py: number;
 };
-
-type ResolvedTheme = "light" | "dark";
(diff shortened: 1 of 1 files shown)


michaelneale (approved):

dsowsy:
Will there be a way to toggle to get the mesh overview back? This seems less informative and bad in light mode. 
2e864b0e-479f-41fa-af8e-040f3dc54c6f

i386:
@dsowsy it didn't scale very well for the public mesh. We could bring the old visualization back?

Issue #278, opened 2026-04-14
draft: Separate configuration communication channel


ndizazzo:
# Communication Investigation

## Executive summary

Today, mesh-llm has one public peer-to-peer QUIC control plane that carries multiple kinds of traffic: gossip, route-table requests, tunnel-map exchange, peer liveness messages, plugin traffic, and now owner-gated config subscribe/push streams. Discovery is separate and uses Nostr JSON events. Mixed-version compatibility is preserved by negotiating either `mesh-llm/1` (protobuf) or `mesh-llm/0` (legacy JSON/raw payloads).

That means the codebase already has some of the primitives needed for owner-to-node control, but they are not isolated enough for the direction you described. The strongest current building blocks are owner attestation, trust policy, signed config push with CAS semantics, long-lived config watch streams, and local runtime model load/unload APIs. The clean next step is not to add more stream IDs to the existing mesh connection. It is to introduce a second, private owner-control plane with its own QUIC endpoint and ALPN, direct owner-attestation handshake, protobuf-only command set, and no gossip/Nostr/public API multiplexing.

### QUIC

The main node-to-node transport is iroh QUIC. The preferred ALPN is `mesh-llm/1`, which uses protobuf framing for the control-plane messages, and the compatibility fallback is `mesh-llm/0`, which preserves the legacy JSON/raw formats. This is defined in `mesh-llm/src/protocol/mod.rs` and described in `mesh-llm/docs/message_protocol.md` and `mesh-llm/docs/DESIGN.md`.

On the public mesh plane, one QUIC connection is multiplexed by 1-byte stream type prefixes. The documented stream set in `mesh-llm/docs/message_protocol.md` covers:

- `0x01` gossip
- `0x02` raw RPC tunnel
- `0x03` tunnel map
- `0x04` raw HTTP tunnel
- `0x05` route-table request/response
- `0x06` peer-down broadcast
- `0x07` peer-leaving broadcast
- `0x08` blackboard
- `0x09` plugin channel
- `0x0a` plugin bulk transfer

The code has since grown two additional owner-gated config streams in `mesh-llm/src/protocol/mod.rs`:

- `0x0b` `STREAM_CONFIG_SUBSCRIBE`
- `0x0c` `STREAM_CONFIG_PUSH`

That doc/code mismatch matters. The code already treats remote config as part of the same QUIC mesh plane, even though the published protocol doc still presents the older stream map.

### Protobuf

[...]

PR #293, opened 2026-04-15
Split config protobuf from mesh protocol types

i386:
## Summary
- move config sync messages into a new `mesh-llm/proto/config.proto`
- keep mesh control-plane protobuf types in `node.proto` and expose them through `crate::proto::mesh`
- update protocol, mesh, inventory, and test code to use the split generated modules
- refresh protocol and testing docs to describe separate mesh and config protobuf schemas

## Testing
- `just build`


ndizazzo (commented):
There's some stuff in #278 that I was hoping we'd get with this work that will set up and unlock some things down the road in the config protocol, like command execution so that I can trigger model refreshes. Would you mind taking a look at the capability diff from the issue and this PR. It also has words for owner attestation built in, which I think is important:

| Capability | Implemented? |
|---|---|
| Split config protobuf messages into `proto/config.proto` | Partial match — aligns with the spec’s direction of separating config concerns, but only at the schema level |
| Separate generated protobuf modules for mesh vs config (`crate::proto::mesh` / `crate::proto::config`) | Partial match — useful groundwork, not the separate communication channel the spec asks for |
| Keep mesh-state protobufs in `node.proto` | Partial match — organizationally aligned, but not transport isolation |
| Update code to use split mesh/config protobuf namespaces | Partial match — mechanical refactor supporting future separation |
| Update docs to describe split mesh/config protobuf schemas | Partial match — documentation alignment only |
| Document config streams `0x0b` / `0x0c` in protocol docs | Partial match — clarifies existing behavior, but does not create a new channel |
| Create a separate QUIC endpoint for owner-control traffic | No match — not implemented |
| Introduce a distinct ALPN like `mesh-llm-control/1` | No match — still uses `mesh-llm/1` |
| Isolate config traffic from the existing public mesh control plane | No match — config still rides the shared control plane via existing stream types |
| Add a direct owner-attestation handshake at connection setup | No match — current auth still depends on existing peer/gossip-derived owner state |
| Create a truly separate private owner-control plane | No match — this is the main missing requirement |
| Add a dedicated remote inventory refresh command on the new control path | No match — not implemented |
| Preserve protobuf-based config messages | Match — consistent with the spec’s proposed protobuf-only owner-control direction |
| Preserve config get/watch/apply message family semantics | Partial match — messages still exist, but they were reorganized rather than moved to a new isolated plane |
| Improve protocol/schema organization beyond what the spec strictly requires | Extra capability — helpful refactor, but not part of the spec’s core acceptance bar |
| Rename/repartition internal proto API surface for clearer ownership (`mesh` vs `config`) | Extra capability — internal cleanliness benefit not explicitly required by the spec |
| Refresh tests to use the split proto modules | Extra capability — maintenance benefit not central to the spec |

i386:
@ndizazzo will come back to this after I sort out new hf crate

ndizazzo (approved):

i386:
Looking at another way to do this

PR #294, opened 2026-04-15
chore(upgrade): bump CUDA image to 13.2 to match self-hosted runner

ndizazzo:


ndizazzo:
Downgraded the runner instead

PR #298, opened 2026-04-16
feature: revise node states to be simpler

### LINKED ISSUE #241: Simplify model for node states
## Goal
- **Primary outcome**
  - Operators and API/UI consumers see a smaller, clearer node state model that matches current product behavior.
  - `/api/status` and the web console no longer imply there is still a first-class “Idle” mode after plain `mesh-llm` was changed to print help and exit.
- **Non-goals**
  - Changing routing, election, or model assignment behavior.
  - Building a new automatic load/unload orchestration system.
  - Fully redesigning the runtime state model beyond removing or replacing the obsolete `Idle` label.
## Context
- **Existing capabilities**
  - Plain `mesh-llm` prints help and exits without binding `:3131`.
  - `/api/status` exposes a derived `node_status` string that the UI consumes directly.
  - Runtime state is already split across:
    - mesh role: `Client | Host | Worker`
    - runtime lifecycle: `starting | ready | stopped`
  - Other user-visible labels already exist, including `Standby`, `Serving`, `Serving (split)`, `Worker (split)`, and `Client`.
- **Current limitations**
  - `Idle` still exists as a derived label in API status logic and UI fallback logic.
  - That label describes a mode the product no longer exposes.
  - State semantics are spread across API derivation and UI fallback mapping, which makes them harder to reason about and easier to drift.
- **Why this is needed now**
  - Issue #123 calls out the mismatch directly.
  - The issue discussion already points toward a simpler model centered on `Client`, `Standby`, `Loading`, and `Serving` modes.
## Surfaces
- **Surface**
  - API
  - UI
  - internal status derivation
- **Entry points**
  - `GET /api/status`
  - Web console status rendering in `mesh-llm/ui/src/App.tsx`
- **Inputs**
  - Current node role
  - whether a model is loaded or assigned
  - peer presence
  - runtime readiness / startup state
- **Defaults**
  - Status is derived from existing runtime and topology facts.
## Behavior
- **Execution flow**
  1. The API derives node status from runtime and topology facts.
  2. If the node has an active concrete role or serving condition, the API returns that non-`Idle` label.
  3. If the node is connected but not actively serving, the API returns the chosen fallback label, expected to be `Standby` rather than `Idle`.
  4. The UI renders the API-provided status and aligns any local fallback logic to the same model.
  5. Tests and docs are updated to match the new semantics.
- **Validation**
  - Valid:
    - all emitted labels map to real current runtime conditions
    - no codepath emits `Idle` after the change
    - existing `Client`, `Serving`, and worker-specific behavior remain intact
  - Rejected:
    - reintroducing `Idle` without a concrete runtime condition
    - API and UI disagreeing on fallback status semantics
- **Error handling**
  - On failure, the API should still return a valid status payload using the supported label set.
  - No new user-visible error flow is expected.
  - Silent failures are not allowed.
## State & Data
- **New data introduced**
  - none
- **Existing data modified**
  - `/api/status.node_status` semantics
  - UI fallback mapping for labels, tooltips, and tone
  - tests and docs that refer to `Idle`
- **Persistence**
  - not required
- **Backward compatibility**
  - API consumers that currently match on the literal string `Idle` will need to accept the replacement label.
## Integration Points
- **Depends on**
  - `mesh-llm/src/api/mod.rs`
  - `mesh-llm/src/api/status.rs`
  - `mesh-llm/src/runtime/mod.rs`
  - `mesh-llm/src/mesh/mod.rs`
- **Affects**
  - `mesh-llm/ui/src/App.tsx`
  - `mesh-llm/ui/src/topology/layouts/types.ts`
  - API tests
  - UI tests
  - status-related docs
- **External systems**
  - none
### Interface Layer
- `/api/status` no longer returns `Idle`.
- The UI no longer renders `Idle` in labels, tooltips, or status tone logic.
### Application Layer
- Status derivation is simplified to match actual runtime behavior.
- UI fallback logic is aligned with API output.
### Core Logic
- No routing, election, or model-loading behavior changes.
- This is a semantics and presentation cleanup only.
### Data Layer
- none
### External / Infrastructure
- none
### Observability
- **Logs**
  - none required
- **Metrics**
  - none
- **Alerts**
  - none
## Edge Cases
- A node with no loaded model and no peers must no longer report `Idle`.
- A client-only node must still report `Client`, not fall through to the standby/default state.
- A split worker must keep its worker-specific status and not collapse into a generic fallback state.
- A node in runtime startup (`starting`) must not be mislabeled as `Idle`.
- UI fallback code must not continue to show `Idle` after the API stops emitting it.
## Constraints & Tradeoffs
- **Constraint**
  - There is no single first-class enum for all user-visible states today; the behavior is derived across API and UI code.
- **Tradeoff**
  - This change targets semantic cleanup, not a full state-model refactor.
- **Alternative considered**
  - Introduce a fully typed unified state model that also promotes `Loading` to a first-class state.
  - Deferred because the immediate problem is the stale `Idle` label, not the full architecture of runtime status.
## Acceptance Criteria
- [ ] Given a node with no loaded model and no peers, when `/api/status` is requested, then `node_status` is not `Idle` and uses the new fallback label.
- [ ] Given a connected client-only node, when `/api/status` is requested, then `node_status` remains `Client`.
- [ ] Given a split worker node, when `/api/status` is requested, then `node_status` remains worker-specific.
- [ ] Given the UI renders node status, when status data is displayed, then no visible label, tooltip, or tone branch depends on `Idle`.
- [ ] Given status-related tests are updated, when the suite runs, then assertions reflect the new state model and pass.
- [ ] Existing serving/client/worker status behavior does not regress.
## Rollout / Migration
- **Rollout strategy**
  - Ship as a normal code, test, and docs update.
- **Migration required**
  - Yes. Any consumer matching on the literal `Idle` string must be updated.
- **Backward compatibility**
  - Runtime behavior stays the same, but the `Idle` status string is removed or replaced.
## Notes
- `Idle` is not a standalone state type today; it is a computed/display label spread across API logic, UI fallbacks, and tests.
- `Loading` is also not currently a first-class node status; the runtime lifecycle still uses `starting`, `ready`, and `stopped`.
## Related
- Issue: https://github.com/michaelneale/mesh-llm/issues/123
- Discussion: https://github.com/michaelneale/mesh-llm/issues/123#issuecomment-4196251525

ndizazzo:
Users can now read node health with a much simpler live state model. The console shows only `Client`, `Standby`, `Loading`, and `Serving` for live members, while provider-backed suspended capacity appears in a separate `Wakeable Capacity` section instead of being mixed into topology or routing.

After writing up #241 and chatting with @michaelneale about some hosted GPU capacity in the mesh, I decided to extend this state idea a little bit to include the concept of "Wakable Capacity". I don't know yet the mechanism we would actually use to wake up a node (ie: send a AWS / GCP / Azure / whatever API call to boot a VM), but I pre-emptively planned for this in the state changes I was making.

## Summary

- Simplifies the live node display to the approved four-state model.
- Keeps wakeable provider-backed capacity visible, but separate from live peers.
- Preserves the existing routing and peer semantics, so sleeping or waking capacity does not become routable by accident.
- Updates the dashboard to show wakeable inventory as its own section.

## Screenshot
wakeable-capacity-8-hosts
image
(we can remove this / opt not to show it, but I mocked it up here for illustration purposes after our conversation)

## Architecture

- `node_state` and peer `state` are additive machine-readable fields.
- `node_status` stays as the Title Case compatibility alias for older consumers.
- Wakeable inventory uses a separate local seam, so it stays out of `peers[]`, live topology, peer counts, and host selection.

## Protocol

Mixed-version compatibility is preserved. Older peers still classify correctly, and wakeable inventory is not advertised as a live peer or protocol change. It is exposed as separate dashboard and status data only, not as routable capacity.

Closes #241


michaelneale (approved):
why not

PR #299, opened 2026-04-16
Require mesh-llm/1 for mesh joins

i386:
## Summary
mesh joins now use only the `mesh-llm/1` protobuf control-plane protocol.
This removes the legacy `mesh-llm/0` JSON compatibility path from dialing, acceptance, stream handling, and protocol docs.

## User Impact
Nodes running releases below `v0.54.0` will no longer be able to join the mesh, because `v0.54.0` was the first tagged release that contained the protobuf `mesh-llm/1` control-plane implementation.
Meshes are now protobuf-only on the control plane.

## What Changed
- remove `mesh-llm/0` ALPN fallback and legacy JSON control-frame handling
- delete the legacy v0 protocol module and compatibility-only tests/helpers
- update protocol and testing docs to describe `mesh-llm/1` as the only supported control-plane protocol

## Protocol
This is an intentional compatibility break for pre-`v0.54.0` nodes.
Releases below `v0.54.0` cannot negotiate the protobuf control plane and will not be able to join upgraded meshes.

## Validation
- `just build`
- `cargo test -p mesh-llm --lib`


michaelneale (approved):
I am ok with this, but need to update the web app on the main site co-ordinated with some other nodes during release - so if we can work that out? 

Also - how do we kind of test this all out?

i386:
> Also - how do we kind of test this all out?

Run a pre-0.54.0 node and this change. You won't see the new node in pre-0.54.0's mesh peer list.

ndizazzo (approved):
LGTM! My favourite activity is deleting code

Issue #256, opened 2026-04-12
feature: make mesh-llm app load aware to stop/start models depending on local usage


michaelneale:
We want to encourage people to not just run a client leeching off the network, but to contribute compute if they can. 

I think to make this better we can have mesh llm be aware when the user needs CPU/GPU themselves, and stop serving (RPC or llama serve) and/or use modern and platform specific equivalents of `nice` and `cgroups` to deprioritise that (in reality it is probably memory reservations so stop start may be fine). 

This may be most relevant to a map .app distribution. 

Good idea?

ndizazzo:
@michaelneale I think this is only one real angle of mesh expansion - in addition to a person's computer still being usable when running as a mesh node, I think more of a "reciprocity" mechanism would help the behavioural aspects.

**I think we have two options:**
1) If you're a client to a public mesh, your GPU automatically contributes 
2) Build out priority routing / higher queue priorities / better model access / reserved capacity 

Just my 0.02c, I think 2 is the right long-term direction for a public network, but we'd need some routing / throughput data collection to make this work. 

1 is not really the right idea, but I mention it for completeness - people should always have choice and not be strong-armed into anything.

Some examples of this similar problem are:
* Torrents
* Tor
* IPFS 
* Blockchains

And out of those, Blockchains have the strongest case for having a node contribute to the network - compensation for participation.

PR #300, opened 2026-04-16
models: inspect and clean up mesh-managed cache entries safely

IvGolovach:
## Summary

Users can now tell which cached models are mesh-managed versus external, see when mesh-llm last used them, and safely preview or remove only mesh-managed cache entries with `mesh-llm models cleanup`.

## Why

The shared Hugging Face cache can grow quickly, but today it is hard to tell which files mesh-llm actually downloaded or used and which ones belong to other tools. That makes cleanup risky and leaves users guessing about disk usage.

Related: #256

## Diff scope

- add mesh-owned model usage records under the mesh cache directory
- mark model usage from catalog downloads, exact-ref downloads, and cache-backed model resolution paths
- enrich `mesh-llm models installed` with `mesh-managed` vs `external` ownership and `last used`
- add `mesh-llm models cleanup` with dry-run-by-default behavior, `--unused-since`, and `--yes`
- remove only paths that mesh-llm explicitly marked as mesh-managed; external Hugging Face cache entries are never deleted
- update the README with the new cache inspection and cleanup workflow

## Example

```bash
mesh-llm models installed
mesh-llm models cleanup --unused-since 30d
mesh-llm models cleanup --unused-since 30d --yes
```

Representative output:

```text
Qwen2.5-3B-Instruct-Q4_K_M
  ownership: mesh-managed
  last used: 2h ago

Previewing cleanup candidates older than 30d
  would remove: Qwen2.5-3B-Instruct-Q4_K_M
```

## Architecture

- usage and ownership metadata live under the mesh-owned cache, not inside Hugging Face cache internals
- cleanup is keyed off explicit mesh usage records rather than filename heuristics
- tracked usage is limited to cache-backed model paths so arbitrary local files are not pulled into cleanup scope

## Protocol

Not applicable — no wire format or protocol changes.

## Validation

- `cd mesh-llm/ui && npm ci`
- `CARGO_HOME=/tmp/mesh-llm-cargo-home CARGO_TARGET_DIR=/tmp/mesh-llm-model-cleanup-target cargo check -p mesh-llm`
- `CARGO_HOME=/tmp/mesh-llm-cargo-home CARGO_TARGET_DIR=/tmp/mesh-llm-model-cleanup-target cargo test -p mesh-llm models::usage -- --nocapture` -> `3 passed; 0 failed`
- `CARGO_HOME=/tmp/mesh-llm-cargo-home CARGO_TARGET_DIR=/tmp/mesh-llm-model-cleanup-target cargo test -p mesh-llm cleanup_age_parser -- --nocapture` -> `2 passed; 0 failed`
- `CARGO_HOME=/tmp/mesh-llm-cargo-home CARGO_TARGET_DIR=/tmp/mesh-llm-model-cleanup-target cargo test -p mesh-llm models::catalog::tests -- --nocapture` -> `21 passed; 0 failed`
- `CARGO_HOME=/tmp/mesh-llm-cargo-home CARGO_TARGET_DIR=/tmp/mesh-llm-model-cleanup-target cargo test -p mesh-llm models::resolve::tests -- --nocapture` -> `37 passed; 0 failed`
- `git diff --check`

## Known residual risks

- the broader crate test suite still has unrelated existing failures in this environment outside the touched model/cache paths
- cleanup only removes cache entries that mesh-llm explicitly marked as mesh-managed; it does not try to infer ownership for pre-existing third-party cache files

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,15 @@ mesh-llm discover                          # browse available meshes
 mesh-llm gpus                              # inspect local GPUs and stable IDs
 ```
 
+### Inspect and clean the shared model cache
+```bash
+mesh-llm models installed
+mesh-llm models cleanup --unused-since 30d
+mesh-llm models cleanup --unused-since 30d --yes
+```
+
+`models installed` now shows whether a cached model is mesh-managed or external plus the last time mesh-llm used it. `models cleanup` only removes model files that mesh-llm explicitly marked as mesh-managed; by default it prints a dry run preview and requires `--yes` to delete anything.
+
 ### Multi-model
 ```bash
 mesh-llm serve --model Qwen2.5-32B --model GLM-4.7-Flash
--- a/mesh-llm/src/cli/commands/models/formatters.rs
+++ b/mesh-llm/src/cli/commands/models/formatters.rs
@@ -4,6 +4,7 @@ use crate::models::{
 };
 use crate::system::hardware;
 use anyhow::Result;
+use chrono::{DateTime, Utc};
 use serde_json::{json, Value};
 use std::path::{Path, PathBuf};
 
@@ -46,6 +47,8 @@ pub(crate) struct InstalledRow {
     pub(crate) size: Option<u64>,
     pub(crate) catalog_model: Option<&'static catalog::CatalogModel>,
     pub(crate) capabilities: ModelCapabilities,
+    pub(crate) managed_by_mesh: bool,
+    pub(crate) last_used_at: Option<String>,
 }
 
 pub(crate) trait ModelsFormatter: SearchFormatter {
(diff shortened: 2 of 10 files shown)


i386 (changes requested):
ignore - put feedback on wrong PR

i386:
@IvGolovach love this change - I dont think this will track the entire model (which is multiple files over splits and metadata). One approach worth thinking about its:

* key the record by canonical HF identity when available
* store the full resolved managed_paths for that identity
* update last_used_at by identity
* clean up by stored paths, optionally validating they still map back to the same identity before deletion

IvGolovach:
> @IvGolovach love this change - I dont think this will track the entire model (which is multiple files over splits and metadata). One approach worth thinking about its:
> 
> * key the record by canonical HF identity when available
> * store the full resolved managed_paths for that identity
> * update last_used_at by identity
> * clean up by stored paths, optionally validating they still map back to the same identity before deletion

Thank you, I’m glad you pointed that out. After looking at it again from the split-model angle, I agreed the previous version was still treating usage too much like individual file hits instead of one managed bundle.

I reworked it so the record stays anchored to the managed bundle, keeps the full stored path set, refreshes usage through that existing record, and only removes paths during cleanup if they still line up with the recorded HF identity. I also added tests around split shards and cleanup validation so that behavior is exercised directly.

i386 (approved):

i386:
Nice one @IvGolovach - merged

Issue #275, opened 2026-04-13
Make metrics a first-class control plane in mesh-llm


ndizazzo:
## Summary

PR #271 established the right starting philosophy for metrics in mesh-llm: keep them local, bounded, actionable, and close to real routing decisions. It added `inference_perf` as a local TTFT memory keyed by `(model, target)` and exposed it on `/api/status`, which is exactly the kind of metric that helps the mesh decide what to do without immediately turning into a protocol burden.

This proposal expands that idea into a core metrics model for mesh-llm across three layers:

1. **Runtime-level**: metrics the mesh can use for real-time decisions.
2. **Information-level**: metrics surfaced to users/operators in the UI and API.
3. **Strategy-level**: metrics used to evaluate product direction, weak spots, and roadmap priorities.

The key design rule is that every metric must also declare a **scope**:

- `local-only`: measured and consumed on one node.
- `peer-advertised`: shared by a node to peers and therefore part of compatibility-sensitive behavior.
- `mesh-derived`: computed locally from already available peer/API/state snapshots, without becoming a new gossip contract.

That scope contract matters more than the layer alone. In mesh-llm, the dangerous mistake is not adding a metric; it is letting a local heuristic accidentally become mesh-wide truth.

## Why now

mesh-llm already has most of the raw ingredients for a metrics system:

- `/api/status` exposes `inflight_requests`, peer `rtt_ms`, `routing_affinity`, GPU capacity hints, and live node/peer inventory.
- `/api/models` already exposes per-model `request_count`, `last_active_secs_ago`, `node_count`, `mesh_vram_gb`, capability flags, and active node lists.
- runtime logic already uses `active_demand`, serving counts, VRAM fit, and routing state to make decisions.
- gossip already carries model demand, serving/runtime descriptors, and capability metadata.
- PR #271 adds `inference_perf`, a concrete example of metrics as routing memory rather than generic observability.

What is missing is a coherent policy for which metrics are core, who consumes them, and which ones must remain local versus becoming compatibility-sensitive mesh signals.

## Goals

- Define a core metrics taxonomy for mesh-llm.
- Separate metrics by **consumer**, **scope**, and **time horizon**.
- Reuse existing repo terminology whenever possible.
- Prefer additive local/API metrics first.
- Identify missing metrics that are necessary for routing quality, operator trust, and strategic planning.

## Non-goals

- This is not a Prometheus/OpenTelemetry/exporter proposal.
[...]

PR #301, opened 2026-04-16
status: surface routing outcomes and utilization in the management API

IvGolovach:
## Summary

Users can now inspect bounded routing outcome and utilization metrics in `/api/status` and `/api/models`. The management API now shows request success, retry and failover behavior, timeout and availability outcomes, queue and attempt timing, throughput samples, and local-vs-remote service pressure without expanding the mesh protocol.

## Why

Issue #275 proposes making metrics a first-class control plane in mesh-llm and explicitly recommends local, bounded, routing-adjacent metrics first. This PR implements that first local/API phase so operators and future routing policy work can see not just latency, but also outcome quality and node pressure.

Related: #275

## Diff scope

- add a bounded in-memory routing metrics store keyed locally by model and target
- track per-attempt outcomes for success, timeout, unavailable, context-overflow, and reject paths
- track per-request outcomes for success, retry count, failover count, and service locality
- capture queue wait, attempt duration, observed completion tokens, and throughput samples
- expose operator summary fields on `/api/status`
- expose per-model routing metrics on `/api/models`
- keep no-model fallback traffic visible in global status metrics without inventing fake per-model buckets
- keep timeout classification strict to concrete timeout types instead of broad string matching
- keep the change local-only with no gossip, protobuf, or wire-format changes

## Test proof

- `cd mesh-llm/ui && npm ci`
- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm routing_metrics_ -- --nocapture` -> `5 passed; 0 failed`
- `cargo test -p mesh-llm test_is_timeout_error_accepts_concrete_timeout_types_only -- --nocapture` -> `1 passed; 0 failed`
- `cargo test -p mesh-llm test_api_models_include_model_routing_metrics -- --nocapture` -> `1 passed; 0 failed`
- `cargo test -p mesh-llm test_moe_remote_failure_removes_peer_for_faildown -- --nocapture` -> `1 passed; 0 failed`
- `cargo test -p mesh-llm -- --skip network::nostr::integration_tests::publish_discover_round_trip` -> `906 passed; 0 failed; 2 ignored; 1 filtered`
- `git diff --check`

## Verification-pack proof

Not applicable - no CI, workflow, infra, governance, replay, or migration logic changed.

## Migration notes

Not applicable - no protocol or schema migration.

## CI context confirmation

Not applicable - no CI workflow or required context names changed.

## Rollback plan

`git revert <merge_commit_sha>`

No DB downgrade required.

## Known residual risks

- throughput samples depend on upstream responses including usage or completion token fields, so some delivered requests will contribute latency and outcome data without a throughput sample
- metrics are intentionally local-only in this phase, so remote peers do not yet consume them for mesh-wide routing decisions


i386 (changes requested):
Oops added this comment to #300 not here.... 

record_attempt is on the routing hot path, and every call currently takes the same process-wide std::sync::Mutex while also doing pruning, LRU maintenance, hashmap mutation, and string-key bookkeeping. Concurrent requests will serialize here, and the same request can hit the same lock multiple times through observe_inflight, record_attempt, and record_request. Can we reduce the critical section here, for example by moving prune/LRU work off the write path, using atomics for global counters, or sharding per-model metrics so unrelated requests don’t contend on one lock?`

IvGolovach:
> Oops added this comment to #300 not here....
> 
> record_attempt is on the routing hot path, and every call currently takes the same process-wide std::sync::Mutex while also doing pruning, LRU maintenance, hashmap mutation, and string-key bookkeeping. Concurrent requests will serialize here, and the same request can hit the same lock multiple times through observe_inflight, record_attempt, and record_request. Can we reduce the critical section here, for example by moving prune/LRU work off the write path, using atomics for global counters, or sharding per-model metrics so unrelated requests don’t contend on one lock?`

Thanks, that was a very good catch. I took another pass over the hot path and you were right that the single process-wide mutex was doing too much work.

I updated the branch to move the global counters onto atomics, shard the per-model metrics state so unrelated models do not contend on one lock, and move the prune/eviction work off the unconditional write path. I also added concurrency-oriented coverage around bounds, pruning, peak inflight, and aggregate preservation. I re-ran the targeted metrics/API tests plus a full local `cargo test -p mesh-llm -- --skip network::nostr::integration_tests::publish_discover_round_trip`, and everything is green now.

ndizazzo (changes requested):
Thanks for starting us off here @IvGolovach! 

Given that you're the first to break ground here.... A few things I think could round this out are:

1. Can you set up our initial repo-local documentation for the new metrics, ideally in mesh-llm/docs/METRICS.md, with a short table for each new metric group?
2. For every new metric or metric group, declare scope explicitly using the issue’s vocabulary: **local-only**, **peer-advertised**, or **mesh-derived**. This will be extremely helpful for agents/humans working in the codebase to use as guidelines.
3. In the API docs/comments for `/api/status` and `/api/models`, note that `routing_metrics`, `local_node`, pressure, and per-model `routing_metrics.targets` are measured on the current node only and are not mesh-wide aggregates
4. I think we could soften or refine the “utilization” language to “local routing pressure” or “local utilization proxies” unless a stricter utilization definition is added

One thought I had was actually encoding the classification and scope for each new metric group in code as explicit metadata directly, and add a check/test that every exported metric declares `runtime/information/strategy` and `local-only/peer-advertised/mesh-derived` rather than relying on struct nesting, comments, or docs. 

This actually gives us an enforceable self-documenting shape we can potentially use at runtime in the future.

IvGolovach:
> Thanks for starting us off here @IvGolovach!
> 
> Given that you're the first to break ground here.... A few things I think could round this out are:
> 
> 1. Can you set up our initial repo-local documentation for the new metrics, ideally in mesh-llm/docs/METRICS.md, with a short table for each new metric group?
> 2. For every new metric or metric group, declare scope explicitly using the issue’s vocabulary: **local-only**, **peer-advertised**, or **mesh-derived**. This will be extremely helpful for agents/humans working in the codebase to use as guidelines.
> 3. In the API docs/comments for `/api/status` and `/api/models`, note that `routing_metrics`, `local_node`, pressure, and per-model `routing_metrics.targets` are measured on the current node only and are not mesh-wide aggregates
> 4. I think we could soften or refine the “utilization” language to “local routing pressure” or “local utilization proxies” unless a stricter utilization definition is added
> 
> One thought I had was actually encoding the classification and scope for each new metric group in code as explicit metadata directly, and add a check/test that every exported metric declares `runtime/information/strategy` and `local-only/peer-advertised/mesh-derived` rather than relying on struct nesting, comments, or docs.
> 
> This actually gives us an enforceable self-documenting shape we can potentially use at runtime in the future.

Thanks, this was really helpful feedback. I took another pass and rounded it out along those lines.

I added repo-local docs in `mesh-llm/docs/METRICS.md` with a short table for each exported metric group, tagged the new groups in code with explicit layer/scope metadata using the issue vocabulary, and added a test so those exported groups have an enforceable classification shape rather than relying only on comments or struct nesting.

I also updated the `/api/status` and `/api/models` docs/comments to make it explicit that `routing_metrics`, `local_node`, `pressure`, and per-model `routing_metrics.targets` are current-node-only and not mesh-wide aggregates, and I tightened the wording around local routing pressure / utilization proxies.

I kept the code-side metadata scoped to the groups this PR actually exports for now, so it is self-documenting and enforceable without turning this into a larger runtime metadata system in one step. I reran the targeted metrics/API coverage and a broader local crate test pass after the update, and everything is green.

PR #305, opened 2026-04-16
scripts: fix silent build exit when neither ccache nor sccache is installed

### LINKED ISSUE #303: configure_compiler_cache() silent-kill under set -e
# `scripts/build-linux.sh`: `configure_compiler_cache()` silently kills the build under `set -euo pipefail` when neither `ccache` nor `sccache` is installed

## Summary

`configure_compiler_cache()` in `scripts/build-linux.sh` has a bare `return` (no argument) in the `else` branch. Because the script runs under `set -euo pipefail`, the bare `return` **inherits the non-zero exit status of the preceding failed `command -v ccache` check** and propagates that up through `set -e` — killing the entire build ~3 seconds into `just release-build-cuda` with **no error output** and nothing between "HEAD is now at …" and "Recipe failed".

I lost a non-trivial amount of time debugging builds on a Docker image that had everything except `ccache`. Dropping `ccache` from the builder apt install was not an obvious trigger.

## Environment

- **Version**: repo HEAD on 2026-04-16, `412ca60 v0.61.0-rc.1: prerelease`
- **Shell**: GNU bash 5.2.21 (ubuntu:24.04 container default)
- **Invocation**: `just release-build-cuda "80;86;89"` from inside an `ubuntu:24.04` builder container with `build-essential cmake git curl ca-certificates pkg-config libssl-dev python3 python3-pip ninja-build`, `just`, Node 24, and Rust stable — but not `ccache` or `sccache`.

## Steps to reproduce

In a fresh container that has every other build dep but deliberately lacks `ccache`/`sccache`:

```bash
apt-get install -y build-essential cmake git curl ca-certificates pkg-config libssl-dev python3 python3-pip ninja-build
# (intentionally NOT installing ccache or sccache)
curl -fsSL https://just.systems/install.sh | bash -s -- --to /usr/local/bin
curl -fsSL https://deb.nodesource.com/setup_24.x | bash -
apt-get install -y nodejs
curl -fsSL https://sh.rustup.rs | sh -s -- -y --default-toolchain stable --profile minimal
export PATH=$HOME/.cargo/bin:$PATH

git clone https://github.com/Mesh-LLM/mesh-llm
cd mesh-llm
just release-build-cuda "80"
```

Within ~3 seconds the build exits. The only output is:

```
Building Linux backend: CUDA
Using nvcc: /usr/local/cuda/bin/nvcc (12.6)
Cloning Mesh-LLM/llama.cpp pinned to ed279faf43c9f43aeff45be85a9a1376f8cf9ad1...
Cloning into '...'...
HEAD is now at ed279fa mesh hooks default OFF per-request
error: Recipe `release-build-cuda` failed on line 59 with exit code 1
```

Nothing between "HEAD is now at …" and the `just` error. No cmake output, no compile output. `bash -x scripts/build-linux.sh --backend cuda --cuda-arch 80` reveals the actual last trace is `+ return` inside `configure_compiler_cache`, immediately followed by script exit.

## Root cause

The `configure_compiler_cache()` helper (lines ~170-190 in `scripts/build-linux.sh`):

```bash
configure_compiler_cache() {
    local backend="$1"
    local cache_bin=""
    if command -v sccache >/dev/null 2>&1; then
        cache_bin="sccache"
    elif command -v ccache >/dev/null 2>&1; then
        cache_bin="ccache"
    else
        return        # ← bug
    fi
    ...
}
```

When both `command -v sccache` and `command -v ccache` fail, `$?` is non-zero. The bare `return` carries that status out of the function. The caller is running under `set -e` so the whole script dies silently.

This is a well-known POSIX shell footgun — the body of an `if … elif … else … fi` leaves `$?` as whatever the failed branch's test returned, and an explicit-looking `return` with no argument inherits it.

## Fix (one-line)

```diff
--- a/scripts/build-linux.sh
+++ b/scripts/build-linux.sh
@@ -180,7 +180,7 @@ configure_compiler_cache() {
     elif command -v ccache >/dev/null 2>&1; then
         cache_bin="ccache"
     else
-        return
+        return 0
     fi
```

Makes the "no cache tool available" exit explicitly success, which matches the function's actual intent (it's a no-op in that case — the caller doesn't need a cache).

## Suggested same-pattern audit

I'd recommend a one-time grep for bare `return$` inside bash functions across the repo:

```bash
grep -rEn '^\s*return\s*$' scripts/ | grep -v '\.md:'
```

Other scripts that likely follow the same helper pattern and might have the same silent-kill bug:
- `scripts/build-linux-rocm.sh`
- `scripts/build-linux-vulkan.sh` (if it exists)
- `scripts/package-release.sh`

## PR / workaround notes

- Deployer-side workaround until fixed: install either `ccache` or `sccache` in the builder. The function then takes the success path cleanly.
- Happy to send a PR with just the one-line `return 0` fix + an audit of the rest of `scripts/`. Let me know if you'd prefer that scoped narrowly to this file or a broader sweep.

ventz:
Closes #303.

## Summary

`configure_compiler_cache()` in `scripts/build-linux.sh` (and its identical copies in `scripts/build-linux-rocm.sh` and `scripts/build-release.sh`) has a bare `return` in the `else` branch. Under the calling script's `set -euo pipefail`, the bare `return` carries the non-zero exit status from the preceding failed `command -v ccache` check and kills the entire build with no diagnostic output — just ~3 seconds after llama.cpp is cloned.

## Fix

Explicit `return 0` so the no-op "neither cache tool installed" path doesn't poison `$?`. One-line change, applied identically across all three backend scripts that carry the copy-pasted helper.

## Reproduction (from #303)

In an otherwise-complete builder container that deliberately lacks `ccache`/`sccache`:

```
Cloning Mesh-LLM/llama.cpp pinned to ...
HEAD is now at ed279fa ...
error: Recipe `release-build-cuda` failed on line 59 with exit code 1
```

`bash -x scripts/build-linux.sh --backend cuda --cuda-arch 80` confirms the last trace is `+ return` from `configure_compiler_cache`, with the script exiting immediately after (no cmake output). With this patch, the builder proceeds past `configure_compiler_cache` and `cmake` runs normally.

## Test

Verified locally on a fresh `nvidia/cuda:12.6.3-devel-ubuntu24.04` container without ccache/sccache installed: before the patch, `just release-build-cuda "80"` dies silently in 3 s; after the patch, it proceeds to cmake and compiles. Also re-verified the happy path (ccache installed) still reaches the `Using compiler cache: ccache` message.

### DIFF
--- a/scripts/build-linux-rocm.sh
+++ b/scripts/build-linux-rocm.sh
@@ -44,7 +44,7 @@ configure_compiler_cache() {
     elif command -v ccache >/dev/null 2>&1; then
         cache_bin="ccache"
     else
-        return
+        return 0
     fi
 
     echo "Using compiler cache: $cache_bin"
--- a/scripts/build-linux.sh
+++ b/scripts/build-linux.sh
@@ -164,7 +164,7 @@ configure_compiler_cache() {
     elif command -v ccache >/dev/null 2>&1; then
         cache_bin="ccache"
     else
-        return
+        return 0
     fi
 
     echo "Using compiler cache: $cache_bin"
--- a/scripts/build-release.sh
+++ b/scripts/build-release.sh
@@ -28,7 +28,7 @@ configure_compiler_cache() {
     elif command -v ccache >/dev/null 2>&1; then
         cache_bin="ccache"
     else
-        return
+        return 0
     fi
 
     echo "Using compiler cache: $cache_bin"


i386 (approved):
Nice catch. I'll merge when CI is green.

michaelneale (approved):
thanks @ventz forgot about this - bit me a while ago.

PR #306, opened 2026-04-16
Add reusable command bar / filter bars, ship the first model catalog filters

ndizazzo:
This PR adds a reusable filter-bar extension point to the command bar so results can be narrowed with structured controls before text search is applied. The first implementation uses that new surface for the model catalog, where operators can quickly filter by capability and warm/cold status, then refine further with normal search text.

## Summary

The command bar is no longer limited to plain text search. It now supports an optional interstitial filter component that renders between the search input and the result list, making the feature reusable for future command-bar modes beyond models. This PR uses that new capability to add the first concrete filter bar for the model picker.

## Screenshot
image

## What users can do now

Users can open the Models command bar and:

- filter models by capability with toggle controls
- filter by `All`, `Warm`, or `Cold`
- keep typing in the search box to narrow the already-filtered results
- select a filtered model exactly as before, with the selected model reflected in the existing UI surfaces

## Architecture

This is intentionally not a models-only feature. `CommandBarModal` now supports an optional interstitial slot rendered between the header/search area and the results pane, so any future command-bar mode can provide its own lightweight filter UI without changing the shell. The model picker is simply the first consumer of that new extension point.

For the model catalog specifically, structured filter state lives in `App.tsx`, where the mesh model list is prefiltered before being passed into the command bar mode source. That keeps the existing command-bar text filtering and ranking behavior intact, while making text search apply on top of the structured filters rather than replacing them.

## UI and behavior

The first shipped filter bar uses shadcn/Radix-style primitives:

- a multi-select Toggle Group for model capabilities
- a Select dropdown for `All` / `Warm` / `Cold`

The filter bar is compact, stays on one row, keeps the status dropdown visible, and gives enabled capability toggles a clearer selected state. The command bar also retains existing keyboard/search behavior, including clear-search support and scroll-follow for keyboard selection.

## Validation

This change was verified with:

- targeted modal tests for the optional interstitial slot
- app-level integration tests covering capability filtering, status filtering, text search layering, and filtered selection
- clean diagnostics on touched files
- successful UI build
- live manual QA in the dev UI confirming that filter controls render in the right place, narrow results correctly, and still allow selecting a model through the normal flow

## TODO
- [ ] **Disable hotkey and button before merge**, once testing completed.


i386 (approved):
I like it - I assume this is building up to your configuration work?

ndizazzo:
> I like it - I assume this is building up to your configuration work?

@i386 yup, exactly!

PR #309, opened 2026-04-17
ci: dual-lane CUDA releases (cuda on 12.6.3, cuda-blackwell on 12.8)

### LINKED ISSUE #304: A30 / driver 535: v0.60.3 CUDA bundle fails at matmul despite sm_80 cubins being present
# Published CUDA v0.60.3 bundle crashes at first matmul on A30 (SM 8.0) — SM 8.0 cubins ARE present, cause is unclear

## Summary

The `mesh-llm-x86_64-unknown-linux-gnu-cuda.tar.gz` bundle published for `v0.60.3` crashes at the first matmul with `CUDA error: device kernel image is invalid` when run against an NVIDIA A30 (Ampere, compute capability 8.0) on an R535-series driver. By the same apparent mechanism, we'd expect the same failure on A100 (also compute 8.0), though that hasn't been verified on our hardware.

**Note — I originally hypothesized this was a missing-SM-8.0-cubin problem. That hypothesis is wrong** — `cuobjdump --list-elf` on the bundled `llama-server-cuda` binary shows SM 8.0 cubins present for every kernel (alongside 75, 86, 89, 90, 120a). The crash is real and reproducible, but the underlying cause is something else (see "Revised hypotheses" below).

## Environment

- **Version**: `v0.60.3` release bundle (`mesh-llm-x86_64-unknown-linux-gnu-cuda.tar.gz`)
- **GPU**: NVIDIA A30 (24 GB), accessed via a MIG 4g.24gb compute instance (full GPU)
- **Host driver**: 535.288.01 (R535 series, CUDA 12.2 native)
- **Container base for runtime**: `nvidia/cuda:12.6.3-runtime-ubuntu24.04`
  - `/usr/local/cuda-12.6/compat/` removed so the host's libcuda (535) is used instead of the forward-compat shim (the shim silently breaks `cudaGetDeviceCount()` on this driver — a separate issue).
- **Deployment**: `docker run --gpus '"device=MIG-<uuid>"' ... --llama-flavor cuda` (not `--gpus all`; under MIG the parent GPU UUID doesn't work and must be the MIG instance UUID).

## Steps to reproduce

```bash
# Host: A30 with nvidia-container-toolkit, MIG enabled and one 4g.24gb instance created
# (sudo nvidia-smi mig -cgi 0 -C if no instance exists)
MIG_UUID=$(nvidia-smi -L | grep -oP 'MIG-[a-f0-9-]+' | head -1)

curl -fsSL https://github.com/Mesh-LLM/mesh-llm/releases/download/v0.60.3/mesh-llm-x86_64-unknown-linux-gnu-cuda.tar.gz \
  | tar xz
docker run --rm --gpus "\"device=$MIG_UUID\"" \
  -v $PWD/mesh-bundle:/opt/mesh-bundle:ro \
  -v $PWD/models:/models:ro \
  --entrypoint /bin/bash \
  nvidia/cuda:12.6.3-runtime-ubuntu24.04 \
  -c 'apt-get update && apt-get install -y libssl3t64 libgomp1 curl && \
      rm -rf /usr/local/cuda-12.6/compat && \
      /opt/mesh-bundle/mesh-llm serve \
        --gguf /models/<any-gguf>.gguf \
        --listen-all --llama-flavor cuda'
```

## Expected

Model loads and inference serves normally.

## Actual

`rpc-server-cuda` comes up, reports the GPU correctly, claims ~512 MB VRAM. `llama-server-cuda` loads the model weights into VRAM (~21 GB of 24 GB used), then crashes on the first forward pass during the warm-up run:

```
sched_reserve:      CUDA0 compute buffer size =   300.75 MiB
sched_reserve:  CUDA_Host compute buffer size =   136.09 MiB
common_init_from_params: warming up the model with an empty run - please wait ... (--no-warmup to disable)
ggml_cuda_compute_forward: MUL_MAT failed
/__w/mesh-llm/mesh-llm/llama.cpp/ggml/src/ggml-cuda/ggml-cuda.cu:98: CUDA error
CUDA error: device kernel image is invalid
  current device: 0, in function ggml_cuda_compute_forward at
  /__w/mesh-llm/mesh-llm/llama.cpp/ggml/src/ggml-cuda/ggml-cuda.cu:2825
```

## Evidence that SM 8.0 is NOT missing from the bundle

Ran `cuobjdump --list-elf /opt/mesh-bundle/llama-server-cuda` inside `nvidia/cuda:12.8.0-devel-ubuntu22.04`. SM 8.0 cubins are present, 60 of them, one per kernel variant alongside 75/86/89/90/120a:

```
ELF file  1:  llama-server-cuda.1.sm_75.cubin
ELF file  2:  llama-server-cuda.2.sm_80.cubin    ← present
ELF file  3:  llama-server-cuda.3.sm_86.cubin
ELF file  4:  llama-server-cuda.4.sm_89.cubin
ELF file  5:  llama-server-cuda.5.sm_90.cubin
ELF file  6:  llama-server-cuda.6.sm_120a.cubin
ELF file  7:  llama-server-cuda.7.sm_75.cubin
ELF file  8:  llama-server-cuda.8.sm_80.cubin    ← present
ELF file  9:  llama-server-cuda.9.sm_86.cubin
...
(pattern repeats; 60+ kernels × 6 arches each, including sm_80 consistently)
```

Same pattern holds for `rpc-server-cuda`.

## Revised hypotheses (testable)

Differences between the failing upstream build and a from-source build that *does* work on the same A30:

| | Upstream release (fails on A30) | Source build (works on A30) |
|---|---|---|
| CUDA toolkit | 12.8 (`nvidia/cuda:12.8.0-devel-ubuntu22.04` per `.github/workflows/release.yml`) | 12.6 |
| `GGML_CUDA_FA_ALL_QUANTS` | ON (default) | OFF (CI opt-out) |
| `CMAKE_CUDA_ARCHITECTURES` | `75;80;86;87;89;90;100;103;120` | `80;86;89` |
| Targets built | all | `llama-server rpc-server llama-moe-split llama-moe-analyze` only |

Host driver, MIG config, container base, and model are identical across both tests.

Plausible causes, in decreasing order of likelihood:

1. **CUDA-toolkit / driver minor-version mismatch.** Binary assembled by CUDA 12.8 nvcc; host libcuda is 535-series (CUDA 12.2 native). With the forward-compat shim removed (required to even get past `cudaGetDeviceCount` on this driver), cubins tagged as CUDA 12.8 may be rejected at load by the CUDA 12.2 driver for a specific kernel set. The load failure at a *late* stage (first matmul, not context creation) is consistent with a lazy kernel-load path finding the first unloadable cubin on the hot path.

2. **A FlashAttention kernel variation that requires features only present under CUDA 12.8 runtime.** Ruling out by rebuilding upstream's recipe with `MESH_LLM_CUDA_FA_ALL_QUANTS=off` and seeing if it runs would confirm or disprove.

3. **A kernel that legitimately requires `sm_89`/`sm_90` features** slipping into a code path executed on SM 8.0 (e.g. an `#ifdef` that doesn't correctly guard against 8.0, or `__CUDA_ARCH__ >= 890` being the default branch). Would manifest as an `sm_80.cubin` that's technically present but non-functional for that specific kernel.

## Concrete experiments that would disambiguate

1. Rebuild upstream's exact recipe (CUDA 12.8 devel, `FA_ALL_QUANTS=on`, full arch list, all targets) and test on A30 + driver 535. If it still fails identically → not a build-config issue, likely #1.
2. Same as #1 but CUDA 12.6 devel (everything else identical). If that works → confirms CUDA-toolkit/driver mismatch is the root cause and the fix is pinning the release build to CUDA ≤ 12.6, OR baking the appropriate forward-compat libs into the release.
3. Same as #1 but `FA_ALL_QUANTS=off` only. If that works → points at a specific FA kernel family as the offender.

## Workaround (confirmed working on A30 / driver 535)

Build from source with CUDA ≤ 12.6 and the narrower config:

```bash
git clone https://github.com/Mesh-LLM/mesh-llm
cd mesh-llm
MESH_LLM_CUDA_FA_ALL_QUANTS=off \
MESH_LLM_LLAMA_TARGETS="llama-server rpc-server llama-moe-split llama-moe-analyze" \
just release-build-cuda "80;86;89"
```

## Suggested fix (symptom-based — cause-based fix pending the experiments above)

Regardless of which hypothesis turns out right, the release pipeline currently has no functional smoke test that would catch this class of bug. The existing `./mesh-llm --version` check does not exercise the CUDA matmul path. A minimum fix that protects users from the current failure mode:

- Add an A30 or A100 runner step to `.github/workflows/release.yml` that actually serves a tiny GGUF through `mesh-llm ... --llama-flavor cuda` and hits `/v1/chat/completions` for one token. Would have caught this bug before the v0.60.3 release shipped.
- Happy to submit a PR for the smoke-test wiring if you can tell me what runner type is available to the project (A30/A100/other).

## Related

- Detected during a private-LAN deployment of mesh-llm on 3 NVIDIA A30 GPUs.
- Not related to [#303 / PR #305 — `configure_compiler_cache()` silent-kill](https://github.com/Mesh-LLM/mesh-llm/issues/303) (filed separately as a build-script robustness issue).

ventz:
Fixes #304.

## Summary

Splits the published CUDA release into **two lanes** so we can keep supporting both the R535 driver install base (A30/A100, current fleet) AND Blackwell hardware (B100/B200/RTX 50-series). nvcc 12.8 emits cubins the R535 driver rejects at kernel load (the failure described in #304); nvcc 12.6 cannot emit Blackwell cubins at all. One toolkit can't do both.

| Asset suffix | Toolkit | Arch | Driver |
|---|---|---|---|
| `-cuda` (primary) | CUDA 12.6.3 | sm_75, 80, 86, 87, 89, 90 (Turing → Hopper) | R535+ |
| `-cuda-blackwell` | CUDA 12.8 | adds sm_100, 103, 120 | R550+ |

Inner binary filenames stay `-cuda` in both bundles (the mesh-llm runtime `BinaryFlavor` enum has one cuda variant). Users pick the archive matching their driver; both invoke as `--llama-flavor cuda`.

Root cause verified on 29 GPUs on driver 535.288.01 (14× A30, 15× A100):

| Toolkit | A30 (SM 8.0) | A100 (SM 8.0) |
|---|---|---|
| 12.6.3 | works | works |
| 12.8.0 | crash at first matmul | same failure expected |

## Changes

- **`.github/workflows/release.yml`** — `build_linux_cuda` now pulls `nvidia/cuda:${CUDA_VERSION:-12.6.3}-...`; new sibling `build_linux_cuda_blackwell` job pulls `${CUDA_BLACKWELL_VERSION:-12.8.0}-...`, both wired into `publish.needs`. New repo Actions variable `vars.CUDA_BLACKWELL_VERSION` mirrors `vars.CUDA_VERSION`.
- **`.github/workflows/ci.yml`, `.github/workflows/llama-cache-keys.yml`** — pin fallback to 12.6.3; add `cuda_blackwell_fat_cache_key` output so the two lanes don't collide in the warm cache.
- **`.github/workflows/docker.yml`** — narrow `CUDA_ARCH` build-arg to `75;80;86;87;89;90` (docker image tracks primary lane only; Blackwell docker tag is follow-up).
- **`Justfile`** — primary `release-build-cuda` default arch narrowed to sm_75-90; add `release-build-cuda-blackwell`, `release-bundle-cuda-blackwell` (plus Windows mirrors); `docker-build-cuda` gains a `cuda_version` parameter.
- **`docker/Dockerfile.cuda`** — three `nvidia/cuda` FROM lines parameterized via `ARG CUDA_VERSION=12.6.3` so local Blackwell docker builds can override. `ARG CUDA_ARCH` default narrowed to sm_75-90.
- **`scripts/package-release.sh`** — add `cuda-blackwell` to linux/x86_64 `supported_release_flavors`; `binary_flavor_for_release_flavor()` maps cuda-blackwell → cuda so inner binaries stay `llama-server-cuda` etc.
- **`install.sh`** — expose `cuda-blackwell` as an installable flavor and map it to `mesh-llm-x86_64-unknown-linux-gnu-cuda-blackwell.tar.gz`.
- **`mesh-llm/tests/fixtures/release-target-matrix.json`** — rows for cuda-blackwell (linux/x86_64 + windows/x86_64 supported; linux/aarch64 + macos/aarch64 unknown).
- **`README.md`, `RELEASE.md`, `docker/SPEC_COMPLIANCE.md`** — document both lanes and the driver-compat matrix.
- **`docs/cuda-release-lanes.md`** — new, end-to-end rationale + "how to pick" guide.

## Verification

Ran manual bash-level parity tests against `scripts/package-release.sh` and `install.sh` (the same functions `cargo run -p xtask -- repo-consistency release-targets` exercises):

| Check | Result |
|---|---|
| `supported_release_flavors` on linux/x86_64 | `cpu cuda cuda-blackwell rocm vulkan` ✓ |
| `STABLE_ASSET` for cuda | `mesh-llm-x86_64-unknown-linux-gnu-cuda.tar.gz` ✓ |
| `STABLE_ASSET` for cuda-blackwell | `mesh-llm-x86_64-unknown-linux-gnu-cuda-blackwell.tar.gz` ✓ |
| `versioned_asset_name v0.60.0` for cuda-blackwell | `mesh-llm-v0.60.0-x86_64-unknown-linux-gnu-cuda-blackwell.tar.gz` ✓ |
| `bundle_bin_name llama-server` with `MESH_RELEASE_FLAVOR=cuda-blackwell` | `llama-server-cuda` ✓ (inner name preserved) |
| `release_target_support` cuda-blackwell on linux/x86_64 | `supported` ✓ |
| `release_target_support` cuda-blackwell on linux/aarch64 | `unsupported` ✓ |
| `install.sh supported_flavors` on Linux/x86_64 | includes cuda-blackwell ✓ |
| `install.sh asset_name cuda-blackwell` | `mesh-llm-x86_64-unknown-linux-gnu-cuda-blackwell.tar.gz` ✓ |

## Test plan

- [ ] `cargo run -p xtask -- repo-consistency release-targets` passes in CI
- [ ] `build_linux_cuda` job succeeds on `nvidia/cuda:12.6.3-devel-ubuntu22.04`
- [ ] `build_linux_cuda_blackwell` job succeeds on `nvidia/cuda:12.8.0-devel-ubuntu22.04`
- [ ] Docker `mesh-llm:cuda` image builds on 12.6.3 base with narrowed arch list
- [ ] End-to-end: download `mesh-llm-*-cuda.tar.gz` built from this branch, run against A30/A100 on R535 driver with the #304 repro steps — warm-up completes and `/v1/chat/completions` returns a token
- [ ] End-to-end (if Blackwell hardware available): download `mesh-llm-*-cuda-blackwell.tar.gz`, confirm it loads on R550+ driver on Blackwell

## Out of scope / follow-ups

- `mesh-llm:cuda-blackwell` Docker tag (would need a parallel docker.yml invocation).
- An A30/A100 smoke-test step in the release workflow so this class of bug (driver/toolkit cubin mismatch) is caught before publish. Issue #304 requested this; treating as separate PR.
- Teaching `install.sh`'s `recommended_flavor()` to probe `nvidia-smi --query-gpu=compute_cap` and auto-pick `cuda-blackwell` when Blackwell hardware is detected. Currently users must opt in via `MESH_LLM_INSTALL_FLAVOR=cuda-blackwell`.

If `vars.CUDA_VERSION` is currently set at the repo level, it will override the new 12.6.3 fallback — please unset it or set it to `12.6.3` so the R535-compatible primary lane takes effect.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -776,7 +776,7 @@ jobs:
       artifact_path: ci-artifacts/cuda-benchmark
       benchmark_binary: membench-fingerprint-cuda
       runs_on: ${{ vars.CUDA_BENCHMARK_RUNNER || '["ubuntu-latest"]' }}
-      container_image: nvidia/cuda:${{ vars.CUDA_VERSION || '12.8.0' }}-devel-ubuntu22.04
+      container_image: nvidia/cuda:${{ vars.CUDA_VERSION || '12.6.3' }}-devel-ubuntu22.04
       container_options: --gpus all
 
   linux_rocm_benchmark_smoke:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -502,7 +502,7 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
-            CUDA_ARCH=75;80;86;87;89;90;100;103;120
+            CUDA_ARCH=75;80;86;87;89;90
           cache-from: |
             type=gha,scope=docker-rust-deps-amd64-cuda
             type=gha,scope=docker-llama-cuda-amd64
--- a/.github/workflows/llama-cache-keys.yml
+++ b/.github/workflows/llama-cache-keys.yml
@@ -22,8 +22,14 @@ on:
         description: Exact slim CUDA cache key
         value: ${{ jobs.resolve.outputs.cuda_slim_cache_key }}
       cuda_fat_cache_key:
-        description: Exact fat CUDA cache key
+        description: Exact fat CUDA cache key (R535-compatible lane)
         value: ${{ jobs.resolve.outputs.cuda_fat_cache_key }}
+      cuda_blackwell_fat_cache_key:
+        description: Exact fat CUDA cache key for the Blackwell lane
+        value: ${{ jobs.resolve.outputs.cuda_blackwell_fat_cache_key }}
+      cuda_blackwell_version:
+        description: CUDA version fragment used for the Blackwell lane
+        value: ${{ jobs.resolve.outputs.cuda_blackwell_version }}
       rocm_slim_cache_key:
         description: Exact slim ROCm cache key
         value: ${{ jobs.resolve.outputs.rocm_slim_cache_key }}
@@ -49,6 +55,8 @@ jobs:
       rocm_cache_inputs_hash: ${{ steps.cache_inputs.outputs.rocm_hash }}
       cuda_slim_cache_key: ${{ steps.render.outputs.cuda_slim_cache_key }}
       cuda_fat_cache_key: ${{ steps.render.outputs.cuda_fat_cache_key }}
+      cuda_blackwell_fat_cache_key: ${{ steps.render.outputs.cuda_blackwell_fat_cache_key }}
+      cuda_blackwell_version: ${{ steps.cuda_blackwell_version.outputs.value }}
       rocm_slim_cache_key: ${{ steps.render.outputs.rocm_slim_cache_key }}
       rocm_fat_cache_key: ${{ steps.render.outputs.rocm_fat_cache_key }}
 
@@ -64,7 +72,12 @@ jobs:
       - name: Expose CUDA version
         id: cuda_version
         shell: bash
-        run: echo "value=${{ vars.CUDA_VERSION || '12.8.0' }}" >> "$GITHUB_OUTPUT"
+        run: echo "value=${{ vars.CUDA_VERSION || '12.6.3' }}" >> "$GITHUB_OUTPUT"
+
+      - name: Expose CUDA Blackwell version
+        id: cuda_blackwell_version
+        shell: bash
+        run: echo "value=${{ vars.CUDA_BLACKWELL_VERSION || '12.8.0' }}" >> "$GITHUB_OUTPUT"
 
       - name: Resolve llama.cpp SHA
         id: resolve
@@ -100,12 +113,14 @@ jobs:
         env:
           CACHE_NAMESPACE: ${{ steps.cache_namespace.outputs.value }}
           CUDA_VERSION: ${{ steps.cuda_version.outputs.value }}
+          CUDA_BLACKWELL_VERSION: ${{ steps.cuda_blackwell_version.outputs.value }}
           LLAMA_SHA: ${{ steps.resolve.outputs.sha }}
           CUDA_HASH: ${{ steps.cache_inputs.outputs.cuda_hash }}
           ROCM_HASH: ${{ steps.cache_inputs.outputs.rocm_hash }}
         run: |
           set -euo pipefail
           echo "cuda_slim_cache_key=${CACHE_NAMESPACE}-llama-cuda-slim-${LLAMA_SHA}-${CUDA_HASH}-arch89-fa-off-cuda${CUDA_VERSION}" >> "$GITHUB_OUTPUT"
-          echo "cuda_fat_cache_key=${CACHE_NAMESPACE}-llama-cuda-fat-${LLAMA_SHA}-${CUDA_HASH}-arch75-80-86-89-90-120-fa-on-cuda${CUDA_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "cuda_fat_cache_key=${CACHE_NAMESPACE}-llama-cuda-fat-${LLAMA_SHA}-${CUDA_HASH}-arch75-80-86-87-89-90-fa-on-cuda${CUDA_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "cuda_blackwell_fat_cache_key=${CACHE_NAMESPACE}-llama-cuda-blackwell-fat-${LLAMA_SHA}-${CUDA_HASH}-arch75-80-86-87-89-90-100-103-120-fa-on-cuda${CUDA_BLACKWELL_VERSION}" >> "$GITHUB_OUTPUT"
           echo "rocm_slim_cache_key=${CACHE_NAMESPACE}-llama-rocm-slim-${LLAMA_SHA}-${ROCM_HASH}-gfx1100-rocm7.0" >> "$GITHUB_OUTPUT"
           echo "rocm_fat_cache_key=${CACHE_NAMESPACE}-llama-rocm-fat-${LLAMA_SHA}-${ROCM_HASH}-gfx90a-gfx942-gfx1100-gfx1101-gfx1102-gfx1200-gfx1201-rocm7.0" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/llama-cache-keys.yml
+++ b/.github/workflows/llama-cache-keys.yml
@@ -121,6 +121,6 @@ jobs:
           set -euo pipefail
           echo "cuda_slim_cache_key=${CACHE_NAMESPACE}-llama-cuda-slim-${LLAMA_SHA}-${CUDA_HASH}-arch89-fa-off-cuda${CUDA_VERSION}" >> "$GITHUB_OUTPUT"
           echo "cuda_fat_cache_key=${CACHE_NAMESPACE}-llama-cuda-fat-${LLAMA_SHA}-${CUDA_HASH}-arch75-80-86-87-89-90-fa-on-cuda${CUDA_VERSION}" >> "$GITHUB_OUTPUT"
-          echo "cuda_blackwell_fat_cache_key=${CACHE_NAMESPACE}-llama-cuda-blackwell-fat-${LLAMA_SHA}-${CUDA_HASH}-arch75-80-86-87-89-90-100-103-120-fa-on-cuda${CUDA_BLACKWELL_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "cuda_blackwell_fat_cache_key=${CACHE_NAMESPACE}-llama-cuda-blackwell-fat-${LLAMA_SHA}-${CUDA_HASH}-arch75-80-86-87-89-90-100-120-fa-on-cuda${CUDA_BLACKWELL_VERSION}" >> "$GITHUB_OUTPUT"
           echo "rocm_slim_cache_key=${CACHE_NAMESPACE}-llama-rocm-slim-${LLAMA_SHA}-${ROCM_HASH}-gfx1100-rocm7.0" >> "$GITHUB_OUTPUT"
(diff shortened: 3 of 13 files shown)


ventz:
~Don't merge this yet -- I am working on a split so that Blackwell is supported and A30/A100 is too -- a "dual path" if you will.~ Ok, should be good to go.

ventz:
## End-to-end validation (on real `docker build` runs)

| Lane | Toolkit | Arch list built | Kernels compiled | Cubins in binary | Audit |
|---|---|---|---|---|---|
| Primary `-cuda` | CUDA 12.6.3 | `75;80;86;87;89;90` | 435/435 | 6×177 = 1062 | R535-compatible, zero Blackwell leak |
| Blackwell `-cuda-blackwell` | CUDA 12.8.0 | `75;80;86;87;89;90;100;120` | 435/435 | 8×177 = 1416 | includes `sm_100` + `sm_120a` |

Cubin breakdown (`cuobjdump --list-elf`) for each final binary:

**Primary `-cuda` (`llama-server-cuda`, `rpc-server-cuda`):**
177 sm_75   177 sm_80   177 sm_86
177 sm_87   177 sm_89   177 sm_90

**Blackwell `-cuda-blackwell` (`llama-server-cuda`, `rpc-server-cuda`):**
177 sm_75   177 sm_80   177 sm_86    177 sm_87
177 sm_89   177 sm_90   177 sm_100   177 sm_120a

PR should work well!

ndizazzo:
Extremely solid work here @ventz - thank you for the deep debugging and diagnosis. I also just discovered the 12.8 / sm_103 issue during a release and fixed it on main. That is likely locked to CUDA >= 13.2. 

I'm inclined to agree with the overall direction here, I would only add a couple things for people / agents in the future:

* a brief update to docs/CI_GUIDELINES.md referencing the new CUDA doc
* a clear CUDA version / driver matrix to make it clear which jobs are building what:

| Lane | Archive | CUDA toolkit | SMs included |
|---|---|---:|---|
| Primary CUDA | `mesh-llm-x86_64-unknown-linux-gnu-cuda.tar.gz` | 12.6.3 | `sm_75,80,86,87,89,90` |
| Blackwell CUDA | `mesh-llm-x86_64-unknown-linux-gnu-cuda-blackwell.tar.gz` | 12.8.0 | `sm_75,80,86,87,89,90,100,120` |
| Future Blackwell Pro | `mesh-llm-x86_64-unknown-linux-gnu-cuda-sm103.tar.gz` *(example)* | 12.9+ | `sm_103` |

In the future, I think we're likely going to have to build many different of versions of CUDA (and ROCm, etc) for each version, so we'll likely want to devise a matrix strategy that covers this case, exactly for the issue you're running into with your A30/A100 split.

Once this is in, I'll revisit what we're doing for cache optimizations for the different CUDA versions.

ndizazzo:
Please see #355

PR #319, opened 2026-04-17
Fix: Update AMD device naming from HIP to ROCm to match upstream llama.cpp

intenso:
### What does this PR do?
Updates the AMD device naming convention to support the new `ROCm` identifier (e.g., `HIP0` -> `ROCm0`) introduced by upstream `llama.cpp`, while maintaining lenient fallback support for existing `HIP` configurations during the transition. It ensures full dual-support across the Rust backend, internal bash scripts, and CLI surfaces.

### Why is this needed?
Upstream `llama.cpp` recently changed how it identifies and logs AMD/ROCm devices, dropping the `HIP` prefix in favor of `ROCm`. Because `mesh-llm` was still auto-detecting and passing `--device HIP0`, the bundled `llama-server` crashed on startup with `error: unknown device: HIP0`.

Flipping the identifier abruptly carries regression risks for users with existing `HIP{index}` values pinned in their `~/.mesh-llm/config.toml`, or internal pathways expecting the legacy string. To fix the crash safely, this PR aligns the default inventory mapping to `ROCm` while introducing dual/lenient matching in the validation paths so that `HIP` and `ROCm` resolve interchangeably.

### Changes made
**Core Logic & Validation:**
- **`src/system/hardware.rs`**: Updated the hardware inventory format from `HIP{index}` to `ROCm{index}`.
- **`src/inference/launch.rs`**: 
  - Updated primary device candidates to use `ROCm0`.
  - Re-added `HIP0` into the fallback candidate lists (`preferred_devices`) for backward compatibility.
  - Added lenient matching to `resolve_device_for_binary` to safely translate requested `ROCm` <-> `HIP` strings depending on what the underlying binary exposes.
- **`src/runtime/mod.rs`**: 
  - Relaxed the exact-match requirement in `startup_rpc_backend_device`. Pinned startup configs using `HIP<N>` will now gracefully match an explicit/resolved `ROCm<N>` device without throwing a conflict error.
  - Added test coverage (`pinned_gpu_startup_rpc_device_allows_lenient_amd_match`) for the lenient config matching.

**Scripts, CLI & Build Paths:**
- **`scripts/detect-llama-device.sh`**: The host detection and device probing logic now correctly returns and prioritizes `ROCm0`. `HIP0` is retained as a fallback in the preferred candidate list for compatibility with older binaries.
- **`src/cli/mod.rs`**: Modified the CLI help description for the `--device` flag to use `ROCm0` as the standard AMD example.
- **Verified Internal Mappings**: Confirmed that build scripts and Windows SDK paths correctly handle the transition, using `rocm` as the primary backend name while preserving necessary `hip` references for SDK/compiler paths.

### How was this tested?
- Compiled and tested locally using `Dockerfile.rocm` on a machine with an AMD Radeon RX 7900 XTX.
- Verified that `mesh-llm serve` successfully loads and serves both the main model and draft models on `ROCm0`.
- Ran `cargo test -p mesh-llm` to verify the new dual-matching and backward compatibility logic passes all existing and newly added unit tests.

### DIFF
--- a/docker/Dockerfile.rocm
+++ b/docker/Dockerfile.rocm
@@ -98,7 +98,7 @@ RUN cd llama.cpp && \
 
 FROM rocm/dev-ubuntu-24.04:7.0 AS runtime
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    ca-certificates libgomp1 libdbus-1-3 \
+    ca-certificates libgomp1 libdbus-1-3 hipblas rocblas rocm-smi \
     && rm -rf /var/lib/apt/lists/*
 COPY --from=rust-builder /out/mesh-llm /usr/local/bin/mesh-llm
 COPY --from=llama-builder-rocm /out/rpc-server-rocm /usr/local/lib/mesh-llm/bin/rpc-server-rocm
--- a/mesh-llm/src/cli/mod.rs
+++ b/mesh-llm/src/cli/mod.rs
@@ -336,7 +336,7 @@ pub(crate) struct Cli {
     #[arg(long, value_enum)]
     pub(crate) llama_flavor: Option<crate::inference::launch::BinaryFlavor>,
 
-    /// Device for rpc-server (e.g. MTL0, CUDA0, HIP0, Vulkan0, CPU).
+    /// Device for rpc-server (e.g. MTL0, CUDA0, ROCm0, Vulkan0, CPU).
     #[arg(long, hide = true)]
     pub(crate) device: Option<String>,
 
--- a/mesh-llm/src/inference/launch.rs
+++ b/mesh-llm/src/inference/launch.rs
@@ -682,7 +682,7 @@ fn resolve_device_for_binary(
             if available.iter().any(|candidate| candidate == device) {
                 return Ok(device.to_string());
             }
-            
+
             // Dual support for ROCm/HIP transition
             let is_amd_requested = device.starts_with("ROCm") || device.starts_with("HIP");
             if is_amd_requested {
--- a/mesh-llm/src/inference/launch.rs
+++ b/mesh-llm/src/inference/launch.rs
@@ -66,7 +66,7 @@ impl BinaryFlavor {
         match self {
             BinaryFlavor::Cpu => &["CPU"],
             BinaryFlavor::Cuda => &["CUDA0", "CPU"],
-            BinaryFlavor::Rocm => &["ROCm0", "CPU"],
+            BinaryFlavor::Rocm => &["ROCm0", "HIP0", "CPU"],
             BinaryFlavor::Vulkan => &["Vulkan0", "CPU"],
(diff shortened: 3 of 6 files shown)


i386 (approved):
LGTM - will merge this on green build and include in rc2

ndizazzo (changes requested):
@intenso thanks for the PR! The team doesn't have access to AMD hardware yet, so this is really helpful. 

After reading through it, I think the safest path is to accept both `HIP0` and `ROCm0` in Mesh-LLM’s resolver / validation paths first, and then converge on `ROCm0` once the whole stack is aligned. Right now the rename feels a bit partial, and partial `HIP0` → `ROCm0` migrations are where I think we can end up with exact-match failures in startup or device resolution even if the GPU is still being detected correctly.

The main places that still seem relevant are:
- `mesh-llm/src/inference/launch.rs`, where AMD launch-time device selection and fallback logic still appear to assume `HIP0`
- `mesh-llm/src/runtime/mod.rs`, where pinned startup stores the resolved `backend_device` and does exact matching against explicit `--device`
- `mesh-llm/src/system/hardware.rs`, where AMD inventory still maps to `HIP{index}` and that value flows outward into runtime / CLI surfaces

So my concern is less with the direction of the change and more that we still have a few legacy/internal `HIP0` expectations, and flipping only part of the stack creates the regression risk. I think one more focused pass to add dual / transition support would put this in a much safer place to merge.

There’s also a smaller benchmark / inventory wrinkle here: the benchmark engine itself doesn’t seem to care about `HIP0` vs `ROCm0`, but `mesh-llm gpus --json` and `mesh-llm gpu benchmark --json` currently emit the live `backend_device` string directly. So this would also change a documented machine-readable output surface, and current benchmark smoke coverage probably wouldn’t catch that.

intenso:
I agree, the initial patch was a bit partial. I was a bit hesitant about the scope since the HIP/ROCm relationship is quite confusing. I'll look into implementing dual support to accept both identifiers.

ndizazzo:
> I agree, the initial patch was a bit partial. I was a bit hesitant about the scope since the HIP/ROCm relationship is quite confusing. I'll look into implementing dual support to accept both identifiers.

No worries! The space moves fast - it's hard, if not impossible to keep up in every area at speed.

Don't hesitate to use an agent to dissect stuff either, we lean on them so we can cover ground (like the AMD GPU ecosystem) we lack hardware coverage in.

intenso:
I've updated the main PR description to reflect the full scope of these changes. Let me know how this looks.

ndizazzo:
@intenso looks good now! I can't rebase your fork, so if you dont mind, could you grab the latest copy of `main` on it, and do a quick linter format correction wit h`cargo fmt -p mesh-ll` , then we can merge 👍🏻

intenso:
Done!

ndizazzo (approved):

PR #322, opened 2026-04-17
Add experimental MoE floor strategies

i386:
This adds experimental MoE shared-core floor strategies to both `mesh-llm moe plan` and `mesh-llm serve`.

Users can now compare alternatives to the current fixed 50% heuristic without changing the default runtime behavior. That makes it possible to test more dynamic shard-floor policies against published `ranking.csv` and `analysis.json` artifacts before deciding whether any of them should become the default.

## What Users Can Do

Try these experimental floor modes in either `moe plan` or `serve`:
- `fixed_50pct`
- `topk_multiplier`
- `mass_threshold`
- `hybrid`

Optional tuning flags:
- `--experimental-moe-topk-multiplier`
- `--experimental-moe-mass-threshold`
- `--experimental-moe-floor-min`
- `--experimental-moe-floor-max`

Examples:

```bash
mesh-llm moe plan unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL \
  --experimental-moe-floor-mode hybrid \
  --experimental-moe-topk-multiplier 4 \
  --experimental-moe-mass-threshold 0.70
```

```bash
mesh-llm serve --model unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL \
  --experimental-moe-floor-mode mass_threshold \
  --experimental-moe-mass-threshold 0.70
```

## How It Works

The shared inputs are:
- `ranking.csv` for expert ordering and cumulative routing mass
- `analysis.json` for `expert_count`, `expert_used_count`, and shard byte accounting

This PR adds a floor-decision step before shard fit estimation. The derived `required_experts_per_node` is then passed into the existing assignment and byte-estimation flow.

Floor formulas:
- `fixed_50pct`: `ceil(expert_count * 0.5)`
- `topk_multiplier`: `expert_used_count * multiplier`
- `mass_threshold`: smallest top-N whose cumulative routing mass reaches the requested threshold
- `hybrid`: `max(topk_multiplier_result, mass_threshold_result)`

That floor is then used to:
- compute assignments with `compute_assignments_with_overlap(...)`
- estimate shard bytes from `analysis.json`
- report feasibility and recommended node count

## CLI Behavior

`moe plan` now:
- accepts the experimental floor flags
- prints the chosen floor strategy and derivation
- uses that derived floor when estimating shard fit

`serve` now:
- accepts the same experimental floor flags
- uses the selected strategy during published MoE preflight before download
- shows the floor mode in preflight progress and result messages
- caches preflight results per `(canonical model ref, floor strategy)` so alternate experiments do not reuse the wrong estimate

## Testing Plan

Goal: find a strategy that is safer and more efficient than `fixed_50pct` without recommending shard plans that fit on paper but fail or degrade badly in practice.

Test this initial strategy matrix:
- `fixed_50pct`
- `topk_multiplier` with `3`, `4`, `6`
- `mass_threshold` with `0.60`, `0.70`, `0.80`
- `hybrid` with:
  - `topk_multiplier=4`, `mass_threshold=0.60`
  - `topk_multiplier=4`, `mass_threshold=0.70`
  - `topk_multiplier=6`, `mass_threshold=0.70`

Use a small model set that includes:
- at least one highly concentrated MoE
- at least one flatter-routing MoE
- at least one smaller MoE
- at least one larger MoE
- at least one model that fits whole on a larger machine
- at least one model that only works as a split

Test each model across three VRAM classes:
- full-fit machine
- borderline machine where only some shard plans should fit
- no-fit machine where neither full model nor conservative split should fit

Phase 1: planner replay
- run `mesh-llm moe plan` for each `(model, VRAM budget, strategy)`
- capture:
  - derived `required_experts_per_node`
  - recommended node count
  - predicted max shard bytes
  - feasibility result

Phase 2: preflight behavior
- run `mesh-llm serve --model ...` with the same strategy matrix
- capture:
  - full-fit / shard-fit / no-fit decision
  - shared expert floor and node count
  - cache reuse on repeated runs
  - correct cache separation when the strategy changes
- `serve` and `moe plan` should agree for the same inputs

Phase 3: actual split validation
- for the most interesting predicted-fit cases, run the real split
- record:
  - predicted shard bytes
  - actual shard GGUF bytes
  - whether the shard actually launches on the target machine
- false positives are the main thing to avoid

Phase 4: quality spot checks
- for a smaller subset of fitted shard plans, run a fixed prompt suite
- compare against either:
  - full-model output on a larger machine, or
  - the current `fixed_50pct` baseline
- look for obvious failure modes:
  - incoherent output
  - repetition or degeneration
  - empty/broken answers
  - dramatic regression versus baseline

Record for each run:
- model and exact revision
- VRAM budget
- strategy and parameters
- derived shared expert floor
- recommended node count
- predicted shard bytes
- actual shard bytes, if tested
- preflight result
- launch result
- quality result

Decision rule:
1. minimize false-positive fit recommendations
2. among safe strategies, minimize required experts per node
3. among efficient safe strategies, prefer the simplest explanation

The most likely first default candidate to test is:
- `hybrid`
- `topk_multiplier=4`
- `mass_threshold=0.70`

## Scope

This PR is still experimental and flag-gated.

Default behavior does not change:
- `fixed_50pct` remains the default for both `moe plan` and `serve`
- users must opt in to other strategies explicitly

## Validation

- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm derive_floor_decision_ -- --nocapture`
- `cargo test -p mesh-llm estimate_startup_fit_`

### DIFF
--- a/mesh-llm/src/cli/commands/moe/mod.rs
+++ b/mesh-llm/src/cli/commands/moe/mod.rs
@@ -15,7 +15,7 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use crate::cli::moe::{HfJobArgs, MoeAnalyzeCommand, MoeCommand};
+use crate::cli::moe::{ExperimentalMoeFloorMode, HfJobArgs, MoeAnalyzeCommand, MoeCommand};
 use crate::cli::terminal_progress::start_spinner;
 use crate::cli::Cli;
 use crate::inference::moe;
@@ -52,6 +52,11 @@ pub(crate) async fn dispatch_moe_command(command: &MoeCommand, cli: &Cli) -> Res
             max_vram,
             nodes,
             dataset_repo,
+            experimental_moe_floor_mode,
+            experimental_moe_topk_multiplier,
+            experimental_moe_mass_threshold,
+            experimental_moe_floor_min,
+            experimental_moe_floor_max,
         } => {
             run_plan(
                 model,
@@ -60,6 +65,11 @@ pub(crate) async fn dispatch_moe_command(command: &MoeCommand, cli: &Cli) -> Res
                 max_vram.or(cli.max_vram),
                 *nodes,
                 dataset_repo,
+                *experimental_moe_floor_mode,
+                *experimental_moe_topk_multiplier,
+                *experimental_moe_mass_threshold,
+                *experimental_moe_floor_min,
+                *experimental_moe_floor_max,
             )
             .await
         }
@@ -104,6 +114,11 @@ async fn run_plan(
     max_vram: Option<f64>,
     nodes: Option<usize>,
     dataset_repo: &str,
+    floor_mode: ExperimentalMoeFloorMode,
+    topk_multiplier: Option<u32>,
+    mass_threshold: Option<f64>,
+    floor_min: Option<u32>,
+    floor_max: Option<u32>,
 ) -> Result<()> {
     if !json_output {
         eprintln!("📍 Resolving MoE model: {model}");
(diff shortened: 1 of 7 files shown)


ndizazzo (commented):
@i386 I'm a bit OOTL with the changes here, but `min_experts_per_node` is what makes the final call on shard assignment, so don’t we need to override/thread that through as well for runtime to match preflight?

i386:
@ndizazzo thanks thats corrected.

PR #339, opened 2026-04-19
Backpressure gate + TTFB-aware auto routing

michaelneale:
## What this does

When a model is slow or overloaded, auto routing now learns from experience and routes around it. A host-side backpressure gate rejects excess requests before they pile up in llama-server.

**Before:** A drowning host (e.g. MiniMax with 43k-token prompts) silently queued 30+ requests in llama-server, causing 5+ minute timeouts for everyone. Auto routing had no load signal and kept sending traffic.

**After:** The host rejects requests beyond its slot capacity with 429. Auto routing tracks response times per model — slow models get demoted, fast ones get picked. A single bad first experience (15s timeout) teaches the client to skip that model for the next 2 minutes.

### Host-side: backpressure gate (backend.rs)

The backend proxy rejects requests with 429 when inflight count reaches `n_parallel + 1`. This is the single choke point for ALL traffic — both local ingress (`:9337`) and remote tunnel requests converge here.

- Cap is set when llama-server goes healthy (default: 5 for 4 slots)
- 429 includes `"server overloaded"` body so clients can distinguish from "model not available"
- Requests beyond the cap never touch llama-server

### Client-side: TTFB tracker (affinity.rs)

Per-model ring buffer of recent time-to-first-byte samples. Purely local to each node.

- 8 samples per model, median-based slowness check
- 15s threshold — models with median TTFB over 15s are marked slow
- 2-minute sample expiry — stale data drops, models get retried
- Auto routing consults the tracker when picking models

### Client-side: auto routing backoff (router.rs, transport.rs, ingress.rs)

- `pick_model_classified_with_backoff` demotes slow models to the back of the candidate list
- `lookup_auto_model_if_responsive` breaks session stickiness when the cached model is slow
- First-auto timeout: 15s for small prompts with no history (learns fast), proportional to recent median otherwise
- Graceful degradation: if all models are slow, still picks one

### What explicit model requests get

No TTFB tracking, no adaptive timeouts. If you ask for MiniMax, you get MiniMax. The 429 gate is the only protection — it prevents the unbounded queue, the client gets a fast rejection.

## Test results

930 tests pass, 16 new (12 TTFB tracker, 4 router backoff).

### Live validation against flooded studio

| Scenario | Result |
|----------|--------|
| 8 concurrent requests to studio | 5 admitted, 3 rejected with 429 in <1ms |
| Auto after MiniMax timeout | Marked slow, next requests routed to gemma/GLM |
| Session stickiness on slow model | Memo forgotten, reclassified to healthy model |
| MiniMax direct request | Still works, 429 gate only protection |

## Files changed

11 files, +820/-118 (excluding upstream merge)

| File | Change |
|------|--------|
| `affinity.rs` | `ModelLatencyTracker`, ring buffer, `is_slow()`, sample expiry, session backoff |
| `backend.rs` | 429 gate in `handle_connection` |
| `router.rs` | `pick_model_classified_with_backoff` |
| `transport.rs` | TTFB recording, first-auto timeout, `RetryableOverloaded` handling |
| `ingress.rs` | Wire tracker through proxy paths |
| `mesh/mod.rs` | `max_inflight`, `is_overloaded()` on Node |
| `election.rs` | Pass Node to backend proxy |
| `runtime/mod.rs` | Set `max_inflight` at startup, create tracker |
| `runtime/proxy.rs` | Thread tracker parameter |
| `runtime/local.rs` | Pass Node to backend proxy |


mesh-llm/src/network/openai/transport.rs
```diff
@@ -2373,7 +2477,7 @@ pub async fn route_moe_request(
             RouteAttemptResult::RetryableContextOverflow => {
                 tracing::warn!("MoE target {target:?} rejected request with context overflow-style 400, trying next");
             }
-            RouteAttemptResult::RetryableUnavailable => {
+            RouteAttemptResult::RetryableUnavailable | RouteAttemptResult::RetryableOverloaded => {
```
i386, line 2480: `RetryableOverloaded` seems different from `RetryableUnavailable`: a 429 from the new backpressure gate means the peer is alive but busy, not dead. Calling `handle_peer_death()` here will evict healthy MoE peers under normal load and may destabilize routing. I think this branch should retry another target without marking the peer dead.

mesh-llm/src/network/openai/transport.rs
```diff
@@ -1990,17 +2051,23 @@ pub async fn handle_mesh_request(
...
             request.response_adapter,
+            auto_timeout,
         )
         .await
         {
             RouteAttemptResult::Delivered { status_code } => {
+                // Record TTFB for all requests (auto and explicit) so the
+                // tracker has full visibility into model responsiveness.
+                if let Some(ref name) = effective_model {
+                    ttfb_tracker.record(name, route_started.elapsed());
```
i386, line 2069: The tracker is described as TTFB-aware, but this sample is recorded after the whole routed request finishes. `route_started.elapsed()` includes relay/copy time and even earlier failed attempts, so long generations or retries will look like bad first-byte latency. That will skew both the slow-model backoff and `first_byte_timeout()` logic. I think this wants to record when the initial response probe succeeds, not after the full response is delivered.

mesh-llm/src/network/openai/backend.rs
```diff
@@ -78,8 +83,34 @@ pub(crate) async fn start_backend_proxy(llama_port: u16) -> Result<BackendProxyH
...
     let _ = stream.set_nodelay(true);
+
+    // Count this request. The guard drops when the connection finishes,
+    // decrementing the counter. This is the single source of truth for
+    // inflight tracking — all traffic (local ingress + remote tunnel)
+    // converges on this backend proxy.
+    let _inflight = node.begin_inflight_request();
+
+    // Backpressure gate: reject before touching llama-server.
+    if node.is_overloaded() {
```
i386, line 100: This looks off by one. We increment inflight first with `begin_inflight_request()`, then reject on `current >= max`. With `max_inflight = 5`, the 5th request is rejected, so the host admits only 4 requests rather than the advertised `n_parallel + 1`. If the intent is "allow 4 active + 1 queued", the overload check probably needs to happen before incrementing, or the predicate needs to use `>` instead of `>=`.

michaelneale:
I don't think I will use this one.

PR #346, opened 2026-04-19
Fix client join hang on unresponsive gossip peer

michaelneale:
## What changed

When `mesh-llm client --auto` joins a mesh, the gossip exchange with the bootstrap peer had no timeout. If the peer accepted the QUIC connection but never responded to the gossip stream, the join blocked forever — the management API never came up and fallback candidates were never tried.

## Root cause

Traced on a Mac Mini (M4 16GB) that consistently failed to join. The Nostr discovery returned 3 mesh candidates. Candidate 0 (peer b42128e8cc) accepted the QUIC connection and bi-stream but never sent a gossip response. `read_len_prefixed` (which calls `recv.read_exact`) hung indefinitely. The join loop never reached candidates 1 or 2.

Debugging sequence:
- `Node::start` — OK
- `PluginManager::start` — OK
- QUIC connect to peer — OK
- `open_bi()` — OK
- Write gossip payload — OK
- `read_len_prefixed` (read gossip response) — **hangs forever**

The blackboard plugin child process also prints `Error: early eof` to inherited stderr when it exits, which made the failure look like the parent process was crashing. The parent was actually alive but stuck on the gossip read.

The reconnect path in `_dispatch_streams` already had a 10s timeout around `initiate_gossip` — only the initial join path was missing it.

## Fix

Wrap `initiate_gossip` in a 15s timeout (matching the existing QUIC connect timeout). On timeout, the error propagates to the join loop, which catches it and tries the next candidate.

## Verified

Deployed to the Mac Mini that was consistently failing:
- Before: hangs forever on peer b42128e8cc, API never comes up
- After: gossip times out after 15s, falls through to candidate 1, joins mesh, API comes up

### DIFF
--- a/mesh-llm/src/mesh/gossip.rs
+++ b/mesh-llm/src/mesh/gossip.rs
@@ -102,7 +102,22 @@ pub(super) fn apply_transitive_ann(
 impl Node {
     /// Open a gossip stream on an existing connection to exchange peer info.
     pub(super) async fn initiate_gossip(&self, conn: Connection, remote: EndpointId) -> Result<()> {
-        self.initiate_gossip_inner(conn, remote, true).await
+        // Timeout the entire gossip exchange. A misbehaving peer may accept the
+        // QUIC connection and even the bi-stream but never send a gossip response,
+        // blocking the join path indefinitely and preventing fallback to other
+        // candidates. The 15s budget matches the QUIC connect timeout.
+        match tokio::time::timeout(
+            std::time::Duration::from_secs(15),
+            self.initiate_gossip_inner(conn, remote, true),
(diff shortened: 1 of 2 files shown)


ndizazzo:
@copilot fix both issues noted in your PR review [here](https://github.com/Mesh-LLM/mesh-llm/pull/346#pullrequestreview-4135513926)

ndizazzo:
@michaelneale not sure if copilot did the right thing in 57491525b9fd52b555941df11d09cee67eec0584 - might want to double check this

michaelneale:
@ndizazzo thanks - I think it did

PR #349, opened 2026-04-19
Add `--headless` flag to disable the web UI while keeping the management API alive

ndizazzo:
Operators running mesh-llm in programmatic or CI environments can now pass `--headless` to serve the full management API (`/api/*`) and inference endpoints (`/v1/*`) while returning 404 for all web console routes. No separate build needed, the UI is still embedded; headless mode just gates the routes at request time.

## What changed
**New flag**: `mesh-llm --headless` (works with `serve` and `client`/`--client` surfaces)
**Route behavior in headless mode**:
- `GET /api/*`, `/v1/*` → served normally (unchanged)
- `GET /`, `/dashboard`, `/chat/*`, `/assets/*`, `.png`/`.ico`/`.webmanifest`, non-API `.json` → 404
**Docker**: new `MESH_HEADLESS=1` env var on the official images
```sh
# API-only client node
docker run -e APP_MODE=console -e MESH_HEADLESS=1 mesh-llm:client
# API-only GPU worker
docker run --gpus all -e APP_MODE=worker -e MESH_HEADLESS=1 mesh-llm:cuda
```

## Architecture
No new state machine, no new startup paths. `--headless` is a boolean on the existing state, checked at the top of handle_request() before the main route match. Default is false; existing behavior is fully preserved.

## Validation
14 new unit tests across cli, api, and runtime modules. All pass. cargo fmt --check clean.

### DIFF
--- a/.github/workflows/docker-precheck.yml
+++ b/.github/workflows/docker-precheck.yml
@@ -74,18 +74,27 @@ jobs:
         run: |
           set -euo pipefail
           file="${{ inputs.dockerfile_path }}"
-          grep -q "COPY Cargo.toml" "$file"
-          grep -q "COPY Cargo.lock" "$file"
-          grep -q "COPY mesh-llm/Cargo.toml" "$file"
-          grep -q "COPY mesh-llm/build.rs" "$file"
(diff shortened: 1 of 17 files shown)


michaelneale (approved):
good idea

PR #354, opened 2026-04-20
auto: only join community mesh, never private named meshes

michaelneale:
## Summary

`--auto` now only joins the **community mesh** — other named meshes published on the same Nostr relay by other people are no longer candidates.

Before this change, `--auto` (with no `--mesh-name`) ranked every discovered mesh and filtered on score. Named meshes got a -200 penalty, but stickiness (+500) or a well-populated named mesh could still push them above zero, at which point auto would join them. You'd also see every named mesh in the `Found N mesh(es)` listing, even though most of them weren't relevant.

After this change, `--auto` only considers:

- Unnamed listings (`name: None`), or
- Listings with the blessed community name `"mesh-llm"` (case-insensitive).

Any other named mesh is still publicly discoverable on Nostr — to join it you must opt in explicitly with `--mesh-name <name>`.

### Terminology note

"Private" in mesh-llm means *not published to Nostr* — peers share the invite token out of band. A named listing that appears in Nostr discovery is **not** private; it's public, it just isn't the default. The PR uses "other named mesh" / "opt in by name" accordingly.

### Example output

```
$ mesh-llm --auto
...
  Found 2 mesh(es) (3 named mesh(es) hidden; use --mesh-name to join)
  · mesh-llm (score: 445, 4 nodes, 96GB, 2 clients)
  · unnamed (score: 115, 1 nodes, 24GB, 0 clients)
  Trying mesh mesh-llm (1/2)...
```

## Changes

- New helper `nostr::is_auto_eligible()` — single source of truth for "is this a `--auto` candidate".
- `smart_auto()` filters by eligibility when `target_name` is `None`.
- Drops the -200 penalty in `score_mesh` for non-community named meshes — it was both moot (they're filtered out now in auto mode) and wrong (when the user targets one via `--mesh-name`, the penalty would skew ranking).
- Runtime discovery printout hides non-eligible named meshes in auto mode and notes they were hidden plus how to opt in.

## Protocol

No wire-format or gossip changes. This is pure client-side join-policy behavior. Mixed-version meshes are unaffected — old nodes continue to do what they always did; new nodes are simply more selective about which meshes they'll auto-join.

## Tests

- New: `smart_auto_excludes_other_named_meshes` — non-community named meshes return `StartNew`, not `Join`.
- New: `smart_auto_mixes_unnamed_and_community` — both unnamed and `"mesh-llm"` eligible; community ranks first.
- New: `other_named_mesh_not_auto_eligible` — direct eligibility test, including case-insensitive `"MESH-LLM"`.
- Updated: `smart_auto_prefers_community_mesh` now compares community vs. unnamed (no longer relies on a ranked named mesh).
- Updated: `score_other_named_mesh_no_community_bonus` (renamed from `score_private_mesh_penalty`) reflects the dropped penalty.

Full lib suite: **962 passed, 0 failed**. `just build` clean.

### DIFF
--- a/mesh-llm/src/network/nostr.rs
+++ b/mesh-llm/src/network/nostr.rs
@@ -814,9 +814,12 @@ pub async fn discover(
 
 /// Is this mesh eligible for `--auto` when the user did not specify `--mesh-name`?
 ///
-/// `--auto` is for the public/community mesh only. Named third-party meshes are
-/// private groups and must be opted into explicitly via `--mesh-name`.
-/// Eligible = unnamed listing OR the blessed community name "mesh-llm".
+/// `--auto` joins the default community mesh. Eligible listings are:
+///   - unnamed (the implicit default), or
+///   - the blessed community name "mesh-llm".
+///
+/// Any other named mesh is still publicly discoverable on Nostr, but it is
+/// not the default — the user must opt in by name via `--mesh-name`.
 pub fn is_auto_eligible(mesh: &DiscoveredMesh) -> bool {
     match mesh.listing.name.as_deref() {
         None => true,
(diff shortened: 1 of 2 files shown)


ndizazzo (approved):

ndizazzo:
I think this might address https://github.com/Mesh-LLM/mesh-llm/issues/348

PR #357, opened 2026-04-20
Add per-model `parallel` config to control the number of parallel request slots per model

ndizazzo:
## Summary

Add per-model `parallel` config to control the number of parallel request slots per model. The global `[gpu].parallel` setting still works as a fallback, but each `[[models]]` entry can now override it independently.

## Background

The number of parallel slots was previously hardcoded at 4. Each parallel slot consumes `n_ctx / parallel_slots` tokens of context window. A node running multiple models at different scales may need different parallelism per model, a large 70B model might need only 1 slot while a small 4B model can handle 8.

### Problem

- A single global `parallel` value forces all models on the node to use the same slot count
- GPU-constrained nodes (like RTX 3080 with 10GB VRAM) may want 1 slot for a large model but 4 for a small one
- No way to express per-model concurrency without running separate node instances

## What changed

Users can now set `parallel` per model in `~/.mesh-llm/config.toml`:

```toml
version = 1

[gpu]
assignment = "auto"
parallel = 4        # global fallback; defaults to 4 when omitted

[[models]]
model = "Qwen3-70B-Q4_K_M"
parallel = 1        # this model gets 1 slot (overrides global)

[[models]]
model = "Qwen3-4B-Q4_K_M"
parallel = 8        # this model gets 8 slots (overrides global)
```

- `parallel: Option<usize>` added to `ModelConfigEntry` in `[[models]]`. Omitting it falls back to `[gpu].parallel`, which itself falls back to 4.
- Setting `parallel = 0` on a model is rejected at validation time with a clear error: `models[N].parallel must be at least 1, got 0`.
- The value propagates: `ModelConfigEntry.parallel` → `StartupModelSpec.parallel` → `StartupModelPlan.parallel` → `ElectionLoopParams.slots` → `ModelLaunchSpec.slots` → `--parallel` CLI arg to llama-server.
- Dynamic/runtime-loaded models (not declared in config) continue to use `[gpu].parallel` as before.

## Precedence

| Input | Behavior |
|-------|----------|
| `[[models]].parallel = N` | Use N slots for that model |
| No per-model `parallel`; `[gpu].parallel = M` | All models use M slots |
| Neither set | Default to 4 |

## Protocol

No wire format changes. `parallel` is a local-only config field — it is not gossiped to peers and does not affect mesh protocol compatibility. Older nodes that lack the field will continue to use their own default of 4 slots.

## Compatibility

Non-breaking. Omitting `parallel` on all models preserves existing behavior. Existing configs with only `[gpu].parallel` continue to work unchanged.

## Example

Multi-model node with mixed GPU budgets:

```toml
version = 1

[gpu]
assignment = "pinned"

[[models]]
model = "Qwen3-70B-A3B-UD-Q4_K_XL"
gpu_id = "pci:00000000:01:00.0"
ctx_size = 65536
parallel = 1        # single slot — large model, tight VRAM

[[models]]
model = "Qwen3-4B-Q4_K_M"
gpu_id = "pci:00000000:06:00.0"
ctx_size = 32768
parallel = 8        # 8 slots — small model, plenty of headroom
```

### DIFF
--- a/mesh-llm/src/inference/election.rs
+++ b/mesh-llm/src/inference/election.rs
@@ -971,6 +971,7 @@ struct MoeElectionParams {
     pinned_gpu: Option<crate::runtime::StartupPinnedGpuTarget>,
     target_tx: Arc<watch::Sender<ModelTargets>>,
     stop_rx: watch::Receiver<bool>,
+    slots: usize,
 }
 
 struct StartLlamaParams<'a> {
@@ -988,6 +989,7 @@ struct StartLlamaParams<'a> {
     binary_flavor: Option<launch::BinaryFlavor>,
     ctx_size_override: Option<u32>,
     pinned_gpu: Option<&'a crate::runtime::StartupPinnedGpuTarget>,
+    slots: usize,
 }
 
 pub struct ElectionLoopParams {
@@ -1009,6 +1011,7 @@ pub struct ElectionLoopParams {
     pub moe_runtime_options: moe::MoeRuntimeOptions,
     pub target_tx: Arc<watch::Sender<ModelTargets>>,
     pub stop_rx: watch::Receiver<bool>,
+    pub slots: usize,
 }
 
 fn spawn_moe_analysis_spinner(
@@ -1516,6 +1519,7 @@ pub async fn election_loop(
         moe_runtime_options,
         target_tx,
         mut stop_rx,
+        slots,
     } = params;
     let mut peer_rx = node.peer_change_rx.clone();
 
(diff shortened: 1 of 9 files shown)


michaelneale (approved):
Is there some heuristics we can add so it isn't always 4 unless overridden? Might be nice to make defaults work well

ndizazzo:
> Is there some heuristics we can add so it isn't always 4 unless overridden? Might be nice to make defaults work well 

@michaelneale Yeah, good idea. I can include that in some of the context work I was talking about a few weeks ago. I'll open up an issue we can jam on.

ndizazzo:
@michaelneale Added #359

PR #358, opened 2026-04-20
Corrects duplicated log extension on logfiles

ndizazzo:


michaelneale (approved):
Looks like conflicts but thanks

PR #363, opened 2026-04-21
Fix: Add full workspace copy to Dockerfiles to resolve missing crates

intenso:
Hi! 
I was trying to build the Docker images from a clean environment and ran into some compilation errors (specifically `can't find crate for iroh`). It looks like the recent restructuring into separate workspace members (`mesh-client`, `mesh-api`, etc.) hasn't fully made its way into the Dockerfiles yet, so `cargo-chef` occasionally misses some internal dependencies during the prepare/cook stages.

To help resolve this and hopefully make the Docker builds a bit more resilient to future workspace additions, I've added a `COPY . .` command to both the `rust-deps` and `rust-builder` stages across the core Dockerfiles.

**Notes on the changes:**
* This ensures `cargo-chef` can see all workspace `Cargo.toml` files and that all `build.rs` scripts (e.g., for Protobufs) have the context they need.
* I noticed that `.github/workflows/docker-precheck.yml` relies on explicit `COPY` commands (like `COPY mesh-llm/proto/`). To keep the CI green, I left these manual `COPY` lines intact right above the new `COPY . .` step. I realize this looks a bit like a workaround/hack. While it might make sense to update the precheck script eventually to allow for a global copy, I figured there might be historical reasons or context I'm missing as to why those strict grep checks are there.

**Testing:**
I've verified the builds locally for `Dockerfile.cuda`, `Dockerfile.rocm`, `Dockerfile.vulkan`, `Dockerfile.cpu`, and `Dockerfile.client`.

*(Minor tip: If anyone is testing this locally and still encounters errors, it might be necessary to clear the BuildKit mount cache first via `docker builder prune --filter type=exec.cachemount`, as the persistent cache can sometimes hold onto the broken state from previous attempts).*

Let me know if this looks okay.

### DIFF
--- a/.github/workflows/docker-precheck.yml
+++ b/.github/workflows/docker-precheck.yml
@@ -75,26 +75,31 @@ jobs:
           set -euo pipefail
           file="${{ inputs.dockerfile_path }}"
           if grep -qE '^\s*COPY \. \.' "$file"; then
-            # COPY . . copies the entire workspace including all manifests — accept as equivalent
-            if grep -qE "echo.*fn main|stub.*lib" "$file"; then
-              echo "FAIL: stub lib.rs shortcut detected in $file" >&2
-              exit 1
-            fi
-            echo "PASS: COPY . . copies all workspace manifests"
-          else
-            grep -q "COPY Cargo.toml" "$file"
-            grep -q "COPY Cargo.lock" "$file"
-            grep -q "COPY mesh-llm/Cargo.toml" "$file"
-            grep -q "COPY mesh-llm/build.rs" "$file"
-            grep -q "COPY mesh-llm/plugin/Cargo.toml" "$file"
-            grep -q "COPY mesh-llm/plugin/build.rs" "$file"
-            grep -q "COPY mesh-llm/proto/" "$file"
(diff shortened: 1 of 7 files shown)


ndizazzo (changes requested):
@intenso we definitely want to avoid blanket `COPY . .` steps, since:

1) Packages can unintentionally be included in the docker image that shouldn't be
2) It balloons the size of the image with specifics that aren't needed for the specific image

I think all we'd need is a `COPY mesh-client/ mesh-client/` if you want to try that out. Once it's sorted, happy to merge!

intenso:
It took a little bit of tuning to get right. Just copying `mesh-client/` broke the build during `cargo chef prepare` and `cargo build --locked` because Cargo requires all workspace members and the root lockfiles to be present to override the chef skeleton. 
 
I also made sure to explicitly keep the specific manifest copies (like `COPY mesh-llm/Cargo.toml`) so that it gracefully passes the GitHub Actions `docker-precheck` job, while appending the rest of the workspace members to satisfy Cargo's workspace resolution.

ndizazzo (approved):

ndizazzo:
Thanks @intenso! These should land in the next release

PR #366, opened 2026-04-21
Add delete option to model CLI flags to remove just a single managed model

ndizazzo:
Users can now remove an installed model from the local Hugging Face cache from the `models` CLI, using the same refs that `models installed` prints.

This change keeps delete aligned with the existing model resolver semantics instead of introducing separate path heuristics, and it deletes only the exact matched GGUF file set from the HF cache.

## How delete works

- `mesh-llm models installed` prints a delete-ready ref for each installed model.
- `mesh-llm models delete <ref>` resolves that input through the existing model resolver semantics.
- Supported delete inputs are the installed forms we already understand, including local installed stems and Hugging Face refs like `org/repo` and `org/repo:VARIANT`.
- For HF-backed installs, delete uses `huggingface_hub_rust` cache metadata to find the exact cached repo revision and matched GGUF file(s).
- If the selected variant is a split GGUF, delete removes the whole matched shard set and does not widen the deletion to unrelated sibling variants in the same snapshot.
- Direct URLs and filesystem paths are rejected for delete.

## Example output

```text
$ ./target/release/mesh-llm models installed
📦 qwen3.5-moe-0.87B-d0.8B.Q2_K
   type: 🦙 GGUF
   size: 627MB 📏
   owner: external
   capabilities: 💬 text
   ref: Flexan/kshitijthakkar-qwen3.5-moe-0.87B-d0.8B-GGUF:Q2_K
   show: mesh-llm models show Flexan/kshitijthakkar-qwen3.5-moe-0.87B-d0.8B-GGUF:Q2_K
   download: mesh-llm models download Flexan/kshitijthakkar-qwen3.5-moe-0.87B-d0.8B-GGUF:Q2_K
   delete: mesh-llm models delete Flexan/kshitijthakkar-qwen3.5-moe-0.87B-d0.8B-GGUF:Q2_K
```

That printed `delete` command is valid for dotted GGUF quant names like `Q2_K`, and split GGUF selector refs remove only the matching shard set.

## Why

The original delete path drifted away from the shared resolver and then expanded deletion from one resolved file into directory-prefix based sibling deletion. That created two problems:

- installed refs shown by the CLI were not always accepted by delete
- delete could remove extra GGUFs that were not part of the resolved match

This change moves delete back onto the shared resolver semantics and keeps the final deletion set driven by exact HF cache matches.

## Validation

- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm models::delete_tests --lib -- --nocapture`
- `cargo test -p mesh-llm models::resolve::tests::dotted_quant_selector_resolves_to_single_file_gguf --lib -- --nocapture`

`cargo check` and tests still show the same pre-existing unrelated warnings in `mesh-llm/src/inference/election.rs` and `mesh-llm/src/runtime/local.rs`.

### DIFF
--- a/mesh-llm/src/models/delete.rs
+++ b/mesh-llm/src/models/delete.rs
@@ -1,10 +1,13 @@
+use std::collections::BTreeSet;
 use std::path::PathBuf;
 
 use anyhow::{bail, Context, Result};
+use hf_hub::types::cache::HFCacheInfo;
 
-use crate::models::usage::delete_model_by_path;
+use crate::models::local::{huggingface_hub_cache_dir, scan_hf_cache_info, split_gguf_base_name};
+use crate::models::usage;
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct DeleteResult {
     pub deleted_paths: Vec<PathBuf>,
     pub reclaimed_bytes: u64,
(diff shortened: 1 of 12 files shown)


ndizazzo:
@i386 fixed split GGUF in 4d20d73!

i386 (commented):
A few blocking review comments on the delete flow and test coverage.

mesh-llm/src/models/delete.rs
```diff
@@ -0,0 +1,307 @@
...
+///
+/// Resolution strategy:
+/// - If `identifier` contains `/`, treat it as a HF repo ID and scan HFCacheInfo for any gguf files
+///   in repos whose repo_id starts with `identifier`.
+/// - Otherwise, treat it as a stem (filename base without `.gguf`) and look for exact filename
+///   matches OR split-GGUF patterns (`stem-00001-of-NNNNN.gguf`).
+///
+/// Returns an error if zero files match (NotFound), or if multiple distinct repos match
+/// and cannot be disambiguated by stem analysis (Ambiguous).
+pub fn resolve_model_identifier(identifier: &str) -> Result<Vec<PathBuf>> {
```
i386: `models delete` still advertises exact-path input, but `resolve_model_identifier` now treats every identifier containing `/` as a Hugging Face repo prefix and never checks the filesystem. On Unix that means `/path/to/model.gguf` always returns `Model not found`, even when the file exists, so the previous exact-path workflow is broken. The resolver needs to detect existing `.gguf` paths before falling back to repo-id matching.

mesh-llm/src/models/delete.rs
```diff
@@ -0,0 +1,307 @@
...
+
+/// Execute a model deletion by identifier, collecting and removing ALL files.
+pub async fn delete_model_by_identifier(identifier: &str) -> Result<DeleteResult> {
+    let primary_paths = resolve_model_identifier(identifier)?;
+
+    if primary_paths.is_empty() {
+        bail!("Model not found: {}", identifier);
+    }
+
+    let primary_path = &primary_paths[0];
```
i386: This code already has the exact `HFCacheInfo` matches in `primary_paths`, but it throws that set away, keeps only `primary_paths[0]`, and then deletes every sibling `.gguf` whose stem starts with `base_stem-`. In a repo snapshot that contains multiple related variants, deleting one model can remove extra GGUFs that were never matched by `huggingface_hub_rust`. To keep this safe, delete the resolved HF files directly and add only explicitly related sidecars or shards.

mesh-llm/src/models/mod.rs
```diff
@@ -1,10 +1,13 @@
 pub mod capabilities;
 pub mod catalog;
+pub mod delete;
```
i386, line 3: The new end-to-end delete coverage lives in `src/models/delete_tests.rs`, but nothing includes that file in the module tree, so Cargo ignores it. That means the split-GGUF path is currently untested, and the stale imports in that file (`delete_model_at_path`, `ResolveError`) also go unnoticed. Add `#[cfg(test)] mod delete_tests;` or move the cases under `delete.rs` before relying on this coverage.

PR #368, opened 2026-04-21
Fix passing parallel via config.toml

ndizazzo:
No idea what happened, but a merge mangled the slot fixes I had. I added tests to capture regressions for this, and added them back.


i386 (approved):

michaelneale (approved):
I like 42 as test value.

PR #371, opened 2026-04-21
Add support to explicitly publish mesh via command-line

### LINKED ISSUE #240: revise mesh publication rules
## Goal
- **Primary outcome**
  - Users can name a mesh without accidentally making it public, and can only make a mesh publicly discoverable through an explicit publication action.
- **Non-goals**
  - Redesigning Nostr discovery itself
  - Changing mesh invite-token behavior for private meshes
  - Changing mesh/network security guarantees beyond clarifying CLI intent

## Context
- **Existing capabilities**
  - `mesh-llm serve --publish` makes a mesh discoverable via Nostr.
  - `mesh-llm serve --mesh-name <name>` currently sets a discovery name.
  - `--auto` and `discover` use Nostr listings to find public meshes.
- **Current limitations / problems**
  - `--mesh-name` currently auto-enables `--publish` in runtime, so naming a mesh silently changes privacy/discoverability.
  - This is documented in the README, but not obvious from CLI semantics.
- **Why now**
  - Issue #197 reports that users can reasonably stumble into joining or creating a public mesh without realizing naming implies publication.

## Surfaces
- **Surface**
  - CLI
  - internal
  - docs
- **Entry points**
  - `mesh-llm serve`
  - CLI help text
  - README usage examples
- **Inputs**
  - `--mesh-name <name>`
  - `--publish`
- **Defaults**
  - Meshes are private by default.
  - Naming alone does not publish.
- **No-input behavior**
  - Existing default behavior remains: unnamed private mesh unless another explicit public-discovery mode is selected.

## Behavior
- **Execution flow**
  1. User starts a mesh with no publication flag.
  2. If they pass `--mesh-name <name>`, the mesh stores and uses that name locally/internally, but does not publish to discovery.
  3. If they explicitly pass `--publish`, the mesh is published to Nostr and becomes discoverable.
  4. If both `--mesh-name <name>` and `--publish` are provided, the published listing uses that mesh name.
  5. Help text, startup logs, and docs make the public/private distinction explicit.
- **Validation**
  - **Valid**
    - `--mesh-name <name>` alone
    - `--publish` alone
    - `--publish --mesh-name <name>`
  - **Rejected**
    - Empty mesh names
    - Invalid future combinations only if new CLI syntax is introduced; otherwise existing valid flags remain accepted
- **Error handling**
  - **On failure**
    - If publish/discovery setup fails, startup reports the failure and does not silently claim the mesh is public.
  - **User-visible errors**
    - Clear CLI/runtime message when publication was requested but could not be established.
  - **Silent failures**
    - not allowed

## State & Data
- **New data introduced**
  - none
- **Existing data modified**
  - CLI semantics for how `mesh_name` and `publish` interact
  - Nostr listing behavior when a mesh name exists without publication
- **Persistence**
  - not required
- **Backward compatibility**
  - Behavior change: `--mesh-name` no longer implies public publication.
  - A deprecation/warning path should be used for old expectations if needed.

## Integration Points
- **Depends on**
  - CLI parsing and validation
  - runtime startup logic
  - Nostr publish/discovery flow
- **Affects**
  - README and CLI help text
  - startup status messages
  - tests covering discovery/public mesh behavior
- **External systems**
  - Nostr relays

### Interface Layer
- Clarify CLI semantics so publication is explicit and naming is separate.
- Update help text to describe private vs public behavior directly.

### Application Layer
- Remove the runtime coupling that force-enables `publish` when `mesh_name` is present.
- Preserve explicit `publish` behavior.

### Core Logic
- Keep discovery/publication logic unchanged except for requiring explicit publication intent.
- Ensure named private meshes do not create public listings.

### Data Layer
- none

### External / Infrastructure
- Public Nostr publication only occurs when explicitly requested.

### Observability
- **Logs**
  - Startup should state whether the mesh is private or published.
  - If published, log the mesh name used in the listing.
- **Metrics**
  - none
- **Alerts**
  - none

## Edge Cases
- User runs `mesh-llm serve --mesh-name "eng-west"` and expects a private departmental mesh; it must stay private unless `--publish` is also set.
- User runs `mesh-llm serve --publish` with no mesh name; publication still works with current unnamed/autogenerated listing behavior.
- User runs `mesh-llm serve --auto --mesh-name "poker-night"`; the name should act as a discovery filter for matching public meshes, but must not itself force publication when starting a new mesh.
- User upgrades from prior behavior and still expects `--mesh-name` to publish; CLI/help/runtime messaging should make the changed behavior obvious.
- Publication fails after explicit `--publish`; the process must report that the mesh is not publicly discoverable.

## Constraints & Tradeoffs
- **Constraint**
  - Keep protocol/discovery internals stable; this is primarily a CLI semantics and UX fix.
- **Tradeoff**
  - Separating naming from publication may add one extra flag for public meshes, but removes a risky implicit privacy change.
- **Alternative considered**
  - Make `--publish` take a name and deprecate `--mesh-name`; rejected for this spec because separating `name` from `public` is clearer and preserves the private-named-mesh use case.

## Acceptance Criteria
- [ ] Given a new mesh, when the user runs `mesh-llm serve --model X --mesh-name "team-a"`, then the mesh starts private and does not publish a Nostr listing.
- [ ] Given a new mesh, when the user runs `mesh-llm serve --model X --publish --mesh-name "team-a"`, then the mesh publishes a discoverable listing named `team-a`.
- [ ] Given a new mesh, when the user runs `mesh-llm serve --model X --publish`, then publication still works without requiring a mesh name.
- [ ] Given invalid input, when the user passes an empty/invalid mesh name, then startup returns a clear error.
- [ ] Defaults behave as specified.
- [ ] No regression in existing private invite-token join behavior.
- [ ] No regression in public mesh discovery for explicitly published meshes.

## Rollout / Migration
- **Rollout strategy**
  - Ship as a CLI behavior change with updated help text and README examples.
- **Migration required**
  - yes: users relying on `--mesh-name` to publish must add `--publish`
- **Backward compatibility**
  - Behavior changes at the CLI level; consider a temporary warning if `--mesh-name` is used without `--publish` to explain the new semantics.

## Notes
- This spec intentionally chooses the “separate naming from publication” direction discussed in issue #197 because it preserves both use cases: private named meshes and explicitly public named meshes.

## Related
- Issue #197: https://github.com/michaelneale/mesh-llm/issues/197

ndizazzo:
Previously, passing `--mesh-name "my-mesh"` implicitly enabled Nostr publication without any warning or user consent. Named meshes were silently discoverable by anyone browsing Nostr, and new meshes created via `--auto` also had publish auto-enabled. This made it unclear whether a mesh was private or public, and users couldn't intentionally keep a named mesh off the directory.

This change makes publication intent **explicit**: only `--publish` enables Nostr discovery. Named meshes stay private by default, with a clear startup message explaining how to make them public. The frontend now derives its public/private state from an explicit `publication_state` field instead of inferring it from a boolean `nostr_discovery` flag.

## What changes for users

| Before | After |
|---|---|
| `--mesh-name "poker-night"` → silently published to Nostr | Prints `"ℹ️ Mesh named 'poker-night' — private by default. Add --publish to make it publicly discoverable."` |
| `--auto` creating a new mesh → auto-enables publish | Shows `"Mesh is PRIVATE — add --publish for public discovery"` |
| UI public/private derived from `nostr_discovery` bool | Derived from explicit `publication_state: "private" \| "public" \| "publish_failed"` |
| No feedback when `--publish` fails (crashes on key load error) | Graceful degradation: prints warning, sets `PublishFailed` state, mesh keeps running privately |

**Recommended usage:**
- Private mesh (default): `mesh-llm serve --model Qwen2.5-3B`
- Public/visible on Nostr: `mesh-llm serve --model Qwen2.5-3B --publish`  
- Named + public (buddy mode): `mesh-llm serve --auto --model Qwen2.5-3B --mesh-name "poker-night" --publish`
- Discover existing meshes: `mesh-llm serve --auto`

## What changes for developers

- **New API field**: `/api/status` now includes `publication_state` (`"private"`, `"public"`, or `"publish_failed"`). The legacy `nostr_discovery` boolean remains for backward compatibility with older UI versions.
- **New Rust enum**: `api::PublicationState` with `Private`, `Public`, `PublishFailed` variants, serializable to JSON strings.
- **Frontend derivation**: `isPublicMesh` in App.tsx checks `publication_state === 'public'` first, falls back to `nostr_discovery` for older APIs.
- **7 new unit tests** covering the publication-state matrix at runtime startup and `start_new_mesh()` helper.

## Architecture

The core change is removing two implicit `cli.publish = true` gates:

1. `runtime/mod.rs`: Removed `if cli.mesh_name.is_some() && !cli.publish { cli.publish = true; }`
2. `runtime/discovery.rs`: Removed auto-enable-publish from `start_new_mesh()` — replaced with truthful status messaging based on whether the user already passed `--publish`.

A new `publication_state` field flows through the entire stack: CLI flag → runtime → `ApiInner` → `/api/status` response → TypeScript types → frontend public/private derivation → AppHeader launcher guidance.

## Protocol Compatibility

This is a **non-breaking additive change**. The existing `nostr_discovery` bool remains in the API response alongside the new `publication_state` string field. Older UIs continue to work via the fallback path. The Nostr publish loop behavior is unchanged when `--publish` is explicitly set.

## Validation

- Rust: 982 tests pass (including 7 new publication-state matrix tests)
- UI: 129 tests pass, TypeScript compiles cleanly
- Full build (`just build`) passes on release profile

Closes #240


michaelneale (approved):

i386 (approved):
Love that this foot-gun is fixed!

PR #376, opened 2026-04-22
paginate long model listings in interactive terminals

i386:
Users can now page through long `mesh-llm models` console listings with `less`-style navigation instead of having large result sets scroll past in one burst.

## What changed

- added a small CLI pager helper that routes buffered output through `less` in interactive terminals
- updated `models search`, `models installed`, and `models recommended` console rendering to buffer their output and send it through the pager helper
- kept JSON output, redirected output, and non-interactive terminals on the existing direct stdout path, with a plain-print fallback when `less` is unavailable

## Why

The long model listing commands were writing directly to stdout, which made interactive browsing awkward once the output exceeded a screenful. This change gives those commands the expected terminal navigation behavior while preserving scripting-friendly output.

## Validation

- `cargo check -p mesh-llm`

## Commands

- `mesh-llm models search qwen`
- `mesh-llm models installed`
- `mesh-llm models recommended`


mesh-llm/src/cli/pager.rs
```diff
@@ -0,0 +1,94 @@
...
+        std::io::stdin().is_terminal(),
+        std::io::stdout().is_terminal(),
+        std::env::var_os("TERM").as_deref(),
+    ) {
+        return print_direct(output);
+    }
+
+    match page_with_less(output) {
+        Ok(()) => Ok(()),
+        Err(err) if pager_missing(&err) => print_direct(output),
```
ndizazzo, line 20: On small NIT - we can probably gate on ANY pager error, including missing or non-zero codes

ndizazzo (approved):
Only one comment that doesn't matter too much right now!

PR #384, opened 2026-04-24
Yield model serving when you need your GPU back

michaelneale:
## What's new

Your laptop can now back off mesh serving when **you** need the GPU, and come back online when you step away. Two ways in:

**Manual on/off** — when you want to take the machine back right now:

```bash
mesh-llm yield    # stop serving; peers will route around this node
mesh-llm resume   # serve again
```

**Automatic presence-based yield** — opt in at startup:

```bash
mesh-llm --auto --yield-on-presence \
         --yield-idle-secs 180 \
         --yield-active-grace 10
```

While you're actively using the machine (keyboard / mouse / trackpad within `--yield-active-grace` seconds), mesh-llm stops local model serving and releases the GPU. After `--yield-idle-secs` of idleness it quietly resumes.

The mesh notices immediately: other peers that host the same models take over, and `/api/status` reports a new `yielded` field (`"manual" | "user_active" | null`) so the UI and peers can show why a node isn't serving right now.

### Safety rules

- **Auto-yield never strands a model.** If this node is the last host for any of its models, a presence-triggered yield is refused. Manual `mesh-llm yield` bypasses this (you explicitly asked for the machine back).
- **Manual yield is sticky.** The presence loop cannot auto-resume a node that was yielded by hand — only `mesh-llm resume` (or `POST /api/resume`) clears it.
- **Fail-open on unknown platforms.** The presence probe is real on macOS (`CGEventSource`) and Windows (`GetLastInputInfo`). Linux returns `Presence::Unknown` and we never auto-yield on unknown.

### Example

```
$ mesh-llm yield
⏸ Yielded — stopped local model serving.
   If other peers host the same models they will take over; otherwise
   those models are unavailable on the mesh until you run `mesh-llm resume`.

$ curl -s localhost:3131/api/status | jq .yielded
"manual"

$ mesh-llm resume
▶ Resumed — reloading 2 model(s).
```

## Architecture

- New `system::presence` module with a `PresenceProbe` trait and per-OS implementations (macOS / Windows real, Linux `Unknown`).
- New `presence_yield_loop` polls every 5s and talks to the existing `RuntimeControlRequest` channel — the same mechanism `Load` / `Unload` use. No new runtime-wide state machine; yield is just two more control messages.
- `yield_owner` lives inside the runtime so stickiness of a manual yield is enforced in one place, not split between the probe loop and the CLI handler.
- `Resume` re-issues `Load` for each previously-yielded model through the control channel rather than re-running election / split setup. Per-Load failures are logged so a broken model file can't silently leave you in a partially-resumed state.

## Protocol

No wire-format change. `RuntimeControlRequest` is an in-process enum; the new `Resume { reason }` field is purely internal. Mesh gossip is unchanged. The `/api/status` response grows an optional `yielded` field which older consumers simply ignore.

## Validation

- `cargo check -p mesh-llm` ✓
- `cargo test -p mesh-llm --lib` ✓ (1061 passed, 0 failed)
- New unit tests cover the `last_host_for_any` guard (4 cases) and the presence loop state machine (yield-then-resume happy path + the regression test for "refused yield must not trigger a spurious resume").

Manual cross-machine validation still recommended per the deploy checklist before rolling to anarchai.org.


ndizazzo (changes requested):
Really good idea! 

One smalll enhancement to the status endpoint might be to add a `yielded_at` or even a `yields: [ {...} ]` array with timestamps, so we can later get data on that too :) 

One thing I think this could use is _exact_ parameter capture for how we start the inference server. That way, when we resume it, we are sending the exact same arguments back into it. We could define some kind of shape to capture the data in a standard way (so when vLLM or other engines arrive later), we can pick that data out from the descriptor and generate CLI arguments to start the inference server with again... 

That would make yield/resume restore the exact serving configuration rather than reloading by name with current defaults... which I see as especially important for startup-managed or election-managed models.

michaelneale:
I am not happy with this. Will come back another day.

PR #385, opened 2026-04-25
Preflight Vulkan launches against primary GPU VRAM

i386:
﻿## Summary

This adds a local-launch preflight for Vulkan and Metal so mesh-llm does not treat aggregate local GPU capacity as if it were available to the primary backend device.

For Vulkan/Metal solo launches, the fit check now uses the first advertised GPU VRAM value. CUDA/ROCm keep aggregate capacity because local multi-GPU row split can be valid there.

The build scripts also stage flavored llama.cpp runtime binaries consistently on Linux, macOS, and Windows, and clear stale flavored server binaries before copying the current backend.

## Validation

- cargo test -p mesh-llm local_launch_preflight
- cargo test -p mesh-llm pinned_gpu_startup_preflight
- cargo test -p mesh-llm backend_device_for_flavor_uses_backend_namespace
- cargo check -p mesh-llm
- cargo fmt --all -- --check
- bash -n scripts/build-linux.sh
- PowerShell parser check for scripts/build-windows.ps1

Could not run zsh -n scripts/build-mac.sh in this Windows environment because zsh is not installed.

### DIFF
--- a/mesh-llm/src/runtime/mod.rs
+++ b/mesh-llm/src/runtime/mod.rs
@@ -1160,11 +1160,11 @@ pub(crate) async fn run() -> Result<()> {
         Some(d) => d.clone(),
         None => detect_bin_dir()?,
     };
-    let rpc_binary_flavor = if config.gpu.assignment == plugin::GpuAssignment::Pinned {
-        launch::resolve_binary_flavor(&bin_dir, "rpc-server", cli.llama_flavor)?
-    } else {
-        None
-    };
+    let rpc_binary_flavor =
+        launch::resolve_binary_flavor(&bin_dir, "rpc-server", cli.llama_flavor)?;
+    if cli.llama_flavor.is_none() {
+        cli.llama_flavor = rpc_binary_flavor;
+    }
     preflight_config_owned_startup_models(
         &config,
         &startup_specs,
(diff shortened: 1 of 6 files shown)


mesh-llm/src/runtime/mod.rs
```diff
@@ -569,11 +569,11 @@ pub(crate) async fn run() -> Result<()> {
...
     };
-    let rpc_binary_flavor = if config.gpu.assignment == plugin::GpuAssignment::Pinned {
-        launch::resolve_binary_flavor(&bin_dir, "rpc-server", cli.llama_flavor)?
-    } else {
-        None
-    };
+    let rpc_binary_flavor =
+        launch::resolve_binary_flavor(&bin_dir, "rpc-server", cli.llama_flavor)?;
+    if cli.llama_flavor.is_none() {
+        cli.llama_flavor = rpc_binary_flavor;
```
ndizazzo, line 1177: Could we make the pinned preflight use the same probed backend-device list as launch-time `resolve_device_for_binary()`? That would keep the preflight and launch paths using the same source of truth, and preserve the pinned-GPU behaviour we want for startup models. 

As written, I think this could pass preflight but still fail at runtime if llamacpp doesn’t expose the synthesized backend device.

ndizazzo: Forgot to tag you @i386

PR #386, opened 2026-04-25
feature(metrics-collector): add support to centralize storage for met…

ndizazzo:
## Summary

This adds a local runtime metrics surface for the llama.cpp process that backs a mesh node. Operators and the web console can now inspect llama.cpp health, selected Prometheus samples, and slot activity through `/api/runtime/llama` without mixing that data into mesh routing metrics.

- Adds `/api/runtime/llama` as a local-only management API payload with endpoint status, last attempt/success timestamps, error state, bounded raw diagnostics, and typed `items` for UI consumers.
- Polls llama.cpp `/metrics` and `/slots` non-fatally: unsupported endpoints are reported as unavailable, transient failures keep the previous successful snapshot, and oversized responses are rejected instead of buffered unboundedly.
- Parses only a permit-listed set of llama.cpp metric names and labels, preserving useful request/token/slot signals while avoiding arbitrary metric cardinality in the API.
- Adds typed slot items that preserve slot index and busy state so the dashboard can show total/busy slot counts reliably even when raw slot diagnostics are capped.
- Documents the runtime metrics classification in `METRICS.md` and clarifies that these local process metrics are separate from routing metrics and do not affect model routing decisions.

## Screenshot
https://github.com/user-attachments/assets/9c4a6cfd-8de3-46b4-9056-e9a245041369

**Basic runtime status collection per node**

## Architecture

Conforms to metrics definitions in METRICS.md.

- Introduces runtime llama metric and slot snapshots under `runtime_data`, published through the existing collector/producer path so API responses and UI status types stay additive.
- Separates schema definitions into `mesh-llm/src/runtime_data/metrics.rs`, giving future runtime metric types a single place to grow instead of expanding the general snapshot module.
- Keeps raw llama.cpp slot data available for local debugging, but adds mesh-owned typed `items` as the preferred stable consumer surface.
- Bounds collection at multiple layers: response byte limits for llama.cpp endpoints, sample and label limits for metrics, slot entry limits, per-slot JSON limits, and a capped dashboard slot preview.

## Compatibility

- API changes are additive: existing `/api/status` consumers keep working, and `/api/runtime/llama` is a new local management endpoint.
- The new UI fields include fallbacks for older payloads so mixed local development states remain readable.
- No mesh gossip, wire protocol, routing, election, or model capability semantics were changed.

## Validation

- `cargo test -p mesh-llm runtime_data_llama`
- `cargo test -p mesh-llm runtime_data_api_routes_remain_payload_stable`
- `cargo check -p mesh-llm`
- UI validation passed: typecheck, targeted runtime UI tests, full UI test suite, and production build.


michaelneale:
main oddness I saw " The 62 const _: assertions in mod.rs are unusual. They statically verify that specific methods exist with expected signatures at compile time — a form of
 compile-time interface documentation / regression guardrail. This is not idiomatic Rust (traits or #[test] are the usual approach), and it's 40% of
 mod.rs by line count. " - deliberate? 
 
 otherwise I think looks good?

michaelneale (approved):

PR #387, opened 2026-04-25
Build llama.cpp from an upstream patch queue

i386:
## Summary

mesh-llm can now build its patched llama.cpp from an upstream ggml-org/llama.cpp pin plus a local patch queue, instead of depending on the Mesh-LLM llama.cpp fork as the build source.

This preserves the current llama behavior for this pass while making the update workflow explicit and preparing the tree for later llama-stage-runtime integration. The ABI/static in-process llama work is intentionally not included here.

## Architecture

- Adds `scripts/prepare-llama.sh` to clone/fetch upstream llama.cpp, check out the pinned upstream SHA, and apply `third_party/llama.cpp/patches/*.patch` with `git am --3way`.
- Adds `third_party/llama.cpp/upstream.txt` as the upstream pin and keeps `LLAMA_CPP_SHA` as a compatibility mirror during the transition.
- Updates local build scripts, Dockerfiles, CI workflows, release packaging, and `Justfile` recipes to use `.deps/llama.cpp` prepared from the patch queue.
- Keeps a best-effort `llama.cpp` compatibility symlink so existing CI/script paths continue to work while the default build directory moves under `.deps/`.
- Documents the patch queue workflow and the later llama-stage integration path.

## Compatibility

The generated patched llama.cpp tree matches the previous pinned Mesh-LLM fork tree exactly for this pass:

```text
9364379d399417ed0b32a55717a739313279f70a
```

No mesh wire protocol changes are included.

## Validation

- `scripts/prepare-llama.sh pinned`
- Compared prepared llama.cpp tree hash against the previous fork pin tree hash
- `just build`

Note: `git diff --cached --check` reports trailing whitespace inside the imported `.patch` files. Those files are preserved as generated from the existing fork so the patch queue reproduces the current tree exactly.


michaelneale:
this seemed to work locally when I tried it - nice one. I guess this is just one first step?

i386:
@michaelneale yeah first step. I'll drop a nightly workflow change that will update the SHA via PR

i386:
@ndizazzo would be good to get your eyes here to ensure I haven't stuffed up CI etc

ndizazzo:
Aw I was just testing this locally lol

ndizazzo (commented):
My local testing / workflow for this worked, retroactive "ship it"

PR #388, opened 2026-04-25
feature(console-output): completely revise mesh-llm console output for JSONL output + pretty printing

ndizazzo:
## Summary

This PR makes the `mesh-llm` terminal experience nicer, while keeping automation-friendly output stable. Users running interactively now get a live TUI-style dashboard for startup, model loading, mesh events, local processes, endpoints, and readiness; scripts can use structured JSONL output without scraping human-readable logs.

## Screenshot
image

## What changed

- Added a typed `OutputEvent` pipeline that drives both pretty terminal rendering and machine-readable JSONL output.
- Added an interactive dashboard mode for real TTY sessions, with fallback line rendering for non-interactive shells and CI.
- Reworked startup/runtime messages for model downloads, llama/rpc launch state, MoE placement, peer discovery, passive/client mode, web/API readiness, warnings, errors, and shutdown.
- Added dashboard snapshot plumbing so the terminal can show live local process, model inventory, plugin, request, and endpoint state.
- Added keyboard handling for interactive sessions, including `h` for help, `i` for an info snapshot, and `q`/Ctrl-C for clean shutdown.
- Documented the terminal output event taxonomy in `mesh-llm/src/cli/output/EVENTS.md` so future output changes preserve the pretty/JSON contract.

## User impact

- Interactive users can see what the node is doing without digging through raw logs.
- Long-running startup flows, downloads, MoE analysis, host election, and runtime model load/unload are easier to follow.
- Headless users and automation get newline-delimited JSON on stdout, while pretty output and tracing stay on stderr.
- Non-TTY environments continue to use safe fallback output instead of attempting to enter the dashboard.

## Validation

- Added coverage for JSON envelope/schema fields across output variants.
- Added coverage for pretty dashboard rendering, TUI fallback behavior, resize/focus handling, scrolling, terminal cleanup, and dashboard snapshot registration.
- Added runtime/dashboard tests for cached inventory snapshots and loaded model metadata display.

## Notes
#386 adds metrics hooks this feature can be wired up to for context.


michaelneale (approved):
why not

i386:
Just double check it doesn’t stuff up in tmux or screen

mesh-llm/src/cli/commands/discover.rs
```diff
@@ -102,8 +102,36 @@ pub(crate) fn run_stop() -> Result<()> {
...
+            killed += 1;
+        }
+        if owner_was_live {
+            live_owner_runtime_dirs.insert(target.runtime_dir.clone());
+        }
+    }
+
+    for target in targets.into_iter().filter(|target| !target.is_owner) {
+        if live_owner_runtime_dirs.contains(&target.runtime_dir) {
+            continue;
```
i386: [P1] Stop can leave child servers alive

This marks a runtime dir as protected whenever the owner was live before termination, then skips all child targets in that dir. `terminate_process_blocking` may escalate from SIGTERM to SIGKILL and still return true, so a wedged mesh-llm process that has to be killed will leave its llama-server/rpc-server children running. That regresses `mesh-llm stop` from the previous scoped cleanup behavior.

mesh-llm/src/runtime/mod.rs
```diff
@@ -2084,22 +2844,31 @@ async fn run_auto(
...
+                if interactive_enabled
+                    && llama_ready
+                    && !interactive_started.swap(true, Ordering::SeqCst)
+                {
+                    if let Some(cs) = interactive_console_state {
+                        interactive::spawn_handler(
+                            interactive_control_tx,
+                            cs,
+                            crate::cli::output::OutputManager::global(),
+                            InitialPromptMode::Deferred,
```
i386: [P2] Fallback input handler is never spawned

`interactive_enabled` is true only for `InteractiveDashboard`, and the call to `spawn_handler` is guarded by it. That means the fallback line handler path in `runtime/interactive.rs` is effectively unreachable for serve/client fallback sessions, so the documented `h`, `i`, and `q` commands do not work in line-oriented pretty mode.

mesh-llm/src/cli/mod.rs
```diff
@@ -232,6 +240,10 @@ pub(crate) struct Cli {
     #[command(subcommand)]
     pub(crate) command: Option<Command>,
 
+    /// Terminal output format for app-owned runtime events.
+    #[arg(long, value_enum, default_value_t = LogFormat::Pretty)]
+    pub(crate) log_format: LogFormat,
```
i386, line 245: [P2] Global log-format fails before serve

The new `--log-format` option is a top-level CLI option, but the serve/client normalizer in `normalize_runtime_surface_args` only recognizes pseudo-subcommands at `argv[1]`. I verified `mesh-llm --log-format json serve --model ...` exits with `unrecognized subcommand 'serve'`, while `mesh-llm serve --log-format json ...` works. Either make this option truly global through the normalizer or document/enforce the supported position.

mesh-llm/src/runtime/mod.rs
ndizazzo: addressed!

mesh-llm/src/cli/commands/discover.rs
ndizazzo: addressed!

mesh-llm/src/cli/mod.rs
ndizazzo: addressed!

PR #390, opened 2026-04-25
Add llama upstream pin canary

i386:
Users and maintainers now get an automated canary for moving the pinned llama.cpp upstream SHA. The workflow checks current upstream, validates the Mesh-LLM patch queue against it, runs CI-like Linux validation plus inference smokes, and only then commits the validated pin update directly to `main`.

Manual runs can optionally provide a specific llama.cpp upstream SHA, which makes it easier to validate a known candidate commit instead of always using `origin/master`.

## Job Flow

```mermaid
flowchart TD
    A[Scheduled trigger or manual trigger with optional SHA] --> B[latest_upstream]
    B --> C{New upstream SHA?}
    C -- No --> D[Write no-change summary]
    C -- Yes --> E[linux_build]
    E --> F[inference_smoke_tests on explicit runner]
    F --> G[update_main_pin]
    G --> H[Checkout latest origin/main]
    H --> I[Commit updated pin files]
    I --> J[Push HEAD to main]
```

## What Changed

- Added `.github/workflows/llama-upstream-canary.yml` with separate jobs for upstream detection, Linux build validation, reusable inference smokes, and direct main update.
- Manual workflow dispatch now accepts an optional `upstream_sha` input for validating a specific llama.cpp commit.
- The canary passes an explicit `runs_on` value to the reusable smoke workflow, and `smoke.yml` now supports that optional input with a default of `ubuntu-latest` for existing callers.
- The pin-update job fetches latest `origin/main`, commits `third_party/llama.cpp/upstream.txt` plus the `LLAMA_CPP_SHA` compatibility mirror, and pushes `HEAD:main` after validation passes.
- Added `just` helpers for preparing latest upstream, updating both pin files, and rendering a pin-change summary.
- Added scripts to update both tracked pin files and to generate a workflow summary with upstream commit/diff context.
- Adjusted `scripts/prepare-llama.sh` so patch application has the blob data needed for `git am --3way`, including compatibility for existing partial clones.

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/llama-upstream-canary.yml"); YAML.load_file(".github/workflows/smoke.yml")'`
- `bash -n scripts/prepare-llama.sh scripts/update-llama-pin.sh scripts/summarize-llama-upstream.sh`
- `just --list | rg 'llama-(prepare|summary|update)'`
- `git diff --check`
- `just llama-summary <current-pin> <current-pin>`

Note: `just llama-prepare-latest` now reaches real patch application, but current llama.cpp `master` conflicts in patch `0001` (`ggml-rpc.cpp`). That means the new canary would currently fail at the compatibility check until the patch queue is refreshed, which is the intended signal.

Direct pushes from Actions still depend on repository branch-protection settings allowing `github-actions[bot]` to update `main`.


i386:
@ndizazzo

.github/workflows/llama-upstream-canary.yml
```diff
@@ -0,0 +1,234 @@
+name: Llama Upstream Canary
+
+on:
+  schedule:
+    - cron: "17 3 * * *"
+  workflow_dispatch:
```
ndizazzo, line 6: NIT: Maybe a SHA here will be useful

.github/workflows/llama-upstream-canary.yml
```diff
@@ -0,0 +1,234 @@
...
+            llama.cpp/build/bin/llama-server
+            llama.cpp/build/bin/llama-moe-analyze
+            llama.cpp/build/bin/llama-moe-split
+          if-no-files-found: error
+
+      - name: Show sccache stats
+        if: always()
+        run: sccache --show-stats || true
+
+  inference_smoke_tests:
```
ndizazzo, line 174: missing explicit `runs_on`

.github/workflows/llama-upstream-canary.yml
```diff
@@ -0,0 +1,234 @@
...
+      - name: Check embedded client dependency purity
+        run: |
+          cargo tree -p mesh-llm-client --prefix=none --no-dedupe > /tmp/mesh-client-deps.txt
+          FORBIDDEN="keyring|rpassword|hf-hub|dirs|clap|include_dir|rmcp|keyring-core"
+          if grep -E "^[[:space:]│├└─-]*($FORBIDDEN)([[:space:]]|$)" /tmp/mesh-client-deps.txt; then
+            echo "ERROR: Forbidden dependency found in mesh-client"
+            grep -E "^[[:space:]│├└─-]*($FORBIDDEN)([[:space:]]|$)" /tmp/mesh-client-deps.txt
+            exit 1
+          fi
+          echo "PASS: No forbidden dependencies found"
```
ndizazzo, line 115: Just an idea, if we find that this is not fast enough, we could create a Docker image with pre-packaged dependencies in the [mesh-llm-container](https://github.com/Mesh-LLM/mesh-llm-runner-images) repo and run the build in docker via DinD on our Linux builder.

ndizazzo (approved):
I don't see anything prickly here, I have some minor notes tho

PR #392, opened 2026-04-25
Remove MoE split serving

i386:
Users no longer see or rely on the `mesh-llm moe` command family or MoE split-serving behavior.

## Summary
- Removed the `mesh-llm moe` CLI, MoE planning/share/analyze command handlers, and MoE split runtime orchestration.
- Removed MoE split target routing (`MoeLocal`/`MoeRemote`), split faildown state, split smoke scripts, and split-specific docs.
- Removed llama.cpp `moe-split` patch queue entries and CI/package references to `llama-moe-split`.
- Kept the `llama-moe-analyze` patch and updated CI/warm-cache targets to keep building it.

## Protocol
MoE model metadata and protobuf fields remain additive for mixed-version compatibility. This removes split-serving behavior without repurposing existing wire fields.

## Validation
- `cargo check -p mesh-llm`
- `cargo check -p mesh-llm-client`
- `cargo test -p mesh-llm --lib` (`982 passed; 0 failed; 3 ignored`)
- `scripts/prepare-llama.sh pinned` in a temporary llama.cpp workdir


ndizazzo:
@i386 OOTL here

```
Users no longer see or rely on the mesh-llm moe command family or MoE split-serving behavior.

Removed the mesh-llm moe CLI, MoE planning/share/analyze command handlers, and MoE split runtime orchestration.
Removed MoE split target routing (MoeLocal/MoeRemote), split faildown state, split smoke scripts, and split-specific docs.
Removed llama.cpp moe-split patch queue entries and CI/package references to llama-moe-split.
Kept the llama-moe-analyze patch and updated CI/warm-cache targets to keep building it.
```

How come? Because of some of the new automated stuff that's on the way?

i386:
@ndizazzo the way we are doing MoE splits was a bit naive and made the model dumber. There might still be a path forward to doe MoE sharding, but this one is a bit of a dead end :(

PR #395, opened 2026-04-26
openai-endpoint plugin: vLLM, TGI, Ollama passthrough

michaelneale:
Route inference to any OpenAI-compatible server. Same plugin pattern as lemonade — 53 lines of plugin code, env var for URL, enable in config.toml.

## Usage

```toml
# ~/.mesh-llm/config.toml
[[plugin]]
name = "openai-endpoint"
```

```bash
MESH_LLM_OPENAI_ENDPOINT_URL=http://gpu-box:8000/v1 mesh-llm serve
```

Works with vLLM, TGI, Ollama, or anything that speaks `/v1/chat/completions`. Default URL: `http://localhost:8000/v1`.

The server does all the inference. mesh-llm discovers models via `GET /v1/models`, health-checks periodically, and routes requests.

## Changes

- `plugins/openai_endpoint/mod.rs`: the plugin (53 lines, same shape as lemonade)
- `plugin/config.rs`: config resolver + spec (same pattern as lemonade/blackboard/blobstore)
- `plugin/mod.rs`: constant + dispatch
- `plugins/mod.rs`: module registration
- `cli/mod.rs`: external backend hint in `--help` output
- `README.md`: section with config + env var examples for vLLM/Ollama/TGI
- `docs/index.html`: feature card


docs/index.html
```diff
@@ -186,6 +186,11 @@ <h3>Inter-model collaboration</h3>
                 <h3>Works with agents</h3>
                 <p>OpenAI-compatible API on <code>localhost:9337</code>. Use with goose, pi, opencode, or any tool that supports custom OpenAI endpoints.</p>
             </div>
+            <div class="feature">
+                <div class="feature-icon">🔗</div>
+                <h3>External backends</h3>
+                <p>Already running vLLM, TGI, or Ollama? Enable the <code>openai-endpoint</code> plugin and set <code>MESH_LLM_OPENAI_ENDPOINT_URL</code>. mesh-llm discovers the models and routes requests — the backend does all the inference work.</p>
```
ndizazzo: It would be really nice to configure this in the plugin definition instead of using an environment variable, which might need:

host
port
api key/token

i386:
@michaelneale can we delete the lemonade plugin while we are here?

i386:
This should also have support for config file not just environment variables

ndizazzo (commented):
Looks good! I think one way we can strengthen is by moving the value definition into the plugin

i386:
Is it possible to run this plugin and not have to load llama?

michaelneale:
yeah will clean it up. My agent didn't want to so it was sent to the naughty corner.

docs/index.html
michaelneale: Done — url is now in the config file:

```toml
[[plugin]]
name = "openai-endpoint"
```

API key/token: good call, not needed for local vLLM/Ollama/TGI but would matter for authenticated endpoints. Will add as a follow-up.

ndizazzo (approved):
Nice one! No more comments that Copilot didn't already catch

PR #396, opened 2026-04-26
WIP: hardware attestation and E2E inference encryption for Apple Silicon

michaelneale:
End-to-end encryption of inference prompts between mesh peers, with hardware attestation for Apple Silicon nodes via Secure Enclave.

- Prompts encrypted (NaCl box) before going over QUIC tunnels — relay nodes see only ciphertext
- Runtime hardening at startup (anti-debug, SIP/RDMA checks, binary integrity)
- P-256 Secure Enclave attestation framework with peer-to-peer verification
- `--require-attested-hosts` flag for client-local trust policy
- Backward compatible — old nodes unaffected, new fields are optional in protobuf gossip

Design doc: LLAMA_SEC.md


mesh-llm/src/network/openai/transport.rs
```diff
@@ -1593,6 +1593,61 @@ async fn route_local_attempt(
...
+            "Invalid inference key for host {}, sending plaintext",
+            host_id.fmt_short()
+        );
+        return prefetched.to_vec();
+    };
+    // Encrypt the entire HTTP payload
+    match crate::crypto::inference_encryption::encrypt_inference_request(prefetched, "", &host_pub)
+    {
+        Ok((payload, _session)) => {
+            // Wire format: MAGIC byte + JSON-serialized EncryptedInferencePayload
```
ndizazzo, line 1660: The ephemeral session is silently dropped here. `encrypt_response_chunk` and `decrypt_response_chunk` already exist in `inference_encryption.rs` but are never called — so responses travel back in plaintext. The E2E in "E2E encryption" is incomplete. Either wire the response decryption path or document explicitly that response encryption is deferred to a follow-up.

mesh-llm/src/network/openai/transport.rs
```diff
@@ -1593,6 +1593,61 @@ async fn route_local_attempt(
...
+            tracing::debug!(
+                "Encrypted inference request for host {}",
+                host_id.fmt_short()
+            );
+            out
+        }
+        Err(err) => {
+            tracing::warn!(
+                "Encryption failed for host {}: {err}, sending plaintext",
+                host_id.fmt_short()
```
ndizazzo, line 1687: When `require_attested_hosts` is set and encryption fails, this silently downgrades to sending plaintext — violating the security policy the user explicitly configured. This should be a hard error that refuses the request, not a warn-and-proceed fallback.

mesh-llm/src/network/openai/transport.rs
```diff
@@ -1593,6 +1593,61 @@ async fn route_local_attempt(
...
+/// Encrypt prefetched HTTP request bytes for a remote host if they have an inference key.
+/// Falls back to plaintext if no key is available or encryption fails.
+async fn maybe_encrypt_for_host(
+    node: &mesh::Node,
+    host_id: iroh::EndpointId,
+    prefetched: &[u8],
+) -> Vec<u8> {
+    // Gate: if --require-attested-hosts, refuse to route to peers without an inference key
+    if node.require_attested_hosts {
+        if node.peer_inference_public_key(host_id).await.is_none() {
```
ndizazzo: This gates on whether the peer has an `inference_public_key` in gossip, but every node running this version auto-generates one at startup. There's no attestation verified here. The flag name (`--require-attested-hosts`) strongly implies SE attestation is checked; either call `verify_attestation` here or rename the flag to something accurate like `--require-encrypted-transport`.

mesh-llm/src/network/openai/transport.rs
```diff
@@ -1593,6 +1593,61 @@ async fn route_local_attempt(
...
+    let Ok(host_pub) = crate::crypto::inference_encryption::parse_public_key(&host_key_b64) else {
+        tracing::warn!(
+            "Invalid inference key for host {}, sending plaintext",
+            host_id.fmt_short()
+        );
+        return prefetched.to_vec();
+    };
+    // Encrypt the entire HTTP payload
+    match crate::crypto::inference_encryption::encrypt_inference_request(prefetched, "", &host_pub)
+    {
```
ndizazzo, line 1658: `encrypt_inference_request` is called with an empty string for the model field. The model is available from the request at this call site — plumb it through so the encrypted payload carries accurate metadata.

mesh-llm/src/network/tunnel.rs
```diff
@@ -320,20 +320,69 @@ async fn handle_inbound_stream(
     Ok(())
 }
 
+/// Magic byte prefix for E2E encrypted tunnel payloads.
+/// Must match the constant in `network::openai::transport`.
+const ENCRYPTED_TUNNEL_MAGIC: u8 = 0xE1;
```
ndizazzo: `ENCRYPTED_TUNNEL_MAGIC = 0xE1` is also defined in `transport.rs`. If either side drifts the two will silently misparse each other's frames. Move this into `crypto/inference_encryption.rs` (or `crypto/mod.rs`) and import it from both sites.

mesh-llm/src/network/tunnel.rs
```diff
@@ -320,20 +320,69 @@ async fn handle_inbound_stream(
...
-    let (tcp_read, tcp_write) = tokio::io::split(tcp_stream);
-    relay_bidirectional(tcp_read, tcp_write, quic_send, quic_recv).await
+    if first[0] == ENCRYPTED_TUNNEL_MAGIC {
+        // Encrypted path: read the rest of the request-direction data, decrypt, forward.
+        let encrypted_json = quic_recv
+            .read_to_end(16 * 1024 * 1024)
+            .await
+            .map_err(|e| anyhow::anyhow!("failed to read encrypted tunnel payload: {e}"))?;
+
+        let payload: crate::crypto::inference_encryption::EncryptedInferencePayload =
```
ndizazzo, line 354: 16 MB is already tight for multimodal requests — images alone routinely exceed this. Either raise the cap to something like 256 MB, or stream the decryption rather than buffering the full payload into memory.

mesh-llm/src/system/hardening.rs
```diff
@@ -0,0 +1,376 @@
...
+
+/// Remove environment variables that could be used for code injection.
+fn scrub_dangerous_env() {
+    for var in DANGEROUS_ENV_VARS {
+        if std::env::var_os(var).is_some() {
+            tracing::warn!("security: scrubbing dangerous env var: {var}");
+            // SAFETY: single-threaded at startup, before any threads spawned.
+            unsafe {
+                std::env::remove_var(var);
+            }
```
ndizazzo, line 194: `std::env::remove_var` is explicitly documented as UB when called in a multi-threaded context, and it's stabilized as `unsafe` for exactly this reason. By the time `harden_runtime` runs from `runtime/mod.rs`, the Tokio runtime is already live with worker threads. Move this call to the very top of `main`, before the Tokio runtime initializes, to be safe.

mesh-llm/src/mesh/gossip.rs
```diff
@@ -99,6 +99,7 @@ pub(super) fn apply_transitive_ann(
     existing.available_models.clear();
     existing.requested_models = ann.requested_models.clone();
     existing.owner_attestation = ann.owner_attestation.clone();
+    existing.inference_public_key = ann.inference_public_key.clone();
     if ann.model_source.is_some() {
```
ndizazzo, line 105: `inference_public_key` is gossiped as a bare string with no signature binding it to the peer's identity. Any peer can advertise an arbitrary key, including impersonating another node. Either sign this field with the node's existing Ed25519 identity key, or only accept it alongside a valid `SignedHardwareAttestation` that ties the inference key to the peer's endpoint ID.

mesh-llm/src/crypto/se_attestation.rs
```diff
@@ -0,0 +1,610 @@
...
+}
+
+/// Sign an attestation blob using a **software** P-256 signing key.
+///
+/// **Testing / development only.** In production on macOS the Secure
+/// Enclave hardware key should be used instead. This helper exists so
+/// that tests and non-macOS environments can exercise the attestation
+/// verification path without SE hardware.
+pub fn sign_attestation_with_software_key(
+    attestation: &HardwareAttestation,
```
ndizazzo, line 437: Both `sign_attestation_with_software_key` and `sign_challenge_with_software_key` are software-only stubs — there's no `Security.framework` or `CryptoKit` call anywhere. The flag name and design doc both promise real SE-backed signing, but there's no path from a running node to an actual Secure Enclave key. What's the plan for wiring this up before the flag ships as non-WIP?

PR #403, opened 2026-04-28
Expose ranked model targets in the API

### LINKED ISSUE #276: Model Targets / Wanted Models Specification
# Model Targets / Wanted Models Specification

## Summary

Add a phased, additive "model interest" feature that lets a user pick a model from the existing search surface, submit explicit interest for that model through the connected host, and have the mesh maintain a collective ranked view of desired models.

The mesh should then:

- expose that ranked wanted state to the UI and API,
- evaluate whether the current mesh has enough capacity to load the model,
- decide how many replicas should exist,
- identify under-utilized models that are eligible for unloading to make room,
- and, when capacity is still insufficient, keep the model on the wanted list with a clear add-capacity hint instead of silently dropping it.

This must ship in small, focused PRs rather than as one monolithic feature.

## Goal

- **Primary outcome**
  - Users can search for a model, register explicit interest in it, and see the mesh treat that model as a collective desired target.
  - The mesh can rank desired models using both explicit interest and observed request demand.
  - The system can evaluate capacity and recommend or perform loading work through existing runtime control surfaces.
  - When the mesh cannot make room, the desired model remains visible as wanted with an explanation and a capacity hint.

- **Secondary outcome**
  - The web console gains a first-class search-and-interest workflow backed by a new `/api/search` endpoint.

## Non-goals

- Replacing the existing `ModelDemand` or `requested_models` semantics.
- Introducing a breaking mesh protocol change.
- Building a full mesh-global persisted desired-state system in the first rollout.
- Enabling aggressive automatic eviction before ranking and capacity evaluation are observable and stable.
- Overloading `/api/models` to serve search queries.

## Context

- The current CLI search surface is `mesh-llm models search` and is implemented in `mesh-llm/src/cli/models.rs` and `mesh-llm/src/models/search.rs`.
- The current management API exposes `GET /api/models`, `GET /api/status`, `GET /api/runtime*`, and runtime load/unload controls, but it does not expose `/api/search`.
- The mesh already tracks:
  - `requested_models` as declared load intent,
  - `ModelDemand` as observed traffic,
  - `active_demand()` as the current active desired-model set,
  - and Nostr `wanted` as a summary of active demand that is not currently served.
- The runtime already performs demand-aware model promotion and uses existing load/unload boundaries.

This spec extends those mechanisms. It does not replace them.

## Problem

Today the mesh has good machinery for reacting to real traffic and declared startup models, but it lacks a first-class user-facing way to say:

- "this model should be considered a desired target for the mesh," even before it has live traffic,
- "show me which desired models the mesh wants most right now,"
- and "tell me why the mesh did or did not make room for that model."

That leaves four gaps:

1. There is no UI/API workflow for searching and selecting a model target from the host the user is connected to.
2. There is no explicit mesh-wide advisory signal for user-submitted interest separate from request traffic.
3. There is no collective ranking surface that combines explicit interest with request demand.
4. There is no advisory capacity evaluator that can explain target replica count, unload candidates, and insufficient-capacity outcomes.

### Search

- `mesh-llm models search` is the existing product surface.
- `mesh-llm/src/models/search.rs` already defines search types and the canonical search behavior.
- Search hits already carry an `exact_ref`, which should become the canonical model identifier for explicit interest.

### Demand and desired state

- `mesh-llm/src/mesh/mod.rs` stores `requested_models` and `ModelDemand`.
- `record_request()` increments demand from routed inference traffic.
- `active_demand()` keeps models active when they are within TTL or pinned by a node's `requested_models`.
- `mesh-llm/src/network/nostr.rs` already computes a `wanted` list from active demand that is not currently served.

### Runtime control

- `mesh-llm/src/runtime/mod.rs` already uses active demand to choose assignments.
- `mesh-llm/src/api/routes/runtime.rs` already exposes:
  - `POST /api/runtime/models`
  - `DELETE /api/runtime/models/:model`
- `mesh-llm/src/runtime/local.rs` enforces that runtime load only works for models that fit locally.

### Status and UI payloads

- `mesh-llm/src/api/status.rs` already carries model-level fields such as `request_count`, `last_active_secs_ago`, `fit_label`, and `fit_detail`.
- The spec should extend existing status/model surfaces before inventing parallel payloads.

## Terminology

- **Search result** — a model candidate returned by the existing search logic.
- **Canonical model reference** — the exact search-selected identifier used for interest submission. This should come from the search result's `exact_ref` semantics.
- **Explicit interest** — an advisory, user-submitted signal that a model should be considered a desired target.
- **Request demand** — observed runtime traffic represented by `ModelDemand`.
- **Wanted model** — a model that the mesh collectively wants but is not currently served at the desired level.
- **Ranked target** — a model after explicit interest and request demand have been combined into a single ranking view.
- **Desired replica count** — how many live serving copies the mesh should ideally maintain for the model.
- **Unload candidate** — a currently served model that is eligible to be removed to free capacity.
- **Deferred wanted model** — a wanted model that cannot currently be activated because the mesh lacks usable capacity.
- **Capacity hint** — user-facing guidance that explains what additional capacity would likely make activation possible.

### 1. Canonical model selection

Explicit interest must be anchored in the same model identity that the existing search flow returns.

### Required behavior

- UI search must call a new `GET /api/search` endpoint.
- `/api/search` must wrap the same search behavior used by `mesh-llm models search`.
- The interest submission flow must use a canonical model reference from the search response rather than accepting arbitrary free-form strings.

### Contract direction

- Use the search result's `exact_ref` as the canonical interest key.
- Also return display metadata useful for the UI, such as:
  - display name
  - repo id
  - artifact kind
  - size label
  - capabilities

If the product later wants `mesh-llm model search` as a singular alias, that should be treated as a CLI UX decision layered over the same search contract. This spec is grounded in the current implementation surface, which is `mesh-llm models search`.

### 2. Explicit interest is a new advisory signal

Explicit interest must not overwrite or redefine `ModelDemand`.

### Why

- `ModelDemand` means observed traffic.
- `requested_models` means node-declared requested load intent.
- User-submitted interest is different: it expresses desire, not proof of traffic.

### Required behavior

- Add a new API surface for submitting and clearing explicit interest.
- Keep explicit interest separate from request demand in storage and propagation.
- Combine them only when computing ranked targets.

### Proposed API surface

- `POST /api/model-interests`
- `DELETE /api/model-interests/:model_ref`
- `GET /api/model-targets`

Example request body:

```json
{
  "model_ref": "Qwen/Qwen3-Coder-Next-GGUF:Q4_K_M",
  "source": "ui"
}
```

`GET /api/model-targets` should return the mesh-wide ranked and explained target list, including active, wanted, and deferred entries.

### 3. Mesh-wide interest propagation must be additive

This is the most compatibility-sensitive part of the rollout.

### Requirement

Do not change the meaning of existing fields such as:

- `requested_models`
- `demand`
- Nostr `wanted`

### Spec direction

- If existing fields cannot safely carry explicit-interest semantics without confusing older peers, add a new optional advisory field.
- That new field must be additive and safe for older nodes to ignore.

### Recommended shape

Add a new optional per-node interest payload in gossip, conceptually like:

```proto
message ModelInterestEntry {
  string model_ref = 1;
  uint32 explicit_interest_count = 2;
  uint64 last_updated = 3;
}
```

This is intentionally advisory. It does not replace `ModelDemandEntry`.

### 4. Ranking must combine interest and demand without hiding either

The system should compute a ranked target list from multiple signals.

### Inputs

- explicit interest count
- number of distinct nodes expressing explicit interest
- request demand recency
- request demand volume
- current serving count
- current desired replica count
- local/mesh fit and availability

### Principles

- Request demand is the stronger signal for immediate promotion and replica expansion.
- Explicit interest can surface a model into the wanted list before it has traffic.
- Explicit interest alone should not immediately force large-scale replication.
- Ranking output must retain component scores so the UI can explain why a model is ranked highly.

### Example ranking outputs per target

- `interest_score`
- `demand_score`
- `combined_rank`
- `current_replicas`
- `desired_replicas`
- `status: serving | wanted | deferred`
- `explanation`

### 5. Capacity evaluation is advisory before it is executable

The mesh must evaluate what it should do before it starts automatically unloading or loading models.

### Inputs

- ranked targets
- mesh-wide serving state
- local and mesh VRAM availability
- fit data and model size
- on-disk availability
- downloadable availability via current catalog resolution
- recent demand activity
- cooldown or anti-flap windows

### Outputs

- desired replica count
- activation recommendation
- unload candidate list
- deferred reason when activation is not possible
- add-capacity hint

### Unload-candidate rules

A model is only an unload candidate if all of the following are true:

- it is below its desired replica count or otherwise over-provisioned,
- it has weak recent demand,
- it has no strong explicit interest keeping it wanted,
- it is outside the cooldown window,
- and unloading it would improve capacity for a higher-ranked target.

### Desired-replica rules

- Explicit interest alone can justify a desired replica count of `1`.
- Additional replicas should require meaningful request demand or clear under-provisioning.
- The first phase of this evaluator should produce decisions and explanations only.

### Deferred behavior

If the evaluator cannot find enough capacity, the target must remain in the wanted list with fields such as:

- `deferred_reason`
- `missing_capacity_gb`
- `suggested_added_nodes`
- `capacity_hint`

### 6. Runtime execution should reuse existing boundaries

When the system eventually becomes executable, it should not invent a new activation control plane.

### Requirement

Use existing runtime boundaries:

- `POST /api/runtime/models`
- `DELETE /api/runtime/models/:model`

### Rollout rule

- Early phases compute advisory decisions only.
- Later phases may reconcile the advisory plan into runtime actions.
- Automatic unload must land after ranking and capacity evaluation are proven stable in mixed-version and multi-node tests.

## UI component

The web console should gain a dedicated wanted-models component, but that work should land only after the backend surfaces, ranking, capacity evaluation, and runtime reconciliation boundaries are stable.

The UI is intentionally the final phase of this rollout. Foundational API, protocol, ranking, and runtime work must land first so the UI binds to stable contracts rather than dragging foundational decisions into frontend work.

### Required UI behavior

- Search box backed by `GET /api/search`.
- Search results show the canonical model reference and summary metadata.
- Each result has a clear **Register interest** action.
- The wanted-models view shows:
  - current rank
  - current serving status
  - desired replica count
  - current replica count
  - deferred reason, if any
  - add-capacity hint, if any
  - the contributing signals at a high level, such as interest vs traffic

### UI placement direction

- Land this UI only in the final rollout phase.
- Add a new wanted-models panel to the dashboard/management experience once the backend contracts are stable.
- Do not hide this behind chat-only model selection.
- Search and wanted-state belong in the management surface because the feature is mesh control, not request composition.

### CLI

- Existing search surface remains the source of truth: `mesh-llm models search`.
- The spec does not require a new CLI mutation command in phase 1.

### Management API

- `GET /api/search`
- `POST /api/model-interests`
- `DELETE /api/model-interests/:model_ref`
- `GET /api/model-targets`
- additive extensions to `GET /api/status` and/or `GET /api/models`

### Mesh protocol

- add an optional advisory explicit-interest field only if existing fields cannot safely express this state
- keep changes additive and mixed-version-safe

### Nostr discovery

- existing `wanted` list may continue to reflect the derived wanted state
- this spec does not require a Nostr schema change in the initial rollout

### `GET /api/search`

Purpose: expose local host search to the UI using the same semantics as the CLI search path.

Query parameters:

- `q` — required search text
- `artifact=gguf|mlx` — optional
- `catalog=true|false` — optional
- `limit=<n>` — optional

Response shape should mirror the CLI JSON search result closely enough that the UI and CLI describe the same model identity.

Example response:

```json
{
  "results": [
    {
      "model_ref": "Qwen/Qwen3-Coder-Next-GGUF:Q4_K_M",
      "repo_id": "Qwen/Qwen3-Coder-Next-GGUF",
      "kind": "GGUF",
      "size_label": "8.0 GB",
      "capabilities": {
        "tool_use": true,
        "reasoning": true
      }
    }
  ]
}
```

### `POST /api/model-interests`

Purpose: register explicit user interest for a canonical model reference.

Rules:

- `model_ref` must come from the `/api/search` result contract.
- duplicate submissions from the same node should be idempotent at the entry level unless the implementation explicitly supports weighted repeated interest.
- the API should return the updated local and mesh-visible state for that model.

### `GET /api/model-targets`

Purpose: expose the derived, explained ranked target list.

Recommended response fields:

- `model_ref`
- `display_name`
- `interest_score`
- `demand_score`
- `combined_rank`
- `current_replicas`
- `desired_replicas`
- `status`
- `unload_candidates`
- `deferred_reason`
- `capacity_hint`

### `GET /api/status` / `GET /api/models`

These surfaces should gain additive fields rather than be replaced.

Recommended additions for model payloads:

- `explicit_interest_count`
- `interest_node_count`
- `combined_rank`
- `desired_replicas`
- `current_replicas`
- `wanted_status`
- `deferred_reason`
- `capacity_hint`

## Compatibility

- Mixed-version mesh compatibility is mandatory.
- Existing peers must continue to ignore unknown additive fields.
- Existing `requested_models`, `demand`, and `wanted` semantics must remain valid.
- If explicit interest cannot be represented without semantic drift, it must use a new optional field rather than redefining an old one.
- Automatic execution phases must be gated by validation against older peers.

## Constraints and tradeoffs

- **Constraint**: current runtime load only supports models that fit locally on the executing node.
- **Constraint**: current request demand already influences promotion and assignment, so explicit interest must not destabilize that path.
- **Tradeoff**: keeping explicit interest advisory first slows down full automation, but it greatly reduces churn and protocol risk.
- **Alternative rejected**: inventing a completely separate model-target subsystem disconnected from demand, requested models, and existing runtime control.

### Phase 0: spec-only

Deliverables:

- `docs/model-targets.md`

Why this stays small:

- aligns terminology and rollout before code lands

### Phase 1: search contract

Deliverables:

- `GET /api/search`
- parity tests against existing search behavior
- search response contract that exposes canonical model references

Why this stays small:

- read-only API only
- no state mutation yet

### Phase 2: local explicit-interest submission

Deliverables:

- `POST /api/model-interests`
- `DELETE /api/model-interests/:model_ref`
- local explicit-interest storage and readback

Why this stays small:

- local-node state only
- no protocol changes yet
- no UI yet

### Phase 3: mesh-wide propagation and ranked target view

Deliverables:

- additive mesh propagation for explicit interest, if needed
- `GET /api/model-targets`
- ranking that combines explicit interest and request demand
- status/model payload extensions for wanted-target visibility

Why this stays small:

- protocol-sensitive work is isolated
- still advisory; no automatic runtime execution yet
- still no UI yet

### Phase 4: advisory capacity evaluator

Deliverables:

- desired replica calculation
- unload candidate calculation
- deferred wanted states
- capacity hints

Why this stays small:

- decision engine only
- no automatic load/unload yet
- still no UI yet

### Phase 5: controlled runtime reconciliation

Deliverables:

- reconcile advisory activation decisions through existing runtime control APIs
- optional automatic load for safe cases
- unload remains guarded and policy-driven

Why this stays small:

- reuses existing execution surfaces
- can be gated behind strict safeguards
- still no UI yet

### Phase 6: UI search, submission, and operator explanation

Deliverables:

- dashboard wanted-models search backed by `GET /api/search`
- register-interest action backed by `POST /api/model-interests`
- wanted-models panel with explanations
- capacity warnings and add-capacity hints
- ranking explanation details

Why this stays small:

- UI-only once backend behavior is stable

## Mermaid component diagram

```mermaid
---
config:
  look: neo
  theme: dark
  layout: dagre
---
flowchart TB
 subgraph P1["Phase 1 - Search Contract"]
        SearchAPI["GET /api/search/"]
        SearchCore["Shared Search Logic\nmodels/search.rs"]
  end
 subgraph P2["Phase 2 - Local Interest"]
        InterestAPI["POST, DELETE /api/model-interests/"]
        LocalInterest["Local Explicit Interest Store"]
  end
 subgraph P3["Phase 3 - Mesh Ranking"]
        Gossip["Additive Interest Propagation"]
        Ranker["Ranked Target Aggregator"]
        TargetsAPI["GET /api/model-targets/"]
  end
 subgraph P4["Phase 4 - Advisory Capacity"]
        Capacity["Capacity Evaluator"]
        Unload["Unload Candidate Evaluator"]
        Deferred["Deferred Wanted + Capacity Hint"]
  end
 subgraph P5["Phase 5 - Runtime Reconciliation"]
        RuntimeCtl["Existing Runtime Load/Unload Control"]
  end
  subgraph P6["Phase 6 - UI Explainability"]
        UI["Dashboard Wanted Models UI"]
        StatusAPI["GET /api/status + /api/models/"]
        Explain["Wanted Rank + Hint Rendering"]
  end
    SearchAPI --> SearchCore
    SearchAPI --> InterestAPI
    InterestAPI --> LocalInterest
    LocalInterest --> Gossip & Ranker
    Gossip --> Ranker
    Ranker --> TargetsAPI & Capacity
    Capacity --> Unload & Deferred & RuntimeCtl
    Unload --> RuntimeCtl
    RuntimeCtl --> StatusAPI
    SearchAPI --> UI
    InterestAPI --> UI
    TargetsAPI --> Explain
    Deferred --> Explain
    StatusAPI --> Explain
    Explain --> UI 
```

## Acceptance criteria

- [ ] Given a user in the web console, when they search for a model, then the UI uses `GET /api/search` and receives results aligned with the current CLI search behavior.
- [ ] Given a search result, when the user registers interest, then the system stores explicit interest using the canonical model reference rather than a free-form string.
- [ ] Given multiple nodes registering interest, when mesh state converges, then the mesh exposes a collective ranked target view.
- [ ] Given request traffic and explicit interest for the same model, when ranking is computed, then both signals are visible and request demand remains the stronger immediate activation signal.
- [ ] Given a wanted model that fits and has room, when the evaluator runs, then it produces a positive activation recommendation and desired replica count.
- [ ] Given a wanted model that does not fit without making room, when the evaluator runs, then it identifies safe unload candidates or returns a deferred wanted state with a capacity hint.
- [ ] Given insufficient capacity and no safe unload candidates, when the evaluator runs, then the model remains on the wanted list and is not silently discarded.
- [ ] Given older peers in the mesh, when explicit-interest support is introduced, then older peers continue to interoperate by ignoring additive fields.
- [ ] Given the advisory phases only, when the evaluator runs, then it does not automatically evict models yet.

### API and contract tests

- `/api/search` parity with CLI search semantics
- canonical model reference round-trip tests
- explicit-interest submission idempotency tests

### Mesh and compatibility tests

- additive protocol compatibility with older peers
- multi-node convergence tests for explicit interest propagation
- ranking determinism tests

### Capacity-evaluator tests

- fits now
- fits after unloading
- does not fit even after unloading
- interest-only target requests one replica
- demand-driven target requests additional replicas
- cooldown prevents unload thrash

### Integration tests

- two-node interest propagation
- ranked target visibility in `/api/model-targets`
- deferred wanted state with capacity hint
- controlled runtime reconciliation through existing runtime APIs

### UI tests

- wanted-model search invokes `/api/search`
- register-interest action updates target state
- deferred capacity hints render correctly
- ranking explanation is visible without exposing raw internal-only fields

## Rollout and migration notes

- Ship phases as separate PRs.
- Keep protocol-sensitive work isolated from UI work.
- Keep executable automation behind later, validation-heavy phases.
- Document any future CLI aliasing or naming cleanup separately from this feature.

## Open questions

- Should explicit interest remain runtime-only in the first rollout, or persist locally across restarts?
- Should repeated interest submissions be weighted, or should interest be idempotent per node/user target?
- Should `GET /api/model-targets` be the sole ranked-target surface, or should `/api/status` and `/api/models` eventually be sufficient for most UI needs?
- Should Nostr listings remain summary-only, or should a future phase advertise richer wanted-model details publicly?
- If the product wants `mesh-llm model search` as a public-facing alias, should that land as a separate CLI compatibility change?

IvGolovach:
## Summary
Expose ranked model targets through the management API so operators can see which model refs the mesh wants and why.

This adds `/api/model-targets` and enriches `/api/status` plus `/api/models` with wanted/ranking metadata derived from explicit interest, active demand, and serving visibility.

## Why
The mesh already tracks demand and explicit model interest, but that signal was not available as a clean operator-facing API view. This makes debugging, scheduling decisions, and follow-up routing work easier without changing serving behavior.

## Scope
- API: `/api/model-targets`, `/api/status`, `/api/models`
- Runtime views: ranked target aggregation from snapshot/gossip/runtime state
- Mesh/protocol: additive `explicit_model_interests` protobuf field 35
- Docs/tests: README, design notes, testing notes, API/model-target/protobuf coverage

Protocol compatibility: the protobuf change is additive. Older peers can ignore the unknown field; peers that do not send it simply do not contribute explicit-interest signal.

## Branch status
Base: `main`

- `origin/main`: `2e465f8c1f6b40cd197db057c09a01970cf868a2`
- Commit: `ffba8d250efd179b9a0228a091035d6a44a6a68a`
- Ahead/behind: `1 / 1`
- Merge-base: `310800ba0e1597ada8a3564236b53a8007fe2bbf`
- Fast-forward safety: not currently fast-forward safe because `origin/main` advanced after validation
- Merge conflict precheck: `git merge-tree --write-tree origin/main HEAD` returned exit `0`

This should be refreshed onto current `main` and revalidated before being marked ready.

## Validation
Run before `origin/main` advanced:

```text
cargo check -p mesh-llm: PASS
cargo test -p mesh-llm model_targets -- --nocapture: PASS, 3 passed
cargo test -p mesh-llm explicit_model_interests -- --nocapture: PASS, 3 passed
cargo test -p mesh-llm test_api_status_and_models_surface_wanted_targets -- --nocapture: PASS, 1 passed
cargo test -p mesh-llm test_proto_round_trip_with_bandwidth_and_tflops -- --nocapture: PASS, 1 passed
cargo test -p mesh-llm -- --skip show_model_variants_accepts_selected_quant_ref: PASS, lib 1223 passed, 3 ignored, 1 filtered out; integration 22 passed
```

`just build`: not run locally because `just` is not installed in this environment.

## Runtime safety
Reviewed API, runtime_data, mesh gossip, protocol conversion, and inference signal plumbing. The change adds aggregation/read-path visibility and additive gossip metadata; it does not add blocking locks, unbounded queues, or remove existing invariants.

## Rollback
After merge:

```bash
```

No DB downgrade or data repair required. Rollback removes the ranked-target API view and related wanted/ranking metadata.

## Known residual risks
- Branch is behind current `origin/main` by one commit; dry merge precheck is clean, but refresh and revalidation are still needed before ready state.
- Repo-preferred `just build` still needs to be run in an environment with `just` installed.


michaelneale:
seems like a good idea, especially for larger meshs where people may want to try different models that aren't there yet

mesh-llm/src/api/mod.rs
```diff
@@ -501,6 +607,198 @@ impl MeshApi {
...
+
+        let mut targets = targets.into_values().collect::<Vec<_>>();
+        for target in &mut targets {
+            target.serving_node_count = serving_count_by_ref
+                .get(&target.model_ref)
+                .copied()
+                .unwrap_or_default();
+        }
+
+        targets.sort_by(|left, right| {
```
ndizazzo: `target_rank` gives us a prioritized ordering and not really neutral aggregation of signals... We haven't really defined any way to weight or figure out precedence a model might get in this computation. 

We should probably formalize the ranking rules somewhere in the docs as a starting point for how you've implemented this - then we can iterate as needed.

mesh-llm/src/api/mod.rs
```diff
@@ -501,6 +607,198 @@ impl MeshApi {
...
 
+    pub(super) async fn model_targets(&self) -> Vec<ModelTargetPayload> {
+        self.model_target_lookup().await.targets
+    }
+
+    async fn wanted_model_refs(&self) -> Vec<String> {
+        self.model_target_lookup().await.wanted_model_refs
+    }
+
+    async fn model_target_lookup(&self) -> ModelTargetLookup {
```
ndizazzo: This method is huge... Recommend split it up and test segments instead

mesh-llm/src/api/mod.rs
```diff
@@ -501,6 +607,198 @@ impl MeshApi {
...
+            .map(|(index, target)| ModelTargetPayload {
+                rank: index + 1,
+                model_ref: target.model_ref,
+                display_name: target.display_name,
+                model_name: target.model_name,
+                explicit_interest_count: target.explicit_interest_count,
+                request_count: target.request_count,
+                last_active_secs_ago: target.last_active_secs_ago,
+                serving_node_count: target.serving_node_count,
+                wanted: target.serving_node_count == 0
```
ndizazzo: `wanted` collapses multiple signals into a decision that simplifies / hides lots of valuable inputs... Maybe we can extract this into some kind of heuristic method that isn't strictly a boolean signal - that way we might have a better view mesh-wide of what to serve instead of binary yes/no signals on serving a model.

mesh-llm/src/api/mod.rs
```diff
@@ -501,6 +607,198 @@ impl MeshApi {
...
+                    && (target.requested
```
ndizazzo: A model with high `request_count` gets sorted early by `count`, then if it's also in `requested_models`, it gets an additional boost from the `(requested && serving==0)` tier... the same underlying demand signal counted twice in the sort chain. 

The wanted flag compounds this because `request_count > 0` satisfies wanted independently of requested. The heuristic method suggested above might help not double count this.

mesh-llm/src/api/routes/model_targets.rs
```diff
@@ -0,0 +1,19 @@
...
+struct ModelTargetListResponse {
+    model_targets: Vec<ModelTargetPayload>,
+}
+
+pub(super) async fn handle(stream: &mut TcpStream, state: &MeshApi) -> anyhow::Result<()> {
+    respond_json(
+        stream,
+        200,
+        &ModelTargetListResponse {
+            model_targets: state.model_targets().await,
```
ndizazzo: It might be worth grouping these so the underlying data stays separate from any computed hints (like ranking or “wanted”), making the intent clearer and easier to evolve later, something like:

```json
{
  "signals": {
    "explicit_interest_count": 3,
    "request_count": 5,
    "serving_node_count": 0,
    "requested": true
  },
  "derived": {
    "target_rank": 1,
    "wanted": true
  }
}
```

Right now everything is just put together (raw signals with derived wantedness)

ndizazzo (changes requested):
I think this is definitely the right track - none of my comments are really about functionality and moreso just forward thinking towards what #276 defines as an end goal and what the shape of that might be

I have a first round of suggestions/observations .. I left a comment on some larger methods that could help to get split up for readability + testing ease

IvGolovach:
Thanks again for the thoughtful review. I took another pass through the branch and updated it to address the feedback more explicitly.

I looked closely at the model-target ranking path, the API response shape, and the way the different demand signals interact. The main thing I wanted to avoid was making this endpoint look authoritative in places where it is really exposing an advisory scheduling/routing hint, so I tightened that contract in both code and docs.

Changes made:

- Split the model-target aggregation logic out of `api/mod.rs` into a dedicated `api/model_targets.rs` module.
- Documented what `target_rank` means and how the ranking is computed.
- Changed `/api/model-targets` so raw inputs live under `signals`, while computed values live under `derived`.
- Made the `wanted` heuristic explicit with `wanted_reason`, instead of hiding the source behind one boolean.
- Adjusted the ranking so `requested` does not double-count a model that already has active request demand.
- Added focused tests around the ranking/wanted behavior and the API response shape.

I also reran the local validation after the update:

- `cargo test -p mesh-llm model_targets -- --nocapture`
- full `cargo test -p mesh-llm -- --test-threads=1`
- `just build`

The intent is that the endpoint now gives operators the useful ranked view, while still making it clear which parts are raw observed mesh state and which parts are derived hints.

PR #404, opened 2026-04-28
Revalidate sticky auto routing for media requests

IvGolovach:
## 1. Summary

Auto-routed chat sessions now reuse a sticky cached model only when that model can satisfy the current request's media requirements.

**Image or audio follow-up requests no longer stay pinned to a text-only cached auto pick.**

## 2. Why

Sticky auto routing improves latency for follow-up turns, but the cached model choice must not bypass per-request media capability checks.

This closes a concrete routing correctness gap where an image/audio request in the same auto session could be sent to a model that is still served but cannot handle the request modality.

## 3. Diff Scope

**Runtime routing**

- `mesh-llm/src/network/openai/transport.rs`: revalidates cached auto-routed models against media requirements before reuse, forgets incompatible cache entries, and keeps mesh-hook behavior tied to the auto-request decision.

**Router capability checks**

- `mesh-llm/src/network/router.rs`: adds a shared media capability predicate used by both cached routing and reclassification filtering.
- Added focused Rust tests for cached text-model rejection on image requests and modality predicate behavior.

## 4. Branch Integrity

**Base branch:** `main`  
**`origin/main`:** `2e465f8c1f6b40cd197db057c09a01970cf868a2`  
**Ahead / behind:** `1 ahead, 0 behind`  
**Merge base:** `2e465f8c1f6b40cd197db057c09a01970cf868a2`  
**Fast-forward safety:** PASS — `origin/main` is an ancestor of `HEAD`; no divergence.

## 5. Commit Integrity

**Introduced commit**

`1cfb1fcde643cbc6d91395973fbb61205c1ac484 Revalidate sticky auto models for media requests`

**One logical change per commit:** confirmed.

**Ledger-Id trailers:** Not applicable — `scripts/ledger/` does not exist in this repository.

**Unique Ledger-Id for squash merge:** Not applicable — `scripts/ledger/` does not exist in this repository.

## 6. Ledger Proof

**Not applicable** — `scripts/ledger/` does not exist in this repository.

## 7. Diff Hygiene

**Changed files**

- `M mesh-llm/src/network/openai/transport.rs`
- `M mesh-llm/src/network/router.rs`

**`git diff --check origin/main...HEAD`:** PASS, no output.

**Forbidden-file confirmation:** no `.env` files, local environment files, secrets, tokens, credentials, `ops/reports` artifacts, `__pycache__`, `*.pyc`, `.DS_Store`, coverage artifacts, build outputs, cache files, or unrelated generated files are included.

## 8. Test Proof

**`cargo test -p mesh-llm`: PASS**

- Unit tests: `1221 passed`, `3 ignored`, `0 failed`
- Integration tests:
  - `protocol_compat_v0_client`: `10 passed`
  - `protocol_convert_matrix`: `11 passed`
  - `virtual_llm_injection`: `1 passed`
- Doc tests: `0 passed`, `0 failed`

**`/tmp/mesh-llm-just-tool/bin/just build`: PASS**

Built llama.cpp, the UI, and the release `mesh-llm` binary.

**Pytest/coverage:** Not applicable — this is a Rust runtime change and this repository has no Python test-runner or coverage gate for this path.

## 9. Verification-Pack Proof

**Not applicable** — `scripts/ops/` does not exist in this repository and no infra/governance/replay/invariant/verification files changed.

## 10. Migration Notes

**Not applicable** — no DB migration changed.

## 11. CI Context Confirmation

**Required contexts:** unchanged for this repository.

**CI context names unchanged.**

**Workflow changes:** Not applicable — no CI/workflow context changes.

## 12. Runtime Safety

**Reviewed runtime files/modules**

- `mesh-llm/src/network/openai/transport.rs`
- `mesh-llm/src/network/router.rs`

**No new blocking locks:** the change only computes media requirements and checks model capabilities before reusing an existing cached auto pick.

**No new unbounded queues:** no queues, channels, background tasks, or buffering paths were added.

**No invariant removal:** the existing served-host check remains, incompatible cached picks are forgotten, and the existing fallback behavior is preserved when no media-capable model is available.

**No invariant regression introduced.**

## 13. Documentation Integrity

**Not applicable** — no README, runbook, documented command, or operational behavior changed.

**Runbook mirror parity:** Not applicable — no runbooks changed.

**Stale references:** Not applicable — no docs changed.

## 14. Rollback Plan

**Squash merge rollback:** `git revert <post_merge_commit_sha>`

**DB downgrade required:** no  
**Data repair required:** no  
**Operational caveats:** none expected; rollback restores the previous sticky auto-cache behavior.

## 15. Known Residual Risks

**No known functional residual risks.**

The successful `just build` emitted existing environment/tooling warnings, including local Node version warning, npm audit output, Vite chunk-size warnings, and llama.cpp compiler warnings. The command exited successfully, and these warnings are unrelated to the changed Rust files.

### DIFF
--- a/mesh-llm/src/network/openai/transport.rs
+++ b/mesh-llm/src/network/openai/transport.rs
@@ -1853,6 +1853,14 @@ fn should_learn_affinity(status_code: u16) -> bool {
     (200..400).contains(&status_code)
 }
 
+fn cached_auto_model_satisfies_media_requirements(
+    model: &str,
+    media: &router::MediaRequirements,
+) -> bool {
+    let caps = crate::models::installed_model_capabilities(model);
+    router::model_satisfies_media_requirements(&caps, media)
+}
+
 // ── Model-aware tunnel routing ──
 
 /// The common request-handling path used by idle proxy, passive proxy, and bootstrap proxy.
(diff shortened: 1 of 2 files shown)


michaelneale:
@IvGolovach yeh I like this - the one reason avoided this was mostly for the default/main web experience, where there are likely not enough image ready ones - so it can do it in the browser, but for api usage I think this is great. wonder if there is a way we can accomodate the special case of they just happen to be using the web console, in which case the browser does image decomposition and description, vs the general case? but I like this.

IvGolovach:
> @IvGolovach yeh I like this - the one reason avoided this was mostly for the default/main web experience, where there are likely not enough image ready ones - so it can do it in the browser, but for api usage I think this is great. wonder if there is a way we can accomodate the special case of they just happen to be using the web console, in which case the browser does image decomposition and description, vs the general case? but I like this.

Thanks, that makes sense, and I agree with the distinction.

I took another pass through the current web console path to make sure this PR does not accidentally break that special case.

**What I checked:**

- The web console already handles image attachments in the browser before the request reaches routing.
- For image attachments, `buildAttachmentBlocks` runs the browser-side image description path and then sends the result as `input_text`.
- It does **not** send the image as `input_image` / `image_url` in that console flow.
- The attachment policy also already treats images and PDFs as browser-handled, so they do not require model media capability from the selected model.
- Only audio and generic file attachments still require model capability checks from the backend/model side.

**What I saw:**

The web console case is already accommodated by the existing frontend pipeline. If the browser decomposes/describes the image first, the router only sees a text request, and this PR will still allow a cached text-only auto model to be reused.

This PR only changes behavior when the actual downstream request still contains real media blocks, like `image_url`, `input_image`, or `input_audio`. In that general API case, rejecting a cached text-only auto pick seems like the right behavior, because the cached model cannot actually satisfy the request payload.

**Test thought:**

I also considered adding a small regression test for the console path, something like: image attachment becomes text content, not an image block.

I held off for now because this PR is currently a narrow Rust routing/runtime fix, and adding UI test coverage would widen the diff beyond the routing bug itself. Happy to add that as a small follow-up if you think it would be useful to make the web-console contract explicit.

ndizazzo (approved):

ndizazzo:
Some non-blocking follow-ups we would do: 

- Add transport-level tests for cached auto invalidation on audio-only and mixed image+audio follow-ups.
- Add one end-to-end sticky-cache test for cache hit → media mismatch → forget → reclassify.
- Document or test the intentional fallback-to-all-served-models behavior so it isn’t mistaken for a missed constraint.

PR #405, opened 2026-04-28
Recovered peers can rejoin the mesh after dead_peers TTL expires

michaelneale:
Peers that temporarily lose connectivity (relay hiccups, cross-region timeouts) can now be re-discovered through gossip after 5 minutes instead of being permanently blocked.

Previously, if a host went unreachable for even a moment, it was added to `dead_peers` and could never come back through transitive gossip — only through a direct inbound reconnection. In relay-partitioned meshes (e.g. Fly console in Sydney, hosts on US relays), this caused stable divergence where different nodes saw different models.

Now: dead peer entries expire after 5 minutes. Once expired, the peer can be re-learned normally through gossip from bridge peers. If the peer is genuinely gone, no bridge peer mentions it and it stays forgotten. No reconnect storms — expiry just removes the block, the next gossip round does the rest.

Verified locally against the live mesh: previously-invisible hosts (GLM on `79af`) re-appeared after TTL expiry, inference routes correctly including `model: auto`.

## What changed

- `dead_peers`: `HashSet<EndpointId>` → `HashMap<EndpointId, Instant>` with 5-minute TTL
- TTL-aware checks in `connect_to_peer` and `update_transitive_peer`
- GC of expired entries in the heartbeat loop (every 60s)
- No wire protocol changes — fully compatible with all existing mesh nodes

### DIFF
--- a/mesh-llm/src/cli/mod.rs
+++ b/mesh-llm/src/cli/mod.rs
@@ -1164,7 +1164,7 @@ mod tests {
         let cli = Cli::parse_from(normalized.normalized);
 
         assert_eq!(cli.log_format, LogFormat::Json);
-        assert_eq!(cli.model, vec!["Qwen3-8B-Q4_K_M".to_string()]);
+        assert_eq!(cli.model, vec![std::path::PathBuf::from("Qwen3-8B-Q4_K_M")]);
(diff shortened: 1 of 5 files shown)


ndizazzo (approved):
LGTM! :shipit:

PR #410, opened 2026-04-30
feature(runtime): add model context info to runtime API

ndizazzo:
## Summary
- `/api/runtime/processes` now returns `context_length` per process, so clients compute effective context (tokens/slot = `context_length / slots`) from a single API call instead of fetching model metadata separately.

## What Changed
Added `context_length: Option<u32>` through the full runtime processes data flow:
| Layer | Change |
|---|---|
| **API** (`RuntimeProcessPayload`) | New optional field, `skip_serializing_if` omits when unknown |
| **Internal** (`RuntimeProcessSnapshot`) | Matching optional field; conversions pass through directly |
| **Runtime** (`local_process_snapshot/payload`) | Accepts `context_length` from `LocalRuntimeModelHandle`, wraps as `Some(...)` |
| **Call sites** (3 in `runtime/mod.rs`) | Pass `process.context_length` / `handle.context_length` |

## Why
In a follow up PR, work will need to be done to fix a bug with OpenCode, where writing the model configuration to the `opencode.json` file does not take into account the number of slots the host reports. By passing the context length the server was started with along with the slots, we will be able to calculate the EFFECTIVE `ctx` for the model in the configuration file.

## Example Output
```json
{
    "name": "qwen2.5-0.5b-instruct-q4_k_m",
    "backend": "llama",
    "status": "ready",
    "port": 58175,
    "pid": 71237,
    "slots": 4,
    "context_length": 65536
}
```

### DIFF
--- a/mesh-llm/src/api/mod.rs
+++ b/mesh-llm/src/api/mod.rs
@@ -1291,6 +1291,7 @@ mod tests {
                     port: 9337,
                     pid: 100,
                     slots: 4,
+                    context_length: None,
                 },
                 RuntimeProcessPayload {
                     name: "Llama".into(),
@@ -1299,6 +1300,7 @@ mod tests {
                     port: 9444,
                     pid: 101,
                     slots: 4,
+                    context_length: None,
                 },
             ],
         );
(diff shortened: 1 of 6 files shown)


michaelneale (approved):
nice

PR #417, opened 2026-05-01
Peer stability: reduce premature removal and gossip noise

michaelneale:
## What this does

Peers in the public mesh stop disappearing prematurely. Relay-only nodes no longer trigger cascading false death reports, and ghost peers stop getting re-introduced after removal.

**Before:** relay-partitioned nodes report peers dead → console confirms (3s timeout too tight) → removes peer → stale gossip re-introduces it → repeat 6× for the same peer. Meanwhile, 0ms RTT artifacts trigger spurious re-elections.

**After:** probes are more patient with relay peers, confirmed deaths are quarantined against gossip re-introduction, and unreliable reporters get silenced after false claims.

## Changes (76 lines added, 9 removed)

1. **RTT measurement fix** — move timing after network read (was measuring local buffer time → 0ms); reject 0ms as invalid
2. **Quarantine on confirmed remote PeerDown** — insert `dead_peers` entry so transitive gossip can't immediately re-introduce the ghost
3. **Quarantine on PeerLeaving** — same protection for graceful shutdowns
4. **Probe timeout increase** — 3s→5s for open_bi, 5s→8s for connect_mesh (relay peers need more time)
5. **Relay-only heartbeat threshold** — 2→3 consecutive failures before death declaration
6. **Reporter cooldown** — track (reporter, target) rejections; suppress repeated false reports for 10 minutes

## Risk

Low. All changes are local decision-making — no wire format changes, fully backward compatible with all existing node versions (0.55–0.65). Worst case: a genuinely dead relay peer takes ~60s longer to detect via heartbeat (data-path failures still catch it instantly during active inference).

## Testing

- `cargo check` ✅
- `cargo test -p mesh-llm --lib` — 1235 tests pass ✅
- `cargo fmt --check` ✅

### DIFF
--- a/crates/mesh-llm/src/api/mod.rs
+++ b/crates/mesh-llm/src/api/mod.rs
@@ -1411,7 +1411,7 @@ mod tests {
             .as_array()
             .expect("processes array");
         assert!(
-            !processes[1].get("context_length").is_some()
+            processes[1].get("context_length").is_none()
                 && processes[1]["context_length"].is_null()
         );
     }
--- a/crates/mesh-llm/src/api/model_targets.rs
+++ b/crates/mesh-llm/src/api/model_targets.rs
@@ -382,12 +382,12 @@ fn model_name_for_model_ref(model_ref: &str, index: &CatalogTargetIndex) -> Opti
     index.model_name_by_ref.get(model_ref).cloned()
 }
 
-fn ensure_model_target<'a>(
-    targets: &'a mut HashMap<String, ModelTargetAccumulator>,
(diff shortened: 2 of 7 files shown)


ndizazzo (approved):

michaelneale:
@copilot can you look at merge conflicts

ndizazzo:
@michaelneale does this PR work address [this ROADMAP item](https://github.com/Mesh-LLM/mesh-llm/blob/main/ROADMAP.md#connection-stability) you've defined?

PR #419, opened 2026-05-01
task(improve-tui): improve TUI startup sequence while preserving raw console commands

### LINKED ISSUE #413: TUI: hangs when downloading / loading model and there is some kind of error
The new TUI needs to be revised so that it loads ASAP, and any errors or issues are surfaced in the new panel. 

Image

Reported by `frag` on Discord

ndizazzo:
## Summary

Users can now see the TUI earlier in the active serve path, while startup is still in flight, and the screen stays honest about what is pending, partial, ready, or failed. Startup history is retained, fatal rpc and llama launch failures stay visible, and fallback output now keeps the same durable failure story when the TUI cannot open.

## What changed

* The active serve path can open the TUI before full runtime readiness.
* Startup now has one canonical lifecycle state, instead of inferring meaning from missing events.
* Startup history is kept for late TUI attach, so earlier events still explain what happened.
* Fatal rpc and llama startup failures now surface durably in the TUI and fallback output.
* Mesh and discovery startup failures map into the startup lifecycle only before readiness completes.
* Endpoint and process rows stay truthful during bootstrap instead of looking ready too early.
* The dashboard now shows an explicit startup summary and failure state.
* Fallback mode and shutdown race handling were hardened.
* The final fix keeps post ready generic errors from reopening startup failure, and it reaps timed out startup children before returning.

## Before vs After

```mermaid
flowchart LR
  subgraph After
    A1[Console bootstrap completes] --> A2[Open TUI early on active serve path]
    A2 --> A3[Track one canonical startup lifecycle]
    A3 --> A4[Retain startup history for late attach]
    A4 --> A5[Show durable failures and truthful row states]
  end

  subgraph Before
    B1[Console boot starts] --> B2[Wait for llama_ready] --> B3[TUI appears late]
    B3 --> B4[Startup events can be missed on late attach]
    B4 --> B5[Failures may disappear into unwind or fallback output]
  end


```

## Validation

* Targeted exact name regressions passed.
* `cargo fmt --all -- --check` passed.
* `cargo check -p mesh-llm` passed.
* `cargo test -p mesh-llm --lib` passed, `1269 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out`.

## Scope

This stayed within TUI and runtime plumbing. It did not change `/api/status`, `NodeState`, `runtime_data::collector`, protocol behavior, or web console behavior.

Fixes #413

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -302,7 +302,13 @@ pkill -f mesh-llm
 
 If the embedded runtime fails to load, check mesh-llm stderr/log output and
 `~/.mesh-llm/runtime/` for the active instance metadata. The old external
-`llama-server` and `rpc-server` log files are no longer produced.
+`llama-server` and `rpc-server` log files are no longer produced. Embedded
+skippy/llama.cpp native logs are redirected away from the TUI into the active
+instance runtime directory:
+
+```text
+<runtime-root>/<pid>/logs/skippy-native.log
+```
 
 To override the runtime root (e.g., for tests or systemd):
 - `MESH_LLM_RUNTIME_ROOT=/path/to/custom/root` — highest priority
--- a/crates/mesh-client/src/models/gguf.rs
+++ b/crates/mesh-client/src/models/gguf.rs
@@ -340,18 +340,18 @@ pub fn scan_gguf_compact_meta(path: &Path) -> Option<GgufCompactMeta> {
         }
     }
 
-    if meta.head_count > 0 {
-        if meta.key_length == 0 {
-            if let Some(key_length) = meta.embedding_size.checked_div(meta.head_count) {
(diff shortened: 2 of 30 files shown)


michaelneale (approved):
I think ok - that is a LOT of lines of code changed, a lot of tests I guess, but looks like better coupling as mod.rs in runtime seens to have less cli-isms in it - fair to say?

ndizazzo:
> that is a LOT of lines of code changed, a lot of tests I guess, but looks like better coupling as mod.rs in runtime seens to have less cli-isms in it - fair to say?

Just acknowledging in here, it's mainly to do with hooking into events for skippy so we can track active states for model loading: 
image

PR #433, opened 2026-05-05
e2e and attestation

michaelneale:
E2E inference encryption (NaCl box, ephemeral keys, forward secrecy) + Secure Enclave hardware attestation + runtime hardening. Automatic when peer advertises an inference key — no opt-in needed.

## Validation so far

Local validation on this branch:

```bash
cargo fmt --all -- --check
cargo check -p mesh-llm
just build
LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm --lib
```

Result: `1306 passed; 0 failed; 3 ignored`.

### Implementation note: encrypted streaming

The encrypted remote path is expected to preserve live SSE/token streaming. The host encrypts response chunks as they arrive, and the client-side tunnel decrypts those chunks into a bounded async pipe that feeds the normal plaintext HTTP response relay. This keeps `/v1/responses` adapters, context-overflow retry detection, error remapping, and token accounting aligned with the plaintext path without buffering the full response before the downstream client sees events.

## Next manual validation: 2-Mac node test

Use two macOS machines, ideally Apple Silicon for Secure Enclave attestation coverage.

### 1. Build and bundle locally

From this branch:

```bash
just build
just bundle
./target/release/mesh-llm --version
```

### 2. Clean both Macs before starting

On **both** Mac A and Mac B:

```bash
pkill -9 -f mesh-llm || true
ps -eo pid,args | grep -E 'mesh-llm' | grep -v grep || true
```

Expected: no remaining `mesh-llm` processes.

### 3. Deploy the same bundle to both Macs

Copy the generated tarball from `dist/` to Mac B, extract it, and codesign if required by local macOS policy.

On both Macs verify the same binary is running:

```bash
mesh-llm --version
```

Expected: both report the version/build from this PR.

### 4. Start Mac B as the model host

On Mac B:

```bash
RUST_LOG=mesh_llm=debug mesh-llm --auto
```

Watch logs for startup, model hosting, and if on Apple Silicon, hardware attestation logs such as:

```text
hardware attestation: SE key bound to node ...
```

Verify status:

```bash
curl -s http://localhost:3131/api/status | jq
curl -s http://localhost:3131/v1/models | jq
```

### 5. Start Mac A and join/discover Mac B

On Mac A, start this PR binary using the normal test mesh flow for your environment, for example auto/discovery or an explicit join token:

```bash
RUST_LOG=mesh_llm=debug mesh-llm --auto
```

or, if using an invite/join flow:

```bash
RUST_LOG=mesh_llm=debug mesh-llm --join '<invite-token-from-mac-b>'
```

Verify both nodes see each other:

```bash
curl -s http://localhost:3131/api/status | jq '.peers // .mesh.peers // .'
curl -s http://localhost:3131/v1/models | jq
```

Expected:

- Mac A sees Mac B as a peer.
- Mac A sees models hosted/routable on Mac B.
- Peer version in status matches this PR build.

### 6. Prove remote encrypted inference works

From **Mac A**, call a model hosted by **Mac B**. Use the exact model id shown in `/v1/models` if needed.

Streaming chat completions:

```bash
curl -N http://localhost:3131/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "auto",
    "messages": [{"role": "user", "content": "Return exactly: encrypted remote path ok"}],
    "stream": true
  }'
```

Non-streaming chat completions:

```bash
curl -s http://localhost:3131/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "auto",
    "messages": [{"role": "user", "content": "Say remote non-stream ok in one sentence."}],
    "stream": false
  }' | jq
```

Expected:

- Responses succeed through Mac A while inference is hosted on Mac B.
- Mac A debug logs show encrypted request/response handling, e.g. messages like:

```text
Encrypted inference request ...
Decrypted response from ...
```

- Mac B debug logs show encrypted tunnel handling, e.g. messages like:

```text
Decrypted inbound HTTP tunnel ...
Encrypted streaming response ...
```

If those log lines do not appear, the test may only be exercising plaintext routing.

### 7. Test `/v1/responses` through the same remote path

This is important because the encrypted response handler should reuse the normal response relay/adapter path.

From Mac A:

```bash
curl -s http://localhost:3131/v1/responses \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "auto",
    "input": "Return exactly: responses endpoint ok"
  }' | jq
```

If streaming responses are enabled in the environment:

```bash
curl -N http://localhost:3131/v1/responses \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "auto",
    "input": "Stream three short words.",
    "stream": true
  }'
```

Expected:

- Valid `/v1/responses` shape.
- No broken chunking or hanging.
- Debug logs still show encrypted tunnel path.

### 8. Test attestation-required behavior

On Apple Silicon Macs, run Mac A with attestation required:

```bash
RUST_LOG=mesh_llm=debug mesh-llm --auto --require-attested-hosts
```

Then repeat a remote inference request from Mac A to Mac B.

Expected when Mac B has valid Secure Enclave attestation:

- Routing succeeds.
- Logs show attestation/security posture and encrypted tunnel path.

Expected when Mac B is not attested, or when testing against a non-Apple-Silicon/older host:

- Routing fails closed.
- Mac A logs a refusal similar to:

```text
refusing to route to unattested host ... (--require-attested-hosts)
```

- No silent plaintext fallback occurs while `--require-attested-hosts` is enabled.

### 9. Mixed-version compatibility check

Run one Mac on this PR and the other Mac on the latest released binary.

Test both useful directions:

1. PR Mac as client/API, released Mac as host.
2. Released Mac as client/API, PR Mac as host.

Expected:

- Gossip/discovery still works.
- `/api/status` shows peers.
- `/v1/models` works.
- Inference still works.
- New-to-old routing should fall back to plaintext because the old host will not advertise an inference key.
- Old-to-new routing should still work because the new host still accepts plaintext tunnels.

### 10. Failure-mode smoke

While a streaming request from Mac A to Mac B is active, stop Mac B:

```bash
pkill -f mesh-llm
```

Expected:

- Mac A does not hang indefinitely.
- Request fails cleanly or retries another host if available.
- Logs classify the tunnel failure as timeout/unavailable rather than panicking.

Also send a malformed request from Mac A:

```bash
curl -s -i http://localhost:3131/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"bad": true}'
```

Expected:

- Error response is relayed/remapped cleanly through the encrypted path.
- No decrypt errors on successful requests.

### Pass criteria

Call the manual 2-Mac validation successful when:

- Both Macs run this PR binary and see each other.
- Remote inference from Mac A to Mac B succeeds.
- Logs prove the encrypted tunnel path is used.
- Streaming and non-streaming `/v1/chat/completions` work.
- `/v1/responses` works through the same path.
- `--require-attested-hosts` accepts an attested Apple Silicon host and rejects unattested hosts.
- Mixed-version PR/release nodes interoperate.
- Killing the remote host during a stream fails cleanly without hangs or panics.


michaelneale:
don't merge without hand testing (at least 2 people)

michaelneale:
Closing — branch too stale after crate restructure. Porting to fresh branch micn/attestation-privacy.

PR #436, opened 2026-05-05
feat: auto-resolve layer packages from meshllm/catalog

michaelneale:
## Remote catalog integration for model resolution

Adds the `meshllm/catalog` HuggingFace dataset as a model resolution source, serving both split and solo modes.

### Split mode (`--split`)

When `--split` is active, the resolver checks the remote catalog for a pre-split layer package before downloading the full monolithic GGUF. If found (e.g. `meshllm/Qwen3-Coder-480B-A35B-Instruct-UD-Q4_K_XL-layers`), each node downloads only its assigned layers via `hf download --include`.

### Solo mode

When a model name isn't found in the baked-in catalog, the resolver falls back to the remote catalog. This means newly added models become available without shipping a new binary.

### Resolver fallback chain

1. Local file on disk → use it
2. Baked-in catalog match → download via catalog
3. Remote `meshllm/catalog` match → download from source repo (solo) or use layer package (split)
4. Canonicalize as HF ref → try that
5. Error

### Changes

| File | What |
|---|---|
| `models/remote_catalog.rs` | New: fetch/cache/query `meshllm/catalog` HF dataset |
| `models/mod.rs` | Register module |
| `models/resolve/mod.rs` | Add remote catalog fallback for solo mode |
| `inference/skippy/package.rs` | `identity_from_layer_package()` — build identity from manifest only |
| `inference/skippy/mod.rs` | Export new functions |
| `runtime/local.rs` | Use `LoadMode::LayerPackage` when layer package found |
| `runtime/mod.rs` | Check catalog before full GGUF download in split mode |
| `skippy-runtime/package.rs` | Selective layer download (`--include` per layer file) |

### Catalog entries added to `meshllm/catalog` HF dataset

- `meshllm/Qwen3-Coder-480B-A35B-Instruct-UD-Q4_K_XL-layers` (62 layers, 294 GB)
- `meshllm/DeepSeek-V3.2-UD-Q4_K_XL-layers` (61 layers, 382 GB)
- `meshllm/Qwen3-235B-A22B-UD-Q4_K_XL-layers` (94 layers, 134 GB)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -407,6 +407,8 @@ jobs:
       LLAMA_STAGE_BACKEND: cuda
       LLAMA_STAGE_CUDA_ARCHITECTURES: "89"
       MESH_LLM_SKIP_UI: "1"
+      # CI containers have no GPU driver (libcuda.so.1). Disable VMM to avoid
+      # linking the CUDA driver library, which isn't needed for validation builds.
       GGML_CUDA_NO_VMM: "1"
     outputs:
       benchmark_artifact_ready: ${{ steps.benchmark_gate.outputs.ready }}
@@ -449,16 +451,7 @@ jobs:
         if: steps.llama_cache.outputs.cache-hit == 'true'
         run: cargo build --release -p mesh-llm
       - name: CLI smoke test
-        run: |
-          # The CUDA devel container has libcuda.so stub but not libcuda.so.1.
-          # Find the stubs directory and create the .so.1 symlink for the dynamic linker.
-          CUDA_STUB="$(find -L /usr/local/cuda /usr/local/cuda-* -path '*/stubs/libcuda.so' -print -quit 2>/dev/null || true)"
-          if [ -n "$CUDA_STUB" ]; then
-            mkdir -p /tmp/cuda-stubs
-            ln -sf "$CUDA_STUB" /tmp/cuda-stubs/libcuda.so.1
-            export LD_LIBRARY_PATH="/tmp/cuda-stubs:$(dirname "$CUDA_STUB"):${LD_LIBRARY_PATH:-}"
-          fi
-          target/release/mesh-llm --version
+        run: target/release/mesh-llm --version
       - name: Build CUDA benchmark binary
         if: ${{ steps.benchmark_gate.outputs.ready == 'true' }}
         run: nvcc -O3 -o target/release/membench-fingerprint-cuda crates/mesh-llm/benchmarks/membench-fingerprint.cu
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -450,6 +450,8 @@ jobs:
         run: cargo build --release -p mesh-llm
       - name: CLI smoke test
         run: |
+          # The CUDA devel container has libcuda.so stub but not libcuda.so.1.
+          # Find the stubs directory and create the .so.1 symlink for the dynamic linker.
           CUDA_STUB="$(find -L /usr/local/cuda /usr/local/cuda-* -path '*/stubs/libcuda.so' -print -quit 2>/dev/null || true)"
           if [ -n "$CUDA_STUB" ]; then
             mkdir -p /tmp/cuda-stubs
--- a/crates/mesh-llm/src/inference/skippy/deployment.rs
+++ b/crates/mesh-llm/src/inference/skippy/deployment.rs
@@ -43,6 +43,7 @@ pub(crate) fn remote_stage_load_request(
         layer_start: stage.layer_start,
         layer_end: stage.layer_end,
         model_path: Some(context.package.package_ref.clone()),
+        source_model_bytes: context.package.source_model_bytes,
         projector_path: None,
         selected_device: None,
         bind_addr: "127.0.0.1:0".to_string(),
@@ -237,6 +238,7 @@ mod tests {
         assert_eq!(request.package_ref, "hf://Mesh-LLM/demo-package");
         assert_eq!(request.manifest_sha256, "manifest");
         assert_eq!(request.load_mode, LoadMode::LayerPackage);
+        assert_eq!(request.source_model_bytes, Some(100));
         assert_eq!(
             request.model_path.as_deref(),
             Some("hf://Mesh-LLM/demo-package")
--- a/crates/mesh-llm/src/inference/skippy/materialization.rs
+++ b/crates/mesh-llm/src/inference/skippy/materialization.rs
@@ -184,13 +184,12 @@ pub(crate) fn resolve_hf_package_to_local(
     // Collect the files we need to download
     let mut needed_files: Vec<String> = Vec::new();
 
-    // Always need shared/metadata.gguf
-    if let Some(path) = manifest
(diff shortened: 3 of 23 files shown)


michaelneale:
## Testing: 2-node layer package split on Qwen3-Coder-480B

Tested end-to-end on two Mac Studios (M3 Ultra 256GB + M1 Ultra 128GB) with the layer package at `meshllm/Qwen3-Coder-480B-A35B-Instruct-UD-Q4_K_XL-layers` (62 layers, 257GB total).

### What works

- Both nodes discover each other and agree on model name via `model_id` from `model-package.json`
- Split topology plans correctly (layers 0..20 on 128GB node, 20..62 on 256GB node)
- `hf://` refs resolve to local HF cache without touching the HF SDK (avoids NFS flock failures and async-in-sync deadlocks)
- **Layer files load directly into GPU memory via `open_from_parts`** — no disk materialization. 175GB + 82GB loaded across the two nodes
- KV cache allocates, compute graph compiles, Metal kernels JIT
- API reports `serving`, `llama_ready: true`, model appears in `/v1/models`

### What fails

Inference requests return `"encode output activation payload"` — the inter-stage activation forwarding from stage-0 to stage-1 fails with an F32 payload size mismatch. The QUIC tunnel between nodes is established and relaying data, but the activation tensor from the model forward pass has an unexpected size vs what `encode_f32_activation_payload` expects for `activation_width=4096`.

This is a binary transport / activation encoding issue, not related to layer package loading. It would likely reproduce with a materialized GGUF too.

### Changes made during testing

Key fixes pushed to this branch:

1. **`open_from_parts` instead of materialization** (`skippy-server/runtime_state.rs`, `skippy-server/binary_transport.rs`) — for `LoadMode::LayerPackage`, load individual GGUF layer files directly into memory via `llama_model_load_from_parts`. Skip the `write_gguf_from_parts` disk concatenation entirely. Also fixed `kv_cache_inspection_path` to use the metadata GGUF instead of materializing the full model.

2. **Skip materialization in coordinator stage-0** (`inference/skippy/mod.rs`) — `load_stage0_config` was still calling `materialize_stage_config`. Now resolves `hf://` to local and lets `load_runtime` use `open_from_parts`.

3. **Resolve stage load without materialization** (`inference/skippy/materialization.rs`, `stage.rs`) — replaced `materialize_stage_load` with `resolve_stage_load_package` that only resolves `hf://` to local path. Stage handler sets `model_path` and `package_ref` to the resolved local dir.

4. **Local HF cache bypass** (`materialization.rs`) — before hitting the HF SDK (which uses `flock` and `block_on`), check if files exist in the local HF cache by reading `refs/` and `snapshots/` directly. Critical for NFS-backed caches.

5. **Canonical model identity** (`runtime/mod.rs`, `package.rs`) — read `model_id` from `model-package.json` so all nodes agree on model name regardless of local filesystem path. Convert local HF cache paths to `hf://` refs via `hf_ref_from_cache_path`.

6. **Layer package detection in split path** (`runtime/local.rs`) — `start_runtime_split_model` checks `is_layer_package_ref` and uses `identity_from_layer_package` instead of `synthetic_direct_gguf_package`.

7. **Timeout increases** — stage control QUIC timeout 180s→900s, binary stage ready probe 120s→900s (large models on NFS can take 5+ minutes to load).

### Next step

Debug the activation encoding mismatch — likely MoE-specific activation shape vs the expected `n_embd * token_count * sizeof(f32)`.

michaelneale:
## Update: Qwen3-Coder-480B inference working end-to-end

The activation encoding issue is resolved. The root cause was `activation_width` defaulting to 4096 when the manifest did not include it — Qwen3-Coder-480B has `n_embd=6144`. Two fixes:

1. **Manifest fix** (upstream): `model-package.json` on HF now includes `"activation_width": 6144`
2. **Runtime fallback** (this branch): `inspect_layer_package` now infers `activation_width` from the first layer `attn_norm.weight` tensor when the manifest omits it

### Verified working

- 480B model (257GB) split across M3 Ultra 256GB (layers 20..62) + M1 Ultra 128GB (layers 0..20)
- No disk materialization — layer GGUFs loaded directly into GPU memory via `open_from_parts`
- Inference producing correct output through `/v1/chat/completions`
- Published as mesh name "keitha" via `--publish --mesh-name keitha`

### Remaining items

- **Timeouts**: stage control (180→900s) and binary stage ready (120→900s) were bumped for NFS. These should be adaptive based on model size or configurable, not hardcoded at 900s
- **HF SDK sync path**: `resolve_hf_package_to_local` falls back to the HF SDK which calls `block_on` inside tokio — deadlocks if the local cache bypass misses. Needs `spawn_blocking` wrapper or full async conversion
- **`HF_HUB_CACHE` requirement**: nodes need `HF_HUB_CACHE` set correctly when the HF cache is on a non-default path (NFS mount, external drive). Could auto-detect from the model path

michaelneale:
## TODO before merge

1. **HF download path deadlocks on fresh nodes.** When a node joins a mesh and does not have the layer files cached locally, `resolve_hf_package_to_local` calls the HF SDK synchronously inside a tokio runtime → `block_on` deadlock. Downloads must happen during `resolve_startup_models` (before the stage handler), not inside the stage control loop. The local cache bypass I added only helps when files are already present.

2. **Stage load timeouts are fixed at 900s.** Bumped from 120/180s to get the 480B model loading from NFS. For the download-from-HF case, even 900s may not be enough for a 170GB download. These should scale with the expected data size or be replaced with a progress-aware mechanism.

Everything else (layer package identity, `open_from_parts`, activation width inference, coordinator stage-0 path) is working and tested with the 480B model across two machines.

michaelneale:
## TODO before merge

1. **HF download path deadlocks on fresh nodes.** `resolve_hf_package_to_local` calls the HF SDK synchronously inside a tokio runtime → `block_on` deadlock. Downloads must happen during `resolve_startup_models` (before the stage handler), not inside the stage control loop. The local cache bypass only helps when files are already present.

2. **Stage load timeouts are fixed at 900s.** Bumped from 120/180s to unblock NFS loading. For remote HF downloads even 900s may not be enough. Should scale with expected data size or use a progress-aware mechanism.

3. **Splitter must write `activation_width` to `model-package.json`.** `skippy-model-package write-package` (`crates/skippy-model-package/src/main.rs`) does not include `activation_width` in `PackageManifest`. The runtime falls back to reading it from the first layer's `attn_norm.weight` tensor, but the splitter should write it at split time so the manifest is self-contained. The Qwen3-Coder-480B manifest was patched manually on HF. The splitter repo is at `/Users/micn/Documents/code/hf-mesh-skippy-splitter`.

michaelneale:
3. **Splitter must emit `activation_width` in `model-package.json`.** `skippy-model-package write-package` (`crates/skippy-model-package/src/main.rs`) builds the `PackageManifest` struct without an `activation_width` field. The runtime falls back to reading it from the first layer GGUF tensor (`attn_norm.weight`), but the manifest should include it so nodes can plan topology without opening GGUF files. The splitter has access to the source model — read `{arch}.embedding_length` from the GGUF metadata and write it into the manifest. The splitter repo is at https://github.com/Mesh-LLM/hf-mesh-skippy-splitter.

ndizazzo:
succeeded by #452

PR #450, opened 2026-05-06
Support duplicate local runtime instances for the same model

IvGolovach:
## Summary
Operators can now load and manage multiple local runtime instances for the same model without one instance replacing the other. Runtime status, process views, and load/unload responses include a stable `instance_id`, so duplicate same-model processes can be inspected and unloaded precisely.

## Why
Local serving was still keyed by model name in several runtime management paths. That made duplicate same-model serving collapse into a single entry, removed sibling local targets, and made unload ambiguous. This keeps public model routing model-centric while adding node-local runtime identity for management surfaces.

## What Changed
- Key local runtime handles, dashboard/process snapshots, and runtime data snapshots by `instance_id` when present.
- Preserve multiple local runtime target ports for the same model.
- Return `instance_id` from runtime load paths, including `/api/runtime/models` and `/mesh/load`.
- Add `DELETE /api/runtime/instances/{instance_id}` for precise unload.
- Keep `DELETE /api/runtime/models/{model}` backwards-compatible when only one instance matches; return `409` when multiple loaded instances make model-name unload ambiguous.
- Add an `Instance` column to `mesh-llm runtime status`.
- Fix Skippy materialization test fixture isolation so repeated lib test runs do not fail on reused fixture paths.

## Compatibility
- No protobuf or mesh protocol schema changes.
- `/v1/models` remains model-centric; duplicate runtime instances are not exposed as duplicate OpenAI models.
- `instance_id` is additive and optional in management payloads, so older payload consumers can ignore it.
- Existing model-name unload still works for unambiguous single-instance cases.

## Validation
Local validation from final HEAD `6e05519a09ab6cfe9c29043982da9ac6048ea2d7` against `origin/main` `20375f4fd6eaba6e5a39e22849584549a677a495`:

- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo check -p mesh-llm`: PASS
- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm materialized_stage_preview_matches_source_removal_candidates`: PASS
- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm --lib`: PASS (1253 passed, 3 ignored)

## Branch Integrity
- Base branch: `main`
- `origin/main`: `20375f4fd6eaba6e5a39e22849584549a677a495`
- Ahead/behind: `0 1`
- Merge-base: `20375f4fd6eaba6e5a39e22849584549a677a495`
- Fast-forward safety: `origin/main` is an ancestor of HEAD.
- Introduced commit: `6e05519a09ab6cfe9c29043982da9ac6048ea2d7 Add runtime instance identity for duplicate local models`

## Rollback
Revert this PR.

DB downgrade: not applicable.  
Data repair: not applicable.  

## Residual Notes
- `just build` was not run locally because this is a Rust-only change and UI assets were not touched.

### DIFF
--- a/crates/mesh-llm/src/cli/commands/runtime.rs
+++ b/crates/mesh-llm/src/cli/commands/runtime.rs
@@ -45,24 +45,53 @@ async fn display_runtime_result(
     verb: &str,
 ) -> Result<()> {
     let action_inf = if verb == "Loaded" { "load" } else { "unload" };
-    if resp.status().is_success() {
-        eprintln!("✅ {verb} runtime model");
-        eprintln!();
-        eprintln!("Model: {model_name}");
-        eprintln!("Scope: Local node");
+    let is_success = resp.status().is_success();
+    let body = resp.json::<serde_json::Value>().await.ok();
+    if is_success {
+        for line in runtime_success_lines(model_name, verb, body.as_ref()) {
+            eprintln!("{line}");
+        }
     } else {
         eprintln!("❌ Failed to {action_inf} runtime model");
         eprintln!();
         eprintln!("Model: {model_name}");
-        let reason = resp
-            .json::<serde_json::Value>()
-            .await
-            .ok()
-            .and_then(|v| v["error"].as_str().map(str::to_owned))
+        let reason = body
+            .as_ref()
+            .and_then(|value| value["error"].as_str().map(str::to_owned))
             .unwrap_or_else(|| "unknown error".to_string());
         eprintln!("Reason: {reason}");
     }
     Ok(())
+}
+
+fn runtime_success_lines(
+    model_name: &str,
+    verb: &str,
+    body: Option<&serde_json::Value>,
+) -> Vec<String> {
+    let response_model_key = match verb {
+        "Loaded" => "loaded",
+        "Unloaded" => "dropped",
+        _ => "model",
+    };
+    let display_model = body
+        .and_then(|value| value[response_model_key].as_str())
+        .filter(|value| !value.trim().is_empty())
+        .unwrap_or(model_name);
+    let instance_id = body
+        .and_then(|value| value["instance_id"].as_str())
+        .filter(|value| !value.trim().is_empty());
+
+    let mut lines = vec![
+        format!("✅ {verb} runtime model"),
+        String::new(),
+        format!("Model: {display_model}"),
+    ];
+    if let Some(instance_id) = instance_id {
+        lines.push(format!("Instance: {instance_id}"));
+    }
+    lines.push("Scope: Local node".to_string());
+    lines
 }
 
 /// Percent-encode a string for use as a URL path segment.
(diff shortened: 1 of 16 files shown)


crates/mesh-llm/src/runtime_data/processes.rs
```diff
@@ -3,6 +3,7 @@ use crate::api::RuntimeProcessPayload;
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub(crate) struct RuntimeProcessSnapshot {
     pub model: String,
+    pub instance_id: Option<String>,
```
i386, line 6: [P2] Duplicate runtime telemetry collapses by model

The new process identity supports multiple instances of the same model, but llama runtime snapshots are still stored under only `model` in `RuntimeDataCollector::replace_llama_slots_snapshot` (`crates/mesh-llm/src/runtime_data/collector.rs:173-177`). When two local `Qwen` instances publish slot/runtime data, they overwrite each other in `llama_runtime_by_model`, so `/api/runtime/llama` and dashboard lane data can show the wrong instance or flicker between siblings. The telemetry key needs to include `instance_id` or otherwise expose per-instance snapshots.

crates/mesh-llm/src/cli/commands/runtime.rs
```diff
@@ -121,11 +121,12 @@ pub(crate) async fn run_status(port: u16) -> Result<()> {
...
 
     println!(
-        "{:<42} {:<8} {:<10} {:<8} {:<6}",
-        "Model", "Backend", "State", "Pid", "Port"
+        "{:<42} {:<12} {:<8} {:<10} {:<8} {:<6}",
+        "Model", "Instance", "Backend", "State", "Pid", "Port"
     );
     for model in models {
         let name = model["name"].as_str().unwrap_or("unknown");
+        let instance = model["instance_id"].as_str().unwrap_or("-");
```
i386, line 158: [P2] Load CLI hides the instance id needed for duplicate unloads

Successful load responses now include `instance_id`, but the CLI success path ignores the JSON body and only prints the model name. In the duplicate-model workflow, the next `mesh-llm unload <model>` returns the new ambiguity error and tells the user to unload by instance id, but the load command did not show that id. Parse the success body and print `Instance: runtime-N` when present.

i386 (commented):
Review findings from Codex.

i386 (changes requested):
could you have a deeper look at where we might be collapsing loaded model identity by name across the code base? I think you may have uncovered something nasty.....

@ndizazzo would be worth getting your eyes on this with a lens on the config work you've been doing

IvGolovach:
Thanks again for the careful review. I went back through the runtime identity path and did a deeper pass on the places where duplicate same-model instances could still collapse back to model-name identity.

What I changed:

- Runtime llama snapshots are now retained by `instance_id`, so two local instances serving the same model keep separate slot/metric state.
- The existing model-level projection is still kept as a compatibility fallback, but it no longer acts as the only source of truth for duplicate instances.
- Dashboard lanes now prefer the process `instance_id` snapshot, which should prevent same-model instances from showing the wrong slots or flickering between each other.
- `/api/runtime/llama` now exposes per-instance snapshots via `instances[]`, and the slot payload includes `instance_id` when known.
- The CLI load/unload success path now prints `Instance: runtime-N` when the API returns it, so duplicate same-model unloads are discoverable from the command line.

I also added focused regression coverage for the exact failure mode: two same-model runtime snapshots stay distinct, unloading one instance does not erase the other, dashboard lanes bind to the matching instance, `/api/runtime/llama` exposes the instance view, and CLI output includes the instance id.

Local validation is green, including the full library suite. Thanks for catching this — the first pass fixed runtime control identity, but your review showed that the telemetry and dashboard paths still needed to use the same `instance_id` boundary.

i386 (approved):
Merge at will

PR #443, opened 2026-05-06
Auto-split Skippy runtime models that exceed local capacity

IvGolovach:
## Summary

`mesh-llm serve --model ...` can now automatically use the Skippy split runtime when a model exceeds the local node's runtime capacity but can be served by the mesh collectively.

- Oversized startup models now try split loading instead of failing on the solo local-fit guard.
- Models that still fit locally keep the existing solo Skippy path unless `--split` is explicitly requested.
- Explicit `--split` remains a forced split path.
- Runtime capacity checks now count all split GGUF parts before deciding whether a local model fits.

## Why

The documented two-node behavior says a model that is too large for one node should split automatically across eligible peers. The previous startup path only entered split loading when `--split` was set, so a model could fail locally even when the mesh had enough aggregate capacity.

This closes that capacity-planning gap for Skippy runtime startup without changing protocol compatibility or broad routing behavior.

## Architecture

- Added a small startup runtime planner that chooses `Local` vs `Split` from the explicit split flag, local capacity, and total model footprint.
- Added aggregate split-capacity validation before launching a topology.
- Added per-stage capacity validation using each planned stage's parameter bytes and assigned node VRAM.
- Reused `election::total_model_bytes` for runtime fit checks so split GGUF shards are counted as one model footprint.

## Diff scope

- `crates/mesh-llm/src/runtime/local.rs`
  - Adds startup split/local planning.
  - Validates split aggregate capacity and per-stage capacity.
  - Carries `parameter_bytes` in runtime stage plans.
  - Adds focused unit coverage for local-fit, auto-split, forced-split, aggregate rejection, and per-stage rejection.
- `crates/mesh-llm/src/runtime/mod.rs`
  - Uses the startup runtime planner before launching the runtime.
  - Keeps the solo path for fitting models.
  - Updates demand/promotion capacity checks to count full split GGUF footprints.
  - Adds split-GGUF capacity coverage.

## Branch integrity

- Base branch: `main`
- Validated base SHA: `679d23826043b874a9b3bc067b0d3851cba5a973`
- Ahead / behind vs `origin/main`: `0 / 1`
- Merge-base SHA: `679d23826043b874a9b3bc067b0d3851cba5a973`
- Fast-forward safety: `origin/main` is an ancestor of this branch.

## Commit integrity

- Introduced commit: `8cf07108af469a25a571b54e135a754843d36f72 Fix Skippy split capacity planning`
- One logical change: confirmed.
- Final diff contains only intended runtime files.

## Diff hygiene

[...]

PR #452, opened 2026-05-06
feat: auto-resolve layer packages from meshllm/catalog rebased

ndizazzo:
Original PR #436 - GitHub was choking on a 90+ comment, 21 commit PR page

## Remote catalog integration for model resolution

Adds the `meshllm/catalog` HuggingFace dataset as a model resolution source, serving both split and solo modes.

### Split mode (`--split`)

When `--split` is active, the resolver checks the remote catalog for a pre-split layer package before downloading the full monolithic GGUF. If found (e.g. `meshllm/Qwen3-Coder-480B-A35B-Instruct-UD-Q4_K_XL-layers`), each node downloads only its assigned layers via `hf download --include`.

### Solo mode

When a model name isn't found in the baked-in catalog, the resolver falls back to the remote catalog. This means newly added models become available without shipping a new binary.

### Resolver fallback chain

1. Local file on disk → use it
2. Baked-in catalog match → download via catalog
3. Remote `meshllm/catalog` match → download from source repo (solo) or use layer package (split)
4. Canonicalize as HF ref → try that
5. Error

### Changes

| File | What |
|---|---|
| `models/remote_catalog.rs` | New: fetch/cache/query `meshllm/catalog` HF dataset |
| `models/mod.rs` | Register module |
| `models/resolve/mod.rs` | Add remote catalog fallback for solo mode |
| `inference/skippy/package.rs` | `identity_from_layer_package()` — build identity from manifest only |
| `inference/skippy/mod.rs` | Export new functions |
| `runtime/local.rs` | Use `LoadMode::LayerPackage` when layer package found |
| `runtime/mod.rs` | Check catalog before full GGUF download in split mode |
| `skippy-runtime/package.rs` | Selective layer download (`--include` per layer file) |

### Catalog entries added to `meshllm/catalog` HF dataset

- `meshllm/Qwen3-Coder-480B-A35B-Instruct-UD-Q4_K_XL-layers` (62 layers, 294 GB)
- `meshllm/DeepSeek-V3.2-UD-Q4_K_XL-layers` (61 layers, 382 GB)
- `meshllm/Qwen3-235B-A22B-UD-Q4_K_XL-layers` (94 layers, 134 GB)


ndizazzo:
@copilot Please re-review the PR in a new session

i386:
@michaelneale @ndizazzo is this an opportunity to delete the baked-in catalog catalog? we can follow up in another PR?

ndizazzo:
> opportunity to delete the baked-in catalog catalog? we can follow up in another PR?

@i386 Probably... Not a huge fan of hand-maintaining a list if we can potentially fetch it from our HF source, but I defer to @michaelneale on the call since he wrote it!

i386:
I think there is a second behavior problem when this lands after #443.

#443 added auto-split when a requested startup model exceeds local runtime capacity, but this PR only checks the remote `meshllm/catalog` layer-package path when `--split` was explicitly set. With no `--split`, startup still resolves/downloads the monolithic GGUF first, then #443 may auto-split from that local GGUF. That misses the main benefit of the remote layer package: avoiding the original giant GGUF download.

Desired behavior: layer-package discovery should happen before deciding local-vs-split for remote/catalog models, not only when `--split` is set. If a matching `-layers` package exists, use its manifest for planning. If the machine can run the whole model locally, download all package layers/shared files and load from the layer package. If it cannot, use the same package for split planning and have each node download only its assigned range. Only fall back to the original monolithic GGUF when no suitable layer package is found or the layer-package path fails under an explicit fallback policy.

```mermaid
flowchart TD
    A["Startup model request"] --> B["Normalize model ref<br/>name, HF ref, repo:selector, local path"]

    B --> C{"Local path exists?"}
    C -->|"Yes, GGUF"| D["Inspect local GGUF size"]
    C -->|"Yes, layer package dir"| E["Inspect layer package manifest"]
    C -->|"No / HF / catalog ref"| F["Check remote catalog for matching -layers package"]

    F -->|"Found"| G["Download/read package manifest only"]
    F -->|"Not found"| H["Fallback: resolve/download original GGUF"]

    G --> I["Plan from manifest:<br/>layer count, source bytes, per-layer bytes, activation width"]
    E --> I
    D --> J["Plan from monolithic GGUF metadata"]

    I --> K{"Fits on this node?"}
    K -->|"Yes"| L["Download all required package files:<br/>metadata, embeddings, all layers, output"]
    L --> M["Load as LayerPackage locally<br/>no original GGUF download"]

    K -->|"No"| N{"Mesh can split it?"}
    N -->|"Yes"| O["Download assigned layer range only<br/>per node"]
    O --> P["Load split LayerPackage stages"]
    N -->|"No"| Q["Fail with capacity message<br/>or fallback if policy allows"]

    J --> R{"Fits on this node?"}
    R -->|"Yes"| S["Load local GGUF normally"]
    R -->|"No"| T{"Mesh can split it?"}
    T -->|"Yes"| U["Split RuntimeSlice from GGUF<br/>requires original GGUF source"]
    T -->|"No"| Q

    H --> J
```

In short: `--split` should force split topology, but it should not be the only path that probes for a `-layers` package. Layer packages should become the preferred remote artifact for both solo and split startup, with the single GGUF as fallback.

i386:
@ndizazzo one more follow-up that I think would make this cleaner: port the existing built-in `catalog.json` data into `meshllm/catalog` first, then remove the hard-coded catalog entirely and make the HF-backed catalog the single resolver source.

That gives us one catalog path for:

- friendly curated names
- exact `repo:selector` / `repo@rev/file` refs
- source GGUF coordinates
- split GGUF `extra_files`
- `mmproj`
- draft model refs
- MoE metadata
- layer-package pointers

Then the resolver can return candidate artifacts from one source of truth instead of checking bundled catalog, then remote catalog, then fallback refs. It should also make the no-`--split` layer-package planning path much cleaner because the catalog entry can expose both the monolithic GGUF source and any `-layers` package candidates before runtime chooses local vs split.

Important guardrail: **we MUST NOT move catalog data that does not have a valid Hugging Face coordinate.** If a bundled entry cannot be represented as source repo + revision + file, it should stay out of the HF catalog migration until it has real coordinates or is intentionally dropped. No raw blob URL or coordinate-less legacy entry should become catalog-of-record data.

Suggested order:

1. Port only HF-coordinate-backed entries from `crates/mesh-llm/src/models/catalog.json` into `meshllm/catalog`.
2. Verify parity for name, source repo/revision/file, size, description, draft, `extra_files`, `mmproj`, and MoE metadata.
3. Port the #330-style HF-backed catalog loader into the current `crates/mesh-llm` layout.
4. Fold this PR's `remote_catalog` lookup into the main `catalog` API instead of keeping two catalog paths.
5. Remove the bundled catalog once parity is proven.

That should leave runtime policy with a simpler question: catalog says what artifacts exist; runtime decides which artifact to load.

i386:
Follow-up architecture suggestion: move the model resolution core into a new workspace crate named `model-resolver`.

This is starting to get complicated enough that a crate boundary would help: startup resolution, catalog lookup, exact-ref normalization, layer-package discovery, local HF cache identity, and fallback behavior are all becoming one decision surface. A small resolver crate would force that code to return structured artifact candidates instead of mixing resolver, download, and runtime policy concerns.

Suggested narrow scope for `model-resolver`:

- parse and normalize model requests:
  - friendly catalog names
  - `owner/repo:selector`
  - `owner/repo@revision/file`
  - local paths as input coordinates
- load/query the HF-backed `meshllm/catalog`
- support local fixture/catalog-root overrides for tests
- use stale cached catalog data when HF is unavailable
- validate HF coordinates for catalog entries
- return candidate artifacts for exactly these four cases:

```rust
pub enum ModelArtifactCandidate {
    LocalGguf(LocalGguf),
    RemoteGguf(RemoteGguf),
    LocalLayerPackage(LocalLayerPackage),
    RemoteLayerPackage(RemoteLayerPackage),
}
```

Explicit non-scope:

- no `RemoteMoePackage`
- no MoE package resolution
- no expert/trunk package layouts
- no MoE package publication or loading
- no local-vs-split runtime decisions
- no aggregate VRAM planning or per-node layer assignment
- no Skippy startup or mesh peer coordination
- no CLI/API formatting
- no usage tracking / cleanup policy

We developed the remote catalog for MoE packages originally in https://github.com/Mesh-LLM/mesh-llm/pull/330, but that is now dead, so this resolver should stay focused on GGUF and layer-package artifacts only: local GGUF, remote GGUF, local layer package, remote layer package.

Then `mesh-llm` can ask the crate, "what artifacts can satisfy this model request?" and runtime can separately decide, "given this machine and mesh, should I load all layers locally or assign layer ranges across nodes?"

i386 (changes requested):
I think it's worth doing the overhaul now and getting this working nicely before it gets any more complicated. Killing builtin catalog + introducing `model-resolver` is a great way to simplify.

michaelneale:
yeah I think that makes sense too! want me to take a look @ndizazzo or you grinding?

michaelneale:
the based rebase - good luck! Not sure, but we do need to make sure we don't shell out to hf cli, current main does that I believe which wont' really cut it.

i386:
I found two separate issues that need different fixes:

1. **Catalog schema mismatch blocks remote catalog loading**

`model-resolver` currently declares `CuratedMeta::draft` as `Option<bool>` and `mmproj` as `Option<String>`, but the live `meshllm/catalog` entries use draft model refs as strings and multimodal projectors as objects. For example, `Qwen3-8B-GGUF.json` has `"draft": "Qwen3-0.6B-Q4_K_M"`, and `Qwen3.5-9B-GGUF.json` has object-valued `mmproj` metadata.

That means `ensure_catalog()` can fail while deserializing live entries, disabling the remote catalog and therefore the layer-package discovery path this PR depends on. It also loses draft metadata later because `remote_model_from_variant()` currently sets `draft: None`.

Suggested fix:
- Change `CuratedMeta::draft` to `Option<String>`.
- Change `CuratedMeta::mmproj` to accept the live object shape, either as a structured type or `serde_json::Value` with validation in `remote_catalog`.
- Preserve `variant.curated.draft.clone()` in `remote_model_from_variant()` instead of hardcoding `None`.

2. **Layer-package lookup misses exact repo selector refs**

`ModelResolver::resolve()` does fuzzy catalog matching before it normalizes parseable model refs. A common startup ref like:

```text
unsloth/Qwen3-8B-GGUF:Q4_K_M
```

will not match `entry.source_repo == unsloth/Qwen3-8B-GGUF` because the query still includes `:Q4_K_M`. The later `ModelRef::parse` fallback returns only a `RemoteGguf` with no package candidates, so `resolve_split_layer_package()` falls back to downloading the monolithic GGUF even when the catalog has a matching `-layers` package.

Suggested fix:
- Parse `ModelRef::parse(input)` before the catalog loop.
- During catalog matching, match by parsed repo against `entry.source_repo` / `variant.source.repo`.
- If a selector is present, only accept variants whose variant name or `variant.source.file` quant selector matches it.
- Keep returning `RemoteLayerPackage` before `RemoteGguf`.
- Add a unit test that resolving `unsloth/Qwen3-8B-GGUF:Q4_K_M` returns `RemoteLayerPackage("meshllm/Qwen3-8B-Q4_K_M-layers")` first.

PR #453, opened 2026-05-06
fix: wire --discover into serve path, add --name to discover subcommand

michaelneale:
--discover <name> was completely ignored on the serve/client startup path — only worked for the MCP path. Users running 'mesh-llm --discover mic22' would silently create a new standalone mesh instead of joining mic22.

Changes:
- --discover triggers Nostr discovery on serve path (same as --auto)
- --discover name is used as target mesh name filter
- discover subcommand gains --name flag for filtering by mesh name
- Suppress misleading 'private by default' warning when joining
- Fix discover subcommand help text (was suggesting nonexistent 'discover --join')
- MeshFilter gains name field for case-insensitive exact mesh name matching

Tested: --join token, --discover name (serve), client --discover name, discover --name, relay-only join

Next: want to diagnose joining when only relay endpoint is accessible

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,11 +195,19 @@ mesh-llm serve --auto --model GLM-4.7-Flash-Q4_K_M --mesh-name "poker-night" --p
 ```
 Everyone runs the same command. First person creates it, everyone else discovers "poker-night" and joins automatically. Use `--publish` to make your named mesh discoverable on Nostr; without it the mesh is private but still joinable via invite token.
 
+You can also join a named mesh explicitly:
+```bash
+mesh-llm serve --discover "poker-night"    # discover by name, join, and serve
+mesh-llm client --discover "poker-night"   # discover by name, join as client
+```
+
 ### Auto-discover
 ```bash
 mesh-llm serve --auto                      # discover, join, and serve a model
 mesh-llm client --auto                     # join as API-only client (no GPU)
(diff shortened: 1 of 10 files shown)


ndizazzo:
@michaelneale agent-assisted review after our testing session earlier:

- Bug: Bare MCP --discover fails to behave like --auto. cli.discover parses as Some("") in `crates/mesh-llm/src/cli/mod.rs:254-256`, and join_mesh_for_mcp passes that directly to smart_auto in `crates/mesh-llm/src/runtime/mod.rs:3147-3161`, where Some("") exact-filters out all normal meshes in `crates/mesh-llm/src/network/nostr.rs:983-995`.
 
- Regression: Serve/client --discover joins initially, but misses --auto lifecycle recovery. Rediscovery/watchdog/takeover still gate on cli.auto only in `crates/mesh-llm/src/runtime/mod.rs:3504`, 3542, 4136, and 4552, despite the PR/docs saying bare --discover behaves like --auto.

- Consistency issue: MCP discovery joins bypass the new retry path. Explicit MCP --join and main join use join_with_retry, but MCP auto/discover candidates still call node.join(token).await at `crates/mesh-llm/src/runtime/mod.rs:3175`, making that path less resilient to transient join failures.

ndizazzo (changes requested):
See above

ndizazzo (approved):
👌🏻👍🏻

PR #454, opened 2026-05-07
Recover Skippy split serving after stage peer loss

IvGolovach:
## Summary

Skippy split-serving now recovers when an active stage peer leaves the mesh instead of keeping a stale split topology alive.

When a replacement topology can still be formed, the coordinator accepts a degraded but valid topology even if it has fewer stages or participants. When quorum is no longer available, the coordinator asks the runtime loop to withdraw the active split generation so the model is no longer advertised through a broken topology.

## Why

The split topology coordinator already replanned on membership changes, but the acceptance gate only treated a candidate as useful when it had more stages, more participants, or a better balance score. That optimized for upgrades, but it missed a correctness case: after a stage owner disappears, the replacement topology may be smaller and less balanced while still being the only valid topology.

Keeping the old generation in that case leaves routing pointed at a stage peer that is no longer present. This change makes peer-loss recovery a separate correctness condition from normal topology improvement.

## Diff Scope

- `crates/mesh-llm/src/runtime/local.rs`
  - Adds loss-aware active-stage participant detection.
  - Accepts replacement candidates when the active topology references a missing stage owner.
  - Adds a fail-closed withdrawal event when a stage peer is lost and split quorum is no longer available.
  - Keeps existing planner validation, stage-0 locality checks, and load-before-cutover behavior.
  - Adds focused unit coverage for degraded recovery and unused-participant loss.

- `crates/mesh-llm/src/runtime/mod.rs`
  - Handles split coordinator withdrawal events in the startup runtime loop.
  - Emits a warning and lets the existing shutdown/cleanup path remove local routing, unregister the runtime instance, and stop the split generation.

## Branch Integrity

- Base branch: `main`
- Validated base: `origin/main`
- `git rev-parse origin/main`: `e2a0b5770a91ea4fddb2c2cae7f067a5c7f34d4b`
- Ahead count: `1`
- Behind count: `0`
- Merge base: `e2a0b5770a91ea4fddb2c2cae7f067a5c7f34d4b`
- Fast-forward safety: `git merge-base --is-ancestor origin/main HEAD` exited `0`; `origin/main` is an ancestor of `HEAD`.

## Commit Integrity

Introduced commit:

- `7697dcd977ea6c0903581224b71c888513c1d908 Recover Skippy split topology after stage peer loss`

## Diff Hygiene

```text
M	crates/mesh-llm/src/runtime/local.rs
M	crates/mesh-llm/src/runtime/mod.rs
```

## Validation Mode And Proof

Validation mode: **Mode 3 - shared backend/runtime behavior**.

Reason: this changes Skippy split runtime coordination and failover behavior for distributed serving, while keeping the change scoped to runtime split topology recovery.

Local validation:

```text
cargo fmt --all -- --check
PASS

cargo check -p mesh-llm
PASS

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm split_replan --lib
PASS - 4 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm split_missing_active_stage_nodes --lib
PASS - 1 passed

cargo test -p skippy-topology --lib
PASS - 16 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm runtime::proxy::tests::test_api_proxy_integration_streaming_response_arrives_incrementally --lib -- --exact
PASS - 1 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm --lib
PASS - 1291 passed, 3 ignored
```

Note: one earlier full `mesh-llm --lib` run hit a timing-sensitive assertion in the unrelated proxy streaming integration test. The exact test passed on isolated rerun, and the final full library suite passed.

Not run: `just build` - not required for this Rust-only runtime change under the selected validation mode; the Rust crate check and full library suite passed locally.

## CI Context Confirmation

Required CI context names unchanged.

## Runtime Safety

Reviewed runtime modules:

- `crates/mesh-llm/src/runtime/local.rs`
- `crates/mesh-llm/src/runtime/mod.rs`

Runtime safety notes:

- No new protocol or public API schema fields.
- No new blocking locks.
- No new unbounded queues.
- No in-flight request resume is attempted; active generations fail naturally and later requests use a fresh topology.
- Existing stage-0 locality validation remains in place.
- Existing topology planner and capacity validation remain in place before replacement candidates can be accepted.
- Replacement generations are still loaded before cutover.
- When quorum is lost, the runtime fails closed by withdrawing the active split generation instead of keeping a stale route advertised.

## Documentation Integrity

Not applicable - no documented command, runbook, protocol, or public API contract changed.

## Migration Notes

Not applicable - no DB migration changed.

## Rollback Plan

Rollback: `git revert <post_merge_commit_sha>`.

DB downgrade required: no.

Data repair required: no.

Operational caveats: rollback restores the previous Skippy split replan acceptance behavior.

## Known Residual Risks

- No known functional residual risks after local validation.
- This PR does not implement in-flight split request resume; failed active generations are intentionally allowed to fail, and the next request uses the replacement topology once ready.

### DIFF
--- a/crates/mesh-llm/src/runtime/local.rs
+++ b/crates/mesh-llm/src/runtime/local.rs
@@ -621,6 +621,28 @@ struct SplitGenerationLoadSpec<'a> {
 async fn load_split_runtime_generation(
     spec: SplitGenerationLoadSpec<'_>,
 ) -> Result<SplitRuntimeGenerationHandle> {
+    let mut cleanup_on_error = false;
+    let result = load_split_runtime_generation_inner(&spec, &mut cleanup_on_error).await;
+    if let Err(error) = &result {
+        if cleanup_on_error {
+            tracing::warn!(
+                model_ref = spec.model_ref,
+                topology_id = %spec.generation.topology_id,
+                run_id = %spec.generation.run_id,
+                generation = spec.generation.generation,
+                error = %error,
+                "cleaning up split runtime generation after failed load"
+            );
+            stop_split_generation(spec.node, spec.generation, spec.generation.generation).await;
+        }
+    }
+    result
+}
+
+async fn load_split_runtime_generation_inner(
+    spec: &SplitGenerationLoadSpec<'_>,
+    cleanup_on_error: &mut bool,
+) -> Result<SplitRuntimeGenerationHandle> {
     let stage0 = spec
         .generation
         .stages
@@ -665,6 +687,7 @@ async fn load_split_runtime_generation(
         skippy_stage_activation_width(spec.package.activation_width, spec.model_ref)?;
 
     for stage in spec.generation.stages.iter().skip(1).rev() {
+        *cleanup_on_error = true;
         let load = skippy::StageLoadRequest {
             topology_id: spec.generation.topology_id.clone(),
             run_id: spec.generation.run_id.clone(),
@@ -777,7 +800,7 @@ async fn load_split_runtime_generation(
         downstream.stage_id,
         downstream.stage_index,
         downstream_endpoint,
-        spec.projector_path,
+        spec.projector_path.clone(),
         spec.ctx_size,
         spec.slots as u32,
         kv_cache,
@@ -2277,6 +2300,7 @@ mod tests {
     use iroh::SecretKey;
     use sha2::{Digest, Sha256};
     use std::fs;
+    use std::sync::{Arc, Mutex as StdMutex};
 
     fn make_id(seed: u8) -> iroh::EndpointId {
         let mut bytes = [0u8; 32];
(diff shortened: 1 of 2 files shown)


i386 (commented):
`SPLIT_DEFAULT_MIN_PARTICIPANTS` is still useful, but it should be treated as the minimum size for a split stage topology, not as a general model-availability check. Now that split serving is no longer just an explicit CLI mode and the runtime can decide between local and split serving automatically, the recovery path needs to make that distinction explicit.

If the active split loses a required stage peer, recovery should be based on whether we can actually keep serving the model: either form and load a valid replacement split topology with at least two stage participants, or fall back to local serving if this node can fit the whole model. If neither is possible, withdraw the advertised model.

As written, the withdrawal path is tied mostly to `participants.len() < 2`, but `participants.len() >= 2` does not prove recovery is possible. The planner can still fail because the remaining nodes do not have enough aggregate or per-stage capacity, leaving the stale active topology advertised. Please make the fail-closed decision depend on missing active stage peers plus replacement/local fallback failure, while still enforcing the two-or-more participant minimum whenever we actually choose a split topology.

i386 (changes requested):

crates/mesh-llm/src/runtime/local.rs
```diff
@@ -938,60 +972,170 @@ impl SplitTopologyCoordinator {
...
+                    active_generation = self.active.generation,
+                    candidate_topology_id = candidate.topology_id,
+                    candidate_generation = candidate.generation,
+                    missing_stage_nodes = ?split_node_labels(&missing_stage_nodes),
+                    active_stages = ?split_stage_plan_labels(&self.active.stages),
+                    candidate_stages = ?split_stage_plan_labels(&candidate.stages),
+                    participants = ?split_participant_labels(&candidate.participants),
+                    "split topology lost an active stage peer; loading replacement split generation"
+                );
+                match self.load_and_publish_candidate(reason, candidate).await {
```
i386, line 1065: @IvGolovach If `load_split_runtime_generation` fails after preparing or loading one or more downstream stages, this recovery path can fall back locally or withdraw the active generation and then stop the coordinator without stopping that partially loaded candidate generation. That leaves those candidate stages consuming resources on peers with no later cleanup path. Please make candidate load failure call `stop_split_generation` for the candidate generation, or otherwise add scoped cleanup inside `load_split_runtime_generation` for already prepared/loaded stages.

i386 (commented):
Added one inline cleanup finding.

crates/mesh-llm/src/runtime/local.rs
IvGolovach: Thanks for catching this — I walked through the failure path again, and this was a real cleanup gap.

I updated `load_split_runtime_generation` so that once a candidate generation has started preparing/loading downstream stages, any later load failure now tears down that candidate generation with `stop_split_generation` before the coordinator moves on to local fallback or withdrawal.

I also added a focused regression test for the partial-load case: one downstream stage loads, the next stage fails, and we verify that the candidate stages receive the cleanup `Stop` request with the candidate generation.

Appreciate you pointing this out. This makes the failure path much cleaner, and the test should keep this from quietly coming back later 🙂

PR #455, opened 2026-05-07
feat: huggingface split job for skippy individually downloadable layers

michaelneale:
This PR makes layer-package creation a first-class mesh-llm workflow for turning Hugging Face GGUF models into individually downloadable skippy layer packages.

The user-facing command is:

```bash
mesh-llm models package unsloth/Qwen3-8B-GGUF:Q4_K_M
```

By default this is a dry run: it resolves the selected GGUF quant, chooses CPU HF Jobs hardware for the splitter, prints the effective timeout, shows the maximum job cost, and emits the job spec without submitting anything. Add `--confirm` to actually submit the HF Job:

```bash
mesh-llm models package unsloth/Qwen3-8B-GGUF:Q4_K_M --confirm
```

The previous top-level `mesh-llm model-prepare` entrypoint is retained as a hidden compatibility shim, but the implementation crate is now named `model-package`.

## Packaging Behavior

- Accepts mesh-llm/HF model coordinates like `org/repo:Q4_K_M` instead of splitting the quant into a separate generated `--quant` argument.
- Selects high-quality 4-bit GGUF variants by preference (`UD-Q4_K_XL`, `UD-Q4_K_M`, `Q4_K_XL`, then `Q4_K_M`).
- Plans HF Jobs as CPU and I/O bound splitter work; hardware does not need enough VRAM or RAM to hold the full model.
- Uses live HF Jobs hardware pricing to estimate a maximum cost from the selected CPU flavor and effective timeout.
- Automatically raises requested timeouts by model size for cost planning.
- Prints the HF Jobs console URL after confirmed submission.
- Supports JSON output for package planning and job-management commands.
- Uses repo-consistent HF auth/cache behavior (`HF_TOKEN` or `HUGGING_FACE_HUB_TOKEN`, plus the shared HF cache dir).
- Redacts secrets from dry-run output while passing HF Jobs secrets through the Jobs API for runtime injection.
- Updates the bucket splitter script automatically during confirmed submissions if the embedded script and bucket script differ.
- Drains `--logs` without blocking forever on running jobs; follow mode reconnects if the HF log stream closes before terminal status.

## Scheduled Unsloth Queue

This PR adds a daily GitHub Actions workflow, `Queue Unsloth layer packages`, that compares recent and popular `unsloth/*-GGUF` repos, selects preferred 4-bit quants, filters out models that fit comfortably in 8 GiB VRAM using the runtime 10% headroom rule, skips packages already present in `meshllm/catalog`, and queues remaining split candidates by selected quant size descending with a one-per-family default.

Workflow defaults:

- `max_jobs`: `5`
- `confirm`: `false` (dry run)
- `mesh_llm_ref`: `main`
- `quant_preference`: `UD-Q4_K_XL,UD-Q4_K_M,Q4_K_XL,Q4_K_M`
- package job hardware: `auto`, currently choosing the CPU splitter baseline (`cpu-upgrade`) when available
- requested package timeout: `1h`, raised automatically by model size for max-cost estimation

When `confirm` is true, the workflow passes `--confirm`, submits at most the requested number of HF Jobs, monitors every submitted job, and fails unless every job completes successfully.

A sample dry run with `--max-jobs 1 --recent-limit 3 --popular-limit 3` selected:

```text
unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL -> meshllm/Qwen3.6-35B-A3B-UD-Q4_K_XL-layers
hardware=cpu-upgrade (8 vCPU, 32 GB), timeout=4h, max_cost=$0.12 USD
```

## Catalog And Publishing

The embedded HF job script builds `skippy-model-package`, writes and validates the layer package, publishes package artifacts to the target model repo, updates `meshllm/catalog`, and uploads a model card. Catalog updates preserve the expected entry schema and key variants by the resolved GGUF distribution/variant name.

## In-Repo Skill

Adds `.agents/skills/hf-layer-package-jobs/SKILL.md` so future agents have the packaging workflow, dry-run/confirm rules, CPU-only splitter planning assumptions, and validation commands close to the code.

## Required GitHub Secret

Configure a repository or environment Actions secret named `HF_TOKEN` before running the workflow with `confirm: true`. It should be a Hugging Face token for an account that can run HF Jobs and has write access to the `meshllm` namespace, the `meshllm/catalog` dataset, and the target layer-package model repos. Dry runs do not submit HF Jobs.

## Validation

```text
cargo fmt --all -- --check
cargo test -p model-package
cargo check -p mesh-llm-host-runtime
cargo run -p model-package --bin queue-unsloth-layer-packages -- --max-jobs 1 --recent-limit 3 --popular-limit 3 --dry-run
```


crates/model-prepare/src/scripts/split-model-job.sh
```diff
@@ -0,0 +1,335 @@
...
+try:
+    existing_path = api.hf_hub_download(
+        repo_id=catalog_repo,
+        filename=entry_path,
+        repo_type="dataset",
+    )
+    entry = json.load(open(existing_path))
+except Exception:
+    # Create new entry
+    entry = {"schema_version": 1, "variants": []}
```
i386: [P1] New catalog entries are written with the wrong schema

When the source repo is not already in `meshllm/catalog`, this creates `{schema_version, variants: []}` and omits the required `source_repo` field. Current catalog parsing expects `variants` to be a map and `source_repo` to be present, so the PR path for adding a new model would publish an entry mesh-llm cannot load.

crates/model-prepare/src/scripts/split-model-job.sh
```diff
@@ -0,0 +1,335 @@
...
+
+# Build variant name from source file path
+variant_name = model_id or source_file.split('/')[-1].replace('.gguf', '')
```
i386: [P1] Catalog updates key variants by MODEL_ID instead of the GGUF variant

`MODEL_ID` is always set by the Rust submitter, so this never falls back to the source file stem. For `unsloth/Qwen3-8B-GGUF --quant Q4_K_M`, the script updates/creates a variant like `unsloth/Qwen3-8B-GGUF` instead of `Qwen3-8B-Q4_K_M`, so existing catalog variants are missed and selector lookups will not line up with the generated package.

crates/mesh-llm-host-runtime/src/cli/commands/model_package.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+    if let Some(created) = &info.created_at {
+        eprintln!("Created: {created}");
+    }
+    Ok(())
+}
+
+async fn run_logs(client: &HfJobsClient, job_id: &str) -> Result<()> {
+    let (namespace, id) = parse_job_id(job_id).await?;
+    let mut stream = std::pin::pin!(client.stream_logs(&namespace, &id).await?);
```
i386, line 368: [P2] `--logs` can block forever on running jobs

This calls the SSE log stream with no non-follow timeout or exit condition. Hugging Face's own Jobs client treats non-follow logs as a short drain because the endpoint keeps the connection alive waiting for future events; here `mesh-llm model-prepare --logs <running-job>` may never return instead of just fetching the buffered logs.

i386 (commented):
@michaelneale I found three issues while reviewing this. Dropping them inline so the fixes are easier to track.

i386:
@michaelneale trying this out myself. I prefer the `mesh-llm models package $model` nesting of the command over a top level `model-prepare`

i386:
also this needs to use our hf coordinates

`unsloth/Qwen3-8B-GGUF:Q4_K_M` not `unsloth/Qwen3-8B-GGUF --quant Q4_K_M`

i386:
Few things:

1. I want to see the max $ cost to be paid like we did on https://github.com/Mesh-LLM/mesh-llm/pull/330
2. make the default a dry run so you can determine if you want to spend the $ and add a --confirm switch to actually submit the job
3. https://github.com/Mesh-LLM/mesh-llm/pull/330 used to check the spot price and find the cheapest hardware that could fit the model and get the packaging done in a decent amount of time. can we look at bringing that back too?
3. `Bucket script is out of date` no idea what that is or why I should care. can it just automatically update it if it needs to?
4. add json output for `mesh-llm model-prepare` commands
5. move `mesh-llm model-prepare` and its sub commands to `mesh-llm models package`
6. add an in repo skill for this (low priority)

```
jdumay@mac mesh-llm % ./target/release/mesh-llm model-prepare unsloth/Kimi-K2-Thinking-GGUF --quant UD-Q4_K_XL
⚠ Bucket script is out of date (bucket: 12419 bytes, expected: 12785 bytes)
  Run: mesh-llm model-prepare --update-script

🔑 Checking permissions...
🔍 Resolving source...
   Repo:   unsloth/Kimi-K2-Thinking-GGUF
   File:   UD-Q4_K_XL/Kimi-K2-Thinking-UD-Q4_K_XL-00001-of-00014.gguf (00014 shards)

🔑 Permissions: jamesdumay (meshllm org member)
   Target:  meshllm/Kimi-K2-Thinking-UD-Q4_K_XL-layers
   Catalog: meshllm/catalog (direct commit)

📋 Job: cpu-xl, timeout 3h, mesh-llm@main

🚀 Submitted: 69fd25c6aff1cd33e8f3019a
   Status:  mesh-llm model-prepare --status 69fd25c6aff1cd33e8f3019a
   Logs:    mesh-llm model-prepare --logs 69fd25c6aff1cd33e8f3019a
```

i386:
Things seem to be splitting nicely on this run

```
  Building skippy-model-package...
   Compiling typed-builder v0.20.1
   Compiling fs4 v0.13.1
   Compiling hyper v1.9.0
   Compiling hyper-util v0.1.20
   Compiling rustls-webpki v0.103.10
   Compiling tokio-rustls v0.26.4
   Compiling rustls-platform-verifier v0.6.2
   Compiling hyper-rustls v0.27.7
   Compiling reqwest v0.13.2
   Compiling reqwest-middleware v0.5.1
   Compiling xet-runtime v1.5.1
   Compiling reqwest-retry v0.9.1
   Compiling xet-core-structures v1.5.1
   Compiling xet-client v1.5.1
   Compiling xet-data v1.5.1
   Compiling hf-xet v1.5.1
   Compiling huggingface-hub v0.0.1 (https://github.com/huggingface/huggingface_hub_rust.git?rev=c9749ef76a294d5e97acae6b644071e39854aeb8#c9749ef7)
   Compiling model-hf v0.1.0 (/tmp/build/crates/model-hf)
   Compiling skippy-model-package v0.1.0 (/tmp/build/crates/skippy-model-package)
    Finished `release` profile [optimized] target(s) in 1m 14s
  ✓ Built: /tmp/build/target/release/skippy-model-package

=== [4/7] Splitting model ===
  Source: /source/UD-Q4_K_XL/Kimi-K2-Thinking-UD-Q4_K_XL-00001-of-00014.gguf (46G)
  ```

**Comment by i386** [comment]
Sometimes this command just stops but the job hasn't finished


`./target/release/mesh-llm model-prepare --logs 69fd25c6aff1cd33e8f3019a`

**Comment by i386** [comment]
@michaelneale it should also output a link to the Huggingface job console

**Comment by i386** [comment]
Addressed the outstanding review threads in the latest pushes:

- Renamed the helper crate from `model-prepare` to `model-package`, including workspace, Docker, workflow, and dependency references.
- Kept the user command under `mesh-llm models package`, with the old top-level `model-prepare` retained only as a hidden compatibility command.
- Aligned Jobs API token lookup with the repo’s HF token helper (`HF_TOKEN` / `HUGGING_FACE_HUB_TOKEN`) and the shared HF cache directory.
- Confirmed against the installed `huggingface_hub` Jobs implementation that `secrets` is a value map in the Jobs REST spec; added an explicit code comment and kept dry-run output redacted.
- Reworked bucket script checks/uploads to use per-run temp dirs and added mismatch reasons for size vs SHA differences.
- Changed non-follow `--logs` to drain buffered logs with an idle timeout, and made follow mode reconnect if the HF log stream ends before the job is terminal.
- Added the HF job console URL after submission.
- Made the job script checkout work with branch, tag, and commit refs.
- Added unit coverage for quant grouping and shard first-file selection.

Validation run locally:

```text
cargo fmt --all -- --check
cargo test -p model-package
cargo check -p mesh-llm-host-runtime
cargo run -p model-package --bin queue-unsloth-layer-packages -- --max-jobs 1 --recent-limit 3 --popular-limit 3 --dry-run
```

IvGolovach (commented):
I took a look with the roadmap in mind, and honestly I’m really excited to see this coming together.

From my outside-ish perspective, this feels like one of those quiet but important pieces that makes the Skippy work much more real for people. The roadmap talks about split-job workflow, smoother downloads/orchestration, automating popular model splits, and eventually cloud/wakeable capacity. This PR seems to connect directly to that: it takes something that could stay as a hand-run expert workflow and turns it into something a normal mesh-llm user can actually reach for :) 

A few things I really liked:

- `mesh-llm models package ...` feels much more discoverable than a separate top-level command.
- Dry-run by default + explicit `--confirm` is a really nice safety choice for anything that can spend money or publish artifacts.
- Showing the max cost and HF Jobs console URL makes the workflow feel much less mysterious.
- The log handling improvements matter a lot in practice. Long-running jobs are already stressful enough without the CLI silently hanging or disappearing.
- Auto-updating the bucket script on confirmed runs removes a bit of “tribal knowledge” from the flow.
- The catalog/package work feels aligned with the bigger layer-package direction, not just bolted on.
- I also like the in-repo skill. This is exactly the kind of workflow future contributors or agents would otherwise have to rediscover from scratch.

On the earlier `secrets` discussion: from what I can tell, the value-map shape looks intentional for the current HF Jobs behavior/docs, and the important practical bit is that dry-run output stays redacted. So I personally wouldn’t treat that as a blocker.

The only reason I wouldn’t hit approve quite yet is just that the latest CI run is still in flight. I saw the branch got updated again, so I’d wait for that to come back green.

One small thing I’d still consider cleaning up before merge: `docs/plans/HF_SPLIT_JOB.md` seems to still have a few older-design bits, like the `--quant` examples, “submits immediately / no confirmation” wording, and the older `secrets: {"HF_TOKEN": true}` example. Since this PR is making the workflow first-class, it might be worth syncing that doc with the final CLI shape so people don’t get confused later.

Anyway, really nice work. This makes the Skippy/package path feel a lot less like specialist machinery and a lot more like something the project can build on.

i386:
This is working well now. I split a Gemma from unsloth and it works https://huggingface.co/meshllm/gemma-4-26B-A4B-it-UD-Q4_K_M-layers

just need to update the script so it doesn't use hf:// in the examples then it can be merged.

michaelneale:
thanks @IvGolovach yeah - could delete that WIP design doc now I think. 
and yeah the jobs get the secret when running so they can push to the appropriate place on HF itself (from HF)

michaelneale:
@i386 :blessed: when ready

PR #456, opened 2026-05-07
feature(wire-ui): tweak, fix, wire all existing preview-ui to API

ndizazzo:
image

test with `just ui-preview-public`

This is not really reviewable by hand - lots of ~little~ big corrections were needed to make the API and the new frontend work. Highly suggest agents review code instead, and human testing of UI.


crates/mesh-llm/ui-preview/src/features/chat/api/mesh-connection.ts
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  if (!response.body) {
+    throw new Error('Response body is null')
+  }
+
+  let messageStarted = false
+  let firstDeltaAt: number | undefined
+
+  for await (const event of parseSSEStream(response.body, abortSignal)) {
```
i386, line 201: @ndizazzo The backend frames streaming failures as `event: error` with an OpenAI error body before `[DONE]`, but this parser strips the `event:` line and only yields JSON `data:` objects. Since the run loop only handles `response.output_text.delta` and `response.completed`, those error bodies are ignored and the adapter emits `RUN_FINISHED` instead of surfacing `chat.error`. A model/router failure during streaming will look like an empty successful response and leave the composer without the retry/error path.

crates/mesh-llm/ui-preview/src/lib/test/setup.ts
```diff
@@ -13,3 +13,13 @@ class TestResizeObserver implements ResizeObserver {
 if (typeof globalThis.ResizeObserver === 'undefined') {
   globalThis.ResizeObserver = TestResizeObserver
 }
+
+Object.defineProperty(HTMLElement.prototype, 'scrollIntoView', {
```
i386, line 78: @ndizazzo `npm test` currently fails before exercising most assertions because `window.localStorage` has no `getItem`/`setItem`/`removeItem`/`clear` methods under the Node 25 runtime used here. The new setup file adds DOM shims but never replaces Node experimental placeholder with a real test Storage implementation or configures jsdom to provide one, so 129 tests fail immediately across chat, app tabs, preferences, and data-mode suites. Add a localStorage shim in setup or set jsdom environment options so the Storage API is valid before tests run.

PR #457, opened 2026-05-07
Verify Skippy layer packages before stage load

IvGolovach:
## Summary

Skippy layer packages resolved from `hf://` are now verified before a stage can use them. Cached Hugging Face snapshots no longer bypass package integrity checks, and stage status keeps source model identity tied to the package manifest instead of exposing local cache paths as the source model.

## Why

Layer packages are now part of the runtime path for distributed Skippy serving. That makes package provenance a runtime safety boundary, not only a packaging-tool concern.

Before this change, package consumers validated manifest shape, required files, and sizes, but cached `hf://` package snapshots could be reused without rechecking selected artifact SHA-256 values. That left a practical gap: a corrupted or tampered cached layer file could still be selected as long as it existed and had the expected size.

This PR closes that gap while keeping local development workflows usable: remote `hf://` package refs get mandatory selected-artifact verification, while direct local package directories keep checksum verification opt-in through the existing environment path.

## Diff Scope

- `crates/skippy-runtime/src/package.rs`
  - Adds explicit package integrity options and a structured verification report.
  - Verifies selected layer/shared artifacts and declared projectors against manifest SHA-256 values.
  - Adds a private verification cache keyed by package/artifact identity and file metadata, without storing raw local package paths.
  - Tightens manifest validation for source SHA-256 fields, unsafe artifact paths, duplicate layers, missing layers, and metadata-only inspection.

- `crates/mesh-llm/src/inference/skippy/materialization.rs`
  - Runs the integrity gate for `hf://` package resolution, including direct snapshot cache hits and ref-resolved snapshot cache hits.
  - Keeps verification errors tied to manifest-relative artifact paths.
  - Carries safe source model identity from the package manifest after resolution.

- `crates/mesh-llm/src/inference/skippy/stage/mod.rs`
  - Preserves resolved local package refs internally for loading.
  - Uses manifest source identity for stage config/status instead of reporting the local cache path as `source_model_path`.

- `crates/mesh-llm/src/inference/skippy/stage/tests.rs`
  - Adds coverage for package source identity precedence over local package refs.

- `docs/specs/layer-package-repos.md`
  - Documents the consumer-side integrity requirement for `hf://` package refs and the privacy expectation for verification cache records.

## Branch Integrity

- Base branch: `main`
- Validated base SHA: `7ea662e0e5242cdfffe5e9887207671cb19218b1`
- Ahead count: `1`
- Behind count: `0`
- Merge-base SHA: `7ea662e0e5242cdfffe5e9887207671cb19218b1`
- Fast-forward safety: `origin/main` is an ancestor of `HEAD`; no divergence.

## Commit Integrity

Introduced commit:

- `3f933628079dfa126b7c8f4d3dfb71a25dadf5e1` — `Gate Skippy layer packages with integrity verification`

One logical change: confirmed. The diff is scoped to Skippy package integrity, `hf://` package resolution, stage identity handling, focused tests, and the matching layer-package spec.

## Ledger and Version Proof

Ledger: not applicable — `scripts/ledger/` does not exist in this repository.

Version: not applicable — this is not a release commit, release metadata did not change, and repository release versioning remains handled by the existing release flow.

## Diff Hygiene

```text
M	crates/mesh-llm/src/inference/skippy/materialization.rs
M	crates/mesh-llm/src/inference/skippy/stage/mod.rs
M	crates/mesh-llm/src/inference/skippy/stage/tests.rs
M	crates/skippy-runtime/src/package.rs
```

## Validation Mode and Proof

Validation mode: Mode 3 — shared backend/runtime behavior. The change affects package validation and Skippy stage-load behavior, but does not change wire protocol, migrations, auth, CI, deployment tooling, or UI assets.

Local validation:

```text
cargo fmt --all -- --check
PASS
```

```text
LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p skippy-runtime --lib
PASS — 17 passed, 0 failed
```

```text
LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm inference::skippy --lib
PASS — 68 passed, 2 ignored, 0 failed
```

```text
LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo check -p mesh-llm
PASS
```

Not run:

- `just build` — not required for the selected validation mode because this is a Rust/docs-only change and no UI, bundle, release, or llama.cpp patch path changed.
- Full `cargo test -p mesh-llm` — not required locally for this scoped Skippy runtime change; targeted runtime and Skippy subsystem tests passed, and mandatory remote CI should provide final full-branch proof after PR creation.

## Verification-Pack Proof

Not applicable — no ops, governance, replay, release, or verification-pack tooling changed.

## Migration Notes

Not applicable — no database schema, migration, persisted data model, or data repair path changed.

## Runtime Safety

Reviewed runtime paths:

- `crates/skippy-runtime/src/package.rs`
- `crates/mesh-llm/src/inference/skippy/materialization.rs`
- `crates/mesh-llm/src/inference/skippy/stage/mod.rs`

Runtime safety notes:

- No new blocking locks.
- No new queues, channels, background workers, or buffering paths.
- No mesh protocol, protobuf, or plugin protocol changes.
- Package manifest schema remains version `1`; this is consumer-side validation hardening.
- First use of a large `hf://` package may hash selected artifacts before stage load; repeat resolutions use the verification cache when file identity is unchanged.
- No invariant removal: existing manifest/file/size checks remain, with checksum and path-safety coverage added around them.

## Documentation Integrity

Updated `docs/specs/layer-package-repos.md` to match the new consumer-side behavior for `hf://` packages and to document that verification cache records should avoid raw local package paths.

No README, runbook, CLI command examples, deployment instructions, or release procedure changed.

## Rollback Plan

Rollback: revert this PR.

```bash
```

DB downgrade required: no.
Data repair required: no.
Operational caveats: rollback restores the previous package consumer behavior, where cached `hf://` package snapshots are not forced through selected-artifact checksum verification before stage load.

## Known Residual Risks

- Large remote package refs can pay a first-use hashing cost for selected artifacts and projectors. The verification cache is intended to keep repeat resolutions cheap without storing raw local paths.

### DIFF
--- a/crates/mesh-llm/src/inference/skippy/materialization.rs
+++ b/crates/mesh-llm/src/inference/skippy/materialization.rs
@@ -7,7 +7,9 @@ use anyhow::{bail, Context, Result};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use skippy_protocol::{LoadMode, StageConfig};
-use skippy_runtime::package::{self, LayerPackageInfo, PackageStageRequest};
+use skippy_runtime::package::{
+    self, LayerPackageInfo, PackageIntegrityOptions, PackageStageRequest,
+};
 
 use super::StageLoadRequest;
 
@@ -97,6 +99,14 @@ pub(crate) struct StagePackageLayerInfo {
 pub(crate) struct MaterializedStageArtifact {
     pub(crate) path: PathBuf,
     pub(crate) manifest_sha256: String,
+    pub(crate) source_model_path: String,
+    pub(crate) source_model_sha256: String,
+    pub(crate) source_model_bytes: Option<u64>,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct ResolvedStagePackage {
+    pub(crate) local_ref: String,
     pub(crate) source_model_path: String,
     pub(crate) source_model_sha256: String,
     pub(crate) source_model_bytes: Option<u64>,
@@ -216,6 +226,49 @@ fn resolve_local_package_files(
     Ok(package_dir.to_string_lossy().to_string())
 }
 
+fn package_integrity_cache_dir() -> PathBuf {
+    crate::models::mesh_llm_cache_dir().join("skippy-package-integrity")
+}
+
+fn verify_resolved_hf_package_files(
+    package_dir: &Path,
+    layer_start: u32,
+    layer_end: u32,
+    include_embeddings: bool,
+    include_output: bool,
+) -> Result<String> {
+    let local_ref = resolve_local_package_files(
+        package_dir,
+        layer_start,
+        layer_end,
+        include_embeddings,
+        include_output,
+    )?;
+    let request = PackageStageRequest {
+        model_id: "hf-layer-package".to_string(),
+        topology_id: "hf-layer-package-resolver".to_string(),
+        package_ref: local_ref.clone(),
+        stage_id: format!("layers-{layer_start}-{layer_end}"),
+        layer_start,
+        layer_end,
+        include_embeddings,
+        include_output,
+    };
+    let report = package::verify_layer_package_integrity(
+        &request,
+        &PackageIntegrityOptions::verify_with_cache(package_integrity_cache_dir()),
+    )
+    .map_err(|error| anyhow::anyhow!("verify resolved HF layer package artifacts: {error:#}"))?;
+    tracing::debug!(
+        artifacts = report.artifacts,
+        verified_artifacts = report.verified_artifacts,
+        cached_artifacts = report.cached_artifacts,
+        manifest_sha256 = %report.manifest_sha256,
+        "verified resolved HF layer package artifacts"
+    );
+    Ok(local_ref)
+}
+
 pub(crate) fn resolve_hf_package_to_local(
     package_ref: &str,
     layer_start: u32,
@@ -257,7 +310,7 @@ pub(crate) fn resolve_hf_package_to_local(
         .join("snapshots")
         .join(&revision_cache_path);
     if direct_snapshot_dir.join("model-package.json").is_file() {
-        return resolve_local_package_files(
+        return verify_resolved_hf_package_files(
             &direct_snapshot_dir,
             layer_start,
             layer_end,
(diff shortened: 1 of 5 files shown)


crates/mesh-llm/src/inference/skippy/materialization.rs
```diff
@@ -216,6 +226,49 @@ fn resolve_local_package_files(
...
+        model_id: "hf-layer-package".to_string(),
+        topology_id: "hf-layer-package-resolver".to_string(),
+        package_ref: local_ref.clone(),
+        stage_id: format!("layers-{layer_start}-{layer_end}"),
+        layer_start,
+        layer_end,
+        include_embeddings,
+        include_output,
+    };
+    let report = package::verify_layer_package_integrity(
```
i386: @IvGolovach Metadata-only HF package inspection now fails here. `inspect_stage_package()` and `identity_from_layer_package()` call `resolve_hf_package_to_local(..., 0, 0, false, false)` to inspect only the manifest/metadata for an `hf://` package. This new wrapper always calls `verify_layer_package_integrity()`, which goes through `validate_manifest()` and rejects `layer_start >= layer_end`, so every metadata-only HF inspection now errors with `stage layer_start must be less than layer_end`. That breaks inventory/prepare and package identity discovery before a stage even gets a real layer range. The verifier needs a metadata-only path, or this wrapper should skip stage-range verification when `layer_start == layer_end`.

IvGolovach: Thanks for catching this- you were exactly right.

I went back through the metadata-only callers and fixed this as a separate integrity scope rather than loosening the real stage-load path. Metadata/identity discovery for `hf://` packages can now validate the manifest plus shared metadata artifact without requiring a non-empty layer range, while actual stage loads still require the non-empty range and full selected-artifact verification.

I also added regression coverage for both sides of that boundary: metadata-only inspection no longer fails on `0..0`, and the metadata artifact itself is still checksum-verified so cached `hf://` snapshots do not bypass integrity checks. Really appreciate the sharp review here — this was a subtle one. 🙂

PR #458, opened 2026-05-07
Share Skippy package artifacts between split peers

IvGolovach:
## Summary

Skippy split peers can now fetch authorized layer-package artifacts from another mesh node instead of requiring every serving box to have direct Hugging Face access for the same package files.

This revision also folds the follow-up review feedback into the transport shape: mesh owns the generic `STREAM_SUBPROTOCOL` admission/open envelope on `mesh-llm/1`, while Skippy owns only the `skippy-stage/1` subprotocol payload and artifact-transfer frames.

## Why

Layer-package serving should be able to recover from partial local materialization without forcing every worker to redownload the same artifacts independently. That matters for split topologies where one peer already has the required shard/package files and another peer can safely pull them over the mesh.

The transport boundary is intentionally split:

- Mesh validates the generic subprotocol open frame and advertised subprotocol capability.
- Skippy keeps ownership of stage-specific stream kinds, artifact requests, and artifact responses.
- Activation/runtime transport remains on the direct `skippy-stage/1` path; artifact body transfer uses the mesh-owned subprotocol envelope.

## Diff scope

- Added the shared mesh subprotocol stream type and validation in `crates/mesh-llm-protocol`.
- Re-exported the shared protocol surface from `crates/mesh-client` after the crate-decomposition merge.
- Advertised supported mesh subprotocols through peer announcements.
- Routed Skippy artifact-transfer streams through `STREAM_SUBPROTOCOL` over `mesh-llm/1`.
- Kept the legacy inbound direct `skippy-stage/1` artifact-transfer path for rolling/mixed-version peers.
- Preserved Skippy-owned framing for artifact request/response bodies in `skippy-protocol`.
- Kept prefetch awaited inside prepare state, retained per-read body idle timeouts, and guarded partial `.part` files on transfer failure.
- Added/updated focused tests for protocol validation, artifact transfer, authorization, timeout/failure behavior, API catalog fixture serialization, and resolver fixture hermeticity.

## Protocol and compatibility

- `PeerAnnouncement.subprotocols` is additive.
- The old artifact-transfer flag remains reserved instead of being repurposed.
- `MeshSubprotocol` and `MeshSubprotocolOpen` are additive protobuf messages.
- Older peers can ignore the new announcement field.
- Newer peers still accept the legacy direct `skippy-stage/1` inbound artifact-transfer stream, so rolling upgrades do not strand peers that have not picked up the generic mesh subprotocol envelope yet.
- `STAGE_STREAM_TRANSPORT` is explicitly kept off the generic mesh subprotocol path and remains on the direct Skippy stage transport.

## Branch integrity

- Base branch: `main`
- Validated base SHA: `7ffb4b9d9ae8f449cc54bd27f836bdb413b02eef`
- `git rev-list --left-right --count origin/main...HEAD`: `0 9`
- Ahead count: 9
- Behind count: 0
- Merge-base SHA: `7ffb4b9d9ae8f449cc54bd27f836bdb413b02eef`
- Fast-forward safety: `origin/main` is an ancestor of `HEAD`; no divergence.

## Commit integrity

Introduced commits:

```text
46bd5629fc9fb779ff6e3c1970dbc3fcdacc5a19 Add peer package artifact transfer
b6cc3bc2e5c1c4efa1ba4f3748135b930adf26bb Fix artifact transfer clippy warnings
bf70260b0e126d065571c17184fe20f4efbeb18c Harden Skippy artifact transfer boundaries
ddb214a5a9c33e0c5c6142f1ce52197aa0000882 Satisfy clippy for package prefetch fallback
377646abe3f17b958b52b664123a76d46214d6a3 Merge main into p2p artifact transfer
93cc6f2fe3077c95259084d9267dbf5d8465b2dd Route Skippy streams through mesh subprotocol
bfb73b27094867d352865d600a2a1d56df78f4cc Merge main into p2p artifact transfer
767cf4c3e14bb2bf25b49ece3d40e8e4cb936e86 Serialize catalog API search override tests
10745abdb0f5639e8adc63f34b3f2aec9948c7f7 Merge main into p2p artifact transfer
```

The PR contains one logical change: peer-assisted Skippy layer-package artifact transfer, updated across review feedback and the current crate decomposition.

## Ledger/version proof

- Ledger: not applicable — not required for this change family.
- Version: not applicable — not required for this change family.

## Diff hygiene

```text
M	crates/mesh-client/tests/protocol_wire.rs
M	crates/mesh-llm-host-runtime/Cargo.toml
M	crates/mesh-llm-host-runtime/README.md
M	crates/mesh-llm-host-runtime/src/api/tests.rs
M	crates/mesh-llm-host-runtime/src/inference/skippy/materialization.rs
M	crates/mesh-llm-host-runtime/src/inference/skippy/mod.rs
M	crates/mesh-llm-host-runtime/src/inference/skippy/stage/inventory.rs
M	crates/mesh-llm-host-runtime/src/inference/skippy/stage/mod.rs
M	crates/mesh-llm-host-runtime/src/inference/skippy/stage/tests.rs
M	crates/mesh-llm-host-runtime/src/mesh/gossip.rs
M	crates/mesh-llm-host-runtime/src/mesh/mod.rs
M	crates/mesh-llm-host-runtime/src/mesh/tests.rs
A	crates/mesh-llm-host-runtime/src/models/artifact_transfer.rs
M	crates/mesh-llm-host-runtime/src/models/mod.rs
M	crates/mesh-llm-host-runtime/src/models/resolve/tests.rs
M	crates/mesh-llm-host-runtime/src/protocol/convert.rs
M	crates/mesh-llm-host-runtime/src/protocol/mod.rs
M	crates/mesh-llm-host-runtime/src/runtime/local.rs
M	crates/mesh-llm-host-runtime/src/runtime/mod.rs
M	crates/mesh-llm-protocol/proto/node.proto
M	crates/mesh-llm-protocol/src/proto/node.rs
M	crates/mesh-llm-protocol/src/protocol/mod.rs
M	crates/mesh-llm-system/README.md
M	crates/skippy-protocol/README.md
M	crates/skippy-protocol/proto/stage.proto
M	crates/skippy-protocol/src/lib.rs
M	docs/SKIPPY.md
M	docs/design/TESTING.md
M	docs/design/message_protocol.md
M	third_party/llama.cpp/patches/0065-Support-broad-llama-family-runtime-slices.patch
```

- Forbidden-file confirmation: no `.env` files, local environment files, secrets, credentials, tokens, pycache files, build output, coverage output, or unrelated generated files are included in the PR diff.

## Validation mode and proof

Validation mode: Mode 3 — shared backend/protocol runtime change. The diff touches mesh stream admission, protobuf compatibility, Skippy package materialization, and artifact transfer behavior.

Local validation:

```text
rustup run stable cargo fmt --all -- --check
PASS

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo check -p mesh-llm
PASS

rustup run stable cargo test -p mesh-llm-client
PASS

rustup run stable cargo test -p mesh-llm-client --test protocol_wire
PASS, 23 passed

rustup run stable cargo test -p skippy-protocol --lib
PASS, 18 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo test -p mesh-llm-host-runtime mesh_subprotocol_open_roundtrips_and_validates --lib
PASS, 1 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo test -p mesh-llm-host-runtime artifact_transfer_stream_serves_authorized_stage_artifact --lib
PASS, 1 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo test -p mesh-llm-host-runtime artifact_transfer --lib
PASS, 10 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo test -p mesh-llm-host-runtime prepare_layer_package --lib
PASS, 2 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo test -p mesh-llm-host-runtime test_api_search --lib -- --test-threads=2
PASS, 4 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo test -p mesh-llm-host-runtime models::resolve::tests --lib -- --test-threads=1
PASS, 51 passed

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo test -p mesh-llm-host-runtime --lib -- --test-threads=1
PASS, 1220 passed, 5 ignored

LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal rustup run stable cargo test -p mesh-llm --lib
PASS, 0 tests
```

Remote gates:

- Pending — required GitHub checks started after the latest push and must pass before merge.

## Runtime safety

Reviewed runtime paths:

- `crates/mesh-llm-host-runtime/src/mesh/mod.rs`
- `crates/mesh-llm-host-runtime/src/models/artifact_transfer.rs`
- `crates/mesh-llm-host-runtime/src/inference/skippy/materialization.rs`
- `crates/mesh-llm-host-runtime/src/inference/skippy/stage/mod.rs`
- `crates/mesh-llm-protocol/src/protocol/mod.rs`

Safety notes:

- No new unbounded queues were added.
- No invariant was removed.
- Mesh subprotocol opens are length-bounded and validated before dispatch.
- Artifact body reads use per-read idle timeouts.
- Failed or incomplete transfers clean up guarded partial files.
- Stage activation transport is rejected on the mesh subprotocol path and remains on direct Skippy stage transport.

## Documentation integrity

Updated protocol and Skippy docs to match the new artifact-transfer path and the mesh/Skippy ownership split.

## Migration notes

Not applicable — no database migration changed.

## Rollback plan

- Rollback: revert this PR.
- DB downgrade: not applicable.
- Data repair: not applicable.
- Operational caveats: rollback restores the prior behavior where peers without local package artifacts must materialize/download them directly instead of fetching them from a peer.

## Known residual risks

- Required remote CI is pending for the latest pushed SHA.
- A live multi-machine mixed-version Skippy split smoke was not rerun in this local pass. The branch keeps the legacy inbound `skippy-stage/1` artifact-transfer path and validates it in local coverage, but final merge should still wait for required GitHub checks and any maintainer-requested remote split smoke.


i386:
Thanks @IvGolovach, this is super interesting. @michaelneale and I have talked a few times about being able to ship files around the mesh for model artifacts and other uses, including a more capable `blobstore`-style primitive.

A couple of concrete review notes first:

1. **`Prepare` can race peer prefetch and fail early**

   In `crates/mesh-llm/src/mesh/mod.rs`, the `Prepare` path starts peer artifact prefetch in a detached task, then the normal stage preparation task can immediately inspect the layer package, fail because the peer transfer has not completed yet, and mark the stage preparation as `Failed`. The coordinator wait loop treats that failed preparation as terminal, so a slow-but-working peer transfer can cause split startup to fail instead of falling back cleanly.

   I think the transfer should either be part of the prepare task/state machine, or prepare should avoid marking layer-package source resolution failed while peer prefetch is still in progress.

2. **Artifact body reads need an idle timeout**

   The stream open and response header are wrapped in a 30s timeout, but once the response is accepted the body read loop can wait indefinitely if a peer stalls mid-transfer. For large artifacts a total transfer timeout may be awkward, but an idle/read timeout would keep this optimization from hanging stage preparation resources and leaving partial files around until the connection eventually dies.

The larger thing I’m still unsure about is the protocol boundary. This PR describes the feature as Skippy layer-package peer cache transfer, and the authorization model is based on active Skippy topology, assigned layer ranges, manifest SHA, stage 0/final-stage files, etc. That makes me wonder whether this should live in the Skippy stage protocol rather than the mesh control protocol.

The tradeoff I see:

1. **Mesh protocol:** convenient because peers already have admitted QUIC connections and gossip can advertise a simple capability, but it makes Skippy package transfer part of the stable mesh compatibility surface.
2. **Skippy protocol:** seems closer to the ownership boundary, because the transfer is only meaningful for Skippy stage execution and can evolve with `skippy-stage/1` without broadening `mesh-llm/1`.
3. **Separate blob/artifact layer:** possibly the right long-term primitive if we want general file distribution for models, plugins, request blobs, cache warming, and other mesh-owned artifacts. This feels less like a plugin and more like a core mesh primitive: a content-addressed artifact service that other subsystems can use, with its own protocol boundary and compatibility story. That might eventually use BitTorrent/IPFS-like mechanics, or start as a simpler mesh-native transfer protocol, but it should probably be designed as reusable infrastructure rather than a Skippy-specific stream.

My leaning is: keep the immediate Skippy layer-package optimization in the Skippy protocol, and only add a mesh-level artifact protocol once we have a clearer generic blobstore/content-addressed design. If that generic path is where we want to go, I’d treat it as a core mesh primitive rather than a plugin API, since model/package distribution feels foundational to the mesh itself.

i386:
Small clarification on my “Skippy protocol” point: I don’t mean this necessarily needs a completely separate physical transport or QUIC connection.

I think it’s reasonable for Skippy messages to ride over the admitted mesh connection / `mesh-llm` ALPN as custom payloads. The boundary I care about is ownership of the schema and compatibility contract: mesh should provide identity, admission, connectivity, stream muxing, and maybe feature/subprotocol discovery; Skippy should own the stage/artifact message schema, generation, semantics, authorization rules, and compatibility behavior.

One possible shape:

```proto
// Mesh-owned, generic discovery only.
message PeerAnnouncement {
  repeated MeshSubprotocol subprotocols = 37;
}

message MeshSubprotocol {
  string name = 1;             // "skippy-stage"
  uint32 major = 2;            // 1
  repeated string features = 3; // ["artifact-transfer"]
}
```

Then over the existing admitted mesh connection:

```text
[stream type: STREAM_SUBPROTOCOL]
[len][MeshSubprotocolOpen { name: "skippy-stage", major: 1 }]
[len][opaque skippy-protocol protobuf bytes...]
[optional raw bytes owned by skippy-protocol...]
```

The artifact request itself would stay in `skippy-protocol`, for example:

```proto
// Skippy-owned protocol surface.
message SkippyArtifactTransferRequest {
  uint32 gen = 1; // skippy protocol generation
  string topology_id = 2;
  string run_id = 3;
  string stage_id = 4;
  string package_ref = 5;
  string manifest_sha256 = 6;
  string relative_path = 7;
  uint64 offset = 8;
}
```

So the flow would be:

1. Gossip advertises “I support `skippy-stage/1` with `artifact-transfer`.”
2. Worker sees the coordinator supports that feature.
3. Worker opens a generic subprotocol/custom-payload stream over the already-admitted mesh connection.
4. Mesh validates admission and subprotocol registration, then hands the stream to Skippy.
5. Skippy decodes the request, checks topology/layer authorization, and owns the response/body framing.

That keeps mesh responsible for the mesh things, while avoiding a first-class `mesh-llm/1` `ArtifactTransferRequest` whose semantics are currently Skippy-specific.

i386:
Thinking about this a bit more, there are probably useful places to apply this boundary to the way Skippy works today.

The cleanest near-term fit seems to be Skippy stage control over admitted mesh transport. Messages like:

- `PrepareStage`
- `LoadStage`
- `GetLayerInventory`
- `GetStageStatus`
- `CancelPrepareStage`
- `StageStatusUpdate`

are already Skippy-owned concepts. Mesh could gossip that a peer supports `skippy-stage/1` plus feature bits like `stage-control` and `artifact-transfer`, then carry those Skippy protobufs over a generic mesh subprotocol/custom-payload stream. Mesh still owns identity, admission, connectivity, and muxing; Skippy keeps owning the protobuf schema, generation, and semantics.

Artifact transfer then becomes a natural part of `PrepareStage`: while preparing, a worker can discover that package artifacts are missing and use a Skippy-owned artifact-transfer message to fetch them from the coordinator. That also avoids the race where peer prefetch is detached from the prepare state machine, because the transfer can be represented directly as `Preparing/Downloading/Available/Failed` state.

I would *not* use gossip itself for the Skippy messages or file movement. Gossip should just advertise support/discovery, e.g. “this peer supports `skippy-stage/1` with `artifact-transfer`.” The actual messages and bytes should be unicast over an admitted stream.

I’d also be cautious about activation traffic. Skippy activation bytes are hot-path and latency-sensitive, so I’d keep the current stage transport shape unless/until there’s a clear reason to unify it. Control, inventory, prepare, and artifact fetch feel like the safer first targets for a generic mesh subprotocol envelope.

So an incremental path could be:

1. Gossip advertises `skippy-stage/1` and feature bits.
2. Mesh adds a generic `STREAM_SUBPROTOCOL` / custom payload stream.
3. Existing Skippy stage control messages ride through that stream unchanged.
4. Artifact transfer becomes a Skippy-owned message used during `PrepareStage`.
5. Activation transport stays separate for now.

IvGolovach:
Thanks a lot for the careful review — this was exactly the right set of boundaries to tighten. 🙂

I went back through all three points and changed the implementation to make the ownership model much cleaner:

- The mesh protocol now only advertises admitted subsystem capability through `MeshSubprotocol` discovery.
- The artifact transfer request/response schema moved into `skippy-protocol`, alongside the rest of the Skippy stage contract.
- The transfer stream now runs under `skippy-stage/1` as a Skippy-owned stream, instead of adding a mesh-owned artifact stream.
- `Prepare` now waits on peer package prefetch inside the stage preparation state machine, so it cannot race a detached background fetch and fail early.
- Artifact body reads now have an idle timeout, and failed/partial downloads clean up their `.part` files instead of leaving stale cache state behind.

I also added focused coverage for the protocol boundary, the prepare/prefetch race, transfer authorization scope, body idle timeout, and partial-file cleanup. The local Rust validation is green.

michaelneale:
I think this is worthwhile and directionally correct - as long as it doesn't break it, but not all boxes may have a HF_TOKEN, so if it can get things from a peer, why not? 
Can use quic tunnels to open a connection and pull things via any protocol it wants really.

i386:
Reviewing again - thanks @IvGolovach !

i386:
Follow-up review pass on the earlier feedback:

| Feedback | Status |
|---|---|
| `Prepare` race with peer prefetch | **Done.** Prefetch is now awaited inside the prepare task/state path, with `Downloading` status while pending and failure only after prefetch plus local resolution fail. |
| Artifact body idle timeout | **Done.** Body reads now go through a per-read idle timeout, and partial `.part` files are guarded/removed on failure. |
| Move artifact schema out of mesh protocol | **Done.** `node.proto` reserves the old artifact flag and advertises `MeshSubprotocol`; artifact request/response live in `skippy-protocol`. |
| Keep transfer Skippy-owned | **Mostly done.** Transfer uses Skippy-owned frames and `STAGE_STREAM_ARTIFACT_TRANSFER` under `skippy-stage/1`. |
| Exact suggested generic `STREAM_SUBPROTOCOL` over `mesh-llm/1` | **Not done exactly.** The PR chose a separate `skippy-stage/1` ALPN/stream-kind path instead of adding a generic mesh subprotocol envelope. This still satisfies the main ownership concern, but not the exact shape suggested in the comment. |

I’d call the earlier review feedback handled at the code/design level. Remaining caveats before merge: CI is still pending/unstable right now, and the PR’s commit notes say two-node mixed-version Skippy transfer validation was not run.

PR #465, opened 2026-05-08
task(swap-ui-preview): add new UI into main, old to ui-legacy

ndizazzo:


michaelneale:
might be an easier way so diff is less diff-ey?

PR #473, opened 2026-05-08
Harden Skippy stage control recovery

IvGolovach:
## Summary

Skippy split-stage serving now has a more reliable control plane for cancelled prepares and failed active stages.

This PR preserves full stage-status visibility for capable peers, makes prepare cancellation mutate real local state, applies status updates to inventory, and lets split topology recovery react when an active stage is `Failed` or `Stopped` even if the peer is still present.

## Why

This closes concrete Skippy orchestration gaps from the short-term roadmap: stage control should not lose multi-stage status evidence, cancelled prepare work should not be overwritten by late background tasks, and split recovery should not wait for a peer to disappear before recovering from a failed active stage.

## What Changed

- Added additive `StageStatusList` support to stage-control responses.
- Advertised the new `status-list` capability through existing `skippy-stage/1` subprotocol metadata.
- Kept mixed-version compatibility by sending legacy singular `StageStatus` responses to peers that do not advertise `status-list`.
- Persisted `CancelPrepare` as real `Cancelled` preparation state and aborts the tracked prepare task.
- Rejected stale shutdown-generation cancels and stale late status writes.
- Made `StatusUpdate` upsert preparation inventory with validation and explicit ack errors.
- Extended split recovery to treat `Failed`/`Stopped` active stage status as unavailable capacity, reusing the existing replacement split, local fallback, and withdraw paths.

## Protocol

Additive only:

- New protobuf field: `StageControlResponse.stage_statuses = 8`.
- New feature flag: `status-list` under `skippy-stage/1`.
- No existing fields, stream kinds, or legacy behavior are removed or repurposed.

## Validation

Base: `main` at `8d12c0be26fb3af4ed309fde6df65acfabff0162`  
Head: `e13559b923c08f6ea083ae95c9e1b7b214f7a4b4`  
Ahead/behind: `0 1`  
Merge-base: `8d12c0be26fb3af4ed309fde6df65acfabff0162`  
Fast-forward safety: `origin/main` is an ancestor of `HEAD`.

```text
cargo fmt --all -- --check: PASS
rustup run stable cargo clippy -p mesh-llm -- -D warnings: PASS
cargo test -p skippy-protocol --lib: PASS, 18 passed
cargo test -p mesh-llm-host-runtime inference::skippy::stage::tests --lib: PASS, 16 passed
cargo test -p mesh-llm-host-runtime stage_control --lib: PASS, 9 passed
cargo test -p mesh-llm-host-runtime split_loss_recovery --lib: PASS, 6 passed
cargo test -p mesh-llm-host-runtime split_unavailable_active_stage_nodes --lib: PASS, 1 passed
cargo check -p mesh-llm: PASS
```

The Rust commands above used:

```text
LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama.cpp/build-stage-abi-metal
```

Broader local-suite note: `cargo test -p mesh-llm-host-runtime --lib` is blocked in this local environment by `runtime::tests::initial_pretty_session_mode_allows_dashboard_only_for_serve_surface`. The same exact test also fails on clean `origin/main` at the validated base SHA with the same `InteractiveDashboard` vs `None` assertion, so this is treated as a baseline/local-environment caveat rather than a regression from this diff.

Remote CI: PASS on the updated head SHA. `CI` completed successfully, including `linux`, `macos`, `Linux Vulkan`, `Linux CUDA slim`, `Linux ROCm slim`, and Skippy inference smoke tests. `docker-client` also completed successfully.

## Runtime Safety

- No new unbounded queues or broad background pipelines were added.
- Prepare tasks are now tracked so cancel/replacement paths can stop stale work explicitly.
- Cancelled or newer generations are guarded against late background writes.
- Existing split recovery outcomes are reused; the health signal is expanded from missing peers to missing-or-failed active stages.

## Rollback

Rollback: revert this PR.  
DB downgrade: not applicable.  
Data repair: not applicable.  
Operational caveats: rollback restores the previous stage-control status, prepare-cancel, and failed-stage recovery behavior.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/inference/skippy/stage/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/inference/skippy/stage/mod.rs
@@ -1,4 +1,12 @@
-use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
+use std::{
+    collections::HashMap,
+    net::SocketAddr,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
 
 use anyhow::{anyhow, Context, Result};
 use skippy_protocol::{FlashAttentionType, LoadMode, PeerConfig, StageConfig};
@@ -7,7 +15,10 @@ use skippy_server::{
     telemetry::TelemetryLevel,
     EmbeddedServerHandle,
 };
-use tokio::sync::{mpsc, Mutex};
+use tokio::{
+    sync::{mpsc, Mutex},
+    task::JoinHandle,
+};
 
 mod inventory;
 #[cfg(test)]
@@ -29,7 +40,13 @@ struct RunningStage {
 struct StageControlState {
     stages: HashMap<String, RunningStage>,
     preparations: Arc<Mutex<HashMap<String, StagePreparationStatus>>>,
+    preparation_tasks: HashMap<String, StagePreparationTask>,
     package_prefetcher: Option<Arc<dyn StagePackagePrefetcher>>,
+}
+
+struct StagePreparationTask {
+    cancelled: Arc<AtomicBool>,
+    handle: JoinHandle<()>,
 }
 
 #[async_trait::async_trait]
@@ -73,14 +90,11 @@ impl StageControlState {
                 self.prepare(request).await?,
             )),
             StageControlRequest::CancelPrepare(cancel) => Ok(
-                StageControlResponse::PreparationStatus(preparation_status_from_cancel(cancel)),
+                StageControlResponse::PreparationStatus(self.cancel_prepare(cancel).await),
             ),
-            StageControlRequest::StatusUpdate(_status) => {
-                Ok(StageControlResponse::StatusAck(StageStatusAck {
-                    accepted: true,
-                    error: None,
-                }))
-            }
+            StageControlRequest::StatusUpdate(_status) => Ok(StageControlResponse::StatusAck(
+                self.apply_status_update(_status).await,
+            )),
         }
     }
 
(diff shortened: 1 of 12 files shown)


crates/mesh-llm-host-runtime/src/inference/skippy/stage/mod.rs
```diff
@@ -165,18 +179,95 @@ impl StageControlState {
             .lock()
             .await
             .insert(key.clone(), status.clone());
```
i386: This unconditional insert lets a stale `Prepare` resurrect a stage after a newer/equal `CancelPrepare` has already persisted `Cancelled`. The later `update_preparation()` guard no longer helps because the cancelled row has been overwritten before the background task starts. Please check any existing status here and reject/preserve it when `existing.shutdown_generation >= request.load.shutdown_generation` and the existing state is `Cancelled`.

crates/mesh-llm-host-runtime/src/mesh/mod.rs
```diff
@@ -808,6 +809,7 @@ pub struct PeerInfo {
     pub served_model_runtime: Vec<ModelRuntimeDescriptor>,
     pub owner_attestation: Option<SignedNodeOwnership>,
     pub artifact_transfer_supported: bool,
+    pub stage_status_list_supported: bool,
```
i386, line 812: Now that peer capability tracking has a separate `stage_status_list_supported` field, the peer update paths need to refresh it too. Existing peers created before they advertise this flag never get upgraded because neither the direct gossip update path nor `apply_transitive_ann()` copies `ann.stage_status_list_supported`, and `peer_meaningfully_changed()` also ignores it. That means capable peers can keep receiving legacy singular status responses until the peer entry is recreated.

i386 (commented):
Found two correctness issues in the Skippy stage-control recovery changes.

i386 (changes requested):
After these are addressed, please merge! 👍

PR #475, opened 2026-05-08
Add Flash-MoE SSD backend support

IvGolovach:
## Summary

Users can now run Flash-MoE as a built-in mesh-llm backend for single-node SSD expert streaming: giant MoE models can live on local NVMe while mesh-llm handles backend lifecycle, endpoint discovery, and OpenAI-compatible routing.

Supported modes:

- **Managed process mode**: mesh-llm starts and supervises the Flash-MoE `infer` binary, allocates the local serving port, and appends `--serve <port>` itself.
- **Existing endpoint mode**: mesh-llm attaches an already-running Flash-MoE `/v1` endpoint and advertises it through the existing plugin inference path.

## Why

This follows the roadmap direction for SSD expert streaming without pushing SSD streaming into llama.cpp internals and without changing Skippy, model packages, or mesh protocol compatibility.

## Diff Scope

- Added built-in `flash-moe` plugin support in `mesh-llm-host-runtime`.
- Registered the plugin in built-in dispatch and plugin resolution.
- Added plugin-scoped env wiring for built-in backend adapters.
- Added config validation: exactly one of `command` or `url`, `args` only with `command`, and no user-provided `--serve` because mesh-llm owns the port.
- Added docs in `README.md`, `docs/plugins/flash-moe.md`, `docs/plugins/README.md`, `ROADMAP.md`, and `crates/mesh-llm/TODO.md`.
- Updated one stale host-runtime test expectation to match the existing dashboard-mode behavior.

## Architecture / Protocol

Flash-MoE is integrated as a host-runtime plugin and registers an OpenAI-compatible endpoint with endpoint id `flash-moe`.

No mesh wire protocol, protobuf, ALPN, gossip schema, Skippy stage protocol, model-package format, or llama.cpp patch queue changes are included.

## Branch / Commit Integrity

- Base branch: `main`
- Validated base SHA: `8d12c0be26fb3af4ed309fde6df65acfabff0162`
- Merge-base SHA: `8d12c0be26fb3af4ed309fde6df65acfabff0162`
- Introduced commit: `d923e037bd7ffb38f632307228d64d61c4eb0640 Add Flash-MoE SSD backend plugin`

## Validation

Validation mode: **Tier 3 — shared backend/runtime integration**.

Local proof:

- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm-host-runtime flash_moe`: PASS, 10 passed, 0 failed
- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm-host-runtime --lib`: PASS, 1242 passed, 0 failed, 5 ignored
- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo check -p mesh-llm`: PASS
- `/tmp/mesh-llm-just-tool/bin/just build`: PASS

Remote CI is pending until this PR is opened. This PR should not be considered merge-ready until required GitHub checks pass on the final PR SHA.

## Runtime Safety

- No new mesh-wide blocking locks.
- No new unbounded queues or background buffering paths.
- Managed Flash-MoE process ownership is scoped to the plugin lifecycle.
- Health probing uses a bounded timeout and keeps warm-up states non-fatal unless the child process exits.
- No mesh protocol invariant is removed.

## Rollback Plan

Rollback: `git revert <post_merge_commit_sha>`.

DB downgrade: not applicable. Data repair: not applicable. Operational caveat: rollback removes the built-in Flash-MoE adapter and restores the previous plugin/backend surface.

## Known Residual Risks

- Real Flash-MoE Qwen3.5-397B hardware smoke was not run locally because the required Flash-MoE binary and model artifacts are not available in this environment.
- This intentionally does not automate Flash-MoE artifact preparation or model conversion; those remain separate follow-up work.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -86,9 +86,9 @@ This is a single-node strategy. The goal is running e.g. Qwen3.5-397B-A17B (~209
 
 Today mesh-llm has two MoE modes: **solo** (model fits in memory, run it whole) and **split** (model doesn't fit, shard experts across nodes). SSD streaming would be a third mode: model doesn't fit in memory but *does* fit on one node's SSD. No mesh coordination, no cross-node traffic, no splitting — just one machine streaming experts from disk.
 
-**Plan:** Use flash-moe directly as an alternative backend, not hack SSD streaming into llama.cpp. llama.cpp's `ggml_mul_mat_id` assumes all expert weights resident in one contiguous tensor — changing that is deep surgery across ggml, the Metal backend, and the model loader. Flash-moe is a working engine. Mesh-llm spawns it like it spawns llama-server — process management + HTTP wrapper.
+**Status:** Initial mesh-llm integration exists as a built-in `flash-moe` plugin adapter. Mesh-llm can spawn a local Flash-MoE `infer --serve` process or attach an already-running OpenAI-compatible `/v1` endpoint, then route to it through the plugin inference path. The adapter intentionally does not vendor Flash-MoE, automate Flash-MoE installation, or prepare SSD-streaming artifacts.
 
-Only supports Qwen3.5-397B for now (hardcoded architecture). That's fine — it's the model we want to run.
+Flash-MoE only supports Qwen3.5-397B for now (hardcoded architecture). That's fine for the target model. Remaining mesh-llm work is install/artifact-prep documentation, real-machine smoke coverage for the target path, and revisiting out-of-tree adapter packaging once the plugin SDK and crates.io surface are stable enough.
 
 ## Blackboard ✅
 
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ mesh-llm serve --auto --headless
 | Join by invite token | `mesh-llm serve --join <token>` | [docs/MESHES.md](docs/MESHES.md) |
 | Run an API-only client | `mesh-llm client --auto` | [docs/MESHES.md](docs/MESHES.md) |
 | Run a big model with splits | `mesh-llm serve --model hf://meshllm/<repo>@<rev> --split` | [docs/SKIPPY_SPLITS.md](docs/SKIPPY_SPLITS.md) |
+| Attach a Flash-MoE SSD backend | `mesh-llm serve` with `[[plugin]] name = "flash-moe"` | [docs/plugins/flash-moe.md](docs/plugins/flash-moe.md) |
 | Use Goose, OpenCode, Claude Code, or Pi | `mesh-llm goose`, `mesh-llm opencode`, `mesh-llm claude`, `mesh-llm pi` | [docs/AGENTS.md](docs/AGENTS.md) |
 | Build or contribute | `just build` | [CONTRIBUTING.md](CONTRIBUTING.md) |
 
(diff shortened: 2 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -501,12 +502,42 @@ jobs:
...
       - name: Benchmark gate
         id: benchmark_gate
+        shell: bash
         run: |
           if [[ "${{ github.event_name }}" == "workflow_dispatch" || "${{ needs.changes.outputs.benchmarks }}" == "true" ]]; then
             echo "ready=true" >> "$GITHUB_OUTPUT"
           else
             echo "ready=false" >> "$GITHUB_OUTPUT"
           fi
+      - name: Bootstrap rustup
```
michaelneale: curious why needed this and bash for CI ? I guess it is testing for the plugin? seem ok then

ROADMAP.md
```diff
@@ -81,7 +81,7 @@ This is a single-node strategy. The goal is running e.g. Qwen3.5-397B-A17B (~209
...
 
-**Plan:** Use flash-moe directly as an alternative backend, not hack SSD streaming into llama.cpp. llama.cpp's `ggml_mul_mat_id` assumes all expert weights resident in one contiguous tensor — changing that is deep surgery across ggml, the Metal backend, and the model loader. Flash-moe is a working engine. Mesh-llm spawns it like it spawns llama-server — process management + HTTP wrapper.
+**Plan:** Use flash-moe directly as an alternative backend, not hack SSD streaming into llama.cpp. llama.cpp's `ggml_mul_mat_id` assumes all expert weights resident in one contiguous tensor — changing that is deep surgery across ggml, the Metal backend, and the model loader. Flash-moe is a working engine. Mesh-llm integrates it through a built-in `flash-moe` plugin: process management, OpenAI-compatible endpoint registration, model discovery, and routing.
```
michaelneale: could probably drop this now as this implements it?

michaelneale:
yeah this is pretty good use of plugins - I think makes sense as a plugin. would be good to round out the install experience (and make it clear that flash moe has to be installed separately right?). 

Flash moe repo hasn't had any commits in 2 months however so I wonder if that repo is still alive, or it is somewhere else? could see a lot of people benefitting from it assuming flash-moe is solid enough.

i386:
Do we publish enough to crates.io that is could live in its own repo?

IvGolovach:
I refreshed this against current main and narrowed the PR back to the Flash-MoE adapter itself. The unrelated CI changes are gone, the README/runtime conflicts are resolved, and the old review threads are now outdated. Local Flash-MoE tests pass and the full GitHub matrix is green now. Thanks again for the guidance here — I think this is a much cleaner review unit.

ndizazzo (approved):
@IvGolovach Agent caught one thing: 

```
Medium: Flash-MoE managed backend startup failures happen after the plugin runtime already sends InitializeResponse, so missing/bad commands can look like plugin crash/restart loops instead of clean startup failures with install hints.
```

This wouldn't happen in all cases, so marking this approved and you can fast-follow with anything to address it.

ndizazzo:
@IvGolovach fire up a PR for the results in [this comment](https://github.com/Mesh-LLM/mesh-llm/pull/475#pullrequestreview-4302233340), trying to get our PR backlog down

PR #472, opened 2026-05-08
Add a Skippy package certification gate

IvGolovach:
## Summary
Adds a repeatable `mesh-llm models certify` gate for Skippy layer packages. The command can now produce a structured confidence report before a split package is treated as ready for release or rollout.

The gate covers package resolution, manifest/artifact integrity, representative local stage materialization, and optional runtime smoke checks against an already running OpenAI-compatible mesh endpoint.

## Why
Skippy layer packages are moving from experimental artifacts toward a real rollout path. Maintainers need a small, repeatable command that can answer the practical questions before a package is trusted:

- Does this model ref resolve to the intended layer package?
- Do the selected package artifacts pass integrity checks?
- Can representative first/final stage ranges be materialized locally?
- If runtime gates are requested, does the served model actually appear in `/v1/models` and return text-bearing outputs through both chat completions and Responses API paths?

## What changed
- Added `mesh-llm models certify`.
- Added a structured JSON certification report with package, manifest, source model, materialized stage, and gate status details.
- Added `--package-only` mode for package integrity and local materialization without claiming runtime readiness.
- Added runtime smoke gates for `/v1/models`, `/v1/chat/completions`, and `/v1/responses`.
- Hardened runtime smoke gates so a successful HTTP status is not enough; chat and Responses gates must return real text-bearing output.
- Added CLI validation for ambiguous modes, missing `--api-base`, invalid runtime base URLs, empty runtime prompts, and `--max-tokens 0`.
- Documented the command and mode contract in `docs/CLI.md`.

## Compatibility
No mesh protocol change.
No plugin protocol change.
No database migration.
The new CLI command is additive.

## Branch integrity
Base branch: `main`.
Validated base SHA: `8d12c0be26fb3af4ed309fde6df65acfabff0162`.
Branch status against `origin/main`: ahead `1`, behind `0`.
Merge-base SHA: `8d12c0be26fb3af4ed309fde6df65acfabff0162`.
Fast-forward safety: `origin/main` is an ancestor of this branch.

## Commit integrity
Introduced commit:

- `ccf84df8cf7ccb9fbbb4a383b6c29f995af404d5 Add Skippy package certification gate`

## Diff scope
Changed files:

- `crates/mesh-llm-host-runtime/src/cli/commands/models/mod.rs`
- `crates/mesh-llm-host-runtime/src/cli/mod.rs`
- `crates/mesh-llm-host-runtime/src/cli/models.rs`
[...]

PR #478, opened 2026-05-08
Add unified HF model certification jobs

i386:
This follows up #472 by adding Hugging Face Jobs workflows for remote family certification and layer-package splitting of large GGUF families, while keeping the public CLI to two model verbs: `models package` and `models certify`.

## What changes

- Uses one `mesh-llm models package` command for both local layer-package writing and remote HF Jobs splitting.
- Uses one `mesh-llm models certify` command for both local package/runtime certification and remote HF Jobs family certification.
- Makes remote HF Jobs explicit with `--hf-job`; without it, `models package` and `models certify` stay local.
- Removes the legacy hidden top-level `model-prepare` / `model-package` entrypoints and removes `models certify-family` outright.
- Removes the public `--quant` split argument for these flows; model refs use the colon selector form, for example `unsloth/MiMo-V2-Flash-GGUF:IQ4_XS`.
- Adds a shared HF Jobs Docker image that prebuilds Rust, the pinned llama.cpp stage ABI, `skippy-model-package`, `skippy-correctness`, `skippy-server`, and `llama-spec-bench`.
- Adds `.github/workflows/hf-jobs-image.yml`, which publishes `ghcr.io/mesh-llm/mesh-llm-hf-jobs:latest`, `:main`, and SHA tags from `main` builds.
- Updates HF job scripts to detect `/opt/mesh-llm` in that image and skip the slow apt/Rust/llama.cpp/tool rebuild path when the image ref matches the requested `MESH_LLM_REF`.
- Splits CPU auto-selection by workflow: package jobs keep the cheap splitter baseline, while certification jobs auto-select the cheapest CPU flavor with enough advertised RAM plus headroom for runtime correctness lanes.
- Updates the HF family-certification agent skill to certify on the smallest representative same-family text GGUF first, then package the largest useful 4-bit quant only after certification passes.
- Refreshes the llama.cpp pin/candidate docs for Unsloth MiMo 2.5 certification.

## Local vs remote behavior

`models package` defaults to local packaging. It invokes the bundled/sibling `skippy-model-package write-package`, requires `--out-dir`, accepts local package identity/projector flags, and writes artifacts on the caller's machine. Local package mode rejects remote job controls such as `--confirm`, `--dry-run`, `--follow`, and `--target` so an operator does not think they queued cloud work when they did not.

`models package --hf-job` switches to remote Hugging Face Jobs packaging. It resolves the source GGUF from a colon-selector ref, chooses HF hardware, timeout, target repo, and job image, then prints a dry-run plan by default. It submits only when `--confirm` is present. Remote package mode rejects local-only flags such as `--out-dir`, `--projector`, `--source-repo`, `--source-revision`, and `--source-file`.

`models certify` defaults to local certification. It certifies an existing package/catalog ref or running endpoint using `--package-only` or `--api-base`, can write `--report-out`, and never talks to HF Jobs. Local certify mode rejects family-certification job controls such as `--family`, `--artifact-repo`, `--confirm`, `--follow`, split-layer knobs, dtype matrix knobs, and state/cache lane knobs.

`models certify --hf-job` switches to remote family certification. It requires a source GGUF colon-selector ref and `--family`, plans the HF hardware/image/timeout/cost, and submits only with `--confirm`. Auto hardware selection is memory-aware for this path. Certification should use the smallest representative same-family text quant, not necessarily the production 4-bit package target: for MiMo Flash, `TQ1_0` dry-runs to `cpu-xl`, while larger `IQ4_XS` dry-runs to `cpu-performance`. Remote certification may publish artifacts with `--artifact-repo`; local report flags such as `--report-out`, `--package-only`, and `--api-base` are rejected.

Management flags for existing HF Jobs stay on the same verbs: `models package --status|--logs|--cancel|--list` for package jobs and `models certify --status|--logs|--cancel|--list` for certification jobs.

## Representative certification quant policy

Family certification is meant to prove that Skippy can split and execute the family topology, not to prove every production-sized quant. For large families, the workflow should first list available GGUF variants and certify the smallest same-family text quant that still exercises the same llama.cpp architecture/topology. After that certification passes, package the largest useful 4-bit production candidate separately.

For MiMo Flash, the current representative certification target is `unsloth/MiMo-V2-Flash-GGUF:TQ1_0` rather than `IQ4_XS` or larger 4-bit variants. Current dry runs show:

- `TQ1_0` certification with `--timeout 2h`: `cpu-xl`, about `$2.00` max.
- `IQ4_XS` certification with `--timeout 2h`: `cpu-performance`, about `$37.40` max.
- MiMo package splitting: `cpu-upgrade`, about `$0.06` max.

The HF family-certification agent skill now encodes this policy: list variants first, reject VL/multimodal repos for text-family evidence, prefer tiny same-family quants such as `TQ1_0` / `UD-IQ1_S` / `UD-IQ1_M`, and keep certification evidence distinct from the later package target.

## End-to-end workflow

1. A relevant build on `main` publishes the shared HF Jobs image to GHCR.
2. A caller dry-runs remote packaging or certification with `--hf-job`. The default `--mesh-llm-ref main --job-image auto` resolves to `ghcr.io/mesh-llm/mesh-llm-hf-jobs:latest`.
3. The dry run prints the resolved model file, target/artifact repo, hardware flavor, timeout, maximum cost, and redacted HF Job spec. No HF Job is submitted without `--confirm`.
4. Submitting with `--hf-job --confirm` runs the HF Job. The container uses the prebuilt tools when the baked image ref is compatible, then mounts the source GGUF repo and bucket workspace.
5. Package jobs split the model and publish the layer package/catalog entry. Certification jobs run the family certification script and optionally publish artifacts when `--artifact-repo` is provided.
6. For branch or exact-SHA validation, callers should pass a pushed `--mesh-llm-ref` plus an explicit matching `--job-image ghcr.io/mesh-llm/mesh-llm-hf-jobs:sha-...`. If `--job-image auto` is used with a non-`main` ref, the job falls back to `ubuntu:22.04` and builds in-job so it does not accidentally run stale `latest` code.

## CLI examples

Local package write:

```bash
mesh-llm models package ./model.gguf \
  --out-dir ./model-package \
  --model-id org/repo:Q4_K_M
```

Remote splitter dry run:

```bash
mesh-llm models package unsloth/Kimi-K2.5-GGUF:Q4_1 \
  --hf-job \
  --json
```

Local package certification:

```bash
mesh-llm models certify hf://meshllm/Qwen3-8B-Q4_K_M-layers \
  --package-only \
  --report-out cert.json
```

Remote family certification dry run using the smallest representative MiMo text quant:

```bash
mesh-llm models certify unsloth/MiMo-V2-Flash-GGUF:TQ1_0 \
  --hf-job \
  --family mimo2 \
  --artifact-repo meshllm/family-certification-runs \
  --json
```

Remote family certification submit after reviewing the dry-run max cost:

```bash
mesh-llm models certify unsloth/MiMo-V2-Flash-GGUF:TQ1_0 \
  --hf-job \
  --family mimo2 \
  --artifact-repo meshllm/family-certification-runs \
  --confirm \
  --follow \
  --json
```

## Validation

- `cargo test -p model-package`
- `cargo test -p model-package hf_jobs_image_auto_uses_main_image_only_for_main_ref`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo check -p mesh-llm-host-runtime`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo check -p mesh-llm`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm-host-runtime models_ --lib`
- `bash -n crates/model-package/src/scripts/certify-model-job.sh`
- `bash -n crates/model-package/src/scripts/split-model-job.sh`
- `actionlint .github/workflows/hf-jobs-image.yml`
- `git diff --check`
- `mesh-llm model-prepare --help` exits as unrecognized
- `mesh-llm models certify-family --help` exits as unrecognized
- `mesh-llm models package --help` exposes `--hf-job` and does not expose `--quant`
- `mesh-llm models certify --help` exposes `--hf-job` and does not expose `--quant`
- Dry-run package spec emits `dockerImage: ghcr.io/mesh-llm/mesh-llm-hf-jobs:latest` for `mesh-llm@main`
- Dry-run certifier spec emits `dockerImage: ghcr.io/mesh-llm/mesh-llm-hf-jobs:latest` for `mesh-llm@main`
- MiMo `TQ1_0` certification dry run with `--timeout 2h` selects `cpu-xl` and estimates max cost at about `$2.00`; MiMo `IQ4_XS` certification dry run selects `cpu-performance` at about `$37.40`; package splitting dry run still selects `cpu-upgrade` at `$0.06`

## Notes

- This PR is intentionally based on an upstream copy of #472 (`jd/pr-472-certification-base`) so the diff is only the HF-family-certification follow-up.
- Earlier exploratory HF MiMo certification jobs were canceled; no certification job is currently running.
- A one-off PR SHA image was manually built and pushed for development: `ghcr.io/mesh-llm/mesh-llm-hf-jobs:sha-026dd91d`.


i386:
Updated #478 to follow #477's mounted bucket workspace pattern and fixed the timeout policy:

- package jobs now inherit #477's writable `/bucket/job-work` workspace behavior
- `certify-family` also mounts `/bucket` writable and stores family-certification work under `/bucket/job-work`
- package `recommended_cpu_timeout_seconds` now tops out at a realistic 2h instead of 12h
- `certify-family --timeout` is now the hard HF max-cost timeout and is no longer raised by package size floors

Verified dry-run for MiMo 2.5 after the push:

```text
mesh-llm-ref: 7e4ec6ae51a754a46607dce8f65ecee1c6744c24
hardware: cpu-xl (16 vCPU, 124 GB)
requested timeout: 2h
effective timeout: 2h
timeout_bumped_to_minimum: false
max cost: $2.00004
/bucket: writable mount
```

IvGolovach:
Thanks for pushing this forward — this feels like a really useful next layer after #472. 🙂

I did a first pass, and the overall direction makes sense to me: keeping local package certification separate from remote family certification feels like the right split, especially for the larger GGUF families where local certification is just not realistic.

I also like the safety shape here: dry-run first, explicit `--confirm-max-cost-usd` before any spend-bearing HF Job, and the `/bucket/job-work` workspace pattern from #477. That all feels like the right kind of friction for this workflow.

The only thing I would wait on before calling this ready is the current `linux` CI failure.
 It looks mechanical rather than design-level: Clippy is failing in `recommended_cpu_timeout_seconds()` because the `gib <= 128.0` branch and the final `else` both return `2 * 60 * 60`.

Once that is cleaned up and CI is green, I’d be happy to take one more focused pass on the cost guardrails, token handling, and the #477 workspace behavior. Overall, this looks like the right next step for family certification.

i386:
MiMo 2.5 certification attempt update:

- Submitted HF job https://huggingface.co/jobs/meshllm/69fe6311aff1cd33e8f3162e against `7e4ec6ae51a754a46607dce8f65ecee1c6744c24` with the requested 2h hard cap on `cpu-xl`.
- The remote build passed with llama.cpp upstream `9f5f0e689c9e977e5f23a27e344aa36082f44738` and the current patch queue.
- Certification did **not** pass: first correctness lane reported `single-step: fail (exit 1)`.
- I canceled the job after that lane failed to stop spend. Final job stage is `CANCELED`.

One workflow issue surfaced: the job script cleaned the failed work dir, so this first run only gave us the lane-level failure, not the per-lane log/artifacts. I pushed follow-up commit `85de42d1` on this PR so failed certifications preserve/upload available artifacts before returning failure.

Next certification options from dry-run, not submitted:

- `cpu-xl`: 16 vCPU / 124 GB, 2h cap = **$2.00004**. The MiMo `IQ4_XS` quant resolves to 164,445,782,752 bytes across 4 shards, so this is likely memory-constrained.
- `cpu-performance`: 32 vCPU / 256 GB, 2h cap = **$37.40004**. This is the first CPU flavor that looks memory-sane for this quant, but needs explicit spend confirmation before submission.

No MiMo certification artifact should be promoted as passing yet.

michaelneale (approved):
I think a good improvement - and makes hf explicit, and no more --quant which felt wrong. 

As long as people realise if they do it localy could be prodigious disk usage (but the slow download will be a hint of that!)

PR #479, opened 2026-05-09
Allow peers to hydrate model files through the mesh

IvGolovach:
## Summary

Nodes can now hydrate missing Hugging Face model files from capable mesh peers before falling back to the normal Hugging Face download path. This helps mesh nodes without direct HF access or a local HF token reuse model files already present on another peer.

## Why

Skippy/runtime flows can coordinate serving across machines, but model acquisition still assumed each node could independently download the same files. This adds a peer-assisted path for exact model assets while keeping the existing HF downloader as the fallback whenever peer hydration is unavailable, incomplete, disabled, or unsafe.

## What changed

- Added a `mesh-model-transfer` stream over the existing mesh subprotocol path.
- Advertised peer model-transfer support through additive `MeshSubprotocol` gossip.
- Refreshed the new capability through direct and transitive peer update paths.
- Added a peer-aware HF asset hydrator before the normal HF fallback.
- Wired hydration into auto assignment, `/mesh/load`, and runtime-slice stage preparation once a `Node` is available.
- Installed transferred files into the normal HF cache layout, resolving mutable refs such as `main` to immutable cached snapshot commits.

## Protocol and compatibility

- No new ALPN was added.
- The new advertised subprotocol is `mesh-model-transfer` major `1`, with `file-get`, `range`, and `sha256` features.
- Older peers can ignore the unknown subprotocol and continue using the existing HF fallback.
- No existing protobuf fields are removed, repurposed, or made required.

## Safety and privacy

- No model cache inventory is gossiped or exposed through status.
- Peers are only tried when they explicitly advertise model-transfer support.
- Requests are exact `repo` / `revision` / `file` lookups.
- Unsafe repo ids, revision refs, traversal paths, absolute paths, and backslash paths are rejected.
- Serving peers only read from managed HF cache snapshots and reject symlink escapes.
- Requesters validate generation, immutable resolved revision, offset, size, and sha256 before installing.
- Partial downloads use `.mesh-peer.part`, support safe resume, and remove unsafe symlink partials.
- `MESH_LLM_MODEL_TRANSFER=0|false|off|no` disables this path.

## Validation

Branch proof:

- Base: `origin/main` at `b9c9c67b0e61c8bda541db6965f42a50db1f89f8`
- Head: `d4aa0d5b1de0ba519a8fc697d99653374c6eabd5`
- Ahead / behind: `0 1`

Local checks:

- `cargo test -p mesh-llm-host-runtime model_transfer --lib`: PASS, 14 passed
- `cargo test -p mesh-llm-host-runtime peer_hydration_plan --lib`: PASS, 2 passed
- `cargo test -p mesh-llm-host-runtime download_hf_repo_file_uses_hydrator_before_hf_download --lib`: PASS, 1 passed
- `cargo check -p mesh-llm-host-runtime`: PASS
- `cargo test -p mesh-llm --lib`: PASS
- `cargo test -p mesh-llm --test protocol_compat_v0_client`: PASS, 10 passed
- `cargo test -p mesh-llm --test protocol_convert_matrix`: PASS, 11 passed
- `rustup run stable cargo clippy -p mesh-llm -- -D warnings`: PASS

## Migration and rollback

- DB migration: not applicable.
- Rollback: revert this PR, or run `git revert <post_merge_commit_sha>` after merge.
- Operational caveat: rollback restores the previous HF-only model acquisition behavior.

## Known residual risks

Real multi-node transfer behavior should still be exercised during review/deployment on representative mesh nodes, especially for machines without direct Hugging Face access.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
@@ -37,6 +37,7 @@ pub(super) fn peer_meaningfully_changed(old: &PeerInfo, new: &PeerInfo) -> bool 
         || old.served_model_runtime != new.served_model_runtime
         || old.artifact_transfer_supported != new.artifact_transfer_supported
         || old.stage_status_list_supported != new.stage_status_list_supported
+        || old.model_transfer_supported != new.model_transfer_supported
         || old.version != new.version
         || old.owner_summary != new.owner_summary
         || old.gpu_reserved_bytes != new.gpu_reserved_bytes
@@ -110,6 +111,7 @@ pub(super) fn apply_transitive_ann(
     existing.served_model_runtime = ann.served_model_runtime.clone();
     existing.artifact_transfer_supported = ann.artifact_transfer_supported;
     existing.stage_status_list_supported = ann.stage_status_list_supported;
+    existing.model_transfer_supported = ann.model_transfer_supported;
     if ann.experts_summary.is_some() {
         existing.experts_summary = ann.experts_summary.clone();
     }
@@ -461,7 +463,9 @@ impl Node {
             existing.owner_summary = owner_summary.clone();
             existing.served_model_descriptors = ann.served_model_descriptors.clone();
             existing.served_model_runtime = ann.served_model_runtime.clone();
+            existing.artifact_transfer_supported = ann.artifact_transfer_supported;
             existing.stage_status_list_supported = ann.stage_status_list_supported;
+            existing.model_transfer_supported = ann.model_transfer_supported;
             if ann.version.is_some() {
                 existing.version = ann.version.clone();
             }
@@ -685,6 +689,7 @@ impl Node {
                     owner_attestation: p.owner_attestation.clone(),
                     artifact_transfer_supported: p.artifact_transfer_supported,
                     stage_status_list_supported: p.stage_status_list_supported,
+                    model_transfer_supported: p.model_transfer_supported,
                 })
                 .collect()
         };
@@ -752,6 +757,7 @@ impl Node {
             artifact_transfer_supported:
                 crate::models::artifact_transfer::artifact_transfer_enabled(),
             stage_status_list_supported: true,
+            model_transfer_supported: crate::models::model_transfer::model_transfer_enabled(),
         });
         announcements
     }
@@ -807,6 +813,7 @@ mod tests {
             owner_attestation: None,
             artifact_transfer_supported: true,
             stage_status_list_supported: true,
+            model_transfer_supported: true,
         }
     }
 
@@ -896,6 +903,15 @@ mod tests {
     }
 
     #[test]
+    fn test_meaningfully_changed_model_transfer_support() {
+        let old_peer = test_peer(Some(100));
+        let mut new_peer = test_peer(Some(100));
+        new_peer.model_transfer_supported = !old_peer.model_transfer_supported;
+
+        assert!(peer_meaningfully_changed(&old_peer, &new_peer));
+    }
+
+    #[test]
     fn test_apply_transitive_ann_refreshes_explicit_model_interests() {
         let mut existing = test_peer(Some(100));
         let mut ann = test_announcement(Some(100));
@@ -921,6 +937,18 @@ mod tests {
         assert!(existing.stage_status_list_supported);
     }
 
+    #[test]
+    fn test_apply_transitive_ann_refreshes_model_transfer_support() {
+        let mut existing = test_peer(Some(100));
+        existing.model_transfer_supported = false;
+        let mut ann = test_announcement(Some(100));
+        ann.model_transfer_supported = true;
+
+        apply_transitive_ann(&mut existing, &test_addr(0x33), &ann);
+
+        assert!(existing.model_transfer_supported);
+    }
+
     #[tokio::test]
     async fn test_add_peer_refreshes_stage_status_list_support() {
         let node = Node::new_for_tests(NodeRole::Worker).await.unwrap();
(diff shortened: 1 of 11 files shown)


crates/mesh-llm-host-runtime/src/mesh/mod.rs
```diff
@@ -4611,6 +4750,321 @@ impl Node {
...
+                current_time_unix_ms(),
+                std::process::id(),
+                peer_id.fmt_short()
+            ),
+            repo: asset.repo.clone(),
+            revision: asset.revision.clone(),
+            file: asset.file.clone(),
+            offset: requested_offset,
+            expected_size: None,
+            expected_sha256: None,
```
i386, line 4880: Small trust-boundary wrinkle here: in the real hydration path we ask the peer for bytes without an independent expected size or sha, then later validate against the sha the same peer reports. That catches corrupted transport, but not a stale/bad peer handing us a self-consistent wrong file. Might be worth only doing peer hydration when we already have an expected digest/size, or limiting this to trusted peers.

crates/mesh-llm-host-runtime/src/mesh/gossip.rs
```diff
@@ -752,6 +757,7 @@ impl Node {
             artifact_transfer_supported:
                 crate::models::artifact_transfer::artifact_transfer_enabled(),
             stage_status_list_supported: true,
+            model_transfer_supported: crate::models::model_transfer::model_transfer_enabled(),
```
i386, line 760: Since this advertises by default, any admitted mesh peer can discover that we serve model-transfer and then request exact HF cache files. That feels a little spicy for private/gated model caches. I’d be more comfortable if serving files were opt-in, or tied to the ownership/trust policy, rather than default-on for every mesh.

crates/mesh-llm-host-runtime/src/mesh/mod.rs
```diff
@@ -4611,6 +4750,321 @@ impl Node {
...
+        let installed_sha = tokio::task::spawn_blocking(move || {
+            crate::models::model_transfer::file_sha256_hex(&destination_for_hash)
+        })
+        .await
+        .context("join installed model sha256 task")??;
+        anyhow::ensure!(
+            installed_sha.eq_ignore_ascii_case(&expected_sha),
+            "installed model sha256 mismatch"
+        );
+        crate::models::model_transfer::install_hf_revision_ref(
```
i386, line 5060: One subtle stale-cache case: for mutable refs like `main`, we accept the peer’s resolved commit and install that as our local HF ref. A well-meaning peer with an old cache could pin us to an older model even when Hugging Face would resolve newer. Maybe peer-hydrate only immutable revisions, or verify mutable refs independently before writing the ref.

i386 (commented):
Left a few notes on the model-transfer path. The feature shape is promising; the things I called out are mostly around trust boundaries and mutable refs before this becomes a default-on mesh behavior.

i386:
Thinking about this more, I think this needs a wider design discussion before we merge it.

The useful thing this PR proves is that peers should be able to help move model/runtime bytes around the mesh. But I’m no longer convinced we want a standalone `mesh-model-transfer` path, and I’m also having second thoughts about the Skippy layer/package transfer path we already added. These are really the same product surface: peer-to-peer blob transfer for model assets, layers, package artifacts, and maybe future runtime files.

I’d rather pause here and design one generic, trustworthy blob-transfer layer that can carry all of those typed asset kinds, with public-mesh safety as a first-class requirement. In particular, public peers should be able to move bytes, but they should not be able to define what those bytes mean. That probably means content-addressed transfer, typed requests, independent digest/manifest verification, and per-kind authorization instead of plain mesh admission.

So my preference is: don’t merge this as a second transfer lane yet. And we should probably disable the existing Skippy layer/package peer transfer until we answer the same question there: what is the unified blob protocol, what is safe on public mesh, and what requires owned/allowlisted trust?

i386 (changes requested):

IvGolovach:
Thanks for thinking this through so carefully. I read through the review notes again, and this framing makes sense to me.

I think this PR is still useful as a proof of concept: it shows that peer-assisted movement of model/runtime bytes is a real capability the mesh probably wants. But I agree that the standalone `mesh-model-transfer` lane is not the right merge shape. It would be too easy to end up with separate transfer paths for model files, Skippy layers/packages, and future runtime artifacts, all solving the same trust problem slightly differently.

So I’m happy to pause this PR as-is and treat it as useful prior art rather than trying to merge it as a second transfer lane.

The direction I’d suggest is:

1. First, make the existing Skippy layer/package peer transfer safe by default for public meshes, either by disabling peer serving by default or gating it behind explicit owned/allowlisted trust.
2. Then write down the unified blob-transfer design: content-addressed requests, typed asset kinds, independent digest/manifest verification, mutable HF ref handling, and per-kind authorization.
3. Then migrate Skippy package artifacts onto that shared layer, since Skippy already has manifest hashes and artifact digests.
4. After that, rework model hydration on top of the same layer, with public peers only moving bytes whose digest/size/manifest the requester already trusts independently.

The rule that clicks for me is: public mesh peers can help move bytes, but they should not get to define what those bytes mean.

I’m happy to take a first pass at the safety/design slice if that sounds like the right next step. Thanks again for calling this out before we accidentally grow two protocols we’d later have to unwind 🙂

i386:
@IvGolovach thanks for taking this constructively! 

Should we disable or limit the model layer downloads we have today in skippy?

i386:
For the generic blob transfer protocol - could we start with a spec in GitHub issues? I'd like to get the whole brain trust thinking about it before we write code

IvGolovach:
Yeah, I think that’s the right call.

For the Skippy layer/package downloads we already have today, I’d lean toward limiting them immediately rather than removing the path entirely. The current path has some useful guardrails already: stage/topology authorization, manifest SHA checks, artifact size/SHA verification, safe relative paths, and HF fallback behavior. But I agree the default serving surface is still too broad while the shared blob-transfer design is unsettled.

My suggested first safety slice would be:

- don’t advertise or serve Skippy artifact transfer by default on public meshes;
- keep it available only behind explicit operator opt-in and/or owned/allowlisted trust policy;
- preserve the existing requester fallback path so split/package flows degrade back to normal resolution instead of hard failing;
- add tests around advertisement, peer selection, and rejected untrusted requests.

And yes, starting the generic blob-transfer protocol as a GitHub issue/spec sounds right to me. I can draft that around the main questions we surfaced here: content addressing, typed asset kinds, independent digest/manifest verification, mutable HF refs, public-vs-owned mesh policy, and migration from the current Skippy artifact path.

So I’d treat this as two separate next steps: a small safety PR for today’s Skippy transfer behavior, and then a spec issue for the shared blob protocol before writing the larger replacement.

i386:
@IvGolovach im aligned with that - could you make that change to skippy and leave an issue open to remove it once the new blob stuff has landed?

IvGolovach:
Yep, absolutely. I’ll take that as the next concrete step.

I think I’ll do it as a separate small PR from current `main`, rather than folding it into this model-transfer PR. This PR is useful as prior art for the future blob-transfer work, but the Skippy safety change is a narrower fix to behavior that already exists on `main`, so keeping it separate should make review and merge much cleaner.

For that PR, I’ll stop Skippy peer layer/package transfer from being advertised or served by default for public mesh usage, while keeping an explicit opt-in / trusted-owner path for deployments that intentionally want it. I’ll also make sure the normal fallback behavior stays intact so package flows don’t hard-fail just because peer transfer is unavailable.

I’ll add focused tests around the advertisement/serving gate and peer-selection behavior, then open a tracking issue to remove or migrate this stopgap once the unified blob-transfer layer lands.

PR #483, opened 2026-05-09
Prevent tool-call grammar crashes in Skippy

i386:
## Summary

Skippy can now survive and avoid Qwen-style lazy tool-call grammar crashes when a model begins a `<tool_call>` block.

This adds two carried llama.cpp patches:

- contain native chat grammar sampler exceptions so they cannot escape across the Rust FFI boundary and abort the process
- trigger lazy chat grammar on the structural marker before trailing whitespace, changing the Qwen/Nemotron-style trigger from `<tool_call>\n` to `<tool_call>` while preserving the grammar itself

## Root Cause

The lazy grammar trigger was waiting for the full marker including trailing newline. That left the token after `<tool_call>` unconstrained. If that token included the newline plus invalid following bytes, the lazy replay path could throw from native grammar code. Because that exception crossed into Rust, the whole process aborted.

## Validation

- `LLAMA_WORKDIR=$(mktemp -d /tmp/mesh-llm-llama.XXXXXX) scripts/prepare-llama.sh pinned` succeeded in a fresh checkout
- `just build` succeeded locally on Metal
- `PATH=/usr/local/cuda/bin:$PATH just build cuda 120` succeeded on `white.local`
- Restarted the public test node on `white.local`; it served `bartowski/Qwen2.5-Coder-7B-Instruct-GGUF:Q4_K_M`
- Sent the repro request against the embedded Skippy OpenAI endpoint; native log showed `Grammar triggered on regex: '<tool_call>'` and the process stayed alive

## Protocol

No mesh wire protocol changes.

### DIFF
--- a/third_party/llama.cpp/patches/0078-Contain-chat-grammar-sampler-exceptions.patch
+++ b/third_party/llama.cpp/patches/0078-Contain-chat-grammar-sampler-exceptions.patch
@@ -1,0 +1,86 @@
+From 0e97eb7cd891a876f7e68a22f9efcc219554c349 Mon Sep 17 00:00:00 2001
+From: Mesh-LLM CI <ci@mesh-llm.local>
+Date: Sat, 9 May 2026 13:23:46 +1000
+Subject: [PATCH] Contain chat grammar sampler exceptions
+
+---
+ src/skippy.cpp | 42 ++++++++++++++++++++++++++++++++++++++----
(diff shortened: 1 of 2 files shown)


ndizazzo (approved):
@i386 Nice, how are you finding these? Just browsing llamacpp repo?

i386:
@ndizazzo trying to find that bug you reported running opencode! It tripped something bad on my CUDA machine. We certainly need to be dogfooding more.

PR #484, opened 2026-05-09
Require skippy smoke before llama pin updates

i386:
## Summary
- run the full skippy CI smoke inside the llama upstream canary before it updates the pin
- install the Hugging Face CLI and smoke-test system dependencies in the canary runner
- mark the canary summary when the smoke gate has run

## Validation
- actionlint .github/workflows/llama-upstream-canary.yml
- shellcheck scripts/summarize-llama-upstream.sh
- SKIPPY_CI_SMOKE=1 scripts/summarize-llama-upstream.sh 44dbe8c5218829be1603a08dee8847f4d0bca323 046e2844370208007c116fab448ed4033d77653f .deps/llama.cpp | sed -n '/## Validation/,'

## Notes
The previous one-off smoke passed locally on the current pin; this PR makes that smoke a required workflow gate for future pin bumps.


.github/workflows/llama-upstream-canary.yml
```diff
@@ -56,12 +59,21 @@ jobs:
...
+        run: scripts/skippy-ci-smoke.sh
+
       - name: Update llama.cpp upstream pin
         if: steps.sha.outputs.changed == 'true'
         run: scripts/update-llama-pin.sh
 
       - name: Generate summary
         if: steps.sha.outputs.changed == 'true'
+        env:
+          SKIPPY_CI_SMOKE: "1"
```
ndizazzo: Better to use repo variables for this, so disabling does not require a commit: https://github.com/Mesh-LLM/mesh-llm/settings/variables/actions

ndizazzo (changes requested):
Small QoL fix will harden this

.github/workflows/llama-upstream-canary.yml
michaelneale: OTOH - if there isn't a commit if like me, you may forget what you set the env var to or that you changed it in github settings (vs source), but but other than that, probably does make sense to set it as one.

PR #485, opened 2026-05-09
Gate Skippy artifact transfer behind trust policy

IvGolovach:
## Summary
Skippy layer-package peer artifact transfer is now opt-in instead of advertised and served by default.

Public mesh nodes now fail closed unless a deployment explicitly enables transfer:

- unset / `off`: do not advertise artifact transfer and reject inbound artifact requests;
- `trusted`: allow transfer only for same verified owner or explicitly trusted-owner peers;
- `open`: explicit lab/public opt-in for topology-authorized peers.

Normal local/Hugging Face package fallback remains intact when peer transfer is unavailable, so package-backed split planning should not hard-fail just because a peer cannot be used as an artifact source.

## Why
This follows the discussion in https://github.com/Mesh-LLM/mesh-llm/pull/479. The broader model/blob-transfer work is useful directionally, but current Skippy peer artifact transfer already exists on `main`. Until a unified blob-transfer layer lands, this keeps the existing Skippy path usable for intentional deployments while avoiding default public-mesh advertisement or serving.

## Related
- Related discussion: https://github.com/Mesh-LLM/mesh-llm/pull/479
- Follow-up tracking issue to open separately: remove or migrate this Skippy-specific stopgap once the unified blob-transfer layer lands.

## Diff scope
- `crates/mesh-llm-host-runtime/src/models/artifact_transfer.rs`: adds the shared artifact-transfer policy parser and the advertise/allow predicates.
- `crates/mesh-llm-host-runtime/src/mesh/gossip.rs`: advertises artifact transfer only when the local policy allows it.
- `crates/mesh-llm-host-runtime/src/mesh/mod.rs`: applies effective artifact-transfer eligibility to subprotocol negotiation and inbound serving.
- `crates/mesh-llm-host-runtime/src/runtime/local.rs`: uses effective peer eligibility for package-aware split participant planning.
- `crates/mesh-llm-host-runtime/src/mesh/tests.rs`: covers default-deny behavior, trusted-owner eligibility, explicit opt-in serving, and public-mesh rejection.
- `docs/design/TESTING.md` and `docs/specs/layer-package-repos.md`: document the new opt-in policy and testing expectations.

## Compatibility
- No protobuf schema changes.
- No new stream type or ALPN change.
- Existing stage-control compatibility is preserved; the trust gate is applied to artifact-transfer use, not to the legacy stage-control fallback.
- Older peers that only expose the raw `artifact_transfer_supported` bit are no longer treated as effective artifact sources unless the local policy also allows transfer.

## Branch integrity
- Base branch: `main`
- Validated base SHA: `6d5bd98dfc4259266dc1dd77255285c0524bcf19`
- Head SHA: `dd3baa7efea2a6941f1e7572885073cbafddc531`
- Merge base: `6d5bd98dfc4259266dc1dd77255285c0524bcf19`
- Fast-forward safety: `git merge-base --is-ancestor origin/main HEAD` exited `0`.

## Commit integrity
- Introduced commit: `dd3baa7efea2a6941f1e7572885073cbafddc531 Gate Skippy artifact transfer by trust policy`
- One logical change per commit: confirmed.
- Ledger: not applicable — this repository has no `scripts/ledger/` directory and this change family does not require a ledger record.
- Version: not applicable — no release/version metadata changed.

## Diff hygiene

```text
M	crates/mesh-llm-host-runtime/src/mesh/gossip.rs
M	crates/mesh-llm-host-runtime/src/mesh/mod.rs
M	crates/mesh-llm-host-runtime/src/mesh/tests.rs
M	crates/mesh-llm-host-runtime/src/models/artifact_transfer.rs
M	crates/mesh-llm-host-runtime/src/runtime/local.rs
M	docs/design/TESTING.md
```

## Validation
Validation tier: Tier 3 — shared runtime behavior, because this touches mesh gossip advertisement, Skippy artifact-transfer serving, and package-backed split participant eligibility.

Local validation:

```text
cargo fmt --all -- --check
PASS

git diff --check origin/main...HEAD
PASS

LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama.cpp/build-stage-abi-metal cargo check -p mesh-llm-host-runtime
PASS

LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm-host-runtime artifact_transfer --lib
PASS, 16 passed

LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm-host-runtime split_inventory_package_signal --lib
PASS, 2 passed

LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama.cpp/build-stage-abi-metal cargo test -p mesh-llm --lib
PASS, 0 tests
```

Remote validation:

- Required PR CI: pending until this PR is opened.
- Current `main` CI for `6d5bd98dfc4259266dc1dd77255285c0524bcf19`: PASS.

## Runtime safety
- No new blocking locks.
- No new unbounded queues, background tasks, or buffering paths.
- Existing topology authorization, requester identity validation, manifest digest validation, artifact digest validation, partial-file cleanup, and atomic install behavior remain in place.
- Public/default behavior is stricter than before: unset configuration no longer advertises or serves peer artifact transfer.

## Documentation integrity
Updated the layer-package spec and testing guide so the documented behavior matches the new opt-in transfer policy.

## Rollback plan
Rollback: revert this PR.

DB downgrade: not applicable.
Data repair: not applicable.
Operational caveats: rollback restores the previous default-on Skippy peer artifact-transfer behavior.

## Known residual risks
- This is an intentional stopgap until the unified blob-transfer layer lands; the follow-up tracking issue should remove or migrate the Skippy-specific gate at that point.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
@@ -640,6 +640,7 @@ impl Node {
         let my_explicit_model_interests = self.explicit_model_interests.lock().await.clone();
         let my_mesh_id = self.mesh_id.lock().await.clone();
         let my_owner_attestation = self.owner_attestation.lock().await.clone();
+        let my_owner_summary = self.owner_summary.lock().await.clone();
         let my_demand = self.get_demand();
         let stale_cutoff =
             std::time::Instant::now() - std::time::Duration::from_secs(PEER_STALE_SECS);
@@ -750,7 +751,7 @@ impl Node {
             served_model_runtime: my_model_runtime_descriptors,
             owner_attestation: my_owner_attestation,
             artifact_transfer_supported:
-                crate::models::artifact_transfer::artifact_transfer_enabled(),
+                crate::models::artifact_transfer::artifact_transfer_advertised(&my_owner_summary),
             stage_status_list_supported: true,
         });
         announcements
--- a/crates/mesh-llm-host-runtime/src/mesh/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/mod.rs
@@ -4592,17 +4592,22 @@ impl Node {
         peer_id: EndpointId,
         feature: &str,
     ) -> bool {
-        let state = self.state.lock().await;
-        let Some(peer) = state.peers.get(&peer_id) else {
+        let peer = {
+            let state = self.state.lock().await;
+            state.peers.get(&peer_id).cloned()
+        };
+        let Some(peer) = peer else {
             return false;
         };
         match feature {
             // Current PR peers advertise `stage-control` together with the
             // artifact-transfer feature. Older peers fall back to the legacy
             // skippy-stage ALPN control path.
-            skippy_protocol::STAGE_SUBPROTOCOL_FEATURE_STAGE_CONTROL
-            | skippy_protocol::STAGE_SUBPROTOCOL_FEATURE_ARTIFACT_TRANSFER => {
+            skippy_protocol::STAGE_SUBPROTOCOL_FEATURE_STAGE_CONTROL => {
                 peer.artifact_transfer_supported
+            }
+            skippy_protocol::STAGE_SUBPROTOCOL_FEATURE_ARTIFACT_TRANSFER => {
+                self.artifact_transfer_allowed_for_peer(&peer).await
             }
             skippy_protocol::STAGE_SUBPROTOCOL_FEATURE_STATUS_LIST => {
                 peer.stage_status_list_supported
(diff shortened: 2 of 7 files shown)


i386 (changes requested):
One thing I think we need to fix before landing: this now uses the artifact-transfer trust policy as part of split participant eligibility. With the new default of `MESH_LLM_ARTIFACT_TRANSFER` being disabled, a worker that is missing package artifacts gets excluded up front instead of being selected and then falling back to the local/HF resolver during prepare. That means default public-mesh nodes can only join package-backed splits if they already have the full package cached, which feels stricter than the intended 'no peer cache transfer unless trusted/open' behavior.\n\nI think the transfer policy should gate whether we dial peer artifact transfer, but it shouldn't make otherwise-valid HF-download-capable workers ineligible.

IvGolovach:
Good catch. I pushed a follow-up that separates split participant eligibility from the artifact-transfer trust policy: HF package-backed workers with missing artifacts can still be selected and fall back to local/HF prepare, while peer artifact transfer remains gated behind trusted/open policy. I also kept missing local-only packages gated unless transfer is allowed, and added focused tests for both paths. 🙂

PR #488, opened 2026-05-09
Emit fresh mesh invite tokens

i386:
## Summary
Mesh startup now emits the invite token from the node's current endpoint address at the moment the invite event is published, instead of reusing the token captured immediately after accepting starts.

## Impact
This reduces the chance of sharing a stale invite token if relay or public endpoint address information changes during startup.

## Validation
- `LLAMA_STAGE_BUILD_DIR=.deps/llama.cpp/build-stage-abi-metal cargo check -p mesh-llm`
- `git diff --check`


michaelneale (approved):
Worth a try

PR #491, opened 2026-05-09
fix(peer-latency): use QUIC transport latency measurements

ndizazzo:
# Use QUIC transport measurements for peer latency

## Summary

- Surface real peer latency from QUIC transport/path RTT instead of  seeing `0 ms`
- Add explicit latency provenance to `/api/status` so clients can distinguish direct measurements, estimated/transitive values, stale samples, and unknown latency
- Update the current dashboard and ui-preview network surfaces to show `Unknown`, `Estimated`, and `Stale` states with clearer tooltips and tests

## Why

Peer latency is operator-facing information, but the UI previously had cases where missing latency could look like a valid `0 ms` measurement. 

That made remote peers appear impossibly fast and hid whether a latency value was directly observed, inferred from another mesh node, stale, or unavailable. This branch separates display latency from routing-grade RTT so users get more truthful latency labels without weakening the existing routing behaviour that keeps the best known direct RTT for split decisions.

## What changed

- Added display-facing latency metadata to peer announcements and status payloads:
  - `latency_ms`
  - `latency_source` (`direct`, `estimated`, `unknown`)
  - `latency_age_ms`
  - `latency_observer_id`
- Added a 15-second RTT refresh task that reads iroh's maintained path RTT and updates peer display latency more frequently than the existing 60-second heartbeat.
- Preserved the existing routing RTT semantics: `rtt_ms` remains the routing-grade direct RTT, while display surfaces can use fresh direct samples or propagated estimates with age/provenance.
- Propagated latency observations through mesh gossip for transitive peers, including observer identity and sample age, while ignoring invalid/unknown propagated samples.
- Extended `/api/status` serialization to keep unknown latency explicit as `null` instead of dropping the field or implying `0 ms`.
- Updated ui-preview status adaptation, peer tables, mesh visualization, hover cards, and node drawer latency rendering to show direct/estimated/stale/unknown states.
- Updated the current dashboard status helpers, topology types, node sidebar, and tests to display latency hints/tooltips consistently.

## Protocol and compatibility

The protobuf changes are additive: new optional peer-announcement fields are assigned tags `36` through `39`, plus a new `LatencySource` enum. Existing fields are not removed or repurposed.

Compatibility tests cover both directions:

- legacy protobuf decoders ignore the new additive latency fields;
- current decoders accept legacy payloads that omit latency metadata;
- latency metadata round-trips through the current protobuf conversion path.

## Validation

- `cargo check -p mesh-llm`
- `npm run typecheck` in `crates/mesh-llm/ui-preview`
- `npm test -- src/features/network/components/PeersTable.test.tsx` in `crates/mesh-llm/ui-preview`
- LSP diagnostics clean for the touched ui-preview conflict-resolution files

Supersedes #460


i386:
Review notes:

- `cargo fmt --all -- --check` currently fails because `crates/mesh-llm-host-runtime/src/mesh/tests.rs` has a malformed `PeerAnnouncement` initializer around the scanned-model metadata test.
- The API serializes `latency_source` as lowercase strings, but the UI types/tests treat it as a numeric enum, so live values like `"estimated"`/`"unknown"` will not match the `LatencySource` switch cases.
- Unknown latency is still emitted as `latency_ms: 0`, which brings back the misleading `0 ms` display this PR is trying to remove. Unknown should stay `null`/absent for `latency_ms`.
- Gossip announcement builders set all new latency fields to `None`, so the propagated-latency consumer exists but no node actually advertises latency observations transitively.
- There is also a likely follow-on compile issue in `protocol/mod.rs`: one test initializer uses `available_metadata`, but the struct field is still `available_model_metadata`.

ndizazzo:
@copilot review the CI failures and correct cargo format checks so they pass

ndizazzo:
> Review notes:
> 
> - `cargo fmt --all -- --check` currently fails because `crates/mesh-llm-host-runtime/src/mesh/tests.rs` has a malformed `PeerAnnouncement` initializer around the scanned-model metadata test.
> - The API serializes `latency_source` as lowercase strings, but the UI types/tests treat it as a numeric enum, so live values like `"estimated"`/`"unknown"` will not match the `LatencySource` switch cases.
> - Unknown latency is still emitted as `latency_ms: 0`, which brings back the misleading `0 ms` display this PR is trying to remove. Unknown should stay `null`/absent for `latency_ms`.
> - Gossip announcement builders set all new latency fields to `None`, so the propagated-latency consumer exists but no node actually advertises latency observations transitively.
> - There is also a likely follow-on compile issue in `protocol/mod.rs`: one test initializer uses `available_metadata`, but the struct field is still `available_model_metadata`.
> 

@copilot examine these issues to determine if they need addressing in this PR, and summarize in a table in your reply. 

If there are items that should be corrected, make those adjustments now, and then run all tests, formatters, and clips to ensure it passes before making a commit and pushing to this branch.

i386:
Merge at will 👍

PR #492, opened 2026-05-09
Use fast Rust linker

i386:
## Summary
- require LLVM `lld` for Rust builds on every supported platform
- make development builds produce debug Rust binaries by default, while release recipes explicitly build release binaries
- update CI, release, and GPU cache workflows to install/use `lld` instead of `mold`

## Build Behavior
- Development: `just build` now builds `target/debug/mesh-llm` unless `MESH_LLM_BUILD_PROFILE=release` is set.
- Release: `just release-build`, GPU release recipes, and release workflows force `--release` / `BuildProfile release`.
- Linux: requires `ld.lld` and passes `-C link-arg=-fuse-ld=lld`.
- macOS: requires `ld64.lld` and passes an absolute `-fuse-ld=/path/to/ld64.lld`.
- Windows: requires `rust-lld.exe` or `lld-link.exe` for MSVC targets.

## Install Notes
- Linux: install your distro's `lld` package, for example `sudo apt-get install -y lld`, `sudo dnf install lld`, `sudo pacman -S lld`, or `sudo zypper install lld`.
- macOS: `brew install lld`.
- Windows: `rustup component add llvm-tools-preview`, or install LLVM with `winget install LLVM.LLVM` / `choco install llvm`.

## Validation
- `scripts/build-mac.sh` passed locally with `ld64.lld` and produced `target/debug/mesh-llm`.
- Linux Docker smoke passed with `ld.lld` and `RUSTFLAGS="-C link-arg=-fuse-ld=lld"`.
- `actionlint .github/workflows/ci.yml .github/workflows/release.yml .github/workflows/gpu-warm-cache-job.yml`
- `just --list`
- `bash -n scripts/build-linux.sh scripts/build-release.sh`
- `zsh -n scripts/build-mac.sh`
- `git diff --check`

## Compatibility
Builds now intentionally fail when `lld` is unavailable, with platform install instructions. Active Windows Rust CI jobs are not present on this branch, so Windows remains validated by script/config review rather than Actions coverage.


michaelneale:
what is benefit of lld? (could also have the justfile check and alert for it)

i386:
@michaelneale it makes the build process super fast. Why run onto two cylinders when you can run on eight

michaelneale (approved):

PR #493, opened 2026-05-09
Improve TUI token copying and text selection

i386:
## Summary

Users can now see clear feedback after copying the TUI join token, read long invite tokens in fullscreen without horizontal slicing, and use native terminal text selection while the dashboard is open.

## Root Cause

The terminal dashboard enabled mouse capture on startup, so terminals sent mouse gestures to the app instead of letting users select visible text. The join-token panel also rendered the token as a single horizontally sliced line, including in fullscreen, and copy feedback was only a terse title-bar state.

## Changes

- Stop enabling terminal mouse capture when entering the TUI so native text selection remains available.
- Show `Copied` / `Failed` feedback directly on the join-token copy control and expand the copied status text.
- Render the join token as wrapped text in fullscreen mode while keeping the compact dashboard view horizontally scrollable.
- Update the terminal dashboard smoke checklist to cover text selection.

## Validation

- `just build`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime tui_join_token --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime tui_full_screen_join_token --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime tui_enter_does_not_enable_mouse_capture --lib`


i386:
@ndizazzo for you

ndizazzo (commented):
@i386 Few things that I noticed:

1) I can't copy the token anymore using "c" without focusing the panel first - was this change intentional?
2) The "copied" state could clear after 5s for a better UX

> The join-token panel also rendered the token as a single horizontally sliced line, including in fullscreen

I don't see the fullscreen support on this branch - are you testing with a merged or rebased local copy?

PR #502, opened 2026-05-10
[codex] Box large runtime futures

i386:
## Summary

Prevents split-serving startup from putting the heaviest runtime futures directly on Tokio worker stacks.

The runtime entrypoint, startup model loops, and split topology coordinator now heap-box their large async state machines before awaiting them. This keeps behavior unchanged while reducing stack pressure in the path that handles `mesh-llm serve --model ... --split`.

## Root Cause

`clippy::large_futures` showed the top-level runtime future around 98 KB and the startup model loop future around 103 KB, with another large split coordinator future on the distributed split path. Those futures can consume a meaningful chunk of a Tokio worker stack when nested with split startup and dashboard/runtime work.

## Validation

- `cargo fmt --all -- crates/mesh-llm-host-runtime/src/lib.rs crates/mesh-llm-host-runtime/src/runtime/mod.rs crates/mesh-llm-host-runtime/src/runtime/local.rs`
- `cargo check -p mesh-llm`
- `just build`
- Focused `cargo clippy -p mesh-llm-host-runtime -- -W clippy::large_futures` grep confirmed no remaining large-future warnings for `runtime::run`, `startup_local_model_loop`, or the split coordinator `run()`.

Note: a direct `cargo test -p mesh-llm-host-runtime runtime::local --lib` without `LLAMA_STAGE_BUILD_DIR` failed to link `llama-common`; `just build` successfully prepared and built the supported Metal llama ABI path.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/lib.rs
+++ b/crates/mesh-llm-host-runtime/src/lib.rs
@@ -26,11 +26,11 @@ use std::time::Duration;
 pub const VERSION: &str = "0.65.1+skippy.20260504.kv.2";
 
 pub async fn run() -> Result<()> {
-    runtime::run().await
+    Box::pin(runtime::run()).await
 }
 
 pub async fn run_main() -> i32 {
-    match run().await {
+    match Box::pin(run()).await {
         Ok(()) => 0,
         Err(err) => {
             let _ = cli::output::emit_fatal_error(&err);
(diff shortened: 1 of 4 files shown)


michaelneale:
sitll working on this, latest one kind of works but hoping a more elegant solution appears overnight.

michaelneale:
solved elsewhere

PR #504, opened 2026-05-10
Fix worker stack overflow by heap-allocating large spawned futures

michaelneale:
## What changed

`startup_local_model_loop` is a 669-line async function that generates a multi-MB future state machine. It was spawned as:

```rust
tokio::spawn(async move { startup_local_model_loop(params).await })
```

This constructs the enormous future **on the Tokio worker thread's 2 MB stack**, causing stack overflow on macOS aarch64 with large models.

### The fix

```rust
tokio::spawn(Box::pin(startup_local_model_loop(params)))
```

The future is constructed on the **caller's stack** (`run_auto`, which runs on the main thread with 8 MB), then `Box::pin` immediately moves it to the heap. The Tokio worker only sees a `Pin<Box<...>>` (~16 bytes on stack).

Same treatment for `SplitTopologyCoordinator::run` (260 lines).

### Why the previous Box::pin attempt failed

The earlier attempt (`jd/box-large-runtime-futures` kv.3) did:

```rust
tokio::spawn(async move { Box::pin(f(params)).await })
```

This still overflows because the `async move` block runs on the worker — it constructs `f(params)`'s future on the worker stack *before* boxing it. The fix must ensure the future is **never constructed on a worker thread**.

### Regression gate

`main.rs` now reads `MESH_TOKIO_STACK_SIZE` to constrain Tokio worker stacks. CI runs the inference smoke test with **1 MB worker stacks** — any future regression that bloats a spawned future past ~1 MB will crash the smoke and fail the build.

No 64 MB thread stacks, no dedicated OS threads, no runtime architecture changes — just `Box::pin` at the right place.


ndizazzo (approved):
Supersedes #502

PR #514, opened 2026-05-11
Add split coordinator fencing and smarter Skippy serving defaults

i386:
## Summary

This PR makes split Skippy serving safer and more usable under real agent traffic.

Users running split inference now get explicit coordinator fencing for split topologies, split-aware context planning, bounded OpenAI generation admission, and smarter default lane/concurrency planning so a huge context window does not silently collapse serving to a single fragile lane.

## What Changed

- Added the `skippy-coordinator` crate for pure coordinator policy: coordinator claims, majority quorum math, monotonic term supersession, same-term conflict rejection, lease expiry, and load fencing.
- Added stage-control protocol fields/messages for coordinator claims, terms, coordinator IDs, and lease expiry.
- Updated split startup to claim coordinator ownership from a majority of planned stages before loading.
- Updated stage runtime control to reject unfenced/stale split loads, reject stale stops, and stop older running work when a newer coordinator term supersedes it.
- Kept local/non-split loads unfenced with term `0` for existing local stage flows.
- Made split and layer-package context planning use GGUF metadata and local layer fraction, so split stages size KV cache from their local layer share instead of falling back to small/default context.
- Simplified KV cache defaults to Q8_0 unless the user explicitly passes `--cache-type-k` / `--cache-type-v`.
- Co-planned context and Skippy concurrency so long-context models preserve at least useful lane capacity when affordable, while explicit `ctx_size` and `parallel` overrides still win.
- Added bounded OpenAI generation admission: immediate lane use when available, a bounded queue behind busy lanes, and retryable `429 rate_limit_exceeded` with `Retry-After` when saturated.
- Repaired host-runtime tests and expanded CI coverage so host-runtime and model-artifact library tests run in CI.

## Architecture

Coordinator ownership is intentionally not Raft. It is a fencing token for split runtime ownership:

- majority claim before load
- monotonic term replacement
- same-term conflict rejection
- stale load/stop rejection
- lease expiry

The `skippy-coordinator` crate stays pure policy. Mesh transport, protobuf conversion, iroh IDs, Tokio tasks, runtime process handling, and gossip remain in `mesh-llm-host-runtime`.

Context and concurrency planning now share the same KV budget model:

```text
KV memory ~= context_length * lane_count * kv_bytes_per_token(local layers)
```

The planner uses user overrides first. Without overrides, it uses GGUF metadata, KV quantization, and local layer fraction to choose a context/slot pair that preserves practical serving concurrency instead of always maximizing single-lane context.

## Protocol

This is a breaking Skippy stage-control protocol change. `ClaimCoordinator` and coordinator lease fields were added to the stage control surface, and split loads now expect claim/term fencing.

The public OpenAI error shape remains OpenAI-compatible. Rate-limit responses can now include `Retry-After` for bounded generation admission failures.

## Validation

- `cargo test -p skippy-coordinator --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime inference::skippy::stage::tests:: --lib`
- `RUST_MIN_STACK=8388608 LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime runtime::local::tests::load_split_runtime_generation_stops_candidate_stages_after_partial_load_failure --lib`
- `cargo check -p mesh-llm`
- `cargo test -p openai-frontend --lib`
- `git diff --check`

Full `mesh-llm-host-runtime --lib` was also run earlier. It compiled and ran, but still had unrelated/flaky failures in proxy streaming timing, console session-mode expectation, and an HF cache test that fetched a real snapshot instead of the fake fixture path.


michaelneale:
yeah seems reasonable - one thing I was thinking, say a 3rd node joins and was happily between 2 - does it rebalance if it is suboptimal in its ram/performance? or leave things as stable?

Issue #369, opened 2026-04-21
When a node hosts multiple instances of a model, the UX is unclear


ndizazzo:
As a test, I started 4 independent copies of the same model (so they all had the same amount of context available), and noticed that the UX could use some improvement:

**Topology View**
<img width="243" height="248" alt="Image" src="https://github.com/user-attachments/assets/9fdcb102-443e-4f09-925f-211ae72fc492" />

This element can use de-duplication with some kind of marker `(4)`, to reduce visual clutter.

**Node View**
<img width="663" height="1049" alt="Image" src="https://github.com/user-attachments/assets/9aa269d9-bfc1-471d-8d86-4d49fdfae145" />

This instance can be re-worked to list out each instance of a model, by adding a row to the table, and an "instance" number column.

**Model View**
<img width="653" height="602" alt="Image" src="https://github.com/user-attachments/assets/56cec8ad-9841-4974-bb5a-31b10db7269d" />
While we're in here, we can fix the overlapping `back` button.

PR #519, opened 2026-05-11
Allow multiple local model instances when capacity remains

### LINKED ISSUE #370: Starting multiple instances of a model on a node has unclear behaviour
## Summary

When attempting to define multiple instances of a model (separate llama.cpp) servers, the node does not correctly load them. The console output affirmatively sets up logs for each llama.cpp instance, and joins the mesh.

**However, the models stay cold and are not served.**

## Use Case
There are times when you still have VRAM available for a model, but want the model's FULL CONTEXT window available, instead of splitting the available context. 

This is different & a simple alternative to rope / yarn, which boosts the context window so you can use parallel. It won't be as performant due to having entirely separate caches, but I think it should still be supported.

## Proof

```terminal
⚡ API ready (bootstrap): http://localhost:9337
  Requests tunneled to mesh while GPU loads...
⏳ Starting rpc-server on port 36629... (logs: /run/user/1001/mesh-llm/runtime/319098/logs/rpc-server-36629.log)
⚡ Bootstrap proxy handing off to full API proxy
🔀 Multi-model mode: 5 additional model(s)
  + Qwen3.5-0.8B-UD-Q4_K_XL
  + Qwen3.5-0.8B-UD-Q4_K_XL
  + Qwen3.5-0.8B-UD-Q4_K_XL
  + Qwen3.5-2B-Q4_K_M
  + Qwen3.5-2B-Q4_K_M
🗳 [Qwen3.5-0.8B-UD-Q4_K_XL] Running as host (65.9GB capacity for 0.6GB model, serving entirely)
  Serving entirely (66GB capacity)
⏳ Starting llama-server... (logs: /run/user/1001/mesh-llm/runtime/319098/logs/llama-server.log)
🗳 [Qwen3.5-0.8B-UD-Q4_K_XL] Running as host (65.9GB capacity for 0.6GB model, serving entirely)
  Serving entirely (66GB capacity)
⏳ Starting llama-server... (logs: /run/user/1001/mesh-llm/runtime/319098/logs/llama-server.log)
🗳 [Qwen3.5-0.8B-UD-Q4_K_XL] Running as host (65.9GB capacity for 0.6GB model, serving entirely)
⏳ Starting llama-server...
  Serving entirely (66GB capacity)
⏳ Starting llama-server... (logs: /run/user/1001/mesh-llm/runtime/319098/logs/llama-server.log)
🗳 [Qwen3.5-2B-Q4_K_M] Running as host (65.9GB capacity for 1.3GB model, serving entirely)
  Serving entirely (66GB capacity)
⏳ Starting llama-server... (logs: /run/user/1001/mesh-llm/runtime/319098/logs/llama-server.log)
🗳 [Qwen3.5-2B-Q4_K_M] Running as host (65.9GB capacity for 1.3GB model, serving entirely)
  Serving entirely (66GB capacity)
⏳ Starting llama-server... (logs: /run/user/1001/mesh-llm/runtime/319098/logs/llama-server.log)
🗳 [Qwen3.5-0.8B-UD-Q4_K_XL] Running as host (65.9GB capacity for 0.6GB model, serving entirely)
  Serving entirely (66GB capacity)
⏳ Starting llama-server... (logs: /run/user/1001/mesh-llm/runtime/319098/logs/llama-server.log)
```

Yet later, when using `mesh-llm status`, the output shows nothing runs:

```
> ./target/release/mesh-llm status
⚙️  Runtime

📦 Models served locally: 0

No local models are currently being served.
```

**Yet, the node appears in the mesh with the models that have been specified registered:**

Image

Compare this to a node where I have not specified duplicate models:

```
./target/release/mesh-llm status
⚙️  Runtime

📦 Models served locally: 2

Model                                      Backend  State      Pid      Port  
gemma-4-26B-A4B-it-UD-Q4_K_XL              Llama    Ready      54386    43661 
Qwen3.6-35B-A3B-UD-Q4_K_XL                 Llama    Ready      54387    35507
```

---

Changing the configuration back for the node to supply two unique models, the issue stops:

Image

IvGolovach:
## Summary

Allow a node to run multiple local runtime instances of the same model, or multiple local runtime instances across different models, as long as the aggregate local runtime capacity still fits.

This addresses the backend behavior reported in #370: duplicate local model starts could appear to launch, but the node did not reliably keep those runtime instances admitted, tracked, and unloadable as distinct local processes. The UI clarity work from #369 is intentionally left as a follow-up rather than mixed into this backend correctness PR.

## Why

Operators sometimes want independent copies of a model so each copy keeps its full context window instead of sharing capacity through one runtime instance. That should be supported when the machine still has enough local runtime memory.

The previous admission path checked each start against the full node capacity independently, and one runtime-control duplicate check was keyed by model name instead of instance identity. That made duplicate local instances ambiguous: capacity accounting did not represent the aggregate local runtime footprint, and duplicate instances could be rejected or lose correct lifecycle ownership.

## What Changed

- Add an instance-aware `RuntimeCapacityLedger` for local runtime starts.
- Reserve local runtime capacity by `instance_id`, not by model name.
- Allow duplicate model instances when there is remaining capacity in the relevant local capacity pool.
- Reject overcommit with an explicit capacity error that reports required, reserved, available, total capacity, and shortfall.
- Keep pinned GPU reservations isolated from the node-wide capacity pool.
- Release reservations on unload, unexpected runtime exit, shutdown, and split cutover.
- Pass the remaining capacity budget into local runtime planning so each later instance plans against the memory left after earlier local reservations.
- Update stale protocol/mesh test fixtures that had drifted from the current descriptor shape.

## Architecture

The new ledger is deliberately local to runtime orchestration. It does not change mesh gossip semantics, routing contracts, or the public API payload shape.

Reservations are RAII guards owned by runtime model handles or startup-local runtime tasks. Dropping the guard releases capacity. Same-instance replacement uses a generation check so an older stale guard cannot accidentally release a newer reservation for the same `instance_id`.

The PR covers full local runtime starts and local fallback starts after split-stage loss. Split-stage aggregate placement itself is not changed here.

## Protocol and Compatibility

No mesh wire protocol change.

No gossip schema change.

No mixed-version compatibility break: capacity admission remains local process state only. Older peers continue to see the same model/runtime gossip surface they already understand.

## Validation

Local validation was run from the PR branch against `origin/main`.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS
- `git diff --check`: PASS
- `git diff --cached --check`: PASS
- `cargo check -p mesh-llm-host-runtime`: PASS
- `cargo check -p mesh-llm`: PASS
- `cargo test -p mesh-llm-host-runtime runtime::capacity --lib`: PASS, 5 passed
- `cargo test -p mesh-llm-host-runtime duplicate --lib`: PASS, 6 passed
- `cargo test -p mesh-llm-host-runtime process_snapshots_keep_distinct_same_model_instances --lib`: PASS
- `cargo test -p mesh-llm-host-runtime gossip_frame_roundtrip_preserves_scanned_model_metadata --lib`: PASS
- `cargo test -p mesh-llm-host-runtime transitive_peer --lib`: PASS, 5 passed
- `cargo test -p mesh-llm-host-runtime owner_fields_roundtrip_through_proto_announcement --lib`: PASS
- `cargo test -p mesh-llm-host-runtime test_peer_announcement_first_joined_mesh_ts_roundtrip --lib`: PASS
- `cargo test -p mesh-llm-host-runtime initial_pretty_session_mode_allows_dashboard_for_runtime_surfaces --lib`: PASS

Local note: Rust checks used the repo's local llama stage ABI build directory through `LLAMA_STAGE_BUILD_DIR`; the machine-specific absolute path is intentionally omitted from the PR text.

Not completed locally:

- Full `cargo test -p mesh-llm-host-runtime --lib` was not completed locally after encountering an unrelated proxy streaming test failure in `crates/mesh-llm-host-runtime/src/runtime/proxy/tests.rs:1182` and long-running HF/cache tests. The targeted tests for the touched runtime, status, duplicate-instance, gossip fixture, and protocol fixture paths passed.

Remote CI:

- Running on GitHub for the published PR.

## Related

- Closes #370
- Related follow-up: #369, for UI deduplication and clearer display of multiple instances of the same model

## Rollback

Revert this PR.

DB downgrade: not applicable.

Data repair: not applicable.

Operational caveats: any local runtime instances admitted only because of this change will no longer be admitted after rollback.

## Known Residual Risks

- This PR is backend/runtime correctness only. It does not attempt the topology and model-view UI cleanup tracked in #369.
- Split-stage aggregate placement is not reworked here; this change reserves full local runtime starts and local fallback starts.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/api/status.rs
+++ b/crates/mesh-llm-host-runtime/src/api/status.rs
@@ -819,7 +819,10 @@ pub(crate) fn classify_runtime_error(msg: &str) -> u16 {
         404
     } else if msg.contains("already loaded") || msg.contains("multiple loaded instances") {
         409
-    } else if msg.contains("fit locally") || msg.contains("runtime load only supports") {
+    } else if msg.contains("fit locally")
+        || msg.contains("runtime load only supports")
+        || msg.contains("runtime capacity")
+    {
         422
     } else {
         400
--- a/crates/mesh-llm-host-runtime/src/api/tests.rs
+++ b/crates/mesh-llm-host-runtime/src/api/tests.rs
@@ -517,6 +517,10 @@ fn test_classify_runtime_error_codes() {
         classify_runtime_error("runtime load only supports models that fit locally"),
         422
     );
+    assert_eq!(
+        classify_runtime_error("runtime capacity for model 'x' exceeds node pool"),
+        422
+    );
     assert_eq!(classify_runtime_error("bad request"), 400);
 }
 
--- a/crates/mesh-llm-host-runtime/src/runtime/capacity.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/capacity.rs
@@ -1,0 +1,454 @@
+use std::collections::HashMap;
+use std::fmt;
+use std::sync::{Arc, Mutex};
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub(super) enum RuntimeCapacityPool {
+    Node,
+    PinnedGpu(String),
+}
+
+impl fmt::Display for RuntimeCapacityPool {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Node => f.write_str("node"),
+            Self::PinnedGpu(stable_id) => write!(f, "pinned GPU {stable_id}"),
+        }
+    }
+}
+
+impl RuntimeCapacityPool {
+    fn overlaps(&self, other: &Self) -> bool {
+        // Node-wide local loads are not pinned to a backend device, so they
+        // share the physical placement domain with every pinned GPU pool.
+        // Pinned GPU pools remain independent from each other.
+        self == other || matches!((self, other), (Self::Node, _) | (_, Self::Node))
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(super) struct RuntimeCapacityRequest {
+    pub(super) instance_id: String,
+    pub(super) model_name: String,
+    pub(super) pool: RuntimeCapacityPool,
+    pub(super) capacity_bytes: u64,
+    pub(super) required_bytes: u64,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct RuntimeCapacityAllocation {
+    model_name: String,
+    pool: RuntimeCapacityPool,
+    required_bytes: u64,
+    generation: u64,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(super) struct RuntimeCapacityError {
+    pub(super) model_name: String,
+    pub(super) pool: RuntimeCapacityPool,
+    pub(super) capacity_bytes: u64,
+    pub(super) reserved_bytes: u64,
+    pub(super) required_bytes: u64,
+}
+
+impl RuntimeCapacityError {
+    pub(super) fn available_bytes(&self) -> u64 {
+        self.capacity_bytes.saturating_sub(self.reserved_bytes)
+    }
+
+    pub(super) fn shortfall_bytes(&self) -> u64 {
+        self.required_bytes.saturating_sub(self.available_bytes())
+    }
+}
+
+impl fmt::Display for RuntimeCapacityError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "runtime capacity for model '{}' exceeds {} pool: requires {}, available {}, reserved {}, capacity {}, short by {}",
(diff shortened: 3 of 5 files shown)


ndizazzo (changes requested):
@IvGolovach Because I use a config file, one thing I found is that the ledger only sums reservations from the same pool, but runtime-control loads reserve the Node pool with pinned_gpu: None.

That means a startup model pinned to a GPU can reserve most of that GPU’s VRAM, then an unpinned runtime-control load can still be admitted against the empty node-wide pool and try to use the same physical GPU.

I think node-wide reservations need to account for overlapping pinned-GPU reservations, at least on single-GPU hosts. Otherwise we can over-admit local models even though the physical GPU capacity is already reserved.

ndizazzo (approved):

PR #523, opened 2026-05-11
Add Skippy WAN Docker lab

i386:
## Summary

This adds a reproducible Docker lab for testing Skippy staged serving over WAN-like links. Operators can now bring up a four-stage CPU-backed chain with Linux `tc netem` shaping, query the OpenAI-compatible endpoint on `:9337`, attach an interactive `skippy-prompt binary` REPL to stage0, and collect run telemetry through `metrics-server`.

The lab is intentionally launched through `docker/skippy-wan-lab/up.sh`. That script makes sure `meshllm/gemma-4-26B-A4B-it-UD-Q4_K_M-layers` exists in the host Hugging Face cache and verifies the package manifest, layer count, activation width, required artifacts, and artifact byte sizes before Docker starts. Containers then mount the host cache read-only, so a missing or partial model package fails early instead of producing a half-started lab.

## What Changed

- Added `docker/skippy-wan-lab/` with a compose topology for `metrics`, `stage0`, `stage1`, `stage2`, and `stage3`.
- Added WAN shaping configuration through `.env` / `.env.link`, including `WAN_RTT_MS`, `WAN_DELAY_MS`, `WAN_RATE_MBIT`, and `WAN_ENABLE`.
- Added `docker/skippy-wan-lab/up.sh` as the supported launcher and host-cache preflight path.
- Added `docker/skippy-wan-lab/prompt.sh` to attach an interactive prompt session to a running stage0 container without starting containers or downloading model artifacts.
- Added `scripts/skippy-wan-calibrate.sh` to derive WAN delay settings from a target host and optionally record `iperf3` bandwidth.
- Added `SKIPPY_PROTOCOL_TODO.md` to track RTT-reduction work for prefill/decode pipelining, reply paths, speculative windows, transport efficiency, and staged state lifecycle costs.
- Added Just targets for the WAN lab workflow.

## Architecture

The Docker lab uses Linux traffic control inside the containers rather than Skippy's artificial downstream delay flags. `stage0` owns the OpenAI frontend and first layer range, then forwards binary activation traffic through `stage1`, `stage2`, and `stage3`. Telemetry is emitted over OTLP/gRPC to the colocated `metrics-server`, with report data stored in a Docker volume.

The model package remains host-owned. Docker only receives a read-only cache mount after the launcher has downloaded and validated the layer package, which keeps package integrity checks outside the container lifecycle.

## Usage

```bash
cp docker/skippy-wan-lab/.env.example docker/skippy-wan-lab/.env
scripts/skippy-wan-calibrate.sh 100.90.121.70 docker/skippy-wan-lab/.env.link
docker/skippy-wan-lab/up.sh
```

Attach an interactive prompt once the lab is running:

```bash
docker/skippy-wan-lab/prompt.sh --max-new-tokens 64 --no-think
```

Check the OpenAI-compatible endpoint:

```bash
```

## Validation

Not run per request. I did not start Docker, run the lab, or download the Hugging Face package after the request to only set this up for a draft PR. I only ran shell syntax checks for the helper scripts.


michaelneale (approved):
excellent idea - helpful too

PR #532, opened 2026-05-12
Pipeline embedded stage0 prefill replies

i386:
## Summary

This pipelines embedded stage0 prefill ACK handling so stage0 can keep computing and forwarding later prefill chunks while earlier downstream ACKs are still pending.

- Adds a `reply_credit_limit` path to embedded OpenAI stage0 generation and uses a default prefill credit window of 3.
- Defers non-final prefill ACK reads until the credit window is full or prefill is complete.
- Drains pending prefill ACKs before cancellation return and before decode continues, preserving downstream cache stats.
- Emits debug telemetry for each prefill chunk, including credit limit, pending replies, credit waits, and deferred replies drained.

## Why

The previous embedded stage0 path waited for a downstream ACK after every forwarded prefill chunk. On shaped WAN links, that exposed an RTT per chunk even when stage0 had more local prefill work it could compute. This overlaps stage0 compute and downstream ACK latency without changing the lab topology.

## Validation

- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-cpu cargo test -p skippy-server --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-cpu cargo test -p mesh-llm inference::skippy --lib`

Note: `cargo test -p skippy-server --lib` without `LLAMA_STAGE_BUILD_DIR` fails locally because the native llama static libraries are not on the default linker path.


ndizazzo (changes requested):
Needs rebasing after #538 and #536

i386:
@ndizazzo rebased 👍

PR #534, opened 2026-05-12
Mixture-of-Agents gateway

michaelneale:
## Mixture-of-Agents gateway

Use `model: "mesh"` to fan out requests to all models in the mesh in parallel, arbitrate with deterministic logic, and return one coherent response. The client thinks it talks to one model.

```
Agent / Goose / pi
    │
    │  POST /v1/chat/completions { "model": "mesh" }
    ▼
 Mesh proxy (ingress.rs)
   │  intercepts model == "mesh", builds GatewayConfig
   ▼
 MoA handle_turn()
   ├─ role-shaped context slicing (real agent context, not synthetic)
   ├─ parallel fan-out via ModelBackend trait
   ├─ early-exit on consensus (cancel slow workers)
   ├─ deterministic arbiter (code, not models)
   ├─ reducer escalation only on genuine conflict
   └─ native tool_call support through full pipeline
```

### What's new

- **`model: "mesh"`** — virtual routing directive (like `auto`), not advertised in `/v1/models`. Activates when ≥2 distinct models are available.
- **`mesh-mixture-of-agents` crate** — standalone, zero mesh dependency, 2714 LOC, 29 unit tests. Testable against any OpenAI-compatible endpoint via `HttpBackend`.
- **`ModelBackend` trait** — crate defines the abstraction; host runtime implements `LocalModelBackend` (direct HTTP to skippy port) and `RemoteModelBackend` (HTTP-over-QUIC tunnel).
- **`handle_turn()`** — stateless free function, no persistent `Gateway` struct. Fresh per request, agent client manages multi-turn.
- **`mesh_hooks: false`** on all worker requests — prevents recursive virtual LLM consultations.
- **Early-exit consensus** — 2+ agreeing workers → return immediately, abort the rest. Failed workers trigger early-exit when only one survivor remains.
- **Tool calling through the full pipeline** — workers propose tools via arbiter, tool results go to reducer only.
- **SSE streaming** — non-streaming arbitration wrapped as SSE for streaming clients.

### Context packing

Workers get slices of the real agent context — the agent's actual system prompt, messages, and tool definitions. The gateway varies depth per role, not content.

| Role | Messages | Tools | Max tokens |
|------|:---:|:---:|:---:|
| Fast | last user msg | names only | 256 |
| Specialist | last 4 | name + description | 512 |
| Strong | last 10 | full schemas (native) | 1024 |
| Reducer | worker outputs | full schemas (native) | 2048 |

### Transport

| Backend | Transport | Use |
|---------|-----------|-----|
| `LocalModelBackend` | Direct HTTP to `127.0.0.1:{skippy_port}` | Local model on this node |
| `RemoteModelBackend` | HTTP-over-QUIC tunnel | Remote mesh peer |
| `HttpBackend` | Plain HTTP to any URL | Standalone testing |

### Live test results

Tested with GLM-4.7-Flash (local) + Qwen3-8B, Qwen3.5-9B, MiniMax-M2.5 (remote mesh peers):

| Test | Time | Result |
|------|-----:|--------|
| Factual ("capital of Japan") | 2.0s | ✅ Correct |
| Tool call ("read README.md") | 2.3s | ✅ `read_file({"path":"README.md"})` |
| Reasoning (syllogism) | 7.9s | ✅ Correct logic |
| Streaming ("7*8") | 1.8s | ✅ SSE chunks |
| Goose agent integration | — | ✅ Tool calls end-to-end |

### Relationship to existing systems

- **`auto`** routes to best single model. **`mesh`** fans out to ALL models.
- **Hooks** (`virtual_llm.rs`) fire reactively during inference. **MoA** fires proactively before inference. Independent, complementary.
- **Pipeline** (`pipeline.rs`) does 2-model plan→execute for code tasks. Still active at ingress line 279.
- **Consult** (`consult.rs`) does single peer consultation. MoA does parallel multi-peer.

### Crate structure

`crates/mesh-mixture-of-agents/` — 6 modules, 29 tests:

| Module | Purpose |
|--------|---------|
| `lib.rs` | `handle_turn()`, `ModelBackend` trait, `GatewayConfig`, fan-out |
| `arbiter.rs` | Deterministic arbitration + early-exit (11 tests) |
| `normalize.rs` | 3-tier dirty output parsing: JSON → KV → heuristic (12 tests) |
| `session.rs` | Canonical transcript, tool tracking, turn classification (3 tests) |
| `context.rs` | Role-shaped context packing from real agent context |
| `worker.rs` | Role assignment, think-tag/reasoning stripping (3 tests) |

Design doc: `docs/design/MOA_GATEWAY.md`


ndizazzo (commented):
Agent found some things to address:

- **High** — tool-result turns drop tool output. handle_turn creates a fresh Session per request, but pack_for_tool_result_turn only includes session.recent_tool_results(). Follow-up OpenAI tool-result requests therefore lose the actual role:"tool" content and reducer only sees the last user text.
-  **High** — /v1/responses + model:"mesh" returns chat-completions schema. The ingress MoA intercept bypasses request.response_adapter, sending chat.completion JSON/SSE chunks directly, so Responses API clients will fail to parse MoA replies.
- **Medium** — Docker/Fly image builds omit crates/spec-prefill-poc. Root workspace includes the new crate, but both Dockerfiles only copy mesh-mixture-of-agents; Cargo resolves all workspace members before building and will fail on the missing path.
- **Medium** — malformed confidence: NaN can panic request handling. KV parsing accepts NaN into WorkerOutput.confidence, then partial_cmp(...).unwrap() in arbiter/best-answer paths can panic.
- **Low** — spec-prefill-poc acceptance metrics skip the first proposed token. It verifies draft tokens after prompt prefill but compares predicted[i] to target_draft_tokens[i + 1], so immediate divergence is not counted correctly.

michaelneale:
Superseded by #566 — fresh branch  off latest , with the prefill-speculative work split out into its own branch/PR. Closing this PR; please review #566 instead.\n\nDifferences vs this branch:\n- Dropped the auto-retry-5xx-fallback commits (PR #563 — also closed; not the right shape of fix for the goose-on-auto streaming case).\n- Dropped the prefill-speculative commits; moved to their own draft PR for separate review.\n- Added live-test fixes from today: size-aware role assignment, hallucinated-tool-name filtering, reducer-candidate fallback on 5xx.

PR #536, opened 2026-05-12
Split skippy-server frontend module

i386:
## Summary

- Rename the skippy-server OpenAI-named module to `frontend` so the module name matches its role as the compatible serving adapter.
- Split the former monolithic frontend implementation into focused modules for backend dispatch, generation flow, local/binary/embedded execution, prefix cache handling, prefill planning, speculative verification, prompting, and wire message construction.
- Update mesh host runtime imports to use the crate-root context budget export instead of `skippy_server::openai`.

## Validation

- `cargo check -p mesh-llm`

## Notes

- `cargo test -p skippy-server --lib` is currently blocked in this environment because `skippy-ffi` cannot find the native static library `llama-common`; the failure occurs before tests run.


ndizazzo (approved):
This seems g2g to me - as long as no consumers of `skippy_server::openai::...` are broken after the decomp

PR #537, opened 2026-05-12
Trim single-stage Skippy decode overhead

i386:
## Summary
Single-stage Skippy decode now avoids debug-only and hook-only work on the normal serving path. This brings Skippy decode throughput back to roughly llama-server parity for the Qwen3-4B single-stage benchmark sweep.

## Root Cause
The single-stage Skippy OpenAI path was doing extra per-token bookkeeping even when the request did not enable mesh hooks and telemetry was not in debug mode. Every decoded token still collected generation signals, built debug telemetry attribute maps, and measured per-token timing fields before discovering that nothing would consume them.

That overhead was small per token, but large enough to show up as a steady decode regression. In the benchmark run, Skippy was about 9-10% slower than vanilla llama for 512-4096 token prompts, while prefill and end-to-end first-token time were already near parity. That pointed at per-token frontend/runtime bookkeeping rather than model load, Metal offload, or prompt processing.

## What Changed
- Added a `Telemetry::is_debug_enabled()` predicate.
- Added a `generation_hooks_active()` helper for the Skippy OpenAI frontend.
- Gated per-token generation signal collection behind active mesh hooks.
- Gated per-token decode telemetry attribute construction behind debug telemetry.

Hook behavior is preserved when mesh hooks are enabled, and detailed per-token debug spans are still available when telemetry level is `Debug`.

## Benchmark
Host: `studio54.local`  
Model: `unsloth/Qwen3-4B-GGUF:Q4_K_M`  
Runtime shape: single-stage Skippy vs vanilla `llama-server`  
Prompt tokens: `128 512 1024 2048 4096`  
Generation: `64` tokens, `3` runs, concurrency `1`  
Context: `8192`  
Metal offload: `37/37 layers to GPU` for both runtimes

| prompt | new Skippy tok/s | old Skippy tok/s | llama tok/s |
|---:|---:|---:|---:|
| 128 | 98.22 | 88.78 | 70.96 |
| 512 | 96.76 | 87.98 | 98.13 |
| 1024 | 95.14 | 86.19 | 95.67 |
| 2048 | 92.67 | 83.37 | 91.76 |
| 4096 | 87.26 | 77.12 | 84.29 |

Raw patched-run artifacts are on the benchmark host at:

```text
/tmp/skippy-lean-decode-bench-20260513-083906
```

The baseline comparison run is at:

```text
/tmp/skippy-vs-llama-bench-20260513-082951
```

## Validation
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p skippy-server --lib`
- Rebuilt patched `skippy-server` on `studio54.local` and reran the focused benchmark sweep above.

### DIFF
--- a/crates/skippy-server/src/frontend/local_generation.rs
+++ b/crates/skippy-server/src/frontend/local_generation.rs
@@ -359,6 +359,9 @@ impl StageOpenAiBackend {
                 .expect("checked non-empty prompt");
             let mut hook_request = request.hook_request;
             let hook_runtime = request.hook_runtime;
+            let generation_hooks_active =
+                self.generation_hooks_active(&hook_request, hook_runtime.as_ref());
+            let emit_token_debug = self.telemetry.is_debug_enabled();
             let mut post_prefill_hook_checked = false;
             let mut last_mid_generation_hook_at = None;
             while decoded_tokens < request.max_tokens as usize {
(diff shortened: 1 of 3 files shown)


ndizazzo (approved):
We have a `--debug` flag now in the mesh-llm binary, that generates additional debug log output when hosting a model.

Suggest gating your token benchmark changes to pass it to mesh-llm, and wire `Telemetry::is_debug_enabled()` up to presence of that flag

PR #538, opened 2026-05-12
Wire mesh debug to Skippy telemetry

i386:
Users can now opt in to Skippy's detailed token-level telemetry through the existing `mesh-llm --debug` path, while normal mesh serving keeps embedded Skippy telemetry disabled.

Follows #537, which is now merged into `main`.

## What was wrong

#537 made per-token Skippy debug work conditional on `Telemetry::is_debug_enabled()` so production decode avoids building debug timing maps on every token. In the mesh-hosted path, embedded Skippy was still always constructed with `TelemetryLevel::Off`, so mesh's `--debug` flag could not make that condition true. That meant the new debug gate was effectively unreachable when hosting a model through `mesh-llm`.

## What changed

- Added embedded Skippy telemetry options with an off-by-default configuration.
- Wired `mesh-llm --debug` to use Skippy `TelemetryLevel::Debug` for local embedded Skippy model loads.
- Added a hidden `--skippy-metrics-otlp-grpc` endpoint for sending those debug spans to `metrics-server` during diagnostic runs.
- Propagated the same telemetry setting through direct GGUF, layer-package, and split stage-0 local load paths.

## CLI

Diagnostic collection can now be run as:

```bash
metrics-server serve \
  --db /tmp/metrics.sqlite \
  --http-addr 127.0.0.1:18080 \
  --otlp-grpc-addr 127.0.0.1:14317

mesh-llm --debug \
  --skippy-metrics-otlp-grpc http://127.0.0.1:14317 \
  serve --model /path/to/model.gguf
```

Production benchmarks should continue to omit `--debug`, because debug mode intentionally restores per-token instrumentation overhead.

## Validation

- `cargo check -p mesh-llm`
- Attempted `cargo test -p mesh-llm --lib`; local link failed before tests because this worktree does not have the prepared llama.cpp static library (`llama-common`).

### DIFF
--- a/crates/mesh-llm-host-runtime/src/cli/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/cli/mod.rs
@@ -247,6 +247,10 @@ pub(crate) struct Cli {
     #[arg(long)]
     pub(crate) debug: bool,
 
+    /// OTLP/gRPC endpoint for embedded Skippy debug telemetry, for example http://127.0.0.1:14317.
+    #[arg(long, hide = true)]
+    pub(crate) skippy_metrics_otlp_grpc: Option<String>,
+
     /// Show all options (including advanced/niche ones).
     #[arg(long, hide = true)]
     pub(crate) help_advanced: bool,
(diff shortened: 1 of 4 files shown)


ndizazzo (approved):

michaelneale (approved):
nice

PR #543, opened 2026-05-13
Add min_p support and align skippy sampling defaults with llama-server

michaelneale:
Skippy's sampling defaults were wide open (temp=1.0, top_p=1.0, top_k=0, no min_p) while llama-server defaults to temp=0.8, top_p=0.95, top_k=40, min_p=0.05.

This caused models with forced `<think>` templates (MiniMax M2.5, Qwen3.5) to degenerate into repetitive loops on ambiguous prompts — the unconstrained sampling let models wander into low-probability repetitive dead ends.

**What changed:**
- Add min_p sampler to the skippy ABI (replaces the unused `reserved` field)
- Wire min_p through FFI → runtime → protocol → request parsing
- Align default sampling with llama-server: temp=0.8, top_k=40, top_p=0.95, min_p=0.05
- Clients can override all params including min_p via the request body

**Tested locally** with Qwen3.5-4B — the degenerate repetition loops are gone. Extended thinking on ambiguous prompts still runs long (model behavior, same on llama-server), but the output is coherent rather than looping.


ndizazzo (approved):
Oooh! Thank you. I had some configs I needed to drop this from. Nice change

PR #545, opened 2026-05-13
fix(ui): many UI fixes after console was deployed

ndizazzo:
* Chats correctly persist from IndexDB
* Default system prompt is now correctly applied
* Version now displays correctly
* Updated Fly dockerfile to allow for the VITE server host
* Fixed mobile responsiveness when chatting and keyboard avoidance
* Implemented markdown table rendering
* Implemented code syntax highlighting
* Restored hidden node age field


i386:
Small thing I noticed: `ChatLayout` now shows the floating sidebar button whenever `!showDesktopSidebar`, even when `hideSidebar` is true. On the empty/first-chat state `ChatPage` passes `hideSidebar`, so this can leave an "Open chat sidebar" button over the composer that opens an empty sidebar. Probably just needs the old guard back: `!hideSidebar && !showDesktopSidebar`.

ndizazzo:
@i386 Good catch. Didn't see it because my chats were being restored. Was able to repro after clearing: 
image

fix inc.

PR #546, opened 2026-05-13
Fix chat persistence, mobile sidebar, public mesh detection

michaelneale:
## What changed

Chat history now survives page refreshes, the mobile sidebar is always accessible, and the Fly console correctly shows public mesh branding.

### Chat persistence
Chat history was silently lost on every refresh. The old UI created the IndexedDB store with `keyPath: "id"` (v1), but the new UI expected no keyPath and used different record keys. Both declared version 1 so the upgrade never fired — saves threw `DataError` (silently caught) and loads returned `undefined`.

**Fix:** Bump DB to version 2 with a migration that reads old v1 data before the upgrade, recreates the store without keyPath, then writes migrated data under the new key `"chat-state:live"`. Existing users get their old history back.

### Mobile chat sidebar
The floating action button for the conversation list was hidden when `conversations.length === 0`, which made chat history unreachable on narrow viewports (and on fresh load before persistence worked). Now always shown on mobile.

### Public mesh detection
The Fly console runs `--client --auto` which joins the public mesh via nostr, but `publication_state` stayed `"private"` because the publish/watchdog paths skip clients. The UI showed private-mesh invite flows instead of "Welcome to the public mesh." Now sets `publication_state: "public"` for clients with `nostr_discovery: true`.

### Docs links
Removed broken path-based links (`/agents`, `/models` → 404). Kept working anchor links (`#install`, `#agents`). Removed dead "Common patterns" footer link.

### Dead code
Removed `hooks/useChatSession.ts` — nobody imported it, but it contained the conflicting v1 IndexedDB schema.


ndizazzo:
This PR now targets #545 to take both combined sets of changes

ndizazzo (approved):

PR #525, opened 2026-05-12
Add network doctor for direct mesh UDP

i386:
## Summary

Users can now run `mesh-llm network doctor --bind-port <port>` to diagnose direct mesh UDP readiness from the local machine.

The doctor command reports the public IPv4 candidate discovered by iroh, the local router WAN IPv4 reported by UPnP/IGD, whether those addresses match, iroh direct candidate counts, and iroh portmapper counters. This makes CGNAT, double NAT, missing public candidates, and fixed-port forwarding mistakes visible without falling back to external STUN services.

Startup also warns when a fixed `--bind-port` is used and the router appears to sit behind CGNAT/private upstream NAT while iroh discovers a different public IPv4.

## User-Facing Changes

- Adds `mesh-llm network doctor --bind-port <port>`.
- Adds `mesh-llm network doctor --bind-port <port> --json`.
- Adds `direct_connectivity` to `/api/status` with iroh direct candidates and portmapper counters.
- Adds a startup `info` event summarizing iroh direct candidates and portmapper counters.
- Adds `docs/CGNAT.md` and links it from the docs index.

## Implementation

- Uses iroh relay-discovered endpoint candidates as the public-address source for invite-port rewriting and direct-connectivity diagnostics.
- Does not query external STUN servers; iroh's relay/discovery path is the source of public endpoint candidates.
- Treats `100.64.0.0/10` shared address space as non-public invite candidates.
- Uses UPnP/IGD `GetExternalIPAddress` as a best-effort router WAN check when `--bind-port` is set.
- Emits a normal `warning` event when UPnP reports a private/shared WAN address while iroh has discovered a different public IPv4.
- Keeps UPnP/direct-connectivity heuristics in `mesh::direct_connectivity` instead of growing `mesh/mod.rs`.

## Doctor Output

Live run on this network, with public IPs obscured here:

```text
mesh-llm network doctor --bind-port 53238

🩺 Direct UDP readiness
  bind port:               53238
  🌍 iroh public IPv4:     159.196.xxx.xxx
  🛜 router WAN IPv4:      159.196.xxx.xxx
  ✅ router WAN type:      public IPv4
  📣 invite candidate:     159.196.xxx.xxx:53238
  ✅ UPnP/IGD:             discovered router
  ✅ WAN/public match:     router WAN matches iroh public candidate
  🧭 iroh candidates:      5
  🗺️ iroh portmapper:      attempts=2, external_updates=1

[...]

PR #547, opened 2026-05-13
Add explicit bind IP selection for mesh QUIC

### LINKED ISSUE #542: [feat] bind-ip
`mesh-llm serve --split` cannot form a working multi-node mesh on Linux hosts that have more than one kernel-visible network interface where any of those interfaces share an identically-numbered address with the same interface on other hosts   (the textbook example: every host has a `docker0` at `172.17.0.1` / `172.23.0.1`).

Workers say `OK Joined mesh` and form direct connections **among themselves**, but the seed (the first node, the one whose invite token was used) **never registers any worker** — its log shows perpetual `Waiting for peers...` even though it is reading 60k+ UDP datagrams. The model never loads; `/v1/models` stays empty; `mesh=partial` forever.

Reproduced on `v0.66.0-rc3`, `v0.66.0-rc4`, and a build off `main` HEAD. Affects both the bundled `--llama-flavor cuda` and stock binaries.
ENV is:

- Linux x86_64 hosts (16 total in our cluster — bug is reproducible with 2).
- mesh-llm running inside `docker run --network host --gpus all`, with the bundled binary bind-mounted from a shared filesystem.
- Container image based on `nvidia/cuda:12.6.x-devel-ubuntu24.04`.
- Each host has:
  - A primary management network interface (`10.x.x.x/24`) — routable host-to-host on UDP (verified independently with `nc -uvz` and `nstat`).
  - Multiple `docker*` / CNI bridge interfaces in the `172.16.0.0/12` and `172.17.0.0/16` ranges, **each with identical addresses across hosts** (every host's `docker0` is `172.17.0.1`, every host's `docker_gwbridge` is `172.23.0.1`, etc.).
- Single `mesh-llm serve --split --gguf <large-model.gguf> --ctx-size <N> --mesh-name foo --headless --llama-flavor cuda --listen-all` on the first host.
- `mesh-llm serve --split --join <token> --gguf ... --mesh-name foo --listen-all` on each of the others.

Decoding the invite token printed by the seed reveals iroh's auto-discovered direct-address set leaking *all* kernel-visible interfaces:
json
{
  "id": "17d4bba6...",
  "addrs": [
    {"Ip":"10.x.x.x:47916"},
    {"Ip":"10.x.x.x:47916"},
    {"Ip":"172.23.0.1:47916"},
    {"Ip":"172.23.1.1:47916"},
    {"Ip":"172.23.3.1:47916"},
    {"Ip":"172.24.1.1:47916"}
  ]
}
When a worker dials this token, iroh's connection establishment races all 6 addresses. The `172.23.x` / `172.24.x` entries resolve to the **worker's own docker bridge** (because every host's docker bridge sits at the same address), not the     seed's. Most dial attempts complete the QUIC handshake against the dialing host's own loopback bridge — meaning the worker mistakenly thinks it has reached the seed when it has actually only reached itself, and never establishes a real seed    connection.

IvGolovach:
## Summary
- Add `--bind-ip <IP>` so multi-interface Linux/Docker hosts can bind mesh QUIC to the routable host-to-host address.
- Filter advertised direct addresses in invite tokens and gossip when a bind IP is selected, keeping the selected IP, relay candidates, and public IPv4 candidates while dropping overlapping Docker/CNI/CGNAT/local ranges.
- Document the multi-interface Docker/Linux workflow and the expected verification path.

## Why
#542 reports split meshes failing on host-network Docker/CNI machines where hosts advertise identical bridge addresses. iroh can race those direct candidates and end up dialing the wrong local bridge instead of the intended seed. `--listen-all` only affects the HTTP API/console listeners; it does not select the mesh QUIC interface.

This PR gives operators an explicit, reviewable way to select the mesh-facing IP used by the QUIC endpoint and by the advertised direct-address set.

## Details
- Introduces a typed QUIC bind selection that carries the requested IP and optional port through runtime startup.
- Preserves existing `--bind-port` behavior when no bind IP is supplied.
- Uses `IP:0` when `--bind-ip` is supplied without `--bind-port`, so the OS can still choose the UDP port.
- Reuses the same filtered advertisement path for invite-token creation and gossip self-announcements.
- Keeps the invite-token envelope and mesh wire protocol unchanged.

## Compatibility
- No QUIC ALPN, protobuf, gossip schema, or invite-token format change.
- Existing `--bind-port` behavior remains `0.0.0.0:<port>`.
- New seed + old worker remains compatible because the token is still a normal iroh endpoint address with a narrower, better candidate set.
- Old seed + new worker remains unchanged; the seed must opt into `--bind-ip` for this specific network-selection fix to apply.

## Validation
- `git diff --check`
- `git diff --cached --check`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime bind_ip --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime quic_bind_addr --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime cli_accepts_global_mesh_discovery_mode_before_serve --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime mesh::tests::artifact_transfer_stream_serves_authorized_stage_artifact --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo check -p mesh-llm-host-runtime`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm --lib`

## Notes
- Not run locally: the two-host Docker/CUDA repro from #542. That still needs the external multi-host environment that motivated the issue.

Fixes #542.
Related: #525.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/mod.rs
@@ -18,6 +18,7 @@ use iroh::{Endpoint, EndpointAddr, EndpointId, SecretKey, TransportAddr};
 use prost::Message;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, VecDeque};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use std::sync::Arc;
 
 use tokio::sync::{watch, Mutex};
@@ -32,6 +33,7 @@ use crate::protocol::*;
 use skippy_protocol::proto::stage as skippy_stage_proto;
 
 const PRETTY_LOCAL_REQUEST_WINDOW_SECS: u64 = 24 * 60 * 60;
+const EPHEMERAL_QUIC_PORT: u16 = 0;
 
 fn emit_mesh_info(message: String) {
     let _ = crate::cli::output::emit_event(crate::cli::output::OutputEvent::Info {
@@ -67,14 +69,30 @@ const ARTIFACT_TRANSFER_OPEN_TIMEOUT: std::time::Duration = std::time::Duration:
 const ARTIFACT_TRANSFER_READ_IDLE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
 const ARTIFACT_TRANSFER_BUFFER_BYTES: usize = 1024 * 1024;
 
-fn quic_bind_addr(bind_port: Option<u16>) -> Option<std::net::SocketAddr> {
-    if let Some(port) = bind_port {
-        return Some(std::net::SocketAddr::from(([0, 0, 0, 0], port)));
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub struct QuicBindSelection {
+    pub ip: Option<IpAddr>,
+    pub port: Option<u16>,
+}
+
+fn quic_bind_addr(bind: QuicBindSelection) -> Option<SocketAddr> {
+    if let Some(ip) = bind.ip {
+        return Some(SocketAddr::new(
+            ip,
+            bind.port.unwrap_or(EPHEMERAL_QUIC_PORT),
+        ));
+    }
+
+    if let Some(port) = bind.port {
+        return Some(SocketAddr::from(([0, 0, 0, 0], port)));
     }
 
     #[cfg(target_os = "windows")]
     {
-        Some(std::net::SocketAddr::from(([127, 0, 0, 1], 0)))
+        Some(std::net::SocketAddr::from((
+            [127, 0, 0, 1],
+            EPHEMERAL_QUIC_PORT,
+        )))
     }
 
     #[cfg(not(target_os = "windows"))]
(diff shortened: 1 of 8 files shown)


crates/mesh-llm-host-runtime/src/mesh/mod.rs
```diff
@@ -67,9 +68,19 @@ const ARTIFACT_TRANSFER_OPEN_TIMEOUT: std::time::Duration = std::time::Duration:
...
-        return Some(std::net::SocketAddr::from(([0, 0, 0, 0], port)));
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub struct QuicBindSelection {
+    pub ip: Option<IpAddr>,
+    pub port: Option<u16>,
+}
+
+fn quic_bind_addr(bind: QuicBindSelection) -> Option<SocketAddr> {
+    if let Some(ip) = bind.ip {
+        return Some(SocketAddr::new(ip, bind.port.unwrap_or(0)));
```
ndizazzo:
```rust
        return Some(SocketAddr::new(ip, bind.port.unwrap_or(0)));
```

Should this be a default port constant instead of 0? Or is that functionally required to allow this to operate and not start conflict-free?

ndizazzo (approved):
Small minor comment, but LGTM!

crates/mesh-llm-host-runtime/src/mesh/mod.rs
IvGolovach: Thanks, good call. I kept port `0` intentionally here because `--bind-ip` without `--bind-port` should still let the OS choose a conflict-free ephemeral QUIC port.

I pushed a small follow-up that names this as `EPHEMERAL_QUIC_PORT` and adds a short comment at the STUN path so the intent is explicit rather than looking like a magic default.

ndizazzo: > should still let the OS choose a conflict-free ephemeral QUIC port

Ok got it - so 0 is for the OS to pick something. Makes sense!

Issue #544, opened 2026-05-13
Skippy: return reasoning_content for thinking models


michaelneale:
Thinking models (Qwen3.5, DeepSeek-R1, etc.) generate `<think>...</think>` blocks. llama-server parses these and returns the thinking in a separate `reasoning_content` field on the assistant message, with only the answer in `content`. Skippy currently puts everything in `content` with no separation.

## What needs to happen

1. **Persist `reasoning_format` in chat metadata** — `skippy_common_chat_metadata_json()` in `stage-chat.cpp` needs to include `reasoning_format` from `common_chat_params`. The parse side (`skippy_parse_chat_response_json`) needs to read it back into `common_chat_parser_params`.

2. **Add `reasoning_content` to `AssistantMessage`** — in `crates/openai-frontend/src/chat.rs`, add an optional `reasoning_content: Option<String>` field (skip-serialize-if-none). Update all constructors.

3. **Call `parse_chat_response_json` for all chat responses** — currently only called for tool-call extraction. The parsed JSON already contains `reasoning_content` when the parser is configured correctly. Wire this into `chat_response_from_generated_text()` in `crates/skippy-server/src/frontend.rs`.

4. **Streaming** — the streaming path (`ChatCompletionChunk`) also needs a `reasoning_content` delta, matching the OpenAI streaming format.

## Context

- llama-server uses `common_chat_parse()` with `reasoning_format = COMMON_REASONING_FORMAT_DEEPSEEK` (set by `common_chat_templates_apply`). The PEG parser extracts thinking blocks and puts them in `common_chat_msg::reasoning_content`.
- Skippy already calls the same `common_chat_parse` via `skippy_parse_chat_response_json`, but `reasoning_format` defaults to `NONE` because it is not round-tripped through the metadata JSON.
- The generation prompt for thinking models ends with `<think>\n`. The model output starts AFTER that prefix — the generated text does NOT include `<think>` but DOES include `</think>`.

## Files involved

- `.deps/llama.cpp/common/stage-chat.cpp` — metadata serialization + parse params
- `crates/openai-frontend/src/chat.rs` — `AssistantMessage`, `ChatCompletionChunk`
- `crates/skippy-server/src/frontend.rs` — response assembly
- `crates/skippy-server/src/frontend/backend.rs` — parse call site
- `crates/skippy-server/src/frontend/prompting.rs` — `parse_chat_response_json` wrapper

ndizazzo:
[...]

PR #552, opened 2026-05-15
Return first-class reasoning_content for Skippy thinking models

IvGolovach:
## Summary

This makes Skippy return thinking-model reasoning as first-class OpenAI-compatible reasoning data instead of mixing it into the visible assistant answer.

- Adds optional `reasoning_content` fields to chat completion messages and streaming deltas.
- Persists llama.cpp `reasoning_format` through Skippy chat metadata so `common_chat_parse` can extract thinking traces correctly.
- Parses Skippy chat output for every chat response, not only the tool-call path.
- Streams reasoning deltas separately from visible answer deltas and translates them to `response.reasoning_text.delta` for the Responses API bridge.
- Updates the console adapter to consume first-class reasoning events while preserving the existing `<think>`-compatible transcript fallback.

## Why

Thinking models such as Qwen and DeepSeek-style models can emit reasoning traces separately from the final answer. llama.cpp already supports this via `reasoning_content`, but Skippy was losing the parser mode because `reasoning_format` was not round-tripped through the stage chat metadata. As a result, clients saw reasoning mixed into `content` instead of receiving a clean final answer plus a separate reasoning field.

## Implementation

- Round-trip `reasoning_format` in the durable llama.cpp stage-chat ABI patch and restore it into `common_chat_parser_params` before parsing generated output.
- Add `reasoning_content` to the shared OpenAI chat message and stream delta types, omitted when absent for existing clients.
- Replace the tool-call-only parse path with a parsed chat-message path that can carry visible content, reasoning content, and tool calls together.
- Add a Skippy streaming parser that tracks the cumulative generated text, emits only new reasoning/content suffixes, and still emits tool-call chunks at final parse time.
- Add Responses stream translation for upstream `delta.reasoning_content` as `response.reasoning_text.delta`.
- Map Responses reasoning SSE events to AG-UI reasoning events in the console, while saving thinking parts back into the existing `<think>...</think>` transcript format for compatibility with current rendering and storage.

## Compatibility

- API changes are additive: `reasoning_content` is skipped when absent.
- Non-thinking models continue to return normal `content` only.
- Existing `<think>` parsing/rendering remains supported in the console.
- No mesh gossip, peer protocol, or plugin protocol compatibility change is introduced.

## Validation

- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p openai-frontend`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p skippy-server`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime test_api_proxy_translates_streaming_reasoning_content_events`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime test_api_proxy_translates_streaming_responses_events_incrementally`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo check -p mesh-llm-host-runtime`
- `pnpm --dir ./crates/mesh-llm-ui exec vitest run src/features/chat/api/mesh-connection.test.ts src/features/chat/api/use-chat-messages.test.ts`
- `pnpm --dir ./crates/mesh-llm-ui exec vitest run src/features/chat/pages/ChatPage.test.tsx --testNamePattern thinking`
- `just build`

## Related

Related: #544

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
@@ -1267,6 +1267,28 @@ async fn relay_translated_responses_stream<R: AsyncRead + Unpin>(
                 .choices
                 .first()
                 .and_then(|choice| choice.delta.as_ref())
+                .and_then(|delta| delta.reasoning_content.as_deref())
+            {
+                let sequence_number = next_sequence_number();
+                let event = serde_json::to_string(
+                    &response_adapter::responses_stream_reasoning_delta_event_with_sequence(
+                        &item_id,
+                        delta,
+                        sequence_number,
+                    ),
+                )
+                .context("serialize response.reasoning_text.delta event")?;
+                response_adapter::write_chunked_sse_event(
+                    tcp_stream,
+                    Some("response.reasoning_text.delta"),
+                    &event,
+                )
+                .await?;
+            }
+            if let Some(delta) = chunk
+                .choices
+                .first()
+                .and_then(|choice| choice.delta.as_ref())
                 .and_then(|delta| delta.content.as_deref())
             {
                 if !output_item_emitted {
--- a/crates/mesh-llm-host-runtime/src/network/openai/response_adapter.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/response_adapter.rs
@@ -6,6 +6,7 @@ pub(crate) use openai_frontend::{
     responses_stream_content_part_done_event, responses_stream_created_event_with_sequence,
     responses_stream_delta_event_with_logprobs_and_sequence,
     responses_stream_output_item_added_event, responses_stream_output_item_done_event,
+    responses_stream_reasoning_delta_event_with_sequence,
     responses_stream_text_done_event_with_sequence, stream_usage_to_responses_usage,
     translate_chat_completion_to_responses,
 };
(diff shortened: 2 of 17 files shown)


crates/mesh-llm-host-runtime/src/network/openai/transport.rs
```diff
@@ -1263,6 +1263,28 @@ async fn relay_translated_responses_stream<R: AsyncRead + Unpin>(
                 .await?;
                 created_emitted = true;
             }
+            if let Some(delta) = chunk
```
ndizazzo, line 1266: I think reasoning-first streams can emit events out of order here...

If the first chunk has only `reasoning_content`, we send `response.reasoning_text.delta` before `response.output_item.added`, so clients may see a delta for an unknown item. 

We should emit the output item scaffold before the first reasoning delta so clients can attach the reasoning text to a specific output item.

ndizazzo (approved):
Approving so you can push a fix then merge.

ndizazzo:
@IvGolovach merging - can you follows this in a new PR for [this comment](https://github.com/Mesh-LLM/mesh-llm/pull/552#discussion_r3251761361)?

PR #553, opened 2026-05-15
Advertise multimodal capabilities from runtime evidence

IvGolovach:
## Summary

This makes multimodal capability advertisement come from the runtime that actually loaded the model, rather than from catalog/name heuristics alone.

- Advertises vision/audio support from runtime-verified media capability evidence once a model is warm.
- Keeps served model descriptors authoritative for `/v1/models`, `/api/models`, cached auto-routing, and OpenAI ingress routing.
- Requires runtime evidence before routing image/audio requests to a model, avoiding false-positive multimodal matches from model names or catalog hints.
- Preserves the existing fallback capability inference for cold/catalog-only models where no runtime evidence exists yet.

## Why

The multimodal design depends on routing decisions reflecting what the active backend can actually serve. Name-based or catalog-based capability inference is useful before startup, but it can over-advertise a model as vision/audio-capable when the runtime has not loaded the required projector/media path.

This PR makes the warm runtime descriptor the source of truth: once the embedded runtime knows whether media support is actually available, that verified descriptor is what clients, routing, and runtime snapshots consume.

## Implementation

- Adds runtime media capability evidence to the model capability layer.
- Captures verified media support from the embedded runtime startup path and publishes it through the served model descriptor before model-ready advertisement.
- Teaches `/v1/models`, cached auto-routing, and OpenAI ingress routing to prefer warm served descriptors over colder catalog/name-derived metadata.
- Updates media routing checks so image/audio requests require runtime-verified support when choosing an active model.
- Surfaces verified capability details in runtime data snapshots so the management/UI model view can distinguish likely catalog capability from runtime evidence.

## Protocol / Compatibility

- No new protobuf fields, ALPN changes, stream types, or plugin protocol changes.
- Existing capability fields continue to be used; this only changes how this node derives them for warm served models.
- Cold/catalog-only descriptors still retain the previous inferred capability behavior until runtime evidence is available.

## Validation

- `git diff --check origin/main...HEAD`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime --lib runtime_media_verification`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime --lib runtime_verified_served_model_descriptor`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime --lib test_model_satisfies_media_requirements_matches_required_modalities`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime --lib models_list_`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime --lib cached_auto_model`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime --lib runtime_data_model_snapshot_keeps_warm_descriptor_media_capabilities_authoritative`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime --lib`
- `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm --lib`

## Related

Related to the multimodal capability and routing model in `docs/design/MULTI_MODAL.md`.

## Known residual risk

I did not run a mixed-version two-node runtime test with a released binary and a real multimodal projector in this local pass. The change intentionally avoids wire/protocol changes, but the final end-to-end proof for real multimodal hardware should still come from a multi-node runtime check.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/mod.rs
@@ -478,6 +478,7 @@ fn infer_remote_served_descriptors(
             };
             ServedModelDescriptor {
                 identity,
+                capabilities_known: false,
                 capabilities: crate::models::ModelCapabilities::default(),
                 topology: None,
             }
@@ -669,6 +670,7 @@ fn descriptor_from_identity(
     capabilities.moe = false;
     ServedModelDescriptor {
         identity,
+        capabilities_known: true,
         capabilities,
         topology,
     }
@@ -3020,7 +3022,14 @@ impl Node {
 
     async fn refresh_served_model_descriptors(&self) {
         let serving_models = self.serving_models.lock().await.clone();
-        let descriptors = if let Some(primary_model_name) = serving_models.first() {
+        let existing_by_name: HashMap<_, _> = self
+            .served_model_descriptors
+            .lock()
+            .await
+            .iter()
+            .map(|descriptor| (descriptor.identity.model_name.clone(), descriptor.clone()))
+            .collect();
+        let mut descriptors = if let Some(primary_model_name) = serving_models.first() {
             let model_source = self.model_source.lock().await.clone();
             let primary_model_path = crate::models::find_model_path(primary_model_name);
             infer_served_model_descriptors(
@@ -3032,6 +3041,15 @@ impl Node {
         } else {
             Vec::new()
         };
+        for descriptor in &mut descriptors {
+            if let Some(existing) = existing_by_name.get(&descriptor.identity.model_name) {
+                descriptor.capabilities = existing.capabilities;
+                descriptor.capabilities_known = existing.capabilities_known;
+                if existing.topology.is_some() {
+                    descriptor.topology = existing.topology.clone();
+                }
+            }
+        }
         self.set_served_model_descriptors(descriptors).await;
     }
 
--- a/crates/mesh-llm-host-runtime/src/mesh/tests.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/tests.rs
@@ -253,6 +253,62 @@ async fn make_test_node(role: super::NodeRole) -> Result<Node> {
 }
 
 #[tokio::test]
+async fn set_serving_models_preserves_existing_known_descriptor_capabilities_when_adding_model(
+) -> Result<()> {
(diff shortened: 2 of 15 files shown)


ndizazzo (changes requested):
I like this idea! Agent picked up a couple things that I think will conflict with work in #519:

```
Medium — Starting another runtime can erase existing verified multimodal 
descriptors. register_runtime_instance() only reapplies verified capabilities for the 
newly loaded model, but add_serving_assignment() calls set_serving_models(), 
which refreshes/replaces all served descriptors. If a verified vision model is already 
serving and a second model starts later, the first descriptor can be rebuilt from 
static/default inference and lose its verified vision capability.


Medium — Empty/legacy descriptors are treated as authoritative text-only capabilities. 
capabilities_for_model() uses descriptor capabilities whenever a descriptor exists, 
and the runtime-data snapshot skips heuristics when descriptor.is_some(). In mixed-
version meshes or pre-verification states where descriptors carry default capabilities, 
likely vision/audio models become text-only for /v1/models and auto-routing despite 
existing heuristic signals.
```

And some media related issues:

```
High — Media auto-routing falls back to text models after a hard media miss. In 
ingress.rs and transport.rs, when no model satisfies 
model_satisfies_media_requirements, the code restores the full candidate list and 
calls pick_model_classified. For an image/audio request on a mesh with only text 
models, this forwards media to a text-only runtime instead of returning no capable 
model. The fallback should only apply to soft ranking filters, not hard media 
requirements.

High — Runtime verification drops audio support for every model. 
models/capabilities.rs unconditionally sets caps.audio = CapabilityLevel::None, while 
routing now requires supports_audio_runtime() for audio requests. Any runtime that 
can actually handle audio will be advertised as non-audio, so auto audio requests 
cannot select it.
```

I don't agree that they're "high" priority, but I left it's comments intact

ndizazzo (approved):

PR #556, opened 2026-05-15
E2E inference encryption and attestation foundations

michaelneale:
Every node now generates an X25519 inference keypair at startup, runs runtime hardening checks, and advertises both in gossip. Peers can verify each other's security posture and prefer or require hardened hosts for inference routing.

## What you can do

- Run `mesh-llm --require-hardened` to only route inference to hosts with a verified security posture (debugger blocked, core dumps disabled, SIP enabled, RDMA disabled)
- Without the flag, hardened hosts are still preferred as a soft routing preference
- Check `/api/status` to see per-peer `e2e_encryption` and `hardened` booleans

## What's included

- **Runtime hardening** — `PT_DENY_ATTACH`, core dump disable, dangerous env var scrubbing (`DYLD_INSERT_LIBRARIES`, `LD_PRELOAD`), SIP/RDMA verification, binary SHA-256 hash
- **Inference keypair** — X25519 keypair generated at startup, public key advertised in gossip
- **Attestation library** — P-256 ECDSA signing via Apple Secure Enclave (macOS), cross-platform verification, challenge-response nonce signing (not yet wired to gossip)
- **Encryption library** — NaCl box with per-request ephemeral keys for forward secrecy, chunked streaming encryption (building block for future use)
- **Gossip wiring** — inference public key and security posture advertised, propagated via transitive gossip, stored per peer
- **Routing enforcement** — `--require-hardened` drops non-hardened peers from routing candidates; soft preference sorts hardened peers first
- **API exposure** — `e2e_encryption` and `hardened` fields on peer payloads in `/api/status`

## Protocol

Two new optional fields added to `PeerAnnouncement` proto:
- Field 44: `inference_public_key` (string, base64 X25519 public key)
- Field 45: `security_posture` (message: sip_enabled, rdma_disabled, debugger_blocked, core_dumps_disabled, binary_hash)

New `SecurityPosture` proto message. Additive — older nodes ignore unknown fields. No breaking change.

## Validation

- 14 crypto tests (7 encryption + 5 attestation + 2 hardening)
- 2 routing preference tests (require-hardened filter + soft preference)
- 1355 host-runtime tests pass
- 132 system tests pass


ndizazzo (changes requested):
Ran this though Opus 4.7 out of curiosity... agent caught the following:

## High 
`--require-hardened` security posture is entirely self-reported

Any peer can set `sip_enabled: true, rdma_disabled: true, debugger_blocked: true, core_dumps_disabled: true, binary_hash: Some(any_string)` and pass `is_hardened()` without doing any of those things.
`gossip.rs:115-119`, `mod.rs:3893-3902`:
```rust
let hardened = p.security_posture.as_ref().is_some_and(|sp| sp.is_hardened());
```
`apply_transitive_ann` also merges `security_posture` from transitively-received announcements, so a malicious node Y can forge the posture of an unrelated peer X.
The CLI help reads: *"Only route inference to peers that report a hardened security posture (debugger blocked, core dumps disabled, SIP enabled, RDMA disabled)."* Users reading that will assume meaningful protection. Options:
- Reword the help text to make clear this is self-reported with no cryptographic backing
- Drop `security_posture` from `apply_transitive_ann`
- Gate the flag on attestation being wired
---
## Medium

Hash rotation breaks hardened-first routing

`hosts_for_model` (`mod.rs:3921-3934`) sorts hardened hosts to the front, then immediately destroys the boundary with `rotate_left(hash % len)`. For 2 hosts, the hardened host is first only ~50% of the time.
```rust
hosts.sort_by(|(id_a, h_a), (id_b, h_b)| h_b.cmp(h_a).then(id_a.cmp(id_b)));
let mut result: Vec<EndpointId> = hosts.into_iter().map(|(id, _)| id).collect();
// ...
let idx = (hash as usize) % result.len();
result.rotate_left(idx);
```
With 3 hosts `[H_hardened, U1, U2]`:
- `idx == 0` → `[H_hardened, U1, U2]` ✓
- `idx == 1` → `[U1, U2, H_hardened]` ✗
- `idx == 2` → `[U2, H_hardened, U1]` ✗
Fix: rotate within the hardened subset only, or partition first and apply the modulo against `hardened_count`. Only affects the soft-preference path — `--require-hardened` (which filters before sorting) is unaffected.
---
`e2e_encryption: true` is misleading

`collector.rs:1024`:
```rust
e2e_encryption: peer.inference_public_key.as_ref().map(|_| true),
```
This fires whenever a peer advertises any string in `inference_public_key`, including garbage. All encrypt/decrypt functions are `#[allow(dead_code)]` with no non-test callers. A fresh build will show `e2e_encryption: true` in `/api/status` while all inference traffic remains cleartext.
Options: rename to `e2e_encryption_supported`, derive the value from a successful `parse_public_key` call, or hold this field back until the request path actually uses the keys.

## Low

Test doesn't assert ordering

`soft_preference_puts_hardened_first` (`tests.rs:5790-5830`) only checks membership and count, not that `hosts[0] == hardened_id`. That's exactly the contract the test name promises to verify, and the gap that hides the hash rotation bug above.

---

RDMA check hardcoded `true` on Linux
`hardening.rs:137-156` — `check_rdma_disabled` returns `true` unconditionally on non-macOS regardless of actual RDMA state. Currently masked because `sip_enabled` is always false on Linux so `is_hardened()` never fires. Return `false` (conservative direction) to avoid wrong-direction defaults for future Linux hardening work.

---

`write_encrypted_chunk` silently truncates length to `u32`
`inference_encryption.rs:251`:
```rust
let payload_len = (24 + ciphertext.len()) as u32;
```
Silently truncates on overflow. The reader enforces a 16 MB cap but the writer doesn't. Use a checked conversion that returns `EncryptionFailed` to match reader semantics.

michaelneale:
Yeah that doesn't surprise me. It's a 3rd take on this. I am still trying to (human) think through what we can really guarantee with this. Mostly things are running as untamed as we can assure they are (the attestations). quic already is end to end encrypted so I think that covers most of the transit privacy (for TEE devices - I am not sure at what level to implement that, is is skippy or is it more other engines). Any thoughts @ndizazzo ?

michaelneale:
@ndizazzo some feedback addressed, I still consider this a WIP though

michaelneale:
## Follow-up TODOs — what's needed before `--require-hardened` is a defensible user-facing claim

Posted from a review pass focused on **"what can this actually prove to a consumer today vs. what does it need to prove."** The PR is a solid foundation; the items below are the wiring gaps between "primitives exist" and "we can honestly tell users their host is verifiable."

Happy to take any of these as follow-up PRs — flagging here so they don't get lost.

### Priority 1 — Blessed-binaries registry (blocks the whole story)

Without this, `binary_hash` in `SecurityPosture` is decorative: it proves the binary matches *some* binary, not the *project's* binary. `AttestationStatus::UnblessedBinary` is already enumerated as a TODO marker — this fills it in.

- [ ] Decide trust-root model (worth a short design doc before code):
  - Sigstore-style transparency log
  - Project-signed Ed25519 manifest, pinned/fetched per release
  - Nostr-published signed manifest (leverages existing infra)
- [ ] Define manifest schema: `{ version, accepted_hashes[], signing_pubkey, expiry }`
- [ ] Client-side verification path: fetch + verify signature + check `binary_hash` membership
- [ ] Surface `UnblessedBinary` in `/api/status` and in routing decisions
- [ ] Decide revocation semantics (compromised release → how do peers learn?)

### Priority 2 — Attestation verification on direct gossip receive

The PR correctly drops `inference_public_key` / `security_posture` from `apply_transitive_ann` (great, prevents bridge-forge). But direct `apply_announcement` still accepts these fields with no signature check, so posture booleans are still effectively self-reported even though the SE key is generated.

- [ ] Add a `signed_attestation: Option<SignedHardwareAttestation>` field to `PeerAnnouncement` (optional → backward compatible like the existing posture fields)
- [ ] On direct receive, verify:
  - SE signature over the canonical signing bytes
  - `attestation.inference_public_key == announcement.inference_public_key`
  - `attestation.node_endpoint_id == sender endpoint id`
  - `attestation.timestamp` is within a freshness window (replay defense)
- [ ] Only set `PeerInfo.security_posture` / `inference_public_key` if verification passes
- [ ] Bubble `AttestationStatus` to `/api/status` so UI can show "attested" vs. "self-reported"
- [ ] Update `--require-hardened` filter in `hosts_for_model` to require `AttestationStatus::Verified`

### Priority 3 — Liveness via challenge-response at inference setup

Proves the SE is *currently present and willing to sign*, not just that an attestation blob was once produced and replayed. `challenge_signing_bytes` already exists in `attestation.rs` — this is wiring.

- [ ] Define a pre-inference handshake message: client sends 32-byte nonce
- [ ] Host signs `CHALLENGE_DOMAIN || se_public_key || nonce` with the SE key
- [ ] Client verifies signature against the SE pubkey from the (verified) attestation
- [ ] Fail closed if signature missing/invalid under `--require-hardened`

### Priority 4 — Bind X25519 session to the SE-attested key

Without this, an MITM with code execution on the host could substitute their own X25519 key even if the SE attestation is replayed. The X25519 ephemeral handshake needs to be bound to the SE identity.

- [ ] Include `se_public_key` + a fresh SE-signed nonce in the session establishment envelope
- [ ] Verify on the client before deriving the SalsaBox
- [ ] Consider rolling this into the Priority 3 challenge to avoid two round-trips

### Priority 5 — Actually wire `crypto/inference_encryption` into the proxy path

All of `encrypt_inference_request` / `decrypt_inference_request` / chunk helpers are currently `#[allow(dead_code)]`. Worth being explicit about the design before coding it.

- [ ] Document where in the proxy the encrypt/decrypt boundary sits (entry API → host deserialization)
- [ ] Decide on cleartext metadata: model name is currently cleartext in `EncryptedInferencePayload` for routing — confirm that's acceptable, document why
- [ ] Document the `ENCRYPTED_TUNNEL_MAGIC = 0xE1` framing alongside existing tunnel framing so it's clear there's no collision
- [ ] Note in design doc: **this is redundant with QUIC for transit privacy**. Its real value is keeping cleartext inside the verified binary's process — i.e., it matters at the host's deserialization boundary, not on the wire.

### Smaller items (nice-to-have, not blocking)

- [ ] **Truth-in-advertising on `--require-hardened`**: until Priority 2 lands, either rename (`--require-self-reported-hardened`) or gate behind `--unstable-*`. Currently an attacker just sets four bools to pass the filter.
- [ ] **`binary_hash` participation in `is_hardened()`**: drop it from `is_hardened()` until Priority 1 lands, or document the field as informational only. Right now it just rejects `SecurityPosture::default()`, which any peer trivially defeats.
- [ ] **`InferenceKeypair` key rotation**: same X25519 secret protects every inference for the process lifetime. Per-request forward secrecy via the ephemeral session is fine, but a rotation hook (timer-based, re-announce on rotate) is worth having before "compliance" is anyone's word for this.
- [ ] **`zeroize` the inference secret key** in memory. Threat model is "operator can inspect the process" — the host's own private key is the most valuable thing to grab and it's currently sitting in plain memory. `secret_bytes()` returns a copy that also isn't zeroized. Long-term, this wants to live in the SE.
- [ ] **`InferenceKeypair::from_secret_bytes`** is currently unused. Decide if the key persists across restarts (then this matters + persisted-on-disk concerns kick in) or it doesn't (then drop the helper).
- [ ] **`csrutil` / `rdma_ctl` parsing fragility**: `stdout.contains("enabled")` has broken across macOS versions before. Worth a comment + maybe a more structured parse. Also: `rdma_ctl` not found returns `true` (RDMA "disabled") — conflates "absent" with "disabled," probably fine but worth a code comment.
- [ ] **`local_security_posture: Arc<Mutex<Option<...>>>`** — set once, read every gossip tick. `OnceLock` or `ArcSwap` expresses the intent better and avoids `.lock().await` on a hot path. Minor.
- [ ] **Comment in the `.proto`** mirroring the `gossip.rs:114-118` warning, so future contributors don't re-add `inference_public_key` / `security_posture` to transitive announcements.

### User-facing messaging note (for whoever writes the launch copy)

When this ships to end-users, please don't phrase it as **"your prompts are private from the host."** That claim is false on commodity hardware without a TEE, and a single counter-example demolishes trust in the whole feature.

The honest claim, once Priorities 1–4 land, is something like:

> *"The peer is provably running mesh-llm binary vX.Y on Apple Silicon with debugger access blocked and System Integrity Protection enabled, signed by a non-extractable Secure Enclave key. Observing your inference would require tampering this attestation says didn't happen."*

That's a real, defensible step up from "trust me bro" — but it's *not* "the host literally cannot see your prompts." Reaching that strong claim needs a TEE (H100 CC, SEV-SNP/TDX, or Apple PCC-class silicon), none of which apply to a consumer Mac mesh.

---

cc @micn @ndizazzo — happy to take any of these as follow-up PRs. Suggest Priority 1 (design doc first) and Priority 2 (verification on direct receive) as the natural next two pieces of work.

PR #558, opened 2026-05-15
Add mixed-version owner-control QA harness

IvGolovach:
## Summary

Adds an executable QA harness for proving mixed-version mesh compatibility while keeping operator config and inventory traffic on the dedicated owner-control lane.

- Adds `scripts/qa-control-plane-mixed-version.sh` for released/current binary checks across public-mesh probes, private loopback coexistence, and owner-control bootstrap/get-config validation.
- Captures reviewable evidence as `manifest.json`, `commands.jsonl`, `results.jsonl`, `summary.md`, `summary.json`, `versions/*.txt`, and grouped `logs/`, `status/`, `models/`, `chat/`, and `control/` payloads.
- Adds `--print-plan` so CI/review automation can validate the planned checks without starting mesh processes or writing evidence.
- Adds Rust contract tests covering harness presence/executable mode, help text, argument validation, and side-effect-free print-plan behavior.
- Updates `docs/design/TESTING.md` and `docs/USAGE.md` with the new evidence contract and usage examples.

## Why

After the owner-control split in #539, config and inventory mutation are intentionally exclusive to `mesh-llm-control/1`; public mesh join, gossip, routing, and inference still need to remain mixed-version compatible. The bind-IP/network work in #547 also made it clear that distributed regressions need repeatable evidence, not just local unit coverage.

This harness gives maintainers a single command that can collect that evidence from a released binary and a current branch binary, while keeping legacy mesh-plane config streams reserved but inactive.

## Compatibility

- No mesh wire protocol or protobuf schema changes.
- No runtime behavior changes outside the new QA script and docs.
- The harness does not publish meshes by default and does not modify protocol state.
- Temporary runtime homes and owner keystores are isolated under a short-lived `/tmp/mllmqa-*` work root; evidence output does not include the temporary owner keystore.
- Status payload checks fail if owner-control endpoint data leaks into normal mesh status evidence.

## Validation

- `git diff --check`
- `git diff --cached --check`
- `bash -n scripts/qa-control-plane-mixed-version.sh`
- `scripts/qa-control-plane-mixed-version.sh --released-binary <true> --current-binary <true> --evidence-dir /tmp/mesh-harness-print-plan --local-only --config-only --print-plan`
- `cargo test -p mesh-llm --test qa_control_plane_harness` — 5 passed
- `cargo test -p mesh-llm --test protocol_compat_v0_client` — 15 passed
- `cargo test -p mesh-llm --test protocol_convert_matrix owner_control_legacy_json_maps_to_structured_error` — 1 passed
- `cargo test -p mesh-llm-client --test control_plane_fallback` — 4 passed
- `cargo test -p mesh-llm-client --test protocol_wire owner_control_legacy_json_rejects_with_structured_error` — 1 passed
- `cargo test -p mesh-llm-host-runtime --lib control_plane_legacy_compat` — 2 passed
- `cargo check -p mesh-llm`
- `scripts/qa-control-plane-mixed-version.sh --released-binary target/debug/mesh-llm --current-binary target/debug/mesh-llm --evidence-dir /tmp/mesh-harness-smoke --local-only --config-only --skip-cargo-tests --max-wait 60 --stable-probes 2 --base-port 24140` — PASS, 18 PASS / 1 PREREQ for intentionally skipped embedded cargo tests

`just build` was not run locally because this is verification tooling plus docs/tests, with no UI or release artifact changes.

## Related

Related: #539, #547

### DIFF
--- a/scripts/qa-control-plane-mixed-version.sh
+++ b/scripts/qa-control-plane-mixed-version.sh
@@ -1,0 +1,1070 @@
+#!/usr/bin/env bash
+# qa-control-plane-mixed-version.sh - certify mixed-version public mesh and
+# owner-control behavior across a released and current mesh-llm binary.
+
+set -euo pipefail
+
+RELEASED_BINARY=""
+CURRENT_BINARY=""
+EVIDENCE_DIR=".sisyphus/evidence"
+LOCAL_ONLY=false
+CONFIG_ONLY=false
+MODEL=""
+RELEASED_MODEL=""
+CURRENT_MODEL=""
+BASE_PORT="${MESH_QA_BASE_PORT:-19640}"
+MAX_WAIT="${MESH_QA_MAX_WAIT:-180}"
+STABLE_PROBES="${MESH_QA_STABLE_PROBES:-3}"
+CHAT_MAX_TIME="${MESH_QA_CHAT_MAX_TIME:-120}"
+CTX_SIZE="${MESH_QA_CTX_SIZE:-512}"
+TMP_ROOT="${MESH_QA_TMP_ROOT:-${RUNNER_TEMP:-${TMPDIR:-/tmp}}}"
+RUN_CARGO_TESTS=true
+KEEP_LOGS=false
+REQUIRE_PUBLIC=false
+PRINT_PLAN=false
+
+RUN_DIR=""
+LOG_DIR=""
+STATUS_DIR=""
+MODELS_DIR=""
+CHAT_DIR=""
+CONTROL_DIR=""
+VERSIONS_DIR=""
+RESULTS_JSONL=""
+COMMANDS_JSONL=""
+MANIFEST_JSON=""
+SUMMARY_JSON=""
+SUMMARY_MD=""
+WORK_ROOT=""
+PIDS=()
+EXIT_STATUS=0
+START_NODE_PID=""
+WAIT_TOKEN_VALUE=""
+WAIT_MODELS_MODEL_ID=""
+NODE_INDEX=0
+
+usage() {
+    cat <<'EOF'
+Usage:
+  scripts/qa-control-plane-mixed-version.sh \
+    --released-binary /path/to/released/mesh-llm \
+    --current-binary /path/to/current/mesh-llm \
+    --evidence-dir .sisyphus/evidence [options]
+
+Purpose:
+  Produce executable evidence that a current mesh-llm build remains compatible
+  with a released mesh-llm binary while owner-control stays on its dedicated
+  mesh-llm-control/1 ALPN.
+
+Required:
+  --released-binary PATH   Released/reference mesh-llm binary.
+  --current-binary PATH    Current-branch mesh-llm binary.
+
+Options:
+  --evidence-dir DIR       Evidence root (default: .sisyphus/evidence).
+  --local-only             Skip public-mesh probes.
+  --config-only            Owner-control lane only; implies --local-only.
+  --model REF_OR_PATH      Model used for both local routing smokes.
+  --released-model REF     Model for released-binary serving direction.
+  --current-model REF      Model for current-binary serving direction.
+  --base-port PORT         First reserved local port (default: 19640).
+  --max-wait SECONDS       Readiness timeout (default: 180).
+  --stable-probes N        Consecutive peer/model probes required (default: 3).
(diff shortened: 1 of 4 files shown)


scripts/qa-control-plane-mixed-version.sh
```diff
@@ -0,0 +1,998 @@
...
+        record_result "PASS" "${label}.no-control-leak" \
+            "status payload does not expose owner-control endpoint data" "path=$path"
+        return 0
+    fi
+    record_result "FAIL" "${label}.no-control-leak" \
+        "status payload exposed owner-control endpoint data" "path=$path"
+    return 1
+}
+
+start_node() {
```
ndizazzo, line 641: Every call to this gets run in a subshell...  PIDS and NODE_INDEX mutations are lost:

- `cleanup()` iterates an empty `PIDS` array → every spawned process is leaked on exit
- Cleanup records emit PASS cleanup processes=0 unconditionally → false PASS in summary.md/summary.json
- `NODE_INDEX` resets to 1 every call → all nodes share the same WORK_ROOT/h-n1 HOME and runtime root instead of isolated dirs

Fix: capture PID in parent shell directly (as ci-two-node-client-serving-smoke.sh does with HOST_PID=$!), and call start_node without command substitution.

scripts/qa-control-plane-mixed-version.sh
```diff
@@ -0,0 +1,998 @@
...
+        export MESH_LLM_RUNTIME_ROOT="$runtime"
+        export MESH_LLM_EPHEMERAL_KEY=1
+        exec "$binary" "$@"
+    ) >"$log" 2>&1 &
+    local pid=$!
+    PIDS+=("$pid")
+    echo "$pid"
+}
+
+assert_process_alive() {
```
ndizazzo, line 664: Every call site ignores the return value for this... When a binary exits during startup, the harness silently continues and reports cascading FAILs rather than failing fast. 

A simple fix could be to just add || return 1 at each call

scripts/qa-control-plane-mixed-version.sh
```diff
@@ -0,0 +1,998 @@
...
+            ;;
+        --skip-cargo-tests)
+            RUN_CARGO_TESTS=false
+            shift
+            ;;
+        --require-public)
+            REQUIRE_PUBLIC=true
+            shift
+            ;;
+        --print-plan)
```
ndizazzo, line 179: The binary executable check runs before the `--print-plan` early-exit, so dry-run validation in CI requires actual binaries. 

Moving the -x check after the print-plan branch would match the documented "no side effects" contract.

scripts/qa-control-plane-mixed-version.sh
```diff
@@ -0,0 +1,998 @@
...
+append_summary "- Released binary: \`$RELEASED_BINARY\`"
+append_summary "- Current binary: \`$CURRENT_BINARY\`"
+append_summary "- Local only: \`$LOCAL_ONLY\`"
+append_summary "- Config only: \`$CONFIG_ONLY\`"
+append_summary "- Evidence: \`$RUN_DIR\`"
+append_summary "- Work root: \`$WORK_ROOT\`"
+append_summary ""
+
+binary_version() {
+    "$1" --version 2>/dev/null | head -1 || true
```
ndizazzo: `|| true` swallows errors from `--version`, leaving versions/released.txt empty. 

An empty version field should be a PREREQ, not a PASS

ndizazzo (changes requested):
Minor comments:

- Plan check names don't match record_result names — automation diffing plan vs. results.jsonl won't find matches.
- Help text asserts PASS for checks that become PREREQ when cargo is missing; entrenched by a contract test that asserts the literal strings in --help.
- Doesn't follow existing CI script conventions (=$! PID capture, /tmp/ logs, env-var config).
- record_result mixes `%s/%q` formatting on stdout (JSON is fine; console output is inconsistent).

IvGolovach:
I took another pass through the review and treated each point as a harness-contract issue rather than just cleanup.

The update now aligns `--print-plan` with the actual `results.jsonl` names, removes fixed PASS promises from help/docs, fixes parent-shell PID tracking and fail-fast startup handling, and records empty `--version` output as PREREQ instead of a false PASS. I also added focused contract coverage around those paths so these edges should be harder to regress.

Really appreciate the careful review here — these were exactly the right places to tighten.

ndizazzo (approved):

PR #559, opened 2026-05-15
chore(ui-size): adjust UI size to use more horizontal space

ndizazzo:
## What Changed?

This PR adjusts sizing of the app to be more readable. Not perfect, but work ongoing.

## Screenshot
image

## Adjustments
* Font size 8% larger
* Padding adjustments everywhere
* Layout fixes
* Settings sizing fixes / alignment corrections

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,3 +6,10 @@ rustflags = ["-C", "link-arg=-Wl,-z,max-page-size=16384"]
 
 [target.x86_64-linux-android]
 rustflags = ["-C", "link-arg=-Wl,-z,max-page-size=16384"]
+# BEGIN Mesh-LLM lld config
+[target.aarch64-apple-darwin]
+rustflags = ["-C", "link-arg=-fuse-ld=/opt/homebrew/bin/ld64.lld"]
+
+[target.x86_64-apple-darwin]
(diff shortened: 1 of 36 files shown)


.cargo/config.toml
```diff
@@ -6,3 +6,10 @@ rustflags = ["-C", "link-arg=-Wl,-z,max-page-size=16384"]
 
 [target.x86_64-linux-android]
 rustflags = ["-C", "link-arg=-Wl,-z,max-page-size=16384"]
+# BEGIN Mesh-LLM lld config
+[target.aarch64-apple-darwin]
```
michaelneale, line 10: any reason why these are needed?

michaelneale (approved):

.cargo/config.toml
ndizazzo: Every time I ask an agent to fix tests, it adds this to the code repeatedly. Local test runs fail with an error without it.

I think it has to do with the lld linker changes we had a while back. It's happened on every PR for the last 3-4 now, so I finally gave in and let it have it's way... 

@michaelneale want me to remove it?

michaelneale: Probably is ok then as long as works for you?

PR #560, opened 2026-05-15
mockup: Reserves high-fidelity UI mockup

ndizazzo:
## What Changed?

This PR adds a **mockup** for reserved capacity from cloud providers / local capacity as idle machines. 

This feature would make it possible (**but is not designed yet**) to have a fleet of mesh-capable nodes that can be CONTROLLED to determine whether they join the active mesh or not.

## Screenshots
Screenshot 2026-05-15 at 4 45 43 PM
Screenshot 2026-05-15 at 4 45 52 PM

## Plan

1) Add high fidelity mockup behind UI feature flags
2) Iterate on core idea
3) Propose changes to mesh-llm backend to support reserve capacity
4) Wire up UI to completed backend components

## Caveats

Since we don't want to have setting and coordinator synchronization concerns with transient mesh nodes, we likely need a CENTRALIZED EXTERNAL COORDINATOR SERVICE for to manage mesh nodes. This should either be offered, or self-hosted as well. This solution needs discussion.


michaelneale:
I do like this design!

PR #562, opened 2026-05-16
chore(version): synchronize version bump everywhere

ndizazzo:
Simple version bump - we had a lot of placeholder versions I sync'd. Not sure if we want this though.


michaelneale:
does the release job not do this? (or pipeline?) confused how it got here.

ndizazzo:
> does the release job not do this? (or pipeline?) confused how it got here.

@michaelneale The release job only bumped versions in select packages and on the final binary, this PR makes all of the crates have a version that's tied to the workspace version (0.6.6 currently). Before, they were just hardcoded at 0.1.0. I figured if we were going to publish crates we'd want this. 

I haven't changed how the release job bumps the versions, it just needs to change less now that the crates are using an "inherited" version number from the cargo workspace

michaelneale (approved):
just waiting for it to be clean

PR #563, opened 2026-05-16
auto: retry with fallback models when peer returns 5xx

michaelneale:
## What changed

When using `model: auto`, if the router picks a model whose only peer returns a 5xx server error (e.g. an old binary that crashes on tool calls), the request now automatically retries with the next model in the ranked list instead of forwarding the error to the client.

This fixes the issue where `auto` with Goose or any tool-using client would fail ~50% of the time because the router randomly picked a broken peer.

### Before
```
auto attempt 1: FAIL — skippy ABI call failed: RuntimeError: llama_decode failed
auto attempt 2: OK (tool_calls) → MiniMax
auto attempt 3: FAIL — skippy ABI call failed: RuntimeError: llama_decode failed
```

### After
```
auto attempt 1: OK (tool_calls) → MiniMax  (fell back from broken 32B peer)
auto attempt 2: OK (tool_calls) → MiniMax
auto attempt 3: OK (tool_calls) → MiniMax
```

Goose agentic workflow (read → edit → verify) now works reliably with `GOOSE_MODEL=auto`.

## How it works

1. **`RetryableServerError` variant** — when a remote peer returns 5xx, the response is buffered without writing to the client stream, and the attempt is marked retryable.

2. **`try_route_model_request`** — new function that returns `RouteModelResult::AllTargetsServerError` when every target for a model returned 5xx, allowing the caller to try a different model.

3. **`rank_models_classified`** — returns all candidate models in priority order (not just the top pick), giving the auto router a fallback list.

4. **Auto fallback loop in ingress** — when the primary model fails, iterates through fallback models until one succeeds or all are exhausted.

## Validated

- 10/10 `auto` + tools requests succeed (was 3/5 before)
- Goose + auto: full agentic edit workflow (cat → edit → python3 verify) ✅
- 1341 lib tests pass, 0 failures
- Existing `route_model_request` callers unchanged (wrapper preserved)

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
@@ -157,7 +157,9 @@ pub(crate) async fn api_proxy(
                         return;
                     }
 
-                    let (effective_model, classification) = if request.model_name.is_none()
+                    let (effective_model, classification, auto_fallback_models) = if request
+                        .model_name
+                        .is_none()
                         || request.model_name.as_deref() == Some("auto")
                     {
                         request.ensure_body_json();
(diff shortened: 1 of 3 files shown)


crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -308,23 +323,104 @@ pub(crate) async fn api_proxy(
...
                                 )
                                 .await;
-                                proxy::release_request_objects(
-                                    &node,
-                                    &request.request_object_request_ids,
-                                )
-                                .await;
-                                debug_assert!(routed);
-                                return;
+                                match result {
```
ndizazzo, line 336: We're getting a bit nesty here... it seems like returning a routing result directly from the ingress seems like we could organize this a bit better - maybe passing off to some kind of builder-pattern for a given routing request?

ndizazzo (approved):
Looks fine to me... nothing really to add on top of what copilot mentioned

michaelneale:
Superseded by cleaner rewrite — see new PR

PR #564, opened 2026-05-16
feature(configuration): add as many configuration settings from llamacpp to wiring

ndizazzo:
## Summary
Operators can now tune model behavior through a rich, structured TOML configuration rather than being limited to a handful of top-level fields. 

This PR wires the full set of llama.cpp-level settings into the config schema and resolver pipeline, so tuning preferences (context size, KV cache policy, hardware placement, throughput profiles, speculative decoding, request defaults, and more) are expressed once in `~/.mesh-llm/config.toml` and automatically applied at request time—without needing to pass them on every API call.

- `[defaults.*]` and per-model `[models.*]` config blocks now supported for 8 setting namespaces:
  `model_fit`, `hardware`, `throughput`, `skippy`, `speculative`, `request_defaults`, `multimodal`,
  `advanced`.
- Multi-level resolver merges: per-model → global defaults → family policy → built-in fallbacks; explicit request values always win.
- Request defaults applied at the OpenAI frontend boundary only; never propagate into `StageConfig`  or the wire protocol.
- Macro presets (`kv_cache_policy`, `tuning_profile`) expand into concrete settings at resolve time;  explicit fields always override.
- 804 new resolver tests + 242 frontend tests.
- `docs/skippy/CONFIGURATION.md` — new operator reference documenting every config key path,  supported modes, translation targets, validation rules, and rejection boundaries.
- `docs/skippy/manual-smoke/` — Python smoke test harness covering all config categories.
- `docs/USAGE.md` and `docs/CLI.md` updated with annotated TOML examples.

## Example

```toml
# ---------------------------------------------------------------------------
# Example 3: MoE model with CPU expert offload
# ---------------------------------------------------------------------------
[[models]]
model = "bartowski/Mixtral-8x7B-Instruct-v0.1-GGUF/Mixtral-8x7B-Instruct-v0.1-Q4_K_M.gguf"

[models.hardware]
device         = "cuda:0"
gpu_layers     = 32
cpu_moe        = true
n_cpu_moe      = 4             # route 4 experts to CPU
split_mode     = "row"
placement      = "auto"
fit_target_mib = 20480

[models.model_fit]
ctx_size        = 131072
kv_cache_policy = "saver"

[models.throughput]
parallel = 2
threads  = 8

[models.advanced.server]
alias = "mixtral-8x7b"
```

## Architecture

Resolver pipeline lives in `crates/mesh-llm-host-runtime/src/inference/skippy/resolver/`:
- `resolution.rs` — top-level entry point; drives the merge cascade
- `request_defaults.rs` — merges `request_defaults` config into OpenAI request fields
- `speculative.rs` — speculative decoding config resolution
- `translation.rs` — maps resolved config to `StageConfig` / llama.cpp args
- `support.rs` / `types.rs` — shared resolver types and helpers

Request-defaults merging (`apply_chat_request_defaults` / `apply_completion_request_defaults`) in`crates/skippy-server/src/frontend/request.rs`.

Live config state tracking in `crates/mesh-llm-host-runtime/src/runtime/config_state.rs`.


ndizazzo:
UI PR to follow, supporting configuration of these settings through the web console

michaelneale:
some feedback from a few agents: The resolver's output is never consumed by runtime/local.rs — resolve_skippy_config and
 ResolvedSkippyConfig::to_{model_load_options,stage_config,embedded_openai_args} have no production callers, so
 all new non-legacy keys (request_defaults, skippy.*, speculative.*, throughput.threads,
 hardware.stage_layer_*/placement/tensor_split/cpu_moe, etc.) are dead at launch.

michaelneale (changes requested):
looks like things are not fully wired in (are they meant to be?)

ndizazzo:
@michaelneale @i386 could I get a re-review here?

i386:
Code review findings:

1. **Pinned defaults are effectively unusable** (`crates/mesh-llm-host-runtime/src/plugin/config.rs`)

   `validate_config` validates `[defaults]` and each `[[models]]` entry independently. Under `gpu.assignment = "pinned"`, `[defaults.hardware] device = "..."` does not satisfy per-model validation because `validate_model_entry` never receives defaults. Conversely, if every model has a device but `[defaults]` exists for unrelated defaults, `validate_model_defaults` can still require `defaults.hardware.device`. This contradicts the docs saying pinned device can be in every model “or in `[defaults.hardware]`”, and rejects valid-looking generated configs.

2. **Large `hardware.gpu_layers` values wrap instead of being rejected** (`crates/mesh-llm-host-runtime/src/inference/skippy/resolver/support.rs`)

   `parse_gpu_layers` casts `i64` to `i32` with `as`, while config validation only checks `>= -1`. A value like `2147483648` passes validation and then wraps to a negative `i32` before reaching `StageConfig.n_gpu_layers`, causing an invalid or unintended launch setting. This should use `i32::try_from` during validation/resolution.

3. **Invalid full config payloads silently degrade to legacy fields** (`crates/mesh-llm-host-runtime/src/protocol/convert.rs`)

   `proto_config_to_mesh` falls back to the legacy snapshot if `config_toml` is present but fails to parse. In owner-control apply, that means a malformed full payload can still apply as a degraded legacy config, dropping nested defaults instead of rejecting the request. For apply paths, presence of `config_toml` should probably be authoritative: parse failure should be a bad request, not data loss.

Validation note: I attempted a focused `cargo test -p mesh-llm-host-runtime ...` in a temporary PR worktree, but it failed before running tests because direct Cargo could not find the prepared `llama-common` native library. This repo’s `just` build prep is required for that path.

PR #565, opened 2026-05-17
auto: retry with fallback models when peer returns 5xx

michaelneale:
When `model: auto` picks a model and the peer returns a 5xx server error (e.g. a stale binary crashing on tool requests), the router now tries the next ranked model instead of forwarding the error to the client.

## What changed

Previously, 5xx responses were relayed to the client immediately — the model tried, it failed, done. But a broken peer (stale binary, ABI crash on tools) isn't a legitimate model failure — it's an availability issue. Auto should route around it.

Now `relay_probed_response` intercepts 5xx before writing to the client stream when `retry_server_error` is set. The existing target loop retries other targets for the same model (same as timeout/unavailable). If all targets for that model fail, the ingress auto-routing loop tries each fallback from `rank_models_classified` in priority order.

Only auto-routed requests retry across models. Explicit model requests still forward the real 5xx.

## Changes

- `RetryableServerError` variant on `RouteAttemptResult` — intercepts 5xx before relay  
- `retry_server_error` param threaded through `relay_probed_response` → routing fns
- `drain_response` helper — consumes upstream response without relaying to client
- `RouteModelResult` enum — `route_model_request` returns `Handled | AllTargetsServerError`
- `route_model_request` takes `&mut TcpStream` (was owned) so caller can retry
- `rank_models_classified` returns all candidates in priority order (was `pick` → single)
- Auto fallback loop in ingress when primary model's targets all 5xx

## Context

A public mesh peer running Qwen3-32B on a stale binary (v0.65.1) crashes with `llama_decode failed` on any request containing tools. Auto picks 32B ~50% of the time → ~50% Goose tool failures. After this fix, the failed 32B request falls back to MiniMax or another model automatically.

Supersedes #563 (rewritten from scratch — cleaner, ~half the diff).

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
@@ -157,7 +157,9 @@ pub(crate) async fn api_proxy(
                         return;
                     }
 
-                    let (effective_model, classification) = if request.model_name.is_none()
+                    let (effective_model, classification, auto_fallback_models) = if request
+                        .model_name
+                        .is_none()
                         || request.model_name.as_deref() == Some("auto")
                     {
                         request.ensure_body_json();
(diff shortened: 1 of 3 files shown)


michaelneale:
actually not warranted. the problem is an outdated node

PR #566, opened 2026-05-17
MoA: mesh mode and many inference critical fixes, and quic keep alive

michaelneale:
## What this is

**Mixture-of-Agents (MoA) as a mesh-native routing mode.** A client requests `model: "mesh"`; the gateway fans out in parallel to N heterogeneous workers across the mesh, a deterministic arbiter returns the first consensus answer, and a hedged reducer ladder produces an arbitrated answer when workers disagree. Same OpenAI API surface as a normal chat completion.

WHY: on a wan mesh where RTT (round trip) is too high to practically do pipeline splits, we can "fall back" to having many more nodes run distinct models, and use this MoA approach to improve quality of inference without being bound to accumulated latency (and it also can cope with some nodes being intermittent). This is complementary for mesh and would want to use it (if it can work!) in cases where we cannot do a split! but lets us make use effectively of vram. 

## Works from any node

The MoA gateway runs **wherever the request lands** — on a serving host, a standby, or a pure `--client --auto` node with zero local models. The worker pool is built from gossip: every model advertised by any peer is a candidate. On a host node, locally-served models are wired directly to their skippy port. On a client node, all backends are remote QUIC tunnels.

```bash
mesh-llm client --auto
curl http://localhost:9337/v1/chat/completions -d '{
  "model": "mesh",
  "messages": [{"role":"user","content":"What is 11 times 7?"}]
}'
```

```
x-moa-elapsed-ms: 1330
x-moa-turn: early-exit
x-moa-workers: 2
x-moa-workers-ok: 2
x-moa-reducer: false
x-moa-reducer-attempts: 0

{"choices":[{"finish_reason":"stop","message":{"content":"11 times 7 equals **77**."...
```

## Live-verified end-to-end

From a `--client --auto` node on a laptop, joined to the public mesh (4 reachable models, all remote):

**Plain chat** — `x-moa-workers: 2`, `early-exit`, 1.3s, clean answer.

**Goose agentic loop with developer extension** — fan-out, tool proposal, shell exec, tool-result turn, final answer. Server log of the live run:

```
MoA config: 4 workers (0 local, 4 remote)
moa: turn=Fresh, 4 models, tools=true
moa: dispatching to 4 workers: [Qwen3.5-9B(fast), Qwen3-8B(specialist), Qwen3-32B(specialist), MiniMax-M2.5(strong)]
moa: worker Qwen3-32B failed after 321ms: HTTP 502 Bad Gateway
moa: worker Qwen3-8B → ToolProposal conf=0.90 (2670ms)
moa: worker MiniMax-M2.5 → ToolProposal conf=0.90 (2802ms)
moa: early exit — 2 workers agree on tool 'shell', 1 still pending
moa: 2803ms, 2/3 workers, kind=early-exit
```

Tool-result turn went through reducer fallback with sole-survivor recovery; goose printed `DONE` and exited 0.

## Per-turn observability via response headers

`x-moa-*` headers (ignored by normal OpenAI clients) expose what happened:

| header | meaning |
|---|---|
| `x-moa-elapsed-ms` | wall-clock for the MoA turn |
| `x-moa-turn` | `fanout` / `early-exit` / `tool-result` / `failed` |
| `x-moa-workers` | configured worker count |
| `x-moa-workers-ok` | workers that returned successfully |
| `x-moa-reducer` | reducer model name (or `false`) |
| `x-moa-reducer-attempts` | candidates actually spawned (accurate on success **and** failure) |

## Architecture

New crate `crates/mesh-mixture-of-agents` (standalone, no host-runtime dep):

| Module | Responsibility |
|---|---|
| `lib.rs` | Gateway entrypoint, `GatewayConfig`, `TurnResult`, response builders |
| `backend.rs` | `ModelBackend` trait, HTTP backend, retry-after parsing |
| `fanout.rs` | Incremental worker gather with early-exit |
| `reducer.rs` | Candidate selection + hedged ladder; carries `attempts` on **both** Ok and Err |
| `tool_guard.rs` | Filter hallucinated tool names against declared `tools` |
| `arbiter.rs` | Deterministic arbitration (consensus + tool-name agreement) |
| `normalize.rs` | 3-tier dirty-output parsing |
| `session.rs` | Canonical transcript, tool tracking, turn classification |
| `context.rs` | Role-shaped context packing (fast / specialist / strong / reducer) |
| `worker.rs` | Size-aware role assignment, canonical `strip_thinking` + `truncate_chars` |

Host-runtime integration in `crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs`. **Single** implementation called from both `api_proxy` (host path) and `handle_mesh_request` (client/passive path), so MoA behaviour is identical regardless of which path the request takes.

### Mesh-wide orchestration from any node
- `build_moa_config` now reads `Node::models_being_served()` (mesh-wide union of local + every peer's advertised models). Stopped depending on the local routing table for the worker list.
- Local routing table is now only used as an *optimization*: locally-served models bypass the QUIC tunnel and go straight to a skippy port. Pure client nodes use all-remote backends.
- `model: "mesh"` intercept added to `handle_mesh_request` so client-mode nodes orchestrate locally instead of forwarding verbatim to a random peer.

### Reducer-attempts accounting
- `hedged_reducer_call` now returns `Result<HedgedReducerOk, HedgedReducerErr>` where `Err` carries `attempts` too. Previously the all-fail path reported `attempts=0` regardless of how many candidates actually ran, producing nonsense like *"Reducer failed (tried 0): remote timeout after 15s"*. Surfaced by the live goose test.

### Code quality from Copilot review
- UTF-8-safe truncation helper (`moa::truncate_chars`) replaces 2 panicking byte-slice sites. New tests pin char-boundary handling for both 2- and 3-byte sequences.
- Remote read cap bumped 256 KiB → 4 MiB.
- CR/LF sanitization on `x-moa-*` header values.
- Dead `("unknown", 0)` defensive fallback removed from `reducer_candidates` — was masking real "no candidates" bugs by silently calling `backend_index=0` with a bogus model name.
- Three byte-identical `strip_thinking` copies consolidated onto one canonical implementation.
- Tracing warnings on response-write failures rather than silently swallowing.

### MoA crate refactored (earlier in this branch)
- `lib.rs` 1267 → 454 LoC; split into `backend` / `reducer` / `fanout` / `tool_guard`.
- 8 new `context.rs` tests pin role-shaped packing contract.
- Total now 63 unit tests in `mesh-mixture-of-agents`.

### `auto` routing is now tok/s-aware
Independent of MoA but landed in the same branch because it surfaced during public-mesh testing. `pick_model_classified` was picking uniformly inside a tier despite `RoutingMetrics.avg_tokens_per_second` being recorded all along. Now: `RoutingCandidate` struct carries `tps_hint` + `throughput_samples`; weighted pick with `weight = tps.clamp(5, 100)` when samples ≥ 3, neutral 25 otherwise; 15% exploration probability prevents lock-in. 5 new behaviour tests; all 17 router tests pass.

Closes #534.


crates/mesh-mixture-of-agents/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+                }) => (spawned, None),
+            };
+
+            match chosen {
+                Some(reduced) => match reduced.kind {
+                    normalize::OutputKind::ToolProposal => {
+                        if let (Some(name), Some(args)) =
+                            (reduced.tool_name.as_ref(), reduced.tool_arguments.as_ref())
+                        {
+                            (tool_call_response(name, args), true, attempts)
```
ndizazzo: This reducer path can emit a `tool_calls` response even when the original request didn't declare  tools. `enforce_allowed_tools` returns early for an empty allowed list, and this conversion does not check `has_tools`.

That can produce a function call the client never advertised and may not be able to execute. 

Could we demote/suppress tool proposals whenever `has_tools == false`, including reducer outputs?

ndizazzo:
@michaelneale I spent some time manually using MoA through the chat, and then had my 3 nodes on this build (clean, rebuilt) with Qwen3.5/3.6 models that I know perform decently. It seemed okay, but the only real difference I noticed was the how requests were flowing. That part seemed to work. Then, I hammered it with Opencode, and generated some automated calls to endpoints to collect multiple runs of data... I had the data output to JSON / raw text, and then distilled it with my agent:

---

# MoA PR 566 high-level report

## Executive summary

The MoA API path handled basic requests successfully. In testing, `model: "mesh"` returned OpenAI-shaped responses, included MoA-specific headers, and supported streaming through the same endpoint.

The happy path works via chat often enough to be promising, but the system feels inconsistent around routing, worker accounting, exact-answer quality, and tool handling.

## What models were in the run

The test used three machines and four configured model workers, on the MoA branch with clean builds:

- **Carrack** was the main gateway for most client requests.
  - It hosted `unsloth/Qwen3.6-27B-GGUF:UD-Q4_K_XL`.
  - It also hosted `unsloth/Qwen3.5-4B-GGUF:UD-Q4_K_XL`.
- **Lemony** hosted `unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL`.
- **Lemony29** hosted `unsloth/Qwen3.5-9B-GGUF:UD-Q8_K_XL`.

So the intended MoA pool was:

- Carrack 27B
- Carrack 4B
- Lemony 35B
- Lemony29 9B

## Where requests actually routed

The routing story is mixed. The user-facing MoA route partially worked, local direct routing worked, and direct routing from the main gateway to worker-hosted models failed.

For normal MoA use, the important model name is `mesh`. In PR 566, `mesh` is a virtual model route. It gets intercepted before normal model routing, and it isn't meant to be advertised as one of the concrete models in `/v1/models`.

A client that already knows to call `model: "mesh"` doesn't need `/v1/models` first, and it doesn't need to call the individual worker model IDs.

We checked `/v1/models`, `/api/model-targets`, and direct worker-model calls as diagnostics. Those checks tested whether the gateway had a consistent view of the concrete worker pool behind `mesh`, whether model IDs stayed stable, and whether worker-hosted models were reachable through the normal non-MoA routing path.

- **Direct calls to Carrack's local models worked.**
  - Carrack 27B returned HTTP 200.
  - Carrack 4B returned HTTP 200.
  - So local model routing through the gateway was functional.

- **Direct calls from Carrack to worker-hosted models didn't work.**
  - Carrack to Lemony 35B returned HTTP 404.
  - Carrack to Lemony29 9B returned HTTP 404.
  - This doesn't block a client that only calls `model: "mesh"`.
  - It does show that the gateway's diagnostic/direct-routing view of the worker pool was incomplete.

- **The virtual `model: "mesh"` route used the MoA path, but worker participation varied.**
  - `x-moa-workers` values were 2, 3, and 4 across the run.
  - `x-moa-workers-ok` values ranged from 0 to 3.
  - Some MoA calls used multiple workers successfully.
  - The run wasn't stable enough to say “all four models answered reliably.”

- **The artifacts didn't give exact per-response model attribution.**
  - They showed the configured pool.
  - They showed MoA worker-count headers.
  - They didn't give a clean per-response list of which model names contributed to each `model: "mesh"` answer.
  - A successful MoA response shouldn't be read as “all four models answered.”

- **Origin testing worked at the basic API level.**
  - Carrack returned the expected origin-specific answer.
  - Lemony returned the expected origin-specific answer.
  - Lemony29 returned the expected origin-specific answer.
  - The worker counts still differed by origin, so model participation was inconsistent.

## What looks ready

Several basic API behaviors are in place:

- The `model: "mesh"` request path is present and functional for simple calls.
- The smoke test returned the exact expected answer.
- Responses had the expected OpenAI-compatible shape.
- MoA headers confirmed that requests went through the MoA path.
- Streaming returned valid SSE framing and ended with `[DONE]`.
- The too-few-models case returned a clear HTTP 503 with a useful error message.
- The reduced network impairment lane returned HTTP 200 across the tested delay/loss cases.
- Churn tests kept returning HTTP 200 while workers were stopped and restarted.

That failure behavior for too few models is especially good. It's the kind of response clients can handle cleanly.

## What worked well

MoA did well on some simple, constrained prompt types:

- Exact string reproduction passed cleanly.
- Structured JSON was strong.
- Order-sensitive prompts performed well.
- Benign refusal-style prompts performed well.

The small tool-call probes were also encouraging:

- MoA didn't duplicate tool calls.
- The fake side-effect test didn't repeat the side-effectful call.
- That matters because duplicate tools are one of the fastest ways an agent system becomes unsafe.

Raw OpenAI compatibility is partially there:

- A raw tool-schema request through the MoA endpoint worked.
- The long-prompt extraction probe returned the expected sentinel string.
- OpenCode could be configured to send requests to the MoA endpoint.

In short: clients can talk to the endpoint. Basic chat works. Basic streaming works. Some classes of exact output work.

## What didn't work

The biggest issue isn't that users must call the worker model IDs. They normally shouldn't. The issue is that the gateway's view of the concrete worker pool behind `mesh` was inconsistent:

- `model: "mesh"` answered, but it didn't consistently show that all configured workers participated.
- The main gateway didn't consistently show all worker models in `/v1/models`.
- `/api/model-targets` didn't expose all worker models as routable targets.
- Direct calls to worker-hosted models through the gateway returned HTTP 404.

Model ID inconsistency is a secondary signal, not the main user-facing MoA problem:

- Some IDs in `/v1/models` dropped quant suffixes.
- Other endpoints used the full model refs.
- That matters for diagnostics, direct model calls, UI pickers, validation, and generic clients that inspect available models.
- It matters less for a client that only ever calls `model: "mesh"`.
- It doesn't mean `model: "mesh"` itself needs to appear in `/v1/models`.

Worker accounting was inconsistent:

- Similar requests reported different `x-moa-workers` values.
- Similar requests reported different `x-moa-workers-ok` values.
- Some successful responses used fewer workers than expected.
- Some churn responses still appeared to report stale worker counts.
- One concurrency response reported zero successful workers.

Failure signaling needs work:

- One concurrency request returned HTTP 200 even though the response body said all MoA workers failed.
- That's a bad client contract.
- If all workers fail, the API should probably return a proper error, not a successful-looking response with failure text inside it.

Correctness was mixed:

- MoA handled some simple prompt types pretty well.
- It struggled with prompts that required short, exact answers.
- The numeric probe was only 5/10 correct.
- The negation/direct-answer probe was only 1/10 correct.
- A common failure mode was exposing reasoning text or giving a verbose answer when the task needed a direct answer.
- Some of that is probably model behavior, not uniquely MoA behavior.
- The direct single-model baseline also showed reasoning-style output on exact-answer prompts.
- So the result shouldn't be read as “MoA alone caused every wrong answer.”
- The endpoint still owns the user-visible failure, though.
- If MoA is going to sit in front of reasoning-heavy models, the reducer needs to normalize those answers better: strip reasoning, choose the direct answer, and enforce the requested format.

The tool-result path isn't ready for agent loops:

- A tool-result follow-up was treated like another fanout turn.
- It wasn't handled like a controlled reducer/synthesis turn.
- For agent harnesses, that matters a lot.
- Tool results should be handled carefully and predictably, not sprayed back through the whole fanout path.

OpenCode-style usage didn't work:

- The coding fixture didn't edit files.
- The target file stayed unchanged.
- The tests still failed.
- The OpenCode event logs showed text-only turns with zero actual tool-use events.
- In the read-tool probe, the model wrote text that looked like a tool call instead of actually invoking the read tool.

## How it feels to use

For simple chat-style requests, MoA behaves like a promising prototype:

- You can hit the endpoint.
- You can get a response.
- You can see MoA-specific headers, so you know the request went through the MoA path.
- The worker-count headers showed that multiple workers were attempted on some requests.
- Those headers don't identify the exact model names that contributed to each answer, so they shouldn't be read as proof that all four configured models answered every request.

For serious application use, it feels unpredictable:

- Worker counts change in ways that are hard to reason about.
- Remote model routing isn't dependable.
- Some failures come back as HTTP 200 responses.

For agent use, it isn't usable yet:

- It can speak OpenAI-compatible HTTP.
- It doesn't reliably produce actual tool calls through OpenCode.
- It didn't complete a basic coding task.
- That makes it risky for workflows where the model needs to inspect files, edit code, run commands, and continue across tool-result turns.

For quality, the tests didn't prove that MoA is better than one strong direct model:

- MoA was much slower than direct calls.
- The quality gain wasn't consistent enough to justify the added latency in this run.

## What needs to be fixed

The fix list is straightforward:

1. **Model discovery and routing.**
   - The gateway needs a consistent view of which worker models are available.
   - Direct worker-model calls should either route correctly or be clearly documented as unsupported.
   - This is mainly a routing/readiness signal, not the primary MoA user path.

2. **Model ID normalization.**
   - The same model should have the same ID everywhere.
   - That includes `/v1/models`, status output, routing tables, request payloads, and logs.
   - This matters most for direct model calls, diagnostics, and clients that inspect available models.

3. **Worker accounting.**
   - `x-moa-workers` and `x-moa-workers-ok` need to reflect what actually happened for that request.
   - Stale or surprising worker counts make it hard to trust the system.

4. **Failure signaling.**
   - If MoA can't get a valid answer from workers, the API should return a proper error.
   - It shouldn't return HTTP 200 with failure text in the response body.

5. **Reducer behavior, especially after tool calls.**
   - Tool-result turns should go through a controlled synthesis path.
   - They shouldn't behave like ordinary fanout prompts.

6. **Agent-tool compatibility.**
   - OpenCode-style harnesses still need actual working tool calls:
     - actual file reads
     - actual edits
     - stable multi-turn behavior
     - Text that merely looks like a tool call isn't enough

## Bottom line

PR 566 shows that the MoA path can handle basic API traffic. It can answer, stream, emit headers, survive some pressure, and handle some simple correctness tasks well.

The main blockers are routing, worker accounting, failure signaling, reducer/tool behavior, and exact-answer quality.

PR #567, opened 2026-05-17
Speculative prefill decode (experimental, opt-in) — verify draft tokens against target model

michaelneale:
## What this is

**Speculative prefill decode** — an opt-in experimental feature where skippy-server verifies a draft response (text or token IDs) against the target model in a single batched forward pass before entering the decode loop. When the draft is correct, generation skips ahead to the divergence point, reducing time-to-first-token.

This is **specifically for skippy's layer-split decode latency** — not standard speculative decoding. The motivation is that distributed layer-split inference pays per-token network latency on every decode step; verifying many tokens in one batched forward is much cheaper than decoding them sequentially.

**Status: experimental, opt-in, disabled by default.** Zero startup cost, zero per-request overhead unless `--prefill-speculative` is set.

## How it works

1. Client sends a chat completion request with a `draft_response` (text) or `draft_tokens` (pre-tokenized IDs).
2. skippy-server tokenizes the draft (if text), runs `verify_tokens()` in chunks of 8 against the target model's KV cache.
3. Finds the contiguous-acceptance prefix using combined criteria: greedy match OR per-position probability ≥ 0.5.
4. Trims KV cache at the divergence point and continues sequential decode from there.
5. Returns the response including `completion_token_ids` if `return_token_ids: true`.

Capped-draft pattern documented for distributed agentic workloads: a small local model drafts up to N tokens; the strong model verifies and continues.

## Try it

```bash
mesh-llm --model <strong-target-model> --prefill-speculative <small-local-draft-model.gguf>

curl http://localhost:9337/v1/chat/completions -d '{
  "model": "<strong-target-model>",
  "messages": [{"role":"user","content":"What is the capital of France?"}],
  "draft_response": "The capital of France is Paris.",
  "return_token_ids": true
}'
```

When `--prefill-speculative` points at a local GGUF, mesh-llm auto-generates draft tokens via DraftRunner. Otherwise the client supplies the draft via `draft_response` / `draft_tokens`.

## Known limitations (preserved from PoC)

The integration commit (`bdea9529`) documents the experimental finding:

- ✅ **Short factual responses**: 100% acceptance, ~55% time savings
- ⚠️ **Longer responses**: 1–14% acceptance — verify_tokens (batched) produces different logits than sequential decode at the same positions due to float precision differences in batch accumulation
- ✅ Tokenizer round-trip is NOT the cause — token IDs and text produce identical acceptance rates
- ✅ Per-position probability analysis + chunked verify (size 8) was added to investigate, and `docs/design/SPECULATIVE_PREFILL_DECODE_SKIPPY.md` documents the full findings

So this is a **research-grade feature** — works well in narrow regimes (short factual answers), needs more work for general use. The branch lands the wiring and the analysis tooling so the work can continue.

## CLI

- `--prefill-speculative <path/to/draft.gguf>` — local GGUF used as the draft model (DraftRunner)
- `--prefill-speculative-max <n>` (hidden, default 8) — max tokens to draft per request

## Architecture

New code:
- `crates/spec-prefill-poc` — standalone PoC binary that exercises verify_tokens against a target model with various drafts
- `crates/skippy-server/src/frontend/spec_prefill.rs` — production integration: chunked verify, contiguous-acceptance prefix detection, per-position probabilities, KV cache trim on divergence
- `crates/openai-frontend` — request/response surface: `draft_response`, `draft_tokens`, `return_token_ids` fields
- `crates/skippy-server/src/runtime_state.rs` — `verify_tokens()` + `token_logprob_at()` wrappers (local, non-staged)
- One new llama.cpp patch: `0081-Expose-per-position-token-signal-and-logprob-ABI.patch` (read full token signal + log-probability at any batch position after `verify_tokens()`)

Mesh integration:
- `crates/mesh-llm-host-runtime/src/inference/skippy/` plumbs the draft runner through to skippy-server
- `--prefill-speculative` CLI flag wired through node config → server runtime

Docs (`docs/design/SPECULATIVE_PREFILL_DECODE_SKIPPY.md` — 1100+ lines):
- Tokenizer-agnostic design
- Mesh-native draft model selection (context subsetting, candidates, pairings)
- Distinction from standard speculative decoding (complementary, not exclusive)
- Capped-draft pattern for distributed agentic workloads
- PoC results, opt-in design, automation strategy tiers
- Server integration findings (acceptance rate caveats)

## Rebase notes

This branch was reconstructed off fresh `origin/main` from the historical work that was previously interleaved with MoA on `micn/mixture-of-agents`. Two commits were dropped during the rebase because their net effect was undone later in the chain:

- `d4230e41 Auto-detect speculative prefill model from local HF cache` — replaced by explicit opt-in
- `1b58c54a Make speculative prefill best-effort: never fail model loading` — only modified auto-detect code path

The final commit `c6623fdd Make speculative prefill opt-in only` retains its meaningful change (removing the `--no-draft` CLI flag).

Three real conflicts resolved during cherry-pick:
- `skippy-server/src/frontend.rs`, `frontend/backend.rs`, `frontend/generation_flow.rs` — main refactored `parse_tool_call_output` → `parse_chat_output` (with `ParsedChatMessage`) and added `prefill_reply_credit_limit`. Merged carefully to preserve both lines of evolution.

## Status

**Draft.** Compiles clean (`cargo check --workspace` and `just build` both green). Not live-tested in this session — the previous PoC and integration findings are documented and that's the working baseline.

Follow-ups for separate review:
- Investigate batch-vs-sequential numerical divergence (the 1–14% acceptance rate problem)
- Mesh-native draft model selection (currently `--prefill-speculative` is local-only; design doc covers the broader picture)
- Real benchmarks on agentic workloads (capped-draft pattern)
- Cross-tokenizer support


michaelneale:
## Live test findings + K-tolerant acceptance fix

Spent a session live-testing this on the canonical Qwen3-0.6B → Qwen3-8B pair and discovered the production behavior was much worse than the PoC suggested. Root cause turned out to be the **acceptance strategy**, not the FFI or chunked verify path.

### Original symptom

Production showed ~1/11 acceptance on prompts where the PoC binary reported 100%. The per-position probability dump had a suspicious alternating-1.0 pattern that initially looked like an off-by-one indexing bug.

### What it actually was

Wrote a standalone diagnostic that calls `verify_tokens_chunked` + `token_logprob_at` directly against the same models, with and without `configure_chat_sampling` mixed in. **All paths produced correct, identical results.** The C ABI, the chunked verify, and the Rust wrappers were all fine.

The real issue: the production path uses the **draft model's tokens** (from a separate Qwen3-0.6B `DraftRunner`) while the PoC self-drafts with the target. When draft and target are different models, they agree on most positions but rarely have a long *contiguous* matching run. The strict-prefix acceptance loop broke on the first divergence and threw away all subsequent matches.

Concrete data — "List the seven days of the week.":
- Draft proposed: `<think>, \n, Okay, ',', the, user, asked, to, list, the, seven, days`
- Target argmax: `\n, <think>, ',', the, user, is, to, list, the, seven, days, of`
- Match status:  ✓, ✗, ✓, ✓, ✓, ✗, ✓, ✓, ✓, ✓, ✓ (9 of 11)
- Strict-prefix accepted: **1** (breaks at position 1)
- K-tolerant accepted: **11** (absorbs the two isolated rejections)

The "first divergence at position 1" was almost universal because of a Qwen3-8B thinking-template quirk where it emits `<think>` twice in a row. That one off-by-one in template behavior was tanking the whole feature.

### Fix in efb8c716

`MAX_CONSECUTIVE_MISMATCHES: usize = 1` — the acceptance loop walks all positions, tolerates isolated rejections (still emits the draft token, continuing the prefix), and breaks only when a run of >K consecutive mismatches accumulates. On break it rewinds to the last matching position so we never end on a rejected token.

This is explicitly a **latency optimization** — accepting past a mismatch means we emit a draft token the target wouldn't have chosen, so output drifts slightly from pure sequential decode. For chat output it's imperceptible; for strict structured output (JSON, tool calls) operators can set `MAX_CONSECUTIVE_MISMATCHES = 0` to recover strict-prefix behavior. The tradeoff is documented in the module docstring and the design doc has a new "Revised conclusion" section.

### Live test results (10 prompts, Qwen3-0.6B → Qwen3-8B, 12-token drafts)

| Strategy | Mean accepted | Mean raw matches | Range |
|---|---|---|---|
| Strict prefix | ~9% | ~9% | 0-11/11 |
| K-tolerant (K=1) | **~95%** | ~86% | 9-11/11 |

Every prompt across the sweep accepted 9-11 of 11 draft tokens. Emitted output was fluent and on-topic for list-of-days, capital-of-australia, math, code-completion, haiku, translation, open-ended explanation. Drift at tolerated positions was typically a word-level synonym choice (e.g. `is` vs `asked`) that the surrounding sentence absorbed cleanly.

### Telemetry

`SpecPrefillResult` now carries `raw_matches` alongside `accepted_tokens`, surfaced both in the `spec_prefill:` stderr line and as separate `raw_matches` / `tolerated_mismatches` attrs on the `stage.openai_spec_prefill` event. Operators can see how much of the accepted prefix came from real agreement vs tolerated drift.

### Updated design doc conclusion

The original "1-14% acceptance, feature is research-grade" finding has a new "Revised conclusion: K-tolerant acceptance" section that updates the diagnosis and records the live-test data. The original conclusion is preserved verbatim above the revision for context.

### Still in scope for follow-ups
- Tune `MAX_CONSECUTIVE_MISMATCHES` per-request via a CLI/API knob
- Sparse-splice strategy (re-decode at mismatches instead of accepting drift) — more complex, faithful, requires KV manipulation
- True speculative-decode style with rejection sampling (mathematically equivalent to target distribution; needs ABI work to expose draft logprobs at decode time)

michaelneale:
CI green ✅ — both previously-failing checks now pass:

-  (was: missing  workspace member)
-  (was: smoke scripts still passing removed  flag)

Status: 20 pass / 2 pending heavyweight builds / 0 fail.

PR is ready for review.

michaelneale:
### Design doc updated with current state + next steps

Added a **"Current State (May 2026)"** section near the top of [`docs/design/SPECULATIVE_PREFILL_DECODE_SKIPPY.md`](https://github.com/Mesh-LLM/mesh-llm/blob/micn/prefill-draft/docs/design/SPECULATIVE_PREFILL_DECODE_SKIPPY.md#current-state-may-2026) so readers landing on this PR can get the snapshot in 30 seconds:

- **What is wired up** — `--prefill-speculative` CLI flag (opt-in), `DraftRunner`, K-tolerant acceptance, telemetry, PoC binary
- **What is intentionally NOT wired up** — auto-detect, protocol changes, mesh-native draft selection, streaming during verify
- **Measured single-node performance** — ~240ms overhead vs ~18ms saved per accepted token; 95% mean acceptance with K-tolerant
- **Cost/benefit picture** — break-even on local 8B, clear win at 32B+ local, large win in distributed pipeline parallel (the real target)
- **Six concrete next-step experiments** to move from validated PoC → ready to merge:
  1. Larger local target (32B / 30B-A3B)
  2. 2-node skippy pipeline split
  3. Tool-call / structured-output drift study (K=0 vs K=1)
  4. Mesh-native draft selection over QUIC tunnel
  5. Calibrate `SPEC_PREFILL_DRAFT_CAP` under K-tolerant
  6. `skippy_verify_tokens_frame` distributed validation

This is the recommended reading order now: design-doc top → revised K-tolerant conclusion → concrete next steps.

Branch stays draft. Performance is the right *shape* on the single-node 8B case but the meaningful payoff lives at larger model sizes and distributed splits, which is what experiments 1 and 2 will confirm.

PR #570, opened 2026-05-17
Gate Skippy split planning by stage protocol generation

IvGolovach:
## Summary

Skippy split coordinators now advertise and consume an explicit current stage protocol generation capability before treating peers as stage-compatible.

- Adds a `stage-generation-2` feature to the existing `skippy-stage/1` subprotocol advertisement.
- Persists stage protocol generation support through proto gossip, `PeerAnnouncement`, `PeerInfo`, and direct/transitive peer updates.
- Excludes peers from split participant collection when they want the model but do not advertise the current stage protocol generation, with the explicit exclusion reason `stage_protocol_generation`.
- Decouples stage-control stream selection from artifact-transfer support so artifact transfer is no longer used as a proxy for stage protocol compatibility.

## Why

Stage control frames already validate `STAGE_PROTOCOL_GENERATION`, but split planning did not have a first-class way to know whether a peer could speak the current generation before probing inventory or issuing stage-control requests. In mixed-version meshes, that can fail late and make eligibility look like a model-source or transfer problem instead of a protocol compatibility problem.

This makes the compatibility boundary explicit: new peers advertise support, and new coordinators keep older/non-advertising peers out of current-generation Skippy split plans.

## Protocol compatibility

This is an additive protocol change.

- The existing `skippy-stage/1` subprotocol remains unchanged.
- The new `stage-generation-2` marker is advertised as an additional subprotocol feature.
- Older nodes should ignore the unknown feature as before.
- New coordinators require the marker only for current-generation Skippy split participation.
- No protobuf fields are removed or repurposed, no ALPN major version changes, and no legacy mesh compatibility path is removed.

## Implementation

- Adds `STAGE_SUBPROTOCOL_FEATURE_STAGE_GENERATION` in `skippy-protocol`.
- Round-trips stage generation support in proto conversion and peer state.
- Updates gossip meaningful-change detection, direct peer refresh, transitive peer refresh, and self-announcement generation.
- Adds a split participant preflight helper so client / VRAM / model-interest / protocol-generation exclusions are classified before any inventory probe.
- Keeps existing artifact-transfer trust policy behavior unchanged.

## Validation

- `git diff --check`
- `git diff --cached --check`
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime stage_compatible --lib`
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime owner_fields_roundtrip_through_proto_announcement --lib`
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime split_peer_preflight_requires_current_stage_protocol_generation --lib`
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime stage_protocol_generation_support --lib`
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime split_topology_planner_reports_exclusions_on_capacity_failure --lib`
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo check -p mesh-llm-host-runtime`
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo check -p mesh-llm`

Not run: `just build` — not required for this Tier 2 runtime/protocol change; targeted Rust protocol/runtime checks cover the changed behavior.

## Rollback

Revert this PR.

## Known residual risk

I did not run a live two-node mixed-version mesh in this local pass. The PR adds targeted protocol conversion, gossip propagation, and runtime eligibility coverage for the compatibility gate itself.


crates/skippy-protocol/src/lib.rs
```diff
@@ -12,6 +12,7 @@ pub const STAGE_ALPN_V1: &[u8] = b"skippy-stage/1";
 pub const STAGE_SUBPROTOCOL_NAME: &str = "skippy-stage";
 pub const STAGE_SUBPROTOCOL_MAJOR: u32 = 1;
 pub const STAGE_SUBPROTOCOL_FEATURE_STAGE_CONTROL: &str = "stage-control";
+pub const STAGE_SUBPROTOCOL_FEATURE_STAGE_GENERATION: &str = "stage-generation-2";
```
ndizazzo: Why `...-2`?

crates/mesh-llm-host-runtime/src/runtime/local.rs
```diff
@@ -2168,6 +2142,42 @@ async fn collect_split_participants(
...
+    peer: &mesh::PeerInfo,
+    model_name: &str,
+    model_ref: &str,
+) -> Option<SplitParticipantExclusionReason> {
+    if matches!(peer.role, NodeRole::Client) {
+        return Some(SplitParticipantExclusionReason::Client);
+    }
+    if peer.vram_bytes == 0 {
+        return Some(SplitParticipantExclusionReason::MissingVram);
+    }
```
ndizazzo: Is this a rudimentary way of ensuring the peer can host? 

IE: `client` don't have VRAM reported. In that case, we should likely check the peer type instead of / in addition to the VRAM.

IvGolovach:
Thanks, good catches. I took another pass over both spots and tightened the code so the intent is less implicit.

For the `...-2` question: the suffix is intentional because this is tied to `STAGE_PROTOCOL_GENERATION == 2`, not generic stage-control support. I kept the wire token stable, renamed the internal constant to make the V2 meaning explicit, and added a small assertion so the feature name stays aligned with the current generation.

For the client/VRAM question: the behavior was already role-first, but the code made that too easy to miss. I split stage host eligibility out from the capacity check and added focused coverage for “client with VRAM is still excluded by role” and “worker with zero VRAM is excluded by capacity.”

Also rebased this onto current `main` before pushing the update, so it includes the latest split recovery work cleanly. 🙂

PR #576, opened 2026-05-18
mesh: drop peers below v0.60.0 from gossip ingest and re-broadcast

michaelneale:
## Summary

Defend the local mesh table from gossip-layer noise — specifically the v0.57.0 client swarm currently visible on the public mesh — with two narrowly-scoped ingest gates and a belt-and-braces outbound filter, all in `crates/mesh-llm-host-runtime/src/mesh/gossip.rs`.

## What it does

1. **Version floor at `v0.60.0`** — reject peers below the floor in `add_peer` (direct ingest), `update_transitive_peer` (transitive ingest), and `collect_announcements` (outbound). `v0.60.0` is the cut where the on-wire `hardware` block landed; below it, gossip metadata is missing fields the current mesh relies on. Build metadata (`+skippy.…`) and pre-release tags (`-rc5`) are stripped before comparison. Peers that advertise no version at all are conservatively accepted.

2. **Drop idle transitive clients** — in `update_transitive_peer` only, refuse to ingest `Client`-role peers that aren't asking for any model, aren't serving anything, and aren't hosting anything. Such peers contribute nothing through transitive propagation:
   - not routable to (no model to serve)
   - not findable (clients-don't-dial-clients by design)
   - no demand signal (empty `requested_models`)
   - not relaying for us (purely transitive means there's no connection through which they could relay)

   Direct ingest in `add_peer` is unaffected — a client we actually connect to is still admitted regardless of what they advertise.

## Live observation (before / after)

Same machine, same public mesh, fresh `client --auto` runs:

| | `main` | this branch |
|---|---:|---:|
| Total peers in `/api/status` | 329 | **8** |
| v0.57.x peers | 306 | 0 |
| Idle transitive clients | ~320 | 0 |
| Real hosts/workers/clients visible | yes | yes |

Inference routing to real hosts above the floor was verified working end-to-end via `/v1/models` and `/v1/chat/completions`.

## What is intentionally **not** in this PR

- **Inbound per-frame cap** — discussed and rejected for now. Looks defensive but doesn't actually bound the attack (an attacker just opens more connections), and risks breaking propagation in healthy meshes. Needs measured attack-pattern data before designing.
- **Owner signing / `TrustPolicy` changes** — separately useful long-term but adoption is 0% today; not a short-term defense.
- **`NODE_PROTOCOL_GENERATION` bump** — that lever exists in `protocol/mod.rs`; it's the right tool for **intentional** wire-compat breakage, not for spam defense.

## Test coverage

20/20 `mesh::gossip::tests` pass, including 8 new tests:
- `version_allowed_for_rebroadcast_handles_floor`
- `version_allowed_for_rebroadcast_handles_metadata_and_prerelease`
- `version_allowed_for_rebroadcast_is_conservative_on_unknown`
- `transitive_ingest_rejects_below_version_floor`
- `direct_add_peer_rejects_below_version_floor`
- `peer_is_idle_transitive_client_basic_shapes`
- `transitive_ingest_drops_idle_clients_but_keeps_clients_with_demand`
- `direct_add_peer_admits_idle_clients`

## Live test plan

- ✅ `client --auto` joins public mesh — peer table contains only real ≥v0.60 peers
- ✅ `serve --auto` joins, talks only to real ≥v0.60 peers, gossip events clean
- ✅ Real release-binary deployed and verified on a v0.65.1 Mac Studio host

## Architecture

- All four checks live in one file (`mesh/gossip.rs`). The version-floor ingest gates check `version_allowed_for_rebroadcast(ann.version.as_deref())` before any other work. The idle-transitive-client gate runs only inside `update_transitive_peer`, never `add_peer`. The outbound `collect_announcements` filter is now belt-and-braces: ingest gates already reject sub-floor peers, but if one slipped through (e.g. version mutated after ingest) we still exclude it from outbound gossip.
- Two new public-within-the-module helpers: `version_allowed_for_rebroadcast` and `peer_is_idle_transitive_client`. Both pure functions, both tested in isolation.

## Protocol

**Not a breaking change.** No wire format changes, no protobuf changes, no `NODE_PROTOCOL_GENERATION` bump. Older peers continue to gossip normally with newer peers; the only change is what each peer chooses to do with what they receive.

Mixed-version interop is preserved:
- A v0.65 peer running this change will still happily gossip with a v0.65 peer running `main`.
- A v0.57 peer trying to participate is simply not added to our local table; they remain in the broader mesh exactly as before from any other node's perspective.

## Includes

- AGENTS.md note on `--headless` vs `--log-format json` (recurring agent mistake observed during this work)
- Small clippy fix on the version-comparison structure (`absurd_extreme_comparisons` warning)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -214,6 +214,29 @@ When making changes that touch gossip, routing, proxy, election, or capability a
 
 Run `cargo` commands serially. Do not run multiple `cargo` commands in parallel (including parallel test runs), because this repo frequently hits Cargo lock conflicts (`package cache` / `artifact directory`) under concurrent invocation.
 
+## Running mesh-llm locally
+
+Default the launch to a normal foreground run (TUI visible) unless you have a
+specific reason to suppress UI surfaces. Most observation/debug tasks do not
+need the TUI suppressed.
+
+- `mesh-llm client --auto` — normal foreground run with the TUI. Use this by
+  default.
+- `--log-format json` — emits machine-parseable JSON log lines. Use this when
+  you want to programmatically read events.
+- `--headless` — disables the **embedded web UI**, not the TUI. The TUI still
+  draws. Only use `--headless` when you are intentionally avoiding the
+  management web console — it is **not** the way to get a quiet background run.
+- `--no-console` — fully disables the management console (HTTP API on the
+  console port).
+- `nohup … &` with a foreground binary that draws a TUI will appear to run but
+  often exits or behaves oddly when the TUI cannot attach to a terminal. Prefer
+  letting the developer launch the binary in their own terminal and observing
+  via `/api/status`, `--log-format json`, or by reading stderr.
+
+Do not reach for `--headless` to "go quiet" — that is a recurring mistake. If
+you want quiet output, use `--log-format json` and parse what you need.
+
 ## Pre-Commit Checklist
 
 Before committing, run the local checks most likely to fail in CI for the files you touched. Do not rely on CI to catch basic formatting, compile, or stale UI build issues.
--- a/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
@@ -49,11 +49,12 @@ pub(super) fn version_allowed_for_rebroadcast(version: Option<&str>) -> bool {
     let Some(minor) = parts.next().and_then(|s| s.parse::<u64>().ok()) else {
         return true;
     };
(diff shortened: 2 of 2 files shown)


ndizazzo:
I'm not sure this will entirely fix it... and I think we will still have the problem. From the analysis doc I generated, it's only a few "bridge" nodes that connect to the mesh and advertise all of the peers flooding the network. 

We have to stop the bridge nodes that are broadcasting mesh connectivity to non-existent peers

ndizazzo (approved):

PR #577, opened 2026-05-18
Route around recently unhealthy inference targets

IvGolovach:
## Summary

This adds outcome-aware routing cooldowns for inference targets. When a target recently times out, fails as unavailable, or returns a server-side failure, the local proxy temporarily routes around it when another eligible target exists.

- Adds a bounded local target-health tracker keyed by model and full inference target.
- Applies cooldown filtering after context-capacity ordering and before prefix, sticky, or round-robin target selection.
- Records routing outcomes from both explicit model routing and mesh remote-host routing.
- Keeps context-overflow responses, ordinary 4xx rejections, and client disconnects out of target cooldowns.
- Preserves availability by falling back to the original candidate order when every candidate is currently cooling.

## Why

Sticky routing and prefix affinity are useful when a target is healthy, but they can keep sending follow-up traffic to a target that just failed. This keeps the existing affinity behavior while giving the router a short local memory of unhealthy outcomes, so a transiently bad target does not keep winning while healthy alternatives are available.

## Design

The cooldown state is intentionally local-only. It is not gossiped, serialized into the mesh protocol, or exposed as a new public API contract.

- `Timeout`, tunnel/unavailable failures, and delivered 5xx responses cool the specific `(model, target)` pair.
- A successful attempt clears the cooldown for that pair.
- Context overflow remains a request-shape/context-capacity signal, not a peer-health signal.
- Cooldowns are bounded, expire automatically, and are scoped per model and target.
- If all candidates are cooling, routing still proceeds with the original candidates instead of failing closed.

## Compatibility

No mesh gossip, protobuf, plugin protocol, or public API schema changes are introduced. Mixed-version meshes continue to interoperate because this only affects local target ordering inside the current process.

## Validation

- `git diff --check`
- `git diff --cached --check`
- `LLAMA_STAGE_BUILD_DIR=<local llama-stage ABI build dir> cargo test -p mesh-llm-host-runtime target_health --lib`: PASS, 8 passed
- `LLAMA_STAGE_BUILD_DIR=<local llama-stage ABI build dir> cargo test -p mesh-llm-host-runtime network::affinity --lib`: PASS, 14 passed
- `LLAMA_STAGE_BUILD_DIR=<local llama-stage ABI build dir> cargo check -p mesh-llm-host-runtime`: PASS

Not run: `just build` - not required for this narrow runtime routing change; no UI or build artifact changed.

## Rollback

Revert this PR.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/affinity.rs
+++ b/crates/mesh-llm-host-runtime/src/network/affinity.rs
@@ -1,6 +1,7 @@
 //! Prefix affinity and sticky routing helpers for inference target selection.
 
 use crate::inference::election;
+use crate::network::target_health::{TargetHealth, TargetHealthOutcome};
 use iroh::EndpointId;
 use serde::Serialize;
 use serde_json::Value;
@@ -87,6 +88,7 @@ struct AffinityState {
 pub struct AffinityRouter {
     inner: Arc<Mutex<AffinityState>>,
     config: Arc<AffinityConfig>,
+    target_health: TargetHealth,
 }
 
 impl AffinityRouter {
@@ -94,6 +96,7 @@ impl AffinityRouter {
         Self {
             inner: Arc::new(Mutex::new(AffinityState::default())),
             config: Arc::new(AffinityConfig::from_env()),
+            target_health: TargetHealth::default(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


michaelneale:
I like this - and need something like this!

michaelneale (approved):
nice one @IvGolovach

PR #580, opened 2026-05-18
Sampler parity (min_p) + bounded max_tokens default + KV reclaim hardening

michaelneale:
edit: https://github.com/Mesh-LLM/mesh-llm/pull/581 - has the new one. 

Three independent fixes for the wedge we hit on the studio running MiniMax-M2.5.

## min_p sampler parity (`b41a7022`)

Upstream llama-server's default sampler chain includes `min_p` between `top_p` and `temperature` (`common.h:245`, default `0.05`). Our skippy sampler chain skipped it entirely. The Rust FFI struct already carried a `min_p` field; the C side was reading those four bytes as `uint32_t reserved` and discarding them.

Added `llama_sampler_init_min_p` to both `skippy_build_sampling_chain` and `skippy_sample_token` in the right chain position, and swapped `reserved` for `float min_p` in `skippy_sampling_config`. ABI bumped 23→24. Wire layout unchanged; only the field interpretation changes.

Carried as `third_party/llama.cpp/patches/0081-Add-min-p-sampler-to-skippy-sampling-chain.patch`. The other 80 patches show as modified because `git format-patch` regenerates mail headers when the queue length changes; semantic content is unchanged and validated by re-applying to a fresh upstream checkout.

## Bounded default max_tokens (`28e142b6`)

The embedded mesh-llm wiring passed `CONTEXT_BUDGET_MAX_TOKENS = u32::MAX` as `default_max_tokens`. When a client omitted `max_tokens`, the server gave the request the entire remaining context window — on MiniMax that's ~262k tokens, hours of generation on Studio hardware.

Added `DEFAULT_EMBEDDED_MAX_TOKENS = 8192` and switched the three host-runtime callsites (`runtime/local.rs` ×2, `inference/skippy/mod.rs` builder default) to use it. `CONTEXT_BUDGET_MAX_TOKENS` is preserved as an opt-in sentinel.

The standalone `serve-openai` CLI already had a sane explicit default and is untouched.

## KV-slot reclaim hardening (`ab933b1d`)

`RuntimeState::drop_session_timed` previously returned `Err` via `?` when a lane's `reset()` failed. That left `session_token_counts` and `session_checkpoints` holding stale entries for the now-removed `session_id`, and dropped the lane on the floor with no log record. Every caller swallows the error (`if let Ok(...)` / `let _ = ...`), so the failure was invisible. Accumulating bookkeeping drift is consistent with the post-incident `failed to find a memory slot` symptom on Studio.

Rewrote so the function always returns `Ok`, always clears per-session bookkeeping, and on reset failure explicitly drops the lane (which fires `StageSession::drop` → `skippy_session_free` — the authoritative native KV release path). Added `lane_discarded: bool` and `lane_discard_reason: Option<String>` to `RuntimeSessionDropStats` for telemetry visibility.

## Validation

Local end-to-end test against Qwen2.5-0.5B with a process that was the same shape as the Studio incident:

| Test | Result |
|---|---|
| Coherent English, no max_tokens | `finish_reason=stop` at 219 tokens via normal EOS |
| **Adversarial Russian-drift prompt, no max_tokens** | **`finish_reason=length` at exactly 8192 tokens** — the cap fires |
| 10 back-to-back small requests | All clean, no `lane_discarded`, no `failed to find a memory slot` |

Standard cargo validation:

```
cargo fmt --all -- --check
cargo check -p skippy-server -p mesh-llm-host-runtime
cargo test -p skippy-server --lib    # 77 pass
cargo test -p skippy-runtime --lib   # 33 pass
just build                            # llama.cpp + workspace
```

## What's still open

- min_p stopping the language flip itself (vs just bounding it) needs verification on MiniMax specifically. The 0.5B model still flipped on the loaded prompt — either expected (the Russian tokens were genuinely top-1% given the prompt) or min_p is less effective on small models. Either way, this PR brings the chain to upstream parity and bounds the damage window.
- The reset-failure path of the KV reclaim fix is correct by inspection but isn't exercised by happy-path tests. Requires a real wedge reproduction to exercise the new `lane_discarded` path.
- Studio deploy + soak test on MiniMax is the next step after merge.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/inference/skippy/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/inference/skippy/mod.rs
@@ -28,7 +28,7 @@ use skippy_runtime::ModelInfo;
 use skippy_server::{
     binary_transport::WireCondition, embedded_openai_backend, telemetry::Telemetry,
     telemetry::TelemetryLevel, EmbeddedOpenAiArgs, EmbeddedRuntimeOptions, EmbeddedRuntimeStatus,
-    EmbeddedServerHandle, EmbeddedState, SkippyRuntimeHandle, CONTEXT_BUDGET_MAX_TOKENS,
+    EmbeddedServerHandle, EmbeddedState, SkippyRuntimeHandle, DEFAULT_EMBEDDED_MAX_TOKENS,
 };
 
 pub(crate) use certification::{
@@ -177,7 +177,7 @@ impl SkippyModelLoadOptions {
             n_ubatch: None,
             flash_attn_type: FlashAttentionType::Auto,
             generation_concurrency: 1,
-            default_max_tokens: CONTEXT_BUDGET_MAX_TOKENS,
+            default_max_tokens: DEFAULT_EMBEDDED_MAX_TOKENS,
             layer_end: None,
             selected_device: None,
             package_identity: None,
--- a/crates/mesh-llm-host-runtime/src/runtime/local.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/local.rs
@@ -1128,7 +1128,7 @@ async fn load_split_runtime_generation_inner(
             config,
             activation_width,
             slots,
-            skippy_server::CONTEXT_BUDGET_MAX_TOKENS,
+            skippy_server::DEFAULT_EMBEDDED_MAX_TOKENS,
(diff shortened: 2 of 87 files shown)


michaelneale:
Superseded by the clean version that doesn't carry 80 cosmetic mail-format-only patch file diffs. See the new PR for the same three commits with a sensible diff shape.

PR #581, opened 2026-05-18
Sampler min-p parity, bounded default max_tokens, and KV-slot reclaim hardening

michaelneale:
Independent fixes for the wedge incident on the Mac Studio MiniMax host, plus review-feedback commits.

## 1. min-p sampler in skippy chain

Upstream `llama-server`'s default sampler chain is `penalties → top_k → top_p → min_p → temp → dist`. Our skippy ABI chain was missing `min_p`. The Rust side has been parsing and sending `min_p=0.05` end-to-end the whole time; the C side declared the field as `uint32_t reserved` and discarded those 4 bytes.

Without `min_p`, low-probability tail tokens survive `top_p` at temperatures > 0 on multilingual models. This is consistent with the rare-token / language-drift behaviour observed during sustained generation on MiniMax M2.5.

Changes:
- `skippy.h`: `uint32_t reserved` → `float min_p` (offset 44..48). Wire layout unchanged.
- `skippy.cpp`: add `llama_sampler_init_min_p` to both sampler-chain builders, in the upstream position (between `top_p` and `temperature`).
- `SKIPPY_ABI_VERSION_PATCH` 23 → 24.

No Rust changes — the FFI struct already had `min_p` at the right offset.

## 2. Bounded default `max_tokens` for embedded serving

Embedded mesh-llm chat serving wired `default_max_tokens = u32::MAX` (`CONTEXT_BUDGET_MAX_TOKENS`). When a client omitted `max_tokens`, the generation loop ran until the entire remaining context window was consumed. On MiniMax M2.5 with a small prompt, that's ~260k tokens of permission — multiple hours per request on Studio hardware.

Introduces `DEFAULT_EMBEDDED_MAX_TOKENS` (defined in `crates/skippy-server/src/frontend.rs`) and wires it through three host-runtime call sites. Whatever value the constant carries is the bounded fallback when a client omits `max_tokens`. `CONTEXT_BUDGET_MAX_TOKENS` remains exported as the opt-in sentinel for callers who genuinely want full-context completion.

## 3. KV-slot reclaim hardening

`RuntimeState::drop_session_timed` is the cleanup path called at the end of every chat completion. It used to propagate `Err` via `?` if the underlying StageSession failed to reset. That had three consequences:

- The lane was dropped on the floor instead of returned to the idle pool — no record anywhere, only released via `StageSession::drop → skippy_session_free` on stack unwind.
- `session_token_counts` and `session_checkpoints` for the now-removed session_id leaked permanently (slow bookkeeping drift).
- Every caller swallows the error (`if let Ok(...) = ...` or `let _ = ...`), so the failure never surfaced to telemetry.

This is consistent with the post-incident symptom on Studio where new requests hit `decode: failed to find a memory slot for batch of size 1` after the wedge cleared.

Changes:
- `drop_session_timed` is now self-consistent and always returns `Ok`. On reset failure: log to stderr, discard the lane explicitly (its `Drop` calls the authoritative C-side free), continue full bookkeeping cleanup.
- `session_token_counts` and `session_checkpoints` are always cleared for the session_id, regardless of reset outcome.
- `RuntimeSessionDropStats` gains `lane_discarded: bool` and `lane_discard_reason: Option<String>` so telemetry can distinguish clean reset from forced discard. No caller pattern-matches the struct exhaustively (verified).

## Local validation

Ran four scenarios against a local Qwen2.5-0.5B serving with the rebuilt binary:

| Test | Result |
|---|---|
| Plain English request, no `max_tokens` | `finish_reason=stop`, 219 tokens, coherent — sampler chain works |
| Adversarial Russian-drift prompt, no `max_tokens` | Model wandered into Russian and looped; **stopped at `DEFAULT_EMBEDDED_MAX_TOKENS` with `finish_reason=length`**. Pre-fix this runs to context exhaustion. |
| 10 back-to-back small requests | All `finish_reason=stop`, zero `lane_discarded` events, zero `failed to find a memory slot` errors |

End-to-end MiniMax verification on Studio is the next step.

## Build / test

- `cargo fmt --all -- --check` clean
- `cargo check -p skippy-server -p mesh-llm-host-runtime` clean, no warnings
- `cargo test -p skippy-server --lib` 81 pass
- `cargo test -p skippy-runtime --lib` 33 pass
- `just build` rebuilds llama.cpp static libs and Rust workspace cleanly
- `scripts/prepare-llama.sh pinned` applies all 81 patches cleanly to a fresh upstream checkout at `389ff61d`

## Replaces

#580 — same content, accidentally committed with a regenerated-mail-format patch queue (80 extra cosmetic file diffs). This PR has the clean diff shape (7 files).


crates/skippy-server/src/runtime_state.rs
```diff
@@ -60,6 +60,13 @@ pub struct RuntimeSessionDropStats {
     pub reset_session: bool,
     pub reset_ms: f64,
     pub preserved_resident_prefix: bool,
+    /// True when the lane could not be returned to the idle pool because
+    /// the underlying StageSession failed to reset cleanly. The lane is
+    /// dropped (which invokes the C-side skippy_session_free) and the
+    /// pool capacity is restored on the next prewarm/admission cycle.
+    pub lane_discarded: bool,
+    /// Reset-error detail, when [`Self::lane_discarded`] is true.
+    pub lane_discard_reason: Option<String>,
```
ndizazzo, line 83: I see `lane_discarded` and `lane_discard_reason` described as telemetry fields, but none of the `*_session_stop` call sites actually emit them on the span... they only are used in the`eprintln!`. 

Fine for now, just flagging in case the intent was OTLP visibility.

ndizazzo (dismissed):
One thing I noticed during testing using the TUI, was that when I had my agent use a slot - I could see the slot go active and it never returned back to a "free" state... I was thinking it was a TUI bug, but then I asked my agent why, and it found a legitimate issue:


> There's a latent wedge condition introduced by this PR's discard path. RuntimeState uses next_lane_index (line 23) as a monotonically incrementing counter to assign C-side session slots. create_indexed_lane_resource increments it on every lane creation and bails with "all execution lanes are busy" once it reaches lane_count — it never decrements.
>
> When a lane is discarded here, sessions.remove() pulls it from the active map and the lane_session drop frees the C-side memory, but next_lane_index stays at its current value. The result: sessions.len() < lane_count (capacity appears available), but the next call to create_lane_session() bails immediately because the index counter is exhausted. The runtime is wedged.
>
> Concrete scenario with lane_count=2: both lanes allocated (next_lane_index=2), one session fails to reset and gets discarded. Now sessions.len()=1, but any new session request falls through take_idle_session() (nothing in the pool), checks sessions.len() < lane_count (passes), calls create_lane_session(), and hits bail!("all execution lanes are busy") — permanently, for the rest of that RuntimeState's lifetime.
>
> This existed in the pre-PR code too (a reset failure also silently dropped the lane), but returning Ok here makes the condition reachable without any error signal to the caller, and the docstring's claim that "pool capacity is restored on the next prewarm/admission cycle" describes what should happen rather than what does. The fix is to either decrement next_lane_index when the discarded lane is the most-recently-allocated slot, or maintain a free-list of reclaimable indices for non-tail discards.

@michaelneale this is probably contributing to why it was getting wedged

crates/skippy-server/src/frontend.rs
ndizazzo: `min(DEFAULT_EMBEDDED_MAX_TOKENS, ctx_size - prompt_tokens)` would be safer when we check for this value, on any model with `ctx_size <= 8192`, the check can fail

michaelneale: 👍

crates/skippy-server/src/runtime_state.rs
```diff
@@ -296,38 +303,109 @@ impl RuntimeState {
...
+                    }
+                }
             }
         }
+
+        // Always clear per-session bookkeeping. The previous version
+        // skipped these when reset returned Err, which leaked entries.
         self.session_token_counts.remove(session_id);
         self.session_checkpoints.remove(session_id);
+        // session_resident_prefixes was already removed above (or absent).
```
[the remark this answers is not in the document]
michaelneale: Already addressed in commit `e271b27c` (`skippy-server: always clear session_resident_prefixes in drop_session_timed`). `session_resident_prefixes.remove(session_id)` is now in the unconditional cleanup block at the end of the function alongside `session_token_counts` and `session_checkpoints`, so idempotent / stale-id calls also clear it.

crates/skippy-server/src/runtime_state.rs
```diff
@@ -296,38 +303,109 @@ impl RuntimeState {
...
+                        self.idle_sessions.push(lane_session);
+                    }
+                    Err(reset_err) => {
+                        lane_discarded = true;
+                        let reason = format!("reset() failed ({reset_err:#})");
+                        eprintln!(
+                            "skippy::runtime_state: drop_session_timed: \
+                             discarding lane {lane_index} for session {session_id}: \
+                             {reason}"
+                        );
```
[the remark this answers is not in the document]
michaelneale: Fixed in commit `723f1cc5` (same change as the comment above).

crates/skippy-server/src/runtime_state.rs
```diff
@@ -296,38 +303,109 @@ impl RuntimeState {
...
+                                lane_discarded = true;
+                                let reason = format!(
+                                    "trim_session({}) failed ({trim_err:#}); fallback reset() also failed ({reset_err:#})",
+                                    prefix.token_count
+                                );
+                                eprintln!(
+                                    "skippy::runtime_state: drop_session_timed: \
+                                     discarding lane {lane_index} for session {session_id}: \
+                                     {reason}"
+                                );
```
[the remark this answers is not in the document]
michaelneale: Fixed in commit `723f1cc5`. Each format string is now a single source line — no `\` line continuation, no embedded indentation in the emitted log message.

crates/skippy-server/src/runtime_state.rs
michaelneale: Good catch — fixed in commit `fe4e7039`. Both `lane_discarded` (bool, always emitted) and `lane_discard_reason` (string, only when discarded) are now inserted into the attrs map at all 5 emit sites:

- `frontend/local_generation.rs:557` (`stage.openai_session_stop`)
- `frontend/generation_flow.rs:450` and `:861` (both `stage.openai_session_stop`)
- `frontend/embedded_generation.rs:1215` (`stage.openai_session_stop`)
- `binary_transport.rs:501` (`stage.binary_session_stop`)

The two other `drop_session_timed` callsites discard `drop_stats` via `let _ = ...` so they don't emit a span and needed no change.

ndizazzo (approved):

PR #586, opened 2026-05-19
Rescue audio requests through Virtual LLM peer handoff

IvGolovach:
## Summary

Text-only Skippy/Virtual LLM requests with audio can now ask an audio-capable mesh peer for concise audio context, inject that context into the local request, and then continue locally as a real text-only request.

This also closes the reviewed case where the original handled `input_audio` / `audio_url` part survived after successful rescue. The handled media part is consumed before Skippy prompt preparation, so text-only Skippy no longer sees residual `prompt.media` and no longer falls into the multimodal/projector path after rescue.

## Why

The existing Virtual LLM path already recognized `audio_no_support` and could find an audio-capable peer, but the actual audio payload was not fully wired through. After the first implementation, i386 pointed out one remaining correctness gap: successful rescue injected `[Audio context: ...]`, but left the original audio content block in the forwarded request. Skippy prompt preparation still treated that as media and could reject the request with `multimodal request requires a configured projector`.

The fix keeps the rescue semantics explicit: successful pre-inference media rescue means "consume this exact handled media part, then inject the text context." If rescue does not succeed, media remains untouched. If there are multiple media parts, only the exact handled part is removed and unhandled media still routes through the normal multimodal path.

## Diff scope

- Extracts audio from `audio_url`, `input_audio.url`, inline `input_audio.data` + `format`, and the legacy `mesh_audio_url` preservation path.
- Routes `audio_no_support` through Virtual LLM peer consultation and injects `[Audio context: ...]` back into the prompt.
- Adds location-aware `ChatMediaRef` fields and a shared `ChatHookAction::ConsumeMedia` action for successful media rescue.
- Uses the shared hook outcome application in both the generic OpenAI hook backend and Skippy backend paths.
- Preserves peer consultation loop safety by sending the audio rescue request with `mesh_hooks: false`.
- Adds regression coverage proving rescued audio becomes text-only before Skippy prompt/media extraction, while unrelated second media remains visible.
- Updates the Virtual LLM design notes.

## Compatibility

- No mesh gossip, QUIC ALPN, protobuf, plugin protocol, Skippy ABI, or public API compatibility change.
- Audio support remains capability-gated by existing runtime capability advertisement.
- Existing image rescue behavior is preserved, with the same handled-media consumption semantics when rescue succeeds.
- Audio payloads are not logged; logs only record routing/action context.

## Branch integrity

- Base branch: `main`
- Validated base: `origin/main@e9a920b3079e026ebaa0a934d51b3691c78fe9af`
- Branch state after rebase: `0 behind / 1 ahead`
- Merge base: `e9a920b3079e026ebaa0a934d51b3691c78fe9af`
- Head commit: `6921ee175f0c09a01803306954ae3867d228f444`

## Commit integrity

Introduced commit:

- `6921ee175f0c09a01803306954ae3867d228f444` - `Add Virtual LLM audio rescue`

This is one logical PR commit. The final diff is limited to the intended OpenAI hook, Virtual LLM consultation, Skippy prompt-prep, and design-note files.

## Diff hygiene

- `git diff --name-status origin/main...HEAD`: PASS, 9 intended files.
- Forbidden files: none observed.

## Validation

Validation tier: Tier 3 - shared OpenAI/Skippy runtime hook path for Virtual LLM media rescue, refreshed on current main; no protocol/schema or release metadata change.

Local validation:

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, `origin/main` at `e9a920b3079e026ebaa0a934d51b3691c78fe9af`.
- `cargo test -p openai-frontend hooks --lib`: PASS, 10 passed.
- `cargo test -p openai-frontend --lib`: PASS, 154 passed.
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build-dir> cargo test -p mesh-llm-host-runtime inference::skippy::hooks::tests --lib -- --test-threads=1`: PASS, 15 passed.
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build-dir> cargo test -p mesh-llm-host-runtime inference::consult::tests --lib -- --test-threads=1`: PASS, 6 passed.
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build-dir> cargo test -p mesh-llm-host-runtime inference::virtual_llm::tests --lib -- --test-threads=1`: PASS, 3 passed.
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build-dir> cargo test -p skippy-server rescued --lib`: PASS, 2 passed.
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build-dir> cargo test -p skippy-server --lib`: PASS, 106 passed.
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build-dir> cargo check -p mesh-llm`: PASS.
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build-dir> cargo-clippy clippy -p openai-frontend -p skippy-server -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS.

Remote gates on head `6921ee175f0c09a01803306954ae3867d228f444`:

- `PR Quality Checks`: PASS.
- `PR Builds`: PASS.
- Required build/test/smoke jobs passed, including Linux/macOS CPU, Linux unit/protocol/skippy/sdk-api/skippy-smoke, OpenAI inference smoke, client-auto boot, and two-node scripted smokes.
- Conditional jobs skipped as expected: matrix GPU builds, optional SDK smoke jobs, `agent_live_smokes`, and `ui-quality`.

Version: not applicable - no release/version sync required for this non-release runtime hook fix.

Not run:

- `just build` - not required for selected validation tier; no UI assets or release bundle changed.
- Live two-node audio-capable peer smoke - no local audio-capable mesh peer endpoint was available; deterministic OpenAI hook, host-runtime policy, and Skippy prompt-prep tests cover the reviewed failure path.

## Runtime safety

No protobuf, gossip schema, Skippy ABI, plugin protocol, workflow, or release metadata change.

The change only mutates the current OpenAI request body after a successful media rescue, and only for the exact handled media part. Unhandled media remains visible to the normal multimodal path.

## Rollback

Revert this PR.

DB downgrade: not applicable.
Data repair: not applicable.

## Known residual risks

A live two-node smoke with a real audio-capable peer would still be useful reviewer confidence, but the reviewed failure mode is covered locally and remotely: after successful rescue, the handled audio media part is removed before Skippy prompt/media extraction.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/api/routes/mesh_hook.rs
+++ b/crates/mesh-llm-host-runtime/src/api/routes/mesh_hook.rs
@@ -68,11 +68,7 @@ async fn dispatch_hook(state: &MeshApi, payload: Value) -> Value {
     let messages: Vec<Value> = payload["messages"].as_array().cloned().unwrap_or_default();
 
     match hook {
-        "pre_inference" => {
-            let trigger = payload["trigger"].as_str().unwrap_or("unknown");
-            let (image_url, user_text) = virtual_llm::extract_image(&payload);
-            virtual_llm::handle_image(&node, trigger, &model, &image_url, &user_text).await
-        }
+        "pre_inference" => dispatch_pre_inference(&node, &payload, &model).await,
         "post_prefill" => {
             let entropy = payload["signals"]["first_token_entropy"]
                 .as_f64()
@@ -92,5 +88,19 @@ async fn dispatch_hook(state: &MeshApi, payload: Value) -> Value {
             tracing::warn!("mesh hook: unknown hook type: {hook}");
             serde_json::json!({ "action": "none" })
         }
+    }
+}
+
+async fn dispatch_pre_inference(node: &crate::mesh::Node, payload: &Value, model: &str) -> Value {
+    let trigger = payload["trigger"].as_str().unwrap_or("unknown");
+    let (media_url, user_text) = pre_inference_media(payload, trigger);
+    virtual_llm::handle_image(node, trigger, model, &media_url, &user_text).await
+}
+
+fn pre_inference_media(payload: &Value, trigger: &str) -> (String, String) {
+    if trigger == "audio_no_support" {
+        virtual_llm::extract_audio(payload)
+    } else {
+        virtual_llm::extract_image(payload)
     }
 }
--- a/crates/mesh-llm-host-runtime/src/inference/consult.rs
+++ b/crates/mesh-llm-host-runtime/src/inference/consult.rs
@@ -8,6 +8,7 @@
 //! Three consultation patterns:
 //!
 //! - **Caption** — send an image to a vision-capable peer, get a text description
+//! - **Audio rescue** — send audio to an audio-capable peer, get concise text context
 //! - **Summarize** — send conversation history, get a condensed summary
 //! - **Second opinion** — send the same question to a different model, get its answer
 
@@ -249,6 +250,46 @@ pub async fn caption_image(
     chat_completion(node, peer_id, model, messages, 256, TIMEOUT_CONSULTATION).await
 }
 
+/// Ask an audio-capable peer to extract useful text context from audio.
+/// `audio_url` should be a URL or a data URL accepted by the peer's OpenAI
+/// chat surface.
+pub async fn transcribe_audio(
+    node: &mesh::Node,
+    peer_id: EndpointId,
+    model: &str,
+    audio_url: &str,
+    user_text: &str,
+) -> Result<String> {
+    chat_completion(
+        node,
+        peer_id,
+        model,
+        audio_rescue_messages(audio_url, user_text),
+        512,
+        TIMEOUT_CONSULTATION,
+    )
+    .await
+}
+
+fn audio_rescue_messages(audio_url: &str, user_text: &str) -> Vec<Value> {
+    let prompt = if user_text.is_empty() {
+        "Extract concise text context from this audio. If it contains speech, transcribe the speech. If it contains non-speech audio, describe the audible events. Return only the useful context."
+            .to_string()
+    } else {
+        format!(
+            "The user asked: \"{user_text}\"\n\nExtract concise text context from this audio for a text-only model. If it contains speech, transcribe the relevant speech. If it contains non-speech audio, describe the audible events relevant to the user's request. Return only the useful context."
+        )
+    };
+
+    vec![serde_json::json!({
+        "role": "user",
+        "content": [
+            {"type": "text", "text": prompt},
+            {"type": "input_audio", "input_audio": {"url": audio_url}}
+        ]
+    })]
+}
+
 /// Ask a peer for a second opinion on the user's question.
 ///
 /// Sends only the last user message (not the full conversation) and asks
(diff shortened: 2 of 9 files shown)


michaelneale:
@IvGolovach how have you found the hooks performing - were you able to try this with an audio model?

IvGolovach:
Thanks, good question - that’s exactly the smoke test this should get next.

I took a careful pass through the hook path rather than treating this as just payload parsing: text-only target detects unsupported audio, `pre_inference` fires `audio_no_support`, Virtual LLM extracts the original audio part, consults an audio-capable peer with `mesh_hooks: false`, and injects concise `[Audio context: ...]` back into the prompt.

I added the coverage around those boundaries because the main risks here were losing the audio payload or accidentally making the hook recurse.

I haven’t run a live two-node audio-model smoke yet, so I don’t want to overclaim that part. Locally/contract-wise the path stays cheap unless hooks are enabled and the selected model can’t handle audio. If you have a preferred small audio model, I’m happy to use that; otherwise I’ll pick one and report timings / failure shape.

michaelneale:
yeah don't have an audio one - only audio things I Have tried were whisper or whisper-fast voice to test non LLM models.

i386:
Blocking issue I found while reviewing:

The audio rescue path still leaves the original audio part in the local Skippy request after the hook injects `[Audio context: ...]`. `first_chat_media` now recognizes `input_audio`/`audio_url`, so `MeshAutoHookPolicy::before_chat_completion` can consult an audio-capable peer and prepend text. But the hook outcome only injects text; it does not remove or neutralize the original media part.

As a result, prompt preparation still sees `prompt.media`, routes into `generate_multimodal_text`, and text-only Skippy runtimes can still fail with `multimodal request requires a configured projector` after the peer handoff succeeds.

Relevant spots:
- `crates/openai-frontend/src/hooks.rs`: `media_url` now recognizes `input_audio` and builds a media ref.
- `crates/mesh-llm-host-runtime/src/inference/skippy/hooks.rs`: `before_chat_completion` turns that media ref into an audio rescue hook result.
- `crates/skippy-server/src/frontend/generation_flow.rs`: any remaining `prompt.media` routes to multimodal generation and rejects runtimes without a projector.

I think the fix needs to strip or replace handled media after a successful pre-inference rescue, or otherwise mark the request as text-only before prompt preparation.

IvGolovach:
Thanks for the careful review here. I rebased this onto current main and tightened the rescue path around the James note.

The important fix is that a successful pre-inference audio/image rescue now consumes the exact handled media part before the request reaches normal Skippy prompt preparation. That keeps the rescued request text-only for the local model path, while leaving unrelated or unhandled media parts intact so they still surface normally instead of being silently dropped.

I also moved the hook mutation into the shared OpenAI frontend hook outcome path, so the proxy and Skippy frontend apply the same semantics, and added regressions for both sides: forwarding consumes the rescued media, matching is location-safe, and Skippy no longer sees the already-rescued media as prompt.media.

Validation is green locally and remotely on the updated head. Thanks again for catching this edge case — it was the right thing to fix before landing. 🙂

PR #587, opened 2026-05-19
fix(ci): fix CI PR cleanup job to delete in batches

ndizazzo:
Job from @IvGolovach [failed on cleanup](https://github.com/Mesh-LLM/mesh-llm/actions/runs/26075446176/job/76665402394) due to rate limits, this should address it


michaelneale:
Is the swift failure related to this?

ndizazzo:
> Is the swift failure related to this?

Nah - it's because an earlier step didn't produce an artifact to do smoke tests with. I'll throw a fix for that in with this

PR #589, opened 2026-05-19
feature(mesh-genesis): add mesh genesis rules

ndizazzo:
## Summary

This PR lets operators create **requirement-aware meshes**: meshes whose admission rules are fixed at creation time and enforced before peers become routable members.

Operators can require trusted embedded release-build attestations, enforce node version floors or ceilings, require compatible protocol generations, and issue signed bootstrap tokens for restricted meshes. Legacy unrestricted meshes keep the existing unsigned invite-token flow.

Release attestation is now embedded in the packaged `mesh-llm` executable itself. There is no sibling `.attestation.json` file and no runtime sidecar path. Nodes inspect the binary they are actually running, advertise the verified release-attestation result, and remain non-blocking by default unless mesh policy requires certified builds.

Certified-build admission proves release provenance. It is not remote runtime attestation and does not prove the remote OS, hardware, or running process is untampered.

### 1. Require trusted release-attested binaries before a node can join

```bash
mesh-llm serve --model Qwen3-8B-Q4_K_M --publish \
  --require-release-attestation \
  --release-signer-key ed25519:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef \
  --owner-key ~/.mesh-llm/owner-keystore.json \
  --owner-required
```

A joining node presents the embedded attestation from its packaged executable:

```bash
mesh-llm serve --join <signed-bootstrap-token>
```

Operators can inspect the same packaged binary directly:

```bash
cargo run -p xtask -- release-attestation inspect \
  --binary /tmp/test-bundle/mesh-llm \
  --public-key-file /secure/mesh-release-key.pub \
  --json
```

Inspection reports `valid`, `missing`, or `invalid`. Unstamped local/dev builds report `missing`; trusted stamped packages report `valid`; bytes changed after packaging report `invalid`.

Admission failures use stable reasons such as:

```text
certified_binary_required
build_proof_missing
build_proof_invalid
release_signer_untrusted
attestation_policy_mismatch
```

### 2. Stamp the final packaged release artifact

Release attestation is stamped after the package binary's final mutation, so the signed bytes match the artifact operators install and distribute.

```bash
MESH_RELEASE_ATTESTATION_SIGNING_KEY_FILE=/secure/mesh-release-key.json \
MESH_RELEASE_ATTESTATION_PUBLIC_KEY_FILE=/secure/mesh-release-key.pub \
  just release-bundle v0.65.0
```

Smoke and nightly evidence now inspect packaged artifacts rather than raw `target/release/mesh-llm` outputs.

### 3. Require a minimum mesh-llm version for all joining nodes

```bash
mesh-llm serve --model Qwen3-8B-Q4_K_M --publish \
  --min-node-version 0.65.0
```

Peers advertising older versions are rejected before routing:

```text
node_version_below_minimum
```

### 4. Cap or pin accepted node versions

```bash
mesh-llm serve --model Qwen3-8B-Q4_K_M --publish \
  --min-node-version 0.65.0 \
  --max-node-version 0.65.9
```

This lets operators stage upgrade windows without silently admitting incompatible future or older binaries.

### 5. Guarantee protocol-generation compatibility

```bash
mesh-llm serve --model Qwen3-8B-Q4_K_M --publish \
  --min-protocol-version 1 \
  --max-protocol-version 1
```

Peers outside the accepted range are rejected with stable reason codes:

```text
protocol_generation_below_minimum
protocol_generation_above_maximum
protocol_generation_unknown
```

### 6. Combine binary, version, and protocol requirements into one immutable mesh policy

```bash
mesh-llm serve --model Qwen3-8B-Q4_K_M --publish \
  --min-node-version 0.65.0 \
  --max-node-version 0.65.9 \
  --min-protocol-version 1 \
  --max-protocol-version 1 \
  --require-release-attestation \
  --release-signer-key ed25519:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef \
  --owner-key ~/.mesh-llm/owner-keystore.json \
  --owner-required
```

Changing those requirements creates a different requirement-aware mesh instead of mutating the existing mesh in place.

### 7. Configure the same requirements from `config.toml`

```toml
[mesh_requirements]
min_node_version = "0.65.0"
max_node_version = "0.65.9"
min_protocol_version = 1
max_protocol_version = 1
require_release_attestation = true
release_signer_keys = [
  "ed25519:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
]
```

Then start with an owner identity so the genesis policy and bootstrap token can be signed:

```bash
mesh-llm serve --model Qwen3-8B-Q4_K_M --publish \
  --owner-key ~/.mesh-llm/owner-keystore.json \
  --owner-required
```

### 8. Keep public discovery separate from admission

Discovery can advertise that a mesh exists, while admission still rejects nodes that do not satisfy its requirements.

```bash
mesh-llm serve --model Qwen3-8B-Q4_K_M --publish \
  --mesh-name certified-lab \
  --require-release-attestation \
  --release-signer-key ed25519:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef \
  --owner-key ~/.mesh-llm/owner-keystore.json \
  --owner-required
```

Runtime, API, and UI status surfaces now expose release-attestation state, mesh requirements, and recent mesh-rejection evidence.

## Compatibility

- Existing unrestricted private and public meshes keep the current unsigned invite-token behavior.
- Requirement-aware meshes use signed genesis policy, signed bootstrap tokens, and policy-matching direct admission proofs.
- Protocol additions are additive and preserve mixed-version compatibility for unrestricted meshes.
- Release attestation is optional by policy: `missing` or `invalid` binaries can still start unless the mesh requires certified builds.
- The embedded attestation applies only to the packaged `mesh-llm` executable, not SDK, XCFramework, or other native artifacts.
- Owner trust remains separate from release trust and is not part of the immutable mesh requirements hash.

## Validation

- Verified embedded release-attestation inspection for `valid`, `missing`, and `invalid` states.
- Confirmed packaged release flows stamp the final bundled executable after its last mutation.
- Confirmed smoke and nightly evidence inspect packaged artifacts rather than raw release outputs.
- Confirmed runtime, API, and UI surfaces expose release-attestation status.
- Confirmed requirement-aware admission preserves the legacy unrestricted mesh compatibility path.


michaelneale:
taking a look - may need this soon @ndizazzo

IvGolovach (commented):
Thanks for pushing this through. I did a careful pass over the current shape and the earlier review threads.

This is a big change, but the scope feels coherent: requirement-aware meshes get immutable genesis/admission policy, signed bootstrap material, version/protocol bounds, optional release-build attestation, and visible status/API/UI state, while unrestricted meshes keep the legacy open path.

The prior review threads look resolved, and the important edge cases around unsigned fallback, token validity, pre-join validation, optional attestation loading, protobuf layout, and release/test tooling all appear accounted for.

Given the protocol/security surface area, I’d just want the branch refreshed on latest main and the final CI/smoke evidence on that head. With that, I’m comfortable with the direction and think this is a strong foundation for restricted meshes.

michaelneale:
@ndizazzo could this be used to gate admission to a mesh to only blessed peers? (ie protection is bi directional?)

ndizazzo:
> could this be used to gate admission to a mesh to only blessed peers? (ie protection is bi directional?)

@michaelneale absolutely! This is actually the whole point of making this surface a bit more robust than simple if-checks. 

Permitting "allowed" peers doesn't exist right now on this PR, but the framework and structure is in place to make it very easy to extend with this rule. 

Right now, the PR has (all disabled by default):

* Minimum app version needed
* Maximum app version admitted
* Minimun protocol version needed
* Maximum protocol version admitted
* Embedded binary attestation -> Only nodes running a mesh-llm binary with a specifically embedded attestation is permitted
  * IE: We do a post-build process on CI to attach a HSM-signed attestation to our binary, then all mesh-llm nodes that require it must validate before being allowed on the mesh

It would be not much more work to:
* Have each node grant itself a specific identifier through the `mesh-llm auth` flow
* Add some services to maintain a list of auth keys that are permitted
* Only admit nodes that have signed proof they are the correct node

I'd like to get this in as-is before adding support for that, so this doesn't balloon too much. In the meantime, I'll draft up a spec for you to check out in the issues.

michaelneale:
I think this is important and good for maturity. 

Some thoughts (which could be follow ons): 

- Bind min/max_node_version to the signed attestation’s node_version, not the peer’s self-advertised gossip version.
- Fail startup when --require-release-attestation is set but the local binary is missing/invalid attestation
- is there a way to somehow enforce/check/prove the binary is itself signed/attested (depending on the OS, using whatever TEE is available?)

ndizazzo:
@michaelneale GitHub also has this: https://github.com/actions/attest-build-provenance

PR #592, opened 2026-05-19
build: ozempic — slim binary -42 MB / -47 MB

michaelneale:
Slims the shipped `mesh-llm` binary substantially and adds a feature flag so future lib-crate consumers can opt out of the embedded web console.

**macOS arm64 release binary:**

| Build | Size | Δ vs main |
|---|---|---|
| `main` (baseline) | 106 MB | — |
| `micn/ozempic` default | 64 MB | **-42 MB (-40%)** |
| `micn/ozempic --no-default-features` | 59 MB | **-47 MB (-44%)** |

The default-features build is what gets shipped — same user experience, just smaller. The `--no-default-features` build is for future lib consumers who want the runtime without the embedded UI assets.

## What changed

Three focused commits, easy to review independently:

### 1. UI: strip bundled ONNX WASM, lazy-load via CDN

The chat composer's "describe attached image" feature uses `@huggingface/transformers` (Transformers.js + Florence-2). That pulled `onnxruntime-web` into the bundle, which contains literals like

```js
new URL("ort-wasm-simd-threaded.jsep.wasm", import.meta.url)
```

Vite's asset emitter recognises that pattern and aggressively copied the **21 MB WASM** into `dist/assets/` — even though at runtime, Transformers.js defaults `wasmPaths` to a jsDelivr CDN URL and never touches the bundled copy.

A small Vite `transform` plugin rewrites the offending literals in `onnxruntime-web` modules so Rollup never emits the asset. Runtime behaviour is unchanged because ORT's `locateFile` callback still resolves the file from the CDN via `wasmPaths`.

`dist/` size: **26 MB → 5.7 MB** (-20.3 MB).

The Transformers.js JS module (~830 KB) stays dynamic-imported as before, so it's a separate chunk only loaded on first use. The Florence-2 model itself was already CDN-streamed from HuggingFace at runtime; this just applies the same policy to the runtime.

637 UI tests pass, no regressions.

### 2. Tighten release profile

Cargo's `[profile.release]` defaults are conservative: no LTO, 16 codegen units, no symbol stripping. Added:

- `lto = "thin"` — parallel link-time optimization across the crate graph. ~80% of fat-LTO's benefit at ~30% of the link cost.
- `codegen-units = 1` — pairs with thin LTO so the optimizer can inline across the whole crate.
- `strip = "debuginfo"` — drops DWARF from the Mach-O / ELF output. Function symbols stay intact so backtraces from production nodes remain readable.

`panic` stays at the default `unwind` per AGENTS.md (plugin / MCP error recovery relies on unwinding).

Release build time: ~30 s → ~2m45s on M4 Max. Dev builds (which `just build` uses) are unchanged.

### 3. `web-ui` Cargo feature, default on

Adds a default-on `web-ui` feature that propagates through:

```
mesh-llm (binary)
  → mesh-llm-host-runtime
     → mesh-llm-ui (embed-assets feature)
```

When on (default), the React console is embedded into the binary via `include_dir!` exactly as before. When off, `mesh-llm-ui::asset()` and `::index()` return `None`, the `include_dir!` invocation is elided, and the crate no longer depends on `include_dir`. The asset HTTP responders in `mesh-llm-host-runtime` route 404s for asset paths while every other management-API surface keeps working.

Intended for future lib-style consumers of `mesh-llm-host-runtime` who want the runtime/networking without the embedded web payload.

## Architecture

No mesh wire-protocol changes. No skippy ABI changes. No model-routing or election changes. This is purely build-time / packaging work.

Lib-crate prep: with `default-features = false`, `mesh-llm-host-runtime` and `mesh-llm-ui` now compile to a smaller artifact suitable for embedding in other binaries. Follow-up PRs can feature-gate other heavy optional deps (audit findings below) to push the lib-friendly variant smaller.

## Dep audit (not acted on — separate PRs if pursued)

Top opportunities ranked by size × likelihood of being optional:

| Crate | Size | Notes |
|---|---|---|
| rmcp | 1.1 MB | Plugin/MCP bridge. Could feature-gate. |
| aws_lc_sys | 671 KB | rustls TLS provider. Could swap for ring (smaller, simpler). |
| xet + hf-xet | ~940 KB | HF xet protocol for large model downloads. Could feature-gate. |
| mdns-sd | 297 KB | LAN discovery only. Could feature-gate. |
| opentelemetry stack | ~600 KB | Telemetry. Could feature-gate. |

Each requires careful surgery and is out of scope here.

## Validation

- `cargo check -p mesh-llm` ✓ (default features)
- `cargo check -p mesh-llm --no-default-features` ✓
- `cargo test -p mesh-llm-host-runtime --lib` → 1416 passed, 0 failed, 6 ignored
- `cargo test -p mesh-llm-ui --lib` → 2 passed (embed-assets) / 1 passed (stub, --no-default-features)
- `npm run typecheck` ✓
- `npm test` → 637 passed, 3 skipped (no regressions in image-describe call sites)
- `cargo build --release -p mesh-llm` → 64 MB binary
- `cargo build --release -p mesh-llm --no-default-features` → 59 MB binary
- Binary `--version` and `--help` smoke OK

## How to consume the lib variant later

```toml
[dependencies]
mesh-llm-host-runtime = { version = "...", default-features = false }
# or
mesh-llm = { version = "...", default-features = false }
```

That's the path forward for publishing `mesh-llm` as a lib crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,3 +67,24 @@ opt-level = 3
 
 [profile.dev.package.sha2-asm]
 opt-level = 3
+
+# Tightened release profile for shipped binaries.
+#
+# - `lto = "thin"`: parallel link-time optimization across the crate graph.
+#   Roughly 80% of fat-LTO's binary-size and perf benefit at ~30% of the
+#   link cost. Fat LTO would shave a few more MB but pushes release link
+#   from ~30 s to several minutes on this codebase.
+# - `codegen-units = 1`: pairs with thin LTO so the optimizer can inline
+#   across the whole crate without re-link boundaries. Trades parallelism
+#   in codegen for output quality; cargo still parallelises the dep graph.
+# - `strip = "debuginfo"`: drops DWARF from the Mach-O / ELF output. We
+#   keep function symbols so backtraces from production nodes remain
+#   readable (`nm`, `addr2line` on the symbol table still works). DWARF
+#   on macOS arm64 alone is several MB of dead weight in the binary.
+# - panic stays at the default ("unwind"). Plugin / MCP error recovery
+#   relies on unwinding; the AGENTS.md notes explicitly forbid switching
+#   to `abort`.
+[profile.release]
+lto = "thin"
+codegen-units = 1
+strip = "debuginfo"
--- a/crates/mesh-llm-host-runtime/Cargo.toml
+++ b/crates/mesh-llm-host-runtime/Cargo.toml
@@ -4,6 +4,11 @@ version = "0.65.1"
 edition = "2021"
 
 [features]
+default = ["web-ui"]
+# Embed the React web console into the binary. Default on; turn off for
+# lib-style / headless consumers that don't need the console assets.
+# Propagates to `mesh-llm-ui/embed-assets`.
+web-ui = ["mesh-llm-ui/embed-assets"]
 gpu-bench-cuda = ["mesh-llm-system/gpu-bench-cuda"]
 gpu-bench-hip = ["mesh-llm-system/gpu-bench-hip"]
 gpu-bench-intel = ["mesh-llm-system/gpu-bench-intel"]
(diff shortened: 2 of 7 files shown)


ndizazzo (approved):
@michaelneale A small improvement to the UI crate would be to return a specific error like: 

```rust
if !respond_console_index(stream).await? {
    respond_error(stream, 500, "Dashboard bundle missing").await?;
}
```

when in Ozempic mode - that way it's easy to debug. Otherwise, PR looks good.

PR #593, opened 2026-05-19
Harden Skippy binary frame allocation bounds

IvGolovach:
## Summary

Skippy binary stage transport now rejects malformed or oversized peer frames before allocating payload buffers.

- Adds explicit upper bounds for predicted-token replies, sideband counts, chat sampling metadata, state-import bytes, activation payload bytes, and decoded activation payloads.
- Validates outbound state-import and large metadata/sideband frames so local writers do not emit frames the reader would reject.
- Keeps the caps conservative for long-context prefix-cache restore and large ALTUP/RWKV activation frames.
- Adds regression tests that construct oversized wire frames and assert InvalidData before payload allocation.

## Why

The binary codec trusted peer-controlled counts and lengths before allocating Vec buffers. A malformed or incompatible peer could advertise a huge count and force a large reservation/allocation before the codec reached EOF or rejected the frame.

This is a local stage transport hardening change; it does not change the Skippy ALPN, protobuf stage-control generation, mesh gossip, or plugin protocols.

## Compatibility

The wire format is unchanged. Valid existing frames continue to decode. The new limits are intentionally above current long-context prefix-cache and large activation use cases, while rejecting clearly unreasonable allocation requests.

## Validation

- git fetch --no-tags origin main:refs/remotes/origin/main
- git diff --check
- git diff --cached --check
- cargo fmt --all -- --check
- cargo clippy -p skippy-protocol --all-targets -- -D warnings
- cargo test -p skippy-protocol --lib (26 passed)
- LLAMA_STAGE_BUILD_DIR=<local stage ABI build dir> cargo test -p skippy-server --lib (81 passed)
- LLAMA_STAGE_BUILD_DIR=<local stage ABI build dir> cargo check -p skippy-runtime -p skippy-ffi -p skippy-prompt -p skippy-bench -p skippy-correctness -p mesh-llm

## Rollback

Revert this PR.

### DIFF
--- a/crates/skippy-protocol/src/binary/activation.rs
+++ b/crates/skippy-protocol/src/binary/activation.rs
@@ -1,6 +1,8 @@
 use std::io;
 
-use super::{invalid_data, state_flags, WireActivationDType};
+use super::{
+    invalid_data, state_flags, types::MAX_STAGE_DECODED_ACTIVATION_BYTES, WireActivationDType,
+};
 
 pub fn activation_wire_bytes(
     dtype: WireActivationDType,
@@ -42,6 +44,19 @@ pub fn activation_wire_bytes_with_state_flags(
     }
 }
 
+pub(crate) fn activation_decoded_f32_bytes_with_state_flags(
+    token_count: i32,
+    n_embd: i32,
+    state_flag_bits: i32,
+) -> io::Result<usize> {
+    activation_wire_bytes_with_state_flags(
+        WireActivationDType::F32,
+        token_count,
+        n_embd,
+        state_flag_bits,
+    )
+}
+
 pub fn encode_f32_activation_payload(
     dtype: WireActivationDType,
     token_count: i32,
@@ -64,6 +79,11 @@ pub fn encode_f32_activation_payload_with_state_flags(
         n_embd,
         state_flag_bits,
     )?;
+    if expected_f32_bytes > MAX_STAGE_DECODED_ACTIVATION_BYTES {
+        return Err(invalid_data(
+            "decoded activation payload byte count exceeds maximum",
+        ));
+    }
     if f32_payload.len() != expected_f32_bytes {
         return Err(invalid_data("F32 activation payload size mismatch"));
     }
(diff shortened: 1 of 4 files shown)


ndizazzo (changes requested):
@IvGolovach PR looks ok conceptually, but my review agent caught a potential edge case:

---

Low / edge-case consistency issue: the wire and decoded activation caps are both currently set to 512 MiB, but compressed wire dtypes can expand past the decoded cap after the frame has already been accepted.

For example, an `f16` activation frame with a 300 MiB wire payload is below `MAX_STAGE_ACTIVATION_BYTES`, so `read_stage_message()` can accept it, but it expands to 600 MiB of decoded `f32` data and then fails later in `StageWireMessage::activation_f32_payload()`. `q8` has the same shape, with an even larger compressed-to-decoded gap.
That said, this seems unlikely under current defaults. Approximate decoded activation size is:

```text
decoded_f32_bytes = chunk_tokens * n_embd * sideband_multiplier * 4
f16_wire_bytes    = decoded_f32_bytes / 2
q8_wire_bytes     ~= decoded_f32_bytes / 4 + scale overhead
```

With today’s chunk defaults/caps:
- mesh-embedded Skippy default: prefill_chunk_size = 64
- standalone skippy-server default: 256
- adaptive max commonly capped around 512

A 512 MiB decoded activation would require roughly:

n_embd=8192,  multiplier=1  -> >16,384 tokens
n_embd=16384, multiplier=1  -> >8,192 tokens
n_embd=8192,  multiplier=4  -> >4,096 tokens
n_embd=16384, multiplier=4  -> >2,048 tokens

So this is probably not a common-path production failure today, but it is still a boundary-hardening inconsistency: compressed frames can pass the wire-size check and only fail later during decode. Consider deriving the accepted wire cap from MAX_STAGE_DECODED_ACTIVATION_BYTES and the selected wire dtype, or pre-checking the decoded size before accepting the frame into the stage chain.

IvGolovach:
I took another pass over the compressed activation case, and the edge case is real: the reader was bounding the compressed wire payload, while f16/q8 could still expand past the decoded f32 cap later.

I moved the decoded-size check into `read_stage_message()` before allocation/read, kept the decode-time guard as defense in depth, and added regressions for f16, q8, and q8 with the sideband multiplier. I also added the same decoded-cap guard on the encode path so we do not produce frames a downstream peer would reject.

Targeted protocol/server validation is green locally.

i386 (approved):
LGTM

Issue #359, opened 2026-04-20
Improved Heuristics for Context Size and Parallel Slots in `--auto`


ndizazzo:
## Background

When `mesh-llm serve --auto` launches llama-server, it picks two key resource parameters:

1. **Context size (`-c`)** — token budget for all active sequences combined.
2. **Parallel slots (`--parallel`)** — number of concurrent decode streams.

### Current state

**Context size** (`compute_context_size` in `inference/launch.rs`):

```rust
fn compute_context_size(ctx_size_override, model_bytes, my_vram, total_group_vram) -> u32 {
    let vram_after_model = my_vram.saturating_sub(host_model_bytes);
    if vram_after_model >= 30 * GB { 65536 }
    else if vram_after_model >= 12 * GB { 32768 }
    else if vram_after_model >= 6 * GB  { 16384 }
    else if vram_after_model >= 3 * GB  { 8192 }
    else                                { 4096 }
}
```

The thresholds are VRAM-only and completely ignore:
- The model's native max context length (`GgufCompactMeta::context_length`)
- The number of attention heads and head dimension (which drives per-token KV byte cost)
- KV cache quantization in use (Q4/Q8/F16 — already computed separately)
- Number of parallel slots (more slots × context = more KV)

**Parallel slots** (`runtime/mod.rs`):

```rust
let slots = primary_startup_model
    .and_then(|m| m.parallel)
    .or(config.gpu.parallel)
    .unwrap_or(4);  // ← hardcoded default
```

The default of 4 is a fixed number with no relationship to available VRAM, context size chosen, or model size. A node with 80 GB free VRAM gets the same 4 slots as a node with 4 GB free.

---

### Context size

1. **Ignores model's architectural max.** A model trained with a 4 096-token context window cannot usefully run at 65 536. Capping at `min(computed, gguf.context_length)` would be free correctness.

2. **Ignores KV cost per token.** The actual bytes consumed per token per layer is:
   ```
   kv_bytes_per_token = 2 × n_kv_heads × head_dim × bytes_per_element × n_layers
   ```
   With GQA and Q4/Q8 KV quantization, this can be 4–8× cheaper than naive F16 assumptions. A model with 8 KV heads at Q4 has vastly different headroom than a full-attention F16 model of the same size.

[...]

PR #594, opened 2026-05-19
Cover asymmetric KV cache width pricing

IvGolovach:
## Summary

This is now a narrow regression-test follow-up after current `main` absorbed and simplified the runtime auto-planning work from the original branch.

- Keeps current `main` runtime planning behavior, including the conservative unified-KV auto slot ceiling from the recent multi-node fixes.
- Drops the stale `RuntimePlanningProfile` / public-mesh lane escalation from this branch instead of fighting the newer runtime work.
- Adds coverage that GGUF KV-cache pricing handles asymmetric key/value widths independently.

## Why

The maintainer feedback was to update against current `main` and re-check the shape after the multi-node runtime changes. After rebasing, the right professional scope was not to reintroduce the older runtime planning profile. The remaining useful piece is the model-artifact regression coverage for asymmetric K/V metadata, which protects the capacity math used by the planner without changing runtime behavior.

## Diff scope

- `crates/model-artifact/src/gguf.rs`
  - Adds a regression test for independent K/V cache width pricing.

## Compatibility

- Test-only change.
- No runtime behavior change.
- No mesh wire protocol, gossip, protobuf, Skippy ABI, or plugin protocol compatibility change.

## Validation

Local validation:

- `git fetch --no-tags origin main:refs/remotes/origin/main codex/kv-context-slot-auto-planning:refs/remotes/origin/codex/kv-context-slot-auto-planning` - PASS
- `git rebase origin/main` - PASS
- `git diff --check origin/main...HEAD` - PASS, no output
- `git diff --check` - PASS, no output
- `git diff --cached --check` - PASS, no output
- `cargo fmt --all -- --check` - PASS
- `cargo test -p model-artifact gguf --lib` - PASS, 16 passed
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime runtime::context_planning --lib` - PASS, 9 passed
- `cargo check -p model-artifact` - PASS
- `LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo check -p mesh-llm` - PASS
- `/opt/homebrew/bin/cargo-clippy clippy -p model-artifact --all-targets -- -D warnings` - PASS

Remote validation:

- PR Quality Checks - PASS
- PR Builds - PASS
- PR Docker Build - PASS

Not run:

- Live performance smoke - not required after the final diff no longer changes runtime planning behavior. Current `main` runtime planning remains covered by targeted `runtime::context_planning` tests and remote CI.

## Rollback

Revert this PR.

Related: #359

### DIFF
--- a/crates/model-artifact/src/gguf.rs
+++ b/crates/model-artifact/src/gguf.rs
@@ -817,6 +817,23 @@ mod tests {
     }
 
     #[test]
+    fn kv_cache_quant_prices_key_and_value_widths_independently() {
+        let meta = GgufCompactMeta {
+            head_count: 32,
+            kv_head_count: 8,
+            layer_count: 24,
+            key_length: 64,
+            value_length: 256,
+            ..Default::default()
+        };
+        let quant = GgufKvCacheQuant::new(GgufKvCacheType::Q8_0, GgufKvCacheType::Q4_0);
+
+        assert_eq!(quant.k_cache_bytes_per_token(&meta), Some(13_056));
+        assert_eq!(quant.v_cache_bytes_per_token(&meta), Some(27_648));
+        assert_eq!(quant.kv_cache_bytes_per_token(&meta), Some(40_704));
+    }
+
+    #[test]
     fn kv_cache_bytes_per_token_returns_none_when_required_fields_are_missing() {
         let meta = GgufCompactMeta {
             head_count: 32,


crates/mesh-llm-host-runtime/src/runtime/context_planning.rs
```diff
@@ -66,11 +94,20 @@ fn planned_context_length(input: &RuntimeResourcePlanInput<'_>) -> u32 {
...
     if kv_bytes_per_token == 0 {
         return native_context;
     }
-    let max_affordable_context = kv_budget / kv_bytes_per_token;
+    let Some(kv_bytes_per_context_step) =
+        kv_bytes_per_token.checked_mul(u64::try_from(slots.max(1)).unwrap_or(u64::MAX))
+    else {
+        return MIN_AUTO_CONTEXT_LENGTH.min(native_context);
+    };
+    let max_affordable_context = kv_budget / kv_bytes_per_context_step;
```
ndizazzo: I don't think we should multiply KV budget by parallel lane count...  Existing split-planning code assumes we share one unified KV cache.  

If we include the lanes, we will end up with a much shorter context because of the multiplication than the runtime can actually support, reducing the total context depth instead of just reserving an amount for safe concurrency on each lane. Right now our context is created with `n_ctx = ctx_size, n_seq_max = lane_count`, and `kv_unified = true`. 

Directionally, I think we want to keep the KV memory sizing separate from lane concurrency as long as we're using the shared context setting.

ndizazzo (changes requested):

ndizazzo:
@IvGolovach thanks for pushing a fix!

ndizazzo (approved):

michaelneale:
@IvGolovach had to make some mega changes to this based on multi node testing I found - might want to update this to main, and check how it performs (I did consult this one but had to simplify it), am still a little confused on how to avoid things overcommitting and falling over, but glad this is looked at

PR #597, opened 2026-05-20
fix(routing): fail over fast off silently-wedged remote peers

michaelneale:
## What this fixes

Picking **MiniMax-M2.5** (or any model where the first hashed-to peer is silently wedged) from the web app dropdown would hang forever — `/api/responses` opens, browser shows "streaming response", no tokens flow. The user gives up; subsequent requests for the same prompt hit the same wedged peer because sticky prefix routing is deterministic on the prompt hash.

Models with 3+ healthy peers (e.g. Qwen3-8B on my local cluster) were unaffected because the first attempt always delivered.

## Root cause

In `network/openai/transport.rs`, the remote first-byte timeout is **5 minutes** per attempt. A peer that accepts the QUIC stream but never writes a response byte stalls each attempt for the full 5 min before the routing loop tries the next target. With 2 candidates that is up to 10 min — well past any user's patience.

Verified live on main (not ozempic) against the public mesh:

- 2 peers host MiniMax-M2.5; both look healthy in gossip (`state=serving`, fresh latency)
- `request_count` ticks 1 on the first MiniMax request; `attempt_unavailable_count` ticks 1 after the 5-min timeout
- Subsequent requests for the same prompt sticky-route to the same wedged target and also hang
- Qwen3-8B in parallel routes in 1.05 s with `successful_requests` ticking, so the proxy itself is healthy

## The fix

Use a **30 s** first-byte timeout for non-final remote attempts and keep the existing **5-minute** budget for the **last** attempt. The signal "more attempts remain" is already plumbed through the routing loop as `retry_context_overflow`, so the change is local to the remote-probe path and adds no new parameters at the call sites.

- New helper `non_final_remote_first_byte_timeout()` → 30 s
- New picker `remote_first_byte_timeout_for_attempt(more_attempts_remain)`
- New probe `probe_http_response_remote(reader, more_attempts_remain)`
- `route_remote_attempt_after_forward` now calls the new probe

Local attempts (`probe_http_response_local`, 10 min) are untouched — there is rarely a next local target, and local prefill of a big model can legitimately take minutes.

## Tests

- Helper picks short timeout when more attempts remain, full budget on the last attempt
- Probe times out on a wedged peer at the configured budget (fast 150 ms variant for CI)
- Existing `probe_http_response_local` slow-first-byte test still passes
- Full `cargo test -p mesh-llm-host-runtime --lib`: 1425 passed, 0 failed

## Diff size

1 file, +90/-1.

## Protocol

No protocol or wire-format change. Pure local timeout policy; mixed-version meshes are unaffected. Behavior with a single remote candidate (the last-attempt case) is identical to before.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
@@ -923,6 +923,27 @@ fn response_first_byte_timeout() -> Duration {
     Duration::from_secs(5 * 60)
 }
 
+/// Non-final remote first-byte timeout: 30 seconds. Used when more retry
+/// targets remain in the routing loop. Lets the loop fail over fast off a
+/// silently-wedged peer (accepts QUIC, never writes a byte) instead of
+/// stalling the whole 5-minute budget on every attempt before reaching a
+/// healthy peer. The full budget is reserved for the LAST attempt, where
+/// patience is the right behavior because there is nowhere else to go.
+fn non_final_remote_first_byte_timeout() -> Duration {
+    Duration::from_secs(30)
+}
+
+/// Pick the first-byte timeout for a remote attempt based on whether more
+/// targets remain. Short timeout on non-final attempts gives fast failover;
+/// the last attempt gets the full prefill budget.
+fn remote_first_byte_timeout_for_attempt(more_attempts_remain: bool) -> Duration {
+    if more_attempts_remain {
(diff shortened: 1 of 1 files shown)


michaelneale:
Closing — N=1 candidate case means this fix is a no-op for the actual symptom. The 30s-timeout-when-more-attempts-remain idea is correct but doesn't help today's MiniMax wedge (which always hits the final attempt). Not worth landing as theater.

PR #598, opened 2026-05-20
ci(client-auto): assert the node actually joined a mesh

michaelneale:
## What changed

The `Client-auto boot test` CI step now requires the node to actually join the public mesh, not just bind its console port.

Before this change, the step passed as long as `http://localhost:3132/api/status` returned 2xx. It explicitly treated "no public mesh discovered" and "node exited with 'no meshes found'" as inconclusive and exited 0. That meant a silent regression in Nostr discovery, gossip, or the join handshake would not fail CI \u2014 the script was effectively a console-bind smoke, not a public-mesh smoke.

After this change, once the management API is reachable, the script polls `/api/status` for up to 60s and requires:

- `mesh_id` set, **and**
- `peers` non-empty **or** `first_joined_mesh_ts` set.

If that signal never appears, the step fails with the last `/api/status` body for diagnosis. A "No meshes found after" process exit is now also a hard failure rather than an early pass.

## Why

The earlier behavior masked exactly the kind of break it was supposed to catch. If we want a CI signal that `client --auto` works end-to-end against the public mesh, the only honest version is one that fails when the join never happens.

## Tradeoff

This couples the PR/main CI lane to public-mesh availability. If meshllm.cloud / Nostr discovery is genuinely down, this step will fail until it recovers. That's intentional \u2014 we want to know about that, not silently pass over it.

## Diagnostics on failure

On failure, the step prints:

- whether the binary logged `No meshes found yet` (discovery returned nothing)
- the last `/api/status` body, indented for readability

## Files

- `scripts/ci-client-auto-test.sh` only. No workflow YAML changes needed \u2014 `.github/workflows/ci.yml:330` and `.github/workflows/pr_builds.yml:395` already invoke this script.


ndizazzo (approved):
Looks fine, but a small edge case is that it might give a false positive... If no public mesh is found, the client might fall back to creating/using its own standalone mesh

michaelneale:
yeah @ndizazzo good point - I am actually not sure what I am really testing here... so will try one more time and close it if not sure.

PR #607, opened 2026-05-20
Reconcile model targets into local runtime loads

IvGolovach:
## Summary

This makes model-target advice actionable for local runtimes, while keeping the behavior explicit and opt-in.

When `[runtime] reconcile_model_targets = true` is set, a node can automatically runtime-load a model that is already present on local disk when that model is:

- registered as a local explicit interest,
- currently wanted by `/api/model-targets`,
- not already served by the mesh,
- advised as a single-node fit,
- small enough for the current node's local capacity.

## Why

#579 added advisory capacity evaluation for model targets. This follow-up closes the next local-only step: a node that already has a wanted model on disk can act on that advice without requiring manual runtime load calls.

The guardrails are intentionally narrow. This does not download models, start split serving, act on peer-only demand, or change gossip/protocol compatibility.

## Implementation

- Adds `[runtime] reconcile_model_targets = true` as an opt-in local runtime config flag.
- Adds a small reconciliation planner with in-flight, failure, and manual-unload cooldown tracking.
- Wires a periodic reconciliation tick into the current runtime event loop shape.
- Queues normal runtime `Load` requests rather than introducing a separate launch path.
- Records reconciliation load completion/failure through runtime events.
- Reconciles Hugging Face model identities conservatively, so `repo@main:selector` and `repo:selector` match while non-main revisions remain distinct.
- Preserves current main's typed runtime unload flow and records manual-unload cooldowns through the canonical runtime target identity.
- Preserves current main's expanded configuration UI/runtime config and OpenAI guardrail runtime-loop surfaces while carrying the runtime reconciliation config through TOML deserialization, owner-control apply payloads, and the runtime event loop.
- Documents the behavior in the design/testing notes.

## Compatibility

- Disabled by default.
- Local config only; no mesh gossip, peer protocol, protobuf, or plugin protocol changes.
- No model downloads or split-serving automation.
- Existing manual runtime load/unload behavior remains supported.
- Manual unload suppresses immediate auto-reload by model ref/name cooldown, including equivalent Hugging Face main-revision aliases.

## Branch integrity

- Base branch: `main`
- Validated base: `origin/main@148fffde031b4b1005c1d209e109eacd86a8581e`
- Branch state after rebase: `0 behind / 1 ahead`
- Merge base: `148fffde031b4b1005c1d209e109eacd86a8581e`
- Head commit: `edf3e5f0905706fd35a60fbdc5d8cd40167438fd`

## Validation

- Validation tier: Tier 3 - opt-in shared runtime reconciliation from advisory model targets into local runtime loads, with post-review alias/cooldown repair and post-main conflict repair after the configuration UI/runtime config and guardrail surfaces landed on main.
- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, `origin/main` at `148fffde031b4b1005c1d209e109eacd86a8581e`.
- `git rebase origin/main`: PASS, resolved the runtime CLI `MeshConfig` initializer conflict by preserving both current main's `extra` config field and PR #607's `runtime` reconciliation config; resolved the runtime loop context conflict by preserving current main's OpenAI guardrail policy handle plus PR #607's model-target reconciliation policy/state.
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime control_plane_api_cli_builds_apply_request_body --lib -- --test-threads=1`: PASS, 1 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime model_target_reconciliation --lib -- --test-threads=1`: PASS, 17 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime config_sync --lib -- --test-threads=1`: PASS, 22 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime plugin::config --lib -- --test-threads=1`: PASS, 45 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal /opt/homebrew/bin/cargo-clippy clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS
- Remote CI on `edf3e5f0905706fd35a60fbdc5d8cd40167438fd`: PASS, including PR Quality Checks, PR Docker Build, PR Builds, Skippy Inference Smoke Tests, and two-node client/serving smoke.
- Version: not applicable - no release/version sync required for this non-release PR branch.
- Not run: `just build` - not required for selected validation tier; no UI or bundle artifact changed.
- Not run: live multi-node inference smoke - not required for local PR-ready proof; feature is opt-in and remote CI/reviewer runtime validation should gate merge.

## CI context confirmation

Remote CI passed on `edf3e5f0905706fd35a60fbdc5d8cd40167438fd`.

## Rollback

Revert this PR.

DB downgrade: not applicable.
Data repair: not applicable.

## Known residual risks

Live multi-node behavior still deserves maintainer/runtime validation before enabling the opt-in flag broadly.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/cli/commands/runtime.rs
+++ b/crates/mesh-llm-host-runtime/src/cli/commands/runtime.rs
@@ -570,6 +570,7 @@ mod tests {
             owner_control: Default::default(),
             telemetry: Default::default(),
             defaults: None,
+            runtime: Default::default(),
             extra: Default::default(),
         };
 
--- a/crates/mesh-llm-host-runtime/src/plugin/config.rs
+++ b/crates/mesh-llm-host-runtime/src/plugin/config.rs
@@ -26,6 +26,8 @@ pub struct MeshConfig {
     #[serde(default)]
     pub defaults: Option<ModelConfigDefaults>,
     #[serde(default)]
+    pub runtime: RuntimeConfig,
+    #[serde(default)]
     pub models: Vec<ModelConfigEntry>,
     #[serde(rename = "plugin", default)]
     pub plugins: Vec<PluginConfigEntry>,
@@ -47,6 +49,12 @@ pub struct GpuConfig {
     pub assignment: GpuAssignment,
     #[serde(default)]
     pub parallel: Option<usize>,
+}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+pub struct RuntimeConfig {
+    #[serde(default)]
+    pub reconcile_model_targets: bool,
 }
 
 #[derive(Clone, Copy, Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
@@ -548,6 +556,8 @@ struct RawMeshConfig {
     #[serde(default)]
     defaults: Option<ModelConfigDefaults>,
     #[serde(default)]
+    runtime: RuntimeConfig,
+    #[serde(default)]
     models: Vec<ModelConfigEntry>,
     #[serde(rename = "plugin", default)]
     plugins: Vec<PluginConfigEntry>,
@@ -644,6 +654,7 @@ impl<'de> Deserialize<'de> for MeshConfig {
             owner_control: raw.owner_control,
             telemetry: raw.telemetry,
             defaults: raw.defaults,
+            runtime: raw.runtime,
             models: raw.models,
             plugins: raw.plugins,
             extra: raw.extra,
@@ -2737,6 +2748,22 @@ flash_attention = "enabled"
             config.models[0].flash_attention,
             Some(FlashAttentionType::Enabled)
         );
+    }
+
+    #[test]
+    fn runtime_model_target_reconciliation_deserializes_from_toml() {
+        let config: MeshConfig = toml::from_str(
+            r#"
+version = 1
+
+[runtime]
+reconcile_model_targets = true
+"#,
+        )
+        .unwrap();
+
+        assert!(config.runtime.reconcile_model_targets);
+        validate_config(&config).unwrap();
     }
 
     #[test]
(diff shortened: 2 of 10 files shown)


i386 (commented):
I found one issue in the reconciliation path: it checks whether a target is already loaded using strict string equality, but the same model can be represented by different valid refs across the target and runtime-load paths.

For example, a target/ref like `unsloth/Qwen3-8B-GGUF@main:Q4_K_M` can be reconciled by queueing a load from the local GGUF path. The existing runtime load path then derives the served model name from `models::model_ref_for_path(&model_path)`, which can become `unsloth/Qwen3-8B-GGUF:Q4_K_M`. On the next reconciliation tick, `loaded_target()` only checks exact matches against `target.model_ref` or `target.model_name`, so it can miss that the target is already satisfied and queue another load. The same alias mismatch can also bypass the manual-unload cooldown.

I think the right fix is to normalize/alias-match model identities before planning reconciliation: compare target refs/names and loaded model refs through the same catalog/HF-cache identity rules used by `/api/model-targets`, rather than relying on exact string equality. That keeps `/runtime/load` naming behavior unchanged while making reconciliation answer the intended question: is this target already satisfied locally?

Relevant code paths: `reconcile_model_targets_once()` builds actions from `target.model_ref` but queues the load by local path, `run_auto_load_runtime_model()` derives the loaded name from the resolved path, and `loaded_target()` only does exact string checks.

i386:
@IvGolovach I updated this branch with latest `main` and resolved the merge conflict. The PR is no longer marked dirty; checks are queued/pending now.

PR #609, opened 2026-05-20
Restore llama.cpp upstream canary patch replay

IvGolovach:
## Summary

Restores the llama.cpp upstream canary by rebasing the Mesh-LLM patch queue onto current upstream llama.cpp, now replayed on fresh `main`.

- Updates the pinned llama.cpp upstream SHA to `6a257d44633d4a752183ed778b88d2924d0a6b9d`.
- Reworks the Qwen35MoE runtime-slice patch for upstream's current NextN/MTP graph shape.
- Refreshes the dependent Qwen35MoE recurrent-only slice patch.
- Refreshes later patch contexts that now depend on upstream's added `h_pre_norm` graph result path.
- Keeps the PR focused on the llama.cpp patch queue and upstream pin after rebasing onto latest `main`.
- Updates the parity helper's inherited-graph allowlist so Mistral4's DeepSeek2-backed staged graph is recognized during validation.

## Why

The scheduled llama.cpp upstream canary was failing before it could build or smoke-test Skippy because patch `0009-Support-Qwen35MoE-runtime-slice-packages.patch` no longer applied to upstream master.

The root cause is upstream graph churn around Qwen35/Qwen35MoE: the main transformer stack is now separated from NextN/MTP blocks, and the graph also carries a pre-norm output path. Skippy runtime slices need to execute only the main transformer layer range while preserving the upstream MTP/pre-norm behavior for full/output stages.

After review, this branch was rebased onto current `main` so the patch replay is validated with the latest MoA/runtime and Swift smoke fixes already landed upstream.

## Implementation

- Rebased `0009` so Qwen35MoE slices clamp staged execution to `n_layer - nextn_predict_layers`, keep token embeddings only on the first slice, and build output row ids only for full/output slices.
- Rebased `0010` so recurrent-only Qwen35MoE slices still avoid attention-position inputs when the selected slice contains no attention layer.
- Refreshed `0057` and `0076` around upstream's `h_pre_norm` graph result addition while preserving RWKV7 and Gemma3n activation sidebands.
- Rebased `0064` with the same Qwen35 NextN/MTP treatment used for Qwen35MoE.
- Advanced `third_party/llama.cpp/upstream.txt` to the validated upstream SHA.
- Added `mistral4` to the parity helper's inherited staged-graph mapping because upstream dispatches Mistral4 through the DeepSeek2-backed graph implementation.

## Compatibility

- No mesh gossip, protobuf, plugin protocol, or client protocol compatibility change.
[...]

PR #610, opened 2026-05-21
ci(sdk-smoke): install lld in macOS swift smoke job

michaelneale:
## Why

The Swift SDK smoke job has been failing on macOS runners with:

```
clang: error: invalid linker name in argument '-fuse-ld=/opt/homebrew/bin/ld64.lld'
```

This is hitting unrelated PRs (#566, #609) — it is not specific to any one branch.

## What

The transitive cargo build inside `sdk/swift/scripts/generate-swift-bindings.sh` runs in a temp dir and ends up invoking `cc` with `-fuse-ld=/opt/homebrew/bin/ld64.lld`. Apple clang accepts that flag only when the linker binary actually exists on disk; the swift smoke job only installed `jq`, so the link step failed.

The `macos_targets` job (in `pr_builds.yml`) already does `brew install ... lld` for the same reason. This PR mirrors that in the reusable `sdk-smoke.yml` workflow so the swift smoke lane has `lld` available too.

## Follow-up

PR #566 has `swift_sdk_smoke` gated off with a `false &&` while this fix lands. Once this PR is merged and proven green, that gate should be removed (a one-line revert on the affected branch).

### DIFF
--- a/.github/actions/compute-changes/action.yml
+++ b/.github/actions/compute-changes/action.yml
@@ -146,7 +146,7 @@ runs:
         if [[ "${{ inputs.event_name }}" == "workflow_dispatch" ]]; then
           SDK_SMOKE_REQUIRED="true"
         elif [[ -n "$CHANGED_FILES" ]]; then
-          DIRECT_SDK_INPUTS=$(echo "$CHANGED_FILES" | grep -E '(^sdk/|^Package\.swift$|^scripts/ci-(native|kotlin|swift)-sdk-smoke\.sh$|^scripts/ci-sdk-fixture\.sh$)' || true)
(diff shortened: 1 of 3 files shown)


ndizazzo (approved):
yeah idk why we didn't before

michaelneale:
not sure if this actually works as the CI was too smart @ndizazzo how do we get it to actually run the swift smoke?

ndizazzo:
@michaelneale We have support to dispatch the workflow manually for this problem... since we aggressively tune the PR job so that it's as fast as possible, we needed a way to trigger it:

`gh workflow run pr_builds.yml --ref micn/swift-smoke-install-lld` in the CLI, or in the GitHub UI, you can navigate to the workflow and run it with a branch:

image

When dispatched, it sets all the fancy skipping to false.

PR #612, opened 2026-05-21
moa: harden gateway from PR #566 review (panic surface, error propagation, dedup race, dead code)

michaelneale:
## What this gives you

Follow-up to #566. Addresses every actionable item from the Copilot review on that PR \u2014 11 HIGH defects, 4 MEDIUM cleanups, 6 LOW doc corrections \u2014 plus a README note marking MoA as experimental.

The mainstream model surface is untouched. All changes are scoped to MoA-only code paths (`model: "mesh"`), one shared HTTP transport helper, and three doc files. Direct `model=<id>` and `model=auto` requests behave exactly as before.

### What you can now do / see

* Send a malformed worker response (`confidence: NaN`, missing `choices`, primitive `arguments`) and the gateway no longer panics \u2014 it returns a structured error or coerces the shape.
* When a reducer fails on a tool-result turn, you get **HTTP 502** (was 200 with an in-band error body).
* SSE streaming clients (Goose, OpenAI SDKs) now see `finish_reason: "error"` and an explicit error chunk on MoA failure (was a silent `stop`).
* Reducers that produce a tool name without arguments now emit a real `tool_calls` reply with `{}`, matching the fan-out path \u2014 agent harnesses no longer drop tool calls to reducer prose.
* When the shortest-name peer for a model is offline, the worker pool now falls back to a reachable longer-form alias instead of silently dropping the model.
* When the same model is served locally under one alias and by a peer under another, the local skippy-port wins regardless of name length \u2014 no more unnecessary QUIC hops.
* Error responses now distinguish `all_workers_failed` from `all_reducers_failed` so clients can branch on `error.code`.
* `strip_thinking` is now linear (was O(n\u00b2)) on long worker outputs.
* HTTP header names from MoA observability writers are RFC 7230 tchar-validated; CR/LF can't be smuggled.
* README clearly marks `model: "mesh"` as experimental so users know the surface may still change.

### Validation

| Check | Result |
|---|---|
| `cargo test -p mesh-mixture-of-agents --lib` | **86 / 86 pass** (+16 new) |
| `cargo test -p mesh-mixture-of-agents` (incl. integration) | all pass |
| `cargo test -p mesh-llm-host-runtime --lib` | **1454 / 1454 pass** (+4 new) |
| `cargo test -p skippy-cache --lib` | 13 / 13 pass |
| `cargo test -p model-ref --lib` | 10 / 10 pass |
| `cargo clippy ... --all-targets -- -D warnings` | clean |
| `cargo fmt --all -- --check` | clean |
| `just build` | green |

### Live end-to-end validation

Built locally and deployed to a 2-node private mesh (M4 Max with Qwen3-8B + Studio M3 Ultra with MiniMax-M2.5, 138 GB, 131k context). Verified with the same harness suite as PR #566:

* Basic `/v1/chat/completions` to Qwen3-8B (local) \u2014 OK.
* Basic `/v1/chat/completions` to MiniMax (remote via QUIC) \u2014 OK.
* `mini-agent.py model=auto` \u2014 2 turns, correct bug ID.
* `mini-agent.py model=mesh` (MoA) \u2014 2 turns, correct bug ID.
* `mini-agent2.py model=mesh` (multi-file exploration) \u2014 4 turns, correct.
* Goose `model=mesh` \u2014 1 tool call, correct.

Also verified the public-mesh path with `mesh-llm client --auto`: joined the public mesh in ~8 seconds, saw 13 peers, called Qwen3-8B end-to-end and got a clean reply.

## Commit shape

7 commits, each runnable in isolation. Squash-merge-friendly.

| Commit | Phase | What |
|---|---|---|
| `a02d407e` | docs | README: mark `model: "mesh"` MoA as experimental |
| `8eb1b6ac` | HIGH | Close panic surface in worker output normalization (NaN clamp, dead or_else, malformed-response handling) |
| `ff4e4789` | HIGH | Propagate failure signals to HTTP status, SSE finish_reason, and the tool-result reducer path |
| `9e360c58` | HIGH | Group aliases by canonical base before resolving backend (fixes dedup race + forced QUIC hop) |
| `8b6115b2` | HIGH | Strip dead `Session` continuation / running-summary machinery (per-request scope) |
| `6908c1b0` | MEDIUM | Distinct error codes, linear `strip_thinking`, RFC 7230 header-name validation, drop dead branch in ingress |
| `6fa3dfbf` | LOW | Align stale comments and doc claims with current code (5 spots + MOA_GATEWAY.md) |

## Architecture

No protocol or wire-format changes. No new public types. `error_response` now takes a `code` parameter (internal-only) and `Session` lost five accessors that were never called in production. The gateway is documented as request-scoped now \u2014 caller owns the multi-turn loop.

## Protocol

None. MoA gateway is local-decision; nothing crosses node boundaries that didn't before. SSE wire shape gains an optional error chunk on failure, which is additive (clients that didn't read it before still don't have to).

## Notes for review

* The 11 HIGH items are listed (with their PR #566 review comment links) in the per-commit messages so reviewers can trace each fix back to the original Copilot remark.
* The `router.rs:1360` wall-clock-seeded test comment from the Copilot pass is intentionally not addressed \u2014 the test is not flaky in practice and has passed every CI run; will reply on PR #566 with the rationale.
* The Swift SDK smoke gate from PR #566 doesn't exist on this branch; it lives on `main` and is being fixed by PR #610 separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,13 @@ mesh-llm serve --auto --headless
 For a deeper operator guide, see [docs/USAGE.md](docs/USAGE.md). For every CLI
 command and switch, see [docs/CLI.md](docs/CLI.md).
 
-## Mixture-of-Agents (`model: "mesh"`)
+## Mixture-of-Agents (`model: "mesh"`) — experimental
+
+> ⚠️ **Experimental.** The MoA gateway is new in this release. Behavior,
+> routing heuristics, error shapes, and tuning knobs may change between
+> versions while we tune it. Treat `model: "mesh"` as a preview feature
+> rather than a stable production path; use a specific model id when you
+> need stable semantics.
 
 Send a request with `"model": "mesh"` and the proxy fans it out to every
 model available in the mesh in parallel, arbitrates their responses with
--- a/crates/mesh-llm-host-runtime/src/inference/skippy/family_policy.rs
+++ b/crates/mesh-llm-host-runtime/src/inference/skippy/family_policy.rs
@@ -260,14 +260,15 @@ fn resident_kv_policy(activation_wire_dtype: StageWireDType) -> FamilyPolicy {
             // prefixes are recorded, so cache hit rate for the
             // recent workload is preserved.
             //
-            // Live observation: even at 16, sustained Goose `auto`
-            // traffic eventually starves the lanes — see the
-            // `prefix cache leak under sustained agent traffic`
-            // section in `docs/design/MOA_BRANCH_REPORT.md`. The
-            // 16-entry cap pushes the failure point from ~6 to ~16+
-            // requests but does not fully eliminate it. The proper
-            // fix is more invasive in `skippy-server` and out of
-            // scope for PR #566.
+            // The entry-count cap is the *coarse* lever: it bounds
+            // how many distinct prefixes the cache can hold, but with
+            // `kv_unified = true` even 16 long prefixes can pin the
+            // full cell pool. The complementary fine-grained cell
+            // budget (`max_resident_tokens` in
+            // `ResidentCacheConfig::from_stage`, landed in PR #566)
+            // closes the remaining gap by evicting on token pressure
+            // before the cell pool runs out. The 16-entry cap is
+            // still useful as a structural ceiling.
             max_entries: 16,
         },
     }
--- a/crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
@@ -604,20 +604,21 @@ async fn try_handle_moa_intercept(
     if decision.effective_model.as_deref() != Some(moa::VIRTUAL_MODEL_NAME) {
         return MoaInterceptResult::NotMoa(tcp_stream);
     }
-    // try_handle_moa self-gates on the model name; the Some(_) return path
-    // means it declined to handle despite our outer gate — log and treat as
-    // handled (we can't reliably re-use the stream after the handoff).
-    if let Some(_unused_stream) = crate::network::openai::moa_gateway::try_handle_moa(
+    // `try_handle_moa` self-gates on the model name and consumes the
+    // stream when it accepts. The outer gate above guarantees the gate
+    // matches, so the inner call always returns `None` here — the stream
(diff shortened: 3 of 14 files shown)


ndizazzo:
@michaelneale Couple findings from mine too:

1. Streaming MoA failures still return HTTP 200 (run_moa_turn() lines 55-70)
   - write_moa_response() detects failure state, but streaming responses still go through send_moa_as_sse(), which hardcodes HTTP/1.1 200 OK.
   - So stream=true MoA failures like all_workers_failed or all_reducers_failed appear successful at the HTTP layer, unlike non-streaming failures that now return 502.

3. The new SSE error chunk may break OpenAI-compatible stream clients (send_moa_as_sse() lines 453-553)
   - send_moa_as_sse() emits an error-only chunk with choices: [].
   - Many clients assume every chat.completion.chunk has choices[0], so this can cause client-side parse/index errors instead of a clean failure signal.

ndizazzo (commented):
see agent findings

michaelneale:
Thanks @ndizazzo \u2014 both addressed in `45d84013`.

**Finding 1 (streaming failures returned 200)**: My earlier code comment claiming "once the headers are flushed we cannot change the status code" was wrong \u2014 the body is fully available before `send_moa_as_sse` is called, so we can switch transports based on the body. Streaming + failure now collapses to a non-streaming HTTP 502 JSON response with the structured error body. Matches the OpenAI API shape (streaming-endpoint failures come back as non-streaming JSON with the right status) and means streaming and non-streaming failure paths are consistent.

**Finding 2 (error chunk with `choices: []` could break clients)**: Good catch. With finding 1's routing change, failure-shaped bodies never reach `send_moa_as_sse` anymore, so the error-chunk emission is dead code. Dropped entirely. `send_moa_as_sse` now does one clear thing: a delta chunk + a stop chunk, both with a real `choices[0]`. `debug_assert!` on `!is_moa_failure_body` pins the invariant in tests.

Four new unit tests cover the four corners of the routing decision (streaming/non-streaming \u00d7 success/failure). Live-tested both happy paths on the 2-node mesh; mini-agent + Goose harnesses also pass.

ndizazzo (approved):

PR #620, opened 2026-05-21
feat(moa): default to no-think on workers + reducer (escape hatch for callers)

### LINKED ISSUE #617: MoA: fast worker shouldn't think — thinking models can poison fan-out and leak reasoning to chat
## Summary

When MoA dispatches to a thinking model (Qwen3-*, DeepSeek-R1-Distill, etc.) in the **fast worker** slot, the model burns its 256-token budget inside its `<think>...</think>` block and never reaches the actual answer. MoA then sees a worker payload that is all reasoning text with an unclosed `<think>` opener, with no `kind:/confidence:/payload:` envelope.

This produces three concrete failures:

1. **Reasoning leaks to chat.** When the slow / smart worker hasn't returned yet (or is slow on a degraded mesh), the new sole-answer grace path (#615) ships the fast worker's reasoning-only payload. Users see `"Okay, the user wants a haiku. Let me recall…"` instead of an actual haiku.
2. **Reducer poisoning.** When both workers complete, the arbiter routes mixed outputs to the reducer. The reducer prompt now has a candidate "answer" that is reasoning prose, polluting its context.
3. **Confidence threshold gaming.** With no `confidence:` field parsed, `heuristic_classify` defaults to `0.5`, which is just enough to qualify the worker for grace early-exit and consensus.

## Reproduction (2-node private mesh)

* Node A: `mesh-llm serve --gguf <Qwen3-8B-Q4_K_M.gguf> --port 9337 --console 3131`
* Node B: `mesh-llm serve --model unsloth/MiniMax-M2.5-GGUF:Q4_K_M --port 9337 --console 3131 --join <invite>`
* On A: `curl -X POST http://localhost:9337/v1/chat/completions -H 'content-type: application/json' -d '{"model":"mesh","messages":[{"role":"user","content":"Write a haiku about the ocean."}],"stream":false}'`

In a sample of 5+ runs on this topology, \~40% of turns shipped raw reasoning text instead of the answer. Gateway logs show the fast Qwen3-8B worker returning 800-960 char payloads — the unstripped thinking block — while MiniMax returns 80-100 char clean haikus.

## Why this is worse for MoA than for stock inference

Stock single-model inference of a thinking model is fine: callers get `<think>...</think>` followed by the answer in one stream. Total tokens are higher, but `response.reasoning_text.delta` lets clients render them separately and the answer always lands.

MoA pulls the rug out by:
* Capping the fast worker at `max_tokens = 256` in `crates/mesh-mixture-of-agents/src/context.rs` (`pack_fast`).
* Asking for a structured `kind:/confidence:/payload:` envelope that the model never reaches because the think block ate the budget.
* Then treating the half-thought output as a candidate answer.

## Proposed fixes (any one would close the worst cases; B is the right shape)

**A. Bump fast worker budget when the model is a known thinker.** Cheapest, \~10 LoC. `pack_fast` looks at a per-model `wants_reasoning_budget` hint and uses 1024 tokens instead of 256. Plaster, doesn't really solve it.

**B. Tell the fast worker to *not* think.** Right shape. The fast slot is supposed to be "give me a quick first impression" — chain-of-thought defeats its purpose. Two sub-options:
  * Pass `enable_thinking=false` (or the equivalent OpenAI-compat field many backends accept) in the upstream chat request when targeting a thinking family.
  * Inject a system-prompt suffix that explicitly disables reasoning (`Respond with the final answer only. Do not use <think> tags or chain-of-thought reasoning.`) for the fast worker.

**C. Defensive: confidence-demote unclosed-think payloads in `normalize_worker_output`.** \~20 LoC, belt-and-suspenders. If the payload starts with `<think>` and never closes, set `confidence = 0.1`. Grace eligibility (`>= 0.5`) skips it, consensus deprioritizes it, the reducer prefers other candidates. Worth landing alongside B.

## Out of scope here

* Grace itself is fine: it correctly skips agentic / tool turns (`has_tools=true`) and only fires on a single confident answer in chat. See PR #615 for the grace contract.
* Stock single-model inference of thinking models: works as-is. Not a regression to fix in this issue.
* Token streaming through MoA: covered by a different follow-up (MoA emits one-shot delta; would be nice to stream the winning worker's tokens once the arbiter locks a decision).

## Acceptance criteria

* On the 2-node private mesh repro above, 10/10 runs of the haiku prompt return a clean haiku in the chat reply (no `<think>` prefix, no raw reasoning prose).
* On the same repro with `tools=[{read_file}]`, the tool-call path still returns a proper `finish_reason=tool_calls` shape (unchanged).
* A new unit test in `crates/mesh-mixture-of-agents/tests/` constructs a stubbed thinking-style worker output (`<think>blah blah` with no close) and asserts MoA demotes it / does not ship it as the response.

michaelneale:
For `model: "mesh"` (MoA), the OpenAI reasoning knobs were silently dropped, so reasoning models thought regardless. The 256-token fast-worker budget got burned inside an unclosed `<think>` block and never reached the answer; the reducer's quality dropped when candidate answers leaked reasoning prose.

**This PR makes MoA opinionated**: workers and the reducer do NOT think by default. The caller can still explicitly enable thinking for an individual request (escape hatch), but for the chat UI's `Auto` (which routes through MoA) and any other `model: "mesh"` call without explicit knobs, MoA picks: reasoning off.

Closes #617.

## Why opinionated, not opt-in

Workers in MoA are short-budget internal slots, not user-facing reasoning steps. The fast worker has a 256-token budget that doesn't fit `<think>\u2026</think>` + answer. The reducer doesn't want reasoning prose as candidate input. And the chat UX is the worst case for thinking models — every turn pays the reasoning latency penalty without showing the user the reasoning output. Asking every MoA caller to remember a flag is the wrong shape; MoA should just do the right thing.

The escape hatch covers experimentation: any of `reasoning_effort: "low"` (or any non-"none" value), `reasoning: { enabled: true }`, `enable_thinking: true`, etc. turns it back on for that request.

## Implementation

* `SamplingParams` gains `enable_thinking: Option<bool>` and a builder-style `with_thinking(...)` helper.
* New `backend::apply_enable_thinking(body, hint)` injects `chat_template_kwargs.enable_thinking` (canonical llama.cpp knob) and `reasoning_effort: "none"` when disabled. Merges into any existing `chat_template_kwargs`.
* `HttpBackend`, `LocalModelBackend`, and `RemoteModelBackend` call it after building their request bodies.
* `GatewayConfig` gains `enable_thinking: Option<bool>` so the choice flows in one place to every worker (`SamplingParams::worker().with_thinking(...)`) AND the hedged reducer.
* New tiny pure function `effective_enable_thinking_for_moa(body) -> Option<bool>` encodes the policy: `extract_enable_thinking_override(body).or(Some(false))`. Easy to unit-test, easy to flip if we change our minds.

## Live verification on 3-node lab mesh

Same prompt ("reply with one short word"), same mesh, release binary:

| Probe | Result |
|---|---|
| `model=mesh`, no knobs (3 runs) | "Hello" / "Okay" / "Hello" \u2014 clean, zero think leakage |
| `model=mesh`, `reasoning_effort: "low"` | Thinking turns back on as requested \u2014 raw "Thinking Process:\u2026" prose visible |
| `model=mesh`, `tools=[read_file]` | Tool path unchanged. `finish_reason: tool_calls`, args correct, no-think default applied. |

For comparison, the pre-fix smoke (also on the same mesh) showed 32s/17s/32s turns on the same prompt with raw reasoning prose leaking in 2/3.

## Tests

* mesh-mixture-of-agents lib: 98 pass (6 new for `apply_enable_thinking` and `SamplingParams::with_thinking`).
* mesh-mixture-of-agents `tests/sim_enable_thinking_propagation.rs` (new): 3 mock-backend integration tests pinning that `chat_template_kwargs.enable_thinking` reaches every worker AND that no spurious fields appear when no override is requested.
* mesh-llm-host-runtime lib: 1486 pass (14 new):
  * 10 for `extract_enable_thinking_override` covering every JSON shape (`reasoning_effort`, `reasoning: { enabled / effort / max_tokens }`, `enable_thinking`, all THINKING_BOOLEAN_ALIASES, `thinking_budget: 0`, `chat_template_kwargs`).
  * 4 for `effective_enable_thinking_for_moa` covering the opinionated default + the escape hatch + the tool path.
* `cargo clippy --all-targets -- -D warnings`: clean.
* `cargo fmt --all -- --check`: clean.

## Compat

* No mesh wire-protocol change. The new fields travel inside the existing chat-completion request JSON over the QUIC tunnel; peers on older binaries simply forward the body to llama.cpp which already understands `chat_template_kwargs`. Additive only, safe across mixed-version meshes.
* No skippy ABI change. No plugin protocol change.
* **Behavior change for existing MoA callers**: a caller who today relies on MoA producing reasoning output will now get short answers instead. The escape hatch (`reasoning_effort: "low"` or similar) restores the old behavior for that caller. The chat UI is the main consumer of `model=mesh` and benefits from this change.

## Follow-up

Lab will keep running with a `mesh_no_think` probe alongside `mesh_chat` for longer-horizon A/B. Tomorrow report will include latency distribution under a few hours of mixed traffic.


michaelneale:
## Honest testing readout

Wanted to be explicit about how much real-world testing this PR has had so far, in case the original description read more confidently than it should:

**Verified live**, on a clean 3-node private mesh (M4 Qwen2.5-3B + studio MiniMax-M2.5 + mini Qwen3.5-9B), release-build binary:

- 3 paired runs of `mesh_chat` (default) vs `mesh_no_think` (`reasoning_effort: "none"`). 32s/17s/32s vs 1s/2s/1s. Worker logs on the slow runs show Qwen3.5-9B returning 1700+ char unclosed-`<think>` payloads; on the no-think runs all three workers finish in <1.3s with clean 4-5 char answers.
- Restart cycle: studio + mini reconnected after M4 restarted with the new binary, the QUIC mesh + model gossip recovered cleanly.

**Verified via unit/integration tests, not live**:

- The agentic / tool-calling path still bypasses the thinking override correctly. Covered by `sim_enable_thinking_propagation.rs` recording-backend tests, but I didn't manually probe `tools=true` + `reasoning_effort: "none"` in the live lab.
- The reducer path receives the override. Same story \u2014 unit test coverage via the recording backend, but in the 3 live smoke pairs the consensus path won every time so the reducer didn't fire.
- All the JSON shape variants (`reasoning: { enabled: false }`, `thinking_budget: 0`, `chat_template_kwargs.enable_thinking`, the THINKING_BOOLEAN_ALIASES set). Each one has a unit test in `extract_enable_thinking_override` but only `reasoning_effort: "none"` was tried live.

**Not yet verified**:

- CI hasn't run yet; will check.
- Chat UI in the web console with the toggle (no UI toggle exists yet \u2014 that's the follow-up I noted).
- Long-horizon stability (hours of mixed probes). The lab has a new `mesh_no_think` probe alongside `mesh_chat` collecting that A/B data now; expect a meaningful sample by tomorrow.

The smoke result was big enough that I felt OK opening the PR, but the long-horizon and live-tool-path numbers are still TBD.

michaelneale:
Superseded by #629 which combines this PR + #624 into a single MoA chat improvement (40s → ~1.2s p50 on public mesh). The two changes were inseparable in the user journey, and the combined branch is cleaner (no lab notes, no accidental QUIC retry code from earlier merges). Old branch left in place for git history; safe to delete.

PR #621, opened 2026-05-21
feat(moa): chunked output_text.delta so /v1/responses streams smoothly

### LINKED ISSUE #618: MoA: stream the winning worker's tokens instead of one-shot delta
## Summary

MoA currently returns its final answer as a **single `response.output_text.delta` event** containing the whole reduced answer. The Responses-API stream is technically correct (`response.created` → 1 delta → `response.output_text.done` → `response.completed` → `[DONE]`), but the chat UI renders it like a hard pop-in: spinner for 5-20s, then the whole answer appears at once.

Direct single-model calls through `/v1/responses` token-stream cleanly (verified: 26 deltas for a 'count to ten' prompt on Qwen2.5-3B). MoA is the only path with the one-shot delta feel.

## Why it's like this today

`crates/mesh-mixture-of-agents` produces one chat-completion JSON body (the arbitrated / reduced answer). `crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs::send_moa_as_responses_sse` translates that body to Responses-API SSE by emitting a single delta with the full content.

Streaming would require MoA to commit to a winning answer earlier than 'we've reduced and serialized the JSON body' and then forward upstream tokens from that worker as they arrive.

## Proposed approaches

**A. Stream the sole-answer path.** When the grace early-exit or sole-survivor early-exit fires, the chosen worker's chat-completion chunk stream is already on the wire. Instead of buffering it into a chat-completion JSON and re-emitting, splice the upstream `chat.completion.chunk` events into the Responses-API translator (see `relay_translated_responses_stream` in `transport.rs`). User sees real tokens flow ~6s in.

**B. Speculative stream + roll-back on disagreement.** Pick a likely-winner worker early (e.g. the fastest to start producing tokens), stream their output to the client immediately, and if the arbiter later detects disagreement requiring a reducer, send a `response.output_text.done` for the speculation and then a new `response.output_text.delta` block from the reducer. Lossy but maximally responsive. Probably overkill v1.

**C. Stream only after the arbiter commits.** Once the arbiter has selected a winning worker (early-exit path), stream that worker's already-received content as small chunked deltas (`for chunk in split_into_chunks(content, ~5 tokens): emit delta`). Cheaper to implement than A; less honest (content is already buffered) but visually identical from the UI's perspective.

C is the easiest unblock, A is correct, B is interesting later.

## Out of scope

* Streaming through the reducer path (the reducer makes its own LLM call; that would need a separate streaming-aware reducer call).
* Mid-stream tool-call interception. For now agentic turns ship the way they do today.

## Acceptance criteria

* Sending `POST /v1/responses` with `{model: 'mesh', input: 'count up to ten', stream: true}` on a healthy 2-node mesh returns ≥ 5 `response.output_text.delta` events (not 1).
* The visible-rendering test in the chat UI shows incremental text appearing during MoA streaming (manual or via a Vitest with mocked SSE).
* Tool-calling path is unchanged (`finish_reason: tool_calls` still arrives correctly).
* The MoA arbitration + grace contract from #615 is preserved (chat-only grace, agentic always bypasses).

## Related

* PR #615 — chat redirect through MoA, sole-answer grace, Responses-API adapter (currently one-shot delta).
* Issue #617 — fast worker shouldn't think (separate failure mode in MoA).

michaelneale:
Closes #618.

MoA was emitting a single `response.output_text.delta` event with the entire arbitrated answer, making the chat UI feel like a long spinner followed by a hard pop-in. This PR splits the buffered content into ~24-char chunks (roughly 5 tokens of English text) and emits one delta per chunk so the UI renders incrementally.

This is **presentation-layer chunking**, not real worker-stream splicing. By the time MoA decides on a winner the content is already buffered, so the underlying timing is unchanged \u2014 the user-perceived feel just changes from "long spinner then pop" to "text flowing in". Issue #618's option C (cheapest). The bigger option A (splice the winning worker's chat.completion.chunk stream into the Responses-API SSE during arbitration) is left as a separate follow-up.

## Implementation

* New `chunk_for_streaming(&str) -> Vec<&str>` helper. Walks UTF-8 character boundaries, prefers to break at whitespace when that produces a reasonably-sized chunk, falls back to the full window when no whitespace is close. Lossless reassembly is a tested invariant.
* `send_moa_as_responses_sse` now emits one delta per chunk plus the existing created / output_text.done / completed / [DONE] events.
* Empty content still produces a well-formed end-of-stream so clients don't hang.

## Tests

7 new in `crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs`:

* 5 unit tests on `chunk_for_streaming`:
  - empty input \u2192 no chunks
  - short content \u2192 single chunk
  - long content \u2192 multiple chunks with lossless reassembly
  - multi-byte UTF-8 (\u00e9, \u4f60\u597d, \ud83d\udc4b, \u3053\u3093\u306b\u3061\u306f) \u2014 no codepoint splits, lossless reassembly
  - no-whitespace runs \u2014 falls back to char window
* 2 end-to-end SSE-shape tests via the existing TCP-loopback harness:
  - long content emits \u22653 deltas + exactly 1 done + 1 completed
  - empty content emits 0 deltas + still 1 done + 1 completed + [DONE]

`cargo test -p mesh-llm-host-runtime --lib`: 1477 pass (7 new).
`cargo clippy --all-targets -- -D warnings`: clean.
`cargo fmt --all -- --check`: clean.

## Compat

* No wire-protocol change. Same SSE event types (`response.output_text.delta`, etc.), just more of them.
* No mesh / skippy / plugin protocol change.
* Clients that already parse multi-delta streams (any spec-compliant Responses-API client) handle this correctly without modification.

## Live validation

To follow once the binary is rebuilt and pushed to the lab. The release-build is wired into the same lab that already validated PR #620 (opinionated MoA no-think). Once swapped in, the chat UI on M4 should feel like real streaming for `model=mesh` instead of one-shot pop.

### DIFF
--- a/MIC_LAB_NOTES.md
+++ b/MIC_LAB_NOTES.md
@@ -66,14 +66,128 @@ This is **net better** for the connection-pair: paths don't get torn down and re
 
 Takeaway: mini's VPN system extension is the most likely environmental cause. Asking the user to disable it would confirm. But the wild won't always cooperate — we need to be resilient regardless.
 
-## Topological resilience experiments still worth running
+## Topological resilience experiments — results from the v2 lab run
 
-1. **Direction reversal**: have mini initiate connections to M4 (mini becomes the proxy). Does the asymmetry flip with direction, or is it always M4→mini that's broken?
-2. **Key rotation**: delete `~/.mesh-llm/key` on either side and restart. Forces a fresh holepunch from scratch. Does the new connection pick different addrs / behave differently?
-3. **Disable mini's VPN system extension**: would confirm or rule out as the cause.
-4. **Capture iroh debug logs** during a flap and see exactly which addresses it tried (`endpoint=` / `path=` events at debug level).
+### Invite token contents (decoded base64)
 
-These are cheap to try in the lab but not in the wild. The point of the resilience work (same-target retry, sticky relay) is to be okay regardless.
+Each node advertises three addresses in its invite token: the iroh relay URL, its public IP/port via STUN-like discovery, and its LAN IP/port. Decoded:
+
+```
+M4     id=cf2d4edab1...  addrs=[relay, 180.181.228.108:36188, 192.168.86.172:56727]
+Mini   id=462bcc96fb...  addrs=[relay, 180.181.228.108:36119, 192.168.86.60:63108]
+Studio id=d0782f712e...  addrs=[relay, 180.181.228.108:0,     192.168.86.24:61252]
+```
+
+All three nodes are behind the same NAT (`180.181.228.108`, Australia Pacific Networks home router). Studio's public-IP port is `0` — meaning STUN-equivalent didn't return a port yet, or studio simply hasn't done the public-address discovery cycle. Both other nodes have specific ports.
+
+### LAN reachability is fine but jittery (zero loss, wide RTT range)
+
+Direct ICMP probes between nodes (all on the same 192.168.86.0/24 subnet via 192.168.86.1 router):
+
+| path | min | avg | max | stddev |
+|---|---|---|---|---|
+| M4 → studio | 7.5 | **22.9** | 64 | 21 |
+| studio → mini | 10.0 | **52.5** | 125 | 36 |
+| mini → studio | 7.8 | **33.0** | 107 | 37 |
+| M4 → mini | 13.3 | **55.6** | 110 | 29 |
+| mini → M4 | 7.7 | **28.3** | 113 | 33 |
+
+**Mini is the jittery node, in both directions.** Not an M4 firewall issue (mini↔studio is also jittery). Not direction-specific. It's mini's wifi link itself. Studio↔M4 (the one stable pair) is the only pair where neither end is mini.
+
+UDP "reachability" via `nc -uvz` succeeds in both directions, but that's misleading — UDP nc just sends a single datagram and exits without waiting for any acknowledgement. It tells us only that the OS didn't reject the send.
+
+### Why iroh's path teardown happens despite 0% loss
+
+iroh's path-quality estimator uses RTT to decide whether a path is healthy. When mini's wifi swings from 8ms to 110ms repeatedly, iroh sees a degraded path. After enough consecutive bad RTTs, multipath marks the path unhealthy, probes a new one, and abandons the old. When all known paths abandon simultaneously, you get `Connection to 462bcc96fb closed: no viable network path exists: last path abandoned, no new path opened`.
+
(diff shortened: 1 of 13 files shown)


michaelneale:
## Honest result on the lab: this PR does NOT improve TTFT

I just timed each delta arrival from a real lab MoA stream:

```
[  6122ms] response.created
[  6122ms] delta: 'The ocean stretches '
[  6122ms] delta: 'like a vast, heaving '
[  6122ms] delta: 'emerald sea, its '
[  6122ms] delta: 'surface fractured by '
\u2026 21 more deltas all at the same 6122-6123ms timestamp \u2026
[  6123ms] response.output_text.done
[  6123ms] response.completed
```

All 22 deltas arrive in a **1ms window**. That's because MoA still buffers the entire answer before chunking begins (option C from #618 is presentation chunking, not real streaming).

So this PR's actual benefit is **wire-shape correctness** \u2014 the SSE stream looks like a normal streaming response with multiple deltas instead of a degenerate single-delta stream \u2014 not perceived latency. The chat UI rendering them as they arrive will still see a single frame's worth of content arrive at t=6.1s.

For real TTFT improvement we need **option A from #618**: splice the winning worker's upstream chat.completion.chunk stream into the Responses-API SSE during arbitration, so the user sees tokens as the model produces them.

Recommendation: keep this PR as a small wire-shape correctness fix; open a follow-up for the proper option-A streaming work. Or close this PR and go straight to option A. Your call.

michaelneale:
## Recommendation: close this and do option A

After lab measurement and Mic's note ("how do we stream from worker when we don't know what we'll do with it?") I'm convinced this PR is **wire-shape correctness only** \u2014 zero perceived TTFT win because the chunks all arrive in a 1ms window at the client end (content is buffered before chunking).

The real fix is option A from #618: stream from the reducer (when arbiter routes to it) or from the winning worker (when arbiter picks one) by holding the upstream chat.completion.chunk stream open through arbitration. That's a ~100 LoC change in moa_gateway, and it's actually what users would feel.

I'll leave this PR open for now in case you want the wire-shape correctness anyway (multi-delta is the canonical Responses-API shape and some clients may render better with it). But I'd recommend closing and doing option A. Will start option A on a fresh branch unless you'd rather skip it.

michaelneale:
Closing per discussion: this is wire-shape-only and doesn't help TTFT \u2014 chunks all arrive in 1ms at the client end because content is buffered before chunking. Going to do option A (real worker-stream splicing from the reducer / winning worker) on a fresh branch instead. Issue #618 stays open for that work.

Issue #611, opened 2026-05-21
publish_crates job will fail on v0.66.0 GA: mesh-llm-client has path-only workspace deps


michaelneale:
## Symptom

The `publish_crates` job in `.github/workflows/release.yml` will fail on the next non-prerelease release tag (e.g. `v0.66.0` GA). The job runs `cargo publish --locked -p mesh-llm-client`, which will reject with:

```
error: all dependencies must have a version specified when publishing.
dependency mesh-llm-identity does not specify a version
```

The job has not failed in any release run since v0.65.1 because every release since then has been a prerelease tag (`v0.66.0-rc1` → `v0.66.0-rc7`), and the job is gated by:

```yaml
publish_crates:
  if: ${{ needs.metadata.outputs.prerelease != 'true' }}
```

So the breakage has been silently sitting on `main` since May 4. The first real (non-RC) release after that point — i.e. `v0.66.0` GA — is when it will surface.

## Root cause

At `v0.65.1` (May 4), `crates/mesh-client/Cargo.toml` had **zero workspace-internal path deps**. It was a pure leaf crate with crates.io deps only. `cargo publish -p mesh-llm-client` worked.

In `e6b723d1 mesh-llm crate decomposition (#459)` (also May 4), the host runtime was decomposed into peer crates (`mesh-llm-types`, `mesh-llm-protocol`, `mesh-llm-routing`, `mesh-llm-identity`, `model-artifact`, etc.). `mesh-client` grew workspace path dependencies as part of that refactor:

```toml
# crates/mesh-client/Cargo.toml on main today
mesh-llm-identity = { path = "../mesh-llm-identity" }
mesh-llm-protocol = { path = "../mesh-llm-protocol" }
mesh-llm-routing  = { path = "../mesh-llm-routing" }
mesh-llm-types    = { path = "../mesh-llm-types" }
model-artifact    = { path = "../model-artifact" }
```

None of them have `version =`. crates.io requires every dep entry on a published crate to be resolvable from the registry, which means `path = "..."` must be paired with `version = "..."`. `mesh-api/Cargo.toml` does this correctly:

```toml
mesh-client = { package = "mesh-llm-client", version = "0.66.0", path = "../mesh-client" }
```

but `mesh-client` itself does not.

## Impact

Released today (v0.66.0-rc7) — none, because RC tags skip `publish_crates`.

[...]

PR #622, opened 2026-05-21
Fix crates.io publish chain for SDK crates

IvGolovach:
## Summary

Prepares the Rust SDK crates.io publish path so the next non-prerelease GA tag can publish the full SDK dependency chain instead of failing after GitHub release artifacts are created.

- Adds versioned internal path dependencies for the crates published through `mesh-llm-client` and `mesh-api`.
- Makes `model-ref` and `model-artifact` publishable with crate metadata and versioned internal dependencies.
- Adds `scripts/publish-crates.sh` to publish the SDK crate chain in dependency order.
- Updates the release workflow to use the shared publish script for non-prerelease tags.
- Generalizes `scripts/release-version.sh` so future release bumps keep versioned path dependencies in sync.
- Documents the dry-run path and the first-publish registry constraint in `RELEASE.md`.

## Why

Issue #611 identified that `publish_crates` would fail on the next GA release because `mesh-llm-client` gained workspace-internal path dependencies during crate decomposition, but those dependencies were not publishable registry dependencies. The old release job only published `mesh-llm-client` and then `mesh-api`, which is not enough once the client crate depends on newly split internal crates.

## Publish Order

The publish script uses the dependency order needed by crates.io:

1. `model-ref`
2. `mesh-llm-identity`
3. `mesh-llm-protocol`
4. `mesh-llm-routing`
5. `mesh-llm-types`
6. `model-artifact`
7. `mesh-llm-client`
8. `mesh-api`

Real publishing still requires `CARGO_REGISTRY_TOKEN` and keeps a crates.io propagation wait between packages. The dry-run path avoids uploads and reports downstream crates that cannot be fully dry-run until their new upstream package versions exist in the registry.

## Validation

- `git diff --check`
- `git diff --cached --check`
- `bash -n scripts/publish-crates.sh`
- `bash -n scripts/release-version.sh`
- `scripts/publish-crates.sh --allow-dirty`: rejected as expected because `--allow-dirty` requires `--dry-run`
- `CARGO_TARGET_DIR=$(mktemp -d /tmp/mesh-llm-publish-target.XXXXXX) scripts/publish-crates.sh --dry-run --allow-dirty`
- `cargo metadata --locked --format-version 1 --no-deps`
- `cargo check -p mesh-llm-client -p mesh-api`
- `cargo test -p model-ref --lib`: 10 passed
- `cargo test -p model-artifact --lib`: 23 passed

Not run:

- `actionlint`: not installed locally.
- Real `cargo publish` / `just release`: intentionally not run for PR validation.

## Rollback

Revert this PR. No runtime migration, data repair, or protocol rollback is required.

Related: #611

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -528,11 +528,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: dtolnay/rust-toolchain@stable
-      - name: Publish mesh-llm-client
+      - name: Publish crates.io package chain
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        run: cargo publish --locked -p mesh-llm-client
-      - name: Publish mesh-api
-        env:
-          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        run: cargo publish --locked -p mesh-api
+        run: scripts/publish-crates.sh
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -68,3 +68,16 @@ Verify:
 ## Publish
 
 Push a `v*` tag to run `.github/workflows/release.yml`.
+
+On non-prerelease tags, the release workflow also publishes the Rust SDK crate
+chain to crates.io in dependency order:
+
+```bash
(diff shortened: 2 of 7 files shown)


i386 (approved):
LGTM

PR #624, opened 2026-05-21
fix(moa): grace fires on diverse fast answers, not just sole answer

michaelneale:
## Real lab failure

When studio AND mini both joined the public mesh today (so MoA now had 5 workers: Qwen2.5-3B, Qwen3-8B, Qwen3.5-9B, MiniMax, Qwen3-32B), `model=mesh` latency DEGRADED:

| Run | mesh time | what happened |
|----:|----------:|---|
| 1 | 45.5s | waited for slow Qwen3-8B |
| 2 | 36.0s | waited for slow Qwen3-8B |
| 3 | 38.2s | waited for slow Qwen3-8B |
| 4 | **0.7s** | 2 workers textually agreed \u2014 consensus fired |
| 5 | 44.9s | waited for slow Qwen3-8B |

Looking at the gateway log: Qwen2.5-3B answered in 84ms, Qwen3.5-9B in 670ms, MiniMax in 1011ms. **All three fast answers in hand within 1 second.** But because their content didn't textually agree ("Hello" / "Yes" / "Ready" / "Okay"), the arbiter's consensus rule ("\u22652 workers agree on answer") didn't fire. MoA then sat there for 30-45 seconds waiting for the tail worker (Qwen3-8B \u2014 the known-bimodal public-mesh peer).

## The bug

The previous grace eligibility was:
```rust
answers.len() == 1 && answers[0].confidence >= GRACE_MIN_CONFIDENCE
```

That covers the case where one worker has answered and the rest are still pending. It does NOT cover the case where multiple workers have answered fast but their answers disagree.

## Fix

Relax eligibility to "\u22651 Answer-kind output with confidence \u2265 0.5". When grace fires with multiple qualifying answers, pick the highest-confidence one. The previous sole-answer case is preserved trivially (max_by on a single element returns that element).

```rust
let grace_eligible = |outs: &[WorkerOutput]| -> bool {
    if !grace_enabled { return false; }
    outs.iter().any(|o| {
        o.kind == normalize::OutputKind::Answer
            && o.confidence >= GRACE_MIN_CONFIDENCE
    })
};
```

When grace fires:
```rust
let answer = outputs
    .iter()
    .filter(|o| o.kind == normalize::OutputKind::Answer)
    .max_by(|a, b| a.confidence.partial_cmp(&b.confidence)
        .unwrap_or(std::cmp::Ordering::Equal))
    .expect("grace_eligible guaranteed at least one Answer")
    .payload
    .clone();
```

## Tests

[...]

PR #626, opened 2026-05-21
tune(moa): tighten first_answer_grace from 6s to 3s

michaelneale:
## What it does

Lowers the chat-mode `first_answer_grace` default from 6 seconds to 3 seconds. This is a one-line tuning change. Built on top of #624 (which makes grace fire on diverse fast answers \u2014 needed for the chat case where workers don't textually agree).

## Why now

On the public mesh today (5 workers including studio MiniMax, mini Qwen3.5, etc.):
* **MoA consensus rarely fires for short chat answers** \u2014 workers respond with diverse short text like "Hello" / "Yes" / "Ready" / "Okay" that don't textually match.
* PR #624's diverse-answer grace catches this case: when grace expires with \u22651 qualifying answer, pick the highest-confidence one.
* But the 6s grace becomes the chat latency floor \u2014 every turn pays the full 6s.

With #624 landed and the timer being the dominant chat path, **3s gives 2\u00d7 better p50 with no measurable quality regression on short chat prompts**.

## Lab data, public mesh, 10 varied turns

| Prompt type | Result |
|---|---|
| "Symbol for gold?" | 2x: "Au, from Latin aurum\u2026" \u2705 |
| "WW2 end?" | 2x: "1945" \u2705 |
| "Coffee shop name?" | "Brewed" / "Perk" \u2705 |
| "60mph \u00d7 2h?" | 2x: "120" \u2705 |
| "5-word rain sentence" | "Rain drums gently on the roof" / "Rain pours down in relentless sheets" \u2014 reasonable but 6 words not 5 (same loose constraint behavior the model has at 6s grace too) |

Latency distribution:

| grace | min | p50 | max | n |
|---|---:|---:|---:|---:|
| **6s** (current) | 700ms | ~6000ms | 6000ms | 5 |
| **3s** (proposed) | 1299ms | 2002ms | 2007ms | 10 |

## Trade-off

A worker that takes 3\u20136s to land is now excluded from consensus. On the public mesh post-#624 we found textual consensus rarely fires for chat anyway, so this isn't losing a path that was actually firing.

For agentic / tool turns (`has_tools=true`), grace is bypassed entirely and consensus continues to work as before. The 3s number is **chat-only**.

## Compat

* No API change. The default value moves, callers that explicitly build a `GatewayConfig` keep their configured value.
* No protocol / ABI / plugin change.
* No mesh-version compatibility concerns.

## Tests

cargo test -p mesh-llm-host-runtime --lib: 1463 pass.
cargo clippy --all-targets -- -D warnings: clean.
cargo fmt --all -- --check: clean.

The grace eligibility logic itself (which determines when the timer arms) is covered by PR #624's tests. This PR is strictly a config tuning, no behavior-shape change.

## Open question

3s is empirical. A more conservative tuning would be 4-5s. A more aggressive one would be 2s (tested in lab, all 10 turns sub-2.5s but tight enough that any worker with 2s+ network latency is excluded). Happy to land at any number you prefer.

### DIFF
--- a/MIC_LAB_TODO.md
+++ b/MIC_LAB_TODO.md
@@ -1,0 +1,15 @@
+
+## After PR #624 lab measurement — strategy refresh
+
+* **Chat path is now grace-bounded.** `mesh_chat` p50 floor on the public
+  mesh is ~6s = the `first_answer_grace` config. To make chat feel
+  faster, lower the grace (1.5-2s), trade off: more likely to take a
+  first-answer that might be lower-quality than what a longer wait
+  would give. **Cheap experiment in the lab.**
+* **Reducer path is where option A matters.** On the public mesh
+  post-fix, reducer fires 0% of chat turns (grace catches everything).
+  On lab tool turns it fires 36% of the time. Streaming would help
+  tool turns feel responsive.
+* **Tool turns currently buffer reducer output.** Real streaming would
+  show the reducer's tokens as it produces them. ~150 LoC.
+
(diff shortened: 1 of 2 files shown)


michaelneale:
Rolled into #624 \u2014 the eligibility relaxation and the tighter default are the same user-journey change and only meaningful together. Easier to review as one PR.

PR #623, opened 2026-05-21
Add an agent tool-call reliability QA probe

IvGolovach:
## Summary

Adds a lightweight QA probe that lets maintainers validate whether a live mesh endpoint can complete agent-style tool-call loops through the OpenAI-compatible chat-completions API.

- Adds `scripts/qa-agent-tool-call-reliability.py` for deterministic tool-call and tool-result continuation checks.
- Covers both non-streaming and streaming chat-completions, including `finish_reason=tool_calls` and streamed `delta.tool_calls[*]` reconstruction.
- Emits JSONL results plus a side-effect-free JSON `--print-plan`.
- Adds Python unit coverage, a Rust contract test for the script surface, and docs for agent/testing workflows.

## Why

Recent agent and MoA testing made tool-call reliability a practical baseline risk. The existing Goose, OpenCode, and Pi smokes are useful end-to-end checks, but they are heavier and can make raw API-contract regressions harder to isolate. This gives a small direct probe for the `tools -> tool_calls -> tool result -> final answer` path those agents depend on.

## Diff Scope

- `scripts/qa-agent-tool-call-reliability.py`
- `scripts/tests/test_qa_agent_tool_call_reliability.py`
- `crates/mesh-llm/tests/qa_agent_tool_call_reliability.rs`
- `docs/AGENTS.md`
- `docs/design/TESTING.md`

## Compatibility

Opt-in verification tooling only. No runtime behavior, mesh protocol, gossip fields, API schema, release artifacts, or CI requirements are changed.

## Branch Integrity

- Base branch: `main`
- Validated base SHA: `27d46a6c4893f6ef0bc9b17591f85179e5e80b0d`
- Ahead/behind: `0 behind / 1 ahead`
- Merge base: `27d46a6c4893f6ef0bc9b17591f85179e5e80b0d`
- Diff proof computed against freshly fetched `origin/main`.

## Commit Integrity

- `a5ceac7d Add agent tool-call reliability harness`
- One logical tooling/docs commit.
- No forbidden files, build artifacts, caches, secrets, credentials, or local environment files included.

## Diff Hygiene

- `git diff --check origin/main...HEAD`: PASS, no output
- `git diff --check`: PASS, no output
- `git diff --cached --check`: PASS, no output before commit

## Validation

Validation tier: Tier 4 - verification tooling plus docs and contract tests for the OpenAI-compatible agent tool-call QA path.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS
[...]

PR #629, opened 2026-05-22
feat(moa): no-think default on workers + grace fires on diverse fast answers

### LINKED ISSUE #617: MoA: fast worker shouldn't think — thinking models can poison fan-out and leak reasoning to chat

michaelneale:
Combines the two MoA chat improvements from PRs #620 and #624 into one PR \u2014 they're inseparable in the user journey and only meaningful together. Closing those in favour of this.

## What gives

On the public mesh today, with the chat UI's `Auto` routing through `model=mesh` (MoA), the dominant chat experience is:

| | before this PR | after this PR |
|---|---:|---:|
| **p50 latency (public mesh, short chat)** | **~40s** | **~1.2s** |
| think leakage into chat (Qwen3-8B, MiniMax, etc. emitting raw `<think>...` prose) | yes, frequent | no |
| answers correct on factual / arithmetic / subjective prompts | yes | yes |
| tool calling | works, unchanged | works, unchanged |

Validated live on a real public mesh (M4 client `--auto`, joined to public mesh, hitting 20+ peers including studio MiniMax and others).

### 1. Opinionated no-think on MoA workers + reducer (was #620)

Workers in MoA are short-budget internal slots, not user-facing reasoning steps. The fast worker has a 256-token budget that doesn't fit `<think>\u2026</think>` + answer. The reducer doesn't want reasoning prose as candidate input. The chat UI's `Auto` routes everything through MoA, so reasoning models burning their budget inside `<think>` was the dominant chat failure mode.

This PR:
* `SamplingParams` gains `enable_thinking: Option<bool>` + `with_thinking(...)`.
* New `backend::apply_enable_thinking` injects `chat_template_kwargs.enable_thinking` (canonical llama.cpp knob) + `reasoning_effort: "none"` when off.
* All three backends (HTTP / LocalModel / RemoteModel) call it.
* New `effective_enable_thinking_for_moa(body)` policy: `extract_enable_thinking_override(body).or(Some(false))` \u2014 opinionated default off, escape hatch for callers who explicitly pass `reasoning_effort: "low"` etc.

### 2. Grace fires on diverse fast answers (was #624)

The arbiter's consensus rule ("\u22652 workers textually agree") rarely fires on short chat answers across many models. "Hello" / "Yes" / "Ready" / "Okay" are all valid but don't match. Before this PR, MoA then waited for the slow tail worker (~40s on the public mesh's bimodal Qwen3-8B peer).

The old sole-answer grace only armed on `outputs.len() == 1`. This PR relaxes it to "at least one Answer-kind output with confidence \u2265 0.5". When grace fires with multiple answers, pick the highest-confidence one.

Also tightens `first_answer_grace` from 6s to 3s. The previous 6s was conservative because grace rarely fired \u2014 with the eligibility relaxation, grace is the dominant chat path, so a tighter default is right.

## Lab data

Public mesh, 5 runs of "reply with one short word":

| | grace=6s old eligibility (before) | grace=3s new eligibility (this PR) |
|---|---:|---:|
| Run 1 | 45.5s | 3.0s |
| Run 2 | 36.0s | 3.0s |
| Run 3 | 38.2s | 3.0s |
| Run 4 |  0.7s | 1.0s |
| Run 5 | 44.9s | 1.0s |

Varied prompts post-fix:

| Prompt | Time | Answer |
|---|---:|---|
| "What is 2+2?" | 877ms | "2+2 equals 4." |
| "Capital of Japan?" | 1.2s | "Tokyo is the capital of Japan." |
| "Name one Python web framework." | 3.0s | "Flask" |
| "Hello" | 1.0s | "Hello! How can I assist you today?" |
| "What color is the sky?" | 3.0s | "The color of the sky can vary\u2026" |

Tool calling (`mesh` + `tools=[read_file]`): 2/2 success in 2.8\u20135.1s, `finish_reason: tool_calls`, no change from before.

## What this does NOT change

* No QUIC, iroh, keep-alive, relay, discovery, gossip, target_health, or any mesh-network primitive. Purely MoA application logic.
* No mesh wire-protocol change. Backends receive a chat-completion request body with an extra additive field (`chat_template_kwargs.enable_thinking`), which llama.cpp templates already understand. Backwards compatible across mixed-version meshes.
* No skippy ABI change. No plugin protocol change.

## Tests

* mesh-mixture-of-agents lib: 100 pass (10+ new for `apply_enable_thinking`, `SamplingParams::with_thinking`, and grace eligibility/behavior).
* mesh-mixture-of-agents `tests/sim_enable_thinking_propagation.rs` (new): 3 mock-backend integration tests pinning `chat_template_kwargs.enable_thinking` reaches every worker AND no spurious fields appear when no override is requested.
* mesh-llm-host-runtime lib: 1477 pass (14 new for `extract_enable_thinking_override` covering every JSON shape, plus 4 for `effective_enable_thinking_for_moa`).
* cargo clippy --all-targets -- -D warnings: clean.
* cargo fmt --all -- --check: clean.

## Compat

* Behavior change for existing MoA callers: a caller who today relies on MoA producing reasoning output will now get short answers instead. Escape hatch via any recognised reasoning knob (`reasoning_effort: "low"`, `enable_thinking: true`, `reasoning: { enabled: true }`, etc.).
* Chat UI's `Auto` (which routes through MoA per #615) is the main consumer and benefits directly.
* Tool turns continue to bypass grace via `!has_tools` gate \u2014 agentic harnesses unaffected.

Closes #617.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs
@@ -371,8 +371,27 @@ pub async fn build_moa_config(
         // (or sooner on outright failure). Cheap on the happy path, big win on
         // the cold-KV / stale-peer tail.
         hedge_delay: std::time::Duration::from_secs(5),
-        // Chat-only sole-answer grace. Tool turns ignore this.
-        first_answer_grace: std::time::Duration::from_secs(6),
+        // Chat-only grace: after this long since dispatch, if at least
+        // one qualifying Answer is in hand we ship the highest-confidence
+        // one. Tool turns bypass this entirely (consensus continues to
+        // arbitrate tool proposals).
+        //
+        // 3 seconds is empirically good across the public mesh today.
+        // Long enough that slow-but-good workers (studio MiniMax
+        // landing at ~1s, mini Qwen3.5 at ~700ms) finish before the
+        // timer; short enough that chat doesn't sit on a multi-second
+        // ceiling on every turn. Lab data: median mesh_chat dropped
+        // from ~6s (old default) to ~2s with this value, no quality
+        // regression measured on factual / arithmetic / short-creative
+        // prompts.
+        //
+        // The previous 6s was conservative because the original grace
+        // logic only armed on a sole answer — it had to wait for a
+        // second non-matching answer to arrive before becoming useless.
+        // With the relaxed eligibility added in this change, the timer
+        // is the dominant chat path, so a tighter default is the right
+        // default.
+        first_answer_grace: std::time::Duration::from_secs(3),
         // Defaults to leaving each model's thinking behavior alone.
         // `try_handle_moa` overrides this from the inbound request body
         // when the caller has expressed a preference
--- a/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs
@@ -48,13 +48,83 @@ pub async fn try_handle_moa(
         return None;
     };
 
-    let Some(config) = build_moa_config(node, targets).await else {
+    // Pull the caller's reasoning preference off the JSON body before we
+    // hand it to MoA. The body itself gets forwarded to workers minus the
+    // `stream` flag, but we need this knob at the gateway level so it
+    // reaches every worker AND the reducer (workers and the reducer pick
+    // up via `GatewayConfig::enable_thinking`).
+    let enable_thinking = extract_enable_thinking_override(&body_json);
+
+    let Some(mut config) = build_moa_config(node, targets).await else {
         let _ = proxy::send_503(tcp_stream, "MoA requires ≥2 models available in the mesh").await;
         return None;
     };
+    config.enable_thinking = enable_thinking;
 
     run_moa_turn(tcp_stream, body_json, &config, request.response_adapter).await;
     None
+}
+
+/// Pull the caller's "disable / enable thinking" preference out of an
+/// inbound chat-completion or responses JSON body. Mirrors the same
+/// shapes that `openai_frontend::common::normalize_reasoning_template_options`
+/// recognises so MoA users get the same surface as direct callers.
+///
+/// Recognised inputs (any one is enough):
+/// * `reasoning_effort: "none"` (off) or any non-`"none"` value (on)
+/// * `reasoning: { enabled: false }` (off) / `{ enabled: true }` (on)
+/// * `reasoning: { effort: "none" }` / `{ max_tokens: 0 }` (off)
+/// * Any of `THINKING_BOOLEAN_ALIASES` as a top-level field with bool
+/// * `thinking_budget: 0` (off)
+/// * `chat_template_kwargs.enable_thinking` (or any alias) as bool
+///
+/// Returns `None` when the caller hasn't expressed a preference, leaving
+/// each worker's default behavior alone.
+fn extract_enable_thinking_override(body: &serde_json::Value) -> Option<bool> {
+    let obj = body.as_object()?;
+    let mut result: Option<bool> = None;
+
+    // reasoning: { enabled, effort, max_tokens }
+    if let Some(r) = obj.get("reasoning").and_then(|v| v.as_object()) {
+        if r.get("enabled") == Some(&serde_json::Value::Bool(false))
+            || r.get("effort").and_then(|v| v.as_str()) == Some("none")
+            || r.get("max_tokens").and_then(|v| v.as_u64()) == Some(0)
+        {
+            result = Some(false);
+        } else if r.get("enabled") == Some(&serde_json::Value::Bool(true))
+            || r.get("effort").is_some()
+            || r.get("max_tokens").is_some()
+        {
+            result = Some(true);
+        }
+    }
+
+    // reasoning_effort: "none" / "low" / etc.
+    if let Some(effort) = obj.get("reasoning_effort").and_then(|v| v.as_str()) {
+        result = Some(effort != "none");
+    }
+
+    // Top-level boolean aliases (enable_thinking, enable_reasoning, etc.).
+    for alias in openai_frontend::common::THINKING_BOOLEAN_ALIASES {
+        if let Some(b) = obj.get(*alias).and_then(|v| v.as_bool()) {
+            result = Some(b);
+        }
+    }
+
+    if obj.get("thinking_budget").and_then(|v| v.as_u64()) == Some(0) {
+        result = Some(false);
+    }
+
+    // chat_template_kwargs.{enable_thinking, ...}
+    if let Some(kwargs) = obj.get("chat_template_kwargs").and_then(|v| v.as_object()) {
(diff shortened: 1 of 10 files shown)


michaelneale:
Merged latest main (incl. Ivan's #623 tool-call reliability probe) and addressed Copilot review:

**Fix 1 (real bug)**: `apply_enable_thinking` would silently drop the flag if `chat_template_kwargs` was already present as a non-object (string / array / null / number) because `entry().or_insert_with()` doesn't replace an existing non-object, leaving `as_object_mut()` to return None and the override to be lost. We now normalize a bogus shape to `{}` before injecting. New test `apply_enable_thinking_normalizes_non_object_kwargs` pins all four bogus-input shapes.

**Fix 2 (doc clarity)**: `effective_enable_thinking_for_moa` and `extract_enable_thinking_override` had their doc comments merged together because there was no item between them. The merged doc was internally inconsistent (one half about each function). Split into one doc per function and clarified `effective_enable_thinking_for_moa` always returns `Some`.

**Skipped (intentional design)**: Copilot also suggested 400-erroring on invalid types for the top-level thinking aliases (non-bool values like `enable_thinking: "yes"`). I left this alone:

* MoA's `enable_thinking` is a *default-override*, not a primary input. Treating an unparseable hint as "use the MoA default" is the safer-fail mode and matches the spirit of an opinionated-by-default surface.
* Strict shape validation of caller bodies is the OpenAI-frontend's responsibility, enforced upstream of MoA in `openai_frontend::common::normalize_reasoning_template_options`.
* MoA's extractor is best-effort recognition of caller intent across the multiple shapes the OpenAI / responses surfaces use; rejecting malformed shapes here would surface MoA-specific 400s for requests the upstream frontend already accepted, which is a worse UX than gracefully falling back to the default.

Happy to revisit if you'd rather have symmetric strictness across both paths.

### Lab re-validation post-fix on public mesh (20+ peers)

Ivan's #623 probe (regression net):
```
agent tool-call reliability: 4/4 phases passed
PASS model=mesh attempt=1 phase=tool_call: lookup_fixture_fact key=codeword
PASS model=mesh attempt=1 phase=tool_result: final answer included signal-7429
PASS model=mesh attempt=1 phase=stream_tool_call: lookup_fixture_fact key=codeword
PASS model=mesh attempt=1 phase=stream_tool_result: final answer included signal-7429
```

Chat smoke (no-think + grace=3s):
| Prompt | Latency | Answer |
|---|---:|---|
| "What is 2+2?" | 858ms | "2+2 equals 4." |
| "Capital of Japan?" | 1.7s | "Tokyo is the capital of Japan." |
| "Hello" | 1.3s | "Hello! How can I help you today?" |
| "Name one Python framework." | 771ms | "One popular Python framework is Django." |
| "What color is the sky?" | 3.0s | "The color of the sky can vary\u2026" |

All sub-3.1s, no `<think>` leak, all correct.

Tests: 1477 host-runtime lib + 101 MoA lib (+1 new test for fix #1) all pass. clippy + fmt clean.

PR #600, opened 2026-05-20
Advertise peer throughput hints for routing

IvGolovach:
## Summary

Mesh peers can now share bounded observed tokens-per-second hints for hosted models through gossip, so routing can make a better choice when multiple targets are otherwise viable.

- Adds an additive `advertised_model_throughput` gossip/protobuf field for per-model throughput hints.
- Publishes only sanitized local serving observations for currently hosted models; remote and endpoint observations are not advertised as this node's own speed.
- Uses peer hints as an advisory routing signal after context-capacity filtering, with local observations weighted ahead of gossip.
- Adds tests for local hint generation, gossip/protobuf round-trip, peer refresh, routing order, and v0 compatibility.

## Why

This addresses #596. Routing already has local outcome memory, but peers could not advertise observed serving speed to the rest of the mesh. The new hint gives the router a bounded, low-risk signal for choosing between otherwise compatible peers without changing the hard routing constraints.

## Protocol / Compatibility

- The protobuf change is additive: `PeerAnnouncement.advertised_model_throughput` uses new tag 44.
- Older protobuf peers should ignore the unknown field, and the legacy v0 JSON compatibility path is unchanged.
- Values are fixed-point milli tokens/sec to avoid floating-point wire edge cases.
- Hints are capped, sanitized, deduplicated by model, and advisory only.
- No ALPN, plugin protocol, or Skippy ABI change is introduced.

## Diff Scope

- Host runtime routing metrics and OpenAI target ordering.
- Mesh gossip peer announcements and peer state refresh.
- Internal protocol conversion plus protobuf mirror for the additive field.
- Focused test fixture updates for the new announcement field.

## Branch Integrity

- Base branch: `main`
- Validated base SHA: `cb9472ee26d2aa356b4c276463a5491be06bb8bf`
- Ahead/behind against fetched `origin/main`: `0 behind / 1 ahead`
- Merge base: `cb9472ee26d2aa356b4c276463a5491be06bb8bf`
- This is now a standalone PR against `main`, not stacked on #566.

## Validation

Local paths in `LLAMA_STAGE_BUILD_DIR` are normalized to `<repo>` below.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS
- `git diff --check`: PASS, no output
- `git diff --cached --check`: PASS, no output
- `cargo fmt --all -- --check`: PASS
[...]

PR #639, opened 2026-05-22
fix(auto): fall back to next-best model when top pick's peers all cool

### LINKED ISSUE #625: auto-router: model-level fallback when top-pick model has no eligible peers
## Problem

On the public mesh, `model=auto` returns 503 deterministically when its top-pick model has only one peer and that peer is broken. Observed in lab today:

- Auto-router classified `Chat/Quick` prompts and picked `Qwen3-32B`.
- The single peer serving `Qwen3-32B` returns `malformed HTTP response` in ~130ms.
- Auto router 503s every single call.
- After studio (publishing MiniMax) joined, auto routed to MiniMax instead, and it works.

So auto IS reactive to availability when a competent peer exists for the chosen model. But when the chosen model has only broken peers, it has no model-level fallback.

## Root cause

Two things:

1. **`target_health::eligible_candidates` short-circuits when only one candidate exists.** Look at the top of the function:
   ```rust
   if candidates.len() <= 1 {
       return candidates.to_vec();
   }
   ```
   This bypasses the cooldown check entirely for single-candidate models. The 30s base cooldown / 5min max never kicks in for Qwen3-32B (one broken peer).

2. **The auto-router has no model-level fallback.** Even if `eligible_candidates` returned empty for a model, auto wouldn't try a different model for the same prompt class.

## Proposed fix

Two complementary changes:

### A. Apply cooldown to single-candidate models too

```rust
// in eligible_candidates:
// Don't short-circuit on candidates.len() <= 1.
// Always run the cooldown filter; let the caller handle "no eligible
// candidates" by considering the model unreachable.
```

This alone doesn't help unless (B) lands too — without a fallback, returning empty just means an earlier 503.

### B. Auto-router candidate ladder

When `auto` resolves a prompt to model A and `eligible_candidates(model_a)` returns empty, fall through to model B (next-best by classification ranking).

Roughly:
```rust
let candidates = classify_prompt(prompt)
    .into_iter()                            // ordered: best → worst
    .map(|model| (model, affinity.eligible_candidates(&model, &peers_for(&model))))
    .filter(|(_, peers)| !peers.is_empty())
    .next();
```

## Real-world impact

- Public mesh today: would route past the broken Qwen3-32B peer to Qwen2.5-3B (working) or Qwen3-8B (mostly working). Probably 95%+ improvement in `auto` success rate on chat prompts.
- Independent of #596 (gossiped tok/s) — we already track per-peer failure outcomes.
- Doesn't change the `mesh` (MoA) path, which already handles this via fan-out.

## Acceptance criteria

- `model=auto` returns a successful response when *some* peer in the mesh can serve a prompt of the requested class, regardless of which specific model the heuristic prefers.
- Existing behavior preserved when the top-pick model is healthy.
- Unit tests for: single broken candidate cooldown, multi-model ladder fall-through, restoration after cooldown expires.

## Out of scope here

- Per-peer tok/s ranking (see #596).
- Racing top-N candidates (a more aggressive strategy, would benefit from #596).
- MoA changes.

michaelneale:
## What

`model=auto` no longer deterministically 503s when the classifier's top-pick model happens to have only broken peers. It now walks an ordered ladder of candidate models and picks the first one with at least one non-cooling peer.

## Symptom this fixes

On the public mesh today (verified during testing):

```
auto[?] "what is 2+2" -> Qwen3-32B (sole peer 1b7af7fa, 503 always)
```

Auto picks Qwen3-32B, the only peer serving it returns malformed HTTP, and the request 503s back to the caller \u2014 even though MiniMax, Qwen3.5-9B, Qwen3-8B and others would have served the same prompt class fine.

After this PR, the auto router skips the model entirely when its peers are all cooling and falls through to the next-best classifier pick.

### 1. `target_health::strict_eligible_candidates`

Existing `eligible_candidates` short-circuits when there's only one candidate and falls back to "include all cooling targets anyway" when every candidate is cool. Both of those preserve availability for callers that have nowhere else to go \u2014 we keep them as-is.

But the auto-router *does* have somewhere else to go: a different model. So this adds a sibling that exposes the *true* eligible set, even if empty, and never short-circuits on single-candidate inputs. The auto-router uses this to decide whether a model is worth picking.

### 2. `router::pick_model_ladder_classified`

Returns an ordered ladder (default depth 4) instead of a single `Option<&str>`. Top entry uses the same weighted draw as the existing `pick_model_classified` (so existing behavior at the head is unchanged); subsequent entries are weighted draws without replacement from the remaining big-tier candidates, then small-tier fallbacks.

### 3. `resolve_auto_model_request` wiring

Walks the ladder, picking the first model with a non-empty `strict_eligible_candidates` set. Falls back to the raw top pick when every ladder entry is cooling, preserving the existing route-layer availability fallback.

## Tests

`cargo test -p mesh-llm-host-runtime --lib`: **1497 passed, 0 failed** (8 new).

* `network::router::tests::ladder_*` (4): empty input, single candidate, all-distinct exhaustion, big-before-small tier order, max_len enforcement.
* `network::target_health::tests::strict_*` (4): sole-candidate cool, partial cool, all healthy, all cool.

`cargo clippy -p mesh-llm-host-runtime --lib --all-targets`: clean.
`cargo fmt --all -- --check`: clean.

## Live validation

Built locally and joined a fresh client to public mesh during testing. Verified:
* The classifier still picks the same top model in 100% of cases that were working before (no regression on the head).
* Direct `/v1/chat/completions` to a broken `Qwen3-32B` peer continues to 503 / 429 as expected (this PR is auto-only \u2014 explicit model requests are unchanged).
* Could not trigger the fallback path observably during testing because MiniMax-M2.5 was always the top weighted pick on the current public mesh; the unit tests cover the fallback logic directly.

The PR will be most visible when the public mesh has more single-peer models that break (which is exactly when it'll matter most).

## Compatibility

* Local-only routing change. No mesh gossip, protobuf, plugin protocol, or public API change.
* Mixed-version meshes unaffected \u2014 this is a client-side decision about which model to send to.
* Backward compatible at the route layer: when every ladder entry is cooling, the existing `eligible_candidates` fallback (include cooling anyway, preserve availability) still kicks in.

## Out of scope

* Per-peer tok/s ranking across alternative models (#600 plumbing is now there but the data is sparse until 0.66+ adoption rises).
* Racing top-N candidates in parallel \u2014 more aggressive, can re-evaluate once #600 throughput data is denser.
* MoA `mesh` virtual model already handles this via its own fanout; no change needed there.

## Refs

Closes #625.
Builds on #577 (cooldown plumbing) and #600 (throughput-aware tiebreak in the weighted draw).

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/affinity.rs
+++ b/crates/mesh-llm-host-runtime/src/network/affinity.rs
@@ -130,6 +130,18 @@ impl AffinityRouter {
         self.target_health.eligible_candidates(model, candidates)
     }
 
+    /// Strict view of which candidates are not in cooldown. Empty when
+    /// every candidate is currently cooling. Used by the auto-router to
+    /// skip a model entirely in favor of the next-best alternative.
+    pub(crate) fn strict_eligible_candidates(
+        &self,
+        model: &str,
+        candidates: &[election::InferenceTarget],
+    ) -> Vec<election::InferenceTarget> {
+        self.target_health
+            .strict_eligible_candidates(model, candidates)
+    }
+
     pub(crate) fn record_target_outcome(
         &self,
         model: Option<&str>,
--- a/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
@@ -3335,7 +3335,7 @@ async fn resolve_auto_model_request(
     let Some(available) = router::filter_media_compatible_candidates(&with_caps, &media) else {
         return AutoModelResolution::UnsupportedMedia;
     };
-    let picked = router::pick_model_classified(&cl, &available).map(str::to_string);
+    let picked = pick_model_with_fallback(node, affinity, &cl, &available).await;
     if let Some(name) = picked.as_deref() {
         tracing::info!(
             "router: {:?}/{:?} tools={} media={} → {name}",
@@ -3349,6 +3349,51 @@ async fn resolve_auto_model_request(
         }
     }
     AutoModelResolution::Model(picked)
+}
+
+/// Walk the classifier's preferred ladder, skipping any model whose only
+/// peers are currently in target-health cooldown. Falls back to the raw
+/// top pick when every ladder entry is cooling (the existing
+/// availability semantics in `route_eligible_candidates` will then
+/// retry the cooling peer).
+const AUTO_LADDER_DEPTH: usize = 4;
+
+async fn pick_model_with_fallback(
+    node: &mesh::Node,
+    affinity: &AffinityRouter,
+    classification: &router::Classification,
+    available: &[router::RoutingCandidate<'_>],
+) -> Option<String> {
+    let ladder = router::pick_model_ladder_classified(classification, available, AUTO_LADDER_DEPTH);
+    if ladder.is_empty() {
+        return None;
+    }
+
+    for name in &ladder {
+        let hosts = node.hosts_for_model(name).await;
+        if hosts.is_empty() {
+            continue;
+        }
+        let candidates: Vec<election::InferenceTarget> = hosts
+            .into_iter()
+            .map(election::InferenceTarget::Remote)
+            .collect();
+        if !affinity
+            .strict_eligible_candidates(name, &candidates)
+            .is_empty()
+        {
+            return Some((*name).to_string());
+        }
+        tracing::debug!(
+            "router: skipping {name}, all hosts in cooldown ({} candidate(s))",
+            candidates.len(),
+        );
+    }
+
+    // Every ladder model is cooling; fall back to the top pick. The
+    // downstream `route_eligible_candidates` keeps availability by
+    // sending to a cooling peer rather than failing closed.
+    Some(ladder[0].to_string())
 }
 
 async fn lookup_cached_auto_model(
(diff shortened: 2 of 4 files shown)


ndizazzo (changes requested):
My PR testing harness caught:

---

**Severity: Medium**
Cached model=auto sessions can bypass PR #639’s new cooldown-aware fallback.

**What happens**
- resolve_auto_model_request checks lookup_cached_auto_model(...) before calling the new fallback picker.
- On cache hit, it returns the cached model immediately.
- That cache path only rejects the cached model if it is no longer served or cannot satisfy media requirements.
- It does not reject the cached model when all targets for that model are currently cooling down.

**Failure scenario**
1. A sticky model=auto request is cached to model A via prompt_cache_key, user, or session_id.
2. Model A’s only peer times out or returns retryable unavailable.
3. Target health marks that peer as cooling.
4. Model B is healthy and available.
5. The next request in the same session hits the auto-model cache.
6. The router returns model A before running the cooldown-aware fallback, so it never falls through to model B.

**Suggested fix**
Revalidate cached auto-model hits against the same strict cooldown eligibility used by pick_model_with_fallback(...). If the cached model has no non-cooling targets, bypass/forget the cache and let normal fallback choose another healthy model.

michaelneale:
I am still not sure if this is right approach but main thing is to think of the right algo to use when the mesh serves multiple of the same kind. Auto should also ladder through models, it is a trade off.

PR #642, opened 2026-05-22
fix(justfile): default optional args to empty string

fgadaleta:
## Summary
Fixes default variadic argument handling in the `Justfile` for two recipes:
- `bench-corpus`
- `metrics-server`

Both recipes now default `*ARGS` to an empty string, allowing them to run without requiring extra arguments.

## Changes
- Updated `bench-corpus` from `*ARGS` to `*ARGS=""`
- Updated `metrics-server` from `*ARGS` to `*ARGS=""`

## Why
This prevents `just` recipe invocation failures when no additional arguments are provided, while preserving support for passing optional extra arguments when needed.

## Testing
`just build` now runs without errors on Linux

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -204,7 +204,7 @@ skippy-wan-lab-build-bins:
     cargo build --release --locked -p skippy-server -p skippy-prompt -p metrics-server -p skippy-model-package
 
 # Generate a reproducible benchmark corpus for skippy bench tooling.
-bench-corpus tier="smoke" *ARGS:
+bench-corpus tier="smoke" *ARGS="":
     scripts/generate-bench-corpus.py "{{ tier }}" {{ ARGS }}
 
 # Run skippy family certification checks.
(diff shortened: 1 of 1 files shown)


ndizazzo:
Hey @fgadaleta - thanks for the fix! 

Any chance you could add a description with some before / after work so we can get a sense for an at-a-glance understanding of:

1) What the original issue was
2) What was done to solve it
3) The corrected result

fgadaleta:
> Hey @fgadaleta - thanks for the fix!
> 
> Any chance you could add a description with some before / after work so we can get a sense for an at-a-glance understanding of:
> 
> 1. What the original issue was
justfile seemed broken in linux (just returned with error and no build)


```
just build vulkan

error: Non-default parameter `ARGS` follows default parameter
   ——▶ Justfile:207:28
    │
207 │ bench-corpus tier="smoke" *ARGS:
    │                            ^^^^
```

> 2. What was done to solve it
fixed justfile

> 3. The corrected result

`just build` now works without errors (linux)

i386 (approved):
LGTM

PR #644, opened 2026-05-22
Normalize non-stream chat tool call IDs

IvGolovach:
## Title

Normalize non-stream chat tool call IDs

## Original problem

Non-stream `/v1/chat/completions` responses could return assistant `tool_calls` without an `id` when the upstream model or adapter omitted it.

That breaks some OpenAI-compatible agent/tool-call clients, because they rely on `tool_call.id` to correlate tool calls with tool results. The streaming path already had normalization for this shape; the non-stream path did not.

## Diagnostics

This came from the first actionable issue in the stability run: agent-style tool-call traffic could receive non-stream chat completion tool calls without IDs.

I reproduced the response-shape gap locally with a focused regression around the non-stream relay path, then verified the existing streaming behavior remained covered.

## Fix

This PR adds focused normalization for non-stream chat completion responses:

- preserves existing non-empty `tool_call.id` values
- synthesizes stable IDs only when an ID is missing or blank
- avoids collisions across multiple choices and multiple tool calls
- routes non-stream `/v1/chat/completions` responses through the normalizer
- keeps the existing streaming normalization behavior intact

Compatibility: this is response-shape normalization only. It does not change mesh protocol, protobuf fields, model selection, routing semantics, or request payloads.

Out of scope: the separate public-mesh tunnel 503 and visible `<think>` findings from the same stability run.

## Validation

Validation tier: Tier 2 - narrow OpenAI-compatible response-shape compatibility repair in host-runtime transport.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime network::openai::tool_call_ids --lib`: PASS, 5 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime relay_normalized_chat_completion_json_adds_missing_tool_call_id --lib`: PASS, 1 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime test_read_http_request_fragmented_post_body --lib`: PASS, 1 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal /opt/homebrew/bin/cargo-clippy clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS

Version: not applicable - compatibility normalization only; no release/version sync required.

UI screenshots: not applicable - no UI changed.

Live nightly stability rerun: not run - not required for this narrow local proof; the remaining stability-run findings are separate follow-ups.

## Rollback

Revert this PR.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
@@ -9,13 +9,15 @@ use crate::network::affinity::{
     prepare_remote_targets_for_request, AffinityRouter, PreparedTargets, TargetSelection,
 };
 use crate::network::openai::response_adapter;
+use crate::network::openai::tool_call_ids::{
+    normalize_chat_completion_json_body, ChatStreamNormalizationState,
+};
 use crate::network::router;
 use crate::network::target_health::TargetHealthOutcome;
 use crate::plugin;
 use anyhow::{anyhow, bail, Context, Result};
 // moa imports relocated into moa_gateway.rs (sole user after merge)
 use serde::Deserialize;
-use serde_json::{Map, Value};
 use std::collections::HashMap;
 use std::time::{Duration, Instant};
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWriteExt};
@@ -86,6 +88,7 @@ impl BufferedHttpRequest {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ResponseAdapter {
     None,
+    OpenAiChatCompletionsJson,
     OpenAiChatCompletionsStream,
     OpenAiResponsesJson,
     OpenAiResponsesStream,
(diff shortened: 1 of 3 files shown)


crates/mesh-llm-host-runtime/src/network/openai/transport.rs
```diff
@@ -1812,6 +1729,41 @@ async fn relay_translated_responses_json<R: AsyncRead + Unpin>(
...
+) -> Result<RouteAttemptResult> {
+    if retry_context_overflow && probe.retryable_context_overflow {
+        return Ok(RouteAttemptResult::RetryableContextOverflow);
+    }
+
+    if !(200..300).contains(&probe.status_code) {
+        return relay_error_response(tcp_stream, reader, probe).await;
+    }
+    let mut buffered = probe.buffered;
+    reader.read_to_end(&mut buffered).await?;
```
ndizazzo: I think this will cause the connection to hang in certain scenarios using HTTP/1.1: 

Because of the nature of the `read_to_end` call, it'll wait for any upstream socket to close before returning anything... with keep-alive and content-length, any non-streaming completions will just wait for this to finish even though the body already got sent. 

I think it's safe to just copy what you have in the generic success path and stop reading when we see `header_end + Content-Length`

ndizazzo (approved):
One small minor edge case, otherwise looks good

PR #645, opened 2026-05-22
feature(tool-calling): harden tool calling for agent harnesses

ndizazzo:
# Summary

Adds Forge-inspired OpenAI guardrails for hosted Skippy chat so small/open models can more reliably satisfy tool-call and structured-output contracts. The guardrail layer stays behind the `OpenAiBackend` seam instead of changing public OpenAI route handling, mesh protocol, plugin protocol, or the Skippy ABI.

# What changed

- Added reusable guardrail primitives in `mesh-llm-guardrails` and OpenAI frontend guardrail modules for:
  - request-contract parsing (`tools`, `tool_choice`, `parallel_tool_calls`, `response_format`, `mesh_guardrails`)
  - tool-call rescue from malformed model text
  - retry nudges after completed semantic model responses
  - validated structured-output emulation
  - OpenAI-shaped guardrail errors
- Added `GuardedOpenAiBackend` wrapping for hosted Skippy non-streaming chat.
  - Streaming stays pass-through.
  - `/v1/responses` inherits guarded chat behavior through the existing adapter.
  - Legacy completions remain outside v1 guardrails.
- Added operator controls:
  - `mesh-llm serve --mesh-guardrails <disabled|metrics|enforce>`
  - `mesh-llm runtime guardrails --mode <...>`
  - `POST /api/runtime/mesh-guardrails`
- Surfaced privacy-safe runtime status via `/api/status.runtime.openai_guardrails`.
- Added bounded guardrail telemetry for decisions/outcomes without prompt, schema, endpoint, or argument leakage.
- Added docs and benchmark scaffolding:
  - `docs/design/OPENAI_GUARDRAILS.md`
  - CLI/usage docs for runtime controls
  - telemetry docs
  - guardrail corpus runner: `scripts/run-openai-guardrail-corpus.py`

# Why

Agent harnesses and tool-using clients are sensitive to small formatting failures from open models: malformed tool JSON, fake tool text, sentinel leakage, or structured-output drift can break otherwise usable responses. This adds a conservative, opt-in middleware layer that can rescue or retry those cases for Skippy-backed chat while preserving existing API compatibility and keeping enforcement disabled by default until live evidence justifies broader rollout.

# Testing

- `cargo test -p openai-frontend --lib` - 124 passed
- `cargo test -p openai-frontend --test benchy_contract` - 8 passed
- `cargo test -p skippy-server --lib` - 82 passed
- `cargo test -p mesh-llm-host-runtime --lib` - 1446 passed, 6 ignored
[...]

Issue #636, opened 2026-05-22
openai-frontend: chat_template_kwargs.enable_thinking / reasoning_effort='none' not honored on direct model calls


michaelneale:
## What

Direct `/v1/chat/completions` calls to a specific model (e.g. `unsloth/MiniMax-M2.5-GGUF:Q4_K_M`) with explicit no-think hints do NOT actually disable thinking. The model still emits `<think>...` reasoning tokens regardless of the caller's stated preference.

## Repro

Against fly.io console (latest main, commit `41a2663f`) just now:

```bash
curl -sS https://mesh-llm-console.fly.dev/v1/chat/completions \
  -H 'content-type: application/json' \
  -d '{
    "model":"unsloth/MiniMax-M2.5-GGUF:Q4_K_M",
    "messages":[{"role":"user","content":"What is 2+2?"}],
    "reasoning_effort":"none"
  }'
# → content starts with: "<think> The user asks a simple question..."
```

Same outcome with `chat_template_kwargs: {"enable_thinking": false}` at the top level.

## Why this happens

PR #629 added MoA-side enable_thinking propagation in `crates/mesh-mixture-of-agents/src/backend.rs::apply_enable_thinking` — it injects `chat_template_kwargs.enable_thinking` into requests going through the MoA fanout/reducer path.

The direct-model path (`openai-frontend` → routing → staged runtime) bypasses MoA entirely. `openai_frontend::common::normalize_reasoning_template_options` apparently knows about these knobs (Copilot's #629 review references it) but the staged runtime doesn't see the result, or doesn't apply it to the chat template before tokenization.

The relevant llama.cpp patch (`third_party/llama.cpp/patches/0027-Expose-stage-chat-template-thinking-toggle.patch`) adds `override_enable_thinking + enable_thinking` ABI parameters to `skippy_apply_chat_template`, so the staged runtime is capable of honoring it — just not being called with it on the direct path.

## Why this matters

* The mesh virtual model (`mesh` via MoA) honors no-think (per #629). Anyone running an agent harness against `mesh` is fine.
* Anyone using a specific model name (Goose, OpenCode, Pi, Claude Code with a pinned model) sees full thinking burn — including the 13s latency on simple prompts I just measured.
* The escape hatch in #629's PR description was "caller can still explicitly enable/disable thinking via any of the recognised knobs" — but that promise only holds for MoA, not for direct calls.

## Suggested fix

[...]

PR #647, opened 2026-05-23
Honor direct chat no-think requests

IvGolovach:
## Summary

Direct OpenAI-compatible chat requests now preserve explicit no-thinking intent when they are forwarded to downstream model backends.

This canonicalizes caller intent such as `reasoning_effort: "none"`, `reasoning.enabled=false`, provider thinking aliases, and `thinking_budget=0` into `chat_template_kwargs.enable_thinking=false` before forwarding. Existing `chat_template_kwargs` values remain authoritative and are not overwritten.

## Why

Issue #636 showed that direct pinned-model requests could still surface `<think>` content even when the caller explicitly requested no thinking.

The typed Skippy path already understands these options, but the host proxy forwarding path could pass the original request body through without translating the no-thinking intent into the backend-compatible chat-template shape. That meant behavior depended on every downstream backend understanding every caller-specific reasoning knob.

## What changed

- Canonicalizes `/v1/chat/completions` and translated `/v1/responses` bodies into `chat_template_kwargs.enable_thinking` when explicit thinking/no-thinking intent is present.
- Extends host-runtime forwarding detection so reasoning/template knobs trigger JSON body normalization before proxying.
- Preserves existing `chat_template_kwargs` overrides and unrelated request fields.
- Preserves nested no-thinking precedence, so `reasoning.enabled=false` wins over `reasoning.effort="low"` before forwarding.
- Adds regressions for both the normalization layer and the relayed host-runtime forwarding path.
- Adds Skippy-server coverage for the top-level `reasoning_effort: "none"` request shape.

## Scope relative to #645

This PR stays scoped to request-side compatibility normalization before the direct forwarding/proxy path sends a body to a downstream OpenAI-compatible backend.

#645 is the broader guardrail/output-contract layer and should continue to own tool rescue, structured-output retries, and output cleanup. This PR does not try to duplicate that middleware path, and it should remain useful even while #645 settles because it does not depend on opt-in guardrails.

## Compatibility

This is an HTTP-body-only compatibility fix.

No protobuf, gossip, mesh protocol, catalog, or Skippy ABI changes. Older peers and existing model-serving paths should continue to ignore unknown request fields as before.

## Branch integrity

- Base branch: `main`
- Validated base: `origin/main@c7948d7181f26ed8cf99f9389a7deb8988162a32`
- Branch state: `0 behind / 2 ahead`
- Head commit: `a915c43199bbee9046c1e412ef2e24cf5f1ddeba`

## Validation

- Validation tier: Tier 2R - narrow post-review correction on top of the existing OpenAI-compatible direct chat request normalization PR.
- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, origin/main at `c7948d7181f26ed8cf99f9389a7deb8988162a32`.
- `cargo fmt --all`: PASS
- `cargo test -p openai-frontend normalize_chat_reasoning_enabled_false_wins_over_nested_effort --lib`: FAIL before fix, reproduced reviewer edge case.
- `cargo test -p openai-frontend normalize_chat --lib`: PASS, 4 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime chat_reasoning_enabled_false_wins_over_nested_effort_before_forwarding --lib`: PASS, 1 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime chat_reasoning_effort_none_is_canonicalized_before_forwarding --lib`: PASS, 1 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime chat_existing_template_kwargs_survive_forwarding_rewrite --lib`: PASS, 1 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal cargo test -p skippy-server top_level_reasoning_effort_none_turns_off_chat_template_thinking --lib`: PASS, 1 passed
- `LLAMA_STAGE_BUILD_DIR=<repo>/.deps/llama-build/build-stage-abi-metal /opt/homebrew/bin/cargo-clippy clippy -p openai-frontend -p skippy-server -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS
- Version: not applicable - post-review compatibility precedence correction only; no release/version sync required.
- Remote CI: PASS on final SHA `a915c43199bbee9046c1e412ef2e24cf5f1ddeba` (PR Builds, PR Docker Build, and PR Quality Checks).
- Not run: live MiniMax/Qwen direct-model smoke - not required for this deterministic precedence correction.
- Not run: two-node mesh or agent harness - not required for selected Tier 2R correction.

## Rollback

Revert this PR.

DB downgrade: not applicable.  
Data repair: not applicable.  

## Residual risk

The original hosted-model symptom should still be confirmed against the affected live endpoint. This PR proves the deterministic request-normalization and forwarding path locally, but does not claim to solve unrelated `<think>` leakage outside that path.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
@@ -5425,6 +5425,30 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn chat_reasoning_enabled_false_wins_over_nested_effort_before_forwarding() {
+        let body = serde_json::json!({
+            "model": "qwen",
+            "messages": [{"role": "user", "content": "hi"}],
+            "reasoning": {"enabled": false, "effort": "low"}
+        })
+        .to_string();
+        let raw = format!(
+            "POST /v1/chat/completions HTTP/1.1\r\nHost: localhost\r\nContent-Type: application/json\r\nContent-Length: {}\r\n\r\n{}",
+            body.len(),
+            body
+        );
+
+        let request = read_request_from_parts(vec![raw.into_bytes()]).await;
+        let forwarded = parse_json_body_from_http_request(&request.raw).unwrap();
+
+        assert_eq!(
+            forwarded["chat_template_kwargs"]["enable_thinking"],
+            serde_json::json!(false)
+        );
+        assert_eq!(request.body_json, Some(forwarded));
+    }
+
+    #[tokio::test]
     async fn test_read_http_request_large_body_over_32k() {
         let large = "x".repeat(40_000);
         let body = serde_json::json!({
--- a/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
@@ -677,12 +677,23 @@ fn request_requires_json_transform(path: &str, body: &[u8], plugin_manager_prese
 
     body_text.contains("\"max_completion_tokens\"")
         || body_text.contains("\"max_output_tokens\"")
+        || body_text_contains_chat_reasoning_template_options(body_text)
         || (plugin_manager_present
             && (body_text.contains("mesh://blob/")
                 || body_text.contains("\"blob_token\"")
                 || body_text.contains("\"mesh_token\"")
                 || body_text.contains("\"input_audio\"")
                 || body_text.contains("\"input_image\"")))
+}
+
+fn body_text_contains_chat_reasoning_template_options(body_text: &str) -> bool {
+    body_text.contains("\"reasoning\"")
+        || body_text.contains("\"reasoning_effort\"")
+        || body_text.contains("\"thinking_budget\"")
+        || body_text.contains("\"chat_template_kwargs\"")
+        || openai_frontend::THINKING_BOOLEAN_ALIASES
+            .iter()
+            .any(|field| body_text.contains(&format!("\"{field}\"")))
 }
 
 fn parse_json_body_from_http_request(raw: &[u8]) -> Option<serde_json::Value> {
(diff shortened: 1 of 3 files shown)


ndizazzo:
This might conflict with #645 - there's some work in there to correct think tags via the middleware layered approach, which I think is going to be more "canonical" for tweaking model output from backends to shape it

If that PR doesn't already entirely fix the issue, we might consider layering select changes from this one into the surfaces it adds instead

i386:
One thing I noticed: `reasoning.enabled: false` can get flipped back on if the same request also has something like `reasoning.effort: "low"`. That seems different from the existing `normalize_reasoning_template_options` behavior, where `enabled: false` wins and keeps thinking off. Could you make this forwarding normalizer follow the same precedence so direct-model requests do not accidentally re-enable thinking?

IvGolovach:
Thanks both, I looked through this carefully and updated the PR.

For the precedence edge case: `reasoning.enabled=false` now wins over nested `reasoning.effort="low"` before forwarding, matching the typed Skippy behavior. I added regressions at both the OpenAI normalization layer and the host-runtime forwarding path so this does not drift again.

For the #645 overlap: I kept this PR scoped to request-side compatibility normalization for the direct forwarding/proxy path. #645 still feels like the right home for the broader guardrail/output-contract layer, but this fix covers the raw forwarded request shape before any guardrail middleware is involved. The PR body is updated with the scope and validation, and CI is green on the final SHA.

i386:
LGTM

PR #649, opened 2026-05-23
feat(embed): mesh_llm::embed — in-process NodeBuilder/NodeHandle for Rust consumers

michaelneale:
## What you can now do

Run a mesh-llm node from a Rust app without shelling out to the CLI:

```rust
use mesh_llm::embed::{NodeBuilder, NodeRole, QuicBindSelection};

let handle = NodeBuilder::new()
    .role(NodeRole::Client)
    .relays(["https://gated.example/"])
    .relay_auth("https://gated.example/", "<nip98-bearer>")
    .max_vram_gb(Some(0.0))
    .start()
    .await?;

handle.start_accepting();
handle.set_display_name("my-app".to_string()).await;
let invite = handle.invite_token();
```

The full surface is at `mesh_llm::embed`. Builder methods cover role, relays, per-relay auth tokens (works with the new `--relay-auth` feature from #641), QUIC bind, VRAM cap, and hardware enumeration. `NodeHandle` forwards a curated subset of `mesh::Node`: `start_accepting`, `id`, `invite_token`, `join`, `set_display_name`, `set_models`, `models`, `role`, `peers`.

## Scope

**In:** mesh layer only — peer membership, gossip, invite tokens, relay registration, model advertisement.

**Out (deliberately, for now):** HTTP proxy, management console, TUI, local inference, auto-discovery, model downloads. Those stay behind the `mesh-llm` CLI. Open an issue if you need an embed surface for any of them.

## Architecture

- The new module is `crates/mesh-llm-host-runtime/src/embed.rs`, re-exported from the shipped `mesh-llm` crate as `mesh_llm::embed`.
- `NodeHandle` wraps `mesh::Node` (kept `pub(crate)`) so we can keep refactoring internal mesh code without breaking embedders. Likewise, `PeerInfo` is reduced to an `EmbedPeer` view (`id`, `role`, `models`, `rtt_ms`).
- The embed module is documented as the **only stable Rust API** in this crate. Everything else may change in any release.
- See `docs/embedding.md` for the embedder-facing doc, including the relationship to `mesh-llm-client` (which is the *client-only* talk-to-a-mesh crate; embed is for *being part of* a mesh).

## Validation

- `cargo test -p mesh-llm-host-runtime --lib embed::` — unit tests pin builder defaults, chainability, and that `relay_auths(map)` replaces prior `relay_auth(url, token)` entries.
- `cargo test -p mesh-llm --test embed_api_surface` — external-consumer compile/runtime test for the public surface (chains, `NodeRole` default + serde round-trip, `QuicBindSelection` default). Future signature regressions fail this test, which is what an external embedder would feel.
- `cargo clippy --all-targets -- -D warnings` clean on `-p mesh-llm` and `-p mesh-llm-host-runtime`.
- `cargo fmt --all -- --check` clean.

## Follow-ups (not in this PR)

- `NodeHandle::shutdown()` for graceful drop of the endpoint + background tasks. v1 ships with implicit drop only.
- Embed surface for the OpenAI HTTP proxy if there's demand.
- Embed surface for skippy / local inference once the stage-runtime API stabilises further.

Depends on #641 (the `RelayConfig` struct from --relay-auth).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,6 +44,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
+dependencies = [
+ "cfg-if 1.0.4",
+ "getrandom 0.3.4",
+ "once_cell",
+ "version_check",
+ "zerocopy",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -247,6 +260,45 @@ dependencies = [
  "serde",
  "serde_derive",
  "winnow 0.7.15",
+]
+
+[[package]]
+name = "asn1-rs"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56624a96882bb8c26d61312ae18cb45868e5a9992ea73c58e45c3101e56a1e60"
+dependencies = [
+ "asn1-rs-derive",
+ "asn1-rs-impl",
+ "displaydoc",
+ "nom 7.1.3",
+ "num-traits",
+ "rusticata-macros",
+ "thiserror 2.0.18",
+ "time",
+]
+
+[[package]]
+name = "asn1-rs-derive"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+ "synstructure",
+]
+
+[[package]]
+name = "asn1-rs-impl"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b18050c2cd6fe86c3a76584ef5e0baf286d038cda203eb6223df2cc413565f7"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


ndizazzo:
Initial run through the lab gives it a 🟢 light, with some minor housekeeping items:

---

### One observation (not blocking)
**`RelayConfig` is re-exported from `embed` but embed users never touch it.**

`pub use crate::mesh::{NodeRole, QuicBindSelection, RelayConfig};`

`RelayConfig<'a>` has a lifetime and borrows from the builder's owned fields. Embed users never construct it — `NodeBuilder::start()` builds it internally. Re-exporting it from `embed` means it's part of the stable surface even though no embed method signature references it. Consider whether to keep it in the re-export or leave it as `pub` in `mesh` only (embed users who dig into internals can reach it). Minor, and fine either way — just worth being intentional about.

### One test fragility note
**`wrong_token_is_rejected_by_gated_relay` asserts on error message text.**
```
assert!(
    auth_err.contains("not authorized"),
    "expected 'not authorized' in error, got: {auth_err}"
);
```
This string-matches on iroh-relay's error formatting via `format!("{err:#}")`. If the iroh crate changes its error messages or the `Display` impl of the relay error type, this test breaks silently. Since `iroh` is pinned to `1.0.0-rc.0` in dev-dependencies this is stable for now, but it's worth flagging. If this becomes flaky in the future, switch to matching on the error variant type rather than the display text.

The `missing_token_for_gated_relay` test avoids this by only checking `.online()` timeout — that's the right pattern for the negative case.

### Minor: `id()` uses Debug formatting
```
pub fn id(&self) -> String {
    format!("{:?}", self.inner.id())
}
```

This works but ties the output to the Debug impl of whatever `id()` returns. If that changes, embed users get a different format. Consider using an explicit hex-format helper (or `.to_string()`) if the type has a Display impl. Not a bug — just a note for when the API stabilises further.

michaelneale:
Closing — wrong scope and wrong crate. The embed work belongs in #641 alongside the relay-auth feature, and the new published surface needs to live in an existing/new published crate (mesh-api lineage), not in mesh_llm::embed inside the unpublished host-runtime monolith. Will fold into #641.

Issue #652, opened 2026-05-23
skippy: 'decode: failed to find a memory slot for batch of size 2048' on goose/pi after a few tool-laden turns


michaelneale:
## What

When using goose or pi (any harness with a long tool-laden system prompt + grammar-triggered tool extraction) against a locally-served small model (Qwen2.5-3B-Instruct-GGUF:q4_k_m on M4 Max), the second or third tool-calling turn fails with:

```
Server error: skippy ABI call failed: RuntimeError: llama_decode failed.
```

Native log (`~/.mesh-llm/runtime/<pid>/logs/skippy-native.log`) shows the underlying cause:

```
Grammar triggered on regex: '<tool_call>'
Grammar triggered on regex: '<tool_call>'
Grammar triggered on regex: '<tool_call>'
Grammar triggered on regex: '<tool_call>'
Grammar triggered on regex: '<tool_call>'
decode: failed to find a memory slot for batch of size 2048
```

The grammar retry / rescue path fires ~5 times before `llama_decode` runs out of KV-cache memory slots.

## Repro

1. `mesh-llm serve --model Qwen/Qwen2.5-3B-Instruct-GGUF@main:q4_k_m --port 9447 --console 3232 --log-format json --auto --publish` on main (`c7948d71`)
2. `GOOSE_PROVIDER=openai OPENAI_HOST=http://127.0.0.1:9447 OPENAI_API_KEY=mesh GOOSE_MODEL='Qwen/Qwen2.5-3B-Instruct-GGUF@main:q4_k_m' goose run -t 'Read /tmp/foo.txt and tell me what it says'` works once
3. Run a second `goose run -t ...` shortly after \u2014 fails with `llama_decode failed`.

Same failure with `pi --provider mesh --model 'Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m' -p '...'`.

## What is happening

The harness sends a long system prompt with many tool definitions + the user's request. The model emits a tool call. Goose/pi pick up the tool call and immediately fire a follow-up request with the tool result. The grammar rescue path triggers on `<tool_call>` regex multiple times during stream decoding, each apparently consuming a KV slot that doesn't get freed. After 5 such triggers the pool is exhausted.

A simple `curl` with the same body shape doesn't hit this \u2014 only sustained tool-loop pressure does.

## Workaround

[...]

PR #659, opened 2026-05-23
Reduce Skippy unified KV pressure

IvGolovach:
## Summary

Reduces Skippy unified-KV pressure for multi-lane serving by lowering the default raw runtime `n_batch` used when `lane_count > 1`, while preserving single-lane behavior and explicit model/config overrides.

The PR also makes oversized resident-prefix cache candidates nonfatal. If one candidate cannot fit within the resident-prefix token/byte budget, Skippy now treats that candidate as uncacheable instead of failing the record path, so smaller shared-prefix candidates can still be attempted.

## Why

Issue #652 diagnosed `decode: failed to find a memory slot for batch of size 2048` under Goose/Pi-style tool traffic against Skippy unified KV. In that mode, several lanes share one KV cell pool, and upstream llama.cpp no longer performs the old defrag path when a large contiguous slot cannot be found.

This is a conservative mitigation, not a claim that the long-term defrag/root-cause work is finished. It directly reduces the contiguous batch pressure that triggers the reported failure and keeps resident-prefix cache recording from turning an oversized candidate into a hard error.

Refs #652.

## Diff scope

- Adds a Skippy multi-lane unified-KV default `n_batch` of `1024`.
- Keeps single-lane default behavior at the existing llama-server-compatible `2048`.
- Preserves explicit `batch` / `n_batch` and `ubatch` / `n_ubatch` overrides.
- Treats resident-prefix candidates larger than the configured token/byte budget as uncacheable rather than fatal.
- Prevents uncacheable resident-prefix candidates from being saved or retained with a placeholder sequence id.
- Adds regression coverage for both the runtime default and the resident-prefix cache edge case.

## Compatibility

No protobuf, gossip, mesh wire protocol, plugin protocol, or Skippy ABI changes.

This changes only default runtime/cache behavior for Skippy serving. Existing explicit model/config batch overrides remain authoritative.

## Branch integrity

- Base branch: `main`
- Validated base: `origin/main@c36132fd3d9f693cac6b9ceafd976db38545656e`
- Branch state during validation: `0 behind / 1 ahead`
- Head commit: `675576bf8b85527b8a58c557d9568375b65348cd`

## Commit integrity

One logical commit:

- `675576bf8b85527b8a58c557d9568375b65348cd` - Reduce Skippy unified KV pressure

Final diff contains only Skippy runtime/cache/server changes.

## Validation

Validation tier: Tier 2 - narrow Skippy unified-KV pressure mitigation for issue #652; no protocol, ABI, gossip, UI, or release tooling changes.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, validated base `c36132fd3d9f693cac6b9ceafd976db38545656e`
- `git rev-list --left-right --count origin/main...HEAD`: PASS, `0 behind / 1 ahead`
- `cargo test -p skippy-cache oversized_resident_prefix_candidate_is_nonfatal --lib`: FAIL before cache guard, red test expected nonfatal oversized resident-prefix allocation
- `cargo test -p skippy-cache oversized_resident_prefix_candidate_is_nonfatal --lib`: PASS, 1 passed
- `cargo test -p skippy-cache --lib`: PASS, 14 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build> cargo test -p skippy-runtime runtime_config_raw_ --lib`: PASS, 4 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build> cargo test -p skippy-runtime --lib`: PASS, 39 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build> cargo test -p skippy-server --lib`: PASS, 81 passed outside the restricted sandbox; the restricted-sandbox run failed only on a socket/listener test with `Operation not permitted`
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build> cargo test -p mesh-llm-host-runtime runtime::context_planning --lib`: PASS, 9 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build> cargo check -p mesh-llm`: PASS
- `LLAMA_STAGE_BUILD_DIR=<stage-abi-build> cargo-clippy clippy -p skippy-cache -p skippy-runtime -p skippy-server --all-targets -- -D warnings`: PASS
- Version: not applicable - runtime/cache mitigation inside existing Skippy config defaults; no release/version sync required.

Not run:

- Live Goose/Pi Qwen2.5-3B reproduction - no affected live direct-model endpoint was available during local validation. The deterministic runtime-default and resident-prefix cache regressions cover the local failure paths changed by this PR.

## Runtime safety

No new locks, queues, network paths, wire formats, or background tasks are introduced.

The resident-prefix cache now exits early for candidates that cannot fit within a single-record budget, avoiding futile eviction attempts and avoiding invalid save/retain behavior for uncacheable candidates.

## Rollback

Rollback: revert this PR.

DB downgrade: not applicable.
Data repair: not applicable.

## Known residual risks

This PR mitigates the reported pressure path; it does not restore llama.cpp defrag or claim complete root-cause closure for all unified-KV fragmentation cases.

The live Goose/Pi reproduction should still be re-run by someone with the affected Skippy direct-model endpoint available.

### DIFF
--- a/crates/skippy-cache/src/resident/prefix.rs
+++ b/crates/skippy-cache/src/resident/prefix.rs
@@ -47,6 +47,36 @@ pub struct ResidentPrefixAllocation {
     pub seq_id: i32,
     pub evictions: Vec<ResidentPrefixEviction>,
     pub should_save: bool,
+    pub should_retain: bool,
+}
+
+impl ResidentPrefixAllocation {
+    fn existing(seq_id: i32) -> Self {
+        Self {
+            seq_id,
+            evictions: Vec::new(),
+            should_save: false,
+            should_retain: true,
+        }
+    }
+
+    fn new_record(seq_id: i32, evictions: Vec<ResidentPrefixEviction>) -> Self {
+        Self {
+            seq_id,
+            evictions,
+            should_save: true,
+            should_retain: true,
+        }
+    }
+
+    fn uncacheable() -> Self {
+        Self {
+            seq_id: -1,
+            evictions: Vec::new(),
+            should_save: false,
+            should_retain: false,
+        }
+    }
 }
 
 #[derive(Debug, Clone, Copy, Default)]
@@ -124,21 +154,16 @@ impl ResidentPrefixCache {
         }
         self.clock = self.clock.saturating_add(1);
         if let Some(entry) = self.entries.get(page_id) {
-            return Ok(ResidentPrefixAllocation {
-                seq_id: entry.seq_id,
-                evictions: Vec::new(),
-                should_save: false,
-            });
+            return Ok(ResidentPrefixAllocation::existing(entry.seq_id));
+        }
+        if self.candidate_exceeds_single_record_budget(estimated_bytes, token_count) {
+            return Ok(ResidentPrefixAllocation::uncacheable());
         }
 
         let evictions =
             self.evict_until_room_for(estimated_bytes, token_count, &mut drop_evicted)?;
         let seq_id = self.next_sequence_id()?;
-        Ok(ResidentPrefixAllocation {
-            seq_id,
-            evictions,
-            should_save: true,
-        })
+        Ok(ResidentPrefixAllocation::new_record(seq_id, evictions))
     }
 
     pub fn commit_record(
@@ -226,6 +251,16 @@ impl ResidentPrefixCache {
         Ok(evictions)
     }
 
+    fn candidate_exceeds_single_record_budget(
+        &self,
+        estimated_bytes: u64,
+        token_count: u64,
+    ) -> bool {
+        let over_bytes = self.max_bytes > 0 && estimated_bytes > self.max_bytes;
+        let over_tokens = self.max_resident_tokens > 0 && token_count > self.max_resident_tokens;
+        over_bytes || over_tokens
+    }
+
     fn next_sequence_id(&mut self) -> Result<i32> {
         if let Some(seq_id) = self.free_seq_ids.pop() {
             return Ok(seq_id);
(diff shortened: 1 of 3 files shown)


ndizazzo (approved):
🚀  nice, going to add one more thing to this in #661

michaelneale:
Tested stacked on top of #661 against the issue #652 repro (`Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m` + goose).

**Confirmed effect:**
- Native log batch size changes from 2048 to 1024:
  ```
  llama_context: n_batch       = 1024     (was 2048 on main)
  ```
- Failures now read `decode: failed to find a memory slot for batch of size 1024` instead of `2048`.

**End-to-end result vs `main`:**

| scenario | this PR + #661 | #661 only | main |
|---|---|---|---|
| 1× cold tool-using prompt | ✅ pass | ✅ pass | varies |
| 10× sequential tool-using prompts | 1/10 (cold only) | 1/10 (cold only) | 3/20 |

So #659 + #661 stacked is **the same overall pass-rate** as #661 alone on the repro. The cold-start prompt succeeds, every subsequent tool-using prompt fails immediately.

Halving `n_batch` doesn't help here because the failure isn't "batch too big for free cells" — it's "no contiguous N-cell chunk free anywhere in the pool", and that's true at 1024 just as much as 2048.

That said, your PR's other change (oversized resident-prefix candidates → uncacheable instead of `bail!`) is a clean safety improvement on its own. Worth landing for that alone.

**Repro** (build with `cargo build -p mesh-llm`, then):
```bash
./target/debug/mesh-llm serve \
  --model Qwen/Qwen2.5-3B-Instruct-GGUF@main:q4_k_m \
  --port 9447 --console 3232 --log-format json --auto --publish

# in another shell
for i in 1 2 3 4 5 6 7 8 9 10; do
  prompts=("list files here" "show me README" "what's in Cargo.toml" "describe codebase" "find rust files")
  p="${prompts[$((i % 5))]}"
  GOOSE_PROVIDER=openai OPENAI_HOST=http://127.0.0.1:9447 OPENAI_API_KEY=mesh \
  GOOSE_MODEL='Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m' \
    timeout 60 goose run -t "$p"
done
```

Process stays alive through failures, just returns 500 to every subsequent request after the first cold one. Doesn't appear to be the contiguous-batch path you're targeting.

PR #660, opened 2026-05-23
Resume and harden trusted Skippy package artifact transfer

IvGolovach:
## Summary

Workers can now resume trusted Skippy layer-package artifact prefetches from an existing partial file instead of restarting every peer transfer from byte 0.

This also tightens the serving side: peer transfer now requires explicit pinned `hf://namespace/repo@revision` package refs, verifies cached artifact bytes against the package manifest before serving, and keeps partial files only when they are safe to resume.

## Why

The Skippy artifact-transfer protocol already had an `offset` field and the docs describe resume support, but the requester always sent `offset = 0` and wrote to a fresh hidden partial file. Interrupted transfers therefore lost useful progress, which matters for large layer-package artifacts on LAN/private mesh deployments.

While reviewing that path, I also found an integrity gap: a same-size corrupted cached artifact could satisfy manifest-declared size checks and be served with the manifest SHA without hashing local bytes first.

## Diff scope

- Moves partial artifact file handling into a dedicated mesh artifact-transfer IO helper.
- Reuses the largest valid hidden partial for the destination and sends its length as the transfer offset.
- Appends only the missing suffix, then verifies final size/SHA before atomic install.
- Preserves partials after resumable transfer interruptions, but removes them after final size/SHA failure.
- Rejects mutable or implicit package refs for peer transfer; only explicit `hf://namespace/repo@revision` refs are eligible.
- Hash-verifies non-manifest cached artifacts before serving them to a peer.
- Adds regression coverage for resume selection/body append, nonzero stream offset, pinned-ref eligibility, and same-size corrupted cache rejection.

## Protocol / compatibility

No protobuf, stream-byte, ALPN, gossip, or Skippy ABI change.

This continues to use the existing `skippy-stage/1` artifact-transfer stream and the existing `offset` field. Mixed-version behavior remains feature-gated by the existing subprotocol advertisement; peers without artifact-transfer support are skipped by the existing fallback path.

No package inventory, cache paths, or artifact filenames are gossiped.

## Branch integrity

- Base branch: `main`
- Validated base: `origin/main@c36132fd`
- Branch state: `0 behind / 1 ahead`
- Head commit: `e1bd221d`

## Validation

- Validation tier: Tier 3 - mesh/skippy runtime transport hardening for trusted layer-package artifact transfer; no proto/schema change.
- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, validated `origin/main@c36132fd`.
- `cargo test -p skippy-protocol stage_artifact_transfer --lib`: PASS, 1 passed.
- `cargo test -p skippy-topology --lib`: PASS, 26 passed.
- `LLAMA_STAGE_BUILD_DIR=<prepared llama-stage ABI build> cargo test -p mesh-llm-host-runtime artifact_transfer --lib`: PASS, 23 passed.
- `LLAMA_STAGE_BUILD_DIR=<prepared llama-stage ABI build> cargo check -p mesh-llm`: PASS.
- `LLAMA_STAGE_BUILD_DIR=<prepared llama-stage ABI build> cargo-clippy clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS.
- Version: not applicable - no release metadata/version sync required; release flow owns version bumps.
- Remote CI: pending after PR creation.
- Not run: live two-node artifact-transfer smoke - no local multi-machine package cache setup was available; targeted mesh stream tests cover the protocol path, and remote CI should run before merge.

## Rollback

Revert this PR.

DB downgrade: not applicable.
Data repair: not applicable.

## Known residual risks

The live multi-machine benefit should still be confirmed on a private mesh with a large pinned HF layer package already cached on the coordinator. The local tests prove the deterministic transfer, resume, integrity, and policy paths.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/artifact_transfer_io.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/artifact_transfer_io.rs
@@ -1,0 +1,293 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::{Context, Result};
+use tokio::io::{AsyncRead, AsyncWriteExt};
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(super) struct PartialArtifactSelection {
+    pub(super) path: PathBuf,
+    pub(super) offset: u64,
+}
+
+pub(super) struct PartialArtifactGuard {
+    path: PathBuf,
+    cleanup_on_drop: bool,
+}
+
+impl PartialArtifactGuard {
+    #[cfg(test)]
+    pub(super) fn new(path: PathBuf) -> Self {
+        Self {
+            path,
+            cleanup_on_drop: true,
+        }
+    }
+
+    pub(super) fn preserve_on_error(path: PathBuf) -> Self {
+        Self {
+            path,
+            cleanup_on_drop: false,
+        }
+    }
+
+    pub(super) fn disarm(&mut self) {
+        self.cleanup_on_drop = false;
+    }
+
+    pub(super) fn remove_now(&mut self) {
+        let _ = std::fs::remove_file(&self.path);
+        self.disarm();
+    }
+}
+
+impl Drop for PartialArtifactGuard {
+    fn drop(&mut self) {
+        if self.cleanup_on_drop {
+            let _ = std::fs::remove_file(&self.path);
+        }
+    }
+}
+
+pub(super) fn partial_artifact_path(destination: &Path) -> PathBuf {
+    let file_name = artifact_file_name(destination);
+    let unique = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|duration| duration.as_nanos())
+        .unwrap_or(0);
+    destination.with_file_name(format!(
+        ".{file_name}.{}.{}.part",
+        std::process::id(),
+        unique
+    ))
+}
+
+pub(super) fn select_partial_artifact(
+    destination: &Path,
+    max_resume_size: u64,
+) -> Result<PartialArtifactSelection> {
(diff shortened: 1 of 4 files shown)


i386 (commented):
I found two issues worth fixing before merge:

1. **Oversized manifest partials can wedge resume retries.** In `fetch_artifact_from_peer`, the selected partial is preserved by default, but early failures before `transfer_result` do not remove it. For manifests, `artifact_transfer_resume_limit` allows any partial up to `MAX_PACKAGE_MANIFEST_BYTES`, because the real size is not known yet. If a stale/corrupt `.model-package.json.*.part` is larger than the peer's manifest, the requester sends that offset, the peer rejects or the requester hits `offset <= response.total_size`, and the same bad partial is kept for the next retry. I would remove the selected partial or retry from offset `0` when the peer reports an invalid/smaller offset, especially for manifest requests.

2. **Serving path hashes large artifacts synchronously on the Tokio worker.** The new cached-artifact integrity check calls `file_sha256_hex(&path)` directly while resolving a peer transfer request. Layer artifacts can be multi-GB, so this blocks a Tokio runtime worker before the response is sent. The requester side already uses `spawn_blocking` for final SHA verification; the serving side should do the same or use cached verification metadata.

Validation I ran locally: `cargo test -p skippy-protocol stage_artifact_transfer --lib` passed, and `cargo test -p skippy-topology --lib` passed. `cargo test -p mesh-llm-host-runtime artifact_transfer --lib` could not run in this worktree because `skippy-ffi` could not find the prepared native `llama-common` static library; it needs the `LLAMA_STAGE_BUILD_DIR` setup noted in the PR.

PR #661, opened 2026-05-23
fix(kv-eviction): proactive prefix cache eviction before decode

ndizazzo:
## Summary

When the resident prefix cache is at capacity and a grammar-triggered retry needs an extra KV cache slot mid-generation, the decode loop can fail with `"failed to find a memory slot"`. This is especially visible under agent harness tool loops where prefix cache entries pin KV cells across turns while the active lane needs more room for grammar retries.

This PR adds a single proactive eviction step before the decode loop starts: one LRU entry from the resident prefix cache is evicted right after prefill recording, while the runtime lock is still held. This frees KV cells before the decode loop begins, so grammar-triggered retries can use them.

The change is small (one `evict_one_lru_entry()` call per generation) and complementary to the existing `allocate_for_record` eviction path. It fires at a different point in the lifecycle and does not change the cache's own capacity-based eviction behavior.

### What #659 did

1. **Reduced `n_batch` from 2048 to 1024** in multi-lane unified KV mode. Smaller batches mean each `llama_decode` call touches fewer KV cells at once, so the peak cell demand per decode step is lower. This reduces how often the pool is over-committed during the decode loop.

2. **Made oversized prefix cache candidates nonfatal**. Previously if a candidate prefix exceeded the token budget, the cache would hard-error (`bail!`). After #659 it soft-rejects instead, keeping the existing prefix cache intact rather than crashing the lane.

### Why we want this too

Both #659 changes only reduce *how often* the failure triggers. They don't remove the failure mode itself:

- **Grammar retries still consume multiple KV slots** within a single generation. With `MAX_AUTO_PARALLEL_SLOTS=4`, each grammar trigger forks a new sequence in the KV cache. A generation with 5+ grammar triggers still exhausts 4 slots, regardless of batch size. The batch size reduction just means each of those 4 slots needs slightly less headroom per decode, but on turn 3 of an agent loop, the prefix cache pins enough KV cells from the same unified pool that even reduced batch demand can exceed what's left.

- **The prefix cache still pins KV cells across turns**. Each cached prefix occupies KV cells from the unified pool (`n_ctx` total cells shared between active lanes and cached prefixes). After 2-3 agent-harness turns, accumulated cached prefixes pin enough cells that the active lane's remaining pool is small. When grammar triggers then need 4 parallel sequences, the available pool may be too small even with `n_batch=1024`.

- **Session reset bookkeeping drift is not addressed**. The `drop_session_timed` path in `runtime_state.rs` (lines 347-353) can leave C-side KV slots allocated when the Rust-side cleanup fails under KV pressure. This drift accumulates across multiple generation cycles in a tool loop. Neither #659 change affects this path.

## Changes

- **`skippy-cache`**: Added `evict_one_lru_entry` to `ResidentPrefixCache`. Scans entries in LRU order (skipping borrowed entries) and evicts the least-recently-used one. 7 new unit tests cover empty cache, all-borrowed, LRU ordering, seq-id recycling, internal accounting, and interaction with `allocate_for_record`.

- **`skippy-server`**: Added `evict_one_resident_prefix` to `KvStageIntegration`. Bridges the cache eviction to the C-side drop callback.

- **`local_generation.rs`**: Calls `evict_one_resident_prefix` after prefill recording, before entering the decode loop.

- **`generation_flow.rs`**: Same eviction point for the multimodal path.

## Verification

- `cargo test -p skippy-cache --lib`: 24 tests pass
- `cargo test -p skippy-server --lib`: 81 tests pass
- `cargo check -p mesh-llm`: zero warnings
- `just test-all`: all 7 stages pass (fmt, clippy, Rust tests, ESLint/Prettier, tsc, vitest 658/661, Playwright smoke)

### DIFF
--- a/crates/skippy-cache/src/config.rs
+++ b/crates/skippy-cache/src/config.rs
@@ -127,7 +127,13 @@ impl PrefixCandidatePolicy {
             return counts;
         }
         let stride = self.stride_tokens.max(1).min(self.page_size_tokens.max(1));
-        let mut candidate = token_count.saturating_sub(1);
+        let mut candidate = stable_grid_floor(token_count, stride);
+        if candidate == token_count {
+            candidate = candidate.saturating_sub(stride);
+        }
+        if candidate < self.min_tokens {
+            candidate = self.min_tokens;
+        }
         while candidate >= self.min_tokens {
             counts.push(candidate);
             if candidate == self.min_tokens {
@@ -151,32 +157,15 @@ impl PrefixCandidatePolicy {
         let mut selected = Vec::with_capacity(limit);
         selected.push(token_count);
 
-        let lower_bound = candidates
-            .iter()
-            .copied()
-            .filter(|candidate| *candidate != token_count)
-            .min()
-            .unwrap_or(token_count);
         let shared_slots = limit.saturating_sub(1);
         if shared_slots == 0 {
             return selected;
         }
-        for slot in 0..shared_slots {
-            let target = if shared_slots == 1 {
-                lower_bound
-            } else {
-                let span = token_count.saturating_sub(lower_bound);
-                token_count
-                    .saturating_sub(span.saturating_mul((slot + 1) as u64) / shared_slots as u64)
+        for candidate in candidates.iter().copied() {
+            if selected.len() >= limit {
+                break;
             }
-            .max(self.min_tokens)
-            .min(token_count);
-            if let Some(candidate) = candidates
-                .iter()
-                .copied()
-                .filter(|candidate| *candidate <= target && *candidate != token_count)
-                .max()
-            {
+            if candidate != token_count {
                 selected.push(candidate);
             }
         }
@@ -196,6 +185,10 @@ impl PrefixCandidatePolicy {
     }
 }
 
+fn stable_grid_floor(token_count: u64, stride: u64) -> u64 {
+    token_count.saturating_sub(token_count % stride.max(1))
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -209,10 +202,7 @@ mod tests {
             page_size_tokens: 64,
         };
 
-        assert_eq!(
-            policy.candidate_token_counts(160),
-            vec![160, 159, 127, 95, 64]
-        );
+        assert_eq!(policy.candidate_token_counts(160), vec![160, 128, 96, 64]);
     }
 
     #[test]
@@ -224,7 +214,7 @@ mod tests {
             page_size_tokens: 64,
         };
 
-        assert_eq!(policy.record_candidate_token_counts(160), vec![160, 64]);
+        assert_eq!(policy.record_candidate_token_counts(160), vec![160, 128]);
     }
 
     #[test]
@@ -251,7 +241,40 @@ mod tests {
 
         assert_eq!(
             policy.record_candidate_token_counts(160),
-            vec![160, 159, 127, 95, 64]
+            vec![160, 128, 96, 64]
         );
+    }
+
+    #[test]
+    fn same_prefix_different_tail_prompts_share_near_tail_candidate() {
+        let policy = PrefixCandidatePolicy {
+            min_tokens: 256,
+            stride_tokens: 128,
+            record_limit: 2,
+            page_size_tokens: 256,
+        };
+
+        let recorded = policy.record_candidate_token_counts(2214);
+        let lookup = policy.candidate_token_counts(2231);
+        let shared = recorded
+            .iter()
+            .copied()
+            .find(|candidate| lookup.contains(candidate));
+
+        assert_eq!(recorded, vec![2214, 2176]);
+        assert_eq!(shared, Some(2176));
+    }
+
+    #[test]
+    fn non_aligned_min_tokens_still_provides_shared_floor_candidate() {
+        let policy = PrefixCandidatePolicy {
+            min_tokens: 300,
+            stride_tokens: 128,
+            record_limit: 2,
+            page_size_tokens: 300,
+        };
+
+        assert_eq!(policy.candidate_token_counts(350), vec![350, 300]);
+        assert_eq!(policy.record_candidate_token_counts(350), vec![350, 300]);
     }
 }
--- a/crates/skippy-cache/src/resident/prefix.rs
+++ b/crates/skippy-cache/src/resident/prefix.rs
@@ -47,6 +47,36 @@ pub struct ResidentPrefixAllocation {
     pub seq_id: i32,
     pub evictions: Vec<ResidentPrefixEviction>,
     pub should_save: bool,
+    pub should_retain: bool,
+}
+
+impl ResidentPrefixAllocation {
+    fn existing(seq_id: i32) -> Self {
+        Self {
+            seq_id,
+            evictions: Vec::new(),
+            should_save: false,
+            should_retain: true,
+        }
+    }
+
+    fn new_record(seq_id: i32, evictions: Vec<ResidentPrefixEviction>) -> Self {
+        Self {
+            seq_id,
+            evictions,
+            should_save: true,
+            should_retain: true,
+        }
+    }
+
+    fn uncacheable() -> Self {
+        Self {
+            seq_id: -1,
+            evictions: Vec::new(),
+            should_save: false,
+            should_retain: false,
+        }
+    }
 }
 
 #[derive(Debug, Clone, Copy, Default)]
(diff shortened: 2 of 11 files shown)


michaelneale:
@ndizazzo nice I think I was battling that last night. Want me to test it on my setup?

ndizazzo:
@michaelneale yes please!

michaelneale:
Tested this PR locally on M4 Pro against `Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m` with goose as the harness. Mixed results — fixes the simple smoke but does not fix the actual tool-loop case.

**Repro:**

```bash
# start serve (debug build is fine)
./target/debug/mesh-llm serve \
  --model Qwen/Qwen2.5-3B-Instruct-GGUF@main:q4_k_m \
  --port 9447 --console 3232 --log-format json --auto --publish

# in another shell, hit it with goose using tool-triggering prompts
for i in 1 2 3 4 5 6 7 8 9 10; do
  prompts=("list files here" "what files are in this dir" "show me the README" "what is in Cargo.toml" "describe the project structure")
  p="${prompts[$((i % 5))]}"
  GOOSE_PROVIDER=openai \
  OPENAI_HOST=http://127.0.0.1:9447 \
  OPENAI_API_KEY=mesh \
  GOOSE_MODEL='Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m' \
    timeout 60 goose run -t "$p"
done
```

**Results:**

| scenario | this PR | main |
|---|---|---|
| 10× sequential `"hi $i"` (no tool use) | 10/10 pass | 0/10 pass |
| 1× cold tool-using prompt | ✅ passes (33s, lists files correctly) | fails |
| 10× sequential tool-using prompts | 1/10 pass (only run 1) | 3/20 pass |

After run 1 (which triggers grammar and uses tools), every subsequent tool-using run fails in ~3s with `Server error: skippy ABI call failed: RuntimeError: llama_decode failed`. Native log at `~/.mesh-llm/runtime/<pid>/logs/skippy-native.log` shows one fresh `decode: failed to find a memory slot for batch of size 2048` per failed request.

Config from native log on the failing process:

```
llama_context: n_ctx         = 32768
llama_context: n_batch       = 2048
llama_context: n_seq_max     = 4
llama_context: kv_unified    = true
```

Process stays alive through all failures — it's not a crash, just every tool-using request after the first returns the ABI error until restart. The eviction is firing (PR's `evict_one_lru_entry` is reachable), it's just not enough headroom once a tool-call grammar fork has happened.

This may be the "session reset bookkeeping drift" path you call out in the PR description — the proactive eviction reclaims one prefix slot but something else accumulates and isn't released between turns. Or 1 LRU entry per gen isn't enough headroom for the grammar fork pattern.

Happy to test follow-up changes against the same repro.

ndizazzo:
@michaelneale I don't think just one PR alone will do it - I think the default 1024 PR from @IvGolovach PLUS this one might give you more solid results when combined

ndizazzo:
@michaelneale I've merged his PR into this one now - should be an easier 1 shot combined test that way

michaelneale:
Got to the root cause. Posting findings here.

I re-ran on the latest `fix/kv-proactive-eviction` head (`b99bcc1e`, with #659 merged in) — same outcome as before: 1/10 cold pass, 9/10 fail with `decode: failed to find a memory slot for batch of size 1024`. Same `Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m` + goose repro from #652.

I traced the failure end-to-end. The actual cause is in `drop_session_timed` (`crates/skippy-server/src/runtime_state.rs`), specifically the trim+preserve resident-prefix path. **The proactive LRU eviction in this PR helps, but it's downstream of the real leak.** Reasoning:

1. **`session_label` is regenerated server-side per request** (`now_unix_millis()` + `OPENAI_GENERATION_COUNTER`, in `crates/skippy-server/src/frontend.rs:839`). The client never sees or controls it — every OpenAI request gets a brand-new session label, so no future request can match a previously-preserved session.

2. **`take_idle_session()` prefers idles WITHOUT a resident_prefix**, only falling back to ones with prefixes when no clean idle exists (`runtime_state.rs:450`). Combined with #1, this means accumulated resident-prefix lanes are almost never reused.

3. **`drop_session_timed`'s `trim+preserve` branch keeps lane KV cells claimed in the unified pool** after a request finishes — intending to let a future request skip prefill. Given #1 and #2, that future request never comes. The cells stay pinned to the idle lane's seq_id until restart.

4. **Goose's ~13.7k-token system prompt + tool defs gets pinned per session.** I traced `idle_resident_total_tokens` across two goose runs and got `0 → 13722 → 27517` (claimed cells in the unified pool from idle lanes alone). With `n_ctx = 32768`, the next prefill needs ~13700 contiguous cells but only ~5251 are unclaimed → `find_slot` fails. This is independent of `n_batch` size (also fails at 1024), and independent of whether the resident-prefix cache layer evicts entries (the cache and the actual idle-lane cell claims are tracked separately).

To verify: I patched `drop_session_timed` to always go down the `reset()` path instead of `trim+preserve`. On the same repro the goose loop is now **15/15 pass**, 17 grammar triggers, 0 memory-slot failures. Diff is ~40 lines and just collapses the trim+preserve branch into the existing reset branch — the same `idle_sessions.push(lane_session)` after `reset()` succeeds.

I think the right next step is to either:
- **Always reset on drop** (the probe fix — kills the optimization but eliminates the leak), or
- **Only retain on drop when `acquire_resident_prefix_lane` has actually fired for the same page_id at least once** (preserves the optimization for clients that ARE reusing prefixes, but stops the unbounded accumulation for clients like goose that aren't), or
- **Cap total preserved idle resident-prefix cells in `RuntimeState` itself** (separate from `ResidentPrefixCache.max_resident_tokens` which only caps the cache's own bookkeeping, not the actual idle-lane KV claims — these are tracked by different objects).

Happy to open a PR with the probe fix if useful, or to defer to whatever direction you and Ivan want. The proactive eviction in this PR is independently a good thing — it just doesn't address the leak source.

michaelneale:
Filed #662 — while probing this PR I found that the prefix cache misses on "same long shared prefix, different tail" prompts even when the prefix is ~2200 identical tokens (gets ~256 cached out of 2240). Pure curl repro, no agent involved. That's why repeated goose turns still re-prefill 2000+ tokens each — and those fresh prefills are what then exhaust the unified KV pool under the failure in #652.

So #661's proactive eviction and #659's batch reduction are real improvements but cap-limited by how often we have to re-prefill the same content. Probably worth resolving #662 before deciding on further mitigations here.

i386:
I think this PR is a useful improvement, but the main KV pressure path from the tool-loop repro still looks outside the thing being evicted here.

`record_resident_prefix()` still calls `retain_resident_prefix_on_drop()`, and then `drop_session_timed()` preserves that active lane by trimming it and pushing it into `idle_sessions` with `resident_prefix` set. Those preserved idle lanes keep KV cells pinned, but the new proactive eviction only evicts entries from `ResidentPrefixCache`; it does not evict the idle-lane resident prefixes in `RuntimeState`.

Relevant spots:

- `record_resident_prefix()` retains the prefix on drop: `crates/skippy-server/src/kv_integration/resident_prefix.rs:236`
- `drop_session_timed()` preserves the lane instead of resetting it: `crates/skippy-server/src/runtime_state.rs:368`
- `take_idle_session()` prefers clean idle lanes, so preserved resident lanes are not naturally drained first: `crates/skippy-server/src/runtime_state.rs:493`
- the new proactive eviction only touches `self.resident`, not `runtime.idle_sessions`: `crates/skippy-server/src/kv_integration/resident_prefix.rs:159`

That matches the repro already discussed here: tool-loop requests can still fail after the first run because large preserved idle prefixes accumulate outside the resident-cache accounting. I would treat that as still blocking for the stated bug. The next fix probably needs to either reset on drop for these OpenAI request sessions, cap/evict preserved idle resident-prefix lanes in `RuntimeState`, or only preserve when the idle resident-prefix path is actually reusable.

michaelneale:
Pushed `688c8c8c` to this branch — `fix(kv): always reset lane in drop_session_timed`. Sits alongside the eviction + prefix-cache work here.

It removes the lane-side trim+preserve path in `drop_session_timed` which was pinning KV cells on idle lanes hoping for an exact `(page_id, token_count)` match on the next request. Real workloads (goose, chat with varying tails) almost never satisfy that exact-match condition, so the pinned cells just leak in the unified pool in parallel with the cache layer's own cell claims.

Validation on top of your current head:
- Goose loop (`Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m`, local-only, 20× sequential tool-using prompts): **20/20 pass, 0 memory-slot failures, 21 grammar triggers**.
- `cargo test -p skippy-server --lib`: 82 pass
- fmt + clippy clean

Happy to revert if you'd rather keep this PR scoped to your eviction work and land the drop-fix separately — let me know. Filed originally as #663; closing that in favor of this combined branch.

PR #663, opened 2026-05-24
fix(kv): page-aligned prefix cache + drop lane reset (#652)

### LINKED ISSUE #652: skippy: 'decode: failed to find a memory slot for batch of size 2048' on goose/pi after a few tool-laden turns

### LINKED ISSUE #662: Prefix cache misses on same-prefix-different-tail prompts
## Summary

Skippy's resident-prefix cache misses on "same long shared prefix, different tail" prompts, even when the shared prefix is hundreds or thousands of tokens. For typical chat/agent workloads where every turn re-sends the same system prompt + tools and only varies the tail, the cache returns ~256 cached tokens out of ~2200 identical tokens.

## Repro (no agent harness, pure curl)

Start `mesh-llm serve --model Qwen/Qwen2.5-3B-Instruct-GGUF@main:q4_k_m --port 9447 --console 3232 --auto`, then:

```bash
cat > /tmp/body.py <<'PY'
import json, sys, time
sys_blob = ("You are a helpful assistant. Be concise but thorough. " * 200).strip()
body = {
    "model": "Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m",
    "messages": [
        {"role": "system", "content": sys_blob},
        {"role": "user", "content": f"Say hi. Current timestamp: {time.time()}"},
    ],
    "max_tokens": 30,
    "temperature": 0.0,
}
sys.stdout.write(json.dumps(body))
PY

for i in 1 2 3 4 5; do
  python3 /tmp/body.py > /tmp/body.json
  resp=$(curl -sS http://127.0.0.1:9447/v1/chat/completions \
    -H 'Authorization: Bearer mesh' -H 'Content-Type: application/json' \
    --data-binary @/tmp/body.json)
  echo "prompt=$(jq -r .usage.prompt_tokens <<<"$resp") cached=$(jq -r '.usage.prompt_tokens_details.cached_tokens // 0' <<<"$resp")"
done
```

Result:
```
prompt=2240 cached=0
prompt=2240 cached=256
prompt=2240 cached=256
prompt=2240 cached=256
prompt=2240 cached=256
```

A byte-identical body (no timestamp variation) gets `cached=2214` on the same harness. So the cache works for exact-match, but degrades to ~10% hit rate the moment the tail differs by a handful of tokens, even though ~2200 tokens of the prefix are byte-identical.

## Why

`PrefixCandidatePolicy::candidate_token_counts` (in `crates/skippy-cache/src/config.rs`) generates the candidate set the cache probes/records on. For runtime defaults (`min_tokens=256`, `stride_tokens=128`, `page_size_tokens=256`), with `record_limit=2` it stores at:
- the full prompt length
- `min_tokens` (256)

The lookup side, for a *different* prompt length, walks down from its own length in stride steps. So:
- Run 1 (prompt=2214) stores at `[2214, 256]`
- Run 2 (prompt=2231) looks up at `[2231, 2230, 2102, 1974, ..., 384, 256]`

`2214` is not in run 2's lookup grid. Only `256` is. The cache hit collapses to the floor record.

This is independent of how much prefix is actually shared between the two prompts — the candidate sets are derived from each request's own length without coordination.

For comparison, upstream `llama-server` computes `slot.prompt.tokens.get_common_prefix(input_tokens)` — a linear token-by-token scan of the slot's existing prompt against the new prompt to find the longest common prefix. No hashing, no strides, no candidate lists. See `.deps/llama.cpp/tools/server/server-context.cpp:2551` and `.deps/llama.cpp/tools/server/server-common.cpp:462`.

## Why this matters

Every chat/agent workload re-sends the same system prompt + tool definitions on every turn, with only the conversation tail differing. With this miss pattern, each turn re-prefills 2000+ tokens that should have hit. That re-prefill consumes fresh cells in the unified KV pool, which is also at the heart of #652 (decode-fails-to-find-memory-slot under repeated goose-style requests).

I haven't tried to design a fix yet — wanted to file the observation first since it's structural. Linking #661 (proactive prefix eviction) which is where I noticed this while probing the related KV-exhaustion symptom.

Repo'd on `test-661` (current head of #661, includes #659). Same pattern visible on `main`.

michaelneale:
Resolves #652 (`decode: failed to find a memory slot for batch of size N` under repeated tool-using agent requests). Resolves #662 (prefix cache misses on same-prefix-different-tail prompts).

Two related changes that are independently meaningful but additively give the goose-loop result:

## 1. Page-aligned prefix cache candidates

`PrefixCandidatePolicy::candidate_token_counts` previously walked down from `token_count` in stride steps. Two prompts sharing a long prefix but with different total lengths landed on disjoint candidate sets and the cache structurally missed — even when 2000+ leading tokens were byte-identical (the dominant chat/agent pattern: same system prompt + tools, different conversation tail).

Replace with absolute page-aligned boundaries: any two prompts that share the first N*P tokens both probe and record at multiples of P that fit in N*P, so their candidates intersect regardless of length.

## 2. Always reset lane on drop_session_timed

The `trim+preserve` branch pinned lane KV cells on idle lanes intending future-request reuse via `acquire_resident_prefix_lane`. But `session_label` is server-side regenerated per request (frontend.rs ~839) so this never matches. The pinned cells accumulated and starved the pool.

Always reset on drop. Prefix reuse across requests stays the cache layer's job (matched by content `page_id`, not `session_id`).

## Validation

Goose loop on `Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m`, M4 Pro, 10× sequential `goose run -t "<tool-prompt>"`:

| config | pass | fail | slot_fails |
|---|---|---|---|
| main | 1/10 | 9/10 | 9 |
| drop fix only | 7/10 | 3/10 | 3 |
| cache fix only | 1/10 | 9/10 | 9 |
| both fixes | **10/10** | 0 | 0 |

Pure curl repro with long shared system prompt + varying user-message tail:
- Before: `cached=256` of 2240 (~10%)
- After: `cached=2048` of 2240 (~92%)
- Byte-identical case: `cached=2214/2215` (preserved)

30 consecutive goose runs with real tool execution, 23 grammar triggers, process stable for 14+ minutes.

- `cargo test -p skippy-cache --lib` ✓
- `cargo test -p skippy-server --lib` ✓
- `cargo clippy -p skippy-cache -p skippy-server --all-targets -- -D warnings` ✓

## Relation to other in-flight PRs

- **#661** (Nick — proactive prefix eviction): cap-limited mitigation, doesn't address either root cause. Eviction helps under pressure; this PR removes much of the pressure source.
- **#659** (Ivan — `n_batch=1024` for multi-lane): also cap-limited, the failure isn't about batch contiguity. With the fixes here, batch size effectively doesn't matter for the goose path.
- Nick's `fix/kv-proactive-eviction` head has a more recent commit (`a5a8a26b`) that does page-grid alignment with a slightly different stride choice (stride-based, mine is page-size-based). Convergent design.

I think both this and #661 could merge — they're complementary. Happy to drop the cache change in favor of Nick's grid alignment, or to PR these to his branch instead, whichever is cleaner.

### DIFF
--- a/crates/skippy-cache/src/config.rs
+++ b/crates/skippy-cache/src/config.rs
@@ -122,19 +122,34 @@ impl PrefixCandidatePolicy {
         if token_count == 0 {
             return Vec::new();
         }
-        let mut counts = vec![token_count];
-        if self.min_tokens == 0 || token_count <= self.min_tokens {
-            return counts;
+        // Candidates are page-aligned at absolute token positions so two
+        // prompts that share a long prefix land on the same candidate set
+        // regardless of their total length. Previously the candidates were
+        // computed by walking down from `token_count` in stride steps,
+        // which only matched between identical-length prompts; "same
+        // system prompt + different conversation tail" workloads (typical
+        // for chat and agent harnesses) cache-missed despite sharing
+        // thousands of identical leading tokens. See #662.
+        let page = self.page_size_tokens.max(1);
+        let min_t = self.min_tokens.max(page);
+        if token_count < min_t {
+            return vec![token_count];
         }
-        let stride = self.stride_tokens.max(1).min(self.page_size_tokens.max(1));
-        let mut candidate = token_count.saturating_sub(1);
-        while candidate >= self.min_tokens {
-            counts.push(candidate);
-            if candidate == self.min_tokens {
+        let mut counts: Vec<u64> = Vec::new();
+        let floor = (token_count / page) * page;
+        if floor < min_t {
+            return vec![token_count];
+        }
+        if token_count > floor {
+            counts.push(token_count);
+        }
+        let mut c = floor;
+        while c >= min_t {
+            counts.push(c);
+            if c == min_t {
                 break;
             }
-            let next = candidate.saturating_sub(stride);
-            candidate = next.max(self.min_tokens);
+            c = c.saturating_sub(page).max(min_t);
         }
         counts.sort_unstable_by(|a, b| b.cmp(a));
         counts.dedup();
@@ -209,10 +224,7 @@ mod tests {
             page_size_tokens: 64,
         };
 
-        assert_eq!(
-            policy.candidate_token_counts(160),
-            vec![160, 159, 127, 95, 64]
-        );
+        assert_eq!(policy.candidate_token_counts(160), vec![160, 128, 64]);
     }
 
     #[test]
@@ -251,7 +263,7 @@ mod tests {
 
         assert_eq!(
             policy.record_candidate_token_counts(160),
-            vec![160, 159, 127, 95, 64]
+            vec![160, 128, 64]
         );
     }
 }
--- a/crates/skippy-cache/src/config.rs
+++ b/crates/skippy-cache/src/config.rs
@@ -122,14 +122,17 @@ impl PrefixCandidatePolicy {
         if token_count == 0 {
             return Vec::new();
         }
-        // Candidates are page-aligned at absolute token positions so two
-        // prompts that share a long prefix land on the same candidate set
-        // regardless of their total length. Previously the candidates were
-        // computed by walking down from `token_count` in stride steps,
-        // which only matched between identical-length prompts; "same
-        // system prompt + different conversation tail" workloads (typical
-        // for chat and agent harnesses) cache-missed despite sharing
-        // thousands of identical leading tokens. See #662.
+        // Preserve the legacy `min_tokens == 0` semantics: return only
+        // the exact length. Walking a per-token grid here would generate
+        // O(token_count) candidates per lookup and blow up allocations.
+        if self.min_tokens == 0 {
+            return vec![token_count];
+        }
+        // Page-aligned absolute boundaries so two prompts sharing a long
+        // prefix land on the same candidate set regardless of total length.
+        // See #662 for the prior stride-from-prompt-length scheme that
+        // structurally missed on agent/chat workloads (same system prompt,
+        // varying conversation tail).
         let page = self.page_size_tokens.max(1);
         let min_t = self.min_tokens.max(page);
         if token_count < min_t {
--- a/crates/skippy-server/src/kv_integration/identity.rs
+++ b/crates/skippy-server/src/kv_integration/identity.rs
@@ -180,7 +180,8 @@ mod tests {
             .map(|identity| identity.identity.token_count)
             .collect::<Vec<_>>();
 
-        assert_eq!(lookup_counts, vec![160, 159, 127, 95, 64]);
+        // Page-aligned candidates: floor(160/64)*64 = 128.
+        assert_eq!(lookup_counts, vec![160, 128, 64]);
         assert_eq!(record_counts, vec![160, 64]);
     }
 }
--- a/crates/skippy-server/src/runtime_state.rs
+++ b/crates/skippy-server/src/runtime_state.rs
@@ -355,15 +355,21 @@ impl RuntimeState {
         if let Some(mut lane_session) = self.sessions.remove(session_id) {
             let lane_index = lane_session.index;
             // Always release the lane's native KV cells back to the
-            // unified pool. The trim+preserve path was intended to let
-            // a future request with the same session_id reuse the cells
-            // via `acquire_resident_prefix_lane`, but session_labels are
-            // regenerated per request server-side (frontend.rs ~line 839),
-            // so the optimization can never actually match. Holding
-            // cells on idle lanes under that mistaken assumption starves
-            // the pool and produces `failed to find a memory slot`
-            // errors. Prefix reuse across requests is the cache layer's
-            // responsibility (matched by `page_id`, not `session_id`).
+            // unified pool. The trim+preserve path kept the lane's cells
+            // pinned to a specific (`page_id`, `token_count`) pair so a
+            // future request whose content prefix hashed to the *exact*
+            // same `page_id` AND same `token_count` could acquire the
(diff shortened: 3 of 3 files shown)


michaelneale:
Pushed `78fd8d64` addressing Copilot's first two comments (min_tokens=0 corner case + misleading drop_session_timed comment).

On the third comment — Copilot is also right and I should correct my own PR description.

The `cached=2048/2240` result I posted was actually run with `SKIPPY_KV_CACHE_SHARED_RECORD_LIMIT=32` via env override. With the default `record_limit=2` (which this PR doesn't change), the cache fix only contributes the **256-floor crossing** between different-length prompts. I retested the curl varying-tail case on this PR head with no env overrides: `cached=256/2240`, same as main.

Why this PR's combined goose-loop result is still 10/10 then: validated with isolation tests just now — **the drop fix alone is sufficient**. Drop fix only against goose (`Qwen/Qwen2.5-3B-Instruct-GGUF:q4_k_m`, 20× sequential `goose run` with tool prompts): **20/20 pass, 0 memory-slot failures**. The cache fix on top is structurally good (correct lookup grid for shared prefixes; preserves byte-identical exact-match) but doesn't measurably help goose's current pattern at `record_limit=2`.

I could:
1. **Drop the cache fix from this PR** and ship only the drop-session fix. Smaller, more honest scope. The cache lookup-grid change becomes a follow-up that's only useful once record-side can store at the page grid.
2. **Keep both** but call the cache change "structural lookup-grid alignment, no immediate cached-tokens gain until record-side is widened."
3. Bump record_limit too. Tried this — naively storing 50 nested page records per prompt at `n_ctx=32k` blows past the cell pool because the cache layer copies cells per entry (each save_resident_prefix allocates a distinct cache_seq_id with its own copy of token_count cells, so nested entries claim independent cells, not shared ones). That's a deeper architectural refactor.

Leaning toward option 1. WDYT?

i386:
@michaelneale would prefer option 1 also

michaelneale:
Pushing the drop-session reset fix to #661 as commit 688c8c8c, alongside Nick's eviction + prefix-cache work. The cache lookup-grid change is being deferred separately (#662) until record_limit can also be widened safely. Closing this in favor of the combined branch.

PR #665, opened 2026-05-24
fix(swift-sdk): produce versioned bundle layout for macOS xcframework

keranm:
fix(swift-sdk): produce versioned bundle layout for macOS xcframework
 
macOS requires framework bundles to use a Versions/A/ directory structure with top-level symlinks. Flat bundles — where Info.plist sits at the framework root — are rejected by the macOS dynamic linker at load time.

Both build scripts were creating flat bundles for all platforms, which works on iOS/simulator/Catalyst but breaks on macOS. This caused the xcframework to fail when consumed by an Xcode project.

Changes:
 - build-xcframework.sh: split create_framework() into a versioned path for macos and a flat path for iOS/sim/catalyst; Info.plist and PrivacyInfo.xcprivacy move into 
Versions/A/Resources/; add the five canonical Versions/Current symlinks; use cp -RP (not cp -R) in the xcodebuild fallback so symlinks are preserved
- build-host-macos-xcframework.sh: same versioned layout applied to the inline macOS framework construction

### DIFF
--- a/sdk/swift/scripts/build-host-macos-xcframework.sh
+++ b/sdk/swift/scripts/build-host-macos-xcframework.sh
@@ -96,21 +96,26 @@ PY
 
 FRAMEWORK_DIR="$TARGET_DIR/frameworks/macos-host/$FRAMEWORK_NAME.framework"
 rm -rf "$FRAMEWORK_DIR"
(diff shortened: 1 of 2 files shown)


i386 (approved):
LGTM and congrats on your first contribution @keranm!

PR #670, opened 2026-05-24
Add optional SDK console asset serving

i386:
## Summary

SDK users can now opt into serving the Mesh console from package-provided static assets instead of forcing every native SDK binary to embed the web UI by default.

This adds a reusable `mesh-llm-console-server` crate, generalizes `mesh-llm-ui` assets behind embedded and filesystem-backed providers, and exposes `startConsole` hooks through the Swift, Kotlin, and Node SDK surfaces.

## Details

- Adds `mesh-llm-console-server` for localhost static console serving from a validated asset directory.
- Keeps the existing CLI console path embedded and compatible.
- Adds UniFFI, Swift, Kotlin, and Node bindings for optional console startup and shutdown.
- Documents the intended packaging shape: console-enabled SDK packages can include built UI resources and resolve their internal resource path before calling the native hook.
- Updates CI crate lists and repo-consistency expectations for the new crate.

## Caveat

This PR establishes the shared native/static serving layer and SDK API hooks. It does not yet package built console `dist/` assets into every SDK release artifact, and it does not turn the SDK console into a full CLI management API replacement. The static console can be shipped by console-enabled SDK packages without bloating the default native SDK libraries.

## Validation

- `cargo check -p mesh-llm-console-server`
- `cargo test -p mesh-llm-console-server`
- `cargo test -p mesh-llm-ui`
- `cargo check -p mesh-llm-ffi`
- `cargo test -p mesh-llm-ffi --lib`
- `cargo check -p mesh-llm-nodejs`
- `cargo check -p mesh-llm`
- `scripts/check-sdk-contract.sh`
- `./sdk/swift/scripts/build-host-macos-xcframework.sh`
- `swift test` in `sdk/swift`
- `scripts/verify-swift-release-artifact.sh` against a zipped local `MeshLLMFFI.xcframework`
- `./gradlew test` in `sdk/kotlin`
- `npm test` in `sdk/node`
- `git diff --check`


ndizazzo (commented):
Agent led review caught the following: 

---

> 1. Medium — SDK console breaks configuration deep links  
crates/mesh-llm-console-server/src/lib.rs only serves index.html for /, /dashboard, and /chat; direct SDK-console opens/refreshes of /configuration or /configuration/... return 404 before React boots. Existing host API treats those paths as console index routes and has tests for configuration deep links. Fix by sharing/synchronizing the SPA route predicate with host/UI routes.

I think this was the same issue in [this cmt](https://github.com/Mesh-LLM/mesh-llm/pull/675#discussion_r3300042916)

> 2. Medium — filesystem assets can escape root via symlinks  
crates/mesh-llm-ui/src/lib.rs cleans literal .. but then reads root.join(rel) directly. A symlink inside the asset directory can point outside the console bundle and be served, especially risky with listenAll: true. Fix by canonicalizing root and target and checking containment, or rejecting symlinks.

Can't hurt to sanitize these, even though issue probability is really low.

> 3. Low/Medium — request parser assumes one TCP read contains the request line  
crates/mesh-llm-console-server/src/lib.rs does one 16 KiB read() and parses only that chunk. Valid HTTP requests split before \r\n or with larger headers can be misparsed/dropped. Fix by buffering until at least the header terminator with a size cap/timeout, or reuse the existing host request-reader pattern.

Not really too sure about this one, I doubt we'd hit the issue. Take or leave!

PR #675, opened 2026-05-25
Add shared config.toml crate for SDKs

i386:
## Summary
- add `mesh-llm-config` as the shared owner for MeshLLM `config.toml` schema, path resolution, typed load/save, TOML parse/serialize helpers, and validation
- add a high-level authoring API so SDKs and apps can configure local nodes, models, request defaults, and plugins without writing TOML directly
- move runtime selection to `mesh_llm_types::runtime::ModelRuntimeKind` instead of a stringly `hardware.model_runtime` field
- keep runtime/plugin interpretation in `mesh-llm-host-runtime`, while re-exporting the config surface through the existing plugin facade and `sdk::config`
- route runtime config apply, CLI config parsing, protocol config snapshots, and SDK config access through the shared crate

## Examples
Configure a local serving node from an SDK or desktop app:

```rust
use mesh_llm_config::{ConfigStore, GpuAssignment, LocalServingNodeConfig};
use mesh_llm_types::runtime::ModelRuntimeKind;

let store = ConfigStore::default_path()?;
store.update(|config| {
    config.configure_local_serving_node(LocalServingNodeConfig {
        model: "Qwen/Qwen3-8B-GGUF:Q4_K_M".into(),
        runtime: Some(ModelRuntimeKind::Metal),
        device: Some("metal:0".into()),
        context_size: Some(8192),
        parallel: Some(2),
        gpu_assignment: Some(GpuAssignment::Auto),
        owner_control_bind: Some("127.0.0.1:0".parse()?),
        ..LocalServingNodeConfig::default()
    })?;
    Ok(())
})?;
```

Add model request defaults:

```rust
let store = ConfigStore::default_path()?;
store.update(|config| {
    config
        .upsert_model("Qwen/Qwen3-8B-GGUF:Q4_K_M")?
        .max_tokens(1024)
        .temperature(0.2);
    Ok(())
})?;
```

Enable plugins without writing `[[plugin]]` TOML by hand:

```rust
let store = ConfigStore::default_path()?;
store.update(|config| {
    config.enable_builtin_plugin("telemetry")?;
    config.upsert_openai_endpoint_plugin("http://localhost:8000/v1")?;
    config.upsert_external_plugin("custom-tool", "mesh-tool", ["--serve"])?;
    Ok(())
})?;
```

## Architecture
`mesh-llm-config` owns filesystem config correctness: schema, parsing, validation, atomic writes, and typed authoring helpers. Host runtime still owns plugin resolution, live config revisions, process lifecycle, and serving behavior.

Protocol config snapshots now call `config_to_toml` / `parse_config_toml` from the shared crate so serialized full-config payloads use the same validation path as file-backed config.

## Validation
- `cargo test -p mesh-llm-config`
- `cargo check -p mesh-llm`


crates/mesh-llm-config/src/model.rs
```diff
@@ -0,0 +1,867 @@
...
+            model_fit,
+            hardware,
+            throughput,
+            skippy: raw.skippy,
+            speculative: raw.speculative,
+            request_defaults: raw.request_defaults,
+            multimodal,
+            advanced: raw.advanced,
+            gpu_id_from_legacy_shim,
+        }
```
ndizazzo, line 806: The new `hardware.device` gets copied into `gpu_id` while loading, but then we write that copied value back out as the legacy `gpu_id` when it's getting saved. Valid configs using `gpu.assignment = "auto"` with `hardware.device = "cuda:0"` would be invalid after a save/reload cycle... We should probably skip writing `gpu_id` here with the right checks.

ndizazzo (commented):
I like this idea! I have some work around it coming up for plugins that this makes sense to do now. 

One thing, you need to add the new crate to the docker COPY phases (in two dockerfiles) - otherwise the build will fail for the new `crates/mesh-llm-console-server`

crates/mesh-llm-config/src/model.rs
i386: Fixed in a58ba01c7. The serializer now only writes legacy `gpu_id` when the field actually came from an explicit legacy `gpu_id` input; a nested `hardware.device` value is still mirrored in memory for compatibility, but no longer serializes back out as `gpu_id`, so `gpu.assignment = "auto"` + `hardware.device = "cuda:0"` survives save/reload.

I added regression coverage for both sides:
- nested `hardware.device` does not serialize as `gpu_id` and re-parses successfully
- explicit legacy `gpu_id` still serializes for legacy round-trips

I also added the missing `mesh-llm-config` COPY to `ci/linux-test.dockerfile`; the existing Docker runtime files already had it.

Validation:
- `cargo test -p mesh-llm-config`
- `cargo check -p mesh-llm`

i386:
@ndizazzo thanks for the review. Looking forward to plugin changes :)

ndizazzo:
>  thanks for the review. Looking forward to plugin changes :)

@i386 the general idea is that plugins can "provide" settings that get mounted into the configuration file, so you can have:

1) TOML control via `config.toml`
2) UI controls for it automatically displayed in the UI

This crate will def. make that simpler to work with.

fly/Dockerfile
```diff
@@ -38,6 +38,7 @@ COPY crates/mesh-llm-routing/ crates/mesh-llm-routing/
 COPY crates/mesh-llm-guardrails/ crates/mesh-llm-guardrails/
 COPY crates/mesh-llm-system/ crates/mesh-llm-system/
 COPY crates/mesh-llm-types/ crates/mesh-llm-types/
+COPY crates/mesh-llm-config/ crates/mesh-llm-config/
```
ndizazzo, line 41: Also need `crates/mesh-llm-console-server`

ci/linux-test.dockerfile
```diff
@@ -28,6 +28,7 @@ COPY crates/mesh-llm-routing/ crates/mesh-llm-routing/
 COPY crates/mesh-llm-guardrails/ crates/mesh-llm-guardrails/
 COPY crates/mesh-llm-system/ crates/mesh-llm-system/
 COPY crates/mesh-llm-types/ crates/mesh-llm-types/
+COPY crates/mesh-llm-config/ crates/mesh-llm-config/
```
ndizazzo, line 31: also need `crates/mesh-llm-console-server`

.github/workflows/docker-precheck.yml
```diff
@@ -68,6 +68,7 @@ jobs:
             crates/mesh-llm-routing/ \
             crates/mesh-llm-guardrails/ \
             crates/mesh-llm-types/ \
+            crates/mesh-llm-config/ \
```
ndizazzo, line 71: Need `crates/mesh-llm-console-server`

crates/mesh-llm-config/src/model.rs
```diff
@@ -0,0 +1,930 @@
...
+    pub shared_record_limit: Option<u32>,
+    #[serde(default)]
+    pub payload_mode: Option<String>,
+}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct HardwareConfig {
+    #[serde(default)]
+    pub model_runtime: Option<ModelRuntimeKind>,
```
ndizazzo, line 234: This changes `model_runtime` from a free-form string to `ModelRuntimeKind`, but the enum only deserializes lowercase values. 

Existing configs that currently pass validation with values like `"CUDA"` or `"Metal"` will now fail before validation runs... we might want to normalize the incoming value

crates/mesh-llm-console-server/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+    } else {
+        respond_text(&mut stream, 404, "Not Found", "not found").await?;
+    }
+    Ok(())
+}
+
+fn is_index_route(path: &str) -> bool {
+    matches!(
+        path,
+        "/" | "/dashboard" | "/dashboard/" | "/chat" | "/chat/"
```
ndizazzo: It looks like this route list is missing a few entry points that the embedded console already serves, like `/configuration`, `/configuration/*`. Unless you're intentionally keeping those out of this crate, console will hit 404s when refreshing or deep-linking to those pages

ndizazzo (changes requested):
Few more solid fixups and then ready to go!

PR #685, opened 2026-05-26
MoA: streaming progress + progressive paint

michaelneale:
Streaming MoA used to spin for ~3s and then drop the whole answer at once. Now:

1. **Headers go out immediately** (~2ms) so the client knows it's connected.
2. **Progress drips into the thinking pane** every 1s while the arbiter waits — "Routing through mesh…", "Querying peer models…", "Comparing responses…" — using `delta.reasoning_content` for chat-completions and `response.reasoning_text.delta` for Responses-API. Clients that don't have a thinking pane ignore the field; clients that do (goose, mesh-llm web UI, OpenAI SDK) surface it separately so it doesn't pollute the final answer.
3. **Final answer paints progressively** over ~500ms (~25 word-boundary chunks at 20ms gap) once the arbiter commits, instead of arriving as one blob. Tool-call deltas stay atomic.

## Wire (M4 against real mesh, /v1/responses streaming)

| | main | this branch |
|---|---|---|
| Time to headers | ~3200ms | **2ms** |
| Time to first paint | ~3200ms | **1180ms** (reasoning progress) |
| Reasoning-text deltas | 0 | 3 |
| Output-text deltas | 1 (one blob) | 22 (progressive) |
| Total wall time | ~3200ms | ~3640ms |

The extra ~450ms is the artificial chunk-paint delay; it's what makes the answer feel like it's streaming. MoA arbiter latency unchanged.

## Verified

- Goose chat-completions: tool calls atomic, content streams correctly, reasoning routed to thinking pane.
- Pi chat-completions: streams correctly, gracefully ignores `reasoning_content`.
- mesh-llm web console: progress shows in thinking pane (collapsible), real answer in main bubble, no concatenation.

## Architecture

`run_moa_turn_with_progress` (new path) sends HTTP headers up front, then races `moa::handle_turn` against a 1-second ticker. Each tick writes one progress line via `delta.reasoning_content` (chat) or `response.reasoning_text.delta` (Responses). On MoA completion, the existing chunked-paint writers take over with `header_already_sent = true`.

Cancellation-safe: TCP writes happen outside the `tokio::select!` so a partial heartbeat write can't be cut off by MoA's completion (which would corrupt the stream).

Trade-off: HTTP headers must precede the body, so this path loses the post-hoc `x-moa-*` observability headers. Worth it for the live feel; we can revisit with HTTP trailers later if needed.

## Constants, not flags

- `MOA_PROGRESS_INTERVAL = 1000ms` — heartbeat cadence
- `DEFAULT_MOA_STREAM_CHUNKS = 25` — winner-paint chunk count
- `DEFAULT_MOA_STREAM_CHUNK_DELAY_MS = 20` — chunk gap

No env vars. If these need tuning later we can add knobs; defaults landed empirically.

## Tests

48 pass in `moa_gateway.rs`:
- 44 pre-existing (regression guard for SSE shapes, tool atomicity, response-id correlation)
- 4 new for progress drip:
  - `progress_line_steps_through_lines_then_stabilises`
  - `progress_event_chat_uses_reasoning_content_field`
  - `progress_event_responses_uses_reasoning_text_delta`
  - `failure_tail_emits_error_then_done_for_chat_adapter`

Refs #618.


i386 (changes requested):
Found a few issues that should be addressed before merge:

- [P1] `crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs:1466`: The `/v1/responses` progress path resets Responses stream state after progress. `write_progress_response_created` emits `response.created` with one `created_at`, progress deltas use sequence numbers `1..N`, then `send_moa_as_responses_sse_inner` recomputes `created_at` and uses helpers that hard-code `sequence_number: 0` for every output/done/completed event. A slow MoA response will produce non-monotonic duplicate sequence numbers and mismatched `created_at` for the same response id, which can break strict Responses clients. Carry `created_at` and the next sequence number from the progress phase into the body writer and use the `_with_sequence` helpers for the final events.

- [P2] `crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs:377`: If a progress write fails, the code logs “falling back” and then waits for `moa_fut` to finish. A write failure here normally means the client disconnected, so this keeps the full MoA worker/reducer workload running for a dead request, potentially up to the 60s timeouts. Drop/cancel the MoA future on progress write failure instead of awaiting it.

- [P3] `crates/mesh-llm-host-runtime/src/network/openai/moa_gateway.rs:1526`: This PR grows `moa_gateway.rs` from 1,478 to 2,663 lines, over the repo’s “no Rust source files over 2,000 lines” rule. The new progress streaming/chunking code and its tests are a natural extraction candidate.

PR #696, opened 2026-05-26
fix(lint): many linter corrections

ndizazzo:
Follow up from #693


michaelneale (approved):
only merge when all green

PR #697, opened 2026-05-26
Relay-only heartbeat grace + robust path-set classification

michaelneale:
Two small, independent fixes — one commit each.

## 1. Relay-only heartbeat: bigger grace, robust path detection, honest message

Observed today on the public mesh: mini's mesh-llm declared studio dead at 10:26:38 UTC after just 2 missed heartbeats, even though the relay-only policy is supposed to allow 3. The 60–90s that followed produced an HTTP 502 storm on openclaw (4 failed agent calls with `Reducer failed (tried 3)`), because MiniMax-M2.5 (studio) had been removed from the MoA eligible-peer set and the fallback to weaker peers also failed the reducer.

Root cause has three layers:

a. **Threshold of 3 isn't enough.** mini's path to studio is relay-only (mini's macOS VPN system extension blocks the LAN UDP hole-punch even though both peers are on the same Sydney subnet). Steady relay RTT is ~200ms; transient renegotiation spikes it past 10s. With 60s heartbeats and 3 misses, the grace window is ~3 min — and a single iroh relay path renegotiation can eat that. Bumped to **5 (≈5 min)**. Direct peers stay at 2 misses (a real network failure should still surface quickly).

b. **`is_relay_only` detection failed silently mid-failure.** The original check was `selected_path_snapshot(conn).kind == Relay`, but at the moment of heartbeat failure the connection is often between path selections and `selected_path_snapshot` returns `Unknown`. That made the lenient relay threshold *never fire during the failures it was designed to protect against*. New `is_relay_only_connection` looks at every advertised path: if none are IP, treat as relay-only. Also reverses the no-connection default (no live `Connection` object now defaults to relay-only / lenient, not strict).

c. **Misleading recovery message.** `💚 Heartbeat: <peer> recovered (was N/2)` was hardcoded — even a relay-only peer with threshold 3 (now 5) showed `/2`. Now reads the real threshold from the failure policy.

Tests:
- `relay_only_peers_get_extra_heartbeat_grace` (renamed; threshold now 5)
- `direct_peers_use_strict_heartbeat_threshold` (new)
- `is_relay_only_path_set_classifies_correctly` (new; `is_relay_only_connection` extracted as a thin wrapper over an iterator-based helper so it can be unit-tested without constructing an iroh `Connection`)

## 2. skippy-runtime: drop DEBUG lines from the native log callback

llama.cpp's `llama_log_internal_v` invokes the registered callback **unconditionally** with the level argument — level filtering is expected to live inside the callback. mesh-llm's callback (`write_native_log`) was ignoring `_level` and writing every line to the per-instance `skippy-native.log`, so `LLAMA_LOG_DEBUG` output ended up in the file regardless of `GGML_LLAMA_LOG_LEVEL`.

Observed impact: on studio today, **5,431 `Grammar still awaiting trigger after token …` DEBUG lines for 18 actual `<minimax:tool_call>` triggers** — about 300× verbosity per completion. These are per-token traces of MiniMax-M2.5's reasoning phase before the tool-call regex fires.

Fix: drop `ggml_log_level == DEBUG` (1) unless `GGML_LLAMA_LOG_LEVEL=4` (the existing opt-in set by `enable_verbose_native_logs`). Keep `CONT` (continuation) lines unconditionally so multi-line INFO/WARN messages don't get truncated.

Tests:
- `native_log_filter_drops_debug_by_default`
- `native_log_filter_keeps_debug_when_verbose`
- `native_log_filter_keeps_continuation_lines`

Env-var tests run serially via a per-test mutex to avoid racing other env-touching tests in the same module.

## Verification

- `cargo fmt --check`, `cargo check -p mesh-llm`, `cargo clippy -p mesh-llm-host-runtime -p skippy-runtime -- -D warnings` all clean.
- `cargo test -p mesh-llm-host-runtime --lib` and `cargo test -p skippy-runtime --lib` both pass for the new tests (one unrelated network-y test, `owner_control_client_reuses_connection_for_sequential_requests`, intermittently flakes on a busy machine; passes when re-run in isolation).
- Not yet deployed live — but the changes are mechanical and the unit tests cover the policy/classification logic exhaustively. Live verification would require reproducing today's relay flap, which is mini-specific.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/heartbeat.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/heartbeat.rs
@@ -21,9 +21,17 @@ pub(super) fn heartbeat_failure_policy_for_peer(
     let _ = peer;
     HeartbeatFailurePolicy {
         allow_recent_inbound_grace: true,
-        // Relay-only peers are more prone to transient timeouts (relay hiccups,
-        // higher base RTT). Give them an extra cycle before declaring death.
-        failure_threshold: if is_relay_only { 3 } else { 2 },
+        // Relay-only peers are far more prone to transient timeouts.
+        // Observed behaviour: a Sydney<->Sydney relay-only path (mini's VPN
+        // extension blocking the LAN UDP hole-punch) can spike from 200ms
+        // to 10s+ RTT during a single relay hiccup. With 60s heartbeat
+        // intervals, two such cycles is ~2min — not enough grace for the
+        // public mesh's relay to recover. Five cycles = 5min grace, which
+        // covers the typical iroh relay path-renegotiation window.
+        //
+        // Direct paths stay at 2 — when the LAN/internet path is up at
+        // all, two consecutive cycles of silence is a real failure signal.
+        failure_threshold: if is_relay_only { 5 } else { 2 },
     }
 }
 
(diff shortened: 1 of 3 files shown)


i386 (commented):
Two review notes from the diff:

- [P2] `crates/mesh-llm-host-runtime/src/mesh/heartbeat.rs`: `conn.as_ref().map(...).unwrap_or(true)` now treats every peer with no current `Connection` as relay-only, so it gets the new 5-miss threshold. That includes non-relay/direct or transitive peers that are simply disconnected but still present in `state.peers`, which can keep stale model routes alive for up to 5 minutes instead of 2. I would default no-connection to strict unless the peer was previously observed as relay-only.

- [P3] `crates/skippy-runtime/src/lib.rs`: the native log filter only honors `GGML_LLAMA_LOG_LEVEL=4` for DEBUG, but still passes INFO/WARN/ERROR when the env is `0`, `1`, or `2`. Since the callback is now responsible for filtering, `silent`, `error`, and `warn` settings remain ineffective. The filter should parse the threshold and apply it to all levels; `CONT` should ideally follow whether the previous base log line was emitted.

Validation note: `git diff --check` passed. I tried `cargo test -p skippy-runtime native_log_filter`, but it did not reach tests because this checkout has not prepared the llama.cpp static libraries: `could not find native static library llama-common`.

PR #699, opened 2026-05-26
Allow plugins to provide CLI commands

i386:
## Summary

This gives configured plugins access to the top-level CLI experience without moving the agent launchers out of the host runtime. One of the things that blocks plugins being first class is not having access to the cli experience; this adds a `cli.run` path so an installed/configured plugin can own commands such as `mesh-llm <plugin> ...`.

Blackboard is no longer bundled or auto-registered by mesh-llm. It is expected to ship from its own `mesh-llm/blackboard` repository and install through the plugin manager flow documented in #701 (`mesh-llm plugins install blackboard`).

## What changed

- Added plugin-backed CLI dispatch for unknown top-level commands backed by `[[plugin]]` config entries.
- Added `mesh-llm plugin mcp` as the generic stdio MCP entrypoint for configured plugin tools.
- Removed the in-repo blackboard plugin crate, blackboard host module, legacy `/api/blackboard/*` route aliases, blackboard integration test script, and `--blackboard` CLI flag.
- Removed blackboard from workspace members, build scripts, release packaging, Docker copy lists, CI artifacts, and crate-list plumbing.
- Kept Goose, Claude, Pi, and OpenCode under host-runtime CLI command handling, including the OpenCode JSONC config support.
- Added long-running plugin operation support for interactive CLI commands without applying the normal 30-second plugin request timeout.
- Synced the branch with latest `mesh/main` before review.

## CLI examples

External plugin command shape:

```toml
[[plugin]]
name = "example"
command = "mesh-llm-plugin-example"
```

```bash
mesh-llm example --help
mesh-llm example --some-plugin-flag value
```

Blackboard will use the same CLI path once installed/configured externally:

```bash
mesh-llm plugins install blackboard
mesh-llm blackboard --search "QUESTION"
```

Generic plugin MCP bridge:

```bash
mesh-llm plugin mcp
```

Agent launchers remain built-in host-runtime commands, not plugin examples:

```bash
mesh-llm opencode --write
mesh-llm goose --model Qwen3-Coder
```

## Validation

- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-host-runtime external_plugin_can_be_configured --lib`
- `cargo test -p mesh-llm-host-runtime unknown_top_level_command_is_captured_for_plugin_dispatch --lib`
- `cargo test -p mesh-llm-host-runtime plugin_mcp_swarm_capture_cli_creates_recorder --lib`
- `just build`
- Verified `target/debug/mesh-llm-plugin-blackboard` and `target/release/mesh-llm-plugin-blackboard` are not produced after `just build`.
- Smoke-tested `./target/debug/mesh-llm plugin --help`, `./target/debug/mesh-llm --help`, and the unconfigured `./target/debug/mesh-llm blackboard` error path.

## Compatibility

This intentionally removes mesh-llm's bundled blackboard runtime and legacy `/api/blackboard/*` aliases. Existing blackboard behavior should move to the external blackboard plugin repository and plugin-manager install flow.


michaelneale:
main concern is that when we know distribute - there is no single binary, if we want it to launchg with blackboard, which seems a problem? 

Also, practically would need to work out how to add other things when there are any: 
 │ Bundled-plugin registry is duplicated across ~13 places. Adding a second bundled plugin means touching Cargo.toml, Justfile, 5
 │ build scripts, 3 workflows, Dockerfile, bundled_cli_plugin_spec, plus affected-crates / clippy-batch lists. The CI failure that
 │ triggered commit cdc0083a is exactly this class of drift. Suggest adding an xtask repo-consistency bundled-plugins check that
 │ asserts all sites stay in sync (matches existing release-targets / ci-crate-lists pattern).
 │
 │ Bundle layout is a soft compat change. Release tarball now ships two binaries that must be co-located; manual cp mesh-llm
 │ /usr/local/bin/ without the sibling silently breaks mesh-llm blackboard. Worth a line in RELEASE.md and a clearer error than ENOENT
 │  when the sibling isn't found.
 
 
 but mostly it is the non monolithic distribution that is worrying, if we want to ship with blackboard (as I read it)

i386:
@michaelneale blackboard will be shipped in its own repo at [mesh-llm/blackboard](https://github.com/Mesh-LLM/blackboard) and will have a plugin manager that allows you to do this https://github.com/Mesh-LLM/mesh-llm/pull/701 

`mesh-llm plugins install blackboard`

michaelneale:
get this one in when a clean build 🚢

PR #701, opened 2026-05-26
Plugin management 

i386:
## Summary

- Adds the `mesh-llm-plugin-manager` install/update implementation for GitHub-backed native plugin releases and the Hugging Face `plugins.jsonl` catalog.
- Wires `mesh-llm plugins` commands for install, update, enable, disable, delete, info, search, and list with spinner/progress output.
- Loads enabled installed plugins at startup from the local plugin store and reports disabled or broken installs as inactive instead of failing unrelated startup paths.
- Keeps the documented native asset pattern target-triple only, with no GPU flavor suffixes.

## Plugin Catalog

- Created the public Hugging Face dataset: [`meshllm/plugin-catalog`](https://huggingface.co/datasets/meshllm/plugin-catalog).
- Uploaded the initial `plugins.jsonl` with `blackboard` as the first entry.
- Default catalog URL: `https://huggingface.co/datasets/meshllm/plugin-catalog/resolve/main/plugins.jsonl`.
- `MESH_LLM_PLUGIN_CATALOG_URL` can point at a staging catalog or fixture.

Catalog rows are JSONL with this shape:

```json
{"name":"blackboard","description":"Shared mesh blackboard for agent status, findings, questions, answers, and searchable coordination notes.","github_url":"https://github.com/mesh-llm/blackboard","author_email":"maintainers@meshllm.cloud","author_name":"Mesh LLM"}
```

## Command Examples

Catalog search and discovery:

```bash
mesh-llm plugins search
mesh-llm plugins search blackboard
mesh-llm plugins list
mesh-llm plugins info blackboard
```

Install from the catalog or GitHub:

```bash
mesh-llm plugins install blackboard
mesh-llm plugins install blackboard@1.1.0
mesh-llm plugins install mesh-llm/cool-plugin
mesh-llm plugins install mesh-llm/cool-plugin@1.1.0
mesh-llm plugins install https://github.com/mesh-llm/cool-plugin
mesh-llm plugins install https://github.com/mesh-llm/cool-plugin@1.1.0
```

Manage an installed plugin:

```bash
mesh-llm plugins update blackboard
mesh-llm plugins disable blackboard
mesh-llm plugins enable blackboard
mesh-llm plugins delete blackboard
```

Representative install output:

```text
⬇️  Downloading blackboard-0.1.2-aarch64-apple-darwin.tar.gz
   size: 2.7 MiB
✅ Downloaded blackboard-0.1.2-aarch64-apple-darwin.tar.gz
📦 Installed blackboard 0.1.2
✅ Installed blackboard 0.1.2
```

## Native Release Assets

Plugin releases are selected by plugin name, optional version, current target triple, and archive extension. GPU flavor suffixes are intentionally not part of the plugin asset pattern.

Examples:

```text
blackboard-1.1.0-aarch64-apple-darwin.tar.gz
blackboard-v1.1.0-x86_64-unknown-linux-gnu.tar.gz
blackboard-x86_64-pc-windows-msvc.zip
```

Each archive must contain `plugin.toml` and a native executable named after the plugin, for example `blackboard` or `blackboard.exe`.

## Startup Behavior

- Enabled installed plugins are added to the runtime plugin set automatically.
- Disabled installed plugins are listed as inactive.
- Broken installed plugins, such as missing executable files, are listed as inactive with an error instead of preventing unrelated plugin startup.
- Unit tests avoid reading the developer machine default plugin store unless `MESH_LLM_PLUGIN_DIR` is explicitly set.

## Validation

- `cargo test -p mesh-llm-plugin-manager`
- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-host-runtime plugin::tests::`
- `just build`
- `MESH_LLM_PLUGIN_DIR=$(mktemp -d) ./target/debug/mesh-llm plugins list`
- `MESH_LLM_PLUGIN_DIR=$(mktemp -d) ./target/debug/mesh-llm plugins search blackboard`


crates/mesh-llm-plugin-manager/src/archive.rs
```diff
@@ -0,0 +1,139 @@
...
+    let final_dir = install_dir.join(plugin_name);
+    if final_dir.exists() {
+        fs::remove_dir_all(&final_dir)
+            .with_context(|| format!("remove previous plugin install {}", final_dir.display()))?;
+    }
+    fs::create_dir_all(install_dir)
+        .with_context(|| format!("create plugin install dir {}", install_dir.display()))?;
+    fs::rename(&extracted_root, &final_dir)
+        .or_else(|_| copy_dir_and_remove(&extracted_root, &final_dir))?;
+    validate_installed_plugin(&final_dir, plugin_name)?;
```
ndizazzo: We should validate the incoming plugin before removing the existing install... otherwise if it fails we'll have lost the old one

crates/mesh-llm-host-runtime/src/plugin/config.rs
```diff
@@ -143,6 +147,115 @@ pub fn resolve_plugins(config: &MeshConfig, _host_mode: PluginHostMode) -> Resul
...
+            continue;
+        }
+        let spec = ExternalPluginSpec {
+            name: metadata.name.clone(),
+            command: command.display().to_string(),
+            args: Vec::new(),
+            url: None,
+            env: BTreeMap::new(),
+        };
+        externals.push(spec);
```
ndizazzo, line 193: Broken plugins will propagate from here into app startup (possibly stopping it) - we should build some kind of "failed" or incompatable list to display alongside installed so that we collect ones that can't specifically be loaded, and then drop some warnings on startup so the user is aware.

crates/mesh-llm-host-runtime/src/cli/commands/plugin_cli.rs
ndizazzo, line 128: Both of these  connect in a way that resolver will only check our configured plugins, then falls back to `bundled_cli_plugin_spec()`, which doesn't return anything now... I think we need to have `resolve_required_cli_plugin()` and `resolve_configured_cli_plugin()` check the installed plugin store too, not just configured plugins and bundled shims.

ndizazzo (changes requested):

crates/mesh-llm-plugin-manager/src/archive.rs
i386: Fixed in 813021792: extracted archives are now validated in staging before the previous install is removed. I also added a regression test that an invalid archive leaves the old install intact.

crates/mesh-llm-host-runtime/src/plugin/config.rs
i386: Fixed in 813021792: plugin load failures now become inactive error summaries and are logged as warnings instead of aborting app startup. The inactive summaries are included in the manager/list surfaces, with regression coverage for a failed external plugin launch.

crates/mesh-llm-host-runtime/src/cli/commands/plugin_cli.rs
i386: Fixed in 813021792: external CLI command resolution now checks the installed plugin store when a command is not configured before falling back to bundled shims. It also reports installed-but-missing executables clearly and has regression coverage for both paths.

PR #702, opened 2026-05-26
Add plugin mesh streams

i386:
## Summary

Adds the mesh-side stream primitive needed by external plugins to route agent traffic across the mesh.

Plugins can now ask mesh-llm to open a bidirectional byte stream to the same plugin on a remote peer. The host handles peer lookup, mesh stream negotiation, and side-stream bridging into the target plugin process. This keeps A2A/ACP agent behavior in the external `Mesh-LLM/agents` plugin while giving it a real mesh transport for streaming task execution.

## User Impact

This enables a mesh-launched plugin such as `agents` to:

- discover a remote A2A agent through plugin mesh channels,
- call `send_message` locally through mesh-llm's plugin/MCP surface,
- route the task to the owning peer over a mesh stream,
- stream A2A JSON-RPC/SSE bytes back to the caller,
- persist the final remote task result for later `get_task` reads.

## Architecture

- Adds `OpenMeshStreamRequest` and `OpenMeshStreamResponse` to the plugin protocol.
- Adds `PluginContext::open_mesh_stream` / `PluginContext::connect_mesh_stream` for plugin-originated mesh streams.
- Adds `PluginMeshEvent::OpenStream` to hand plugin stream requests from the plugin runtime into the mesh node.
- Adds a dedicated plugin mesh stream type in the mesh dispatcher.
- Bridges the remote mesh stream to the target plugin through the existing plugin side-stream mechanism.
- Keeps agent-specific A2A/ACP code out of mesh-llm; that lives in `Mesh-LLM/agents`.

## Example Flow

A concrete A2A use case looks like this:

1. Node B is running the external `agents` plugin and advertises a `pr-review` agent.
2. Node A discovers that remote agent through plugin mesh data.
3. A user or harness calls Node A's local `agents.send_message` tool.
4. The Node A `agents` plugin opens a mesh stream to Node B's `agents` plugin:

```rust
let stream = context
    .connect_mesh_stream(OpenMeshStreamRequest {
        stream_id: task_id.clone(),
        target_peer_id: owner_peer_id.to_string(),
        plugin_id: "agents".into(),
        channel: "a2a".into(),
        purpose: OpenStreamPurpose::Bidirectional as i32,
        metadata_json: serde_json::json!({
            "agent_id": "pr-review",
            "task_id": task_id,
        })
        .to_string(),
        ..Default::default()
    })
    .await?;
```

5. mesh-llm opens a QUIC stream to Node B, forwards the request to the local `agents` plugin side stream, and bridges bytes in both directions.
6. The `agents` plugin on Node B handles the A2A task, talks to its local ACP/runtime harness, and streams task updates/results back over the same mesh stream.
7. Node A persists the remote task result so later `get_task` reads work locally.

From the caller's perspective this still looks like a local MCP/A2A tool call. The mesh stream is the transport that lets the owning peer execute the task.

## Stack

This PR targets `codex/plugin-manager-docs` from #701 because it depends on the native plugin manager/runtime work.

The companion plugin implementation is in `Mesh-LLM/agents`:

- `8858e4a` Route remote A2A tasks over mesh streams
- `3a26c2f` Refresh agent adverts on peer updates

## Protocol

The mesh and plugin protocol changes are additive:

- new plugin protocol request/response messages are optional for older plugins,
- the new mesh stream type is only initiated by newer hosts/plugins,
- existing gossip, routing, and plugin channels are unchanged.

Older nodes will not understand or initiate plugin mesh streams, so remote A2A stream routing requires compatible peers on both ends. Existing non-agent plugin behavior remains unchanged.

## Validation

- `just build`
- `cargo test -p mesh-llm-plugin --lib`
- `cargo build -p mesh-agents --bin agents --locked`
- `cargo test -p mesh-agents --locked`
- `cargo clippy -p mesh-agents --all-targets --locked -- -D warnings`

End-to-end validation:

- started two local mesh-llm client nodes with isolated homes/configs,
- loaded the `agents` plugin on both nodes,
- advertised a valid `pr-review` A2A Agent Card on node B,
- verified node A discovered the remote agent,
- called node A's `send_message` plugin operation,
- verified the response used `transport: "mesh_stream"`,
- streamed a completed A2A task back from node B through a fake ACP stdio harness,
- verified node A `get_task` returned the persisted remote task result.


i386:
This PR is still valid

michaelneale (approved):
blessed as log as this is updated and build all green.

michaelneale:
@i386 I am not that worried about the stream/open angle I think it seems ok, really up to each plugin to do their own due diligence.

PR #706, opened 2026-05-27
Make model downloads package aware

i386:
## Summary

Make `mesh-llm models download` prefer a Skippy layer package when the requested model ref has a package mapping, while preserving the existing GGUF/model-file download fallback when no package exists.

## Details

- Resolves explicit layer package refs directly.
- Loads the remote catalog before checking package mappings for ordinary model refs.
- Downloads the complete package by resolving the package identity and fetching layer range `0..layer_count` with embeddings and output artifacts.
- Reports layer-package downloads distinctly in console and JSON output.

## Validation

- `just build`
- `cargo test -p mesh-llm-host-runtime model_download --lib`
- `cargo check -p mesh-llm`


michaelneale (approved):
Works - tested with 705

PR #708, opened 2026-05-27
Add mesh MCP HTTP endpoint + auto configure agents

i386:
## Summary

Adds a canonical streamable HTTP MCP endpoint to the mesh management API at `/mcp`.

Codex, Goose, OpenCode, and other MCP-capable harnesses should connect to the running mesh node at `http://127.0.0.1:3131/mcp` instead of launching a separate plugin runtime. The endpoint is backed by the existing plugin MCP aggregator, so plugin tools remain host-owned and are exposed through one management API surface.

## CLI / Agent Setup

- `mesh-llm opencode` now injects a remote MCP config pointed at the management API MCP URL.
- `mesh-llm opencode --write` persists the same MCP config alongside the mesh provider.
- `mesh-llm claude` passes an HTTP MCP config for the mesh endpoint.
- `mesh-llm goose` writes a Goose `streamable_http` MCP extension for the mesh endpoint.

The stdio MCP command path is intentionally not part of this PR.

## Validation

- `cargo check -p mesh-llm`

Attempted focused MCP test in a fresh worktree:

- `cargo test -p mesh-llm-host-runtime management_mcp_endpoint_initializes_streamable_http_session --lib`

That test build stopped before running because the fresh worktree had not prepared the repo's native llama static libraries (`llama-common`). The same focused MCP and launcher tests passed before the PR was restacked onto `main`.


michaelneale (approved):
seems low risk and a nice improvement, much better than the stdio. @i386 can we delete the old stdio code if not already?

i386:
@michaelneale deleted

PR #713, opened 2026-05-27
Fix MoA reducer context for chained tool results

michaelneale:
🤖 This PR fixes `model: "mesh"` tool-result turns when an agent has run multiple tools in sequence.

## Summary
- Preserve the nearest preceding user message when packing MoA reducer context for tool-result turns.
- Keep chained `assistant(tool_calls)` / `tool` pairs together so backend chat templates do not see a leading bare `tool` message.
- Add a regression test covering a weather-style chained web search then fetch flow.

## Validation
- `cargo test -p mesh-mixture-of-agents`
- `cargo check -p mesh-llm`

### DIFF
--- a/crates/mesh-mixture-of-agents/src/context.rs
+++ b/crates/mesh-mixture-of-agents/src/context.rs
@@ -245,19 +245,37 @@ pub fn pack_for_tool_result_turn(
 
     let mut messages = vec![json!({"role": "system", "content": system})];
 
-    // Forward the tail of the conversation that includes tool_call + tool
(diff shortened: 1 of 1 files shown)


ndizazzo:
@michaelneale This is still acting up? What are you seeing?

i386 (approved):

michaelneale:
@ndizazzo was pretty simple - multiple tool calls from my open claw box, using for real, easy fix!

ndizazzo (approved):

PR #719, opened 2026-05-28
fix(context-config): always pass a context length value for runtime models

ndizazzo:
## Summary

`mesh-llm opencode --write` and `pi` agent config files now use the **actual running context length** from the local mesh process, instead of only falling back to GGUF metadata headers.

### User-visible change

Before this fix, the `context_length` in generated OpenCode/pi configs came exclusively from `/api/models`, which reads GGUF metadata headers. That value is often the model's *native* context length (e.g. 32K for Qwen3-8B), not what the user actually configured via `--ctx-size` in CLI flags or `[models.ctx_size]` in `config.toml`. If a user overrode the context length to 16K (or the auto-planner picked a smaller value from VRAM), the config file still showed the GGUF's native value — silently wrong.

After this fix, agent configs use the **runtime process** `context_length` as the primary source — this is the value the model is actually running with (from `--ctx-size`, `config.toml`, or auto-computed from VRAM by `plan_runtime_resources`). GGUF/peer metadata is used only as a fallback for models not running locally (remote/peer models).

### Implementation

**`crates/mesh-llm-host-runtime/src/cli/commands/agent_cli.rs`**

- `fetch_model_context_lengths` now queries **both** `/api/models` and `/api/runtime/processes`, then merges via the new `merge_context_lengths()` pure function.
- `merge_context_lengths()` priority: runtime processes (actual running value) → GGUF/peer metadata (fallback via `or_insert`).
- Added `fetch_json()` helper to eliminate fetch/parse duplication.

### Merge semantics

| Model in runtime processes | Model in /api/models | Result |
|---|---|---|
| Has `context_length: N` | Any | `N` (runtime wins) |
| Has `context_length: null` | Has `context_length: M` | `M` (falls back to metadata) |
| Not present | Present | metadata value |
| Present | Not present | runtime value |

### Testing

5 unit tests on `merge_context_lengths`:

| Test | Scenario |
|---|---|
| `merge_context_lengths_uses_runtime_process_when_api_models_missing` | Mixed: runtime trumps metadata, runtime-null falls back, runtime-only appears |
| `merge_context_lengths_api_models_only` | No runtime processes → pure metadata |
| `merge_context_lengths_runtime_process_only` | No metadata → pure runtime |
| `merge_context_lengths_runtime_process_trumps_api_models` | Explicit override: 16384 (runtime) > 32768 (metadata) |
| `merge_context_lengths_falls_back_to_metadata_when_runtime_null` | Runtime has null → metadata survives |

/api/runtime/processes — ✅ has context_length:

```json
{
  "name": "unsloth/Qwen3.6-27B-GGUF:UD-Q4_K_XL",
  "context_length": 32768,
  "slots": 1,
  "status": "ready"
}
```

### DIFF
--- a/crates/mesh-llm-host-runtime/src/cli/commands/agent_cli.rs
+++ b/crates/mesh-llm-host-runtime/src/cli/commands/agent_cli.rs
@@ -919,16 +919,51 @@ async fn fetch_model_context_lengths(
     client: &reqwest::Client,
     management_models_url: &str,
 ) -> std::collections::HashMap<String, Option<u32>> {
+    let models_json = fetch_json(client, management_models_url).await;
+
+    // Query /api/runtime/processes for the actual running context_lengths.
+    let processes_url = management_models_url.replace("/api/models", "/api/runtime/processes");
+    let processes_json = fetch_json(client, &processes_url).await;
+
+    merge_context_lengths(&models_json, &processes_json)
+}
+
+async fn fetch_json(client: &reqwest::Client, url: &str) -> serde_json::Value {
+    match client.get(url).send().await {
+        Ok(resp) => resp.json::<serde_json::Value>().await.unwrap_or_default(),
+        Err(_) => serde_json::Value::Null,
+    }
+}
+
+fn merge_context_lengths(
+    models_json: &serde_json::Value,
+    processes_json: &serde_json::Value,
+) -> std::collections::HashMap<String, Option<u32>> {
     let mut context_map = std::collections::HashMap::new();
 
-    if let Ok(resp) = client.get(management_models_url).send().await
-        && let Ok(body) = resp.json::<serde_json::Value>().await
-    {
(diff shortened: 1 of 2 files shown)


michaelneale:
this is A ok with me - as long as get the checks passing

michaelneale:
:shipit:

i386:
One change request from review:

The new context-length lookup is only wired into `opencode --write`. The default `mesh-llm opencode` path still does:

```rust
let spec = build_opencode_launch_spec(&models, &chosen, &target.api_base_url);
```

so the `OPENCODE_CONFIG_CONTENT` used for a direct OpenCode launch still has no runtime-derived `limit.context` / `limit.output` values. That means the common non-`--write` flow keeps the old behavior even though it is also generating an OpenCode config on the fly.

Can you wire the non-write path through `fetch_model_context_lengths(...)` + `build_opencode_launch_spec_with_limits(...)` too? That should make `mesh-llm opencode` and `mesh-llm opencode --write` behave consistently.

PR #722, opened 2026-05-28
OpenAI API errors now return clean error messages instead of raw skippy envelopes

### LINKED ISSUE #715: OpenAI API format not returning error messages in the correct structure
## Problem

Simple prompt in OpenCode failing in v0.68.0 on certain models that fail to load successfully, should return the actual error from the backend instead of an unparsable message.

## Steps to reproduce

1. Open `mesh-llm opencode`
2. Select the `mesh-llm` model `unsloth/gemma-4-26B-A4B-it-GGUF:UD-Q4_K_XL` or `unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL`
3. Submit any prompt

## Diagnostics

```
skippy ABI call failed: Unsupported: [json.exception. parse_error.101] parse error at line 1,
column 580: syntax error while parsing value - invalid literal; last read: • "maximum": 0•|
```

## Extra context

Image

Image

ndizazzo:
## Summary

When a model fails to load (e.g. malformed GGUF metadata), the OpenAI-compatible API previously returned errors like:

```
skippy ABI call failed: Unsupported: parse error at line 1, column 580: ...
```

The `skippy ABI call failed:` envelope is now stripped at the source, so the same error surfaces cleanly as:

```
Unsupported: parse error at line 1, column 580: ...
```

Additionally, malformed GGUF `chat_template` or other model metadata JSON no longer causes a fatal load failure. The stage runtime catches the parse error and clears chat sampling gracefully instead of aborting — the model loads and serves normally, just without chat-grammar constraints.

### Fix 1: error envelope (`crates/skippy-runtime/src/lib.rs`)

`ensure_ok()` in the skippy-runtime C ABI wrapper was prepending `"skippy ABI call failed: "` to every C-side error before propagating it through the Rust anyhow chain. Removed that prefix — callers that want context add their own `.with_context()` or `.context()`. The error now reads `{Status:?}: {message}` directly.

No programmatic dependencies on the old string existed — it was purely cosmetic.

### Fix 2: metadata parse crash (`third_party/llama.cpp/patches/0037-*.patch`)

`skippy_session_configure_chat_sampling()` parses model metadata JSON (chat template, special tokens, etc.) with nlohmann/json. When a GGUF file has malformed JSON in its metadata, the `json::parse` throws. The catch blocks were returning `SKIPPY_STATUS_UNSUPPORTED` with the parse error message, which made the model unfriendly. Changed both catch blocks to return `skippy_success(out_error)` after already clearing chat sampling — the model loads, chat sampling is simply absent.

### Fix 3: operation context in error messages (`third_party/llama.cpp/patches/0036-*.patch`)

Two additional catch blocks in `skippy_apply_chat_template_json` and `skippy_parse_chat_response_json` were returning raw `e.what()` from nlohmann parse errors without any context about what operation was failing. Added a prefix: `"failed to apply chat template: "` and `"failed to parse chat response: "` respectively, so users see what was being attempted when an error occurs.

## Protocol / Compatibility

No protocol impact. All fixes are local to the serving node's error handling — no wire format, gossip field, or API contract changes.

## Files changed

| File | Change |
|---|---|
| `crates/skippy-runtime/src/lib.rs` | Strip `"skippy ABI call failed: "` prefix from `ensure_ok()`; extract `format_skippy_error()` helper |
| `third_party/llama.cpp/patches/0036-Apply-chat-template-JSON.patch` | Add operation context prefixes to catch blocks |
| `third_party/llama.cpp/patches/0037-Apply-chat-grammar-during-stage-sampling.patch` | Catch blocks return success instead of fatal error on metadata parse failure |
| `.deps/llama.cpp/src/skippy.cpp` | Same fixes applied in compiled build output (verification copy) |
| `crates/mesh-llm-host-runtime/src/runtime/context_planning.rs` | Doc references match new error format |

### Local
- `cargo check -p skippy-runtime` ✓
- `cargo check -p mesh-llm` ✓
- `cargo test -p skippy-runtime --lib` — 45/45 passed (was 41) ✓
- LSP diagnostics clean on all changed files ✓
- Zero remaining references to `"skippy ABI call failed"` in Rust codebase ✓

### Live (carrack.patio51.com:9337)
Deployed after `just build` + bundle + scp. Server restarted and verified clean.

| Test | Model | Result |
|---|---|---|
| Long prompt + system message (200 tok) | `unsloth/Qwen3.5-4B-MTP-GGUF:UD-Q4_K_XL` | ✅ |
| Multi-turn conversation (context retention) | same MTP | ✅ |
| Streaming (SSE chunks) | same MTP | ✅ |
| Custom parameters (temp 0.3, top_p 0.9, 300 tok) | `unsloth/Qwen3.6-27B-GGUF:UD-Q4_K_XL` | ✅ |
| Simple query (30 tok) | `unsloth/Qwen3.5-4B-GGUF:UD-Q4_K_XL` | ✅ |
| Non-existent model 404 | — | ✅ proper OpenAI error |
| Unsupported feature (json_object) | MTP | ✅ proper OpenAI error |
| `journalctl -u mesh-llm` | error/warn/fail log scan | ✅ clean |

## Tests added

| Test | What it verifies |
|---|---|
| `format_skippy_error_omits_abi_envelope` | Error format does not contain `"skippy ABI call failed"` prefix |
| `format_skippy_error_works_without_message` | Error format works when no C-side message is present |
| `format_skippy_error_covers_all_status_variants` | All 7 `Status` variants produce clean error without ABI envelope |
| `configure_chat_sampling_survives_bad_metadata_json` | Integration test — GGUF with malformed metadata JSON returns Ok instead of fatal error (requires `SKIPPY_CORRECTNESS_MODEL` env var, skipped otherwise) |

Fixes #715

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
@@ -4268,8 +4268,7 @@ pub async fn send_json_with_status_and_headers(
 }
 
 pub async fn send_400(mut stream: TcpStream, msg: &str) -> std::io::Result<()> {
-    let body = serde_json::to_vec(&serde_json::json!({ "error": msg }))
-        .expect("serializing JSON error response should not fail");
+    let body = openai_error_body(400, msg);
     let headers = format!(
         "HTTP/1.1 400 Bad Request\r\nContent-Type: application/json\r\nContent-Length: {}\r\n\r\n",
         body.len()
@@ -4282,13 +4281,20 @@ pub async fn send_400(mut stream: TcpStream, msg: &str) -> std::io::Result<()> {
 
 pub async fn send_error(mut stream: TcpStream, code: u16, msg: &str) -> std::io::Result<()> {
     let status = match code {
+        401 => "Unauthorized",
+        403 => "Forbidden",
         404 => "Not Found",
         409 => "Conflict",
+        413 => "Payload Too Large",
         422 => "Unprocessable Content",
         429 => "Too Many Requests",
+        500 => "Internal Server Error",
+        502 => "Bad Gateway",
+        503 => "Service Unavailable",
+        504 => "Gateway Timeout",
         _ => "Bad Request",
     };
-    let body = serde_json::json!({"error": msg}).to_string();
+    let body = openai_error_body(code, msg);
     let retry_after = if code == 429 {
         "Retry-After: 5\r\n"
     } else {
@@ -4297,7 +4303,7 @@ pub async fn send_error(mut stream: TcpStream, code: u16, msg: &str) -> std::io:
     let resp = format!(
         "HTTP/1.1 {code} {status}\r\nContent-Type: application/json\r\n{retry_after}Content-Length: {}\r\n\r\n{}",
         body.len(),
-        body
+        String::from_utf8_lossy(&body)
     );
     stream.write_all(resp.as_bytes()).await?;
     stream.shutdown().await?;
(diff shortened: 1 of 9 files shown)


michaelneale:
Ha nice. I was amusingly getting familiar with that error.

ndizazzo:
@michaelneale Now you don't have to settle!

The only thing I'm not sure about is if we want this:

> the model loads and serves normally, just without chat-grammar constraints.

Perhaps _yes_, **with** some kind of skippy warning would be appropriate for any embedded template that causes parse issues?

i386 (approved):

PR #723, opened 2026-05-28
Add split readiness diagnostics and Windows capture bundle

IvGolovach:
## Summary

This PR adds a repeatable split-readiness diagnostic path for operators who are trying to bring up Skippy split serving across multiple nodes.

Users can now:

- Ask a running node why a model is or is not split-ready through `GET /api/diagnostics/split-readiness?model_ref=...`.
- Run `mesh-llm doctor split --model-ref <model>` for terminal-readable guidance and optional JSON capture.
- Collect a redacted Windows diagnostic bundle from already-running nodes with `contrib/windows/CollectSplitDiagnostics.ps1`.
- Follow the documented split doctor flow from `docs/SKIPPY_SPLITS.md` and the Windows helper README.

## Why

Recent split setup debugging showed that users can hit symptoms like "only one valid node", repeated manifest transfer, missing stage source, or unclear GPU/backend eligibility without a compact way to prove which node was excluded and why.

This keeps the fix operational and low-risk: it does not change split placement, routing, gossip, protocol payloads, model loading, or artifact transfer behavior. It only exposes the already-known readiness signals in a focused diagnostic surface and adds a shareable capture path for Windows repros.

## Diff scope

- Adds a read-only split-readiness report that classifies local and peer nodes as eligible participants or exclusions.
- Computes readiness verdicts and capacity evidence from the same participant set shown in the report, so excluded high-capacity peers cannot make the diagnostic report `ready`.
- Reports peers with missing stage/package source proof as excluded with `missing_model_source` instead of treating unknown source state as ready.
- Adds exclusion reasons for client role, missing VRAM, missing model interest/source signal, and missing stage protocol generation support.
- Adds recommendations for common operator mistakes, including starting another worker/host with the same model and using distinct ports when testing multiple local processes.
- Adds `mesh-llm doctor split` with human-readable output, `--json`, and `--output-dir` support.
- Adds a Windows collector that captures redacted status/runtime/models/split-readiness payloads, GPU/process facts, optional probe output, and recent Skippy native log tails without starting or stopping mesh processes.
- Updates split and Windows docs to point users at the new flow.

## Compatibility and safety

No mesh wire protocol change.
No plugin protocol change.
No Skippy ABI change.
No routing or model-load behavior change.
The management endpoint is read-only.
The Windows collector redacts token/auth/join-like values before writing captured HTTP payloads.

## Validation

Validation tier: Tier 4 - split-readiness verification tooling, read-only management diagnostics, Windows capture helper, and post-review diagnostic correctness repair; no routing, wire protocol, or model-load behavior change.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, origin/main at `6f6d606c6191fafce68973eff56fe7aa3e035533`.
- `cargo fmt --all -- --check`: PASS, no output.
- `cargo test -p mesh-llm-host-runtime split_readiness --lib -- --test-threads=1`: PASS, 9 passed.
- `cargo check -p mesh-llm`: PASS.
- `cargo-clippy clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS.

Version: not applicable - diagnostics/tooling/docs only; no release/version sync required.

Not run:

- PowerShell syntax/help validation - `pwsh` is not installed in this environment.
- Live two-node split smoke - no local split-capable multi-node runtime/model endpoint was available; deterministic split-readiness tests cover the diagnostic classification and capture path.

## Rollback

Revert this PR.

DB downgrade: not applicable.
Data repair: not applicable.

## Known residual risks

The Windows collector should still be exercised on an actual Windows machine with `pwsh` and a running mesh setup before treating it as fully field-proven. The Rust-side diagnostic classification and CLI output are covered locally.

### DIFF
--- a/contrib/windows/CollectSplitDiagnostics.ps1
+++ b/contrib/windows/CollectSplitDiagnostics.ps1
@@ -1,0 +1,252 @@
+param(
+    [string[]]$ConsoleUrls = @("http://127.0.0.1:3131"),
+    [string[]]$ApiUrls = @("http://127.0.0.1:9337/v1"),
+    [string]$Model = "auto",
+    [string]$OutputDir = $env:TEMP,
+    [string]$MeshLlm = "",
+    [switch]$RunProbe,
+    [switch]$SkipHttp,
+    [switch]$Help
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = "Stop"
+
+function Show-Usage {
+    @"
+CollectSplitDiagnostics.ps1 - capture split-readiness diagnostics for maintainers.
+
+Usage:
+  .\contrib\windows\CollectSplitDiagnostics.ps1 -Model meshllm/Qwen3-8B-Q4_K_M-layers
+
+Options:
+  -ConsoleUrls  Management API URLs, default http://127.0.0.1:3131
+  -ApiUrls      OpenAI API base URLs, default http://127.0.0.1:9337/v1
+  -Model        Model id/ref for split doctor and optional chat probe
+  -OutputDir    Parent output directory, default TEMP
+  -MeshLlm      mesh-llm.exe path, default target\release\mesh-llm.exe or PATH
+  -RunProbe     Run a tiny /chat/completions probe through each API URL
+  -SkipHttp     Capture local system/process facts only
+  -Help         Print this help
+"@
+}
+
+if ($Help) {
+    Show-Usage
+    exit 0
+}
+
+function Resolve-MeshBinary {
+    param([string]$Preferred)
+
+    if ($Preferred -and (Test-Path -LiteralPath $Preferred)) {
+        return (Resolve-Path -LiteralPath $Preferred).Path
+    }
+
+    $ReleaseBinary = Join-Path (Get-Location) "target\release\mesh-llm.exe"
+    if (Test-Path -LiteralPath $ReleaseBinary) {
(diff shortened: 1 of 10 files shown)


i386 (changes requested):
Thanks for putting this together. The shape is useful, but I think this needs a couple fixes before it can be trusted as a split-readiness diagnostic. Right now the report can say `ready` in cases where the actual split planner would still reject the peer set, which is exactly the kind of operator confusion this command is meant to remove.

Two spots I think need tightening:

- `crates/mesh-llm-host-runtime/src/api/split_readiness.rs:103`: this reuses `model_target_lookup` capacity advice, but that advice counts every non-client peer with VRAM. That can include peers the new split-readiness report later excludes for missing model interest or stale stage protocol support. So the diagnostic can show a smaller participant set while borrowing capacity from an excluded high-VRAM peer and still return `ready`. I think the verdict needs to be computed from the same participant set this report returns, or avoid claiming `ready` from that reused advice.

- `crates/mesh-llm-host-runtime/src/api/split_readiness.rs:209`: this preflight does not model the runtime planner's `missing_model_source` exclusion. `collect_split_participants` probes inventory/source availability and excludes peers that cannot stage the package, but this diagnostic currently keeps an interested peer as a participant with `model_source_state: "unknown"` and can still report `ready`. Since missing stage source is one of the failure modes this PR is meant to explain, the diagnostic should either perform the same source/inventory check or downgrade/report those peers as excluded instead of ready participants.

PR #725, opened 2026-05-28
Let plugins provide agent skills

i386:
## Summary

Plugins can now ship Agent Skills that mesh-llm installs into supported agent clients. A plugin exposes skills by bundling skill folders in its native plugin archive, for example:

```text
blackboard/
  plugin.toml
  blackboard
  skills/
    blackboard/
      SKILL.md
```

The mesh plugin installer extracts the archive as before. The new skill installer discovers enabled installed plugins, finds `skills/<name>/SKILL.md` under their install paths, and copies those skills into agent-specific skill directories. It preserves user-owned skill directories unless the explicit installer is run with `--force`.

## CLI and Agent Behavior

Adds:

```bash
mesh-llm skills install
mesh-llm skills install --agent goose --agent codex
mesh-llm skills install --all --dry-run
```

By default, `skills install` writes only to detected agent locations. It can also target explicit agents or all supported locations.

Agent launchers now install available plugin skills before launching the target agent:

- `mesh-llm goose` installs Goose skills
- `mesh-llm pi` installs Pi skills
- `mesh-llm opencode` installs OpenCode skills
- `mesh-llm claude` installs Claude Code skills

Launch-time install failures warn and continue so the agent session is not blocked by a local skill filesystem issue.

## Plugin Author Story

Plugin authors define skills by adding normal Agent Skill directories to their release archive:

```text
skills/<skill-name>/SKILL.md
```

Skill names are validated for portable lowercase ASCII/hyphen naming. Skills should use relative references and avoid OS-specific absolute paths unless the skill explicitly documents platform requirements.

Example plugin PR: https://github.com/Mesh-LLM/blackboard/pull/1 updates Blackboard to bundle its existing `skills/blackboard/SKILL.md` in release archives.

## Validation

- `cargo check -p mesh-llm-skills`
- `cargo check -p mesh-llm-plugin-manager`
- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-skills --lib`
- `cargo test -p mesh-llm-plugin-manager --lib`

Note: `cargo test -p mesh-llm-host-runtime --lib` was attempted but did not run because this local checkout is missing the native `llama-common` static library required by `skippy-ffi`.


crates/mesh-llm-plugin-manager/Cargo.toml
```diff
@@ -16,6 +16,7 @@ anyhow.workspace = true
 dirs = "6"
 flate2 = "1"
 futures-util = "0.3"
+mesh-llm-skills = { path = "../mesh-llm-skills", version = "0.68.0" }
```
michaelneale: version mismatch?

crates/mesh-llm-skills/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+}
+
+impl SkillAgent {
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Goose => "goose",
+            Self::Pi => "pi",
+            Self::Codex => "codex",
+            Self::Opencode => "opencode",
+            Self::Claude => "claude",
```
ndizazzo, line 31: @i386 I know this doesn't really fit as a dedicated "agent", but there is [an open format](https://agentskills.io/client-implementation/adding-skills-support) I think we should also support along with this... if you install them to `~/.agents/skills`, they're available to all harnesses that support that.

It might be worth shoehorning into this PR to support "general" or "global" or something like that.

crates/mesh-llm-plugin-manager/Cargo.toml
ndizazzo: We can use the workspace inherited version instead of hardcoding something specific

crates/mesh-llm-host-runtime/src/cli/commands/skills.rs
```diff
@@ -0,0 +1,180 @@
...
+        eprintln!(
+            "   {label:<17} {:<28} -> {:<8} {}",
+            skill_display_name(action),
+            action.agent.as_str(),
+            action.destination_dir.display()
+        );
+    }
+
+    print_install_summary(report, dry_run);
+}
```
ndizazzo, line 135: Lots of `--log-format json` incompatible println's here. Unifying to one log utility is probably out of scope for this PR, but we need to make sure we don't break other output formats still

ndizazzo (approved):
Really interesting idea - specifically, for Blackboard it makes sense to teach agents how to use it intrinsically. 

Since this is a generally an improvement that depends on individual plugins adding skills, I think it's safe and doesn't violate the boundaries of what mesh-llm provides... However, I do share the same sentiment that @michaelneale mentioned a few days ago that we want to be deliberate about where/what we provide and at what layer of the stack.

Some minor inline comments/stuff, but generally nothing blocking ✅

i386:
@ndizazzo @michaelneale thanks for the reviews 👍

i386:
@ndizazzo @michaelneale updated with suggested changes

PR #731, opened 2026-05-28
Keep plugin runtime responsive during long operations

i386:
## Summary

Makes plugin runtimes accept and execute concurrent host requests instead of serializing the entire plugin process behind one long-running handler.

This fixes the class of failure we hit while testing A2A agents through the mesh MCP endpoint: `agents.send_message` started a long ACP-backed PR review task, then the plugin process stopped responding to host traffic because the runtime was blocked inside that single operation. The host supervisor interpreted that as a dead plugin, restarted it, and the original MCP call failed with:

```text
MCP error -32603: Plugin 'agents' disconnected
```

## Problem

The old runtime processed envelopes serially:

1. read one envelope from the host
2. await the handler to completion
3. write the response
4. read the next envelope

That meant a slow `tools/call`, service invocation, stream open, or plugin event handler blocked every later envelope, including health checks. Async handlers could yield to Tokio, but the runtime read loop still awaited the active handler before accepting the next host message.

The first version of this PR made the read/write loops concurrent, but still held a plugin mutex while each handler executed. That kept the process responsive to envelope reads, but still imposed a runtime-level plugin concurrency limit. That is not what agent plugins need: A2A, ACP-backed tasks, long MCP tools, and remote mesh streams must be able to run concurrently unless the plugin itself chooses to limit them.

## Changes

- Splits plugin connection IO into an independent read loop and serialized write loop.
- Routes all plugin writes through an outbound channel instead of giving handlers direct `&mut LocalStream` access.
- Dispatches host requests into independent tasks so one long operation does not block later host messages.
- Stores the initialized plugin behind an `RwLock`, then clones the plugin handle for each inbound request and drops the lock before invoking the handler.
- Removes the runtime-level handler mutex and the synthetic `busy` health response.
- Routes plugin-originated host responses, such as `open_mesh_stream`, through a pending response map instead of directly reading from the shared stream.
- Adds a regression test that sends two service invocations which must rendezvous with each other; the test times out if handlers are still serialized.
- Keeps the health regression test that verifies health can respond while a slow operation is still running.

## Plugin Concurrency Contract

`PluginRuntime::run` and `PluginRuntime::run_with_stream` now require plugin handles to be `Clone + Sync`.

The runtime does not impose a request concurrency limit. Plugin implementations that need shared state should put that state behind their own `Arc`, `Mutex`, `RwLock`, database pool, queue, or semaphore. Plugin implementations that need a concurrency limit should enforce that limit inside the plugin where the policy belongs.

The built-in `SimplePlugin`, routers, and `InternalRpcPlugin` are cloneable, so macro-built plugins and the in-tree plugins continue to work.

## Compatibility

This changes the source-level bound for custom plugin types passed directly to `PluginRuntime::run`: they now need to be cloneable and sync-safe. The wire protocol is unchanged.

Existing plugins built with the public builders/macros should not need handler rewrites. Custom plugins with internal mutable fields should move those fields into shared interior state, for example `Arc<Mutex<State>>`, if they need mutations to be visible across concurrent request clones.

## Validation

- `cargo test -p mesh-llm-plugin --lib`
- `cargo check -p mesh-llm`

### DIFF
--- a/crates/mesh-llm-plugin/src/context.rs
+++ b/crates/mesh-llm-plugin/src/context.rs
@@ -146,15 +146,11 @@ impl<'a> PluginContext<'a> {
         let mut pending_guard =
             PendingHostResponseGuard::new(request_id, self.pending_host_responses.clone());
 
-        if let Err(err) = self
-            .send_payload(
-                proto::envelope::Payload::OpenMeshStreamRequest(request),
-                request_id,
-            )
-            .await
-        {
-            return Err(err);
-        }
+        self.send_payload(
+            proto::envelope::Payload::OpenMeshStreamRequest(request),
+            request_id,
+        )
+        .await?;
 
         let response = rx.await??;
         pending_guard.disarm();
--- a/crates/mesh-llm-plugin/src/context.rs
+++ b/crates/mesh-llm-plugin/src/context.rs
@@ -1,31 +1,52 @@
 use anyhow::{Result, bail};
 use serde::Serialize;
+use std::collections::HashMap;
+use std::marker::PhantomData;
+use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
+use tokio::sync::{Mutex, mpsc, oneshot};
 
 use crate::{
     PROTOCOL_VERSION,
     helpers::{channel_message, json_channel_message},
-    io::{
-        LocalStream, connect_side_stream, read_envelope, send_bulk_transfer_message,
-        send_channel_message, write_envelope,
-    },
+    io::{LocalStream, connect_side_stream},
     proto,
 };
 
 static NEXT_HOST_REQUEST_ID: AtomicU64 = AtomicU64::new(1);
+const PLUGIN_ORIGINATED_REQUEST_BIT: u64 = 1 << 63;
+
+pub(crate) type PendingHostResponses =
+    Arc<Mutex<HashMap<u64, oneshot::Sender<Result<proto::Envelope>>>>>;
 
 pub struct PluginContext<'a> {
-    pub(crate) stream: &'a mut LocalStream,
-    pub(crate) plugin_id: &'a str,
+    pub(crate) outbound_tx: mpsc::Sender<proto::Envelope>,
+    pub(crate) pending_host_responses: PendingHostResponses,
+    pub(crate) plugin_id: String,
+    pub(crate) _marker: PhantomData<&'a mut ()>,
 }
 
 impl<'a> PluginContext<'a> {
+    pub(crate) fn new(
+        plugin_id: String,
+        outbound_tx: mpsc::Sender<proto::Envelope>,
+        pending_host_responses: PendingHostResponses,
+    ) -> Self {
+        Self {
+            outbound_tx,
+            pending_host_responses,
+            plugin_id,
+            _marker: PhantomData,
+        }
+    }
+
     pub async fn send_channel(&mut self, message: proto::ChannelMessage) -> Result<()> {
         self.send_channel_message(message).await
     }
 
     pub async fn send_channel_message(&mut self, message: proto::ChannelMessage) -> Result<()> {
-        send_channel_message(self.stream, self.plugin_id, message).await
+        self.send_payload(proto::envelope::Payload::ChannelMessage(message), 0)
+            .await
     }
 
     pub async fn send_text_channel(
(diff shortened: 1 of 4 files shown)


crates/mesh-llm-plugin/src/context.rs
```diff
@@ -1,31 +1,52 @@
...
-    },
+    io::{LocalStream, connect_side_stream},
     proto,
 };
 
 static NEXT_HOST_REQUEST_ID: AtomicU64 = AtomicU64::new(1);
+const PLUGIN_ORIGINATED_REQUEST_BIT: u64 = 1 << 63;
+
+pub(crate) type PendingHostResponses =
+    Arc<Mutex<HashMap<u64, oneshot::Sender<Result<proto::Envelope>>>>>;
```
ndizazzo, line 20: If a handler drops its future without awaiting the response, the `oneshot::Sender` stays in this map indefinitely. We should track pending requests in the outbound channel so dropped futures get cleaned up when the sender half is dropped. 

Without something like this, long-running plugins that don't clean up will accumulate memory on every cancelled/aborted request. Another option might be a timeout-based cleanup, but I think that's a lot weaker, we'll never know what timeout is generally appropriate.

crates/mesh-llm-plugin/src/runtime.rs
```diff
@@ -1334,105 +1341,193 @@ impl Plugin for SimplePlugin {
...
+        let plugin_id = plugin.plugin_id().to_string();
+        let (read, write) = stream.into_split();
+        let (outbound_tx, outbound_rx) = mpsc::channel(256);
+        let writer = tokio::spawn(Self::write_loop(write, outbound_rx));
+        let state = Arc::new(RuntimeState {
+            plugin: Arc::new(RwLock::new(plugin)),
+            plugin_id,
+            outbound_tx,
+            pending_host_responses: Arc::new(Mutex::new(HashMap::new())),
+        });
```
ndizazzo, line 1379: This spawns independent read & write tasks, but if `write_loop` ever fails, it'll terminate silently. Then `read_loop` keeps running, but all pending requests will hang forever because responses can never arrive. 

It also doesn't have any shared error state (watch/AtomicBool) to poison any pending responses.

crates/mesh-llm-plugin/src/runtime.rs
```diff
@@ -1334,105 +1341,193 @@ impl Plugin for SimplePlugin {
...
-                    stream,
-                    &plugin_id,
+                    &state,
                     request_id,
                     proto::envelope::Payload::ShutdownResponse(proto::ShutdownResponse {}),
                 )
                 .await?;
                 Ok(false)
             }
+            payload => Self::spawn_payload_handler(state, request_id, payload),
```
ndizazzo, line 1558: Nothing drains pending work when this exits... `Ok(false)` → `read_loop` exits → `run_with_stream` calls `writer.await`. But spawned handler tasks hold `Arc<RuntimeState> `, so the writer keeps running until all active handlers finish and drop their Arcs. 

We need some kind of poison/cancel mechanism for pending responses or queued messages.

crates/mesh-llm-plugin/src/runtime.rs
```diff
@@ -1334,105 +1341,193 @@ impl Plugin for SimplePlugin {
...
+    fn spawn_payload_handler<P: Plugin + Clone + Sync + 'static>(
+        state: Arc<RuntimeState<P>>,
+        request_id: u64,
+        payload: proto::envelope::Payload,
+    ) -> Result<bool> {
+        tokio::spawn(async move {
+            let _ = Self::handle_payload(state, request_id, payload).await;
+        });
+        Ok(true)
+    }
```
ndizazzo, line 1600: Review agent caught one here

---

Previously, all handlers ran sequentially in the event loop. Now ChannelMessage, MeshEvent, BulkTransferMessage, notification handlers, etc. are all spawned. If MeshEvent A arrives before event B, but B's handler finishes first, the plugin processes B before A. For ordering-sensitive events, this creates subtle races.

i386:
@ndizazzo I think we have solved these issues now in b68488b15. The update cleans up dropped pending host responses, poisons/drains pending responses on runtime shutdown or writer failure, and keeps event/notification-style handlers ordered while leaving request handlers concurrent. The focused plugin tests, fmt check, and crate check are passing.

PR #734, opened 2026-05-29
Make auto routing fall back to healthy models

### LINKED ISSUE #625: auto-router: model-level fallback when top-pick model has no eligible peers

IvGolovach:
## Summary

`model=auto` now avoids choosing a model whose only available targets are cooling, unavailable, or too small for the request context when another suitable model can serve the prompt. Explicit requests for a named model keep the existing availability-preserving behavior, but auto routing gets a stricter pre-selection path so agent/chat traffic can fall through to a healthier model instead of repeatedly spending turns on a known-bad top pick.

Fixes #625.

## Why

#625 describes the public-mesh failure mode directly: auto could classify a prompt, pick the preferred model, and then keep routing to that model even when its only peer had just failed. The existing target-health filter intentionally preserved availability for explicit model requests, but auto routing needed model-level admission before it committed to a model.

The review pass also caught an important stickiness edge: a cached `model=auto` session should not abandon its current model merely because that model is cooling if every alternative is cooling too. In that case switching models does not buy a healthier route and can break agent/tool-loop KV locality.

## Diff scope

- Adds a strict target-health path for auto routing while keeping the existing lenient path for explicit model routes.
- Adds an auto-route admission module that checks both request context budget and target health before a model participates in auto selection.
- Applies the same ready-model pool to both passive mesh OpenAI routing and local API ingress routing.
- Reclassifies cached auto-model sessions only when the cached model is missing, cannot satisfy media requirements, or a non-empty ready-model pool excludes it.
- Preserves cached auto-model stickiness when no ready alternate model exists.
- Adds regression coverage for single-candidate cooldown, all-cooling fallback behavior, ready-model pool selection, cached all-cooling stickiness, and switching to a ready alternate.

## Behavior notes

- If at least one auto candidate has a healthy/context-compatible target, auto selects from that ready pool.
- If every candidate is cooling or otherwise ineligible, auto preserves availability by falling back to the original candidate pool instead of manufacturing an immediate empty-route failure.
- Cached auto sessions keep their model when the ready pool is empty; they switch only when another compatible model has a proven eligible target.
- Missing cached models and media-incompatible cached models are still hard reclassification reasons.
- `ContextOverflow`, 4xx rejections, and client disconnects still do not poison target health; only retryable timeout/unavailable outcomes cool targets.

## Compatibility

No mesh protocol, Skippy ABI, package format, release metadata, or CLI contract changes. This is process-local routing behavior only; no new gossip fields or mixed-version incompatibility.

## Validation

Validation tier: Tier 3 - shared OpenAI routing behavior for agent/model auto selection, refreshed on current main and updated for the cached auto-model stickiness review fix.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, `origin/main` at `6d2f6b104f0ebc63f933fced1d840bacbdc15b34`
- `LLAMA_STAGE_BUILD_DIR=<llama-stage-build-dir> cargo test -p mesh-llm-host-runtime openai::transport --lib -- --test-threads=1`: PASS, 63 passed
- `LLAMA_STAGE_BUILD_DIR=<llama-stage-build-dir> cargo test -p mesh-llm-host-runtime auto_route --lib -- --test-threads=1`: PASS, 2 passed
- `LLAMA_STAGE_BUILD_DIR=<llama-stage-build-dir> cargo test -p mesh-llm-host-runtime target_health --lib -- --test-threads=1`: PASS, 11 passed
- `LLAMA_STAGE_BUILD_DIR=<llama-stage-build-dir> cargo test -p mesh-llm-host-runtime strict_eligible --lib -- --test-threads=1`: PASS, 1 passed
- `LLAMA_STAGE_BUILD_DIR=<llama-stage-build-dir> cargo check -p mesh-llm`: PASS
- `LLAMA_STAGE_BUILD_DIR=<llama-stage-build-dir> cargo-clippy clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS

Version: not applicable - no release/version sync required for this non-release routing behavior change.

Not run: live multi-node agent smoke. I did not have a local model/runtime mesh endpoint available; deterministic routing-health and auto-route unit coverage plus shipped-binary check/clippy cover the changed paths.

## Rollback

Revert this PR.

DB downgrade: not applicable.
Data repair: not applicable.

## Known residual risk

This should still get normal PR CI and one real public/private mesh smoke with `model=auto` after opening: force one preferred model target into cooldown/unavailable state, confirm auto falls through to another compatible model, then confirm the preferred model can participate again after health recovers.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/affinity.rs
+++ b/crates/mesh-llm-host-runtime/src/network/affinity.rs
@@ -130,6 +130,15 @@ impl AffinityRouter {
         self.target_health.eligible_candidates(model, candidates)
     }
 
+    pub(crate) fn route_strict_eligible_candidates(
+        &self,
+        model: &str,
+        candidates: &[election::InferenceTarget],
+    ) -> Vec<election::InferenceTarget> {
+        self.target_health
+            .strict_eligible_candidates(model, candidates)
+    }
+
     pub(crate) fn record_target_outcome(
         &self,
         model: Option<&str>,
@@ -888,6 +897,25 @@ mod tests {
         assert_eq!(prepared.ordered.len(), 1);
         assert_eq!(prepared.learn_prefix_hash, None);
         assert_eq!(prepared.cached_target, None);
+    }
+
+    #[test]
+    fn strict_eligible_candidates_drop_single_cooling_auto_target() {
+        let id = make_id(1);
+        let target = election::InferenceTarget::Remote(id);
+        let affinity = AffinityRouter::with_config(true, true);
+
+        affinity.record_target_outcome(Some("qwen"), &target, TargetHealthOutcome::Unavailable);
+
+        assert_eq!(
+            affinity.route_eligible_candidates("qwen", std::slice::from_ref(&target)),
+            vec![target.clone()]
+        );
+        assert!(
+            affinity
+                .route_strict_eligible_candidates("qwen", std::slice::from_ref(&target))
+                .is_empty()
+        );
     }
 
     #[test]
--- a/crates/mesh-llm-host-runtime/src/network/openai/auto_route.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/auto_route.rs
@@ -1,0 +1,149 @@
+//! Auto-route model admission helpers.
+//!
+//! Explicit model routing keeps an availability-preserving fallback when every
+//! target is cooling. Auto routing can be stricter before it chooses a model:
+//! if another model has a healthy target that can fit the request context, it
+//! should use that model instead of spending an agent turn on a target we just
+//! proved unhealthy or too small.
+
+use crate::inference::election;
+use crate::mesh;
+use crate::network::affinity::AffinityRouter;
+use crate::network::router;
+
+fn has_routable_candidate(candidates: &[election::InferenceTarget]) -> bool {
+    candidates
+        .iter()
+        .any(|target| !matches!(target, election::InferenceTarget::None))
+}
+
+async fn target_context_satisfies_request(
+    node: &mesh::Node,
+    model: &str,
+    required_tokens: Option<u32>,
+    target: &election::InferenceTarget,
+) -> bool {
+    let Some(required_tokens) = required_tokens else {
+        return !matches!(target, election::InferenceTarget::None);
+    };
+    let context_length = match target {
+        election::InferenceTarget::Local(_) => node.local_model_context_length(model).await,
+        election::InferenceTarget::Remote(peer_id) => {
+            node.peer_model_context_length(*peer_id, model).await
+        }
+        election::InferenceTarget::None => return false,
+    };
+    context_length
+        .map(|context| context >= required_tokens)
+        .unwrap_or(true)
+}
+
+async fn context_compatible_targets(
+    node: &mesh::Node,
+    model: &str,
+    required_tokens: Option<u32>,
+    candidates: &[election::InferenceTarget],
+) -> Vec<election::InferenceTarget> {
+    let mut compatible = Vec::new();
+    for candidate in candidates {
+        if target_context_satisfies_request(node, model, required_tokens, candidate).await {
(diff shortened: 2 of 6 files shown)


i386 (changes requested):
Thanks for the routing-health work here. I think this needs one behavior fix before merge.

[P2] Preserve cached auto-model stickiness when no alternate ready model exists

`cached_auto_model_reclassify_reason` now forgets the cached auto model whenever `model_has_eligible_remote_host(...)` is false. That check runs before the later ready-pool fallback in `auto_route::pool_for_ready_models`, which intentionally preserves availability by falling back to the full candidate set when no model has a ready target.

Concretely, if the cached model is cooling and every other candidate is also cooling, this path forgets the session cache anyway and lets reclassification choose another cooling model. That contradicts the stated behavior of switching only when another model has a healthy target, and it can break agent/tool-loop model stickiness without gaining any healthier route.

Please defer health-based cached-model invalidation until after confirming there is at least one alternate ready model for this request, or otherwise preserve the cached model when the ready set is empty.

Validation note: I attempted `cargo test -p mesh-llm-host-runtime auto_route --lib -- --test-threads=1`, but this local checkout did not have the prepared llama.cpp static ABI artifacts, so `skippy-ffi` failed before tests with missing `llama-common`.

PR #736, opened 2026-05-29
feat: native rust SDK for rust consumers.

michaelneale:
Needed to have a rust API surface that another rust app could consume to launch nodes, similar to swift/kotlin etc.

Tested this out with block/sprout to check it could consume it.

## Rust embedded usage

```toml
mesh-llm-host-runtime = { git = "https://github.com/Mesh-LLM/mesh-llm.git", rev = "<commit>", default-features = false }
```

```rust
use mesh_llm_host_runtime::sdk::{EmbeddedMeshNodeConfig, start_embedded_node};

let config = EmbeddedMeshNodeConfig::builder()
    .client()
    .auto_join(true)
    .api_port(9337)
    .console_port(3131)
    .console_ui(false)
    .build();

let node = start_embedded_node(config).await?;
let api_base = node.api_base_url();
let status = node.status().await?;

node.stop().await?;
```


i386 (dismissed):
I don’t think `mesh-llm-host-runtime` should be the public crate Rust consumers depend on.

`mesh-llm-host-runtime` is our internal host/CLI runtime crate. It owns a lot more than a stable SDK boundary: CLI startup behavior, runtime orchestration, plugins, config loading, OpenAI frontend wiring, skippy/native build details, dashboard/output state, etc. If we document this as the Rust SDK dependency, we effectively make a large internal API surface part of our public contract.

I’d rather split a dedicated public crate, probably:

```text
crates/mesh-llm-sdk
```

with modules/features like:

```toml
[features]
default = ["client", "serve"]
client = ["dep:mesh-llm-host-runtime"]
serve = ["dep:mesh-llm-host-runtime"]
web-ui = ["mesh-llm-host-runtime/web-ui"]
```

and public modules like:

```rust
mesh_llm_sdk::client
mesh_llm_sdk::serve
```

The API could stay close to what this PR introduces:

```rust
use mesh_llm_sdk::client::EmbeddedClientConfig;

let node = mesh_llm_sdk::client::start(
    EmbeddedClientConfig::builder()
        .auto_join(true)
        .api_port(9337)
        .console_port(3131)
        .build(),
).await?;

let api_base = node.api_base_url();
let status = node.status().await?;
node.stop().await?;
```

and:

```rust
use mesh_llm_sdk::serve::EmbeddedServeConfig;

let node = mesh_llm_sdk::serve::start(
    EmbeddedServeConfig::builder()
        .model("unsloth/Qwen3-0.6B-GGUF:Q4_K_M")
        .api_port(9337)
        .console_port(3131)
        .build(),
).await?;
```

Shared concepts like `EmbeddedNodeHandle`, `NetworkConfig`, `StorageConfig`, status, and errors can live at the SDK root. `client` and `serve` are modes of one embedded-node lifecycle, so I’d keep them as modules/features rather than separate crates unless we later find a real dependency boundary.

This also gives us a cleaner story for Swift/Kotlin/Node: those SDKs should wrap the same Rust SDK boundary via UniFFI/C/NAPI rather than each binding into `mesh-llm-host-runtime` separately. That way lifecycle semantics, config defaults, startup readiness, shutdown, isolated config behavior, and errors are defined once.

A few concrete issues I think should be fixed before we bless this as an SDK surface:

1. `run_embedded_runtime()` still calls the normal CLI entrypoint initialization. That currently includes CLI-only helpers that can call `std::process::exit`, such as the no-arg usage path. Embedded SDK code must never terminate the host application process.

2. If `start_embedded_node()` times out waiting for `/api/status`, it returns an error after spawning the runtime thread, but does not shut that thread down or join it. That can leave an embedded node running with ports still bound and no handle returned to the caller.

3. `skippy-ffi/build.rs` now auto-builds patched llama.cpp by default. That means downstream Rust consumers can trigger a heavyweight native build/source prep as a Cargo build-script side effect. For an SDK dependency, I think that needs to be explicit/opt-in or handled by a more deliberate packaging/sys-crate strategy.

So my preferred shape is: keep the embedded Rust API, but move it behind a narrow `mesh-llm-sdk` crate before documenting it as the public consumer dependency. `mesh-llm-host-runtime` should remain an implementation detail.

i386:
Will probably want https://github.com/Mesh-LLM/mesh-llm/pull/670 too, so you can optionally bundle in the console.

michaelneale:
@i386 thanks. On console, I believe that is already a feature you can just not compile in (and none of it comes, saves a lot of size, did that a while ago).

i386 (approved):
Reviewed the latest head against the prior comments. The SDK boundary, embedded runtime lifecycle, CLI-exit safety, startup cleanup, and console feature direction have been addressed. Remaining native auto-build behavior is already captured in the existing unresolved thread and docs call out the intended runtime packaging direction.

PR #737, opened 2026-05-29
Add coding model capability

i386:
## Summary

Adds `coding` as a first-class model capability for routing and agent model selection.

- Adds `coding: CapabilityLevel` to `ModelCapabilities`, with `none`, `likely`, and `supported` status values.
- Detects code-specialized models from names and metadata such as `coder`, `codestral`, `devstral`, `starcoder`, `deepseek-coder`, `qwen-coder`, and generic code/programming signals.
- Carries `coding` over the protobuf model capability advertisement as an additive field.
- Surfaces `coding` in `/v1/models` capabilities and `coding_status`, alongside existing `tool_use_status`.
- Prefers coding-capable models for requests classified as code, while preserving existing fallback behavior.

## Protocol

This is additive: `ModelCapabilities.coding` uses a new protobuf field number. Older peers ignore it, and newer peers treat missing values as `none`.

## Validation

- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-host-runtime --lib` -> 1654 passed, 0 failed, 6 ignored
- `rustfmt --edition 2024 --check ...`
- `git diff --check`

### DIFF
--- a/crates/mesh-llm-host-runtime/src/cli/commands/models/formatters.rs
+++ b/crates/mesh-llm-host-runtime/src/cli/commands/models/formatters.rs
@@ -211,6 +211,7 @@ pub(crate) fn capabilities_json(caps: ModelCapabilities) -> Value {
         "audio": caps.audio_status(),
         "reasoning": caps.reasoning_status(),
         "tool_use": caps.tool_use_status(),
+        "coding": caps.coding_status(),
     })
 }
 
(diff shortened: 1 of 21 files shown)


i386:
@michaelneale

michaelneale:
Not clear what this does that other signals (like tools) don't - is it just based on name? Not deeper metadata? (Then shouldn't need to bake it into this level). I've never seen models that say they are good at code be good at code so I am wary of classifying them yet this way

michaelneale:
Without this do we miss something and maybe models have improved relative to being classified as "coding" and I didn't notice? (Ie does this help)

ndizazzo (commented):
I'm also a bit confused on this one - mainly the "why" along with what Mic was curious about. 

At first I thought it was limited to the `models installed` or `models show`, but realize there's some protocol + API response changes too, so I guess I wonder to what end?

ndizazzo:
Also `encoder` would trigger as a false positive

i386:
Was for A2A routing. maybe this is a bad idea. let me close this and rethink.

PR #738, opened 2026-05-29
Move openai-endpoint to external plugin

### LINKED ISSUE #689: Clarity on openapi-endpoint plugin
Hi all. 
I've been watching this project since it was introduced into goose. Awesome work and a lot of promise!

The thing I've been hung up on is the `openai-endpoint` plugin. From the cli help menu:

```
External backends (vLLM, TGI, Ollama):
  Add to ~/.mesh-llm/config.toml:
    [[plugin]]
    name = "openai-endpoint"
  Then: mesh-llm serve     (or: mesh-llm client  for client-only mode)
```

I am very keen to try out this feature as it might be the definitive feature that unlocks this tool at work (hooking into existing infrastructure already exposed on openai api). Additionally at home I have some custom llamacpp flags, that I cannot pass through mesh-llm, but can run an instance of llamacpp and could expose that through mesh-llm if this functions.

Both use cases fit my mental model of how this feature works. Which may not be correct.

What I have done, is set up the config as described. I have observed the plugin loading in the debug logs, and have observed it hitting the `/v1/models` endpoint over and over. 
I have not seen any models that are exposed by the endpoints appear in the mesh, and when hitting the `localhost:3131/api/status` endpoint they're not present there either.

Rather than calling it broken, first I am wondering if I am expecting the right thing from this feature.
Am I correct in assuming I should only have to add the config, and url (via config or env variable) and then the models exposed should show up and be usable?
If that is not how it works. I'd love to know what it does so I can toy around with it. 😄 

Many thanks!!

i386:
Users can now install `openai-endpoint` from the plugin catalog or directly from [Mesh-LLM/openai-endpoint](https://github.com/Mesh-LLM/openai-endpoint), then configure `[[plugin]] name = "openai-endpoint"` with a `url` for any OpenAI-compatible backend.

## What Changed

- Removed the in-tree `openai-endpoint` plugin implementation and runtime dispatch.
- Resolve configured external plugins by name from the plugin store when `command` is omitted.
- Pass configured plugin URLs through the generic plugin launch contract as `MESH_LLM_PLUGIN_URL`; the host no longer has OpenAI-endpoint-specific config/runtime hooks.
- Removed the OpenAI-endpoint-specific config SDK helper in favor of generic plugin config editing.
- Updated CLI/docs to install `openai-endpoint` from the plugin catalog or GitHub.
- Fixed #689 by surfacing endpoint-discovered model IDs in `/api/status` and local gossip so they are visible and routeable through the mesh.

## Catalog

Added `openai-endpoint` to the Hugging Face plugin catalog:
https://huggingface.co/datasets/meshllm/plugin-catalog/commit/b5709261bcc48233869f385eb6fe41e5d0c4c51d

Updated the external plugin repo to use the generic plugin URL contract:
https://github.com/Mesh-LLM/openai-endpoint/commit/693aa10

## Validation

- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-config --lib`
- `cargo test -p mesh-llm-host-runtime external_plugin --lib`
- `cargo test -p mesh-llm-host-runtime openai_endpoint --lib`
- `cargo test -p mesh-llm-host-runtime inference_endpoint --lib`
- `cargo test -p mesh-llm-host-runtime --lib` (1659 passed, 0 failed, 6 ignored)
- `cargo fmt --all -- --check` in `Mesh-LLM/openai-endpoint`
- `cargo test --lib` in `Mesh-LLM/openai-endpoint`


ndizazzo (approved):
Seems fine to me

One small duplicate function `append_external_inference_models` that we can extract into a shared spot.

PR #740, opened 2026-05-29
Add native skippy crash breadcrumbs

i386:
## Summary

Adds flushed `mesh-llm:` breadcrumb lines to `skippy-native.log` around the native skippy model/session FFI calls that can hard-crash the process. If llama.cpp aborts before returning to Rust, the native log now shows which call was entered and the runtime configuration used for that call.

## Why

A split-runtime crash can currently leave the native log ending on llama.cpp's last internal line, which makes it hard to tell whether the process died in model open, multipart model open, or session creation. These breadcrumbs provide a last-known Rust-side boundary and include context, lane, batch, device, cache, and layer settings.

## Breadcrumb Examples

Successful single-file model open now looks like:

```text
mesh-llm: skippy_model_open begin path=/models/Qwen3.5-4B.gguf stage_index=0 layers=0..18 ctx=65536 lanes=8 n_batch=512 n_ubatch=128 n_gpu_layers=999 backend=Vulkan0 cache_k=1 cache_v=1 flash_attn=Auto load_mode=RuntimeSlice include_embeddings=true include_output=false filter_tensors_on_load=true
llama_model_loader: loaded meta data with 46 key-value pairs and 426 tensors from /models/Qwen3.5-4B.gguf
...
mesh-llm: skippy_model_open returned status=Ok
```

A native crash during model open would leave the `begin` line without a matching `returned` line:

```text
mesh-llm: skippy_model_open begin path=/models/Qwen3.5-4B.gguf stage_index=0 layers=0..18 ctx=65536 lanes=8 n_batch=512 n_ubatch=128 n_gpu_layers=999 backend=Vulkan0 cache_k=1 cache_v=1 flash_attn=Auto load_mode=RuntimeSlice include_embeddings=true include_output=false filter_tensors_on_load=true
llama_context: constructing llama_context
llama_context: n_seq_max     = 8
llama_context: n_ctx         = 65536
# process exits here; no `mesh-llm: skippy_model_open returned ...`
```

Multipart layer-package loads get the same boundary, using `skippy_model_open_from_parts`:

```text
mesh-llm: skippy_model_open_from_parts begin parts=/cache/pkg/meta.gguf,/cache/pkg/layers-00000-00018.gguf stage_index=0 layers=0..18 ctx=8192 lanes=2 n_batch=512 n_ubatch=128 n_gpu_layers=999 backend=auto cache_k=1 cache_v=1 flash_attn=Auto load_mode=LayerPackage include_embeddings=true include_output=false filter_tensors_on_load=true
...
mesh-llm: skippy_model_open_from_parts returned status=Ok
```

Session creation is also marked, which helps distinguish a model-load crash from a KV/session allocation crash:

```text
mesh-llm: skippy_session_create begin
llama_context: constructing llama_context
...
mesh-llm: skippy_session_create returned status=Ok
```

## Validation

- `just build`
- `cargo test -p skippy-runtime --lib`

### DIFF
--- a/crates/skippy-runtime/src/lib.rs
+++ b/crates/skippy-runtime/src/lib.rs
@@ -657,11 +657,11 @@ fn sanitize_native_log_note(note: &str) -> String {
 
 pub fn write_native_log_note(note: impl AsRef<str>) {
     let note = sanitize_native_log_note(note.as_ref());
-    if let Ok(mut guard) = native_log_file().lock() {
-        if let Some(writer) = guard.as_mut() {
-            let _ = writeln!(writer, "mesh-llm: {note}");
-            let _ = writer.flush();
-        }
+    if let Ok(mut guard) = native_log_file().lock()
+        && let Some(writer) = guard.as_mut()
+    {
+        let _ = writeln!(writer, "mesh-llm: {note}");
+        let _ = writer.flush();
     }
 }
 
--- a/crates/skippy-runtime/src/lib.rs
+++ b/crates/skippy-runtime/src/lib.rs
@@ -649,6 +649,22 @@ fn flush_native_log_writer<W: Write>(writer: &mut Option<LineWriter<W>>) {
     }
 }
 
+fn sanitize_native_log_note(note: &str) -> String {
+    note.chars()
+        .map(|ch| if matches!(ch, '\n' | '\r') { ' ' } else { ch })
+        .collect()
+}
+
+pub fn write_native_log_note(note: impl AsRef<str>) {
+    let note = sanitize_native_log_note(note.as_ref());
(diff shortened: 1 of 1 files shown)


crates/skippy-runtime/src/lib.rs
```diff
@@ -649,6 +649,22 @@ fn flush_native_log_writer<W: Write>(writer: &mut Option<LineWriter<W>>) {
     }
 }
 
+fn sanitize_native_log_note(note: &str) -> String {
```
ndizazzo, line 652: Sanitize makes me think - is there anything we shouldn't bundle in the doctor bundle?

ndizazzo (approved):

PR #742, opened 2026-05-29
Retry uncommitted remote tunnel failures

i386:
## Summary

Retries remote mesh HTTP tunnels once when the tunnel fails before any upstream response has been committed to the downstream client.

This covers transient tunnel/read failures such as a connection loss while waiting for the first upstream response. The retry uses a fresh tunnel and is only applied to uncommitted failures, so requests are not replayed after an upstream response has already been delivered.

## Why

A2A/OpenCode end-to-end testing exposed intermittent 503s from the mesh proxy when a remote host connection dropped before the first response bytes were available:

- `failed to read response from host ... connection lost`
- `all 1 tunnel(s) to hosts for Some("...") failed (mesh request)`

The first fix added the retry wrapper but only wired one routing path. This PR also wires the direct mesh request loop so model-routed remote requests use the same retry policy.

## Validation

- `just build`
- `cargo test -p mesh-llm-host-runtime --lib test_remote_retry_policy_only_retries_uncommitted_failures` passed in the original worktree with this transport change before isolating the PR branch.

In the clean PR worktree, the focused test was retried after `just build` prepared the native llama libraries, but the test-profile link failed because `/private/tmp` ran out of disk space:

- `ld64.lld: error: failed to write output ... No space left on device`

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/transport.rs
@@ -114,6 +114,8 @@ enum RouteAttemptResult {
     ClientDisconnected,
 }
 
+const REMOTE_UNCOMMITTED_RETRIES: usize = 1;
+
 fn route_attempt_result_label(result: &RouteAttemptResult) -> &'static str {
     match result {
         RouteAttemptResult::Delivered { .. } => "delivered",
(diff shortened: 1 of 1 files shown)


ndizazzo (approved):
LGTM!

PR #743, opened 2026-05-29
fix(cuda-crash): add preventitive crash framework for skippy allocation OOMs

ndizazzo:
I've been battling crashes pretty often and had an idea to try and track allocation sizes, and cowardly skipping the `alloc` to prevent the entire process from getting an `abort()` (other models might be loaded). 

We can evolve this to clean up the offending model, or print some kind of warning. 

Next thing we do might be to gate behind an option, or (less usefully) a debug flag.

### DIFF
--- a/crates/skippy-ffi/build.rs
+++ b/crates/skippy-ffi/build.rs
@@ -1,4 +1,5 @@
 fn main() {
+    println!("cargo::rustc-check-cfg=cfg(has_cuda)");
     println!("cargo:rerun-if-env-changed=LLAMA_STAGE_BUILD_DIR");
     println!("cargo:rerun-if-env-changed=LLAMA_STAGE_LIB_DIR");
     println!("cargo:rerun-if-env-changed=LLAMA_STAGE_LINK_MODE");
@@ -134,6 +135,7 @@ fn main() {
     );
     if has_cuda {
         println!("cargo:rustc-link-lib=static=ggml-cuda");
+        println!("cargo:rustc-cfg=has_cuda");
     }
     let has_hip = static_archive_exists(
         &build_dir,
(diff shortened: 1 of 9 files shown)


ndizazzo:
@i386 this is preliminary, but what do you think of this?

i386 (changes requested):
Requesting changes for the CUDA OOM prevention path.

1. **CUDA alloc hook never actually prevents the abort path.** `0083-cuda-alloc-hook.patch` adds a hook that can return `false`, but the callback added in `0084-alloc-tracker-api.patch` always returns `true`. The Rust precheck then only requires `free >= 5% of max_seen`, so a repeat large allocation can still pass the precheck and hit `cuMemCreate` OOM. This does not implement the PR body's "skip the alloc" behavior. The hook needs to veto based on the actual upcoming reserve/allocation requirement, or the precheck needs to require enough free memory for the expected allocation plus margin.

2. **Split-stage activation prefill paths bypass the new memory guard.** `RuntimeState::prefill` and media prefill call `check_memory`, but `prefill_frame_with_positions` and `prefill_final_frame_sampled` do not. Those are what binary split serving calls for `PrefillEmbd` / `PrefillFinalEmbd`, so split serving can still hit the native OOM abort this PR is trying to avoid.

3. **New C ABI symbols are not versioned/exported consistently.** The PR adds Rust bindings for `skippy_backend_device_check_memory` and tracker symbols, but leaves `ABI_VERSION_PATCH` at `24`. Also, the memory-check patch only changes `src/skippy.cpp` and does not add a `LLAMA_API` declaration to `include/skippy.h`, unlike the tracker API. Please bump the skippy ABI version in both C and Rust and expose the new public function in the header.

ndizazzo:
Reopen in the future if needed, in a more polished state

PR #751, opened 2026-05-30
Load versioned native runtimes from release artifacts

i386:
Stacked on #736.

MeshLLM can now ship a single host binary per OS/architecture and resolve version-matched native runtimes separately. Runtime selection, install, cache management, diagnostics, and release manifest handling live in shared Rust-owned code instead of shell installer heuristics.

## Summary
- Add `mesh-llm-native-runtime`, a shared crate for native runtime manifests, flavors, host profiles, compatibility filtering, ranking, cache layout, install/remove/prune policy, resolver diagnostics, and load plans. The crate README documents the manifest formats, resolver API, cache semantics, and loader boundary.
- Add `mesh-llm runtime list`, `install`, `remove`, and `prune`; with no argument, `runtime install` resolves and installs the recommended compatible runtime for this host.
- Add native runtime diagnostics to `mesh-llm doctor`, including selected runtime, cache path, host profile, and installed runtime counts.
- Add a feature-gated dynamic Skippy FFI loader (`dynamic-native-runtime`) that loads the selected cached native runtime libraries before host runtime startup.
- Package true native runtime artifacts from patched llama.cpp/Skippy shared libraries via `scripts/package-native-runtime.sh` and verify them with `scripts/verify-native-runtime-package.sh`.
- Generate `native-runtimes.json` from native runtime archives during release publishing, while keeping native SDK artifacts separate.
- Have install scripts and non-Windows autoupdate fetch `native-runtimes.json`, run `mesh-llm runtime install`, then prune old runtimes for the active MeshLLM version.

## Architecture
- Native runtimes are release artifacts, not implicit Cargo builds.
- `mesh_version` must exactly match the running MeshLLM version; Skippy ABI version is diagnostic metadata, not the compatibility boundary.
- Release host binaries build with dynamic native runtime loading; local developer builds still support the existing embedded/static ABI path unless the dynamic feature is enabled.
- Hardware detection falls back to platform probes before a dynamic runtime is loaded, so `mesh-llm runtime install` can select a runtime without already having one loaded.
- The runtime cache remains versioned as `<cache>/mesh-llm/native-runtimes/<mesh_version>/<native_runtime_id>/`.

## CLI Shape
```bash
mesh-llm runtime list --installed
mesh-llm runtime list --available --manifest native-runtimes.json
mesh-llm runtime install
mesh-llm runtime install cuda
mesh-llm runtime remove meshllm-native-runtime-linux-x86_64-cuda
mesh-llm runtime prune
mesh-llm runtime prune --active-only
mesh-llm doctor --json
```

## Validation
- `scripts/package-native-runtime.sh --build --backend metal --target aarch64-apple-darwin --out target/native-runtime-smoke`
- `scripts/verify-native-runtime-package.sh target/native-runtime-smoke/meshllm-native-runtime-darwin-aarch64-metal.tar.gz`
- `scripts/generate-native-runtime-release-manifest.sh --tag v0.68.0 --repo Mesh-LLM/mesh-llm --out target/native-runtime-smoke/native-runtimes.json target/native-runtime-smoke/meshllm-native-runtime-darwin-aarch64-metal.tar.gz`
- Dynamic binary smoke: install the local Metal runtime artifact and run `mesh-llm doctor --json`, confirming the selected cached runtime loads.
- `cargo check -p mesh-llm --features dynamic-native-runtime`
- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-native-runtime`
- `cargo clippy -p mesh-llm --features dynamic-native-runtime --all-targets -- -D warnings`
- `cargo test -p mesh-llm-host-runtime --lib` (`1744 passed; 7 ignored`)
- `bash -n scripts/package-native-runtime.sh scripts/verify-native-runtime-package.sh scripts/build-llama.sh scripts/build-release.sh scripts/generate-native-runtime-release-manifest.sh`
- `git diff --check`

PowerShell parse validation was not run locally because `pwsh`/`powershell` is not installed in this environment.


michaelneale:
I am not smart enough to review this by hand but I know it is needed and directionally right: 

"A MeshLLM SDK
consumer should be able to source Rust crates from crates.io without compiling
llama.cpp or Skippy native code as a default side effect."

so are the binaries pre-compiled and will be downloaded when needed from crates or from gh releases? (as crates doesnt' like large bins at all)?

crates/mesh-llm-native-runtime/src/flavor.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl std::error::Error for NativeRuntimeFlavorParseError {}
+
+impl NativeRuntimeFlavor {
+    pub fn as_str(&self) -> &str {
+        match self {
+            Self::Cpu => "cpu",
+            Self::Metal => "metal",
+            Self::Cuda => "cuda",
+            Self::CudaBlackwell => "cuda-blackwell",
```
ndizazzo, line 34: We need to remove `CudaBlackwell` in favour of either:

1) Runtime: CUDA12 + Runtime: CUDA13, or
2) Runtime: CUDA + RuntimeVariant: 12 / RuntimeVariant: 13

I prefer option 1. Reason being, I just did some work around this in our release pipeline to add support for NVIDIA DGX Spark, and the table for CUDA support isn't just Blackwell anymore - it's anything on sm_110+, and Rubin is now on the horizon, likely > sm_130. 

The minimum CUDA version for those will be CUDA 13.x

.github/workflows/release.yml
```diff
@@ -242,6 +242,68 @@ jobs:
...
+
+      - name: Upload native runtime
+        uses: actions/upload-artifact@v6
+        with:
+          name: release-native-runtime-${{ matrix.artifact_suffix }}
+          path: |
+            dist/native-runtimes/*.tar.gz
+            dist/native-runtimes/*.sha256
+          if-no-files-found: error
+
```
ndizazzo, line 306: I think we'll likely want a job to compute a JSON blob representing our build structure, given the complexity of all our runtimes, binaries, and smokes, etc.

A big CI refactor will need to likely happen to make this all sane. Not something we should worry about in this PR, but structurally, I'd like for us to use separate workflows for our host binary and runtimes, then have the release / pr workflows call into them. 

This way, we don't have disparate jobs in each file that can drift.

ndizazzo (commented):
Generally, this is where my mind was going too. **I have two requests to add here:**

1) **We need a "fake" / "stub" runtime** That outputs fixtures or expected results. This can be our PR / testing engine for mesh-llm. All of the support code to connect / peers / protocols / can independly be tested without actual CUDA / ROCm / Vulkan stuff, which will make development lightning fast.
2) I want to be able to plug old runtimes in to new mesh-llm host binaries. We need to separate / stabilize the interface between the runtime and the host so calling into them takes a standard shape (within reason - some breaking changes might happen as features arrive). 

For #2 this means - if we totally flip up a release of a runtime, users can still use older versions if there are problems. AND, we can have different channels for experimental features: BETA / Stable

crates/mesh-llm-native-runtime/src/flavor.rs
ndizazzo: In fact - I also want to add something to this where we have system compatibility where we can enumerate the SM_* version, and do lookups to see which runtimes _are compatible to offer the user_. 

That would be a nice usability experience.

Same for ROCm and whatever they do to indicate version support between 7.0, 7.1, and 7.2. I'm sure it's simpler than whatever NVIDIA is doing.

i386:
> so are the binaries pre-compiled and will be downloaded when needed from crates or from gh releases? (as crates doesnt' like large bins at all)?

Yes. You can even bundle the native runtime binaries in your app and just load them.

i386:
> Generally, this is where my mind was going too. **I have two requests to add here:**
> 
> 1) **We need a "fake" / "stub" runtime** That outputs fixtures or expected results. This can be our PR / testing engine for mesh-llm. All of the support code to connect / peers / protocols / can independly be tested without actual CUDA / ROCm / Vulkan stuff, which will make development lightning fast.
> 2) I want to be able to plug old runtimes in to new mesh-llm host binaries. We need to separate / stabilize the interface between the runtime and the host so calling into them takes a standard shape (within reason - some breaking changes might happen as features arrive). 
> 
> For #2 this means - if we totally flip up a release of a runtime, users can still use older versions if there are problems. AND, we can have different channels for experimental features: BETA / Stable 

Totally fine with both of those goals. I would restrict version matching just to skippy abi version  rather than mesh version when deducing if the runtime can be loaded or not

PR #760, opened 2026-05-30
Balance auto context and slots for shared mesh launches

IvGolovach:
## Summary

Auto runtime planning now chooses resource defaults based on launch intent, not public/private mesh status.

Dedicated local launches prefer the deepest safe context. Shared mesh-serving launches preserve concurrency by choosing the deepest context that still supports the default parallel slot target. Shared-serving modes now include `--auto`, `--publish`, `--discover`, and explicit invite joins via `--join`.

Explicit operator overrides still win: CLI/model `ctx_size` and `parallel` settings are preserved when provided.

## Why

The old framing treated public and private meshes as separate planning categories. That is not the right boundary: private invite meshes and large organization meshes can still be shared, multi-workload serving surfaces.

The relevant distinction is whether the node is acting as a dedicated local model launch or as part of a shared mesh-serving surface where preserving slots/concurrency is expected.

## Diff scope

- Adds a `RuntimeResourcePlanningProfile` for dedicated-local vs shared-mesh serving defaults.
- Uses shared-mesh concurrency planning for `--auto`, `--publish`, `--discover`, and `--join`.
- Keeps explicit `ctx_size` / `parallel` overrides ahead of automatic planning.
- Keeps runtime-demand model loading aligned with per-model `ctx_size` overrides.

## Branch integrity

- Base branch: `main`
- Validated base SHA: `ccefb8a8509d82b2bf517e82ef43e7e57bcfaba6`
- Head SHA: `b76d87c50f9287950b88ad7048600f19c1907b38`
- Merge-base: `ccefb8a8509d82b2bf517e82ef43e7e57bcfaba6`
- Diff proof computed against fetched `origin/main`.

## Commit integrity

Introduced commit:

```text
b76d87c50f9287950b88ad7048600f19c1907b38 Balance auto runtime context for shared mesh serving
```

One logical runtime-planning change. Final PR diff contains only the intended runtime files:

```text
M crates/mesh-llm-host-runtime/src/runtime/context_planning.rs
M crates/mesh-llm-host-runtime/src/runtime/local.rs
M crates/mesh-llm-host-runtime/src/runtime/mod.rs
```

## Ledger / Version

Ledger: not applicable — not required for selected validation mode/change family.

Version: not applicable — no release/version sync required for this non-release runtime planning change.

## Diff hygiene

```text
git diff --check: PASS, no output
git diff --cached --check: PASS, no output
```

No `.env`, local environment files, secrets, tokens, credentials, caches, build output, pycache files, or unrelated generated artifacts are included.

## Validation mode and proof

Validation mode: Mode 3 — shared runtime resource planning. This changes runtime defaults for context/slot planning across startup and runtime-demand loading, but does not change protocol, migrations, auth, permissions, deployment tooling, or external service contracts.

Local validation:

```text
cargo fmt --all: PASS
cargo fmt --all -- --check: PASS
LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime context_planning --lib -- --test-threads=1: PASS, 10 passed
LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime runtime_load_ctx_size --lib -- --test-threads=1: PASS, 2 passed
LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo test -p mesh-llm-host-runtime shared_mesh_modes --lib -- --test-threads=1: PASS, 1 passed
cargo test -p model-artifact gguf --lib -- --test-threads=1: PASS, 16 passed
LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal cargo check -p mesh-llm: PASS
LLAMA_STAGE_BUILD_DIR=/Users/Funtland/Downloads/mesh-llm/.deps/llama-build/build-stage-abi-metal /opt/homebrew/bin/cargo-clippy clippy -p mesh-llm-host-runtime --all-targets -- -D warnings: PASS
```

Not run locally: live multi-node mesh startup smoke — no local multi-node runtime mesh endpoint was available; deterministic resource-planning, startup profile, runtime-load override, shipped-binary check, and clippy cover the changed paths. Mandatory remote CI is the final full proof for merge.

## Migration notes

Not applicable — no DB migration changed.

## CI context confirmation

PASS — required CI completed for head `b76d87c50f9287950b88ad7048600f19c1907b38`. CI context names unchanged.

## Runtime safety

No new blocking locks, unbounded queues, protocol messages, gossip fields, or invariant removals were introduced. No invariant regression introduced.

## Documentation integrity

Not applicable — no docs/runbooks/commands changed.

## Rollback plan

Rollback: revert this PR.

DB downgrade: not applicable.

Data repair: not applicable.

Operational caveats: automatic defaults for shared mesh-serving launches would return to the previous context/slot behavior after revert.

## Known residual risks

None known after required CI passed on the final head SHA.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/runtime/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/mod.rs
@@ -18,6 +18,7 @@ use self::capacity::{
     RuntimeCapacityLedger, RuntimeCapacityPool, RuntimeCapacityRequest, RuntimeCapacityReservation,
     model_fits_runtime_capacity,
 };
+use self::context_planning::RuntimeResourcePlanningProfile;
 use self::discovery::{lan_rediscovery, nostr_rediscovery, start_new_mesh};
 use self::interactive::InitialPromptMode;
 use self::local::{
@@ -1220,6 +1221,7 @@ struct StartupLocalModelTask {
     n_ubatch: Option<u32>,
     flash_attention: FlashAttentionType,
     parallel_override: Option<usize>,
+    resource_planning_profile: RuntimeResourcePlanningProfile,
     openai_guardrail_policy: OpenAiGuardrailPolicyHandle,
     split: bool,
     skippy_telemetry: skippy::SkippyTelemetryOptions,
@@ -1299,6 +1301,7 @@ struct StartupLoopContext<'a> {
     n_ubatch: Option<u32>,
     flash_attention: FlashAttentionType,
     parallel_override: Option<usize>,
+    resource_planning_profile: RuntimeResourcePlanningProfile,
     openai_guardrail_policy: OpenAiGuardrailPolicyHandle,
     skippy_telemetry: &'a skippy::SkippyTelemetryOptions,
     survey_telemetry: &'a survey::SurveyTelemetry,
@@ -1371,6 +1374,7 @@ struct StartupLaunchRuntimeContext<'a> {
     n_ubatch: Option<u32>,
     flash_attention: FlashAttentionType,
     parallel_override: Option<usize>,
+    resource_planning_profile: RuntimeResourcePlanningProfile,
     openai_guardrail_policy: OpenAiGuardrailPolicyHandle,
     skippy_telemetry: &'a skippy::SkippyTelemetryOptions,
     survey_telemetry: &'a survey::SurveyTelemetry,
@@ -1841,6 +1845,7 @@ async fn startup_handle_local_fallback_event(
             n_ubatch_override: ctx.n_ubatch,
             flash_attention_override: ctx.flash_attention,
             parallel_override: ctx.parallel_override,
+            planning_profile: ctx.resource_planning_profile,
             openai_guardrail_policy: ctx.openai_guardrail_policy.clone(),
             skippy_telemetry: ctx.skippy_telemetry.clone(),
             survey_telemetry: ctx.survey_telemetry.clone(),
@@ -2173,6 +2178,7 @@ async fn startup_launch_runtime(
         n_ubatch,
         flash_attention,
         parallel_override,
+        resource_planning_profile,
         openai_guardrail_policy,
         skippy_telemetry,
         survey_telemetry,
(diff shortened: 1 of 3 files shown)


crates/mesh-llm-host-runtime/src/runtime/mod.rs
```diff
@@ -5695,6 +5705,22 @@ fn relay_policy_for_mesh_discovery_mode(
     }
 }
 
+fn runtime_resource_planning_profile(cli: &Cli) -> RuntimeResourcePlanningProfile {
+    if cli.auto || cli.publish || cli.discover.is_some() {
```
ndizazzo: I think `--join` should count as a shared/public mesh here too. Otherwise invite-joined meshes may get the wrong profile and end up with lower concurrency than expected.

ndizazzo (approved):
This ones fine - also needs a rebase tho

michaelneale:
hrm not sure - never intended this separation between public and private: even private nodes, the idea is that they may run several workloads not just model hosting, but then same with public as well. I'm not sure about the concept of having this separation along these lines (I can understand an operator makign decisions depending on it) but to me it all can blur together. The public mesh or an organisation with 2000 random nodes are kind of similar in effect, unless I am missing something?

IvGolovach:
Thanks both, this was the right callout.

I had framed the behavior too much around public vs private mesh status, but that is not the clean boundary here. Private invite meshes and org meshes can still be shared serving surfaces, so the runtime defaults should follow launch intent/workload shape instead.

I updated the PR to use a dedicated-local vs shared-mesh planning profile. Shared mesh planning now applies to `--auto`, `--publish`, `--discover`, and explicit invite joins via `--join`, while explicit `ctx_size` / `parallel` overrides still win.

I also added regression coverage for the shared modes, including `--join`, rebased the branch, and the final CI is green on the updated head. Really appreciate the review here — the resulting behavior is clearer and less surprising.

PR #761, opened 2026-05-30
Stream committed MoA answers for Responses clients

### LINKED ISSUE #618: MoA: stream the winning worker's tokens instead of one-shot delta

IvGolovach:
## Summary

This PR makes MoA streaming feel incremental for OpenAI Responses clients once the arbiter has committed a non-reducer answer.

For committed worker/early-exit answers, `/v1/responses` streaming now emits multiple `response.output_text.delta` events instead of a single full-answer delta. Reducer output and tool-call responses remain atomic, which keeps the existing MoA grace and tool-call contracts intact while addressing the visible streaming gap from #618.

Closes #618.

## Why

Issue #618 identified that MoA waited for all worker selection logic and then delivered the winning text as one large SSE delta. That made UI clients look frozen even when the final answer was available. This change implements the scoped unblock: stream the already-committed final answer in chunks, without introducing reducer streaming or mid-stream tool-call behavior.

## Diff scope

- Adds an explicit MoA final-text stream mode for committed non-reducer answers versus reducer answers.
- Applies that mode to both the direct MoA SSE response path and the progress-drip handoff path.
- Splits committed non-reducer text into multiple Responses output deltas.
- Keeps reducer output one-shot until reducer streaming is designed deliberately.
- Keeps tool-call streaming atomic.

## Compatibility

- No mesh protocol, wire schema, Skippy ABI, or release metadata changes.
- No change to MoA worker selection, reducer semantics, or grace timing.
- No change to non-streaming JSON responses.
- Tool-call responses keep the existing atomic chunk behavior.

## Validation

Validation tier: Tier 3 - OpenAI Responses/MoA streaming behavior; committed non-reducer MoA answers now chunk into Responses deltas while reducer and tool-call contracts remain bounded.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, `origin/main` at `fef0f4deddadca3e7d95b8b86e52fa3a60ca0942`.
- `cargo fmt --all`: PASS.
- `cargo test -p mesh-llm-host-runtime responses_sse_keeps_reducer_output_one_delta_for_long_content --lib -- --test-threads=1`: PASS, 1 passed.
- `cargo test -p mesh-llm-host-runtime responses_sse_emits_multiple_deltas_for_long_content --lib -- --test-threads=1`: PASS, 1 passed.
- `cargo test -p mesh-llm-host-runtime final_text_stream_mode --lib -- --test-threads=1`: PASS, 1 passed.
- `cargo test -p mesh-llm-host-runtime chat_sse_tool_calls_remain_atomic --lib -- --test-threads=1`: PASS, 1 passed.
- `cargo test -p mesh-llm-host-runtime network::openai::moa_gateway::tests --lib -- --test-threads=1`: PASS, 46 passed.
- `cargo test -p mesh-llm-host-runtime network::openai::moa_gateway::progress::tests --lib -- --test-threads=1`: PASS, 9 passed.
- `cargo test -p mesh-mixture-of-agents --lib -- --test-threads=1`: PASS, 112 passed.
- `cargo check -p mesh-llm`: PASS.
- `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS.

## Not run

- Live two-node MoA Responses smoke: not run because no local multi-node mesh with live models was available. Deterministic SSE writer coverage, progress-path coverage, MoA arbitration tests, shipped-binary check, and clippy cover the changed paths; PR CI should remain the final merge gate.

## Rollback

- Revert this PR.
- DB downgrade: not applicable.
- Data repair: not applicable.
- Operational caveats: none known.

## Known residual risks

- This intentionally does not stream reducer output or relay live upstream worker tokens before the arbiter commits. It implements the scoped #618 unblock for committed non-reducer answers.
- Final user-visible confidence still benefits from a live multi-node MoA smoke once a model mesh is available.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/mod.rs
@@ -243,14 +243,26 @@ async fn write_moa_response(
         match response_adapter {
             proxy::ResponseAdapter::OpenAiResponsesStream => (
                 "SSE-responses",
-                send_moa_as_responses_sse(tcp_stream, body, extra_headers).await,
+                send_moa_as_responses_sse(
+                    tcp_stream,
+                    body,
+                    extra_headers,
+                    final_text_stream_mode_for_result(moa_result),
+                )
+                .await,
             ),
             // None, OpenAiChatCompletionsStream, OpenAiResponsesJson all
             // get the chat.completion.chunk SSE shape — the JSON-mode
             // adapter caller will never set was_streaming=true.
             _ => (
                 "SSE-chat",
-                send_moa_as_sse(tcp_stream, body, extra_headers).await,
+                send_moa_as_sse(
+                    tcp_stream,
+                    body,
+                    extra_headers,
+                    final_text_stream_mode_for_result(moa_result),
+                )
+                .await,
             ),
         }
     } else if is_failure {
@@ -278,6 +290,22 @@ async fn write_moa_response(
     };
     if let Err(e) = result {
         tracing::warn!("MoA: response write failed ({mode}): {e}");
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(in crate::network::openai::moa_gateway) enum MoaFinalTextStreamMode {
+    OneShot,
+    ChunkedCommittedText,
+}
+
+pub(in crate::network::openai::moa_gateway) fn final_text_stream_mode_for_result(
+    result: &moa::TurnResult,
+) -> MoaFinalTextStreamMode {
+    if result.reducer_used {
+        MoaFinalTextStreamMode::OneShot
+    } else {
+        MoaFinalTextStreamMode::ChunkedCommittedText
     }
 }
 
@@ -753,8 +781,9 @@ async fn send_moa_as_sse(
     stream: TcpStream,
     response: &serde_json::Value,
     extra_headers: &[(&str, String)],
+    text_stream_mode: MoaFinalTextStreamMode,
 ) -> std::io::Result<()> {
-    send_moa_as_sse_inner(stream, response, extra_headers, false).await
+    send_moa_as_sse_inner(stream, response, extra_headers, false, text_stream_mode).await
 }
 
 /// Write the standard SSE response header block, with optional
@@ -782,6 +811,7 @@ pub(in crate::network::openai::moa_gateway) async fn send_moa_as_sse_inner(
     response: &serde_json::Value,
     extra_headers: &[(&str, String)],
     header_already_sent: bool,
+    text_stream_mode: MoaFinalTextStreamMode,
 ) -> std::io::Result<()> {
     if !header_already_sent {
         write_sse_response_headers(&mut stream, extra_headers).await?;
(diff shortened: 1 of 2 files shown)


i386:
@michaelneale FYI

michaelneale (approved):
seems a good fix.

PR #766, opened 2026-05-30
Harden LAN-only mDNS discovery

IvGolovach:
## Summary

LAN-only discovery is safer and less misleading now. A node advertising over mDNS still publishes bounded LAN capacity/model summaries and a token fingerprint, but it only advertises the proof-gated LAN detail endpoint when the management API is actually reachable from LAN peers.

Peers that already have the supplied join token can still fetch the full LAN discovery details through `POST /api/discovery/lan-details`, and that detail response now uses the same mesh name, region, and max-client metadata as the mDNS publisher.

## Why

The LAN/mDNS path should not expose raw join material, and it also should not advertise an HTTP details path that a default loopback-only management API cannot serve to other LAN hosts. Review caught that split correctly: discovery metadata, details reachability, and the local details route need to agree.

## Diff scope

- Adds token-fingerprint and challenge/proof helpers for LAN detail lookup.
- Adds a proof-gated `POST /api/discovery/lan-details` management API route.
- Keeps raw invite tokens out of mDNS TXT records and sanitized LAN detail responses.
- Gates advertised `details`/`proof_challenge` TXT metadata on LAN-reachable management API binding, currently `--listen-all`.
- Stores mesh publication metadata in API state so LAN detail responses match the mDNS publish loop's `mesh_name`, `region`, and `max_clients`.
- Adds LAN rediscovery for supplied-token mDNS peers when joinable LAN peers are lost.
- Keeps mDNS discovery on the LAN-only relay policy path.
- Updates CLI, mesh, design, and testing docs for the LAN detail/recovery contract.

## Compatibility

No mesh protobuf, QUIC ALPN, gossip schema, Skippy ABI, or plugin protocol changes. The management API route is additive, and LAN advertisements remain local mDNS metadata rather than a cross-mesh protocol signal.

## Branch integrity

Base branch: `main`  
Validated base: `f9bd75a97378be014f52c4e68c13e81d3399e65e`  
Branch status: `0 behind / 1 ahead` against `origin/main`  
Introduced commit: `88971f73 Harden mDNS LAN discovery contract`

## Validation

Validation mode: Tier 2R - post-review correction for PR #766. The PR remains runtime-significant, but this update is a narrow same-PR correction around LAN details advertisement reachability and publication metadata consistency.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, `origin/main` at `f9bd75a97378be014f52c4e68c13e81d3399e65e`
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo test -p mesh-llm-host-runtime lan_advertisement --lib -- --test-threads=1`: PASS, 4 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo test -p mesh-llm-host-runtime mdns --lib -- --test-threads=1`: PASS, 8 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo test -p mesh-llm-host-runtime lan_details --lib -- --test-threads=1`: PASS, 4 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo check -p mesh-llm`: PASS
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS

## Required remote gates

PASS - required GitHub CI completed on final SHA `88971f734a91ae9f596c3fe983dc2add03253417`. Review gate remains on the prior CHANGES_REQUESTED review until the reviewer re-reviews the fixed SHA.

## Not run

- Live two-node LAN mDNS smoke: not run because no second LAN node was available in this isolated worktree. Advertisement gating, route metadata, token-proof, mDNS mode, relay policy, shipped-binary check, and clippy cover the changed branches locally.
- Full workspace suite locally: not required for this post-review correction; mandatory PR CI is the final full-suite proof.

## Rollback

Rollback: revert this PR.  
DB downgrade: not applicable.  
Data repair: not applicable.  

### DIFF
--- a/crates/mesh-llm-host-runtime/src/api/routes/discover.rs
+++ b/crates/mesh-llm-host-runtime/src/api/routes/discover.rs
@@ -1,4 +1,7 @@
-use super::super::{MeshApi, http::respond_error};
+use super::super::{
+    MeshApi,
+    http::{respond_error, respond_json},
+};
 use crate::network::{discovery, nostr};
 use tokio::io::AsyncWriteExt;
 use tokio::net::TcpStream;
@@ -42,4 +45,67 @@ pub(super) async fn handle(stream: &mut TcpStream, state: &MeshApi) -> anyhow::R
         Err(_) => respond_error(stream, 500, "Failed to serialize").await?,
     }
     Ok(())
+}
+
+pub(super) async fn handle_lan_details(
+    stream: &mut TcpStream,
+    state: &MeshApi,
+    body: &str,
+) -> anyhow::Result<()> {
+    let request = match serde_json::from_str::<discovery::LanDetailsProofRequest>(body) {
+        Ok(request) => request,
+        Err(err) => {
+            respond_error(stream, 400, &format!("Invalid JSON body: {err}")).await?;
+            return Ok(());
+        }
+    };
+    let (mode, node, mesh_name, mesh_region, mesh_max_clients) = {
+        let inner = state.inner.lock().await;
+        (
+            inner.mesh_discovery_mode,
+            inner.node.clone(),
+            inner.mesh_name.clone(),
+            inner.mesh_region.clone(),
+            inner.mesh_max_clients,
+        )
+    };
+    if mode != discovery::MeshDiscoveryMode::Mdns {
+        respond_error(
+            stream,
+            404,
+            "LAN discovery details are only available in mDNS discovery mode",
+        )
+        .await?;
+        return Ok(());
+    }
+
+    let invite_token = node.invite_token().await;
+    if !discovery::verify_lan_details_token_proof(
+        &invite_token,
+        &request.token_fingerprint,
+        &request.challenge,
+        &request.proof,
+        current_unix_secs(),
+    ) {
+        respond_error(stream, 403, "Invalid LAN discovery proof").await?;
+        return Ok(());
+    }
+
+    let listing =
+        discovery::build_local_mesh_listing(&node, mesh_name, mesh_region, mesh_max_clients).await;
+    let response = discovery::LanDetailsResponse::from_local_listing(
+        listing,
+        request.token_fingerprint,
+        request.challenge,
+        Some(crate::VERSION),
+    );
+    respond_json(stream, 200, &response).await?;
+    Ok(())
+}
+
+fn current_unix_secs() -> u64 {
+    std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_secs()
 }
--- a/crates/mesh-llm-host-runtime/src/network/discovery.rs
+++ b/crates/mesh-llm-host-runtime/src/network/discovery.rs
@@ -1,7 +1,7 @@
 use anyhow::{Context, Result};
 use clap::ValueEnum;
 use mdns_sd::{DaemonStatus, ResolvedService, ServiceDaemon, ServiceEvent, ServiceInfo};
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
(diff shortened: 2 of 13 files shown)


crates/mesh-llm-host-runtime/src/network/discovery.rs
```diff
@@ -503,14 +567,80 @@ pub(crate) fn lan_token_fingerprint(token: &str) -> String {
...
+pub(crate) fn lan_details_challenge(token_fingerprint: &str, now_secs: u64) -> String {
+    lan_details_challenge_for_bucket(
+        token_fingerprint,
+        now_secs / LAN_DETAILS_CHALLENGE_WINDOW_SECS,
+    )
+}
+
+pub(crate) fn lan_details_token_proof(invite_token: &str, challenge: &str) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(b"mesh-llm-lan-details-proof-v1\0");
```
ndizazzo: NIT: inline static constant

crates/mesh-llm-host-runtime/src/network/discovery.rs
```diff
@@ -99,6 +103,12 @@ impl LanMeshAdvertisement {
                 (!listing.invite_token.trim().is_empty())
                     .then(|| lan_token_fingerprint(&listing.invite_token))
             });
+        let proof_challenge = token_fingerprint
+            .as_deref()
+            .map(|fingerprint| lan_details_challenge(fingerprint, current_unix_secs()));
+        let details_path = proof_challenge
+            .as_ref()
+            .map(|_| LAN_DETAILS_PATH.to_string());
```
ndizazzo, line 119: We expose this path,  but only the loopback interface is listened to unless `--listen-all` is set on the CLI options. I haven't verified locally, but I think this would make it so a default CLI option LAN peer can discover the path, but not POST to `host:port` when fetching details. 

Maybe we force binding to `0.0.0.0` on for this mDNS case, or only advertise details when we have `--listen-all` set. Not really sure what the best approach would be here.

crates/mesh-llm-host-runtime/src/api/routes/discover.rs
```diff
@@ -43,3 +46,63 @@ pub(super) async fn handle(stream: &mut TcpStream, state: &MeshApi) -> anyhow::R
...
+        &request.token_fingerprint,
+        &request.challenge,
+        &request.proof,
+        current_unix_secs(),
+    ) {
+        respond_error(stream, 403, "Invalid LAN discovery proof").await?;
+        return Ok(());
+    }
+
+    let listing = discovery::build_local_mesh_listing(&node, mesh_name, None, None).await;
```
ndizazzo: The route should use the same publication metadata as `publish_lan_loop`, or store `region/max_clients` in API state, so we can pass them before serving this response here.

ndizazzo (changes requested):
Some minor comments / improvement notes

IvGolovach:
I did a focused follow-up pass on the review notes here.

What changed:

- Inlined the LAN challenge/proof domain constants instead of keeping the one-off statics.
- Gated LAN details advertisement on the management API being LAN-reachable, so default loopback-only nodes do not publish a details URL that LAN peers cannot fetch.
- Moved the LAN details response onto the same publication metadata used by the mDNS advertisement path.

The branch is now refreshed, GitHub CI is green on the final SHA, and the PR body has the current validation proof. This should be ready for a re-review when you have a moment.

michaelneale:
@ndizazzo I am ok with this - as long as your thoughts are addressed - only hard requirement is that this is opt in and not on by default (in as much as it doesn't break any default experiences).

ndizazzo (approved):

Issue #265, opened 2026-04-13
Behavioral trust scoring for mesh nodes


realpercivallabs:
## Problem

When joining a public mesh via `--auto`, there's currently no way to:

- **Verify** a node actually ran the model it claims to run
- **Score** node reliability or honesty over time
- **Audit** inference requests for quality, latency consistency, or correctness
- **Detect** degraded or malicious nodes before they affect output

Public meshes today are effectively trust-me-bro architecture. Prompts on public meshes are visible to all peers with no behavioral accountability.

This is especially relevant as meshes scale beyond trusted friends to public discovery via Nostr relays.

## Proposed Approach: Behavioral Trust Layer

Rather than requiring hardware attestation (which needs Intel TDX / NVIDIA Confidential Computing), a **behavioral trust layer** scores nodes based on observed inference behavior over time.

I built a proof-of-concept that demonstrates this:

**[Percival-Labs/mesh-llm-trust](https://github.com/Percival-Labs/mesh-llm-trust)** — transparent proxy that sits between clients and mesh-llm, captures behavioral traces, computes per-node trust scores, and publishes them via Nostr.

### Architecture

```
Client (:9338)                    mesh-llm (:9337)
    │                                  │
    ▼                                  │
┌────────────────────────────┐         │
│  mesh-llm-trust proxy     │────────►│
│                            │◄────────│
│  ┌────────┐  ┌──────────┐ │
│  │ tracer │  │  scorer  │ │
│  └───┬────┘  └────┬─────┘ │
│  traces/*.jsonl    │       │
│               ┌────┴───┐   │
│               │ nostr  │   │
│               └────────┘   │
└────────────────────────────┘
         │
    NIP-85 trust scores
    → relay.damus.io, nos.lol
```

### What It Captures (per inference request)

- Latency (TTFB + total)
- Token counts (prompt + completion)
- Model claimed by node
- Success/failure
- Request classification (chat/code/reasoning)

### Trust Score Model

Each node gets a composite score (0-1000) across 4 dimensions:

| Dimension | Weight | What It Measures |
|-----------|--------|------------------|
| Success Rate | 35% | Ratio of successful responses |
| Latency Consistency | 25% | Low variance in response times (coefficient of variation) |
| Uptime | 20% | Heartbeat reliability over 30-min windows |
| Output Quality | 20% | Non-empty completions, reasonable token ratios |

### Nostr Integration (the key insight)

[...]

PR #767, opened 2026-05-30
Add local target reputation routing

IvGolovach:
## Summary

The local OpenAI/mesh proxy now learns from retryable target failures instead of treating every expired cooldown as fully equal again. When a peer times out or returns unavailable, the router keeps a bounded local reputation signal, prefers healthier alternatives when available, and exposes the local counters in `/api/status.routing_affinity.target_reputation`.

This update also adds the requested design note in `docs/NODE_REP.md` and a routing simulation test that exercises a three-peer candidate set, marks one adjacent peer unavailable, and verifies subsequent request flow avoids it until a successful outcome restores routing.

## Why

Issue #265 points at the need for behavioral peer scoring, but a real mesh-wide trust system needs output verification and protocol design. This PR deliberately implements the safe first layer: local routing-health reputation only. It improves availability and diagnostics without pretending to provide cryptographic trust, global peer reputation, or model-honesty attestation.

The review feedback was right that this is a sharp routing surface. The policy is now documented separately from the general mesh docs so future discussion can evolve the design without burying the trust boundary in implementation details.

## Diff scope

- Adds bounded, process-local target reputation state beside existing target cooldowns.
- Penalizes retryable timeout/unavailable outcomes while leaving client-side rejection, context overflow, and disconnect paths out of reputation scoring.
- Gradually recovers a penalized target after successful responses.
- Reorders eligible candidates away from penalized targets while preserving availability fallback semantics.
- Keeps `InferenceTarget::None` from becoming preferred only because a real target is penalized.
- Exposes local reputation counters through the existing affinity/status payload.
- Adds `docs/NODE_REP.md` for the local reputation model, non-goals, signals, status surface, testing boundary, and future cross-node design questions.
- Adds an affinity-routing simulation test for a multi-peer candidate list with a locally penalized adjacent peer.

## Compatibility

No mesh protobuf, QUIC ALPN, gossip schema, Skippy ABI, plugin protocol, or release metadata changes. The status field is additive and the reputation state is process-local.

## Branch integrity

Base branch: `main`  
Validated base: `f9bd75a97378be014f52c4e68c13e81d3399e65e`  
Branch status: `0 behind / 1 ahead` against `origin/main`  
Introduced commit: `9800b62f Add local target reputation routing`

## Validation

Validation mode: Tier 3 - shared OpenAI/mesh routing health behavior plus post-review design/test hardening.

- `git fetch --no-tags origin main:refs/remotes/origin/main`: PASS, `origin/main` at `f9bd75a97378be014f52c4e68c13e81d3399e65e`
- `cargo fmt --all`: PASS
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo test -p mesh-llm-host-runtime affinity --lib -- --test-threads=1`: PASS, 18 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo test -p mesh-llm-host-runtime target_health --lib -- --test-threads=1`: PASS, 13 passed
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo check -p mesh-llm`: PASS
- `LLAMA_STAGE_BUILD_DIR=<stage-build-dir> cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: PASS

## Required remote gates

PASS - required GitHub CI completed on final SHA `9800b62fdd4f4aafa442ebc6022c80683b1a8672`.

## Not run

- Live multi-node reputation/routing smoke: not run because no local multi-node model runtime endpoint was available. Deterministic target-health and affinity-routing simulation tests cover the changed local behavior.
- Full workspace suite locally: not required for this validation mode; mandatory PR CI is the final full-suite proof.

## Rollback

Rollback: revert this PR.  
DB downgrade: not applicable.  
Data repair: not applicable.  

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/affinity.rs
+++ b/crates/mesh-llm-host-runtime/src/network/affinity.rs
@@ -1,7 +1,7 @@
 //! Prefix affinity and sticky routing helpers for inference target selection.
 
 use crate::inference::election;
-use crate::network::target_health::{TargetHealth, TargetHealthOutcome};
+use crate::network::target_health::{TargetHealth, TargetHealthOutcome, TargetReputationStats};
 use iroh::EndpointId;
 use serde::Serialize;
 use serde_json::Value;
@@ -34,6 +34,7 @@ pub struct AffinityStatsSnapshot {
     pub session_routes: u64,
     pub learned: u64,
     pub evicted: u64,
+    pub target_reputation: TargetReputationStats,
 }
 
 fn prefix_only_enabled() -> bool {
@@ -119,6 +120,7 @@ impl AffinityRouter {
         stats.prefix_entries = state.entries.len();
         stats.prefix_enabled = self.config.prefix_enabled;
         stats.sticky_enabled = self.config.sticky_enabled;
+        stats.target_reputation = self.target_health.reputation_stats();
         stats
     }
 
@@ -739,14 +741,45 @@ mod tests {
     use crate::network::target_health::TargetHealthOutcome;
     use iroh::SecretKey;
 
+    const TEST_MODEL: &str = "qwen";
+
     fn make_id(seed: u8) -> EndpointId {
         let mut bytes = [0u8; 32];
         bytes[0] = seed;
         SecretKey::from_bytes(&bytes).public()
     }
 
+    fn remote(seed: u8) -> election::InferenceTarget {
+        election::InferenceTarget::Remote(make_id(seed))
+    }
+
     fn parse_body(body: &str) -> Value {
         serde_json::from_str(body).unwrap()
+    }
+
+    struct SimulatedMeshRouter {
+        affinity: AffinityRouter,
+        hosts: Vec<EndpointId>,
+    }
+
+    impl SimulatedMeshRouter {
+        fn new(host_seeds: &[u8]) -> Self {
+            Self {
+                affinity: AffinityRouter::default(),
+                hosts: host_seeds.iter().map(|seed| make_id(*seed)).collect(),
+            }
+        }
+
+        fn route_order(&self) -> Vec<election::InferenceTarget> {
+            prepare_remote_targets_for_request(TEST_MODEL, &self.hosts, None, &self.affinity)
+                .ordered
+        }
+
+        fn record_peer_outcome(&self, host_index: usize, outcome: TargetHealthOutcome) {
+            let target = election::InferenceTarget::Remote(self.hosts[host_index]);
+            self.affinity
+                .record_target_outcome(Some(TEST_MODEL), &target, outcome);
+        }
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


ndizazzo (commented):
@IvGolovach I think this is going to be a significant enough piece of routing in mesh-llm that it warrants its own doc and careful design. Can you carve out `docs/NODE_REP.md` with the details from the first additive change you have here? 

I also think this can be a really sharp edge for us, so I wouldn't mind building more integration testing in up front. It would be great if we could have some mechanism to simulate a multi-node setup, and set the reputation for an adjacent node and observe how requests flow across the mesh.

Then we can build our implementation out from discussions between @michaelneale @i386 and yourself.

michaelneale:
I kind of like this - it seems fairly lean and minimal. @IvGolovach do you have any rough A/B of using this vs the status quo on public mesh? I think if we can show it improves things in the face of non responsive peers, worth getting it in, as long as it makes it no worse.

michaelneale:
yeah if could add some simulated tests for it would be really nice - I am happy to try this out too if that helps. one more thing - this is all per peer - so each peer keeps its own reputation data on how it sees the world, is that right @IvGolovach ?

IvGolovach:
Yes, this is intentionally local per-peer reputation: each node keeps its own view of how adjacent targets behave from that node's requests and does not gossip or trust another peer's scores.

I added `docs/NODE_REP.md` to make that boundary explicit, linked it from the docs, and added a simulated multi-node router test that sets reputation on adjacent candidates and verifies the request flow changes without needing a live mesh.

I do not have a public-mesh A/B from this branch yet, so I kept the claim narrower in the PR: deterministic local simulation plus the existing affinity/target-health coverage. The live public-mesh comparison is still the right next proof once this is in a deployable build.

michaelneale:
@IvGolovach you could just run this client only against the public mesh with --auto and see how it goes as nodes on other end don't need any change do they? should be able to see how it goes?

IvGolovach:
Yep, you were right: the other public-mesh nodes do not need this branch. I ran a client-only public-mesh smoke from this PR head.

Validation shape:

- Built this branch with `just build`.
- Launched it as a client-only node with `mesh-llm client --auto` on separate local ports.
- Joined the public Nostr mesh successfully.
- `/api/status` showed 13-18 peers and the new `routing_affinity.target_reputation` surface.
- `/v1/models` returned 3 mesh-advertised models.
- A short Qwen3-8B request completed with HTTP 200 in ~6.7s.
- A short MiniMax-M2.5 request completed with HTTP 200 in ~0.8s.

I also hit two Qwen3-8B client-side curl timeouts during the smoke. I would treat that as public-mesh flakiness rather than a clean A/B result: the local client stayed healthy, but those timeouts were client aborts, so they did not become router-observed retryable target failures and did not increment the local reputation penalty counters.

So the branch passes the “client-only can join public mesh and route without upgrading remote peers” smoke. I would not oversell this as a controlled A/B benchmark yet; the next stronger proof would be a branch-vs-main run with a router-observed timeout/unavailable target so we can show the reputation penalty changing candidate ordering live.

michaelneale (approved):
tested it - is nice.

michaelneale:
ok I tested this, seems good - @IvGolovach https://github.com/Mesh-LLM/mesh-llm/pull/772 is a follow up for you to review (which makes it in theory a bit smarter about treating timeouts as a signal) - if that one makes sense please review it there, this one I will merge for now as it is good. Nice one!

PR #772, opened 2026-06-01
Record MoA remote timeouts in target reputation

michaelneale:
🤖 This stacks on #767.

## Summary
- route MoA remote worker host selection through target reputation eligibility
- record MoA remote worker timeout as a local target timeout
- record MoA remote worker success so penalties can recover

## Validation
- cargo test -p mesh-llm-host-runtime moa_remote_host_selection --lib
- cargo check -p mesh-llm-host-runtime
- cargo check -p mesh-llm
- just build
- public mesh smoke: model=mesh returned PATCHED_MOA_OK in 0.719s

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
@@ -692,6 +692,7 @@ async fn try_handle_moa_intercept(
         request,
         decision.effective_model.as_deref(),
         Some(ctx.route.targets),
+        ctx.route.affinity,
     )
     .await;
     proxy::release_request_objects(ctx.route.node, &request.request_object_request_ids).await;
--- a/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/mod.rs
@@ -12,7 +12,9 @@
 
 use crate::inference::election;
 use crate::mesh;
+use crate::network::affinity::AffinityRouter;
 use crate::network::openai::transport as proxy;
+use crate::network::target_health::TargetHealthOutcome;
 use mesh_mixture_of_agents as moa;
 use progress::ProgressContinuation;
 use tokio::io::AsyncWriteExt;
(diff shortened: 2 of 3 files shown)


IvGolovach (changes requested):
Thanks for the focused follow-up here. I like the direction a lot: keeping the reputation signal local/process-bound and extending it into the MoA remote-worker path is the right scope, and it avoids turning this into a gossip/protocol trust change.

The core pieces mostly line up: MoA now receives the shared `AffinityRouter`, remote worker host selection goes through target-health eligibility, remote worker timeouts are recorded as `Timeout`, and successful parsed responses record `Success` so reputation can recover. CI is green, and I also checked the targeted `moa_remote_host_selection`, `target_health`, and `affinity` tests locally.

I do think there is one correctness hole worth fixing before merge, though.

The new selection test covers multiple hosts for the same exact model alias, but MoA worker resolution first groups aliases by canonical model and then tries aliases one by one. In that path, a cooled single-host first alias can still win before the resolver ever tries a healthy second alias from the same canonical group.

Concretely: if peer A advertises `Qwen3-8B-Q4_K_M` and peer B advertises `unsloth/Qwen3-8B-GGUF:Q4_K_M`, both canonicalize to one MoA worker group. If A times out, the next MoA build still tries the shorter alias first. `hosts_for_model(short_alias)` returns only A, and because `select_remote_host_for_moa()` uses preserve-availability eligibility, the single cooling candidate is kept eligible. `add_worker_backend()` then returns true and the loop never reaches B’s healthy alias.

That seems to undercut the main promise of the PR for mixed-alias public-mesh cases.

Could we add a targeted regression test for “cooled first alias, healthy second alias in the same canonical group”, and adjust the resolver so alias fallback gets a chance to prefer a non-cooling route across the canonical group? One clean shape might be to use strict eligibility while probing individual aliases, then only fall back to preserve-availability behavior if no alias in the group has a healthy route.

The timeout/success accounting itself looks good to me after that. Also worth noting, but not necessarily blocking this PR: immediate tunnel/send/recv/parser failures still return errors without recording `Unavailable`; this PR currently records only outer timeout and success outcomes.

michaelneale:
@IvGolovach yeah - I don't think this adds the right stuff, if you want, make a follow on issue and we can work on it fresh again I think. I will close this for now (feel free to pick it up if you want).

PR #782, opened 2026-06-02
Add embedded SDK knobs: disable iroh relays + join-token on running node

tlongwell-block:
## What

Two opt-in, default-false seams the Sprout v1 mesh integration needs on the embedded SDK:

- **`disable_iroh_relays(bool)`** — when true, `bind_mesh_endpoint` uses `RelayMode::Disabled` instead of the existing `RelayMode::Custom` path. Lets a consumer say "no public relay" explicitly; today an empty relay list silently falls back to public `*.iroh.link`. Default `false` preserves existing behavior byte-for-byte.
- **`EmbeddedNodeHandle::join_token(token)`** — forwards an invite token over the existing `RuntimeControl` channel to `node.join_with_retry`, so an already-running embedded node can dial a new `EndpointAddr` without restart (the call-me-now primitive).

Purely additive. Defaults preserve current behavior for all other consumers. +87/−9 across 6 files.

## Base

This PR targets **`micn/sprout-embedded-serve-sdk`** (the branch behind #736), not `main`. The knobs extend the `crates/mesh-llm-sdk` crate, which only exists on the #736 stack — it is not on `main` yet. A PR against `main` would be unreviewable noise (the whole SDK stack would show as the diff). This is a dependent PR; it should land after / with #736.

## Verification

`cargo check -p mesh-llm-sdk --features client,serve` green. Consumed and exercised in Sprout's desktop mesh integration (relay-gated discovery → connect → call-me-now → inference).

Co-authored with Max (Sprout team).

## Connectivity: WAN via STUN, relay transport off

`disable_iroh_relays(true)` → `RelayPolicy::ExplicitlyDisabled` turns off the iroh **relay transport** (no `*.iroh.link` traffic) but **keeps raw STUN on** (`uses_raw_stun()` matches `DefaultPublic | ExplicitlyDisabled`). The STUN-discovered public address is injected into the invite token / `EndpointAddr`, so a relay-signaled peer can hole-punch directly over **WAN** without an iroh relay. STUN is a "what's my public IP" lookup, not a data path — the privacy posture (no relay traffic) holds.

`Disabled` (LAN-only mDNS mode) stays STUN-off, intentionally.

Residual limit: no relay **transport fallback**, so two peers both behind **symmetric NATs** may fail to hole-punch (the case iroh relays normally cover). Fine for the common cases (≥1 side cone-NAT/forwarded/server). (Reviewed by Perci.)

### DIFF
--- a/crates/mesh-llm-host-runtime/src/api/state.rs
+++ b/crates/mesh-llm-host-runtime/src/api/state.rs
@@ -43,6 +43,10 @@ impl Serialize for PublicationState {
 }
 
 pub enum RuntimeControlRequest {
+    Join {
+        invite_token: String,
+        resp: tokio::sync::oneshot::Sender<anyhow::Result<()>>,
+    },
     Load {
         spec: String,
         resp: tokio::sync::oneshot::Sender<anyhow::Result<RuntimeLoadResponse>>,
--- a/crates/mesh-llm-host-runtime/src/cli/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/cli/mod.rs
@@ -536,6 +536,10 @@ pub(crate) struct Cli {
     #[arg(long = "relay-auth", value_parser = parse_relay_auth_pair, hide = true)]
     pub(crate) relay_auth: Vec<(String, String)>,
 
+    /// Disable iroh relay registration. Direct addresses and explicit EndpointAddr dials still work.
+    #[arg(long, hide = true)]
+    pub(crate) disable_iroh_relays: bool,
+
     /// Bind QUIC to a fixed UDP port (for NAT port forwarding).
     #[arg(long, hide = true)]
     pub(crate) bind_port: Option<u16>,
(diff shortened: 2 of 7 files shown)


ndizazzo (changes requested):
Hey @tlongwell-block 👋🏻 good to see you here, and thanks for the PR!

I had a look at the PR and it looks sane. There's been some back and forth in Discord on some node discoverability issues, so I was looking @ this from the critical lens of mDNS and local-only hosting, and spotted a potential issue...

`ExplicitlyDisabled.uses_raw_stun()` -> true means that mDNS/LAN-only mode would let mesh do STUN when we'd want that disabled. The smallest fix could be to keep mDNS/LAN precedence over `disable_iroh_relays`, since you'd not be enabling LAN for Sprout anyways. Mind adding a case to cover that edge and a switch over the relay policy order?

michaelneale:
@ndizazzo I think this may be ok

PR #790, opened 2026-06-04
Reduce Skippy staged decode latency toward 30 tok/s

i386:
## Summary

This PR moves Skippy's staged serving path closer to the 30 tok/s target by removing serialized decode-loop network waits, documenting tip-to-tip placement, and exposing exact-prefix cache savings in telemetry.

The user-facing result is easier to reason about: stage chains now have a direct prediction return path, topology planning can place the first and final logical stages on the same physical peer, and repeated-prefix requests report how many activation bytes were avoided by resident cache hits.

## Protocol

This is a breaking Skippy stage-protocol change. Generation-3 stage peers must support direct prediction return; older chained-reply stages should not be mixed into a generation-3 split topology.

Before, prediction-bearing decode replies walked back through every intermediate stage:

```mermaid
sequenceDiagram
    participant S0 as Stage 0 / driver
    participant S1 as Stage 1
    participant S2 as Stage 2
    participant S3 as Stage 3 / readout

    S0->>S1: DecodeEmbd activation
    S1->>S2: DecodeEmbd activation
    S2->>S3: DecodeEmbd activation
    S3-->>S2: PredictedToken
    S2-->>S1: PredictedToken
    S1-->>S0: PredictedToken
```

After, prediction-bearing replies return directly from the final/readout tip to the driver-facing stage. Intermediate stages can still send cold-path ACK/stats/control replies, but they are no longer on the per-token prediction return path:

```mermaid
sequenceDiagram
    participant S0 as Stage 0 / driver
    participant S1 as Stage 1
    participant S2 as Stage 2
    participant S3 as Stage 3 / readout

    S0->>S1: DecodeEmbd activation
    S1->>S2: DecodeEmbd activation
    S2->>S3: DecodeEmbd activation
    S3-->>S0: PredictedToken direct return
    S3-.->>S2: ACK/stats/control, cold path
    S2-.->>S1: ACK/stats/control, cold path
    S1-.->>S0: ACK/stats/control, cold path
```

Tip-to-tip placement then colocates stage 0 and the final/readout logical stage on the same physical peer. The stage graph still has four logical stages, but the no-spec decode loop serializes only three physical network hops:

```mermaid
flowchart LR
    Client["OpenAI client"] --> Mesh["mesh/openai-frontend"]
    Mesh --> A0["host A<br/>stage 0<br/>driver tip"]
    A0 -->|"activation"| B["host B<br/>stage 1"]
    B -->|"activation"| C["host C<br/>stage 2"]
    C -->|"activation"| AF["host A<br/>final/readout tip"]
    AF -->|"PredictedToken<br/>local return"| A0
    A0 --> Mesh
    Mesh --> Client
```

## Exact Prefix Cache

The cache is resident inside each `skippy-server` stage process. It is not a central cache service and it is not lossy activation compression. Each stage records exact prefix state keyed by model/topology/stage identity, layer range, ABI/layout/dtype, token range, and token IDs. For transformer-only families this is effectively resident KV state; for recurrent/hybrid families the exact-state path can include recurrent state as well.

On a repeated-prefix request, stage 0 attempts an all-stage chain restore before normal prefill. If any downstream stage misses, Skippy drops the partial restore and falls back to normal prefill rather than mixing restored and cold state.

```mermaid
sequenceDiagram
    participant OAI as OpenAI frontend / stage 0
    participant C0 as Stage 0 resident cache
    participant S1 as Stage 1
    participant C1 as Stage 1 resident cache
    participant S2 as Stage 2
    participant C2 as Stage 2 resident cache
    participant SF as Final stage
    participant CF as Final resident cache

    OAI->>C0: Lookup exact prefix identity
    C0-->>OAI: hit
    OAI->>S1: TryRestorePrefill(prefix token sideband)
    S1->>C1: Lookup exact prefix identity
    C1-->>S1: hit
    S1->>S2: TryRestorePrefill(prefix token sideband)
    S2->>C2: Lookup exact prefix identity
    C2-->>S2: hit
    S2->>SF: TryRestorePrefill(prefix token sideband)
    SF->>CF: Lookup exact prefix identity
    CF-->>SF: hit
    SF-->>S2: ACK + cache stats
    S2-->>S1: ACK + cache stats
    S1-->>OAI: ACK + cache stats
    OAI->>S1: Prefill only uncached suffix, or fused first decode
```

This PR adds telemetry to make those cache hits measurable:

- `skippy.kv.chain_cache_hit_stage_count`
- `skippy.kv.chain_cache_stage0_activation_bytes_avoided`
- `skippy.kv.chain_cache_interstage_activation_bytes_avoided_estimate`

## Performance Estimate

Assumption used below: four logical stages with 10 ms one-way delay per serialized stage-to-stage network hop, no speculative decoding, and no compute included unless stated. These are latency floors, not guaranteed end-to-end throughput.

| Path | Serialized network hops per decoded token | Network-only TPOT floor | Network-only tok/s ceiling | Impact |
| --- | ---: | ---: | ---: | --- |
| Old chained reply | 6 | ~60 ms | ~16.7 tok/s | Baseline before direct return |
| Direct return | 4 | ~40 ms | ~25.0 tok/s | Removes two return-chain waits |
| Tip-to-tip | 3 | ~30 ms | ~33.3 tok/s | Removes the final network return wait |

TPOT: direct return should improve no-spec decode TPOT by about 20 ms/token versus chained replies under the 10 ms-hop assumption. Tip-to-tip should improve it by another 10 ms/token versus direct return. Real TPOT is `network floor + stage compute + encode/write/queueing`; reaching 30 tok/s with a 30 ms network floor leaves only about 3.3 ms/token for everything else, so decode-loop overlap and compute-side work still matter.

FTTT: cold first-token total time still includes prefill plus the first decode. The resident exact-prefix cache can remove repeated-prefix prefill compute and activation transfer. For a 256-token cached prefix at activation width 5120:

| Wire dtype | Avoided stage0->stage1 activation bytes | Four-stage interstage estimate |
| --- | ---: | ---: |
| f32 | 5.00 MiB | 15.00 MiB |
| f16 | 2.50 MiB | 7.50 MiB |
| q8 | 1.25 MiB | 3.75 MiB |

At 1 Gbps, the f16 256-token example avoids roughly 63 ms of payload transmission across three inter-stage boundaries, before counting RTT, framing, scheduler waits, and per-stage prefill compute. A 2k-token repeated prefix scales to roughly 60 MiB avoided at f16 across three boundaries, or about 0.5 s of payload time at 1 Gbps.

Effective tok/s: the cache mostly improves request-level effective throughput for short/medium completions by reducing FTTT; it does not materially change steady-state decode TPOT after the first token. Example with 33 ms steady TPOT:

| Completion length | Cold FTTT 800 ms | Cached FTTT 150 ms | Effective tok/s gain |
| ---: | ---: | ---: | ---: |
| 16 tokens | ~12.4 tok/s | ~24.8 tok/s | ~2.0x |
| 128 tokens | ~25.7 tok/s | ~29.5 tok/s | ~1.15x |

## Validation

- `just with-lld cargo check -p skippy-server`
- `just build`


i386:
Superseded by prefix-cache-only PR #791. This stacked PR included unrelated model-fit, direct-return, and tip-to-tip history; #791 is scoped to resident exact-prefix cache telemetry only.

PR #791, opened 2026-06-04
Skip first-token work for repeated Skippy prompts

i386:
## Summary

Repeated exact prompts can now skip the expensive first-token path across a Skippy split chain.

The motivating case is agent and tool-loop traffic: the same chat prefix, or sometimes the exact same prompt, can be sent again while the model only needs a short continuation. Before this PR, Skippy could restore a cached prefix, but an exact repeated prompt still had to run the first decode through every stage before the client saw token 1. On a staged model, that first token pays both compute and the serialized stage-to-stage activation trip.

This PR adds an exact, memory-resident full-prompt first-token cache. After the first request computes token 1, every stage records the full prompt state. Stage 0 also records the predicted first generated token for that exact prompt identity. On an exact repeat, stage 0 restores the full prompt across the chain and emits cached token 1 immediately, then normal decode continues from token 2.

This is intentionally narrow: exact prompt identity only, memory resident only, and no lossy activation compression or central cache service.

## What the Data Shows

I ran a category-shaped corpus built from the local `skippy-bench` prompt sets. The benchmark used a local 3-stage `Qwen3-0.6B-Q4_K_M` split chain, split `0..9`, `9..18`, `18..28`, f16 activation wire, streaming chat, `max_tokens=4`, and temperature `0`.

Raw timing output: `/tmp/prefix-cache-category-bench-20260604-175250/report.json`.
Telemetry proof run: `/tmp/prefix-cache-category-telemetry-20260604-175330/stage0.log`.

| skippy-bench category | Workload story | Prompt tokens | Cold FTTT/TTFT ms | Warm/shared median ms | Delta ms | Speedup | Total elapsed delta ms | Cache path |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| `exact_repeat` | Same prompt repeated 4x | 88 | 312.8 | 17.2 | 295.6 | 18.2x | 298.0 | `chain_full_prompt_first_token` |
| `prefix_cache_exact_repeat` | Longer `prefix_cache` prompt repeated 3x | 185 | 1542.5 | 18.7 | 1523.8 | 82.6x | 1524.7 | `chain_full_prompt_first_token` |
| `repeated_prefix_alpha` | Shared prefix, different user tail | 109 | 597.6 | 72.6 | 525.0 | 8.2x | 522.4 | `chain_prefix` |
| `repeated_prefix_beta` | Shared prefix, different user tail | 79 | 321.8 | 69.4 | 252.4 | 4.6x | 250.0 | `chain_prefix` |
| `control_sanity` | Unrelated one-shot control | 14 | 47.4 | n/a | n/a | n/a | n/a | no expected reuse |
| `control_qa` | Unrelated one-shot control | 38 | 237.7 | n/a | n/a | n/a | n/a | no expected reuse |
| `control_summarization` | Unrelated one-shot control | 39 | 238.2 | n/a | n/a | n/a | n/a | no expected reuse |
| `control_debugging` | Unrelated one-shot control | 52 | 230.2 | n/a | n/a | n/a | n/a | no expected reuse |

The important rows are the first two. They are exact repeats and telemetry confirms they used the new `chain_full_prompt_first_token` path. The 88-token exact repeat moved from `312.8 ms` cold FTTT to `17.2 ms` warm median FTTT. The longer 185-token repeat moved from `1542.5 ms` to `18.7 ms`.

The shared-prefix rows are included to orient reviewers rather than to claim this PR invented that behavior. They still use the existing `chain_prefix` path, and they show that the broader cache machinery remains useful when prompts share a prefix but have different tails. The control rows are unrelated one-shot prompts, so they intentionally have no warm comparison.

## Why This Helps Tokens Per Second

For long generations, steady TPOT after token 1 is mostly unchanged because token 2+ still uses the normal decode loop. The win is FTTT and effective tok/s for short repeated requests, which is exactly where interactive agents and tool loops tend to feel slow.

With a 4-stage chain and 10 ms one-way latency between stages, skipping the first decode avoids roughly one serialized stage-chain trip, about `30 ms` of network latency before compute is counted. The payload avoided is small, so this is not primarily a bandwidth win:

| Wire dtype | One activation frame at width 5120 | 3 boundaries | Payload time at 1 Gbps |
| --- | ---: | ---: | ---: |
| f32 | ~20 KiB | ~60 KiB | <1 ms |
| f16 | ~10 KiB | ~30 KiB | <1 ms |
| q8 | ~5 KiB | ~15 KiB | <1 ms |

The practical win is removing first-decode compute plus the stage-chain latency. That changes effective tok/s most for short completions:

| Completion length | Before: 250 ms FTTT, 33 ms TPOT | After: 180 ms FTTT, 33 ms TPOT | Effective tok/s change |
| ---: | ---: | ---: | ---: |
| 8 tokens | ~15.6 tok/s | ~19.3 tok/s | +24% |
| 16 tokens | ~20.6 tok/s | ~24.0 tok/s | +17% |
| 64 tokens | ~27.1 tok/s | ~28.0 tok/s | +3% |

These estimates are deliberately simple and not benchmark-fitted constants. The measured local run proves the new cache path and gives a lower-scale signal; larger staged models should benefit more when first decode compute is larger.

## Behavior Before

Before this PR, an exact repeated prompt could restore the prefix before the final prompt token, but it still had to run first decode through every stage to produce token 1.

```mermaid
sequenceDiagram
    participant C as Client
    participant OAI as OpenAI frontend / Stage 0
    participant C0 as Stage 0 cache
    participant S1 as Stage 1
    participant C1 as Stage 1 cache
    participant S2 as Stage 2
    participant C2 as Stage 2 cache
    participant SF as Final stage
    participant CF as Final cache

    C->>OAI: Repeat exact prompt
    OAI->>C0: Restore prefix tokens 0..N-2
    C0-->>OAI: hit
    OAI->>S1: TryRestorePrefill(prefix tokens)
    S1->>C1: Restore prefix tokens 0..N-2
    C1-->>S1: hit
    S1->>S2: TryRestorePrefill(prefix tokens)
    S2->>C2: Restore prefix tokens 0..N-2
    C2-->>S2: hit
    S2->>SF: TryRestorePrefill(prefix tokens)
    SF->>CF: Restore prefix tokens 0..N-2
    CF-->>SF: hit
    SF-->>S2: ACK + cache stats
    S2-->>S1: ACK + cache stats
    S1-->>OAI: ACK + cache stats
    OAI->>S1: DecodeEmbd(final prompt token)
    S1->>S2: DecodeEmbd(activation)
    S2->>SF: DecodeEmbd(activation)
    SF-->>S2: PredictedToken(token 1)
    S2-->>S1: PredictedToken(token 1)
    S1-->>OAI: PredictedToken(token 1)
    OAI-->>C: token 1
```

## Behavior After

The first request records a replayable full-prompt state after first decode succeeds.

```mermaid
sequenceDiagram
    participant OAI as OpenAI frontend / Stage 0
    participant S1 as Stage 1
    participant S2 as Stage 2
    participant SF as Final stage
    participant K0 as Stage 0 cache
    participant K1 as Stage 1 cache
    participant K2 as Stage 2 cache
    participant KF as Final cache

    OAI->>S1: DecodeEmbd(final prompt token, full prompt sideband)
    S1->>S2: DecodeEmbd(activation, full prompt sideband)
    S2->>SF: DecodeEmbd(activation, full prompt sideband)
    SF-->>S2: PredictedToken(token 1)
    SF->>KF: Record full prompt state
    S2->>K2: Record full prompt state
    S2-->>S1: PredictedToken(token 1)
    S1->>K1: Record full prompt state
    S1-->>OAI: PredictedToken(token 1)
    OAI->>K0: Record full prompt state + token 1
```

On an exact repeat, every stage restores the full prompt directly and stage 0 emits token 1 from the cache.

```mermaid
sequenceDiagram
    participant C as Client
    participant OAI as OpenAI frontend / Stage 0
    participant K0 as Stage 0 cache
    participant S1 as Stage 1
    participant K1 as Stage 1 cache
    participant S2 as Stage 2
    participant K2 as Stage 2 cache
    participant SF as Final stage
    participant KF as Final cache

    C->>OAI: Repeat exact prompt
    OAI->>K0: Lookup full prompt identity + token 1
    K0-->>OAI: hit(token 1)
    OAI->>K0: Restore full prompt state
    K0-->>OAI: hit
    OAI->>S1: TryRestorePrefill(full prompt tokens)
    S1->>K1: Restore full prompt state
    K1-->>S1: hit
    S1->>S2: TryRestorePrefill(full prompt tokens)
    S2->>K2: Restore full prompt state
    K2-->>S2: hit
    S2->>SF: TryRestorePrefill(full prompt tokens)
    SF->>KF: Restore full prompt state
    KF-->>SF: hit
    SF-->>S2: ACK + cache stats
    S2-->>S1: ACK + cache stats
    S1-->>OAI: ACK + cache stats
    OAI-->>C: cached token 1
    OAI->>S1: DecodeEmbd(token 1) for token 2+
```

If any downstream stage misses or errors during the full-prompt restore, Skippy drops the partial restore and falls back to the existing prefix/first-decode path.

## Implementation Notes

- Adds a stage-0 first-token index keyed by the exact full-prompt cache identity.
- Records exact full-prompt state on stage 0 after first decode, including exact-state payloads where enabled.
- Propagates full-prompt sideband on first cold `DecodeEmbd` so downstream stages can record the same full-prompt state.
- Records full-prompt state in `TryRestorePrefillDecode` too, covering the existing fused prefix-restore + first-decode path.
- Keeps the existing prefix restore fallback: a full-prompt first-token cache hit must restore across all stages before it can emit cached token 1.
- Adds telemetry so cache decisions can be verified as `chain_full_prompt_first_token`, `chain_prefix`, or miss/control behavior.

## Protocol And Compatibility

This changes Skippy stage protocol semantics for first decode messages.

`DecodeEmbd` may now carry full prompt tokens in `message.tokens` as sideband metadata when `token_count == 1`, `decode_step == 0`, and `state.current_token` is the final prompt token. Updated stages execute decode from `state.current_token`; they do not decode `message.tokens[0]` when this sideband is present. The sideband is used to record the full-prompt cache identity after the decode succeeds.

Compatibility impact:

- Mesh gossip and public mesh protocol are unchanged.
- Skippy stage message IDs are unchanged.
- Mixed old/new Skippy stage binaries are not safe for this semantic change: older stages may treat the first sideband token as the decode token. Deploy the stage chain with updated binaries together.
- Sideband recording is bounded by `MAX_STAGE_SIDEBAND_VALUES`; prompts above that limit keep the old one-token decode sideband and do not record the full-prompt first-token entry.

## Validation

- `rustfmt --edition 2024 --check crates/skippy-server/src/kv_integration/mod.rs crates/skippy-server/src/kv_integration/config.rs crates/skippy-server/src/frontend.rs crates/skippy-server/src/frontend/prefix_cache.rs crates/skippy-server/src/frontend/embedded_generation.rs crates/skippy-server/src/binary_transport.rs crates/skippy-server/src/binary_transport/tests.rs`
- `git diff --check`
- `just with-lld cargo check -p skippy-server`
- `just build`

### DIFF
--- a/crates/skippy-server/src/binary_transport.rs
+++ b/crates/skippy-server/src/binary_transport.rs
@@ -978,6 +978,21 @@ fn handle_binary_connection(
             }
         }
 
+        if let Some(full_prompt_tokens) = decode_full_prompt_sideband(&message) {
+            let mut runtime = runtime.lock().expect("runtime lock poisoned");
+            let record = maybe_record_binary_full_prefill(
+                config,
+                &mut runtime,
+                kv,
+                telemetry,
+                &session_key,
+                &message,
+                full_prompt_tokens,
+            );
+            drop(runtime);
+            add_binary_record_stats(&mut message_reply_stats, config, &record);
+        }
+
         let mut forward_write_ms = 0.0;
         let mut forward_activation_encode_ms = 0.0;
         let mut downstream_wait_ms = 0.0;
@@ -2098,6 +2113,19 @@ fn handle_binary_restore_prefill_decode_control(
                 .context("restore-decode downstream miss ACK")?;
             return Ok(());
         }
+        {
+            let mut runtime = runtime.lock().expect("runtime lock poisoned");
+            let record = maybe_record_binary_full_prefill(
+                config,
+                &mut runtime,
+                kv,
+                telemetry,
+                session_id,
+                message,
+                message.tokens.as_slice(),
+            );
+            add_binary_record_stats(&mut control_stats, config, &record);
+        }
         let mut attrs = binary_message_attrs(config, wire_session_id, message);
         attrs.insert("skippy.kv.control_hit".to_string(), json!(true));
         attrs.insert(
@@ -2127,6 +2155,19 @@ fn handle_binary_restore_prefill_decode_control(
         return Ok(());
     }
 
+    {
+        let mut runtime = runtime.lock().expect("runtime lock poisoned");
+        let record = maybe_record_binary_full_prefill(
+            config,
+            &mut runtime,
+            kv,
+            telemetry,
+            session_id,
+            message,
+            message.tokens.as_slice(),
+        );
+        add_binary_record_stats(&mut control_stats, config, &record);
+    }
     let mut attrs = binary_message_attrs(config, wire_session_id, message);
     attrs.insert("skippy.kv.control_hit".to_string(), json!(true));
     attrs.insert(
@@ -3303,6 +3344,9 @@ fn token_sideband_or_fill(message: &StageWireMessage) -> Result<Vec<i32>> {
         .token_count
         .try_into()
         .context("negative token_count")?;
+    if let Some(token) = decode_execution_token(message, token_count) {
+        return Ok(vec![token]);
+    }
     if message.tokens.len() == token_count {
         return Ok(message.tokens.clone());
     }
@@ -3315,6 +3359,55 @@ fn token_sideband_or_fill(message: &StageWireMessage) -> Result<Vec<i32>> {
         0
     };
     Ok(vec![fill; token_count])
+}
+
+fn decode_execution_token(message: &StageWireMessage, token_count: usize) -> Option<i32> {
+    if !matches!(
+        message.kind,
+        WireMessageKind::DecodeEmbd
+            | WireMessageKind::DecodeReadout
+            | WireMessageKind::DecodeLightCtx
+            | WireMessageKind::DecodeReplayEmbd
+            | WireMessageKind::DecodeReplayFinalEmbd
+    ) || token_count != 1
+        || message.state.current_token == skippy_protocol::binary::LLAMA_TOKEN_NULL
+    {
+        return None;
+    }
+    Some(message.state.current_token)
+}
+
+fn decode_full_prompt_sideband(message: &StageWireMessage) -> Option<&[i32]> {
+    if message.kind != WireMessageKind::DecodeEmbd
+        || message.token_count != 1
+        || message.state.decode_step != 0
+        || message.state.prompt_token_count <= 0
+    {
+        return None;
+    }
+    let prompt_token_count = usize::try_from(message.state.prompt_token_count).ok()?;
+    if message.tokens.len() != prompt_token_count
+        || message.tokens.last().copied() != Some(message.state.current_token)
+    {
+        return None;
+    }
+    Some(message.tokens.as_slice())
+}
+
+fn add_binary_record_stats(
+    reply_stats: &mut StageReplyStats,
+    config: &StageConfig,
+    record: &BinaryKvRecordResult,
+) {
+    if record.recorded_pages > 0 {
+        reply_stats.kv_recorded_pages += record.recorded_pages as i64;
+        reply_stats.kv_record_stage_mask |= stage_mask(config.stage_index);
+    }
+    if record.recorded_activations > 0 {
+        reply_stats.kv_recorded_bytes = reply_stats
+            .kv_recorded_bytes
+            .saturating_add(record.recorded_activation_bytes as i64);
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


michaelneale (approved):
Woah

PR #799, opened 2026-06-05
Reuse Skippy decode wire messages (2/3)

i386:
## Summary
Skippy split decode now reuses the per-token decode wire-message envelope in the frontend hot path instead of rebuilding the same `StageWireMessage` shape every token.

This is stacked on #798 and is intentionally scoped to allocation churn around decode message construction. It does not change topology, protocol, sampling, activation dtype, or native stage execution.

## What changed
- Added `ReusableDecodeMessage`, a small frontend wire-message helper that owns the stable decode envelope once.
- Normal split decode mutates only `decode_step`, `pos_start`, `current_token`, and the one-token sideband each iteration.
- Multimodal/split decode reuses the same message and token buffer, including the short exact-replay sideband checkpoint case.
- Kept the existing one-shot `embedded_decode_message` helper for prefix-cache and repair paths.
- Added a focused unit test for reusable decode message mutation and stable request/session/sampling fields.

## Before
```mermaid
flowchart LR
    A["Each decode token"] --> B["Allocate StageStateHeader"]
    B --> C["Clone sampling config"]
    C --> D["Allocate tokens Vec"]
    D --> E["Allocate empty positions/activation/raw Vecs"]
    E --> F["Run stage + forward activation"]
```

## After
```mermaid
flowchart LR
    A["Before decode loop"] --> B["Create reusable DecodeEmbd message"]
    B --> C["Each decode token"]
    C --> D["Mutate step/position/current token"]
    D --> E["Refill existing token sideband buffer"]
    E --> F["Run stage + forward activation"]
```

## Performance Impact
This targets small but repeated CPU-side work in TPOT:

- Removes per-token allocation of the decode message's stable empty vectors.
- Avoids per-token sampling clone in the two split decode loops.
- Reuses the sideband token buffer for exact-replay checkpoint frames instead of allocating a fresh `Vec` every eligible token.

This should be a modest fixed-overhead improvement rather than a model-compute improvement. It is designed to compose with #798 and remain easy to benchmark independently once lab capacity is available.

## Compatibility
No protocol or ABI changes. The emitted decode wire messages keep the same fields and token sideband semantics.

## Validation
- `cargo fmt -p skippy-server -- --check`
- `cargo test -p skippy-server --lib` — 116 passed
- `cargo check -p mesh-llm`

### DIFF
--- a/crates/skippy-server/src/frontend/embedded_generation.rs
+++ b/crates/skippy-server/src/frontend/embedded_generation.rs
@@ -599,6 +599,17 @@ impl StageOpenAiBackend {
                 .expect("checked non-empty prompt");
             let mut context_tokens = request.prompt_token_ids.to_vec();
             let mut exact_replay_tokens = Vec::new();
+            let mut decode_message = ReusableDecodeMessage::new(
+                request.wire_dtype,
+                ReusableDecodeMessageArgs {
+                    request_id,
+                    session_id,
+                    prompt_token_count: request.prompt_token_ids.len(),
+                    base_pos_start: prefill_token_count,
+                    sampling: wire_sampling.clone(),
+                    sideband_capacity: skippy_protocol::binary::MAX_STAGE_SIDEBAND_VALUES,
+                },
+            )?;
             let mut fused_reached_stop = false;
             if let Some(fused) = fused_first_decode.take() {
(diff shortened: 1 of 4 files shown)


ndizazzo (approved):
timberrrrr

PR #800, opened 2026-06-05
Reuse Skippy forwarded decode frames (1/3)

i386:
## Summary
Skippy split decode now reuses the forwarded activation frame envelope and activation encode buffer across decode tokens.

This is stacked on #799. It keeps the existing writer and wire format unchanged, but removes repeated forwarded `StageWireMessage` construction and gives activation encoding a caller-owned buffer to refill.

## What changed
- Added an internal in-place activation encoder: `encode_f32_activation_payload_with_state_flags_into(...)`.
- Kept the existing `Vec`-returning activation encode API by delegating to the in-place helper.
- Added `ReusableForwardedStageMessage` for the decode forwarding hot path.
- Reused forwarded decode frame containers in normal split decode and multimodal split decode.
- Preserved existing one-shot `forwarded_stage_message_timed(...)` for non-loop callsites.

## Before
```mermaid
flowchart LR
    A["Native decode output"] --> B["Encode activation into new Vec"]
    B --> C["Build new forwarded StageWireMessage"]
    C --> D["Clone sampling/tokens/positions"]
    D --> E["write_stage_message_conditioned"]
```

## After
```mermaid
flowchart LR
    A["Before decode loop"] --> B["Create reusable forwarded frame"]
    B --> C["Native decode output"]
    C --> D["Refill activation encode buffer"]
    D --> E["Update forwarded frame fields"]
    E --> F["write_stage_message_conditioned"]
```

## Performance Impact
This targets fixed CPU/allocation overhead on decode TPOT:

- Reuses the activation encode buffer instead of allocating a fresh activation `Vec` every forwarded decode token.
- Reuses token/position/raw-byte containers on the forwarded message.
- Avoids repeatedly cloning stable sampling config in the two split decode loops unless it actually changes.

Expected impact is still modest because GPU forward, network wait, activation wire bytes, and downstream compute remain unchanged. This should be more meaningful than #799 when activation encode allocation shows up in profiles, but it is still a cleanup-class improvement rather than the main 30 tok/s lever.

## High-impact follow-up
The bigger decode lever remains overlap/pipelining: hide stage0 setup, sampler/direct-return handling, or downstream wait behind work already in flight. That can remove or hide milliseconds of TPOT; this PR removes repeated local allocation work.

## Compatibility
No wire protocol, ABI, topology, sampling, or activation dtype changes. The emitted stage messages keep the same fields and activation bytes.

## Validation
- `cargo fmt -p skippy-protocol -p skippy-server -- --check`
- `cargo test -p skippy-protocol --lib` — 34 passed
- `cargo test -p skippy-server --lib` — 117 passed
- `cargo check -p mesh-llm`

### DIFF
--- a/crates/skippy-protocol/src/binary/activation.rs
+++ b/crates/skippy-protocol/src/binary/activation.rs
@@ -73,6 +73,26 @@ pub fn encode_f32_activation_payload_with_state_flags(
     f32_payload: &[u8],
     state_flag_bits: i32,
 ) -> io::Result<Vec<u8>> {
+    let mut out = Vec::new();
+    encode_f32_activation_payload_with_state_flags_into(
+        dtype,
+        token_count,
+        n_embd,
+        f32_payload,
+        state_flag_bits,
+        &mut out,
+    )?;
+    Ok(out)
+}
+
+pub fn encode_f32_activation_payload_with_state_flags_into(
+    dtype: WireActivationDType,
+    token_count: i32,
+    n_embd: i32,
+    f32_payload: &[u8],
+    state_flag_bits: i32,
+    out: &mut Vec<u8>,
+) -> io::Result<()> {
     let expected_f32_bytes = activation_wire_bytes_with_state_flags(
         WireActivationDType::F32,
         token_count,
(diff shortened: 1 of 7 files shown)


ndizazzo:
Tipping the stack

PR #802, opened 2026-06-05
Improve MoA reliability for agent tool loops

michaelneale:
🤖 Improves MoA behavior for agentic tool loops by keeping tool selection tighter and bounding large tool-result context before reducer synthesis.

What changed:
- Preserve exact small tool results for follow-up recall.
- Compact large JSON/text tool results into bounded evidence before reducer calls.
- Prefer explicitly requested tools, so prompts like "use exec" do not broaden into unrelated web tools.
- Add deterministic repair for short tool facts that reducers omit.

Why:
Large tool outputs, especially GitHub/API JSON, could make the post-tool reducer path slow or time out even after the tool call succeeded.

Validation:
- cargo test -p mesh-mixture-of-agents --lib
- cargo fmt --all -- --check
- MESH_LLM_DYNAMIC_NATIVE_RUNTIME=0 just release-build
- Deployed static build to mini and validated OpenClaw via mesh/mesh with a GitHub exec workflow and same-session follow-up recall.

### DIFF
--- a/crates/mesh-mixture-of-agents/src/arbiter.rs
+++ b/crates/mesh-mixture-of-agents/src/arbiter.rs
@@ -237,10 +237,12 @@ pub fn try_early_decision(
     // Two workers saying "Paris" and "Berlin" must not be treated as
     // consensus. Find the largest cluster of content-similar answers
     // and only early-exit if it's ≥2 workers AND a majority of answers.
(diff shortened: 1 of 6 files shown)


michaelneale:
small tested moa fixes, isolated.

PR #814, opened 2026-06-10
Add transport-aware Skippy stage ordering

i386:
## Summary

Teach the Skippy topology planner to consider directed pipeline edge cost when ordering selected stages.

This is the transport-specific slice from the MLX comparison work. It does not change runtime networking, protocol bytes, or artifact planning. It only adds an optional planner input and a new package-aware planning entry point.

## Before

Package-aware planning could score nodes by capacity, cache locality, missing artifact cost, and placement signals. That answers “which nodes should participate?”

For a staged pipeline, that is not enough. Decode and prefill activations move stage-to-stage, so the ordered edges matter:

```mermaid
flowchart LR
  A["stage 0: node A"] -->|"slow"| B["stage 1: node B"]
  B -->|"slow"| C["stage 2: node C"]
  C -->|"fast"| D["stage 3: node D"]
```

The same node set can be valid but unnecessarily slow if adjacent stages are ordered across expensive links.

## How It Works

This PR adds `StageEdgeSignal`:

- `source_node_id`
- `target_node_id`
- optional directed `rtt_ms`
- optional `large_frame_bytes_per_sec`
- `direct_prediction_return_supported`

`plan_package_aware_contiguous_with_transport(...)` first applies the existing package-aware node ordering, then reorders the same candidate nodes by adjacent directed edge cost.

For small stage counts, it checks all permutations. For larger plans, it falls back to a greedy ordering so planning stays bounded.

```mermaid
flowchart LR
  A["stage 0: node A"] -->|"2 ms"| C["stage 1: node C"]
  C -->|"3 ms"| B["stage 2: node B"]
  B -->|"4 ms"| D["stage 3: node D"]
```

The accepted plan emits diagnostics for measured stage edges:

```text
[info] pipeline edge node-a -> node-c after stage-0 has measured rtt 5 ms
```

## Why This Is Good

Skippy split serving is chain-latency sensitive. A single slow adjacent stage edge can dominate token latency even when each node is individually capable.

MLX pipeline mode gets this discipline naturally from rank-neighbor communication. This PR brings the same idea to Skippy without copying MLX’s runtime: the topology planner can now prefer the cheapest activation chain when the selected nodes are otherwise equivalent.

## Scope

- Adds planner-only `StageEdgeSignal`.
- Adds `plan_package_aware_contiguous_with_transport(...)`.
- Keeps the existing `plan_package_aware_contiguous_with_signals(...)` API as a compatibility wrapper.
- Does not infer true peer-to-peer edge costs from coordinator-to-peer RTT.
- Does not change wire protocol or server transport.

## Validation

## Summary by CodeRabbit

* **New Features**
  * Planner orders pipeline stages using measured per-edge network latency (RTT) when available.
  * New API accepts transport/edge signals so planning can prefer lower-latency stage transitions.
  * Ties in edge cost preserve original package order deterministically.
  * Plans include edge-related diagnostics showing per-transition network cost details.

* **Tests**
  * Added transport-aware tests validating ordering, tie behavior, and diagnostic messages.


i386:

crates/skippy-topology/src/edge_order.rs
```diff
@@ -0,0 +1,188 @@
...
+        values.swap(start, index);
+    }
+}
+
+fn order_score(
+    order: &[(usize, NodeSpec)],
+    placement_signals: &[NodePlacementSignal],
+    edge_signals: &[StageEdgeSignal],
+) -> (u64, Vec<usize>) {
+    let edge_cost = order
```
ndizazzo, line 138: When we get a tie, edge ordering can stomp your package-aware warm-node sort since it uses indices assigned before that sort runs. We probably should rebuild that order after package sorting or keep a stable package order on ties.

ndizazzo (changes requested):
minor fixup

PR #815, opened 2026-06-10
Report Skippy artifact cold-start costs

i386:
## Summary

Add cold-start artifact cost diagnostics to Skippy topology planning.

This PR is only about explaining the artifact cost of an accepted package-aware plan. It does not change stage ordering, stage roles, runtime loading, or the binary protocol.

## Before

Package-aware planning could use cache and missing-artifact signals to affect placement, but the accepted plan did not summarize what startup would cost.

That left operators with a plan but not an explanation:

- how much of the selected topology is already cached
- how many bytes are missing
- how many missing bytes can come from peers
- how many bytes must fall back to remote download

## How It Works

`artifact_diagnostics` runs after stages are selected.

It aggregates selected stages:

- `cached_slice_bytes`
- `missing_artifact_bytes`
- transfer-supported missing bytes
- remote-download fallback bytes

It emits a single info diagnostic when artifact signals are present:

```text
[info] artifact cold-start plan: cached=64 bytes, missing=48 bytes, peer-transfer-eligible=32 bytes, remote-download-fallback=16 bytes
```

Plain weighted plans without artifact signals stay quiet, so non-package topology planning does not get noisy diagnostics.

## Why This Is Good

This makes package-aware planning explainable.

The planner can already choose a node because it has cached layers or because missing artifacts are cheap to transfer. This PR lets the accepted topology say that out loud. That matters for Skippy cold starts because slow startup is often artifact movement, not model execution.

It also prepares the next step toward metadata-first package planning: resolve package inventory, plan from cache/transfer signals, then materialize only the accepted stage artifacts.

## Validation

## Summary by CodeRabbit

* **New Features**
  * Enhanced diagnostic reporting for artifact planning: adds detailed breakdowns of total cached bytes, total missing bytes, how much is eligible for peer transfer versus remote-download fallback, and an overall diagnostic severity/code reflecting artifact transfer impact.

* **Tests**
  * Added unit tests that validate artifact diagnostic messages and transfer-support detection across placement scenarios, and ensure no diagnostics are emitted when artifact signals are absent.

### DIFF
--- a/crates/skippy-topology/src/artifact_diagnostics.rs
+++ b/crates/skippy-topology/src/artifact_diagnostics.rs
@@ -1,0 +1,54 @@
+use crate::{DiagnosticSeverity, NodePlacementSignal, PlanDiagnostic, PlanReasonCode, StagePlan};
+
+pub(crate) fn append_artifact_diagnostics(
+    diagnostics: &mut Vec<PlanDiagnostic>,
+    stages: &[StagePlan],
+    placement_signals: &[NodePlacementSignal],
+) {
+    let cached_bytes = stages
+        .iter()
+        .map(|stage| stage.cached_slice_bytes)
+        .sum::<u64>();
+    let missing_bytes = stages
+        .iter()
+        .map(|stage| stage.missing_artifact_bytes)
+        .sum::<u64>();
+
+    if cached_bytes == 0 && missing_bytes == 0 {
+        return;
+    }
+
+    let peer_transfer_bytes = missing_bytes_by_transfer_support(stages, placement_signals, true);
+    let remote_download_bytes = missing_bytes.saturating_sub(peer_transfer_bytes);
+    let code = if missing_bytes > 0 {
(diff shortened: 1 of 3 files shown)


i386:

crates/skippy-topology/src/artifact_diagnostics.rs
```diff
@@ -0,0 +1,54 @@
...
+    placement_signals: &[NodePlacementSignal],
+) {
+    let cached_bytes = stages
+        .iter()
+        .map(|stage| stage.cached_slice_bytes)
+        .sum::<u64>();
+    let missing_bytes = stages
+        .iter()
+        .map(|stage| stage.missing_artifact_bytes)
+        .sum::<u64>();
```
ndizazzo, line 15: Might wanna intersect with layer_start..layer_end so the totals match what cold start actually loads

ndizazzo (approved):

PR #816, opened 2026-06-10
Add Skippy stage role metadata

i386:
## Summary

Add explicit role metadata to Skippy topology stages.

This PR is only about topology semantics. It does not change placement scoring, artifact diagnostics, protocol bytes, or runtime transport.

## Before

Stage behavior was inferred from numeric `stage_index` and layer range:

- stage 0 usually acted as the driver
- the first layer owner implicitly handled embedding-side work
- middle stages were just “not first or last”
- the final layer owner implicitly acted as the readout/logits stage

That worked internally, but it made diagnostics and future UI/API views rely on convention instead of data.

## How It Works

`StagePlan` now includes:

```rust
roles: Vec<StageRole>
```

with serde defaulting so older serialized plans that do not carry role data still deserialize.

Role assignment is deterministic:

- first stage: `Driver`, `Embedding`
- middle stages: `Intermediate`
- last stage: `Readout`
- single-stage plan: `Driver`, `Embedding`, `Readout`

```mermaid
flowchart LR
  S0["stage 0\nDriver + Embedding"] --> S1["stage 1\nIntermediate"]
  S1 --> S2["stage 2\nReadout"]
  S2 -. "prediction return" .-> S0
```

## Why This Is Good

Skippy should not reverse stage numbering just because MLX rank ordering makes readout ownership obvious. Stage 0 has driver/session meaning in Skippy.

Explicit roles give us the useful part without changing the numbering model:

- diagnostics can say what a stage does
- UI/API surfaces can display topology responsibilities directly
- direct-return and readout-stage experiments can reason about `Readout` independently from `stage_index`
- single-stage and split-stage topologies use the same role vocabulary

## Compatibility

The new `roles` field has `#[serde(default)]`, so older topology JSON that lacks role data still reads as an empty role list.

## Validation

## Summary by CodeRabbit

* **New Features**
  * Stages now include role classifications that identify their position and function in the execution pipeline: driver, embedding, intermediate, or readout.

* **Tests**
  * Added comprehensive test coverage to validate stage role assignments across various topology configurations, including multi-stage and single-stage scenarios.

### DIFF
--- a/crates/skippy-topology/src/lib.rs
+++ b/crates/skippy-topology/src/lib.rs
@@ -70,6 +70,8 @@ pub struct StagePlan {
     pub stage_id: String,
     pub stage_index: u32,
     pub node_id: String,
+    #[serde(default)]
+    pub roles: Vec<StageRole>,
     pub layer_start: u32,
     pub layer_end: u32,
     pub layer_count: u32,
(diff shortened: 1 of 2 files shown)


i386:

PR #817, opened 2026-06-10
Add Skippy request cache epoch telemetry

i386:
## Summary

Add a typed request/cache epoch view for Skippy binary stage messages and emit checkpoint generation in server telemetry.

This PR is about request/cache coordination vocabulary. It does not add protocol fields, change encoded bytes, or reject stale frames yet.

## Before

The stage message already carried the pieces needed to identify a cache generation:

- `request_id`
- `session_id`
- `checkpoint_generation`
- `prompt_token_count`
- `decode_step`

But code treated them as separate fields. Telemetry had prompt/decode information, but not checkpoint generation in the common binary-message attribute set.

## How It Works

`StageRequestEpoch` groups the existing identity fields:

```text
request_id + session_id + checkpoint_generation + prompt_token_count + decode_step
```

`StageWireMessage::request_epoch()` derives the value from an existing message.

The helper also supports:

- `same_flow(...)` for request/session matching
- `is_stale_for(...)` for future stale-frame/cache-epoch checks

`skippy-server` now uses the epoch helper in binary transport telemetry and emits:

```text
skippy.checkpoint_generation=3
skippy.prompt_token_count=8192
skippy.decode_step=0
```

## Why This Is Good

Skippy is getting more aggressive about prefix/cache reuse, restore, trim, prefill, and decode overlap. Those paths need one shared way to talk about which cache generation a frame belongs to.

This PR creates that vocabulary without changing behavior yet. It makes future stale restore/trim/decode checks smaller and easier to review because the identity comparison is centralized.

## Compatibility

This is **not** a wire-format change.

- no new encoded fields
- no stage state version bump
- no protobuf/gossip change
- no mixed-version binary framing break

The `skippy-protocol` crate changes, but the bytes sent between stages do not.

## Validation

- `cargo check -p mesh-llm`

## Summary by CodeRabbit

* **Refactor**
  * Improved request-epoch tracking to more reliably identify request/session flows and progression state.
  * Telemetry now reports checkpoint, prompt, and decode progression using consistent attribute keys.
* **New Features**
  * Exposed a compact epoch snapshot used for comparing flow identity and staleness.
* **Tests**
  * Added unit tests validating flow matching and staleness ordering across progression fields.

### DIFF
--- a/crates/skippy-server/src/binary_transport.rs
+++ b/crates/skippy-server/src/binary_transport.rs
@@ -1609,6 +1609,7 @@ fn binary_message_attrs(
     message: &StageWireMessage,
 ) -> std::collections::BTreeMap<String, serde_json::Value> {
     let mut attrs = lifecycle_attrs(config);
+    let epoch = message.request_epoch();
     attrs.insert(attr::SESSION_ID.to_string(), json!(session_id.to_string()));
     attrs.insert(
         attr::REQUEST_ID.to_string(),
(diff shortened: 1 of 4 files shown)


i386:

crates/skippy-server/src/binary_transport.rs
```diff
@@ -1687,13 +1625,14 @@ fn binary_message_attrs(
     );
     attrs.insert("skippy.token_count".to_string(), json!(message.token_count));
     attrs.insert(
-        "skippy.prompt_token_count".to_string(),
-        json!(message.state.prompt_token_count),
+        "skippy.checkpoint_generation".to_string(),
```
ndizazzo: We could pull the new telemetry key into the shared metrics attributes instead of adding another inline string here? Not a blocker, but it would keep telemetry naming easier to track and reduce the chance of drift if this field gets emitted or queried elsewhere. 

This could live alongside `REQUEST_ID` and `SESSION_ID` in `skippy-metrics::attr`.

ndizazzo (commented):
seems fine small sug

PR #820, opened 2026-06-10
Stabilize mesh MoA context and tool loops

michaelneale:
## Summary

This stabilizes the `mesh` MoA virtual model for agent clients without copying the old oversized branch:

- Advertise real context for public aliases and for virtual `mesh`, and filter MoA workers by required context with a small fallback only for unknown tiny prompts.
- Preserve recent same-chat context in MoA worker/reducer prompts so a follow-up question can refer to earlier turns in the same conversation.
- Normalize Responses API tools and function-call outputs into chat shape, and emit proper Responses SSE lifecycle events for both text and function-call streams.
- Harden MoA tool loops around malformed/empty native tool calls, repeated tool IDs, repeated same-tool repair, and disabled-tools output.
- Bias MoA strong/reducer role assignment toward larger known models using existing advertised `parameter_count_b` metadata after context eligibility.
- Update recommended small auto packs toward Gemma-4-E4B and fix mobile console double-tap behavior.

## Protocol

No mesh wire/protobuf schema change. This consumes existing served model descriptors/runtime context metadata and keeps mixed-version gossip additive.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p mesh-mixture-of-agents --all-targets`
- `cargo test -p openai-frontend --lib`
- `cargo test -p mesh-llm-host-runtime --lib` (`1448 passed, 0 failed, 8 ignored`)
- `cargo clippy -p openai-frontend -p mesh-mixture-of-agents -p mesh-llm-host-runtime -p mesh-llm-client -p mesh-llm-node -p mesh-llm --all-targets -- -D warnings`
- `pnpm run typecheck`
- `pnpm test -- src/features/chat/api/use-chat.test.tsx`
- `pnpm exec prettier --check src/components/ui/tooltip.tsx src/features/chat/api/use-chat.ts src/features/chat/api/use-chat.test.tsx src/features/chat/components/composer/ChatComposer.tsx`
- `pnpm exec eslint src/components/ui/tooltip.tsx src/features/chat/api/use-chat.ts src/features/chat/api/use-chat.test.tsx src/features/chat/components/composer/ChatComposer.tsx --max-warnings=0`
- `just build`

Live validation against rebuilt `./target/debug/mesh-llm serve` with local Qwen3-4B-128K + Gemma-4-E4B:

- `/v1/models` advertises both concrete models at `context_length: 131072` and virtual `mesh` at `context_length: 131072`; concrete metadata includes `parameter_count_b` (`4.0`, `7.5`).
[...]

PR #837, opened 2026-06-12
MoA: don't let small-model consensus pre-empt a still-running large model

michaelneale:
## Summary

When `model: "mesh"` fans out across a mixed mesh — one large model (e.g. MiniMax on studio) plus several small models — the large model's answer could be silently discarded: two fast small workers agreeing (or the 3s answer-grace timer) finalized the turn while the large model was still prefilling. Users got small-model consensus even when a much stronger answer was seconds away.

After this change, MoA holds small-tier-only answers for a **bounded patience window** (20s) when a big-tier Strong worker is still running:

- If the strong worker lands in time, its answer participates (and consensus that *includes* the strong answer ships immediately — that's agreement with the strong model, not against it).
- If the strong worker is stuck or slow past the window, the held small-tier consensus ships — the gate switches **hard off** at expiry, so a hung peer can never hold a turn hostage (the failure mode that sank #820). A dedicated wake-up re-evaluates held outputs at expiry rather than waiting for `worker_timeout`.
- **Tool proposals are never held** — they're schema-verified by `tool_guard`, and agent loops (goose / OpenClaw) keep their current latency.
- **Same-tier pools are unaffected**: `has_quality_gap` only arms the gate when the Strong worker is big-tier *and* small-tier workers are present. Many-small-models meshes keep today's early-exit latency, preserving the consensus-lift behavior.
- Answer grace now prefers the Strong worker's qualifying answer over marginally-higher self-reported confidence from a smaller model (confidence is not comparable across models).

## Why

Self-MoA (arXiv:2502.00674) shows MoA output quality is far more sensitive to proposer *quality* than diversity — mixing weak proposers into a strong model's pool actively hurts. "Are More LLM Calls All You Need?" (arXiv:2403.02419) shows majority voting degrades on hard queries. The goal of mesh MoA is to lift intelligence when there are many comparable small models, not to dumb down a large one; this change separates those two regimes.

## Architecture

- `worker::has_quality_gap` — tier analysis over `(model_name, role)` pairs, reusing the existing single-digit-B name heuristic shared with the router.
- `arbiter::StrongGate` — explicit gate state passed into `try_early_decision`; `Off` reproduces pre-change behavior exactly (all existing arbiter tests pass unchanged with `Off`).
- `fanout::GatherPolicy` — groups the timing knobs (`first_answer_grace`, `grace_mode`, new `strong_patience`).
- `GatewayConfig.strong_patience` — zero disables the gate entirely; sim tests run with zero except the new gate-specific sims.

## Protocol

No wire/protocol changes. Purely host-local decision logic inside the MoA gateway; mixed-version meshes unaffected.

## Validation

- `cargo test -p mesh-mixture-of-agents` — 158 lib tests + all sims pass, including new `sim_strong_patience.rs` pinning: held consensus until strong lands, **hard release at patience expiry with a hung strong worker** (anti-#820), and same-tier pools keeping early-exit.
- `cargo test -p mesh-llm-host-runtime --lib` — 1439 passed.
- Clippy `-D warnings` clean on both crates; fmt clean.
- Live multi-node validation (lab mini + public mesh, goose + claw vs baseline) to follow in PR comments.

## Summary by CodeRabbit

* **New Features**
  * Added configurable "strong patience" gating to hold small-tier consensus while higher-quality workers are still producing, with a patience timeout to release decisions.
  * Detects quality gaps between worker tiers to enable tier-aware arbitration and prefer a qualifying strong answer when it disagrees with a small-tier consensus.

* **Refactor**
  * Consolidated grace/timing into a GatherPolicy for clearer tier-gating behavior.

* **Tests**
  * Expanded tests covering gating, patience expiry, preference for strong answers, and all-small pools.


crates/mesh-mixture-of-agents/src/arbiter.rs
```diff
@@ -302,6 +339,75 @@ pub fn try_early_decision(
...
 }
 
+/// Resolve a majority answer cluster against the strong-worker tier gate.
+///
+/// Consensus that includes the strong worker's answer passes through —
+/// that is agreement *with* the strong model, not against it. Small-tier-
+/// only consensus is held back (returns `None`) while the strong worker
+/// is still pending so the model most likely to be right gets to weigh
+/// in. The fan-out loop bounds how long the hold can last.
+fn tier_aware_consensus_decision(
```
ndizazzo, line 349: When we pick a strong worker's answer over a small-tier consensus... it only does a `is_usable_answer`... and we need >= 0.5 for the small ones.

In the case where we have 1 strong worker at 0.0, it can override 3 workers with a strong 0.9 confidence. 

Is that on purpose? If not we could give some kind of alignment with a `GRACE_MIN_CONFIDENCE = 0.5` to. the strong_answer

ndizazzo (approved):

crates/mesh-mixture-of-agents/src/arbiter.rs
michaelneale: I think so - a strong worker (if we really know strong) almost always would be better than weaker ones and the scoring (confidence in scoring is limited) - at least I believe that makes sense, but will need to think about it.

michaelneale: @ndizazzo yeah confirmewd - there is no world in which a strong model in this case would do 0 anyway, but in general yes would prefer it as long as it is an actual response.

PR #842, opened 2026-06-13
feat(skippy): Add native runtime event visibility

ndizazzo:
## Summary

This gives Mesh LLM a first-class way to surface native Skippy/llama.cpp model-open progress without letting native code take over readiness, routing, or recovery decisions.

Operators and developers will now be able to see more of what is happening while embedded Skippy runtimes open models: when native loading starts, which backend device gets selected, how loading progresses, when native loading finishes, and when native code reports a handled open failure. Those facts flow into the existing runtime output surfaces, so startup is easier to understand without changing the meaning of `ModelReady` or `RuntimeReady`.

## What this allows us to do

- Make slow or confusing startup phases visible instead of looking like a silent hang.
- Explain native backend/device selection during model open.
- Distinguish native model-open progress from Rust-owned readiness, so dashboards can show activity without claiming the runtime is ready too early.
- Keep older native runtimes working through feature and symbol fallback.
- Preserve raw native log visibility as a debug fallback while adding structured lifecycle facts for normal operation.
- Build future native lifecycle event families on a documented, versioned boundary.
- Give maintainers one inventory for hook/callback/emitter surfaces and their readiness impact.

## User-facing behavior

Startup output can now include native model-open lifecycle visibility for Skippy-backed loads. These messages are informational or warning-level visibility events only. Readiness is still emitted by the existing Rust startup path after the runtime is actually ready.

If the native runtime does not support the event ABI, Mesh LLM falls back to the existing no-events open path.

## Validation

- `just build`
- `cargo check -p skippy-ffi`
- `cargo test -p skippy-ffi --lib`
- `cargo test -p skippy-runtime --lib`
- `cargo test -p mesh-llm-host-runtime --lib inference::skippy`
- `cargo test -p mesh-llm-tui --lib`
- `cargo check -p mesh-llm`

## Summary by CodeRabbit

* **New Features**
  * Native runtime event reporting for model-open lifecycle and new native/FFI entrypoints.

* **Behavior**
  * Native visibility events are observational and do not advance or replace dashboard readiness/state.

* **Documentation**
  * Added emitter/hooks inventory, native runtime event callback contract, and updated ABI/README guidance.

* **Tests**
  * New unit tests for event translation, ordering, fallbacks, dropped-receiver safety, and dashboard visibility.

* **Chores**
  * Increased CI smoke startup timeout (configurable) and added CLI/server --max-inflight with runtime propagation and prewarm clamping.


ndizazzo:
@coderabbitai review

ndizazzo:
@coderabbitai review

ndizazzo:
@i386 mind taking a quick look at this one? Currently, we do some hand-wavy parsing of debug output from llamacpp, and I'd like some real anchor points in Rust to do state / model loading management from. 

This doesn't add them, but adds all of the scaffolding for us to. I plan to follow this up with stripping out all of the debug parsing for model loading in debug mode as a first use

crates/skippy-runtime/src/runtime_events.rs
```diff
@@ -0,0 +1,387 @@
...
+) -> (*mut RawModel, Status, *mut RawError)
+where
+    OpenFn:
+        FnOnce(*const RawRuntimeEventReporter, *mut *mut RawModel, *mut *mut RawError) -> Status,
+    EventFn: FnMut(RuntimeEvent),
+{
+    let mut registration = ModelOpenEventReporterRegistration::new();
+    let mut raw = ptr::null_mut();
+    let mut error = ptr::null_mut();
+    let status = open_fn(registration.reporter_ptr(), &mut raw, &mut error);
```
i386, line 264: I think this still buffers the native progress events until `open_fn` returns, which means we don’t actually surface progress while the slow model-open work is happening. It also means the 64-event channel can fill during a long/noisy load and drop later events, potentially including the finished/failure event. Can we drain concurrently during model open, or otherwise forward from the callback without waiting for open to complete?

third_party/llama.cpp/patches/0083-skippy-add-model-open-runtime-events-ABI.patch
```diff
@@ -0,0 +1,726 @@
...
+-#define SKIPPY_ABI_VERSION_PATCH 25
++#define SKIPPY_ABI_VERSION_PATCH 26
+ 
+ enum skippy_feature {
+     SKIPPY_FEATURE_RUNTIME_SLICE          = 1 << 0,
+@@ -52,7 +52,8 @@ enum skippy_feature {
+     SKIPPY_FEATURE_EXTERNAL_MEDIA_PREFILL = 1 << 20,
+     SKIPPY_FEATURE_CHAT_TEMPLATE_TOOLS    = 1 << 21,
+     SKIPPY_FEATURE_CHAT_SAMPLING_GRAMMAR  = 1 << 22,
+-    SKIPPY_FEATURE_BACKEND_DEVICES        = 1 << 23,
```
i386, line 62: This looks like it reuses the old backend-devices feature bit for runtime events and moves backend devices from `1 << 23` to `1 << 24`. Since these feature bits are part of the ABI, changing an existing bit can confuse older/newer consumers. I think backend devices should stay on `1 << 23`, and runtime events should take the next unused bit.

i386 (commented):
I left two inline review comments focused on runtime-event behavior and ABI compatibility.

crates/skippy-correctness/src/runner.rs
```diff
@@ -143,6 +145,7 @@ struct BinaryStateHandoffConfig {
     runtime_lane_count: Option<u32>,
     borrow_resident_hits: bool,
     cache_decoded_result_hits: bool,
+    skip_suffix_prefill_check: bool,
```
i386, line 148: Looks like `--max-inflight` is still skipped for the `state-handoff` path. Split/chain pass it through to `serve-binary`, but this state-handoff config does not carry `args.server.max_inflight`, and the source/restore `serve-binary` commands don’t include it either. Can we pass `args.max_inflight` through on both child commands too?

i386 (commented):
One more state-handoff propagation note.

PR #847, opened 2026-06-13
fix(console): render thinking traces as markdown

ndizazzo:
## Original problem

Thinking trace content in the React chat UI rendered as pre-wrapped plain text, so markdown markers like `**bold**`, ordered steps, nested bullets, inline code, and emphasis appeared raw inside reasoning traces. Final assistant responses already rendered through markdown, so the trace panel looked inconsistent and harder to scan.

## Diagnostics

The issue was isolated to `MessageRow.tsx`: normal assistant segments used `AssistantMarkdown`, while thinking segments used a raw `span` with `whitespace-pre-wrap`.

Screenshot of the fixed thinking trace:

Thinking trace markdown fix

## Fix

- Render thinking trace segments through the existing assistant markdown renderer.
- Keep trace-specific muted styling for bold text so the reasoning panel stays visually secondary to the final answer.
- Restore semantic ordered and unordered list rendering so numbered reasoning steps and nested bullets format correctly.
- Add regression coverage for bold, inline code, italics, ordered lists, nested bullets, and hidden raw markdown markers inside thinking traces.

No protocol or compatibility impact; this is UI-only rendering behavior.

## Validation

- [x] `pnpm prettier --check src/features/chat/components/MessageRow.tsx src/features/chat/components/MessageRow.test.tsx`
- [x] `pnpm vitest run src/features/chat/components/MessageRow.test.tsx`
- [x] `just build`
- [x] Browser check in the developer playground confirmed the thinking trace DOM includes `strong`, `code`, `em`, `ol`, and `ul` nodes, with no raw `**` visible.
- [x] UI changes include screenshot above.

## Release Notes

* **New Features**
  * Enhanced markdown rendering for thinking content with support for lists, text emphasis, code formatting, and improved visual styling
  * Simplified thinking content presentation in the chat interface

* **Bug Fixes**
  * Improved text selection behavior and CSS styling for thinking segments to ensure consistent appearance

### DIFF
--- a/crates/mesh-llm-ui/src/features/chat/components/MessageRow.test.tsx
+++ b/crates/mesh-llm-ui/src/features/chat/components/MessageRow.test.tsx
@@ -233,7 +233,7 @@ describe('MessageRow', () => {
     expect(screen.getByText('Thinking trace').closest('[data-thinking-state="complete"]')).toHaveTextContent(
       'Thinking trace'
     )
-    expect(screen.getByText('Check geography facts before answering.')).toHaveClass('select-text')
+    expect(screen.getByText('Check geography facts before answering.').closest('.select-text')).toBeInTheDocument()
     expect(screen.getByText('The capital of France is Paris.').closest('.select-text')).toBeInTheDocument()
   })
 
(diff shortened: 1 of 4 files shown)


michaelneale:
Looks good - probably could relabel it to just "thinking". Dont need to mention trace as extra jargon in gui?

PR #848, opened 2026-06-13
[codex] Use sccache for Rust build scripts

i386:
## Summary

Local Linux and release builds now automatically use `sccache` for Rust compilation when it is installed, matching the existing macOS and Windows behavior. The scripts still respect an explicitly configured `RUSTC_WRAPPER`, so callers can override or disable the wrapper as needed.

## Impact

This should improve repeated Rust build times for `just build` on Linux and `just release-build` on Linux/macOS without changing the llama.cpp C/C++ sccache wiring.

## Validation

- `bash -n scripts/build-linux.sh`
- `bash -n scripts/build-release.sh`
- `git diff --check`

## Summary by CodeRabbit

* **Chores**
  * Optimized build configuration to accelerate compilation times for both Linux and release builds by implementing compiler caching when available.

### DIFF
--- a/scripts/build-linux.sh
+++ b/scripts/build-linux.sh
@@ -94,6 +94,15 @@ EOF
     echo "Using Rust linker: $(command -v ld.lld)"
 }
 
+configure_rust_cache() {
+    if [[ -n "${RUSTC_WRAPPER:-}" ]]; then
+        echo "Using Rust compiler wrapper: $RUSTC_WRAPPER"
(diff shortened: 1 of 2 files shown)


ndizazzo (approved):

ndizazzo:
I could have sworn we had this before... maybe it got stomped in our CI refactoring / churning

i386:
Me too but whatever :)

PR #849, opened 2026-06-13
[codex] Stabilize skippy smoke chain startup

i386:
## Summary

Stabilizes the PR `Linux tests (skippy-smoke)` lane by balancing the dense 3-stage smoke split and making dense chain startup timeout configurable.

## Why

Several unrelated PRs have failed with `stage 2 binary server did not become ready`. The old default dense split `1,2` leaves stage 2 owning nearly the full 30-layer smoke model, so startup can exceed the fixed 60s readiness timeout on current GitHub runners.

## Details

- Leaves `DENSE_SMOKE_SPLIT_1` / `DENSE_SMOKE_SPLIT_2` override support intact.
- When unset, computes roughly one-third/two-third balanced splits from the inspected dense layer count.
- Adds `DENSE_CHAIN_STARTUP_TIMEOUT_SECS`, defaulting to `180`, and passes it to `skippy-correctness chain`.

## Validation

- `bash -n scripts/skippy-ci-smoke.sh`
- `git diff --check`

## Summary by CodeRabbit

* **Chores**
  * Enhanced smoke testing infrastructure with flexible split boundary configuration that no longer enforces default values
  * Introduced configurable startup timeout for chain operations, replacing hardcoded value (default: 180 seconds)


michaelneale:
Interesting wonder why it has to be flexible with this now

PR #846, opened 2026-06-13
Add bounded direct path repair

i386:
## Summary

Adds a targeted mesh-level direct-path repair flow for peers whose selected iroh path has fallen back to relay or is unknown even though a direct UDP candidate is available.

The repair path uses a dedicated mesh stream byte (`0x0e`) and a `DirectPathRequest` frame. It does not use `STREAM_SUBPROTOCOL`, does not gossip, and only asks one peer per maintenance tick to reverse-dial the requester’s current advertised endpoint address.

## Why

In private/mDNS lab meshes, one side may need the other side to initiate the UDP attempt for iroh to converge on a direct path. The previous behavior could sit on relay/unknown path state even when a usable LAN candidate existed. This keeps the fix close to mesh networking and iroh connection management without adding broad discovery chatter.

This follows the iroh layering model: iroh authenticates the endpoint reached by a dial, while mesh-llm constrains which peer-advertised candidates this mesh-level repair path is allowed to hand to iroh.

## Flow Diagram

```mermaid
sequenceDiagram
    autonumber
    participant A as Node A maintenance loop
    participant Conn as Existing mesh QUIC connection
    participant B as Node B mesh dispatcher
    participant Iroh as iroh endpoint

    Note over A,B: Existing admitted mesh connection is already alive
    A->>A: Observe selected path = relay/unknown<br/>and peer has direct UDP candidate
    A->>A: Apply grace period, one-at-a-time gate,<br/>sender cooldown, inflight suppression
    A->>Conn: Open bi stream with byte 0x0e
    Conn->>B: DirectPathRequest{requester_id, gen, EndpointAddr}
    B->>B: Validate generation, requester id,<br/>admitted peer, known direct candidate, receiver cooldown
    B->>Iroh: Dial A using only previously advertised direct candidates
    Iroh-->>B: New QUIC connection if endpoint identity verifies
    B->>B: Install connection, dispatch streams,<br/>initiate gossip
```

## Security / Layering Note

This is only reachable from an already-admitted mesh peer over an existing QUIC connection. iroh still verifies the remote endpoint identity during the reverse dial; a candidate cannot produce an accepted connection unless the remote proves it owns the requested endpoint id.

[...]

PR #851, opened 2026-06-13
Reliably find LAN direct paths in mDNS and default modes, regardless of which node hosts

michaelneale:
Follow-up to #846. With just #846, two nodes on the same LAN still fell back to the relay (≈250 ms) when the **M4 hosted** and the mini joined, and mDNS discovery did not find a direct path in that direction. This makes direct LAN paths land reliably no matter which node starts the mesh — with the default (token-shared, non-public) discovery and with mDNS — by pinning QUIC to the LAN interface and adding a reverse-dial path so the host can dial the joiner back.

## What you get

- Two nodes on the same LAN now connect over a **direct path** (single-digit-to-low-tens-of-ms RTT) instead of falling back to the relay, in both directions:
  - host → joiner and joiner → host
  - default token-shared (non-public) discovery and mDNS discovery mode
- Works regardless of which node starts the mesh.

## How

- **Auto-bind QUIC to the LAN interface** (`68ffa64e`): detect the primary LAN IPv4 and bind QUIC to it, and clear the unused IP transport so relay-less connections don't stall on multipath negotiation.
- **Reverse-dial + LAN beacon** (`b0ece8a0`): the joiner advertises its endpoint address (over mDNS TXT and a small link-local LAN beacon), and the host browses and dials it back. This covers the single-homed-initiator direction that one-way dialing missed. The mDNS instance name is now node-id-based so multiple meshes on one LAN don't clobber each other's records.

## Process

```mermaid
flowchart TD
    start["Node starts in token or mDNS mode"] --> bind["Choose QUIC bind address"]
    bind --> explicit{"--bind-ip set?"}
    explicit -->|yes| use_explicit["Bind QUIC to explicit address"]
    explicit -->|no| detect["Detect primary LAN IPv4"]
    detect --> lan_found{"LAN IPv4 found?"}
    lan_found -->|yes| use_detected["Bind QUIC to LAN IPv4"]
    lan_found -->|no| default_bind["Use default QUIC bind"]

    use_explicit --> endpoint["Build LAN-filtered EndpointAddr"]
    use_detected --> endpoint
    default_bind --> endpoint

    endpoint --> token_path["Invite token / gossip carries reachable candidates"]
    endpoint --> mdns_path["mDNS advert includes ep_addr TXT"]
    endpoint --> beacon_path["LAN beacon emits EndpointAddr"]

    token_path --> joiner["Joiner attempts normal outbound dial"]
    mdns_path --> browse["Peers browse LAN mDNS"]
    beacon_path --> listen["Peers hear multicast or unicast beacon"]

    joiner --> direct{"Direct path established?"}
    browse --> reverse["Peer dials advertised EndpointAddr back"]
    listen --> reverse
    reverse --> direct

    direct -->|yes| ready["Mesh connection uses low-latency LAN path"]
    direct -->|no| retry["Retry on next mDNS/beacon/direct-path maintenance tick"]
    retry --> browse
    retry --> listen
```

## Protocol

Backward compatible. The new `ep_addr` mDNS TXT key and the LAN beacon are additive — older nodes simply ignore the TXT key and don't participate in the beacon, and all existing discovery/dial paths are unchanged. No gossip or QUIC wire-format changes.

## Validation

- `cargo fmt --all -- --check` — clean
- `cargo test -p mesh-llm-host-runtime --lib` — 1449 passed, 0 failed
- Two-node lab (Apple M4 ↔ mini on the same LAN), committed binary on both nodes:

| Mode | Direction | Runs | Result |
|---|---|---|---|
| Default (token) | M4-host → mini-join | 3/3 | DIRECT 8–15 ms |
| Default (token) | mini-host → M4-join | 2/2 | DIRECT 13–14 ms |
| mDNS | M4-host → mini-join | 11/11 | DIRECT 8–53 ms |
| mDNS | mini-host → M4-join | 2/2 | DIRECT 13 ms |

Survived a sleep/wake cycle on both nodes.

## Summary by CodeRabbit

* **New Features**
  * Enhanced local network peer discovery with automatic IP detection and multicast beacon bootstrap for relay-less direct connections.
  * Tier-aware response handling for mixed-quality worker pools, improving consensus decisions.

* **Improvements**
  * Renamed "thinking trace" label to "thinking" for clarity in assistant responses.
  * Enhanced markdown rendering for thinking content with proper styling.
  * Improved mobile experience for catalog table display on small screens.


crates/mesh-llm-host-runtime/src/runtime/mod.rs
```diff
@@ -7896,6 +7919,52 @@ async fn spawn_run_auto_nostr_publisher(
...
+fn spawn_mdns_reverse_dial(options: &RuntimeOptions, node: &mesh::Node) {
+    if options.mesh_discovery_mode != mesh_discovery::MeshDiscoveryMode::Mdns {
+        return;
+    }
+
+    // Ensure an mDNS advertisement (carrying our ep_addr) exists even without
+    // --publish, so peers can discover a dial-back address. The standard
+    // publish path only runs with --publish; spawn a publisher here otherwise.
+    if !options.publish {
+        tokio::spawn(Box::pin(mesh_discovery::publish_lan_loop(
```
i386: This extra mDNS publisher is spawned without its `JoinHandle`, and the same happens below for the reverse-dial loop and LAN beacon. `shutdown_run_auto_runtime()` only aborts `discovery_publisher`, so in shutdown/restart paths these new loops can keep publishing, dialing, and holding the beacon UDP socket until process exit. Please return/store a small guard of these handles and abort it alongside `discovery_publisher` during runtime shutdown.

crates/mesh-llm-host-runtime/src/mesh/mod.rs
```diff
@@ -4797,9 +4985,34 @@ impl Node {
...
         self.state.lock().await.dead_peers.remove(&addr.id);
+        self.remember_join_target(addr.clone()).await;
         self.connect_to_peer(addr).await
     }
 
+    /// Record a join target address so the LAN beacon can unicast a dial-back
+    /// hint to it even before a direct connection forms.
+    async fn remember_join_target(&self, addr: EndpointAddr) {
+        let mut targets = self.join_targets.lock().await;
+        if !targets.iter().any(|t| t.id == addr.id) {
```
i386: This keeps the first address ever seen for a peer and ignores later invite tokens with the same endpoint id. If the peer restarts or rebinds to a new QUIC port, the LAN beacon will keep unicasting dial-back hints to the stale socket address, so this recovery path stops working after a common rebind/restart case. Please replace/update the existing entry when `t.id == addr.id` instead of dropping the new address.

i386 (changes requested):
I rechecked the latest head before posting. The LAN detection issue from the earlier pass looks addressed by the private-interface fallback, so I’m leaving the remaining two actionable findings only.

PR #873, opened 2026-06-18
Require explicit native runtime version selection

i386:
MeshLLM now pairs native runtimes with an explicit MeshLLM version and Skippy ABI instead of accepting any cached runtime with a matching ABI.

## What changed

- Native runtime resolution now rejects stale cached runtime artifacts when their `mesh_version` does not match the selected MeshLLM runtime version.
- Startup defaults to the running binary's MeshLLM release version and current Skippy ABI.
- Added an explicit config override:

```toml
[runtime.native_runtime]
mesh_version = "0.68.0"
skippy_abi = "0.1.25"
```

- `mesh-llm runtime install` and `mesh-llm runtime list --available` use the same explicit selector from config.
- macOS native runtime packaging now rewrites local dylib dependencies to packaged `@rpath/...` names and adds `@loader_path`; package verification now catches absolute packaged-dylib dependencies and missing `@loader_path`.

## Why

Upgrading from a MeshLLM version without separate native runtimes could leave an old cached runtime, such as `0.68.0`, selected by a newer binary. On macOS, that stale runtime could then fail during `dlopen` because packaged dylibs still referenced build-tree paths for sibling llama libraries.

## Runtime install scenarios

| Scenario | Behavior |
|---|---|
| Fresh install, no config | Installs the recommended native runtime for the new MeshLLM release, then prunes with the current release active. |
| Upgrade, no config | The new binary installs the new release runtime; old cached runtimes are rejected by `mesh_version` mismatch and pruned. |
| Startup, matching runtime already cached | Loads the cached runtime directly; no install attempt. |
| Startup, no compatible cached runtime | Attempts a one-shot install for the selected version; if install fails, continues without a dynamic native runtime. |
| Config pins only `mesh_version` | Install derives ABI from that release manifest; startup, prune, and doctor target that version. |
| Config pins `mesh_version` + `skippy_abi` | Resolver requires both exact MeshLLM version and ABI. |
| Config pins `selection` | Startup, install, list, and doctor use that selection, e.g. `cuda12` or `exact:...`. |
| CLI runs `runtime install cuda12` with config selection | CLI positional selection wins for that install invocation. |
| Upgrade while pinned to old version | New binary honors the pin; `runtime prune --active-only` keeps the pinned version instead of pruning it away. |
| Explicit `runtime prune --mesh-version X` | Explicit CLI version wins over config. |
| Bad config has `skippy_abi` or `selection` without `mesh_version` | Config validation fails. |
| `runtime list --installed` with bad config | Still lists cache; it does not load config. |
| `runtime list --available` with no compatible runtime | Prints each artifact plus rejection reasons. |
| Cached runtime manifest omits `mesh_version` | Rejected; it is not rewritten to the selected version. |
| Bundle has same runtime ID but wrong version/ABI | Not used as source for the selected artifact. |
| `mesh-llm doctor` with pin | Shows running version, selected runtime version, configured ABI/selection, and pinned status. |

## Validation

- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-native-runtime --lib`
- `cargo test -p mesh-llm-config --lib`
- `cargo test -p mesh-llm-host-runtime --features dynamic-native-runtime --lib system::native_runtime`
- `bash -n scripts/package-native-runtime.sh`
- `bash -n scripts/verify-native-runtime-package.sh`
- `cargo clippy -p mesh-llm --features dynamic-native-runtime --all-targets -- -D warnings`

## Summary by CodeRabbit

* **New Features**
  * Added `[runtime.native_runtime]` settings to pin native `mesh_version`, `skippy_abi`, and `selection` for native runtime resolution.
  * Native runtime list/install/remove/prune/doctor now render via consistent human/JSON output, including more detailed doctor reporting for pinned versions.
* **Bug Fixes**
  * Native runtime resolution now rejects candidates with mismatched mesh versions (in addition to ABI/compat checks).
  * Install no longer implicitly fills `skippy_abi_version` when not provided.
  * macOS packaging now rewrites staged library paths/dependencies and enforces `@loader_path` in verification.
* **Tests / Documentation**
  * Added config parsing/validation tests and updated the native runtimes pinning documentation.

### DIFF
--- a/crates/mesh-llm-commands/src/runtime_native.rs
+++ b/crates/mesh-llm-commands/src/runtime_native.rs
@@ -1,17 +1,17 @@
+mod formatters;
+
 use anyhow::Result;
-use mesh_llm_native_runtime::{
-    HostRuntimeProfile, NativeRuntimePruneMode, NativeRuntimeResolver, RuntimeSelection,
-};
+use mesh_llm_native_runtime::{NativeRuntimePruneMode, NativeRuntimeResolver, RuntimeSelection};
 use mesh_llm_runtime_install::{
     CURRENT_MESH_VERSION, NativeRuntimeDownloadProgressCallback, NativeRuntimeInstallOptions,
-    NativeRuntimeInstallStatus, NativeRuntimeManifestOptions, host_runtime_profile,
-    install_native_runtime, load_release_manifest, native_runtime_cache,
+    NativeRuntimeManifestOptions, host_runtime_profile, install_native_runtime,
+    load_release_manifest, native_runtime_cache,
 };
-use serde::Serialize;
-use serde_json::json;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
+
+use formatters::{AvailableRuntimeRow, NativeRuntimeDoctorReport, runtime_native_formatter};
 
 #[derive(Clone, Copy, Debug, Default)]
 pub struct NativeRuntimeConfigSelection<'a> {
@@ -37,6 +37,7 @@ pub async fn run_native_runtime_list(
     let mesh_version = configured.mesh_version_or_current();
     let selection = RuntimeSelection::parse(configured.selection)?;
     let cache = native_runtime_cache(cache_dir)?;
+    let formatter = runtime_native_formatter(json_output);
     if available {
         print_configured_selector(configured, json_output);
         if !json_output && manifest_path.is_none() && bundle_dirs.is_empty() {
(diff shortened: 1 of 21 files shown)


ndizazzo (changes requested):
I've been having some struggles with the ergonomics of the CLI lately - so my review is going to be a bit more centred around the experience:

1) `runtime list --available` wont give any rejection reasons if there's nothing comparable... largely because resolver.rs bails out before returning any evaluated candidates against a matched runtime.
2) **native runtime config schema understates the restart boundary.** (my agent found this one): 

```
The new settings are registered via runtime_setting at
[built_in_schema.rs (line 143)]
(mesh-llm-config/src/model/built_in_schema.rs:143), and that helper
marks runtime settings as API-visible / dynamic-validation-only while
inheriting ModelReload restart scope from basic_setting at
[built_in_schema.rs (line 895)]
(crates/mesh-llm-config/src/model/built_in_schema.rs:895).

But the selector is consumed before dynamic native libraries are loaded
during host initialization in [lib.rs (line 63)]
(crates/mesh-llm-host-runtime/src/lib.rs:63).

This should be a process-restart setting, probably with a
native-runtime-specific schema helper and a clearer description.
Otherwise the UI/API can imply that a model reload is sufficient.
```

Otherwise, seems like a good change. I tend to want poweruser-type features and this has the right config additions I would wanna see. A rough improvement there would be helping the discoverability of this (doc might be fine, or add to the sample templates).

PR #878, opened 2026-06-18
model-hf: clean up broken hf_hub cache symlinks after model deletion (#877)

### LINKED ISSUE #877: Deleting a Layer-Package Model Leaves Broken Manifest State
## Summary

After deleting a model that uses a layer package, attempting to download/install the model again fails immediately because the downloader believes the package manifest is already present, but then cannot read it.

## Command

`./target/release/mesh-llm models download unsloth/Qwen3.6-27B-MTP-GGUF:UD-Q4_K_XL`

## Actual Behavior

The downloader reports:

`Ready meshllm/Qwen3.6-27B-UD-Q4_K_XL-layers/model-package.json`

but then fails with:

`No such file or directory (os error 2): read package manifest`

So the downloader appears to think the manifest exists or is ready, while the actual file is missing.

## Expected Behavior

After deleting a model, reinstalling/downloading it should work cleanly.

If the layer package manifest is missing, the downloader should either:

- Re-download `model-package.json`
- Repair the incomplete package state
- Clear stale cache/index metadata before retrying
- Report which exact path is missing

## Impact

- A deleted model cannot be reinstalled.
- The local model cache can get stuck in a broken state.
- Users may need to manually find and delete internal package/cache metadata.
- The error message is confusing because it says the manifest is “Ready” immediately before failing to read it.

## Reproduction

1. Download a model that uses a layer package.
2. Delete the model.
3. Run the same download command again:

   `./target/release/mesh-llm models download unsloth/Qwen3.6-27B-MTP-GGUF:UD-Q4_K_XL`

4. Observe that the downloader reports the manifest as ready, then fails with:

   `No such file or directory (os error 2): read package manifest`

## Suspected Cause

The model deletion path likely removes the package manifest or layer package files, but leaves behind stale state indicating that the manifest/package has already been downloaded.

On the next install, the downloader skips fetching `model-package.json`, marks it as ready, then attempts to read it from disk and fails.

## Suggested Fix

Deletion should fully clean up all layer-package state, including:

- Manifest files
- Layer files
- Shared files
- Download metadata
- Cache/index entries
- Any “ready” markers

The downloader should also validate that “ready” files actually exist before skipping download.

ndizazzo:
Fixes #877.

## Problem
After deleting a layer-package model via `mesh-llm models delete`, re-downloading the same model fails with:
```
Ready meshllm/.../model-package.json
No such file or directory: read package manifest
```

## Root Cause
The delete flow (`delete_model_by_identifier_with_catalog`) calls `collect_delete_paths` which canonicalizes hf_hub cache symlinks (`snapshots/<hash>/file.gguf` → `blobs/<sha256>`) to their blob targets. The blob-deletion loop:
1. Removes the blob file
2. Calls `prune_empty_ancestors` starting from the blob directory

But `blobs/` is never empty (other blobs exist), so the snapshot directory at `snapshots/<hash>/` — which still contains broken symlinks — is never pruned. On re-download, hf_hub cache finds the `CachedRevisionInfo` pointing to the still-present snapshot path, reports "Ready", and the downloader attempts to read the non-existent file.

## Fix
After the blob-deletion loop, iterate the original `resolved_paths` (pre-canonicalization symlink paths), detect broken symlinks (`symlink_metadata().is_ok() && !path.exists()`), remove them, and prune empty ancestors up to the hf cache root. This ensures the snapshot directory is properly cleaned so hf_hub initiates a fresh download on re-request.

## Verification
- `cargo check -p model-hf` passes
- `cargo test -p model-hf --lib` — all 34 tests pass
- `cargo test -p model-resolver --lib` — all 29 tests pass

## Release Notes

* **Bug Fixes**
  * Improved model deletion cleanup by detecting and removing broken Hugging Face Hub cache symlinks, and pruning resulting empty cache directories to prevent stale cache entries and reduce follow-up file-not-found errors.

### DIFF
--- a/crates/model-hf/src/store/delete.rs
+++ b/crates/model-hf/src/store/delete.rs
@@ -439,10 +439,11 @@ pub async fn delete_model_by_identifier_with_catalog(
         // symlink_metadata succeeds for real files AND broken symlinks;
         // exists() returns false for broken symlinks. Combined they identify
         // symlinks whose blob target no longer exists.
-        if std::fs::symlink_metadata(path).is_ok() && !path.exists() {
-            if std::fs::remove_file(path).is_ok() {
-                prune_empty_ancestors(path, &hf_cache_root);
-            }
+        if std::fs::symlink_metadata(path).is_ok()
+            && !path.exists()
+            && std::fs::remove_file(path).is_ok()
+        {
+            prune_empty_ancestors(path, &hf_cache_root);
         }
     }
 
--- a/crates/model-hf/src/store/delete.rs
+++ b/crates/model-hf/src/store/delete.rs
@@ -439,11 +439,21 @@ pub async fn delete_model_by_identifier_with_catalog(
         // symlink_metadata succeeds for real files AND broken symlinks;
         // exists() returns false for broken symlinks. Combined they identify
         // symlinks whose blob target no longer exists.
-        if std::fs::symlink_metadata(path).is_ok()
-            && !path.exists()
-            && std::fs::remove_file(path).is_ok()
-        {
-            prune_empty_ancestors(path, &hf_cache_root);
+        if std::fs::symlink_metadata(path).is_ok() && !path.exists() {
+            match std::fs::remove_file(path) {
+                Ok(()) => {
+                    prune_empty_ancestors(path, &hf_cache_root);
+                }
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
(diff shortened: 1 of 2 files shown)


michaelneale (approved):
if you can get build clean

ndizazzo:
Just a transient ci failure

ndizazzo:
@michaelneale not at my terminal? Can you rekick

PR #913, opened 2026-06-28
Add split overrides and reusable fast binaries

luxiano3990:
Incremental build-oriented patch: adds runtime override knobs for split admission/capacity and reusable binary caching in the fast GitHub workflow.


ndizazzo:
I’m trying to understand the shape of this PR before reviewing the details, because the stated scope sounds small but the diff is quite large.

My read is that the branch is pretty far out of date. One of the two commits is `v0.71.0: prepare release source`, which makes this look like it was carried forward from an older release/source snapshot rather than built incrementally on today’s `main`.

Functionally, I see several separate ideas mixed together:

- split/capacity override knobs via `MESH_LLM_FORCE_*`
- CPU-stage capacity advertisement through gossip/protobuf fields
- split readiness and participant-selection changes for package-backed split serving
- layer-package materialization/runtime-load fixes
- a reusable Mac/Windows fast-binary workflow
- fleet helper scripts/docs
- a separate Flushnet webchat/tool-loop integration

So I think this was opened to preserve your working local/fleet setup for package-backed split serving across a Mac seed and Windows/CPU workers, plus faster binaries to make that setup easier to reproduce. That goal makes sense, but this PR combines too many unrelated changes to review safely.

It's also kinda huge, inflated by things that do not look like core changes:

- `crates/mesh-llm-host-runtime/src/mesh/mod.rs.before-cpu-stage-vram-fallback` adds 10,329 lines and looks like an accidental backup file.
- `crates/mesh-llm-ui/package-lock.json` accounts for 6,904 changed lines, while this repo’s UI build path appears to use `pnpm-lock.yaml`.
- The Flushnet tool loop is another 630 lines and seems unrelated to split admission/capacity or fast CI binaries.

If you think these changes should be in our codebase, please break this into separate PRs for small individual features. 

For example:

- CPU-stage capacity gossip/protobuf + split readiness
- layer-package/runtime-load fixes
- fast reusable binary workflow
- fleet scripts/docs, if these belong in the repo
- Flushnet webchat/tool-loop integration as its own product/security/UI review

That would make each change reviewable on its own merits and avoid hiding unrelated behavior inside one large draft.

PR #917, opened 2026-06-29
fix: validate release native runtimes from explicit matrix

ndizazzo:
## Summary
- teach the release native-runtime validator to accept an explicit required target matrix
- update the release workflow to validate only the native runtime lanes the workflow actually builds
- keep the old asset-inference validator behavior available for direct CLI use

## Validation
- `python3 -m unittest scripts.tests.test_validate_release_native_runtime_matrix`
- `git diff --check`
- `actionlint .github/workflows/release.yml`
- simulated the failed RC5 release manifest step with representative release artifacts: legacy asset inference reproduced the missing linux/windows native-runtime errors, while the explicit release matrix passed`

## Summary by CodeRabbit

* **New Features**
  * Native runtime validation now accepts an explicitly provided set of required targets (CPU and optional GPU/CUDA targets), and can validate using those targets instead of inferring from assets.
  * The release workflow now respects the GPU-bundle skip setting when running native runtime validation.

* **Bug Fixes**
  * Improved detection of missing native runtime entries when required targets are explicitly specified.
  * Invalid required-target labels now produce clearer errors and a nonzero exit code.

* **Tests**
  * Expanded coverage for explicit required-target behavior and CLI argument handling.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -943,14 +943,30 @@ jobs:
       - name: Generate native runtime release manifest
         env:
           RELEASE_TAG: ${{ needs.metadata.outputs.tag }}
+          SKIP_GPU_BUNDLES: ${{ needs.metadata.outputs.skip_gpu_bundles }}
         run: |
           scripts/generate-native-runtime-release-manifest.sh \
             --tag "$RELEASE_TAG" \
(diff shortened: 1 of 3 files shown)


michaelneale (approved):
nice one

PR #918, opened 2026-06-29
fix: RC5 release readiness corrections

ndizazzo:
## Summary

This PR bundles the RC5 release-readiness fixes found across the open testing threads:

- closes the native-runtime matrix gaps exposed by clean `runtime install` testing;
- makes release packaging/signing metadata more explicit and verifiable;
- hardens the RC smoke script around real published artifacts and native-runtime selection;
- reshapes `doctor --json` GPU diagnostics so detailed `/proc` probe data remains available without being reported as fake GPU rows.

### mesh1 RC4 readiness thread

The earlier mesh1 audit showed that a published bundle could serve a model, but a clean `mesh-llm runtime install` path could still fail because release native-runtime artifacts were missing for the host. It also showed that `doctor --json` was not enough as a readiness gate when the selected runtime could be absent or wrong.

This PR carries that lesson into the release workflow by validating explicit native-runtime targets and making the smoke path check the runtime-install surface separately from bundle execution.

### carrack RC5 release audit

Carrack detected `cpu, cuda, vulkan`, but RC5’s published `native-runtimes.json` only offered an x86_64 CPU runtime for that host. The CUDA runtimes present in the manifest were Linux aarch64 targets, so they were correctly rejected as `arch_mismatch`; the resolver then selected CPU.

This PR adds native-runtime jobs for the missing published-platform targets, including the carrack-critical Linux x86_64 CUDA 12/13 runtimes, plus Linux x86_64 ROCm/Vulkan and Windows x86_64 CPU/CUDA/ROCm/Vulkan runtimes.

### release matrix / Actions artifact audit

The RC5 matrix published binary bundles for more platform/backend combinations than it published native-runtime artifacts for. The release validator did not require those runtime artifacts, so the gap reached the prerelease manifest.

This PR updates the workflow and matrix validation so `native-runtimes.json` generation is driven by the intended explicit target list rather than accidentally passing with only a subset of runtime artifacts.

### carrack doctor GPU diagnostics thread

On carrack, `doctor --json` previously reported 25 `host.gpus` rows. Many of those rows were useful raw NVIDIA probe details such as DMA masks, IRQs, BIOS versions, bus locations, and device minors, but they were incorrectly shaped as separate GPUs with `detected-*` stable IDs.

This PR preserves the useful probe information by nesting it under the matching real GPU entry. `host.gpus` now reports the two actual carrack devices, each with a `probe` object containing the `/proc/driver/nvidia/gpus/*/information` fields and raw lines.

## What Changed

- Added Linux x86_64 CUDA 12/13 native-runtime release jobs.
- Added Linux x86_64 ROCm and Vulkan native-runtime release jobs.
- Added Windows x86_64 CPU/CUDA/ROCm/Vulkan native-runtime release jobs.
- Updated required native-runtime targets and validation tests so missing runtime artifacts fail release validation.
- Updated native-runtime packaging verification and smoke scripts around manifest/runtime-install expectations.
- Added signing/attestation release notes and packaging metadata support.
- Added `HostGpuProbe` and nested NVIDIA `/proc` probe details under `HostGpuProfile`.
- Added a carrack-shaped unit test proving NVIDIA proc details are nested under two real GPUs rather than emitted as 25 top-level GPU rows.

## Validation

Local validation run before pushing:

```text
cargo test -p mesh-llm-hardware-profile --lib
cargo test -p mesh-llm-native-runtime --lib
cargo check -p mesh-llm-native-runtime
cargo check -p mesh-llm-hardware-profile
cargo check -p mesh-llm
cargo clippy -p mesh-llm-native-runtime --all-targets -- -D warnings
cargo clippy -p mesh-llm-hardware-profile --all-targets -- -D warnings
cargo clippy -p mesh-llm --all-targets -- -D warnings
cargo fmt --all --check
```

Carrack validation after pulling this branch:

```text
cd /home/ndizazzo/dev/mesh/mesh-llm
just build
cargo test -p mesh-llm-hardware-profile --lib nvidia_proc_details_are_nested_under_matching_gpus
```

Carrack doctor verification from `target/debug/mesh-llm doctor --json`:

```text
gpu_count: 2
NVIDIA GeForce RTX 5090 -> CUDA0, probe source linux_nvidia_proc, IRQ 16, DMA Mask 0xfffffffffffff, Bus Location 0000:01:00.0
NVIDIA GeForce RTX 3080 -> CUDA1, probe source linux_nvidia_proc, IRQ 184, DMA Mask 0x7fffffffffff, Bus Location 0000:06:00.0
bad top-level DMA/IRQ/Bus Location/Video BIOS rows: none
```

Saved carrack artifacts:

```text
/tmp/mesh-llm-doctor-gpu-fix-build.log
/tmp/mesh-llm-doctor-gpu-fix.json
```

Note: the carrack doctor run still reports `healthy: false` in this branch checkout because that local cache does not currently have a compatible native runtime selected. The GPU diagnostics shape is fixed independently and verified from the built binary.

## Summary by CodeRabbit

* **New Features**
  * Expanded native runtime release coverage with additional Linux x86_64 (CUDA/ROCm/Vulkan) and Windows x86_64 (CPU plus CUDA/ROCm/Vulkan) builds.
  * Added release attestation signing keys and conditional attestation stamping support for packaged executables.
* **Bug Fixes**
  * Improved GPU hardware discovery by combining NVIDIA probe results with fallback GPU profiles without losing non-NVIDIA devices.
  * Added stricter, time-bounded release smoke cleanup/verification to ensure no lingering bundle processes remain.
* **Documentation**
  * Documented how to generate, store, and verify attestation signing keys in RELEASE.md.
* **Chores / Tests**
  * Hardened release workflow and packaging scripts (checkout credential handling, Python selection, manifest/architecture parsing, expanded matrix validation tests).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -51,12 +51,10 @@ Set or rotate the repository secrets from the generated files with:
 
 ```bash
 gh secret set MESH_RELEASE_ATTESTATION_SIGNING_KEY_FILE \
-  --repo Mesh-LLM/mesh-llm \
   --app actions \
   < /tmp/mesh-release-attestation/mesh-release-attestation-private-key.json
 
 gh secret set MESH_RELEASE_ATTESTATION_PUBLIC_KEY_FILE \
-  --repo Mesh-LLM/mesh-llm \
   --app actions \
   < /tmp/mesh-release-attestation/mesh-release-attestation-public-key.json
 ```
@@ -72,7 +70,8 @@ cargo run -p xtask -- release-attestation inspect \
 ```
 
 The reported status must be `valid`. A `missing` status means the bundle was
-published without an embedded release-attestation footer.
+published without an embedded release-attestation footer. An `invalid` status
+means a footer was present, but signature verification failed.
 
 ## Build
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -23,6 +23,57 @@ locally.
 - Node/npm installed for the UI build
 - `gh` CLI authenticated if publishing manually
 
+## Release Attestation Signing Keys
+
+The GitHub Actions release workflow stamps packaged `mesh-llm` executables when
+these repository Actions secrets are present:
+
+- `MESH_RELEASE_ATTESTATION_SIGNING_KEY_FILE`
+- `MESH_RELEASE_ATTESTATION_PUBLIC_KEY_FILE`
+
+The secret values are the full JSON contents of the release-attestation private
+and public key files, not paths to files. Generate a production keypair with:
+
+```bash
+umask 077
+mkdir -p /tmp/mesh-release-attestation
+cargo run -q -p xtask -- release-attestation generate-keypair \
+  --private-key-out /tmp/mesh-release-attestation/mesh-release-attestation-private-key.json \
+  --public-key-out /tmp/mesh-release-attestation/mesh-release-attestation-public-key.json
+```
+
+Store the keypair in 1Password before adding or rotating GitHub secrets. The
+production release-attestation keypair lives in the `mesh-llm` vault as
+`GitHub Actions Release Attestation Signing Keys`, with fields named exactly
+after the GitHub Actions secrets above.
+
+Set or rotate the repository secrets from the generated files with:
+
+```bash
+gh secret set MESH_RELEASE_ATTESTATION_SIGNING_KEY_FILE \
+  --repo Mesh-LLM/mesh-llm \
+  --app actions \
+  < /tmp/mesh-release-attestation/mesh-release-attestation-private-key.json
+
+gh secret set MESH_RELEASE_ATTESTATION_PUBLIC_KEY_FILE \
+  --repo Mesh-LLM/mesh-llm \
+  --app actions \
+  < /tmp/mesh-release-attestation/mesh-release-attestation-public-key.json
+```
+
+After publishing, verify at least one packaged release archive by extracting it
+and running:
+
+```bash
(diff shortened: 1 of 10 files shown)


RELEASE.md
```diff
@@ -23,6 +23,57 @@ locally.
...
+```bash
+umask 077
+mkdir -p /tmp/mesh-release-attestation
+cargo run -q -p xtask -- release-attestation generate-keypair \
+  --private-key-out /tmp/mesh-release-attestation/mesh-release-attestation-private-key.json \
+  --public-key-out /tmp/mesh-release-attestation/mesh-release-attestation-public-key.json
+```
+
+Store the keypair in 1Password before adding or rotating GitHub secrets. The
+production release-attestation keypair lives in the `mesh-llm` vault as
```
michaelneale, line 46: is that a shared vault?

michaelneale (approved):
nice - just one question about 1password and shared vault

RELEASE.md
ndizazzo: local on my machine for now, can share with you guys

ndizazzo: Or - if you want to create a shared vault, I can move the secret to it

michaelneale: depends if we need access to it - or just recreate it when needed?

ndizazzo: Yeah, good callout. It's not really needed for referencing outside of the build process, it can just be recreated/replaced and the next RC / release builds get signed with that instead.

PR #934, opened 2026-06-30
chore: synchronize release version management

ndizazzo:
## Summary
- sync mesh-llm release/version surfaces to the current GitHub release, v0.72.1
- add `just release <version>` as a synchronous release-prep and publish command
- expand `scripts/release-version.sh` so future bumps update Rust crates, lockfile entries, SDK/package manifests, docs, config schema metadata, and website source examples together
- add fail-closed release preflight checks for invalid/non-advancing versions, existing tags/releases, dirty worktrees, and non-current `main`

## Rationale
The previous release workflow accepted a version input for build jobs, but it did not reliably produce and build from a source commit that contained the corresponding version bump. That made the repository drift from published releases and left multiple public surfaces stuck on older versions.

This PR makes the release commit the first-class source of truth. `just release <version>` prepares the versioned commit on `main`, pushes it, dispatches the GitHub release workflow against that commit, watches the run via `gh`, and then attaches generated release notes to the GitHub release after a successful non-canary build. That avoids release branches and keeps the release build tied to the commit containing the version update.

The release command also fails closed before it can publish an ambiguous release: the target version must parse as semver, advance from the workspace version, and not already exist as a local tag, remote tag, or GitHub release. This supports RC progression such as `0.72.0-rc1` to `0.72.0-rc3`, while preventing accidental rebuilds of an already released version.

## Validation
- `just test-all`
- `scripts/release-version.sh 0.72.1`
- `bash -n scripts/release.sh scripts/release-version.sh`
- `cargo metadata --format-version 1`
- `just check-release`
- `git diff --check`

Note: the first full `just test-all` attempt exposed the expected config schema snapshot drift after syncing the built-in plugin version, and Playwright Chromium needed to be installed for the updated Playwright version. After updating the snapshot and running `pnpm exec playwright install chromium`, the final full `just test-all` passed.

## Summary by CodeRabbit

* **New Features**
  * Added a new release helper command to streamline preparing and publishing GitHub releases.

* **Documentation**
  * Updated SDK, CLI, and runtime documentation/examples to reference release version `0.72.1`.
  * Refreshed install snippets across Rust, Node.js, Swift, Kotlin, and native runtime docs.

* **Chores**
  * Bumped workspace and package versions to `0.72.1` across the project.
  * Updated the supported version list and UI placeholders to align with the latest release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3613,14 +3613,14 @@ checksum = "11d3d7f243d5c5a8b9bb5d6dd2b1602c0cb0b9db1621bafc7ed66e35ff9fe092"
 
 [[package]]
 name = "llama-quant-ffi"
-version = "0.68.0"
+version = "0.72.1"
 dependencies = [
  "libloading",
 ]
 
 [[package]]
 name = "llama-spec-bench"
-version = "0.68.0"
+version = "0.72.1"
 dependencies = [
  "anyhow",
  "clap",
@@ -3756,7 +3756,7 @@ dependencies = [
 
 [[package]]
 name = "mesh-llm"
-version = "0.68.0"
+version = "0.72.1"
 dependencies = [
  "anyhow",
  "axum",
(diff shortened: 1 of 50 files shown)


michaelneale (approved):
one coderabbit comment about missing version @ndizazzo

PR #939, opened 2026-07-01
Make managed relay defaults tolerate alias cert flips

michaelneale:
🤖 This PR was opened by my AI agent.

## Summary
- add both managed relay aliases per region to the default relay list
- keep TLS verification strict while giving iroh a cert-valid candidate when the paired alias has the wrong SAN
- update the archived relay reference to document the four managed relay aliases

## Verification
- live release-binary simulation with all four relay URLs selected `https://aps1-2.relay.michaelneale.mesh-llm.iroh.link./` and logged `Relay connected`
- `cargo test -p mesh-llm-host-runtime relay_policy_tests::default_policy_uses_managed_relays_when_no_urls_are_given --lib`

## Notes
- This preserves compatibility with the existing `usw1-2` / `aps1-1` defaults while also working with the currently served `usw1-1` / `aps1-2` certs.
- Existing released clients still need the server-side cert/SAN issue fixed; this PR protects new clients.

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded the default relay fallback set for `DefaultPublic` so more region-specific relay endpoints are tried when none are provided.
* **Documentation**
  * Updated the Fly relay reference with the expanded default region aliases and clarified where these defaults are defined.
  * Noted that paired relay aliases are provided to remain usable across relay certificate rotations.
* **Tests / CI**
  * Strengthened the client auto-join CI flow to validate join success via `/api/status` peer presence and improved cleanup/logging on failure.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/mod.rs
@@ -598,7 +598,9 @@ fn effective_relay_urls(policy: RelayPolicy, relay_urls: &[String]) -> Vec<Strin
     match policy {
         RelayPolicy::Disabled | RelayPolicy::ExplicitlyDisabled => Vec::new(),
         RelayPolicy::DefaultPublic if relay_urls.is_empty() => vec![
+            "https://usw1-1.relay.michaelneale.mesh-llm.iroh.link./".into(),
             "https://usw1-2.relay.michaelneale.mesh-llm.iroh.link./".into(),
+            "https://aps1-2.relay.michaelneale.mesh-llm.iroh.link./".into(),
             "https://aps1-1.relay.michaelneale.mesh-llm.iroh.link./".into(),
         ],
         RelayPolicy::DefaultPublic => relay_urls.to_vec(),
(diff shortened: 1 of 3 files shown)


i386:
This is a good mitigation for the alias/cert mismatch, but it does not fully explain or fix the current `Linux client-auto boot test` failure.

The PR's own client-auto check is still failing: https://github.com/Mesh-LLM/mesh-llm/actions/runs/28498453850/job/84470419463

From that log, the new binary does get as far as public discovery and tries the advertised candidates:

- `discovered mesh unnamed (2 peer(s))`
- `Racing 2 auto-join bootstrap candidates`
- `No auto-join candidate completed the fast probe; falling back to serial join`
- `Failed to join any peer — running standalone (Timeout connecting to 55eb35e7ce (30s))`

So adding both aliases helps new clients tolerate the currently flipped relay certs, but this CI check is still coupled to live public mesh state: stale Nostr listings, old peers with old/default relay behavior, dead advertised addresses, or GitHub-runner-to-public-peer connectivity can still make it fail.

I think this PR is still useful for new clients, while the CI fix should be separate: convert `ci-client-auto-test.sh` to a local/direct no-relay join smoke so it verifies `client --auto` boot/join behavior without depending on public Nostr or managed relay health.

i386:
Updated this PR with the CI-side fix we discussed: `scripts/ci-client-auto-test.sh` now creates a local seed client mesh with mDNS/LAN relay policy and has the tested `client --auto` join it via a direct loopback invite token. This keeps the boot/join regression coverage but removes dependency on public Nostr listings or managed relay health. Commit: `54f346b1`.

PR #946, opened 2026-07-02
Server-side tool-call emulation for models without native tool support

### LINKED ISSUE #944: Serving-side tool-call emulation for models without native tool support (port of goose local-inference tool shim)
## Problem

Small / non-tool-trained models served via the mesh /v1 endpoint handle `tools` poorly: the chat template either ignores the schemas or the model loops, re-issuing the same call. goose already solved this for its **local-inference** provider (text-convention emulation + parser), but that path is bypassed when goose talks to a mesh — and every other OpenAI client hits the same wall.

The serving node is the right place to fix it: it is the only party that knows the actual **chat template capability** of the loaded model. Clients only know a model name.

## Proposal

When a /v1/chat/completions request carries `tools` for a model whose chat template does **not** support native tool calling:

1. **Request adaptation**: strip `tools`/`tool_choice` from template input; inject a compact instruction into the system message — tool name + description + **compact parameter schema** (property names/types), and the convention: emit `TOOL_CALL {"name": ..., "arguments": {...}}` on its own line. Rewrite conversation history so the template never sees tool roles (assistant `tool_calls` → text; `role:"tool"` → user "Tool result:" text).
2. **Response parsing**: scan output for `TOOL_CALL` lines → real OpenAI `tool_calls` + `finish_reason: "tool_calls"`. Tolerant of prose and `<think>` blocks. Streaming: hold back partial-marker text (goose's HOLD_BACK pattern), emit standard streaming deltas.
3. **Detection**: gate on template capability, not model size. llama.cpp's template application already reports this — goose checks `parse_tool_calls && parser non-empty` on the template result (`crates/goose/src/providers/local_inference/llamacpp/mod.rs`, `template_result_supports_native_tool_calling`). The same signal should be reachable where `apply_chat_template_json` returns `metadata_json` (skippy-runtime). Tool-capable models: **zero behavior change**.

## Suggested seam

`crates/skippy-server/src/frontend/prompting.rs` is the single point where `request.tools` is serialized into `tools_json` for the template. Emulated calls should ride the same downstream assembly as native ones (`call_mesh_*` ids in `mesh-llm-host-runtime/src/network/openai/transport.rs` + `tool_call_ids.rs`) so ids/streaming behave identically.

## Reference implementations

- **goose local-inference**: `crates/goose/src/providers/local_inference/tool_emulation.rs`, `tool_parsing.rs` (`compact_tools_json`), `prompts/tiny_model_system.md` — the proven prompt + parser for tiny models.
- **Client-side prototype (working, tested)**: `mesh-console` `src-tauri/src/toolshim.rs` — the full request/response transform incl. SSE replay, 7 unit tests, verified live end-to-end (0.6B emitted a parseable `tool_calls` through it).

## Lessons from the prototype (live-tested)

- **Include compact parameter schemas** in the instruction. Name+description alone made Qwen3-0.6B hallucinate argument names (`{"path"}` instead of `{"command"}`).
- **Don't gate on model size.** Qwen3-0.6B handles *native* tool calls fine with a lean toolset — its template is tool-trained. A size heuristic regresses such models; template capability is the correct trigger.
- Multi-turn history rewrite is required or the template errors/mangles on tool roles.

## Why here and not goose

Emulation-at-the-server fixes every client (goose, curl, any OpenAI SDK) at the one place the template is known — same pattern as llama.cpp-server style server-side tool handling. goose-side adaptation (leaner prompts for weak models) remains complementary but can't know template capability remotely.

michaelneale:
Closes #944.

## What you can now do

Point any OpenAI client — goose, curl, an SDK — at a mesh model that is **not**
tool-trained (e.g. a small model whose chat template ignores `tools`), pass the
standard `tools` field, and get back real `tool_calls` with
`finish_reason: "tool_calls"`. Previously those models ignored the schemas or
looped re-issuing the same call.

The serving node is the only party that knows the loaded model's actual
chat-template capability, so it emulates tool calling when — and only when — the
template can't do it natively. Tool-capable models (including lean ones like
Qwen3-0.6B with a small toolset) keep native tool calling and see **zero
behavior change**.

## How it works

1. **Detection** — gated on template capability, not model size. When the chat
   template is applied, the staged runtime returns `metadata_json`; native
   support means `parse_tool_calls == true` **and** a non-empty `chat_parser`
   (the same signal goose checks).
2. **Request adaptation** — for a non-tool-capable template with `tools`, the
   prompt is re-rendered: `tools`/`tool_choice` stripped, a compact instruction
   injected (tool name + description + **compact parameter schema** with `?` for
   optional args), and history rewritten so the template never sees tool roles
   (assistant `tool_calls` → `TOOL_CALL {json}` text; `role:"tool"` → user
   "Tool result:" text). The compact schema matters: the live prototype showed
   name+description alone made a 0.6B model hallucinate argument names.
3. **Response parsing** — output is scanned for `TOOL_CALL {json}` lines →
   OpenAI `tool_calls`, tolerant of surrounding prose and `<think>` blocks.
   Streaming holds back the trailing incomplete line and withholds calls until
   finalization, matching native tool-call streaming semantics. Emulated calls
   ride the same downstream assembly as native ones, so `call_mesh_*` ids and
   streaming behave identically.

Ported from goose's local-inference provider
(`crates/goose-local-inference/src/tool_emulation.rs`, `tool_parsing.rs`,
`prompts/tiny_model_system.md`).

## Architecture

- New self-contained `crates/skippy-server/src/frontend/tool_emulation.rs`
  (detection, request adaptation, response parsing).
- `frontend/prompting.rs` wires it into `prepare_chat_prompt` (adaptation) and
  `parse_chat_output` (parsing), with the streaming-aware
  `parse_emulated_chat_output` bridge.
- Detection needs no new state threading: emulation renders the prompt without
  tools, so the flowing `metadata_json` naturally reports non-native and the
  parse path branches on the same signal.
- New spec: `docs/specs/server-side-tool-call-emulation.md`.

## Validation

- `cargo test -p skippy-server --lib` — 183 passed (19 new unit tests in
  `tool_emulation`, 5 new integration tests in `frontend::tests`).
- `cargo clippy -p skippy-server --all-targets -- -D warnings` — clean.
- `cargo fmt --all --check` — clean.
- `cargo check -p mesh-llm` — clean (reachable from the shipped binary).

Tests cover: capability detection (both signals required), compact schema with
required/optional flags, single/multiple/multi-turn calls, `<think>`-block
tolerance, disallowed-name filtering, malformed JSON treated as prose, history
rewrite (system merge/insert, tool→user, assistant tool_calls→text), streaming
partial withholding, and `parallel_tool_calls:false` truncation.

## Summary by CodeRabbit

* **New Features**
  * Added server-side tool-call emulation for chat completions when the selected chat template doesn’t natively support tools.
  * Injects emulation guidance and rewrites prior tool interactions so models can return tool calls in a compatible format.
* **Bug Fixes**
  * Improved parsing of emulated tool calls (including tolerant handling for partial/streaming output) and converts them into OpenAI-compatible tool calls.
  * Enables early streaming stop as soon as a complete emulated tool call is produced.
  * Enforces tool allowlisting and single-call behavior when parallel tool calls are disabled.
* **Documentation**
  * Added a specification for server-side tool-call emulation behavior and response handling.

### DIFF
--- a/crates/skippy-server/src/frontend.rs
+++ b/crates/skippy-server/src/frontend.rs
@@ -1779,6 +1779,7 @@ struct SplitMultimodalGeneration<'a> {
     downstream_wire_condition: WireCondition,
     lane_pool: Arc<PersistentStageLanePool>,
     prediction_return: Option<PredictionReturnReceiver>,
+    emulation_active: bool,
 }
 
 struct EmbeddedLocalOutput {
--- a/crates/skippy-server/src/frontend.rs
+++ b/crates/skippy-server/src/frontend.rs
@@ -82,7 +82,11 @@ mod prefix_cache;
 mod prompting;
 mod request;
 mod speculative;
+mod tool_emulation;
 mod util;
+
+#[cfg(test)]
+use prompting::parse_emulated_chat_output;
 mod wire_messages;
 
 use self::{
--- a/crates/skippy-server/src/frontend.rs
+++ b/crates/skippy-server/src/frontend.rs
@@ -1994,6 +1994,27 @@ enum TokenControl {
     Stop,
 }
 
+/// Whether generation for this request is running under tool-call emulation:
+/// the request carries tools and the rendered prompt metadata does not report
+/// native tool support (empty grammar triggers). Used to enable early
+/// generation stop once a complete emulated tool call has been produced.
+fn emulation_generation_active(
+    hook_request: Option<&ChatCompletionRequest>,
+    prompt: &PreparedGenerationPrompt,
+) -> bool {
+    let Some(request) = hook_request else {
+        return false;
+    };
+    if !tool_calls_requested(request) {
+        return false;
+    }
+    prompt
+        .chat_parse_metadata
+        .as_deref()
+        .map(|metadata| !tool_emulation::template_supports_native_tool_calls(metadata))
+        .unwrap_or(false)
+}
+
 struct TextGenerationCollector<'a, F>
 where
     F: FnMut(&str) -> OpenAiResult<()>,
@@ -2008,6 +2029,7 @@ where
     completion_tokens: usize,
     finish_reason: FinishReason,
     metrics: GenerationMetrics,
+    emulation_active: bool,
 }
 
 impl<'a, F> TextGenerationCollector<'a, F>
@@ -2031,7 +2053,17 @@ where
             completion_tokens: 0,
             finish_reason: finish_reason_for_generation(true),
             metrics: GenerationMetrics::default(),
+            emulation_active: false,
         }
+    }
+
+    /// Enables early generation stop once a complete emulated tool call is
+    /// generated. Mirrors goose's `tool_call_emitted -> Stop` so the model does
+    /// not ramble after emitting a call. Only active for emulated tools
+    /// requests; native requests are unaffected.
+    fn with_emulation_stop(mut self, emulation_active: bool) -> Self {
+        self.emulation_active = emulation_active;
+        self
     }
 
     fn push_token(&mut self, token: i32) -> OpenAiResult<TokenControl> {
@@ -2066,6 +2098,11 @@ where
             .any(|stop| !stop.is_empty() && self.text.contains(stop))
         {
             self.text = trim_at_stop(&self.text, &self.stop_values).to_string();
+            self.emit_safe_delta(true)?;
+            self.finish_reason = finish_reason_for_generation(false);
+            return Ok(TokenControl::Stop);
+        }
+        if self.emulation_active && tool_emulation::emulated_tool_call_complete(&self.text) {
             self.emit_safe_delta(true)?;
             self.finish_reason = finish_reason_for_generation(false);
             return Ok(TokenControl::Stop);
--- a/crates/skippy-server/src/frontend/generation_flow.rs
+++ b/crates/skippy-server/src/frontend/generation_flow.rs
@@ -227,6 +227,7 @@ impl StageOpenAiBackend {
                     .map(|hub| hub.register(ids.request_id, ids.session_id))
                     .transpose()
                     .map_err(openai_backend_error)?;
+                let emulation_active = emulation_generation_active(hook_request.as_ref(), &prompt);
                 return self.generate_split_multimodal_text(
                     SplitMultimodalGeneration {
                         prompt,
@@ -241,6 +242,7 @@ impl StageOpenAiBackend {
                         downstream_wire_condition,
                         lane_pool,
                         prediction_return,
+                        emulation_active,
                     },
                     on_text_chunk,
                 );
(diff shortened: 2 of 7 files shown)


michaelneale:
## Live spot check (local, real model)

Verified end-to-end against a real non-tool-trained model served through the mesh — `Qwen/Qwen2.5-0.5B-Instruct-GGUF:q4_k_m` on `mesh-llm serve` — driving real `/v1/chat/completions` requests:

- **Tools request** → `finish_reason: "tool_calls"` with a valid `shell` call (`{"command":"ls"}`) and a `call_mesh_*` id. ✅
- **No-tools control** → normal prose (`finish_reason: "stop"`, no tool_calls). Native path untouched. ✅
- **Multi-turn with a prior tool result** → history-rewrite path (assistant `tool_calls`→text, `role:"tool"`→user) worked with no template mangling; model produced a valid follow-up call. ✅
- **Streaming** → proper `delta.tool_calls` with `index`, then `finish_reason: "tool_calls"`. ✅

## CI coverage

The existing `scripts/skippy-ci-smoke.sh` already sends a `tools` request to **SmolLM2-135M**, whose template is not tool-capable — so it was *already* exercising this new emulation path, but only asserted HTTP 200 + `role==assistant`. This branch strengthens that row to assert any emulated `tool_calls` are well-formed (non-empty name + JSON-object arguments). A tool call is not *forced* (a 135M model can't reliably emit one without flakiness), but malformed emulated output now fails the smoke.

michaelneale:
## Correction: detection signal fixed (grammar_triggers)

While verifying your question ("do real-tool-calling models skip this path?") I found a real bug in the first version: the ported goose heuristic — `parse_tool_calls && non-empty chat_parser` — **does not work against mesh-llm's patched llama.cpp**. There, `parse_tool_calls` is `true` for *every* tools request and `chat_parser` is always a non-empty PEG structure, so the check was always true and **emulation never fired**.

Fixed to detect native support from `grammar_triggers`: a tool-capable jinja template yields a tool-call grammar trigger (e.g. `<tool_call>`) when applied with tools; a template with no native tool support yields an empty list.

Verified live with a temporary branch trace:

| Model | Template | `native_supported` | Path |
|---|---|---|---|
| SmolLM2-135M | no tool grammar | `false` | **emulation** |
| Qwen2.5-0.5B-Instruct | `<tool_call>` trigger | `true` | native (unchanged) |
| Qwen3.5-0.8B | `<tool_call>` trigger | `true` | native `get_weather` call ✅ |

So the answer to your question is confirmed with the corrected detector: **tool-capable templates do not trigger emulation.** (Note: my earlier "spot check" comment mis-attributed Qwen2.5-0.5B as emulated — it was actually native. This commit corrects the detection so the routing is now right.)

Trace removed; unit test updated; clippy/fmt/tests clean.

PR #947, opened 2026-07-03
Pass reasoning format through Skippy chat templates

michaelneale:
🤖 Opened by AI agent.

Summary:
- Pass Skippy chat-template reasoning_format into llama.cpp's established reasoning parser via the C ABI.
- Default embedded OpenAI chat templating to hidden reasoning parsing so Gemma 4 thought channels are removed from content but not exposed as reasoning_content unless configured.
- Add UI fallback coverage for real Gemma channel markers.

Validation:
- scripts/prepare-llama.sh pinned with a clean temp LLAMA_WORKDIR
- just build
- cargo fmt --all --check
- cargo test -p skippy-server --lib
- cargo test -p skippy-runtime --lib
- cargo test -p mesh-llm --lib
- cargo check -p mesh-llm
- cargo clippy -p skippy-server --all-targets -- -D warnings
- cargo clippy -p mesh-llm --all-targets -- -D warnings
- pnpm test -- src/features/chat/components/thinking-segments.test.ts

## Summary by CodeRabbit

* **New Features**
  * Added configurable reasoning/thinking visibility for chat templates, including support for formats that hide or expose reasoning content.
* **Bug Fixes**
  * Streaming and non-streaming parsing now consistently respects the selected reasoning visibility, preventing hidden reasoning from being emitted.
  * Improved Gemma “thought” section splitting to handle additional channel variants and newline-prefixed thought content.
  * Chat template rendering now passes through an explicit reasoning-format setting for more accurate template behavior.

### DIFF
--- a/crates/mesh-llm-ui/src/features/chat/components/thinking-segments.test.ts
+++ b/crates/mesh-llm-ui/src/features/chat/components/thinking-segments.test.ts
@@ -24,6 +24,15 @@ describe('splitAssistantThinking', () => {
 
   it('splits Gemma channel thinking from final response text', () => {
     expect(
+      splitAssistantThinking('<|channel>thought\nCheck whether the prompt is a test.<channel|>Test received!')
+    ).toEqual([
+      { kind: 'thinking', text: '\nCheck whether the prompt is a test.', open: false },
+      { kind: 'response', text: 'Test received!' }
+    ])
+  })
+
+  it('splits legacy Gemma channel thinking from final response text', () => {
+    expect(
       splitAssistantThinking('<|channel|>thoughtCheck whether the prompt is a test.<channel|>Test received!')
     ).toEqual([
       { kind: 'thinking', text: 'Check whether the prompt is a test.', open: false },
--- a/crates/mesh-llm-ui/src/features/chat/components/thinking-segments.ts
+++ b/crates/mesh-llm-ui/src/features/chat/components/thinking-segments.ts
@@ -20,8 +20,8 @@ type TagMatch = {
 
 const THINK_OPEN_TAG = '<think>'
 const THINK_CLOSE_TAG = '</think>'
-const GEMMA_THOUGHT_CHANNEL_TAGS = ['<|channel|>thought', '<channel|>thought']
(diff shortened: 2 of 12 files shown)


michaelneale:
@i386 someone smarter than me needs to review this!

michaelneale:
🤖 Context on why this is larger than Goose:

Goose already has this knob exposed by its llama.cpp Rust binding: it sets `OpenAIChatTemplateParams.reasoning_format` and then calls `apply_chat_template_oaicompat` directly. Examples:
- https://github.com/aaif-goose/goose/blob/838631de5d17ae3d68b70d3a53f7359fe34151e3/crates/goose-local-inference/src/llamacpp/inference_engine.rs#L513-L535
- https://github.com/aaif-goose/goose/blob/838631de5d17ae3d68b70d3a53f7359fe34151e3/crates/goose-local-inference/src/llamacpp/mod.rs#L79-L102

Mesh's Skippy path had the same conceptual requirement, but the C ABI boundary did not expose `reasoning_format` yet. This PR adds that missing ABI/runtime/server plumbing so Mesh uses llama.cpp's established reasoning parser instead of adding model-specific string stripping.

ndizazzo:
> Mesh's Skippy path had the same conceptual requirement, but the C ABI boundary did not expose reasoning_format yet. This PR adds that missing ABI/runtime/server plumbing so Mesh uses llama.cpp's established reasoning parser instead of adding model-specific string stripping.

@michaelneale Yeah, I'd keep this and pass it directly into the template/parser... I think this is where we should pick apart model family markers.

We should still keep skippy-server using the parsed chat metadata, so the server can decide if it wants to parse, hide, or expose reasoning before it leaves the OpenAI backend.

One thing we should do though is update `skippy-ffi` to guard and reject runtimes > 0.1.28 before wrappers can call that function, since the `skippy_apply_chat_template_json` changed signatures.

Some tests that would help here would be to see hidden reasoning markers get stripped out from the content, and that the `reasonsing_content` gets populated

ndizazzo (commented):
see above

michaelneale:
@ndizazzo how does this look now?

ndizazzo (approved):
Looks good now!

PR #952, opened 2026-07-08
Gate passive inference streams behind admission when a trust policy is active

michaelneale:
**Draft.**

`stream_allowed_before_admission` exempts `STREAM_TUNNEL_HTTP` (0x04) and `STREAM_ROUTE_REQUEST` (0x05) from the quarantine gate unconditionally. A caller holding a leaked invite token can therefore route inference through a node even after the trust gate rejected it at gossip (`UntrustedOwner` under `TrustPolicy::Allowlist`) — the token acts as a bearer credential for inference regardless of admission policy.

This change scopes the exemption to `TrustPolicy::Off`, so default open-mesh behavior is unchanged. With `PreferOwned`/`RequireOwned`/`Allowlist`, only `STREAM_GOSSIP` bypasses the gate: if a node enforces who may join, the same enforcement covers who may consume.

Found while wiring membership-derived admission into Buzz (embedded SDK): a three-node harness — allowlist serve, trusted client, stranger reusing the trusted invite token — showed the stranger rejected at gossip but still completing a chat completion via 0x04 (harness: [block/buzz `mesh_admission_smoke`](https://github.com/block/buzz/blob/micn/mesh-upgrade/crates/buzz-relay/examples/mesh_admission_smoke.rs)).

Regression test added: `passive_streams_are_gated_when_trust_policy_active`. Existing gate tests updated to the new signature under `TrustPolicy::Off`. `mesh::tests` suite passes (one unrelated flake, `owner_control_client_reuses_connection_for_sequential_requests`, passes in isolation before and after).

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened stream admission checks so only gossip traffic always bypasses quarantine; other stream types now wait until the remote peer is admitted when trust protections are enabled.
  * Updated gating behavior for passive and tunnel-related streams to match the active trust policy, improving consistency in connection handling.

* **Tests**
  * Expanded coverage for admission and quarantine rules across trust-policy modes.
  * Added validation for stream access when trust policy is active versus off.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/mod.rs
@@ -2536,21 +2536,33 @@ pub(crate) fn is_peer_admitted(peers: &HashMap<EndpointId, PeerInfo>, id: &Endpo
 }
 
 /// Returns `true` if the given stream type is permitted before a peer has
-/// been admitted through gossip.
+/// been admitted through gossip, under the node's trust policy.
 ///
-/// Only two streams bypass the quarantine gate:
+/// With `TrustPolicy::Off` (the default open-mesh posture), three streams
+/// bypass the quarantine gate:
 /// - `STREAM_GOSSIP (0x01)`: the admission handshake itself.
 /// - `STREAM_ROUTE_REQUEST (0x05)`: passive/client request-only path — caller
 ///   is NEVER promoted to `state.peers`.
 /// - `STREAM_TUNNEL_HTTP (0x04)`: passive SDK inference path for callers that
 ///   have an invite token but should not need a local `/v1` HTTP listener.
 ///
-/// Every other stream — including raw tunnel (0x02) — requires the remote to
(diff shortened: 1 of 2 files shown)


i386:
@michaelneale I pushed `6de939010` to address a review issue in the admission gate.

`PreferOwned` is advisory: `policy_accepts_peer` accepts unattributed peers, and passive clients intentionally may not gossip. Treating every non-`Off` policy as enforcing therefore blocked legitimate passive `STREAM_TUNNEL_HTTP` and `STREAM_ROUTE_REQUEST` traffic under `PreferOwned`.

The fix gates passive streams only for the ownership-enforcing policies (`RequireOwned` and `Allowlist`) and adds regression coverage confirming that `Off` and `PreferOwned` preserve passive access.

Validated with the focused regression test, `cargo fmt --all --check`, `cargo check` for `mesh-llm-host-runtime` and `mesh-llm`, and warning-denying Clippy for both packages. Please review the update when you have a chance.

PR #959, opened 2026-07-10
Batch independent GLM-5.2 sessions for 32.6 aggregate decode tok/s

i386:
## Title

Batch independent GLM-5.2 sessions for 32.6 aggregate decode tok/s

## Original problem

Skippy could execute independent decode sessions in one native llama batch, but the request batchers drained immediately, so near-simultaneous OpenAI requests often missed each other. GLM-DSA also treated a batched Full-layer top-k result as one global sideband and reused graphs using padded KV capacity instead of the live logical KV route. Under concurrent split inference this caused malformed sideband widths, stale graph shapes, and failed requests.

The GLM llama.cpp work had also accumulated as more than 200 incremental patch files, making the final execution shape difficult to review or rebase.

This PR is stacked on `feat/jianyang-glm-52`; it does not target `main` directly. Native MTP and n-gram drafting were disabled for the throughput milestone.

## Diagnostics

Fixed two-node lab topology:

- `micstudio`: layers 0..60, `192.168.0.10:19031`
- `studio54`: layers 60..79, `192.168.0.5:19032`
- 8 lanes, F16 activation wire, 16,384 context, `n_batch=n_ubatch=2304`
- identical checked-in coding prompts through the OpenAI streaming path

| Run | Completion tokens | Errors | End-to-end completion tok/s | Aggregate decode tok/s |
| --- | ---: | ---: | ---: | ---: |
| Serial corpus, depth 1, 64 tokens/request | 512 | 0 | 8.0276 | 9.3601 |
| Serial hot fresh session, natural EOS | 149 | 0 | 7.9596 | 8.6804 |
| Concurrent, depth 8, 64 tokens/request | 512 | 0 | 23.3653 | 31.4688 |
| Concurrent, depth 8, 256 tokens/request | 2,048 | 0 | 29.1644 | 32.6391 |
| Concurrent, depth 8, 512 tokens/request | 4,096 | 0 | 29.3417 | 30.9864 |

The controlled 64-token A/B is a `2.9106x` end-to-end aggregate throughput increase. The decode metric follows the existing Jianyang convention: exclude each request's first token and measure after TTFT. For concurrent runs it uses the conservative common span from the earliest first token to the final completion. The 512-token run therefore sustains `30.9864 aggregate decode tok/s`; the 256-token run peaks at `32.6391`.

The concurrent rows are coding-agent pool results, not single-request throughput. The valid hot serial baseline is `8.6804 decode tok/s`; reaching 30 serial tok/s still requires a `3.46x` execution gain. End-to-end completion throughput includes prefill and remains below 30 tok/s, so both denominators are shown explicitly.

A 512-token serial stability run exposed a deterministic boundary failure when GLM-DSA top-k saturated at 512 keys. The all-KV flash route no longer consumed the top-k tensor locally, while the split stage still exported it for downstream IndexShare, leaving the saved tensor outside the executable graph with no backend buffer. After explicitly retaining that exported tensor, the same run logged sidebands at positions 510, 511, and 512 with widths 511, 512, and 513, completed 512 tokens with zero errors, and left both stages healthy.

A focused 4-layer Full+Shared+Shared+Shared gate showed exact split-sideband parity, 0.2487% hidden relative RMSE, 0.99999691 cosine similarity, and a 1.82x native batch speedup. Sequential prefill chunks `[4, 4]` produced exact top-k widths `[4, 4]`.

The lab measurements were produced by the integrated experiment tree. This isolated branch has replayed and compiled cleanly; a clean-branch lab rerun remains required before marking the PR ready.

## Fix

- Add an opt-in, bounded decode batch collection window through `SKIPPY_DECODE_BATCH_COLLECTION_US`; it defaults to zero, is disabled for one lane, and is capped at 10 ms.
- Preserve GLM-DSA top-k sidebands per request across native batches and split Full-to-Shared execution.
- Retain the exported IndexShare top-k tensor in the executable graph when all-KV flash bypasses it locally.
- Include live DSA KV usage in graph-route invalidation so graph shapes follow logical session history rather than padded cache capacity.
- Add model-backed serial-vs-native-batch, split-chain, and chunked-prefill parity gates.
- Add a repeatable OpenAI coding-corpus benchmark script.
- Collapse the GLM patch tail from 205 micro-patches into five subsystem patches: MTP/staged runtime, DSA sparse attention/IndexShare, compact decode/MoE routing, runtime policy/validation, and MoE/quantization/multi-session tuning.

The complete queue now contains 88 patches. Before the boundary fix, the five consolidated patches produced the exact same llama.cpp source tree as the old tail (`3df10562509bf092df66dd8840e6affca713a9ae`). The intentional graph-liveness fix is folded into the existing IndexShare/sparse-attention subsystem patch; clean replay now produces `93a1e3bbe39b737ce891663b941b84349d5ae24e`.

There is no mesh protocol or Skippy ABI version change in this PR.

## Validation

- [x] Clean patch replay from pinned llama.cpp upstream
- [x] Old and consolidated queues produce the same Git tree hash
- [x] Clean replay with the all-KV sideband fix matches tree `93a1e3bbe39b737ce891663b941b84349d5ae24e`
- [x] `just llama-build` after queue consolidation
- [x] `just build` (patched llama.cpp CPU + Metal, UI, and Rust integration)
- [x] `cargo test -p skippy-server --lib` (223 passed)
- [x] `cargo clippy -p skippy-server --all-targets -- -D warnings`
- [x] `cargo clippy -p skippy-bench --all-targets -- -D warnings`
- [x] `bash -n scripts/glm52-multi-session-openai-bench.sh`
- [x] 7,829 measured completion tokens across the six successful reported conditions with zero request errors
- [x] Serial boundary validation crossed positions 510, 511, and 512 without a missing top-k buffer
- [x] UI changes do not apply
- [ ] Repeat the full two-node benchmark from this isolated branch before moving out of draft

`cargo test -p skippy-bench` compiled and ran 270 tests: 268 passed. Two existing tests in untouched `glm_dsa_report_aggregate.rs` failed due report-path classification expectations (`compact_flash_native` vs `sparse_mask_flash`, and `compact_flash_forced` vs `compact_flash_fallback`).


i386:
reworking this branch

PR #969, opened 2026-07-12
fix: use GGUF file stems as display names for synthetic local-gguf refs

### LINKED ISSUE #882: Synthetic local GGUF refs leak into model display names
## Summary

The dashboard can show internal synthetic model refs like `local-gguf/sha256-66243256b95c5f7c` as the user-facing model name/display name.

These refs are generated for local GGUF files that cannot be mapped back to a Hugging Face repo/file identity. They are valid internal identifiers, but they should not be used as display labels in the app.

## Observed behavior

A model card appears with a title like:

```text
local-gguf/sha256-66243256b95c5f7c
```

The card can show as Offline, Dense, Unknown node, with local metadata such as size and context length. This looks like a broken model name to users.

## Root cause

`crates/model-hf/src/store/local.rs` creates synthetic local GGUF refs in `synthetic_local_gguf_model_ref()` when `model_ref_for_path()` cannot derive a Hugging Face identity for a local GGUF path.

Local inventory then stores only the synthetic key plus size/metadata:

- `crates/mesh-llm-host-runtime/src/models/inventory.rs`
- `LocalModelInventorySnapshot` currently has `model_names`, `size_by_name`, and `metadata_by_name`
- `scan_local_inventory_snapshot_with_progress()` inserts `entry.model_key` but does not preserve the path or filename/stem

The model API path uses that inventory key as the catalog/model name:

- `MeshApi::mesh_models()` calls `RuntimeDataCollector::build_model_view()`
- `build_model_payload_from_catalog_entry()` calls `installed_model_display_name(name)`
- `installed_model_display_name()` only resolves remote catalog names and otherwise returns `model_name.to_string()`

For synthetic local refs there is no remote catalog hit, so the backend sends `display_name` as the raw `local-gguf/sha256-...` key. The React UI then correctly renders the backend-provided display name.

The model target API has a similar fallback in `crates/mesh-llm-host-runtime/src/api/model_targets.rs`, so target/reconciliation surfaces can leak the same internal ref.

## Expected behavior

Synthetic refs should remain stable internal identifiers for routing, load/unload, and path resolution, but user-facing display fields should use a readable local GGUF label, such as the GGUF filename/stem.

For example:

- `name`: `local-gguf/sha256-66243256b95c5f7c`
- `display_name`: `Qwen2.5-7B-Instruct-Q4_K_M` or another filename-derived label

## Suggested fix

Extend local inventory to preserve human-readable local file information by synthetic model key.

Possible implementation:

1. Add a field to `LocalModelInventorySnapshot`, such as `display_name_by_name: HashMap<String, String>` or `path_by_name: HashMap<String, PathBuf>`.
2. Populate it in `scan_local_inventory_snapshot_with_progress()` from `InventoryScanEntry.path.file_stem()` or a shared helper based on `runtime/local.rs::resolved_model_name()`.
3. In `build_model_payload_from_catalog_entry()`, resolve display name in this order:
   - remote/catalog display name if available
   - local inventory display name if the model is locally known
   - current `installed_model_display_name(name)` fallback
4. Update `api/model_targets.rs` to use the same local display-name source so model target views do not keep leaking synthetic refs.
5. Keep the canonical `name` as `local-gguf/sha256-...`; do not replace the internal key globally with a filename because filenames are not unique and the synthetic key maps back to the exact path.

## Test coverage

Add tests for `runtime_data::collector::build_model_view()` with a `LocalModelInventorySnapshot` containing a synthetic `local-gguf/sha256-...` key and local display/path metadata.

Expected assertions:

- payload `name` remains the synthetic ref
- payload `display_name` is the GGUF filename/stem
- size/context/quantization metadata still comes from local inventory
- remote catalog display names still win when present

Also add a model target display test if `api/model_targets.rs` is updated.

Bortlesboat:
Fixes #882

Implements the approach laid out in the issue:

- `LocalModelInventorySnapshot` gains `display_name_by_name`, populated in `scan_local_inventory_snapshot_with_progress()` from the GGUF file stem — only for keys with the synthetic `local-gguf/` prefix, so display behavior for HF-derived refs is untouched.
- The dashboard collector resolves display names as: loaded remote catalog name → local inventory label → raw ref. Split `installed_model_display_name` into an `Option`-returning `loaded_remote_catalog_display_name` so the middle step can slot in without changing existing callers.
- `CatalogTargetIndex` in `api/model_targets.rs` carries the same labels (fed from `runtime_data_collector.local_inventory_snapshot()`), with identical precedence in `display_name_for_model_ref`, so model target views stop leaking synthetic refs too.
- The canonical `name`/key stays the synthetic ref everywhere — only the display label changes, since filenames aren't unique and the synthetic key maps back to the exact path.

Tests per the issue's expectations: `runtime_data_model_snapshot_uses_local_display_name_for_synthetic_refs` asserts the payload keeps the synthetic `name`, shows the file stem as `display_name`, and still carries local size/context/quantization; `display_name_falls_back_to_local_inventory_label_for_synthetic_refs` covers the model-target precedence (catalog wins, label second, raw ref last).

Full `mesh-llm-host-runtime` lib suite passes (1564 tests, Linux); clippy and fmt clean.

## Summary by CodeRabbit

* **New Features**
  * Improved display-name behavior for locally available GGUF models, including synthetic local references derived from file stems.
  * Synthetic local models now surface a local GGUF-derived label as the display name while retaining the canonical model name and existing metadata.

* **Bug Fixes**
  * Updated display-name resolution order to prefer catalog-provided names, then loaded remote catalog names, then local inventory labels (with sensible fallback).

* **Tests**
  * Added/updated unit tests to verify precedence and synthetic local fallback behavior, including handling of non-UTF8 stems via lossy conversion.


crates/mesh-llm-host-runtime/src/models/inventory.rs
```diff
@@ -317,6 +324,14 @@ where
...
     for entry in entries {
         snapshot.model_names.insert(entry.model_key.clone());
+        if entry.model_key.starts_with(SYNTHETIC_LOCAL_GGUF_PREFIX)
+            && let Some(stem) = entry.path.file_stem().and_then(|stem| stem.to_str())
+        {
+            snapshot
+                .display_name_by_name
+                .entry(entry.model_key.clone())
+                .or_insert_with(|| stem.to_string());
+        }
```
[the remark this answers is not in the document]
Bortlesboat: Fixed in the latest commit. Synthetic GGUF stems now use deterministic lossy conversion, with a Unix regression test for a non-UTF-8 filename. I also moved the model-target assertion through ModelTargetSource/build_model_target_lookup so the local-label wiring is covered end to end.

crates/mesh-llm-host-runtime/src/models/inventory.rs
```diff
@@ -55,6 +62,15 @@ struct InventoryScanEntry {
     missing_cache_file: bool,
 }
 
+fn synthetic_local_display_name(entry: &InventoryScanEntry) -> Option<String> {
+    entry
+        .model_key
+        .starts_with(SYNTHETIC_LOCAL_GGUF_PREFIX)
+        .then(|| entry.path.file_stem())
```
ndizazzo, line 69: @Bortlesboat I think at least one Windows-based test of what a flattened file stem looks like might be good here - we can have a separate test if you want to keep the Unix one separate, or make the existing test not just Unix-only

ndizazzo (approved):
Looks good to me! Minor comment WRT testing Windows results of collapsed file stems.

crates/mesh-llm-host-runtime/src/models/inventory.rs
Bortlesboat: Added a separate Windows regression in fc1923f2. It constructs a filename stem containing an unpaired UTF-16 surrogate and verifies the same deterministic lossy label (`Model-�`) while retaining the synthetic model key. Native Windows probe, `cargo fmt --all --check`, and `git diff --check` pass. The focused crate test compile remains locally blocked by the repository's missing patched llama ABI archives on Windows, before reaching this test.

PR #977, opened 2026-07-13
fix(ui): keep chat streaming in hidden tabs

### LINKED ISSUE #950: Console chat output pauses when switching browser tabs
## Problem

When tabbed out of the window during chat generation, generation is paused and does not output in the background while in a different browser tab.

## Steps to reproduce

1. Open the web console
2. Navigate to the chat tab
3. Ask a model any question that yields a long answer
4. Switch tabs, wait 2-3 minutes
5. Navigate back to the tab
6. Observe the chat output resumes printing from where you tabbed away, instead of populating in the background

## Expected Behaviour

The chat output should be added to the window without needing to be looking at it, so users can do something else while the model gives it's response.

Bortlesboat:
## Summary

- publish streamed chat snapshots immediately while the document is hidden
- cancel a queued animation frame before the hidden-tab update to avoid duplicate publication
- retain animation-frame batching for visible tabs

Closes #950.

## Validation

- `pnpm test -- src/lib/streaming.test.ts` (17 passed)
- `pnpm typecheck`
- `pnpm build`
- `pnpm exec prettier src/lib/streaming.ts src/lib/streaming.test.ts --check`
- `git diff --check`

The repository-wide test run completed with 892 passing, 3 skipped, and 6 unrelated existing UI/timing failures. The repository-wide Prettier check also currently reports 509 files outside this change; both changed files pass Prettier directly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Streaming updates now continue to render promptly when the app is in a background browser tab.
  - Pending animation-frame batching is cancelled when the tab becomes hidden, while the latest buffered content is still delivered.

- **Tests**
  - Expanded coverage for streaming batching behavior, including hidden-tab immediate updates and cancellation during visibility transitions.
  - Restored mocks after each test to keep the suite isolated.

### DIFF
--- a/crates/mesh-llm-ui/src/lib/streaming.test.ts
+++ b/crates/mesh-llm-ui/src/lib/streaming.test.ts
@@ -1,6 +1,21 @@
-import { describe, expect, it } from 'vitest'
+import { afterEach, describe, expect, it, vi } from 'vitest'
 
-import { hasBlobContent, parseApiErrorBody } from '@/lib/streaming'
+import { createRafBatcher, hasBlobContent, parseApiErrorBody } from '@/lib/streaming'
+
+afterEach(() => vi.restoreAllMocks())
+
+describe('createRafBatcher', () => {
+  it('publishes updates without waiting for animation frames in a hidden tab', () => {
+    vi.spyOn(document, 'visibilityState', 'get').mockReturnValue('hidden')
+    const requestFrame = vi.spyOn(window, 'requestAnimationFrame')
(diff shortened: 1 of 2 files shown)


Bortlesboat:
Added the visible-to-hidden transition regression in `928ac7c8`. It verifies that a queued animation frame is canceled and the latest snapshot is published immediately. Focused suite is now 18/18; typecheck and scoped ESLint pass.

michaelneale (approved):
Nice

PR #988, opened 2026-07-14
Recognize thinking field in chat auto-parser

michaelneale:
## Summary

- teach llama.cpp’s generic differential chat-template analyzer that assistant `thinking` is an alias of `reasoning_content`
- keep Command A and similar models on the normal generated PEG parser path
- add a synthetic regression template proving `thinking` yields tag-based reasoning with the inferred start/end markers

This replaces the earlier model-specific Cohere parser. The actual failure was generic: Command A’s `tool_use` template uses `message.thinking`, while the differential fixtures populated only `message.reasoning_content`, so llama.cpp inferred `reasoning_mode: NONE` even though it correctly inferred the response and JSON-action envelopes.

## Validation

- clean `scripts/prepare-llama.sh pinned` patch application
- `just build`
- direct template analysis reproduced the pre-fix `reasoning_mode: NONE` for Command A’s named `tool_use` template
- existing live evidence covers plain response, streaming, reasoning, and required tool-call behavior

No Skippy ABI surface change is required.

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing the `thinking` field as reasoning content during chat parsing.
  * Improved detection of reasoning modes and start/end markers for templates that use thinking sections.

* **Tests**
  * Added coverage for templates that emit tagged thinking content.


i386:
@michaelneale im a bit confused why we have to do this for specific models. shouldnt we be leaning more on vanilla llama for this?

michaelneale:
@i386 yeah that is what I thought too, nut not sure how.

michaelneale:
@i386 take a look again - should be different now. is this any better?

i386:
Yep, this is much better. This is now a generic llama.cpp autoparser fix rather than Command A/Cohere special-casing, and it keeps mesh-llm on the normal generated PEG path without adding runtime model dispatch or ABI surface.\n\nMy only ask is that we submit/link this upstream to llama.cpp and treat the local patch as temporary. I don’t think the broader reasoning-field normalization work needs to block this PR.

i386 (approved):

PR #993, opened 2026-07-14
Improve Skippy downstream readiness and decode batching

i386:
## Title
Improve Skippy downstream readiness and decode batching

## Original problem
Split-stage forwarding establishes a downstream TCP connection on the request path, and both local and binary decode batchers dispatch as soon as the first request arrives. This leaves avoidable connection latency at startup and gives concurrent decode requests no bounded opportunity to coalesce.

## Diagnostics
Review of the embedded and binary transport paths showed that they use separate batch collectors with immediate dispatch, while downstream connections are created only when forwarding begins.

## Fix
Add one shared, bounded decode collection policy used by both batchers. `SKIPPY_DECODE_BATCH_COLLECTION_US` enables a collection window up to 10,000 microseconds and is ignored when the configured maximum batch size is one.

Add opt-in downstream warm preconnection with `SKIPPY_BINARY_WARM_PRECONNECT=1`. A background worker maintains one spare connection and normal on-demand connection remains the fallback.

Both features default off, so existing transport and latency behavior remains unchanged. This PR contains no model-specific sidebands, speculation, MTP return paths, or lab routing behavior.

## Validation
- [x] `cargo test -p skippy-server --lib` (211 passed)
- [x] `cargo clippy -p skippy-server --all-targets --no-deps -- -D warnings`
- [x] `git diff --check`
- [x] No UI changes; screenshots do not apply.

## Summary by CodeRabbit

* **New Features**
  * Added optional downstream connection prewarming to reduce connection setup delays.
  * Added optional cross-request decode batching with a configurable collection window.
  * Added configuration guidance for both features, including supported limits and defaults.

* **Performance**
  * Improved decode batching by collecting requests until the configured window expires or the batch reaches capacity.

### DIFF
--- a/crates/skippy-server/README.md
+++ b/crates/skippy-server/README.md
@@ -132,6 +132,12 @@ unload or replan.
   downstream streams open for the lifetime of that lane. `Stop` resets the
   logical session on a lane and leaves the TCP stream open. A failed lane is
   retired and replaced.
+- Set `SKIPPY_BINARY_WARM_PRECONNECT=1` to establish one downstream binary
+  connection while a stage runtime loads and replenish it after use. This is
+  opt-in and leaves the normal on-demand connection path unchanged.
+- Set `SKIPPY_DECODE_BATCH_COLLECTION_US` to an opt-in collection window for
+  cross-request decode batching. The default is zero added wait; configured
+  values are capped at 10,000 microseconds and ignored for single-lane runtimes.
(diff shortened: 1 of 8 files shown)


michaelneale (approved):
slightly confused why specifically tcp (I would think it is over the quic tunnel?)

i386:
@michaelneale its still a TCP stream over QUIC?

Issue #951, opened 2026-07-08
Split serving: second worker's stage never activated, host stays "ready" alone


takesan923:
# Host never activates the second worker's stage in a 2-node `--split` topology (permanent hang; reported "ready" throughout)

## Summary

In a 2-worker `--split` topology, the node elected as the split runtime coordinator (stage-0) finishes loading its own shard and reaches `ready`, but **never activates/connects to the second worker's stage (stage-1)**. The second worker loops forever logging `Split runtime coordinator is <id>; standing by for stage assignment` and never reaches `model_ready`/`ready`. Depending on when a client hits it, inference requests fail in one of two ways: early on, the OpenAI-compatible API returns a fast rejection —

```json
{"error":{"message":"model \"hf://meshllm/Qwen3-32B-UD-Q4_K_XL-layers\" not currently available — retry later","type":"rate_limit_error","param":null,"code":"rate_limit_exceeded"}}
```

— but once the host itself reports `hosted_models` populated and `/v1/models` lists the model as available (host-only state, second stage still never wired in), requests instead **hang indefinitely with no response at all** (confirmed with a 30s client-side timeout: curl exit 28, `HTTP_000`, no body). Either way the model never becomes actually usable.

This has been reproduced **4 times across 2 separate days** with fresh EC2 instances and fresh join tokens each time (not a one-off network blip) — see the live-diagnostics reproduction below for the most conclusive evidence. It does not always happen — we have one prior anecdotal run (logs not captured) where the same 2-node topology came up correctly and the benchmark ran to completion, so this looks like an intermittent/race-condition bug rather than a deterministic misconfiguration.

## Environment

- mesh-llm version: `0.72.2` (release_attestation shows `build_id: mesh-llm-sha256:...`, `commit: task8-local`, `target_triple: x86_64-linux`)
- Topology: 1 client node (t3.micro, `--client`) + 2 worker nodes (AWS EC2 `g5.xlarge`, 1x NVIDIA A10G / 24GB each, `allocatable_vram_bytes` ≈ 22 GiB)
- Model: `hf://meshllm/Qwen3-32B-UD-Q4_K_XL-layers` (Qwen3-32B, 64 layers, layer-package split)
- Launch command (identical on both workers, only the join token differs per mesh):
  ```
  mesh-llm serve --join <TOKEN> --model hf://meshllm/Qwen3-32B-UD-Q4_K_XL-layers --split --max-vram 15 --log-format=json
  ```
[...]

PR #998, opened 2026-07-15
Clarify active split worker status

michaelneale:
## Summary

- replace the ambiguous recurring split-worker "standing by for stage assignment" message with status derived from authoritative local stage state
- report `stage-1` as starting or ready when the coordinator has already assigned it
- clarify that model requests remain routed through the coordinator

## Why

Issue #951 showed that the old message could continue while the same node was already running a ready downstream stage. That made a loaded worker look unassigned and led operators toward the wrong diagnosis.

This PR intentionally changes reporting only. It does not claim to fix the inference hang in #951, change split election or assignment, alter readiness semantics, or add a request timeout.

## Compatibility

No mesh protocol, plugin protocol, Skippy ABI, API schema, routing, or model-load behavior changes.

## Validation

- `cargo fmt --all --check`
- `cargo test -p mesh-llm-host-runtime --lib split_status -- --test-threads=1` (3 passed)
- `cargo test -p mesh-llm-host-runtime --lib -- --test-threads=1` (1575 passed, 8 ignored)
- `cargo check -p mesh-llm`
- `git diff --check`

## Notes

The full host-runtime suite initially failed inside the restricted filesystem/socket sandbox because socket binding returned `Operation not permitted`. The required rerun outside that sandbox passed completely.

Related: #951

## Summary by CodeRabbit

* **New Features**
  * Added clearer standby status messages for split-runtime coordinators.
  * Status now indicates whether the system is waiting for a local stage assignment or identifies the active stage and its state.

* **Bug Fixes**
  * Improved stage selection so messages reflect the correct active local stage for the requested model.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/runtime/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/mod.rs
@@ -10,6 +10,7 @@ mod options;
 mod proxy;
 mod release_attestation;
 mod split_planning;
+mod split_status;
 pub(crate) mod survey;
 pub(crate) mod wakeable;
 
@@ -1585,10 +1586,7 @@ where
             Ok(SplitRuntimeStart::Standby { coordinator }) => {
                 drop(startup_load_guard);
                 let _ = emit_event(OutputEvent::Info {
-                    message: format!(
-                        "Split runtime coordinator is {}; standing by for stage assignment",
-                        coordinator.fmt_short()
-                    ),
+                    message: split_status::standby_message(node, model_ref, coordinator).await,
                     context: Some(format!("model={model_ref}")),
                 });
                 startup_reset_model_target(target_tx, model_name, console_state).await;
--- a/crates/mesh-llm-host-runtime/src/runtime/split_status.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/split_status.rs
@@ -1,0 +1,138 @@
+use crate::inference::skippy;
+use crate::mesh;
+
+pub(super) async fn standby_message(
+    node: &mesh::Node,
+    model_ref: &str,
+    coordinator: iroh::EndpointId,
+) -> String {
+    let statuses = node.stage_runtime_statuses().await;
+    let active_stage = active_local_stage(node.id(), model_ref, &statuses);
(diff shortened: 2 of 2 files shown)


michaelneale:
@i386 not sure if relevant given the ongoing work but this may help with the wording

ndizazzo:
Agent-assisted literal 50,000 ft review:

1) Status is read from a cache⁠, not queried from the local stage runtime. 👉 Refresh/query the local stage state before formatting, and add a regression test for a cached-ready stage that is no longer running.

2) Stage selection ignores coordinator/topology identity⁠. 👉 Preserve coordinator identity in StageRuntimeStatus and require it to match coordinator (or otherwise bind selection to the active run). Test two ready local stages for one model under different coordinators.

And a NIT:

Inactive-state formatting is unreachable⁠.

active_local_stage only returns Starting or Ready, so the Stopping/Stopped/Failed arms can never produce “inactive.”

---

For the starting / stopping / failed arms, I actually started instrumenting the llamacpp runtime with a patch to output events when we have underlying state changes... I think you could probably extend those to cover these cases as well

I had originally meant to revise the runtime debug and loading events with it so we could get instrumentation and timing metrics. You might want to have an agent investigate what's there and fit some events for the unused statuses

ndizazzo (changes requested):
See above

michaelneale:
@ndizazzo do you want me to close this, sounds like you might have some ideas. this was more helping diagnose but it is based on a report, not my experience.

ndizazzo:
@mic nah! We can keep it open and work toward merging... I just added the context because I spotted a place where some stuff I touched recently could fit.

mic:
> @mic nah! We can keep it open and work toward merging... I just added the context because I spotted a place where some stuff I touched recently could fit.

lgtm :)

mic:
> > @mic nah! We can keep it open and work toward merging... I just added the context because I spotted a place where some stuff I touched recently could fit.
> 
> lgtm :)

btw. you tagged the wrong mic

ndizazzo:
Haha, oops! Sorry. Thought autocomplete finished the username

The real @michaelneale

PR #999, opened 2026-07-15
Fix model catalog discovery and package visibility

michaelneale:
## Summary

- build the public catalog from the canonical `entries/**/*.json` tree used by the CLI, instead of the stale flattened Dataset Viewer projection
- publish a readable same-origin catalog snapshot during normal or manually triggered website deployments, and test the source-to-package mapping in website CI
- show whether catalog results resolve to a GGUF or Skippy layer package and print the exact `serve --model` command in console and JSON output
- clarify the model-selection flow in the catalog, Choose a Model, Running Large Models, CLI, and contributor docs
- distinguish curated catalog membership from split readiness: only entries with a layer package are prepared for multi-machine serving; broad Hugging Face search remains uncurated

This addresses the model discovery and source-ref versus layer-package confusion discussed in https://github.com/Mesh-LLM/mesh-llm/discussions/978.

## Catalog refresh behavior

The public website is static, while the canonical catalog is maintained in the separate `meshllm/catalog` Hugging Face dataset. Each normal or manually triggered website deployment reads the canonical entry tree and embeds a same-origin JSON snapshot in the Pages artifact. Catalog-only changes therefore appear on the next website deployment; a maintainer can manually run Public Website Deploy when an immediate refresh is wanted. Production deployment requires a fresh fetch, while local offline builds may use the checked-in readable snapshot.

## Validation

- `just build`
- `just website-build`
- `npm test` (3 catalog mapping tests)
- `cargo fmt --all --check`
- `cargo check -p mesh-llm`
- focused model formatter, catalog lookup, and search JSON unit tests
- live `models show` verification for `unsloth/Qwen3.6-27B-MTP-GGUF:UD-Q4_K_XL`, resolving to `hf://meshllm/Qwen3.6-27B-UD-Q4_K_XL-layers` while preserving the source ref in the serve command

The generated fallback snapshot contains 106 catalog variants and is kept formatted so its source-to-package mappings remain reviewable.

## Summary by CodeRabbit

* **New Features**
  * `models search`, `models recommended`, and `models show` now include resolved artifact/layer-package details (including for GGUF variants) plus the exact `mesh-llm serve --model ...` command.
* **Documentation**
  * Updated CLI and model/catalog guides to clearly describe curated catalog resolution, artifact ↔ layer-package mapping, and serving flows using the displayed `serve` command/ref.
* **Tests**
  * Added catalog-data mapping coverage for the website and introduced build-time `npm test` checks before publishing the public site.


website/src/assets/catalog-data.js
```diff
@@ -0,0 +1,110 @@
...
+    const packages = layerPackages(variant)
+    const primaryPackage = packages[0]
+    return {
+      schema_version: entry.schema_version || 1,
+      entry_path: entryPath,
+      variant_id: variantId,
+      name: curated.name || variantId,
+      size: curated.size || '',
+      description: curated.description || '',
+      capabilities: Array.isArray(curated.capabilities) ? curated.capabilities.join(' ') : '',
```
ndizazzo, line 33: The generated snapshot currently has capabilities: "" for all 106 rows, so choosing Code, Vision, or Long context always returns zero results. Also, 70 rows use layer counts as size, which turns into “Size pending” and excludes them from every hardware filter.

Deriving these fields during snapshot generation, adding them to the canonical schema, or remove filters that cannot be populated would be an improvement. It would also be good to test that every available filter matches at least one generated row.

website/src/catalog/index.njk
ndizazzo, line 155: Can we emit the canonical model ref in the generated snapshot instead of reconstructing it with this quant regex?

Valid current entries such as GLM-5.1-Q3_K_M-plus and GLM-5.2-Q2_K-MTP-Q8 don't match the regex, so the page drops their selector, displays no quant badge, and copies only the bare repository. Rendering a generated model_ref directly would also avoid duplicating the Rust model-ref parsing rules in JavaScript.

crates/mesh-llm-host-runtime/src/models/search.rs
```diff
@@ -97,6 +97,7 @@ pub fn search_catalog_json_payload(
         .take(limit)
         .map(|model| {
             let model_ref = remote_catalog_model_ref(model);
+            let layer_package_ref = remote_catalog::find_layer_package(&model_ref);
```
ndizazzo, line 100: I think we should resolve the layer-package mappings once for the complete result set

`find_layer_package()` clones the loaded catalog, constructs a new resolver, and scans it for every result. Since this endpoint can return 50 results and the catalog can grow toward 1,000 entries, this adds repeated synchronous work to each request. 

A reusable resolver or model-ref-to-package lookup would avoid that scaling cost.

ndizazzo (changes requested):

PR #1003, opened 2026-07-15
docs: add agent-guided mesh installation runbook

michaelneale:
## Summary

- add an assistant-led workflow for setting up a private mesh
- cover release selection, hardware survey, node enrollment, serving plans, diagnostics, and agent-harness connection
- include secret-safe journaling and real-world failure references

## Status

Draft preserved from the earlier installation-guide work for review and refinement.

## Summary by CodeRabbit

* **Documentation**
  * Added new agent-facing runbooks for setting up a private Mesh LLM, including supervised install flow, capability/model discovery, node enrollment, topology planning, and bounded serving verification.
  * Published the setup guide as a new website page for easy access.
* **Build & Deployment**
  * Updated website build/publishing instructions and artifact output mapping to match the generated site structure.
  * Adjusted the website publishing workflow to include the new setup guide in the deployed pages artifact.


website/.eleventy.js
```diff
@@ -38,6 +38,7 @@ export default function(eleventyConfig) {
   eleventyConfig.addPassthroughCopy("src/assets");
   eleventyConfig.addPassthroughCopy({ "../install.sh": "install.sh" });
   eleventyConfig.addPassthroughCopy({ "../install.ps1": "install.ps1" });
+  eleventyConfig.addPassthroughCopy({ "../install.md": "setup-mesh" });
```
ndizazzo, line 41: This creates `docs/setup-mesh`, but the Pages deployment workflow does not copy that file into `public-website-artifact`, and `install.md` is not included in the workflow path triggers. As written, https://meshllm.cloud/setup-mesh will still return 404 after merge, and future changes limited to `install.md` will not deploy. 

To fix, we can add `install.md` to the deploy triggers and stage `docs/setup-mesh`. The website output check and clean-generated-site manifest should include it too.

docs/setup-mesh
```diff
@@ -0,0 +1,644 @@
...
+- [Configuration](https://meshllm.cloud/docs/config-reference/)
+
+Do not make a mesh public unless the user explicitly asks. Do not post an
+invite token, credentials, host inventory, or private paths to a public service.
+An ordinary private invite is connectivity material, not a strong
+identity/admission policy. For an untrusted network or controlled membership,
+stop and discuss the owner identity and trust-policy options in
+<https://meshllm.cloud/MESHES.md> before enrolling nodes.
+
+### Pick the shortest path for the goal
```
ndizazzo: These changes were made only in the generated artifact. `install.md` is the source copied by Eleventy, and it currently differs from `docs/setup-mesh`, so the next website build will remove this routing table and the installed-model filtering added below. 

We should move the changes into `install.md`, run `just website-build`, and verify the two files match. The AGENTS wording should also point agents to the owning source, since `install.md` is outside `website/src/**`.

install.md
```diff
@@ -0,0 +1,611 @@
...
+The other machine does not need an agent and the main machine does not need
+direct access to it. Generate one copy/paste block containing install, setup,
+and join. Substitute the exact invite and a unique node name. Quote values for
+the target shell and never use placeholders in the final block.
+
+For macOS or Linux:
+
+```sh
+set -eu
+curl -fsSL https://meshllm.cloud/install.sh | bash -s -- --no-setup
```
ndizazzo: This unconditionally downloads and installs the latest release, even when preflight found an existing, custom-path, or source installation. It also assumes the installer used `~/.local/bin`, although `MESH_LLM_INSTALL_DIR` may override that. 

This contradicts the earlier instructions to ask before installing or updating and to avoid duplicate installations. 

Please make the generated block reuse the approved resolved binary and run the installer only when Mesh is absent. The Windows block has the same issue.

install.md
```diff
@@ -0,0 +1,611 @@
...
+curl -fsSL https://meshllm.cloud/install.sh | bash
+```
+
+Windows x64 PowerShell:
+
+```powershell
+irm https://meshllm.cloud/install.ps1 | iex
+```
+
+The installer normally runs `mesh-llm setup` when interactive. Afterwards,
```
ndizazzo: The documented Unix invocation uses `curl ... | bash`, which makes stdin non-TTY. `install.sh` therefore prints the follow-up setup command instead of automatically running it, even from an interactive terminal. 

The explicit setup step below is correct, but this sentence should describe the actual piped-installer behavior so an assistant does not incorrectly report that setup already ran.

ndizazzo (commented):
@michaelneale These are my agent review findings, after feeding it to Sol to follow

michaelneale:
@ndizazzo you want to push some changes and take this one over - mostly I am focussing on the wording to guide the agent, but could use with some eyes on making sure it shows up on website right and does reasonable stuff, so feel free to push away.

ndizazzo:
@michaelneale SG, generally I try to comment only JIC things are still being worked on, but I can take over if you want to focus elsewhere

PR #1009, opened 2026-07-16
long term effort: Serve safetensors models on Metal via MLX (Apple Silicon)

michaelneale:
## What this draft now proves

This branch has a connected Apple-Silicon path from Hugging Face SafeTensors
to both ordinary single-node serving and mesh-managed explicit split serving.

### Off-the-shelf whole-model MLX serving

The normal shipped command now serves both canonical dense checkpoints and
unchanged MLX-LM repositories through mesh-llm's shared OpenAI frontend:

```bash
mesh-llm serve --model HuggingFaceTB/SmolLM2-135M-Instruct
mesh-llm serve --model mlx-community/Qwen3-0.6B-4bit --ctx-size 16384
```

The second command was exercised at HF revision
`73e3e38d981303bc594367cd910ea6eb48349da8`. It loaded the cached model in
about 1.2 seconds, appeared in `/v1/models`, answered `23 * 17` with `391`,
completed SSE with `[DONE]`, and completed a Goose OpenAI-provider request.
The same path now reports the repository as 0.6B rather than mistaking its
`4bit` suffix for a 4B parameter label.

Published MLX-LM affine checkpoints commonly omit `quantization.mode`. The
pinned safemlx correction treats that omission as the standard `affine`
default. The one-commit fork is proposed upstream as
[jbg/safemlx#2](https://github.com/jbg/safemlx/pull/2).

For eligible unquantized dense checkpoints, automatic mode applies
affine-4/group-64 while loading. Inkling, Nemotron-H, GGUF files, and sources
already declaring quantization/compression preserve their native
representation. Qwen3-0.6B's exact benign tied-`lm_head.weight` rejection uses
a narrow native-load retry; all unrelated strict-load failures remain
fail-closed.

### Mesh-managed range-only split serving

With explicit `--split`, the coordinator:

1. resolves config/index/shard headers and tokenizer sidecars without fetching
   checkpoint tensor payloads;
2. advertises and selects additive `backend-mlx` stage capability;
3. plans topology using header-derived per-layer affine-4 estimates, including
   embedding/readout boundary costs;
4. asks each stage to fetch and derive only its assigned tensor ranges;
5. drives the stage chain from the normal OpenAI frontend over Skippy's binary
   activation wire.

A real two-host SmolLM2 run used independent identities and ordinary stage
control:

| Stage | Layers | Derived affine-4 artifact |
| --- | ---: | ---: |
| coordinator | `0..29` | about 70 MB |
| remote final | `29..30` | about 17 MB |

The remote did not store the complete roughly 269 MB source checkpoint.
Non-streaming chat traversed both stages and returned coherent text; SSE
completed with `[DONE]`.

This is deliberately different from whole-model prequantized loading:
prequantized `mlx-community` repositories work unchanged on one node, while
distributed `--split` currently starts from canonical dense SafeTensors and
derives stage-local quantized artifacts. It does not yet slice a published
prequantized repository across nodes.

### Lifecycle and compatibility hardening

- Immutable Hugging Face revisions and strong shard identities are required
  for split tensor ranges.
- Sidecar and derived caches are lock-protected and atomically published.
- Only Prepare derives a missing stage; Load validates and consumes it.
- EOF, error, timeout, or Stop resets touched local sessions and propagates Stop
  downstream.
- Stage/generation I/O is bounded; frontend bind completes before readiness.
- Distributed tokenization includes special template variables and UTF-8-safe
  incremental decoding with final suffix reconciliation.
- MLX capability gossip is additive and advertised only by an Apple-Silicon
  MLX build.
- Missing native llama runtime no longer blocks an MLX-only start; actual
  GGUF/Skippy loads still require it.

### Frontier evidence retained

- Exact-range planning against pinned Inkling BF16 showed four layers contain
  109.84 GiB across 942.99 GiB of source shards, avoiding 833.15 GiB.
- A pinned Nemotron-H Nano MoE layer is range-planned, assembled into bounded
  affine expert banks, strict-loaded, executed, and validated through the
  engine-neutral stage contract and binary wire.
- Manual two-process/two-host SmolLM2, F16/F32 boundary, cache identity,
  cancellation, and portable `mlx.metallib` evidence are documented in
  `crates/skippy-engine-mlx/STAGED_EXECUTION.md`.

## Validation for the latest checkpoint

- `just mlx-build`: pass; emits and signs `mesh-llm` plus sibling
  `mlx.metallib`.
- Exact off-the-shelf Qwen3-0.6B-4bit model listing, non-streaming chat, SSE,
  Goose, and clean shutdown: pass; independently reproduced by the MLX expert.
- Real two-host mesh-managed 29/1 split and stage-local storage check: pass.
- `skippy-engine-mlx --features mlx --lib`: 42 passed.
- `mesh-llm-host-runtime --lib`: 1,602 passed, 8 ignored.
- safemlx fork `safemlx-lm --lib`: 119 passed, 66 ignored.
- Warning-denying Clippy for the MLX engine and shipped MLX binary: pass.
- Formatting, publish-chain consistency, release-target consistency,
  CI-crate-list consistency, and publish dry-run checks: pass.
- Final expert audit found no runtime blocker and its publication/claim issues
  are now stated below rather than hidden.

## Deliberately incomplete

- Partial download currently requires explicit `--split`; automatic split
  selection is not implemented.
- Integrated partial generation supports dense `model_type=llama`, one greedy
  lane. Inkling partial stages and complete Nemotron hybrid topology remain.
- The automatic whole-model policy is affine-4-or-native based on checkpoint
  metadata, not yet a user-facing hardware/quality policy.
- The MLX adapter does not yet offer the full llama.cpp feature surface such as
  tools, sampling controls, logprobs, parallel lanes, or multimodal requests.
- Derived-cache capacity and eviction remain unowned.
- The workspace source build requires its pinned safemlx root patch. Cargo root
  patches do not propagate, and registry `safemlx-lm 0.4.1` lacks APIs used by
  this engine. Standalone crates.io `skippy-engine-mlx --features mlx` support
  therefore remains unusable until compatible safemlx releases land.


crates/model-hf/src/safetensors_stage/http.rs
```diff
@@ -0,0 +1,438 @@
...
+            "HTTP range returned {} bytes, expected {}",
+            bytes.len(),
+            self.expected_bytes
+        );
+        Ok(bytes)
+    }
+
+    pub fn copy_to(mut self, writer: &mut impl Write) -> Result<u64> {
+        let written =
+            std::io::copy(&mut self.response, writer).context("stream HTTP tensor range")?;
```
i386, line 238: These paths validate the byte count only after fully consuming the response. A server can return `206` with a matching `Content-Range` and no `Content-Length`, then stream far more than requested; `read_to_end`/`io::copy` will allocate or write all of it before the equality check, defeating the full-shard-download guard and potentially exhausting memory or disk. Please cap both readers at `expected_bytes + 1` and reject as soon as an extra byte is observed.

i386 (commented):
One partial-download safety issue in the new exact-range reader.

i386 (commented):
Review of current head `75fb8b24`.

One blocking safety issue remains in the new exact-range download path: `ExactRangeResponse::into_bytes()` and `copy_to()` consume the response without a limit and only compare the byte count afterward. A `206` response with the requested `Content-Range` but no `Content-Length` can stream arbitrarily more data, causing unbounded memory growth or disk writes before rejection. Cap both readers at `expected_bytes + 1` and reject immediately when the extra byte is observed. Detailed inline comment: https://github.com/Mesh-LLM/mesh-llm/pull/1009#discussion_r3650136943.

The staged-protocol changes I inspected are additive: older peers can ignore the new fields, and missing weight-quantization values default to `Auto`. The whole-model MLX and explicit split evidence is useful, but this remains a very large 93-file integration (engine abstraction, partial HTTP/materialization, host wiring, docs, and three spikes) and is conflicted with current `main`; splitting/removing the spike work would make the production path substantially easier to review and maintain.

crates/model-hf/src/safetensors_stage/materialize.rs
```diff
@@ -0,0 +1,1133 @@
...
+    /// distributed stage-0 frontend. No model tensor payload is downloaded.
+    pub fn prepare_checkpoint(
+        &self,
+        repo: &str,
+        revision: &str,
+    ) -> Result<PreparedSafetensorsCheckpoint> {
+        let descriptor = self.describe_checkpoint(repo, revision)?;
+        let cache_key = format!(
+            "checkpoint-{:x}",
+            Sha256::digest(format!("{repo}@{revision}").as_bytes())
```
ndizazzo, line 89: [P2] Include HF_ENDPOINT in the sidecar cache identity. The sidecar key uses only `repo@revision`, even though `from_environment` reads `HF_ENDPOINT` and the stage artifact key includes the endpoint. If a process switches between two endpoints serving the same repo/revision, `checkpoint_sidecars_match` accepts the existing entry because the descriptor has no endpoint and only checks config/tokenizer presence. The cached tokenizer and optional chat template from endpoint A can then drive generation against endpoint B’s weights. Include the normalized endpoint in this key or record and validate the sidecar source identity.

ndizazzo (changes requested):
I’m requesting changes. The existing unresolved range-reader finding at `crates/model-hf/src/safetensors_stage/http.rs:238` remains blocking: `into_bytes()` and `copy_to()` consume the entire 206 body before checking its length, so a response without `Content-Length` can grow memory or disk without bound. The inline comment below covers a separate cache identity issue in the new MLX sidecar path.

The available green checks are historical July runs; I did not treat them as current validation, and local tests were blocked by the host’s unaccepted Xcode license.

PR #1011, opened 2026-07-16
Make split serving resilient to memory pressure and network jitter

michaelneale:
## What this does

Makes Skippy stage-split serving usable on real hardware. Previously, splitting a
model across machines could OOM-crash a stage on the first token, swap the host
into unusability, hang bring-up, or tear the whole split down on a brief network
blip. This makes a split stay up and generate under normal home-network and
memory conditions.

## Why

Bringing up a real 2-node Apple Silicon split (MiniMax-M2.7 across a 128 GB + a
64 GB Mac) surfaced several distinct failures:

- **Memory pressure / OOM.** The planner packed each node to its full advertised
  VRAM budget. On Apple Silicon that budget is already ~75% of *total* unified
  memory (`recommendedMaxWorkingSetSize`), so packing it near-full left nothing
  for decode command buffers, the OS, or per-session state — the first token
  failed with `kIOGPUCommandBufferCallbackErrorOutOfMemory` and the host swapped
  hard. Split serving was actually reserving *less* headroom than single-node
  loading, which is backwards.
- **Network fragility.** Stage handshakes did one-shot blocking reads with no
  timeout, so a jitter blip either aborted lane/return-sink setup or hung
  bring-up forever, and the coordinator withdrew the whole model on the first
  missed health tick.

## Changes

- **Planner reserves 20% decode/OS headroom** (`split_planning.rs`) instead of
  effectively zero, and larger than the ~10% solo-load reserve since split
  serving is more memory-hungry (activation frames, persistent lanes, cross-node
  buffers). `--max-vram` still lets you push a node harder.
- **Retry transient handshake failures** with backoff for both the stage lane
  pool (`frontend.rs`) and the prediction-return sink (`direct_return.rs`).
- **Bounded read timeout on the ready handshake** so a stalled downstream no
  longer hangs bring-up (and the retries can actually fire). This also makes the
  default multi-lane pool bring up reliably.
- **Grace period on transient stage-peer loss** (`local.rs`) so a brief gossip
  drop no longer withdraws the model on the first health tick.

## Testing

- New/updated unit tests for headroom sizing, both handshake retry helpers, and
  the peer-loss grace decision. `cargo test`, `clippy -D warnings`, and `fmt`
  clean on `mesh-llm-host-runtime` and `skippy-server`.
- Validated live on a 2-node M5+M4 split: forms, serves, and generates correct
  tokens; multi-lane bring-up reliable; recovered across network storms.

## Caveats

- The **peer-loss grace period** is unit-tested but **not yet integration-validated**
  against a real multi-node peer drop (home Wi-Fi was too unstable for a clean
  controlled test). Worth a deliberate peer-kill test on a stable network before
  relying on it.
- Headroom fraction and grace window are fixed constants; they could later be
  config-exposed if needed.

## Summary by CodeRabbit

- **Reliability Improvements**
  - Added a 75-second grace period for split withdrawal when the only recovery action is `Withdraw`.
  - Added bounded timeouts to downstream and prediction-return “ready” handshakes to avoid indefinite hangs.
- **Resource Planning**
  - Increased per-node VRAM headroom to a 10% cushion.
  - Updated KV-cache memory accounting to include compute-graph buffer reserve (100/85 scaling), adjusting fit/shape decisions.
- **Bug Fixes**
  - Improved transient connection readiness handling and clearer error context.
- **Tests**
  - Added unit coverage for the grace decision logic and updated planning/regression expectations.

### DIFF
--- a/crates/skippy-server/src/binary_transport/direct_return.rs
+++ b/crates/skippy-server/src/binary_transport/direct_return.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::HashMap,
     io,
-    net::{SocketAddr, TcpListener, TcpStream},
+    net::{IpAddr, SocketAddr, TcpListener, TcpStream},
     sync::{
         Arc, Mutex,
         atomic::{AtomicBool, Ordering},
@@ -287,6 +287,58 @@ impl PredictionReturnSinks {
     }
 }
 
+/// Read timeout for the return-sink ready handshake. `recv_ready` is a blocking
+/// `read_exact`; without this a stalled downstream connection hangs the open
+/// forever, which mid-generation blocks the request from ever falling back to
+/// the upstream reply. Kept short so a genuinely stalled peer fails fast to the
+/// fallback; cleared afterwards so the sink's normal reads stay blocking.
+///
+/// This is a *single bounded deadline*, not a retry budget: the sink is opened
+/// on the generation hot path, and `connect_downstream_socket` already bounds
+/// the connect itself, so wrapping this in an outer retry only compounds the
+/// worst-case stall (see PR #1011 review).
+const RETURN_SINK_READY_READ_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Connect to `return_addr`, complete the ready handshake, and send the
+/// prediction-return open message. Single bounded attempt — on failure the
+/// caller falls back to the upstream reply path.
+fn open_return_sink_once(
+    return_addr: SocketAddr,
+    source_ip: Option<IpAddr>,
+    request_id: u64,
+    session_id: u64,
+    wire_dtype: WireActivationDType,
+    not_ready_context: &'static str,
+) -> Result<TcpStream> {
+    let mut stream = connect_downstream_socket(return_addr, source_ip, Duration::from_secs(2))
+        .map_err(|error| anyhow!(error))?;
+    stream.set_nodelay(true).ok();
+    send_client_ready_hello_if_enabled(&mut stream)
+        .context("send prediction return client ready hello")?;
+    // Bound the ready handshake read. `recv_ready` is a blocking `read_exact`;
+    // without a timeout a stalled downstream connection hangs the return-sink
+    // open forever, blocking generation from falling back to the upstream reply.
+    // A single short deadline (no outer retry) fails fast to that fallback.
+    // Both the set and the clear are propagated: if the set fails, `recv_ready`
+    // would be unbounded (defeating the fix); if the clear fails, the handshake
+    // timeout would leak into the sink's later reads.
+    stream
+        .set_read_timeout(Some(RETURN_SINK_READY_READ_TIMEOUT))
+        .context("set prediction return ready read timeout")?;
+    let ready = recv_ready(&mut stream).context(not_ready_context);
+    stream
+        .set_read_timeout(None)
+        .context("clear prediction return ready read timeout")?;
+    ready?;
+    write_stage_message(
+        &mut stream,
+        &prediction_return_open_message(request_id, session_id),
+        wire_dtype,
+    )
+    .context("open prediction return stream")?;
+    Ok(stream)
+}
+
 pub(crate) fn open_prediction_return_stream(
     config: &StageConfig,
     topology: Option<&StageTopology>,
@@ -298,34 +350,15 @@ pub(crate) fn open_prediction_return_stream(
     let endpoint = driver_stage_endpoint(config, topology)?;
     let return_addr = resolve_downstream_endpoint(endpoint)?;
     let source_ip = downstream_source_ip(config)?;
-    let attempts = 1;
-    let mut last_error = None;
-    for _ in 0..attempts {
-        match connect_downstream_socket(return_addr, source_ip, Duration::from_secs(2)) {
-            Ok(mut stream) => {
-                stream.set_nodelay(true).ok();
-                send_client_ready_hello_if_enabled(&mut stream)
-                    .context("send prediction return client ready hello")?;
-                recv_ready(&mut stream).context("prediction return sink did not become ready")?;
-                write_stage_message(
-                    &mut stream,
-                    &prediction_return_open_message(request_id, session_id),
-                    wire_dtype,
-                )
-                .context("open direct prediction return stream")?;
-                return Ok(stream);
-            }
-            Err(error) => {
-                last_error = Some(anyhow!(error));
-                std::thread::sleep(Duration::from_millis(500));
-            }
-        }
-    }
-    Err(last_error
-        .unwrap_or_else(|| anyhow!("timed out"))
-        .context(format!(
-            "connect direct prediction return sink at {endpoint}"
-        )))
+    open_return_sink_once(
+        return_addr,
+        source_ip,
+        request_id,
+        session_id,
+        wire_dtype,
+        "prediction return sink did not become ready",
+    )
+    .with_context(|| format!("connect direct prediction return sink at {endpoint}"))
 }
 
 pub(crate) fn open_downstream_prediction_return_stream(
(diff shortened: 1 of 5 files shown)


michaelneale:
@ndizazzo PTAL

crates/skippy-server/src/frontend.rs
```diff
@@ -1176,6 +1228,41 @@ impl PersistentStageLanePool {
...
         })
     }
+
+    /// Perform a single connect + ready-handshake attempt for one lane.
+    ///
+    /// Returns the connected, ready `TcpStream` on success. Callers retry this
+    /// with a backoff so a transient short read on `recv_ready` does not fail
+    /// lane creation outright.
+    fn connect_lane_once(&self, lane_id: u64) -> Result<TcpStream> {
+        let mut stream = connect_binary_downstream(&self.config, self.timeout_secs)?
```
ndizazzo, line 1194: I think we have an issue here - connect lane calls this 5 times, but this connect_binary_downstream already was consuming the connection budget. We'll probably end up with more than 30 seconds of timeout even when a connection can fail immediately

crates/skippy-server/src/binary_transport/direct_return.rs
```diff
@@ -341,19 +407,22 @@ pub(crate) fn open_downstream_prediction_return_stream(
...
-        .with_context(|| format!("connect downstream prediction return sink at {endpoint}"))?;
-    stream.set_nodelay(true).ok();
-    send_client_ready_hello_if_enabled(&mut stream)
-        .context("send downstream prediction return client ready hello")?;
-    recv_ready(&mut stream).context("downstream prediction return sink did not become ready")?;
-    write_stage_message(
-        &mut stream,
-        &prediction_return_open_message(request_id, session_id),
-        wire_dtype,
+    retry_return_sink(
```
ndizazzo: Same for the downstream sink on stall

crates/skippy-server/src/binary_transport/direct_return.rs
```diff
@@ -298,34 +376,22 @@ pub(crate) fn open_prediction_return_stream(
...
+                source_ip,
+                request_id,
+                session_id,
+                wire_dtype,
+                "prediction return sink did not become ready",
+            )
+        },
+        std::thread::sleep,
+    )
+    .with_context(|| format!("connect direct prediction return sink at {endpoint}"))
```
ndizazzo, line 361: Another delay + timing issue here too... If a peer TCP stalls, it will block generation for much longer before the upstream-reply falback runs. We can probably just have a single short total deadline.

ndizazzo (commented):
Overall I think it's alright but I think this will introduce some excess timing delays without slight adjustments

michaelneale:
@ndizazzo yeah some are necessary though as in a split world you want things to change very very slowly

crates/skippy-server/src/binary_transport/direct_return.rs
```diff
@@ -287,6 +287,84 @@ impl PredictionReturnSinks {
...
+const RETURN_SINK_HANDSHAKE_ATTEMPTS: u32 = 5;
+
+/// Backoff between prediction-return-sink connect + handshake attempts.
+const RETURN_SINK_HANDSHAKE_BACKOFF: Duration = Duration::from_millis(500);
+
+/// Read timeout for the return-sink ready handshake. `recv_ready` is a blocking
+/// `read_exact`; without this a stalled downstream connection hangs the open
+/// indefinitely and the retry loop never fires. Bounded so a stall becomes a
+/// retryable error; cleared afterwards so the sink's normal reads stay blocking.
+const RETURN_SINK_READY_READ_TIMEOUT: Duration = Duration::from_secs(20);
```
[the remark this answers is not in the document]
ndizazzo: Coderabbit spotted the same issue

crates/skippy-server/src/frontend.rs
michaelneale: Good catch — fixed in b1f511ec. Removed the outer `retry_lane_handshake` wrapper entirely; `connect_lane` now makes a single bounded attempt. `connect_binary_downstream` already owns the TCP connect budget, and the only remaining unbounded wait was the blocking `recv_ready`, which now has a single read timeout (`LANE_READY_READ_TIMEOUT`). No more multiplying connect-budget × outer attempts.

crates/skippy-server/src/binary_transport/direct_return.rs
michaelneale: Done in b1f511ec — went with exactly your suggestion: a single short total deadline, no retry. Removed `retry_return_sink`; both `open_prediction_return_stream` and `open_downstream_prediction_return_stream` now do one bounded attempt, and `RETURN_SINK_READY_READ_TIMEOUT` is dropped to 5s so a stalled peer fails fast to the upstream-reply fallback instead of blocking generation.

crates/skippy-server/src/binary_transport/direct_return.rs
michaelneale: Fixed together with the other sink path in b1f511ec — same single-bounded-deadline treatment (5s read timeout, no outer retry), fails fast to the upstream-reply fallback.

michaelneale:
Pushed \`b1f511ec\` addressing the review:

**@ndizazzo (timeouts) — the main fix.** You were right that stacking retries on top of `connect_binary_downstream` (which already owns the connect budget) compounded the worst-case stall. Reworked both handshakes to a **single bounded deadline**, no outer retry:
- `connect_lane`: one attempt; only the blocking `recv_ready` is bounded (`LANE_READY_READ_TIMEOUT`).
- return sink (both paths): one attempt; `RETURN_SINK_READY_READ_TIMEOUT` dropped to **5s** so a stalled peer fails fast to the upstream-reply fallback instead of blocking generation.
- Net −183 lines (removed the retry helpers + their tests).
- The peer-loss **grace period** is intentionally kept slow — that governs topology churn, not the connection hot path.

**CodeRabbit nits:**
- Magic `100/85` → named `KV_COMPUTE_RESERVE_*` constants.
- Grace tests no longer subtract from `Instant::now()` (baseline + add) to avoid the fresh-VM underflow panic.

**Deferred:** the `local.rs` >2000-line extraction. `local.rs` was already ~6k lines before this PR; carving out `SplitTopologyCoordinator` is a worthwhile refactor but is orthogonal to these resilience fixes and risky to land mid-review. Tracking as a follow-up rather than expanding this PR.

All touched crates: tests green, `clippy -D warnings` + `fmt` clean.

crates/skippy-server/src/binary_transport/direct_return.rs
```diff
@@ -287,6 +287,84 @@ impl PredictionReturnSinks {
...
+        .context("send prediction return client ready hello")?;
+    // Bound the ready handshake read. `recv_ready` is a blocking `read_exact`;
+    // without a timeout a stalled downstream connection hangs the return-sink
+    // open forever and the surrounding retry never fires. Cleared afterwards so
+    // the sink's normal reads stay blocking.
+    stream
+        .set_read_timeout(Some(RETURN_SINK_READY_READ_TIMEOUT))
+        .ok();
+    let ready = recv_ready(&mut stream).context(not_ready_context);
+    stream.set_read_timeout(None).ok();
```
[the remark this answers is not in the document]
michaelneale: Good catch — fixed in cfa9bf0a. Both the `set_read_timeout(Some(..))` and the `set_read_timeout(None)` clear now propagate with `?` in `connect_lane_once` (frontend.rs) and `open_return_sink_once` (direct_return.rs). The clear is the important one: since these streams are kept and reused for generation, a silently-failed clear would have leaked the short handshake deadline into later reads and truncated long generations.

Note: `_timeout_secs` on `open_prediction_return_stream` was already an unused parameter on `main` (the fn hardcodes the 2s connect and now relies on the single ready-read deadline); removing it cascades through `configure_prediction_return_stream`/`send_one_off_direct_return`, so I left that signature cleanup out of this PR as orthogonal.

ndizazzo (approved):
@michaelneale 
Image

PR #1018, opened 2026-07-17
Auto-load the native runtime for SDK/embedded serving on startup

### LINKED ISSUE #1016: SDK docs: document native-runtime version-check pattern; clearer error on embedded serve
## What's actually here

Minor DX gap, not a breakage.

When MeshLLM is used as a library (`mesh-llm-sdk` `serving` / embedded serve),
the embedded serve startup path does **not** run the native-runtime load that
the shipped binary runs before serving. If a consumer starts embedded serving
without a compatible native runtime cached for the linked version/ABI, the
failure surfaces late as a cryptic Skippy FFI load error rather than a clear,
actionable message.

## Why this is low priority

Real SDK consumers already handle this correctly. For example, buzz checks the
cache for a runtime matching `CURRENT_MESH_VERSION` (which changes on SDK
upgrade), bails early with a clear message + explicit install step, and calls
`initialize_host_runtime()` itself before serving. So the upgrade case works
today and there is no user-facing bug.

## Reasonable scope if addressed

Documentation-first, not new config:

- Document the recommended SDK pattern (cache-version check + explicit
  `install_native_runtime`, as buzz does) in the SDK README.
- Optionally, have the embedded serve path surface a clear "no compatible native
  runtime installed; run install" error instead of letting it fall through to a
  raw FFI load failure.

Explicitly **not** wanted: implicit network downloads on library startup.

Closing candidate if we don't want to invest in the docs/error-message polish.

michaelneale:
## What you can now do

If you embed MeshLLM as a library (`mesh-llm-sdk` with the `serving` feature —
`start_embedded_node` / `start_embedded_serve` / `MeshNode`), the correct native
runtime is now loaded — and auto-installed if missing — automatically when the
embedded node starts. When you bump your `mesh-llm-sdk` dependency to a new
version, the matching upgraded native runtime is fetched and loaded on the next
start, just like it already is for the shipped `mesh-llm` binary. You no longer
have to call `install_native_runtime()` yourself before serving.

## The problem

Native runtimes are release artifacts selected by exact Skippy ABI +
platform/backend, and the compiled-in release version drives the versioned
release-manifest URL and cache layout. The bootstrap that makes upgrades "just
work" (`initialize_host_runtime_with_config`) loads a compatible cached runtime
or does a one-shot auto-install of the right one.

The shipped binary runs that bootstrap before serving. The SDK/embedded serving
path (`start_embedded_node` → `run_embedded_runtime` → `run_runtime_cli`) never
did — it only reset tracing and the shutdown flag. So an app that upgraded its
`mesh-llm-sdk` dependency linked a new Skippy ABI / release version but never
loaded or auto-installed the matching runtime, and serving failed later at the
Skippy FFI load unless the app manually installed the runtime first (as the SDK
README documented).

Fixes #1016.

## Change

`run_embedded_runtime` now runs the same `initialize_host_runtime_with_config`
bootstrap as the binary before serving, deriving the config path from the
embedded options so a configured `[runtime.native_runtime]` override still
applies. The call is routed through a small injectable seam
(`bootstrap_embedded_native_runtime`) so behavior can be unit-tested without
standing up a full runtime.

## Validation

- `cargo test -p mesh-llm-host-runtime --lib embedded_native_runtime_bootstrap` — 3 new tests pass
  - forwards the embedded config path to the initializer
  - forwards an absent config path
  - propagates initializer failure (embedded serving fails startup instead of failing later at FFI load)

## Summary by CodeRabbit

* **Bug Fixes**
  * Embedded and SDK serving now correctly initialize the configured native runtime before startup.
  * Runtime configuration overrides are consistently honored during embedded startup.
  * Startup failures from native runtime initialization are surfaced correctly.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/runtime/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/mod.rs
@@ -3564,10 +3564,39 @@ pub(crate) async fn run_cli(
 pub(crate) async fn run_embedded_runtime(mut options: EmbeddedRuntimeOptions) -> Result<()> {
     initialize_embedded_runtime_entrypoint()?;
 
+    // Embedded/SDK serving must bootstrap the native runtime exactly like the
+    // shipped binary does before dispatch. Without this, an SDK consumer that
+    // upgrades the `mesh-llm-sdk` dependency links a new Skippy ABI / release
+    // version but never loads or auto-installs the matching native runtime, so
+    // serving fails later at the Skippy FFI load. Honor the embedded config so
+    // a configured `[runtime.native_runtime]` override still applies. See
+    // https://github.com/Mesh-LLM/mesh-llm/issues/1016.
+    bootstrap_embedded_native_runtime(&options, |config_path| async move {
+        crate::initialize_host_runtime_with_config(config_path.as_deref()).await
+    })
+    .await?;
+
     let surface = options.runtime_surface();
     let control_rx = options.control_rx.take();
     let options = options_from_embedded_options(options);
     run_runtime_cli(options, Some(surface), None, control_rx).await
+}
+
+/// Run the native-runtime bootstrap for the embedded/SDK serving path, deriving
+/// the config path from the embedded options so a configured
+/// `[runtime.native_runtime]` override is honored.
+///
+/// The initializer is injected so the ordering/config-path contract can be
+/// exercised in tests without standing up a full runtime.
+async fn bootstrap_embedded_native_runtime<Init, Fut>(
+    options: &EmbeddedRuntimeOptions,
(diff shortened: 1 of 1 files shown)


michaelneale:
Closing. On review this isn't needed: real SDK consumers (e.g. buzz) already handle upgrades correctly by checking the cache for a runtime matching CURRENT_MESH_VERSION and failing with a clear message + explicit install step. The original auto-download-on-startup approach was wrong (a library shouldn't do implicit network I/O on startup), and the revised deny-by-default variant is only a minor DX nicety, not a fix for any real breakage. Not worth the config-schema addition and cross-crate plumbing.

Issue #1025, opened 2026-07-20
WAN split throughput: engage async verify-window pipelining aggressively (the ~10x latency-hiding lever)


michaelneale:
## Summary

Our WAN split-serving throughput is stuck near the **latency floor** (~15–20 tok/s on a 2-node ~25ms-RTT split), and I believe the root cause is that **asynchronous verify-window pipelining effectively never engages** — so every token still pays the full pipeline round-trip in the critical path. This is the single highest-leverage WAN optimization available to us, and we already have most of the primitives; the gap is a conservative admission gate plus defaults tuned for a datacenter mindset.

This proposal draws directly on the [`leyten/shard`](https://github.com/leyten/shard) work, which served **GLM-5.2 (744B) at ~30 tok/s across 6 prosumer GPUs in different US states over the open internet** (WAN edge RTTs 22–75ms), and **gpt-oss-120B at ~40 tok/s across 3 scattered 4090s**. Their measured climb is the map we should follow.

## The core insight (why this matters)

Over WAN, **the round-trip is the scarce resource, not compute.** There are two independent axes:

1. **Proposer quality** (ngram / native MTP / draft model) → sets how many tokens commit *per traversal* (acceptance rate). We've spent most of our effort here.
2. **Async pipelining** (verify chunks in flight at once) → decides *whether WAN latency is in the per-token critical path at all*.

Shard's measured ladder on GLM-5.2 is unambiguous about which axis dominates:

| Step | tok/s | What changed |
|---|---|---|
| plain pipeline decode | 1.87 | latency-bound, 1 token / round-trip |
| + deep-draft spec-decode | 1.99 | one traversal commits several tokens |
| + ring direct-return | 2.94 | tail → coordinator in 1 hop, not relayed back |
| **+ async pipelining** | **16.6** | **overlap many verify traversals → throughput-bound; WAN drops to ~5% of the loop** |
| + CUDA-graphed draft | ~30 | once WAN hidden, make the now-dominant draft cheap |

The **2.94 → 16.6** jump is the whole game, and it is **independent of the proposer**. Shard used a draft model, not MTP — meaning **we can demonstrate this with ngram on models we already have**, no big-model or MTP dependency required to prove the mechanism.

## Where mesh-llm stands (primitives vs gap)

Good news: most of this already exists after #938.

| shard mechanism | mesh-llm status |
|---|---|
[...]

PR #938, opened 2026-07-01
Pipeline MTP-anchored n-gram verify windows

i386:
## Summary

Skippy can now use native MTP, upstream llama.cpp N-gram proposal, or a composite of both while preserving target verification for every committed token. The old synchronous `VerifySpan` hot path is replaced by `VerifyWindow`, and deeper asynchronous verification is enabled only when request-local evidence predicts that overlap will repay stale speculative work.

This also makes speculative strategy a product configuration surface. Model packages can select native MTP, simple prompt lookup, or stateful cache N-gram behavior, while `mesh-llm` and `skippy-server` can apply deployment overrides.

## Architecture

- Replaces internal `VerifySpan` messages and hot-path machinery with versioned `VerifyWindow` request/reply messages. No internal compatibility lane remains.
- Represents a proposal as zero-or-more native MTP prefix tokens followed by zero-or-more N-gram tail tokens.
- Uses N-gram as a fallback proposal source when native MTP produces no token.
- Verifies the combined linear stream in one target operation and commits the valid prefix when a later token rejects.
- Uses llama.cpp's upstream simple prompt lookup and stateful N-gram cache ABIs instead of a separate in-tree N-gram implementation.
- Treats `verify_window_pipeline_depth` as a maximum. A rolling exact-width profile estimates useful continuation overlap versus expected stale work before issuing a dependent future window.
- Keeps synchronous batched `VerifyWindow` verification when the observed shape is not profitable.
- Exposes composite, native-prefix, N-gram-tail, pure-fallback, adaptive-width, pipeline-policy, and stale-discard telemetry.
- Charges native MTP tensors to the final logits-owning stage during direct GGUF split planning.

## Product Configuration

- Adds declarative speculative controls to `model-package.json`, built-in config schema, validation, and package preflight.
- Supports `simple`, `cache`, and disabled N-gram proposal modes.
- Carries package defaults through mesh split resolution into embedded Skippy stage options.
- Adds matching `mesh-llm` and `skippy-server` configuration and CLI documentation.
- Adds a model-agnostic direct N-gram strategy for models without native MTP.

## Protocol

This intentionally breaks the internal Skippy stage decode protocol. Old and new staged-runtime binaries do not interoperate. Public mesh gossip, the OpenAI-compatible API, package selection, and mixed-version public mesh behavior are unchanged.

## Lab Validation

[...]

Issue #936, opened 2026-07-01
latency hiding enhancements for skippy


michaelneale:
From analysis of shard: 

# Decode Latency Enhancement Notes

These notes capture the current gap between Skippy staged decode and the
latency-hiding decode loop prototyped in the Shard repo. Shard references point
to `leyten/shard`; Skippy and mesh references point to `Mesh-LLM/mesh-llm`.

## Main Gap Versus Shard

Shard's `coordinate_pipe` keeps multiple verify chunks in flight, overlaps draft
work with WAN traversal, and discards stale results after divergence. That is the
actual latency-hiding engine:

- [`coordinate_pipe`](https://github.com/leyten/shard/blob/c4588bf846d422c99b4b8249d3aec9d155e01c56/phase0/specpipe.py#L644)
- inflight send/read/discard loop in [`specpipe.py`](https://github.com/leyten/shard/blob/c4588bf846d422c99b4b8249d3aec9d155e01c56/phase0/specpipe.py#L724)

Skippy's decode path still follows a synchronous loop for each decode or verify
span: compute stage 0, forward the activation, wait for a downstream reply, then
decide the next token or next speculative window. The serial wait point is in
[`embedded_generation.rs`](https://github.com/Mesh-LLM/mesh-llm/blob/232c85a86a1bc809edaeb7966fd2c3d458e5438c/crates/skippy-server/src/frontend/embedded_generation.rs#L1496).

## Latency Strategy

For Skippy, the immediate path is:

1. Amortize each downstream wait over more committed tokens.
2. Then make decode windows concurrent.

The current knobs already available or partially available are:

- staged serving;
- direct prediction return;
- Q8/F16 activation wire dtype;
- prefill reply credits;
- prefix and exact replay cache;
- native MTP where package generation metadata supports it;
- external draft mode with a good same-family GGUF.

The next Shard-level step is a Skippy decode scheduler:

- maintain a FIFO of in-flight `VerifySpan` requests per session;
- require direct prediction return for pipelined decode;
- send up to a configured `depth`;
- read replies in order;
- drop stale replies after a rejection;
- re-prime draft generation from the corrected prefix;
- preserve target correctness by verifying every committed token.

This should be implemented as backend-serving machinery in `skippy-server` and
kept behind the existing embedded OpenAI/runtime APIs. Do not make mesh's public
OpenAI surface depend on a standalone draft or n-gram service.

## Predictor-Agnostic Decode

[...]

PR #1026, opened 2026-07-20
🚬 Add bounded MTP + N-gram pipelining for latency-limited split inference

i386:
## What changed

This replaces the earlier repair-oriented experiment with a deliberately breaking, Shard-shaped stage protocol for **native GLM-4.7 MTP + request-local N-gram speculation**. There is no standalone draft model or draft-model experiment in this PR.

- stage protocol v10 makes stage zero the sole acceptance authority;
- full, contiguous multi-token verification chunks use authoritative absolute positions;
- a fixed per-request FIFO stays filled to `verify_window_pipeline_depth`;
- stage forwarding is asynchronous and the tail returns replies directly to the coordinator;
- replies retire in FIFO order; divergence discards dependent work and immediately starts a new epoch;
- fully accepted chunks advance KV/MTP state monotonically without accepted-path rewind or replay;
- pipeline prefill and request-local N-gram horizon refill keep work available;
- follower stages inherit their node-local telemetry policy, and stage zero now emits joinable target-compute spans;
- the old MTP checkpoint ring, repair/replay messages, adaptive depth policy, probation/promotion/slow-start controller, global speculative-credit pool, and their configuration/telemetry were deleted.

The public mesh and OpenAI protocols are unchanged. The internal stage protocol is intentionally incompatible; all stages in a split must run v10.

## Canonical benchmark gate: NVIDIA SPEED-Bench under a locked topology

The tables below are controlled diagnostic microbenchmarks, not the final headline benchmark. Before this is promoted or merged, the performance claim will be rerun with the full upstream NVIDIA SPEED-Bench qualitative split (all 880 samples and all categories, OSL 1024, temperature 0) through the OpenAI-compatible endpoint.

Both A/B arms will use the locked-topology implementation merged from #1050 / `jd/locked-split-topology`:

- immutable GLM-4.7 package revision and `model-package.json` SHA-256;
- identical lock bytes and lock SHA-256 on both hosts;
- two immutable topology files with exact node order and contiguous ranges:
  - primary balanced topology: `micstudio.local:0..25` then `studio54-3.local:25..48` (25/23);
  - asymmetric latency stress: `micstudio.local:0..47` then `studio54-3.local:47..48` (47/1);
- `--split-topology-lock` passed to both nodes;
- realized topology verified through `/api/runtime/stages` before traffic starts;
- fail closed on package, participant, capacity, range, or stage-loss mismatch.

The only serving-policy difference between matched runs will be native MTP versus native MTP + request-local N-gram width 4 / fixed depth 2. The locked 25/23 split is the primary production-like comparison; the locked 47/1 split is explicitly an asymmetric latency-hiding stress case, not the headline topology. Full qualitative runs use request concurrency 1 at natural LAN latency and at 100 ms injected stage latency. Concurrency 8 is a separate saturation run on the balanced split. The report will include per-category and overall decode TPS, end-to-end latency, TTFT, ITL, acceptance, in-flight occupancy, stale work, and joined stage-compute overlap. The SPEED-Bench throughput split will separately test low-, mixed-, and high-entropy load.

## Current diagnostic result

Two-node GLM-4.7 Flash MTP Q4_K_M split, 47/1 layers, eight lanes, F16 activations, 32K context, 100 ms injected inter-stage delay. The selected setting is N-gram 4, verification width 4, fixed depth 2.

| Mixed corpus, c1 | Native MTP | MTP + N-gram, depth 2 | Change |
|---|---:|---:|---:|
| Model-reported decode TPS | 9.3152 | **10.7097** | **+14.97%** |
| Aggregate wall TPS | 7.2815 | **8.0852** | **+11.04%** |
| Draft acceptance | **77.35%** | 38.51% | -38.84 pp |
| Exact output hashes | 4/4 | 4/4 | equal |

This is the result we set out to prove: **MTP + N-gram beats native MTP even though its acceptance rate is substantially lower.** Acceptance is not throughput; overlapping useful target work can outweigh rejected proposals.

On a separate four-prompt novel-code corpus, selected depth 2 was essentially neutral: 8.7018 versus 8.7921 decode TPS (-1.03%), with 4/4 exact hashes and 32.8% acceptance. The win is therefore workload-dependent rather than a universal claim.

## Concurrency

Same mixed corpus and selected depth-2 configuration:

| Request concurrency | Native MTP wall TPS | MTP + N-gram wall TPS | Change | Hybrid acceptance |
|---:|---:|---:|---:|---:|
| 1 | 7.2815 | **8.0852** | **+11.04%** | 38.51% |
| 2 | 11.6076 | **15.1921** | **+30.88%** | 35.19% |
| 4 | 18.4284 | **22.8768** | **+24.14%** | 33.90% |
| 8 | **27.0188** | 26.1993 | **-3.03%** | 31.42% |

Depth is intentionally fixed per request. At c8, depth 2 can place 16 windows against eight lanes and slightly overfill the service. This is the current operating-envelope limit; the deleted global-credit/adaptive machinery is not being smuggled back in. Production policy should choose depth with total admitted concurrency in mind or batch callers into shared pipeline jobs as Shard does.

## Does speculation fill the pipeline?

Yes, on eligible repeated spans. At depth 2 and 100 ms injected delay:

| Occupancy / latency signal | Result |
|---|---:|
| Time-weighted average windows in flight | **1.92 / 2** |
| Time at full configured depth | **94.3%** |
| Stale windows on the fully accepted repeat case | **0** |
| Verify elapsed accumulated across windows | 15.664 s |
| End-to-end decode interval | 8.242 s |
| Mean streaming ITL, native MTP | 83.10 ms |
| Mean streaming ITL, hybrid | **22.24 ms** |
| Mean ITL reduction | **73.2%** |
| p95 ITL, native MTP / hybrid | 169.29 / **127.92 ms** |
| Sub-millisecond token intervals, native MTP / hybrid | 49.87% / **74.67%** |
| TTFT, native MTP / hybrid | 5.295 / 5.309 s |

The queue is not merely configured to two; it is measurably full for almost the entire eligible interval. Accumulated verification time exceeding wall decode time and the burstier ITL distribution show overlapped in-flight target work. TTFT is unchanged because this optimization starts after prefill.

### How much are the stages actually computing in parallel?

Debug traces now join the stage-0 and stage-1 target-compute spans by run, request, session, and window. Each trace matched all 97 windows on both stages; every simultaneous-compute pair was from different windows, which is the pipeline behavior we wanted to observe.

| Split | Verify interval | Stage 0 busy | Stage 1 busy | Both stages busy | Share of S0 / S1 compute overlapped | Different-window overlaps |
|---:|---:|---:|---:|---:|---:|---:|
| 47/1 | 8.284 s | 3.351 s (40.5%) | 1.551 s (18.7%) | **0.516 s (6.23%)** | 15.4% / **33.3%** | 43 |
| 25/23 | 11.586 s | 1.953 s (16.9%) | 8.268 s (71.4%) | **0.520 s (4.49%)** | **26.6%** / 6.3% | 117 |

So the precise answer to the review question is nuanced: **the inference FIFO is full and hides stage/network latency, but both stage runtimes are simultaneously computing for only 4.5–6.2% of the verify interval at depth 2.** The 47/1 tail is intentionally tiny; in the 25/23 split, the slower stage 1 is busy 71.4% of the interval while stage 0 works ahead in shorter bursts. The gain comes from keeping the bottleneck supplied, overlapping different windows, hiding the ~100 ms edge delay (median same-window stage-0-end to stage-1-start was 113–117 ms), and amortizing traversals with wider chunks—not from both GPUs being continuously busy together.

These are host-measured synchronous stage-compute spans, not Metal hardware-counter samples. They measure when each stage runtime is executing target work; they do not claim cycle-level shader occupancy.

## Network-latency sweep

Highly repeatable 384-token workload, 47/1 split, width 4, depth 2; exact output in every row:

| Injected delay | Native MTP decode TPS | Hybrid decode TPS | Change | Hybrid avg/full occupancy |
|---:|---:|---:|---:|---:|
| 0 ms | 39.8221 | **103.9520** | **+161.0%** | 1.90 / 93% |
| 20 ms | 26.3357 | **97.1255** | **+268.8%** | 1.92 / 95% |
| 50 ms | 18.5264 | **68.0052** | **+267.1%** | 1.93 / 96% |
| 100 ms | 12.2195 | **46.5880** | **+281.3%** | 1.92 / 94% |

The delay between stages is therefore substantially hidden when the request-local horizon can keep the FIFO supplied. It is not eliminated: TTFT/prefill remain serial, divergence wastes later work, and low-reuse workloads cannot sustain the horizon.

## Depth and width frontier

The same 100 ms repeat workload establishes the upper bound and where added depth stops buying much throughput:

| Fixed depth | Decode TPS | Wall TPS | Exact | Acceptance |
|---:|---:|---:|:---:|---:|
| Native MTP | 12.2195 | 10.5183 | yes | 100% |
| 1 | 28.1010 | 20.5081 | yes | 100% |
| 2 | 46.8637 | 28.9054 | yes | 100% |
| 4 | 91.6338 | 41.2619 | yes | 100% |
| 8 | 105.5877 | 43.9471 | yes | 100% |
| 16 | **110.2350** | **44.4998** | yes | 100% |

Throughput approaches the bottleneck service rate around depth 8; depth 16 adds only 4.4%. Depth 2 is selected for mixed traffic because it captures overlap without the concurrency/stale-work exposure of a deep per-request queue.

At depth 2, the width sweep separates pipeline overlap from wider target chunks:

| Verification width | Decode TPS | Wall TPS | Avg/full occupancy |
|---:|---:|---:|---:|
| 1 | 13.4691 | 11.4329 | 1.97 / 98.2% |
| 2 | 25.3917 | 19.0012 | 1.94 / 95.9% |
| 4 | **46.5880** | **28.9208** | 1.92 / 94.3% |

Width 1 is already +10.2% over native MTP at 100 ms, showing that overlap alone helps. Full chunks provide the larger gain by amortizing each traversal across more verified tokens.

## Topology

The effect is not limited to the asymmetric 47/1 placement:

| Split | Native MTP decode TPS | Hybrid decode TPS | Change |
|---:|---:|---:|---:|
| 47/1 | 12.2195 | **46.5880** | **+281.3%** |
| 25/23 | 10.0907 | **33.3145** | **+230.1%** |

Both rows use the 100 ms repeat workload, width 4, depth 2, and exact output comparison.

## Comparison with the previous runs

These are separate run families, so the percentage deltas are the honest comparison—not the absolute TPS values.

| Run family | c1 | c2 | c4 | c8 | Design |
|---|---:|---:|---:|---:|---|
| PR #938 reported hybrid uplift | +5.57% | — | — | — | Earlier MTP/N-gram experiment; different run shape |
| PR #938 frozen-control uplift | +11.87% | — | — | — | Earlier control rerun; different run shape |
| Previous #1026 v9 | **+27.07%** | +13.22% | +0.74% | +0.23% | Overlapping one-position windows, checkpoint repair, adaptive/credit controls |
| This v10, selected mixed run | +11.04% wall / +14.97% decode | **+30.88%** | **+24.14%** | -3.03% | Full chunks, fixed-depth FIFO, contiguous positions, no accepted-path repair |

The new path trades the old c1 peak for much stronger c2/c4 scaling and a far simpler execution model. It also exposes the real c8 overfill limit rather than obscuring it with a global credit controller.

## How similar is this to Shard / issue #1025?

This now copies Shard's important execution pattern much more directly:

- explicit fixed depth rather than an inferred magic concurrency of two;
- prefill the depth, retire FIFO, and refill immediately;
- independent in-flight target traversals at absolute positions;
- asynchronous stage forwarding and direct final-stage return;
- invalidate later work after divergence instead of repairing accepted work;
- choose the shallowest depth that covers the pipeline service latency.

The important differences are:

- Shard has an independent draft actor; this path deliberately uses only native GLM-4.7 MTP plus a request-local exact N-gram continuation;
- Shard's gateway batches concurrent callers into shared ring jobs; this implementation still assigns a fixed FIFO depth per admitted request;
- Shard also requires an explicit depth—it does not discover an optimal per-request depth automatically.

That answers the central questions in #1025: parallel speculation can hide stage/network latency when the candidate horizon stays supplied; depth must cover the pipeline rather than be hard-coded to concurrency two; and production concurrency still needs shared batching/admission so independent per-request queues do not overfill the lanes.

## Validation

- `cargo fmt --all --check`
- `cargo test -p skippy-protocol --lib` — 43 passed
- `cargo test -p skippy-server --lib` — 281 passed
- `cargo test -p mesh-llm-config --lib` — 95 passed
- `cargo test -p mesh-llm-host-runtime --lib` — 1,695 passed, 8 ignored
- warning-denying Clippy across all touched crates and `mesh-llm`
- `MESH_LLM_DYNAMIC_NATIVE_RUNTIME=0 just release-build`
- merged `jd/locked-split-topology` / #1050 into this branch;
- clean patch-queue application at llama.cpp `178a6c44937154dc4c4eff0d166f4a044c4fceba`;
- `cargo test -p skippy-coordinator --lib` — 37 passed;
- `cargo test -p mesh-llm-cli --lib` — 72 passed;
- `cargo test -p mesh-llm-host-runtime --lib split_topology` — 10 passed;
- `cargo check -p mesh-llm` and warning-denying Clippy across the topology-touched crates;
- two-host GLM-4.7 split output-hash comparisons across mixed, code, repeat, depth, width, delay, topology, and concurrency sweeps

## Production limitations

- Depth is fixed per request; c8 shows that total depth across callers must be bounded or callers must be coalesced into shared jobs.
- Stage overlap is measured from synchronous runtime spans, not GPU performance counters; hardware-counter validation remains separate.
- The large latency/depth gains are an upper bound on highly reusable text; novel code is neutral and arbitrary low-reuse traffic may not form a deep horizon.
- Divergence correctness is covered by the coordinator/stale-discard path, but a longer adversarial rejection/stop/tool-call soak is still warranted before making this a default package policy.

Addresses the implementation questions raised in #1025; package-default promotion should remain a separate production decision.


i386:
@michaelneale Follow-up plan for answering the parallel-busy-time / WAN-latency question and bringing this closer to Shard.

The positional protocol in this PR is the prerequisite: verify windows carry authoritative positions, stages repair KV locally, and rejection no longer requires checkpoint/restore ACKs or repair/reverify. The remaining work is to turn the current depth-2 burst into a continuously refilled pipeline and prove that it hides latency.

A useful sizing condition is:

`required_depth >= ceil(end-to-end verify latency / verify-window launch interval)`

Once that many independent windows are circulating, propagation delay becomes pipeline fill rather than an inter-token stall; throughput approaches the slowest stage/link service rate. Initial fill, rejection recovery, and link serialization still remain visible.

Implementation plan:

1. **Remove remaining serialization**
   - Replace the 10 ms direct-return polling loop with a blocking, timeout-aware receive.
   - Extend bounded asynchronous activation forwarding from prefill to `VerifyWindow` traffic at stage 0 and intermediate stages.
   - Measure send backpressure and size flow-control capacity for `depth × activation_frame_bytes`.

2. **Use a continuously refilled credit window**
   - Fill all available credits, consume the oldest reply, then immediately refill the freed credit.
   - Tag windows with a speculation epoch and absolute position.
   - On divergence, mark later windows stale and begin the corrected branch as credits reopen; do not stop to synchronously drain the entire stale queue.

3. **Generate enough speculative future**
   - Keep Ngram order as the matching-quality control, but decouple it from speculative horizon.
   - Use MTP for the high-confidence prefix and recursively extend the speculative prefix with the stateful Ngram cache until the target pipeline depth or stale-work budget is reached.
   - This supplies the continuous cheap draft stream needed for depths 4/8.

4. **Make depth latency-driven**
   - Estimate `target_depth = ceil(EWMA(verify_latency) / EWMA(launch_interval)) + safety_credit`, clamped to an operator maximum.
   - Ramp 1→2→4→8 while downstream wait remains exposed; shrink when stale compute exceeds saved wait.
   - Judge profitability in time: `hidden_wait_ms > stale_compute_ms × margin`, allowing lower acceptance when wall TPS and latency still improve.

5. **Measure real occupancy**
   - Time at each in-flight depth and percentage of decode time at full depth.
   - Per-stage compute-busy, send-blocked, and input-idle time.
   - Exposed versus hidden downstream wait, stale compute/bytes, and inter-token p50/p95.
   - Rejection-recovery bubbles and pipeline warm-up separately from steady state.

6. **Controlled proof**
   - GLM-4.7 split at injected 1/10/25/50/100 ms latency.
   - Depths 1/2/4/8, MTP versus MTP+Ngram, with long enough generations for steady state.
   - Compare against the existing PR baseline: MTP 42.85 wall TPS / 8.961 s mean latency versus MTP+Ngram 52.49 wall TPS / 7.316 s mean latency on LAN.

The success signal is not merely reaching depth 8. It is stage idle time collapsing as depth grows, exposed downstream wait approaching zero, TPS plateauing at slowest-stage throughput, and additional depth increasing stale work without improving throughput.

This is the same core loop as [Shard's pipelined coordinator](https://github.com/leyten/shard/blob/master/phase0/specpipe.py#L718-L838): direct return, multiple causally speculative chunks in flight, FIFO completion, and stale-result discard after divergence. The comparison we need to prove is whether our positional repair plus MTP+Ngram can achieve the same WAN latency hiding without reintroducing rollback/recovery serialization.

michaelneale:
One thing worth being explicit about for the WAN goal (#936 / #1025): this reaches depth **2**, but not the deeper "keep the pipe full" regime shard uses, because the proposer is still **coupled** to the verify traversal. The next MTP draft comes from the target's own forward (`NativeMtpDraftOrigin::VerifyNext`) into a single `next_draft` slot (`native_mtp/pipeline.rs`), so proposal N+1 can't exist until reply N returns — a structural depth-2 cap, independent of `permit_pipeline_width`'s profitability gate.

Shard's `coordinate_pipe` fills to `depth` because its predictor runs **ahead, decoupled** from verification (it optimistically appends its own proposals and drafts forward; a divergence just discards the stale in-flight chunks). #936's "Predictor-Agnostic Decode" section is the same idea: a `PredictionProvider` boundary where some source proposes a span the target verifies, and n-gram / prompt-lookup fits it precisely because it needs only the committed prefix, no target reply.

i386:
@michaelneale The structural depth-two cap you identified is removed in `c06bca6df`.

The important clarification is that **two is not configured or steady-state concurrency**. It is now named and emitted as `verify_window_bootstrap_probe_depth`: depth one cannot tell us whether dependent work overlaps verification, so two is the smallest useful exploration point. After that, active depth is selected from `ceil(mean verify elapsed / mean launch interval)`, bounded by the operator cap and recent continuation/stale-work profitability.

The proposer is no longer blocked on the single `VerifyNext` MTP slot. MTP remains the high-confidence anchor, but request-local stateful N-gram extends the optimistic branch and can seed work while an MTP reply or reject cooldown is outstanding. Credits are continuously refilled; replies commit FIFO; later epoch-tagged replies become stale after divergence without blocking the corrected branch on rollback or a synchronous drain. That is the predictor-ahead property needed by #1025 and Shard's `coordinate_pipe`.

On the two-node GLM-4.7 split with 100 ms injected one-way downstream delay, the four-prompt matched result was:

| | MTP depth 1 | MTP + N-gram, adaptive cap 8 |
|---|---:|---:|
| Decode TPS | 11.53 | **29.30 (+154.19%)** |
| Wall TPS | 11.38 | **28.33 (+149.05%)** |
| Mean latency | 33.756 s | **13.554 s (-59.85%)** |
| Acceptance | 91.23% | 69.45% |
| Max selected / actual in-flight depth | 1 | **7** |
| Decode time with >=2 windows in flight | 0% | **34.66%** |
| Time at the controller's current target | 100% at serial target | **24.12%** |

The exact final binary canary then reached selected and actual depth **8**, spent **47.71%** of decode time with multiple windows in flight, and produced 30.75 decode TPS / 13.076 s latency.

So the answer to “are both sides busy in parallel / is latency hidden?” is: **materially, but not completely**. At 100 ms the active depth grows to seven/eight and cuts request latency by ~60% in the matched run. Propagation overlaps while credits are full. Startup, proposal gaps, divergence recovery, bandwidth serialization, and the first corrected reply still leave bubbles; the 35-48% parallel-occupancy result tells us exactly how much room remains. On LAN the controller only reached actual depth two and spent 17.69% parallel, which is the expected behavior when deeper speculation does not repay itself.

This is now directly comparable to [Shard's pipeline loop](https://github.com/leyten/shard/blob/master/phase0/specpipe.py#L718-L838): bounded fill, predictor-ahead speculative chunks, FIFO commit, stale discard. Shard is given a depth bound; it does not infer the economically optimal depth from a serial trace. The extra piece here is the latency/stale-cost controller around that loop, plus positional repair instead of rollback.

The PR description now contains the full LAN, 100 ms, prior-baseline, and #938 comparison tables and the validation record.

i386:
@michaelneale This diagnosis was exactly right for the implementation at the time of the comment: a target-produced `VerifyNext` MTP draft cannot run ahead of its own verify traversal, so MTP alone has a structural depth-two ceiling.

The follow-up at `55db3881d` removes that ceiling from the hybrid path:

- N-gram extends the optimistic prefix while older verifies are outstanding; it no longer waits for the next target MTP draft.
- depth 2 is only the first identifiable overlap probe from depth 1, not configured concurrency.
- the controller derives target depth from verify latency / launch interval, shares latency evidence across adaptive verify widths, and throttles when continuation-weighted stale cost is unprofitable.
- FIFO commit and epoch-tagged stale discard match Shard's useful pipeline mechanism; positional trim replaces rollback.

I also found and fixed a correctness bug while tracing this: batched stage-0 serial correction could execute on the rejected suffix because alignment and decode were not atomic. All earlier `c06bca6df` performance figures are withdrawn. MTP and hybrid now produce identical per-prompt output hashes.

Corrected two-node GLM-4.7 split, controlled 100 ms one-way delay, 3 matched repetitions / 1,152 generated tokens per condition:

| Condition | Decode TPS | Mean latency | Acceptance | Max selected / actual | Time >=2 in flight |
| MTP depth 1 | 10.340 | 23.976 s | 83.68% | 1 / 1 | 0% |
| MTP + N-gram cap 8 | **13.715** | **19.388 s** | 52.80% | **6 / 3** | **15.79%** |

That is **+32.65% decode TPS and -19.14% latency despite 30.88 points lower acceptance**. The hybrid proposed 1,284 tokens and accepted 678, versus MTP's 570 proposed / 477 accepted: lower acceptance ratio, but more useful speculative work completed.

The busy-time answer is workload-dependent and the new description separates it explicitly:

- prompt copy: 11.485 -> **31.109 decode TPS**, selected depth up to 6, actual max 3, **57.25% parallel time**, 81/81 profitable continuations;
- novel code: 9.402 -> 8.797 decode TPS, zero profitable continuations, 4.06% parallel time, controller falls back to depth 1.

So this now runs ahead independently in the Shard sense, but it does not claim continuous saturation: the bounded N-gram horizon plus fill/drain still leave bubbles. It fills enough to hide substantial latency when the text is draftable and refuses to manufacture stale traffic when it is not.

One nuance on “doesn't Shard figure depth out?”: current Shard has an optional acceptance-EMA throttle from 1 to a configured maximum. It starts optimistic at that maximum; it does not infer the maximum from measured traversal latency. This PR adds that latency-derived target on top of the refill/FIFO/stale-discard loop.

The PR description now contains the corrected results, #938 comparison, Shard source comparison, and exact validation details.

i386:
@michaelneale Final follow-up to the structural coupling point in [your comment](https://github.com/Mesh-LLM/mesh-llm/pull/1026#issuecomment-5019834065): `0b3720c90` now removes the remaining proposal-boundary wait as well.

The hybrid no longer has to drain all active windows when one bounded proposal runs out. Before consuming the current horizon, it queries the committed-only N-gram index with the complete optimistic suffix (including dispatched-but-uncommitted tokens), appends any continuation, and immediately spends free verify credits. The suffix is never indexed until target commit. FIFO commit, epoch stale-discard, and positional repair are unchanged.

Fresh 3-process matched GLM-4.7 A/B at controlled 100 ms one-way delay (6 requests / 1,152 generated tokens per condition):

| | MTP depth 1 | Refilled MTP + N-gram cap 8 |
|---|---:|---:|
| Decode TPS | 10.362 | **14.136 (+36.41%)** |
| Wall TPS | 8.020 | **10.113 (+26.10%)** |
| Mean latency | 23.941 s | **18.985 s (-20.70%)** |
| Acceptance | 83.68% | **45.10%** |
| Max selected / actual | 1 / 1 | **5 / 3** |
| Time with >=2 windows active | 0% | **15.44%** |

Outputs matched MTP hashes in all repetitions. Relative to the corrected pre-refill hybrid baseline already on this PR, this is another **+3.06% decode TPS / +2.12% wall TPS / -2.07% latency**.

The busy-in-parallel answer is now measured rather than inferred:

- draftable prompt-copy: actual depth 3, 60.64% parallel time, 35.177 vs 11.506 decode TPS, zero stale windows;
- novel code: controller ends at depth 1, 4.07% parallel time, 8.845 vs 9.425 decode TPS, 21 stale windows.

At 0/50/100 ms delay, hybrid parallel time rises 46.05% -> 51.84% -> 58.35%, while its wall-TPS advantage over MTP rises +34.67% -> +61.73% -> +71.65%. That is the WAN behavior we wanted: propagation is increasingly hidden behind useful work, but fill/drain, finite predictor horizon, and serialized bandwidth remain visible.

Cap 2/4/8 on the 100 ms copy prompt reaches actual depth 2/3/3 and wall TPS 12.046/12.509/12.510. So cap 8 is not "concurrency eight" and forcing it would not help—the scheduler finds the plateau at actual depth 3. This now matches Shard's predictor-ahead fill/FIFO/stale-discard loop; the extra mechanism here is latency-derived depth plus positional repair rather than rollback.

The PR description has the full tables, #938 comparison, exact binary identity, and validation record.

i386:
@michaelneale Final production update on the “how busy are the stages / is WAN latency hidden?” question.

First, one correction to my earlier explanation: **authoritative position is enough to repair attention KV, but not the whole GLM-4.7 MTP sidecar**. The sidecar also owns a recurrent hidden prefix and pending hidden vector. We now restore those from a bounded, process-local checkpoint on speculative trim. There is still no rollback/repair message or replay on the wire.

Final matched sustained copy result at 100 ms one-way delay:

| Request concurrency | Native MTP wall TPS | Hybrid wall TPS | Change | Parallel occupancy | Behavior |
|---:|---:|---:|---:|---:|---|
| 1 | 8.920 | **11.335** | **+27.07%** | 41.03% | pipelined |
| 2 | 17.222 | **19.499** | **+13.22%** | about 45.5% / request | fairly shared |
| 4 | 27.501 | **27.705** | +0.74% | 0% optional | native fallback |
| 8 | 29.395 | **29.463** | +0.23% | 0% optional | native fallback |

So yes: on draftable c1/c2 work, multiple dependent verify windows occupy the split concurrently and materially hide the 100 ms delay. At c4/c8 the independent requests already occupy the pipeline, so the global credit policy removes optional N-gram work instead of oversubscribing it.

The representative code/reasoning/copy/tool matrix is deliberately less exciting: LAN -0.96%, 20 ms +0.82%, 50 ms -0.69%, 100 ms +1.56%, and **mixed WAN +0.61%**. That meets the aggregate no-harm goal, but **does not meet the requested +20% mixed-WAN bar**. The >=20% proof is specific to sustained, highly draftable c1 text.

Relative to Shard, the refill/FIFO/stale-discard mechanism is now similar, but our scheduling adds MTP anchoring, latency/profitability depth selection, and globally fair request credits. Shard still has broader multi-stage and resumable failure machinery. Here, killing stage 1 during prefill now returns a bounded 502 in 29.06 seconds instead of hanging; the active request is not resumed.

The PR description now has the clean final tables, #938 and previous-run comparison, correctness repair, failure results, final binary identity, and production caveat. Commit: `20e457a0c`.

michaelneale:
Let's hold off on merging this one for now — it overlaps with an active research/WIP effort on the split-runtime decode path (both rewrite `skippy-server/src/frontend.rs` and `binary_transport/direct_return.rs`), and I want to reconcile the two designs before either lands rather than merge-then-untangle.

To be clear this isn't a quality concern — the benchmark writeup is excellent and the net simplification (deleting the checkpoint ring / adaptive-depth / credit-pool machinery) is genuinely appealing. I just want to sequence it deliberately against the in-flight work. Also worth flagging: this collides internally with #1034 (skippy-ffi ABI) and #1032 (`preflight.rs`), so merge order across your own stack matters here too.

Will follow up on sequencing. Leaving open, just please don't merge yet.

i386:
@michaelneale — tagging you for review, particularly the new stage-overlap/latency evidence and the Shard comparison in the PR description.

crates/mesh-llm-ui/src/features/configuration/lib/build-toml.test.ts
```diff
@@ -239,14 +231,14 @@ describe('buildTOML', () => {
           label: 'Speculative mode',
           description: 'Speculative runtime mode.',
           inheritedLabel: 'Inherited by speculative defaults',
-          valueSchema: { kind: 'enum', values: ['draft', 'ngram'] },
+          valueSchema: { kind: 'enum', values: ['draft', 'disabled'] },
```
ndizazzo, line 234: @i386 Is this the configuration surface you want for this work? This means that the user can't select Ngram MTP via any options config.

docs/plugins/telemetry.md
```diff
@@ -148,6 +148,8 @@ to an OTLP record.
 | `mesh_llm.guardrail.outcome` | guardrail outcome | Bounded enum: `pass_through`, `valid`, `rescued`, `retried`, `failed`, or `metrics_only_failure`. |
 | `mesh_llm.guardrail.parser_stage` | guardrail outcome | Bounded enum: `none`, `json_exact`, `json_fenced`, or `json_substring`. |
 | `mesh_llm.guardrail.attempt_bucket` | guardrail outcome | Bounded retry bucket: `1`, `2`, or `3_plus`. |
+| `llama_stage.verify_window.direct_return_upstream_opened` | Skippy decode summary | Boolean indicating that the preferred upstream-opened v10 prediction-return sink completed its handshake. |
+| `llama_stage.verify_window.direct_return_reverse_fallback` | Skippy decode summary | Boolean indicating that the final stage used the bounded reverse-open v10 prediction-return fallback after the preferred sink was unavailable. |
```
ndizazzo, line 152: Somewhat broken naming convention for these - not namespaces to `mesh_llm.`

ndizazzo (commented):
Lightweight review pass - found a couple things

michaelneale:
🤖 *Pushed by micn's AI agent.*

Pushed two small WAN-hardening commits onto this branch (`87e691a4`, `34bd3b89`) — ports of the two transport fixes that were validated live on a real ~26ms WAN split (M5 Sydney ↔ AU RTX 4090) on the now-retired `wip/wan-direct-prediction-return` lab branch:

1. **`87e691a4`** — stop re-applying the formation-time `MAX_SPLIT_RTT_MS` ceiling to fresh *operational* stage streams (warn-and-proceed). Without this, per-request direct-return sinks get rejected under normal WAN RTT jitter while pooled forward lanes stay healthy → ready-handshake timeouts → every request 502s on an already-admitted split. Split *admission* still gates on gossiped, hysteresis-smoothed RTT + re-election; this only relaxes the per-stream re-check.
2. **`34bd3b89`** — return-sink ready-handshake timeout 5s→20s. Over WAN the `ready` byte only arrives after the bridge cold-establishes a fresh stage QUIC connection (~10s budget); 5s timed out during that cold setup on a healthy split. Matches the forward-lane budget.

Why: this PR's latency sweep (+161%→+281%) is injected-delay on LAN. These two fixes are what made a *real* internet split serve at all in the lab. Next step is the first real-WAN datapoint for v10: M5 (Sydney) ↔ cheap vast.ai AU 4090, GLM-4.7-Flash-MTP Q4_K_M 2-stage locked-topology split, native MTP vs MTP+ngram width 4 / depth 2. Numbers will be posted here.

Validation: `cargo fmt --all --check` clean; `clippy -D warnings` clean on `mesh-llm-host-runtime`, `skippy-server`, `mesh-llm`; `cargo test -p mesh-llm-host-runtime --lib` 1698 passed; `cargo test -p skippy-server --lib` 281 passed.

Not ported (deliberately): the lab branch's pre-warmed return-sink pool. v10 reworked `direct_return` (per-request sinks keyed by request/session), so it may not need it — will port only if the WAN rehearsal shows cold-open flakiness.

michaelneale:
🤖 *Posted by micn's AI agent (working on this branch with micn).*

## First real-WAN validation of v10 — results

Ran the first real-internet split test of this branch today: **M5 (Sydney, Apple M5 Max, stage 0) ↔ vast.ai RTX 4090 (AU, stage 1, layers 27..48)**, 21–25ms real WAN RTT, GLM-4.7-Flash-MTP Q4_K_M, 2-stage split, greedy decode, `/v1/completions`, 384 tokens, 2 warm + 5 measured per arm. Both nodes at exactly `34bd3b89` (this branch tip, including the two WAN transport commits below).

| Workload | Arm A: native MTP (median dec tok/s) | Arm B: MTP+ngram (min 2 / max 4 / depth 2) | Change |
|---|---:|---:|---:|
| repeat corpus | 35.4 | **148.1** | **+318%** |
| code re-emit w/ rename | 35.6 | **69.8** | **+96%** |

- **Outputs byte-identical between arms** (sha-256 over generated text: repeat `7ab048a22bff`, rename `3d7b1f1f7c85` — same in both arms, stable across every rep).
- Arm B repeat: occupancy 1.99 avg / **99% full-fraction**, stale=0, ngram acceptance 384/384, `direct_prediction_return=true`, `max_in_flight=2` on every run.
- Arm B rename: occupancy ~1.22 / 26% full, 2 recovery epochs — partial fill on mixed-novelty text, matching this PR's own model of when the hybrid pays.
- **Cross-check against this PR's injected-delay sweep**: it predicted ~+269% at 20ms injected; real 22ms WAN measured +318%. The injected-delay bench methodology holds up on the real internet.
- Loopback ceiling (same binary, same flags, both stages on the M5): 206–237 dec tok/s on repeat → WAN at 148 leaves the residual gap as per-window traversal cost at depth 2, as expected.

Raw per-rep JSON (full `timings` blocks + generated text) captured for all four arm×workload runs; happy to attach if useful.

## The two commits at the tip (`87e691a4`, `34bd3b89`)

WAN transport hardening ported from the retired lab branch, where they were validated live on the same Sydney↔AU path:

1. `87e691a4` — stop re-applying the formation-time `MAX_SPLIT_RTT_MS` ceiling to fresh *operational* stage streams (warn-and-proceed). Split admission still gates on gossiped, hysteresis-smoothed RTT + re-election. Without this, per-request direct-return sinks get rejected under normal WAN jitter → ready-handshake timeouts → 502s on an already-admitted split.
2. `34bd3b89` — return-sink ready-handshake timeout 5s→20s to cover cold bridge QUIC setup over WAN; matches the forward-lane budget.

Validation: fmt clean, `clippy -D warnings` clean (`mesh-llm-host-runtime`, `skippy-server`, `mesh-llm`), `cargo test -p mesh-llm-host-runtime --lib` 1698 passed, `-p skippy-server --lib` 281 passed. Today's WAN run is the live proof: with these two commits, 15 consecutive requests over real WAN with zero 502s after formation.

## Two real findings from the run (not yet fixed on this branch)

1. **Cold direct-return sink opens can still fail over WAN.** First request of one bringup hit `direct prediction return upstream-opened sink unavailable ... os error 35` + a lane-5 handshake failure → 502. The lab branch's prepared-return-sink pool (pre-warmed, age-rotated sinks; `2b76ba4e` there) addresses exactly this; we deliberately hadn't ported it. Today says it's worth porting.
2. **Withdraw is terminal — a withdrawn split never re-forms.** After that 502 the coordinator withdrew the topology (`periodic_check`, grace expired) and then sat in standby forever while the healthy worker sat "standing by for stage assignment" — dual restart with a fresh token was the only recovery. Root cause traced: `SplitCoordinatorEvent::Withdraw` → `StartupLoopControl::Break` → `startup_local_model_loop` tears down and **returns**; nothing re-enters the wait-for-participants → relaunch path. Proposed fix: on withdraw (when the model can't fit locally — the `LocalFallback` path already covers fits-locally), loop back to the participant-wait phase instead of ending the task. Working on this next; flagging here first in case there's an intended design for post-withdraw recovery we should follow instead.

michaelneale:
🤖 *Posted by micn's AI agent.*

Follow-up to finding 2 in the WAN report above: pushed `c05355ad` — **withdrawn splits now relaunch when peers return** instead of ending the model task.

Root cause: `SplitCoordinatorEvent::Withdraw` → `StartupLoopControl::Break` → `startup_local_model_loop` tore down and returned; nothing re-entered the participant-wait phase, so a withdrawn split needed a manual dual-restart even with a healthy worker standing by.

Fix shape (second-opinion reviewed): new `RelaunchSplit` outcome runs the full existing teardown, checks the stop channel (explicit shutdown still wins), then loops back to the launch phase and re-enters `wait_for_split_participants`. `LocalFallback` unchanged; the participant-wait loop's 30s cadence + stable-participant gating act as the retry throttle.

Validation: fmt clean, `clippy -D warnings` clean (`mesh-llm-host-runtime`, `mesh-llm`), `cargo test -p mesh-llm-host-runtime --lib` 1698 passed.

Still queued (not yet on branch): prepared-return-sink pool port for the cold-open 502 itself (finding 1).

crates/mesh-llm-ui/src/features/configuration/lib/build-toml.test.ts
ndizazzo: @i386 this comment seems to have been missed - @michaelneale opened #1056

PR #1027, opened 2026-07-20
Avoid stale embedded Metal shaders in llama builds

i386:
## Summary

- avoid using `sccache` for the C compiler launcher on Darwin Metal llama.cpp builds
- keep C++/CUDA/HIP compiler launchers cached where they are safe
- explicitly clear CMake compiler launcher settings when caching is unavailable or disabled

## Why

The embedded Metal shader library is included through an assembly `.incbin` object. `sccache` does not include that embedded payload in the assembly cache key, which can leave a llama.cpp build using stale Metal shader contents after shader source changes.

## Validation

- `bash -n scripts/build-llama.sh`
- `just build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compiler caching configuration for Apple Metal builds.
  * Prevented unsupported or unavailable caching settings from affecting C, C++, CUDA, and HIP compilation.
  * Updated build status messages to accurately reflect when caching is enabled only where safe.

### DIFF
--- a/scripts/build-llama.sh
+++ b/scripts/build-llama.sh
@@ -206,10 +206,14 @@ if [[ "$USE_SCCACHE" != "0" && -n "$SCCACHE_BIN" ]] &&
 fi
 
 if [[ "$USE_SCCACHE" != "0" && -n "$SCCACHE_BIN" ]]; then
-  CMAKE_ARGS+=(
-    -DCMAKE_C_COMPILER_LAUNCHER="$SCCACHE_BIN"
-    -DCMAKE_CXX_COMPILER_LAUNCHER="$SCCACHE_BIN"
-  )
+  if [[ "$(uname -s)" == "Darwin" && "$LLAMA_BACKEND" == "metal" ]]; then
+    # The embedded Metal source is pulled into an assembly object with .incbin.
+    # sccache does not include that payload in the assembly cache key.
+    CMAKE_ARGS+=(-DCMAKE_C_COMPILER_LAUNCHER=)
+  else
+    CMAKE_ARGS+=(-DCMAKE_C_COMPILER_LAUNCHER="$SCCACHE_BIN")
+  fi
+  CMAKE_ARGS+=(-DCMAKE_CXX_COMPILER_LAUNCHER="$SCCACHE_BIN")
   case "$LLAMA_BACKEND" in
     cuda)
       CMAKE_ARGS+=(-DCMAKE_CUDA_COMPILER_LAUNCHER="$SCCACHE_BIN")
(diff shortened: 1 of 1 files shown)


michaelneale:
Holding this one - I don't think it does what it claims, so let's not merge as-is.

The `.incbin` payload compiles as an **ASM**-language source (`ggml-metal-embed.s` via `enable_language(ASM)`), so clearing `CMAKE_C_COMPILER_LAUNCHER` has no effect on it - that's a per-language var for C sources. And in our build the ASM step isn't wrapped by sccache at all: ggml's internal `RULE_LAUNCH_COMPILE` fallback is guarded by `if(GGML_CCACHE AND NOT C_LAUNCHER AND NOT CXX_LAUNCHER)`, and we pass both launchers, so that block never runs. So there's no cache-key bug on that step to fix here.

The likely reason it appeared to fix stale shaders: changing `CMAKE_ARGS` changes the build stamp and forces a reconfigure/rebuild, so the stale object got relinked incidentally.

If we do want a real hardening, the minimal correct version keeps C+CXX cached and just disables ggml's fallback:
```
CMAKE_ARGS+=(
  -DGGML_CCACHE=OFF
  -DCMAKE_C_COMPILER_LAUNCHER="$SCCACHE_BIN"
  -DCMAKE_CXX_COMPILER_LAUNCHER="$SCCACHE_BIN"
)
```
(Note `CMAKE_ASM_COMPILER_LAUNCHER` is not a thing - CMake doesn't implement a launcher for ASM.)

Also minor: the 'sccache not found' branch clears the launchers but leaves `GGML_CCACHE` on, so on a box with ccache-but-no-sccache ggml would still enable global caching despite the 'without caching' message.

Separately - the failing Windows CPU check here is not caused by this PR; it's the pre-existing `autoupdate.rs` `INSTALL_SCRIPT_URL` cfg-gate breakage on main (fixed by #1030).

ndizazzo (approved):

ndizazzo:
> Holding this one - I don't think it does what it claims, so let's not merge as-is.
> 
> The `.incbin` payload compiles as an **ASM**-language source (`ggml-metal-embed.s` via `enable_language(ASM)`), so clearing `CMAKE_C_COMPILER_LAUNCHER` has no effect on it - that's a per-language var for C sources. And in our build the ASM step isn't wrapped by sccache at all: ggml's internal `RULE_LAUNCH_COMPILE` fallback is guarded by `if(GGML_CCACHE AND NOT C_LAUNCHER AND NOT CXX_LAUNCHER)`, and we pass both launchers, so that block never runs. So there's no cache-key bug on that step to fix here.
> 
> The likely reason it appeared to fix stale shaders: changing `CMAKE_ARGS` changes the build stamp and forces a reconfigure/rebuild, so the stale object got relinked incidentally.
> 
> If we do want a real hardening, the minimal correct version keeps C+CXX cached and just disables ggml's fallback:
> 
> ```
> CMAKE_ARGS+=(
>   -DGGML_CCACHE=OFF
>   -DCMAKE_C_COMPILER_LAUNCHER="$SCCACHE_BIN"
>   -DCMAKE_CXX_COMPILER_LAUNCHER="$SCCACHE_BIN"
> )
> ```
> 
> (Note `CMAKE_ASM_COMPILER_LAUNCHER` is not a thing - CMake doesn't implement a launcher for ASM.)
> 
> Also minor: the 'sccache not found' branch clears the launchers but leaves `GGML_CCACHE` on, so on a box with ccache-but-no-sccache ggml would still enable global caching despite the 'without caching' message.
> 
> Separately - the failing Windows CPU check here is not caused by this PR; it's the pre-existing `autoupdate.rs` `INSTALL_SCRIPT_URL` cfg-gate breakage on main (fixed by #1030).

Yeah, this checks out. Since the shader gets compiled as ASM, changing the C launcher won’t really affect that step. Let’s keep the C/C++ launcher flags, (could add -DGGML_CCACHE=OFF to disable the fallback), and dig into the stale shader issue separately. We could start with `ninja -d explain -v` to investigate the build obj.

PR #1028, opened 2026-07-21
Reliable WAN split serving + speculative pipelining over the internet

michaelneale:
## What this enables

Split serving now works reliably over a **WAN** link, and speculative "verify-window" pipelining can actually engage across it. Before this change, a 2-node split across the internet (e.g. a coordinator in one region and a GPU box in another) would **502 every request** — the direct-prediction-return channel that pipelining depends on could not be established, and even when it could, normal WAN latency jitter tore it down mid-request.

After this change:
- A cross-WAN split **serves inference** (previously impossible — it 502'd or hung).
- Speculative pipelining **engages over WAN** (`verify_window_max_in_flight > 1`), the mechanism that hides remote round-trip latency.
- When the direct channel genuinely isn't available, requests **fall back to serial serving instead of failing** — a request is never dropped.

Validated live on a Sydney↔AU 4090 split (~26ms RTT): direct-return confirmed working, warm requests dropped from ~12.7s to ~0.6s once the channel was reliable.

## Architecture

Four changes, smallest-first:

1. **Serve without direct-return, and stop killing operational streams on RTT jitter.** The native-MTP verify path used to hard-require a direct-return sink and 502 without it. It now falls back to serial completion over the forward lane (which polls both routes), so the split serves regardless. Separately, `open_stage_transport_stream` no longer re-applies the formation-time `MAX_SPLIT_RTT_MS` ceiling to every fresh operational stream — split *admission* already gates path eligibility with gossiped, hysteresis-smoothed RTT plus re-election. Re-checking instantaneous per-stream RTT was rejecting per-request return sinks under normal jitter while pooled forward lanes stayed healthy.

2. **Return-sink ready-handshake timeout 5s → 20s.** Over WAN the return sink connects to a local bridge alias, but the remote `ready` byte only arrives after the bridge cold-establishes a fresh stage QUIC connection (~10s budget) and the remote handler dials its local server. 5s timed out during that cold setup; forward lanes already used a 20s budget. Matching it lets the cold return path complete.

3. **Pre-warmed return-sink pool.** A single long-lived maintenance worker keeps `capacity = generation_concurrency` return sockets pre-warmed off the hot path (connect + ready handshake done, not yet bound), age-rotated before they can silently rot, and refilled FIFO on checkout. The hot path binds a pooled socket with a cheap write and falls back to a cold open, then to serial. This removes the cold-handshake intermittency that made direct-return work only sporadically.

4. **Implicit pipeline confirmation.** A successful `PredictionReturnOpen` write does not prove the tail selected the direct route (it may still reply on the forward lane), and the pipelined completion path is direct-only — so a false positive could hang up to 300s. Pipelining now gates on an *actual* direct reply observed **after** generation config, not on bind success. While unconfirmed, serial completion polls both routes (serving promptly) and trips confirmation on the first real direct reply. The flag is reset right after the generation-config ACK so pre-config restore/prefill replies can't falsely confirm.

## Protocol

No wire-format change. All behavior is additive and compatible with mixed-version meshes:
- The return-sink open/bind and confirmation are frontend-local (stage 0 ↔ tail over the existing binary transport); no new message kinds or fields.
- Older peers that reply on the forward lane simply keep the request on the serial path — correct, just not pipelined.
- The per-stream RTT gate change relaxes an operational check only; split *admission* eligibility is unchanged.

## Validation

- New unit tests: prepared-return-pool age/rotation policy, shared return-failure classifier; existing `direct_return`, `split_stage_path`, `split_readiness`, `stage_transport` suites pass.
- Full `skippy-server` lib suite green; `cargo clippy -D warnings` clean on `skippy-server` and `mesh-llm-host-runtime`; `cargo fmt --all --check` clean.
- Live: 2-node WAN split (Sydney M5 ↔ AU 4090, ~26ms) serves HTTP 200; `verify_window_direct_prediction_return=true`; pipelining reached `max_in_flight` up to 4 under forced mode.

## Notes / follow-up (not in this PR)

Depth tuning is intentionally out of scope. A separate A/B on this branch showed that pushing proposal depth from 2→4 on GLM-4.7-Flash over WAN *did* fill the pipeline and hide latency (downstream-wait ~700ms→~335ms) but **reduced** net throughput (~45→~37 tok/s) because divergence discarded far more speculative windows. The throughput lever for that model/workload is speculative **acceptance rate** and **keeping the pipeline full without wasting work on divergence**, not raw depth — tracked separately. This PR is the reliability foundation that makes any of that measurable.

## Summary by CodeRabbit

* **New Features**
  * Added a hidden CLI/config option to force speculative verify-window pipelining when depth is greater than one.
  * Added pre-warmed pooling for direct prediction-return connections to speed up embedded generation.
* **Bug Fixes**
  * Operational traffic no longer aborts on transient split-path rejection; it now warns and continues.
  * Verify-window pipelining is now gated on confirmed direct-return delivery; when unavailable, it falls back to serial behavior.
* **Monitoring**
  * Improved telemetry for direct-return open/bind phase classification, confirmation gating, verify-window force/fallback, and native vs ngram survival accounting.

### DIFF
--- a/SKIPPY_PROTOCOL_TODO.md
+++ b/SKIPPY_PROTOCOL_TODO.md
@@ -16,6 +16,14 @@ chunk and every normal decode token.
 > predicted-token / span returns exist (`binary_transport/direct_return.rs`).
 > Checkboxes that remain `[ ]` are still open or unverified.
 
+> **See also:** [docs/skippy/WAN_SPLIT_TRANSPORT_STATUS.md](docs/skippy/WAN_SPLIT_TRANSPORT_STATUS.md)
+> for the WAN split transport status/handoff (branch
+> `wip/wan-direct-prediction-return`, PR #1028): what is proven over WAN, the
+> landed transport keep-set, why fixed pipeline depth is diagnostic-only, the
+> go-forward suffix-ngram + adaptive-depth direction (PR #1037), and the 2-node
+> bringup config trap. Draft-model speculation is deferred on branch
+> `wip/wan-draft-ahead`.
+
 ## Current Observations
 
 - Prefill and decode use the same neighbor-chain shape:
--- a/crates/mesh-llm-cli/src/parser/commands.rs
+++ b/crates/mesh-llm-cli/src/parser/commands.rs
@@ -604,6 +604,11 @@ pub struct Cli {
     #[arg(long, hide = true)]
     pub speculative_verify_window_pipeline_depth: Option<u32>,
 
+    /// Force pipelined verify windows whenever depth > 1, bypassing the adaptive
+    /// profitability gate (shard-style "bet on the pipeline" mode for WAN splits).
+    #[arg(long, hide = true)]
+    pub speculative_verify_window_pipeline_force: Option<bool>,
+
     /// Draft model for speculative decoding.
     #[arg(long, hide = true)]
     pub draft: Option<PathBuf>,
--- a/crates/mesh-llm-config/src/model.rs
+++ b/crates/mesh-llm-config/src/model.rs
@@ -570,6 +570,7 @@ pub struct SpeculativeConfig {
     pub verify_window_min_tokens: Option<u32>,
     pub verify_window_max_tokens: Option<u32>,
     pub verify_window_pipeline_depth: Option<u32>,
+    pub verify_window_pipeline_force: Option<bool>,
     pub spec_default: Option<BoolOrAuto>,
     pub(crate) legacy_draft_model_path_used: bool,
 }
@@ -623,6 +624,7 @@ impl SpeculativeConfig {
             verify_window_min_tokens: pick!(verify_window_min_tokens),
             verify_window_max_tokens: pick!(verify_window_max_tokens),
             verify_window_pipeline_depth: pick!(verify_window_pipeline_depth),
+            verify_window_pipeline_force: pick!(verify_window_pipeline_force),
             spec_default: pick!(spec_default),
             legacy_draft_model_path_used: overrides
                 .filter(|config| config.draft_model.is_some())
@@ -700,6 +702,8 @@ struct SpeculativeConfigRaw {
     #[serde(default)]
     verify_window_pipeline_depth: Option<u32>,
     #[serde(default)]
+    verify_window_pipeline_force: Option<bool>,
+    #[serde(default)]
     spec_default: Option<BoolOrAuto>,
 }
 
@@ -746,6 +750,7 @@ impl<'de> Deserialize<'de> for SpeculativeConfig {
             verify_window_min_tokens: raw.verify_window_min_tokens,
             verify_window_max_tokens: raw.verify_window_max_tokens,
             verify_window_pipeline_depth: raw.verify_window_pipeline_depth,
+            verify_window_pipeline_force: raw.verify_window_pipeline_force,
             spec_default: raw.spec_default,
             legacy_draft_model_path_used: legacy_used,
         })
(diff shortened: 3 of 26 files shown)


ndizazzo (approved):
Looks fine to me! 

@michaelneale Just FYI, this `verify_window_pipeline_force = true|false` option will _automatically_ appear in the configuration TOML and React options on the frontend, if you are okay with that. This is because of the recent work to enumerate settings through the schema (which was updated in the PR)... So if you don't want it exposed you'll have to hide that setting.

michaelneale:
🤖 *Posted by micn's AI agent after a triage of this branch, `leyten/shard`, and #1037.*

## Where this branch actually stands

The **transport layer is proven** (WAN split 502 → serving, warm 12.7s → 0.6s, direct-return confirmed, depth 4 in flight). The **payoff layer is not**: pipelining only engaged under `--force`, the one real A/B showed deeper speculation *lost* throughput (45→37 tok/s, stale windows 2→16), and the draft-model act (`DraftAheadWorker`, draft seeding) is uncommitted WIP with a known `Drop` deadlock risk. shard's own numbers point the same way: once the WAN was hidden, their draft model was 94% of the loop — the draft is the cost center, not the magic. Their `glm_swarm_nvfp4_ngram_pipe.py` already replaced it with a free suffix lookup.

## Plan

```text
1. SPLIT this PR. Land the transport core on its own — it is proven and
   proposer-agnostic:
     - serial fallback when direct-return unavailable   (c340f741)
     - per-stream RTT gate fix                          (c340f741)
     - 20s return-sink ready-handshake timeout          (46108cfc)
     - pre-warmed return-sink pool + implicit
       pipeline confirmation                            (2b76ba4e)
     - per-source accepted-prefix survival telemetry    (86ff55eb)

2. DROP the draft-model plumbing from this PR:
     - draft_compat.rs (bdf505c7) — no call sites; park with the draft work
     - draft-model survival gate (cfff45db)
     - uncommitted DraftAheadWorker / draft-seeding working-tree WIP

3. SWITCH pipeline fill to #1037's suffix ngram proposer.
   84.9% acceptance at depth = pipeline stays full on re-emissive text
   (agent loops, file edits) — the dominant mesh workload. The Phase 1
   continuous refill (44d31e3b) only pays with this proposer; keep it,
   but expect nothing from it until #1037 lands.

4. MAKE DEPTH ADAPTIVE, not fixed. Deep while suffix matches hold; drop
   to serial on miss (shard's ngram-pipe stalls to one sync traversal).
   Fixed depth 4 is what lost throughput here. Gate depth on the Phase 0
   survival curve, per source.

5. FIX #1037's two blockers before any WAN bench on it:
     - prefix-cache + speculative checkpoint 502 (chain_restore_hit)
     - greedy-equivalence (verified spec decode must be byte-identical;
       shard proves this per run — we should too)

6. DEFER the draft model to an optional later phase for novel-prose
   workloads (ngram survival there is ~0). If built, it must be async
   draft-ahead hidden inside the WAN RTT — not the synchronous seeding
   abandoned here — and MTP+suffix composite likely covers the gap
   anyway, with no external draft to source.
```

## What to keep from this branch

**Keep (proven, land as the split-out PR):**
- `c340f741` — serial fallback + RTT operational-stream gate fix. Root-cause fix; turns hard 502s into served requests. *Note: warn-and-proceed changes failure semantics for LAN splits too — degradation now resolves via re-election, not fast stream failure.*
- `46108cfc` — 20s ready timeout (matches forward-lane budget).
- `2b76ba4e` — return-sink pool + confirmation gating on a *real* direct reply (closes a 300s-hang hazard). Best-tested code on the branch. Known limit: pool covers stage-0→immediate-downstream only, so 3+-stage chains still cold-open.
- `86ff55eb` — survival telemetry. This is the instrument for judging *any* proposer, including #1037.

**Keep conditionally:**
- `44d31e3b` — continuous refill. Sound mechanism (sealed-tail invariant), worthless with the 4-token ngram window, pays with #1037's suffix proposer.
- `e48d4583` / `verify_window_pipeline_force` — bench/CI lever only; never a production default.

**Cut / park:**
- `bdf505c7` — `draft_compat.rs` (unused: no call sites beyond its re-export).
- `cfff45db` — draft-model survival gate.
- All uncommitted draft-ahead working-tree WIP.

Protocol: no wire change anywhere in the keep set; older peers just stay on the serial path. Mixed-version safe.

PR #1031, opened 2026-07-21
Add direct quant layer package tooling

i386:
## Summary

- add direct BF16/FP16 GGUF to quantized layer-package tooling in `skippy-quantize`
- support explicit tensor-type recipes and metadata overrides for custom quant profiles
- add GLM-5.2 quant recipes plus helper scripts for local/HF quant layer-package workflows

## Why

Large-model experiments need to reuse BF16 GGUF artifacts and produce quantized Skippy layer packages without round-tripping through ad hoc intermediate scripts. This makes the quant path reproducible and keeps custom tensor decisions explicit in checked-in recipes.

## Validation

- `bash -n scripts/skippy-bf16-to-quant-layer-package.sh scripts/glm52-q2-routed-down-layer-package.sh scripts/glm52-q2-routed-down-hf-job.sh`
- `cargo check -p skippy-quantize`
- `cargo test -p skippy-quantize`

## Summary by CodeRabbit

- **New Features**
  - Added GLM-DSA model conversion support, including indexer metadata and KV projection splitting.
  - Added tensor-name mapping and quantization recipes for GLM 5.2, including routed-down variants.
  - Added scripts for local routed-down model packaging and BF16-to-layer-package conversion.
  - Existing manifests can now omit tensor recipe settings.

- **Documentation**
  - Documented recipe syntax, regex matching, precedence, validation, and GLM-5.2 usage.

- **Bug Fixes**
  - Improved validation and error reporting for incomplete or conflicting GLM-DSA configuration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7488,6 +7488,7 @@ dependencies = [
  "clap",
  "libc",
  "llama-quant-ffi",
+ "regex-lite",
  "serde",
  "serde_json",
  "skippy-ffi",
--- a/crates/skippy-quantize/Cargo.toml
+++ b/crates/skippy-quantize/Cargo.toml
@@ -21,3 +21,6 @@ llama-quant-ffi = { path = "../llama-quant-ffi" }
 serde.workspace = true
 serde_json.workspace = true
 skippy-ffi = { path = "../skippy-ffi" }
+
+[dev-dependencies]
+regex-lite = "0.1"
--- a/crates/skippy-quantize/README.md
+++ b/crates/skippy-quantize/README.md
@@ -414,10 +414,24 @@ per-tensor policy with `--tensor-type-file` or repeated `--tensor-type`.
 The tensor recipe format is one override per line:
 
 ```text
-blk.*.ffn_gate_exps.weight=Q2_K
-blk.*.ffn_down_exps.weight=Q3_K
-mtp.*=Q8_0
+attn_q_a\.weight$=Q8_0
+ffn_down_shexp\.weight$=Q4_K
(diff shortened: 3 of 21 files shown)


michaelneale (approved):
Approving. Self-contained quant tooling in the `skippy-quantize` leaf crate plus checked-in recipes and helper scripts — no mesh/wire/ABI protocol surface, CI fully green. Reproducible BF16/FP16 → quantized layer-package path with explicit tensor-type recipes is the right direction over ad-hoc scripts. No collision with in-flight work (only the shared `autoupdate.rs` hunk, which is identical across branches and auto-resolves).

crates/mesh-llm-system/src/autoupdate.rs
```diff
@@ -10,8 +10,10 @@ mod release_fetch;
 pub use release_fetch::{latest_release_version, version_newer};
 
 use platform_probe::{installed_bundle_flavor, preferred_bundle_flavor_for_current_host};
+#[cfg(not(windows))]
+use release_fetch::INSTALL_SCRIPT_URL;
 use release_fetch::{
-    INSTALL_SCRIPT_URL, InstallOutcome, PostInstallAction, RELEASES_URL, ReleaseAssetPreference,
+    InstallOutcome, PostInstallAction, RELEASES_URL, ReleaseAssetPreference,
```
ndizazzo, line 16: PR needs a rebase, this was merged

ndizazzo (commented):
Approving! But needs a rebase and Coderabbit had a good callout that we can fix up before merging to keep stuff < 2000

PR #1033, opened 2026-07-21
Add GLM DSA correctness trace reporting

i386:
## Summary
- add GLM-DSA stage-0 trace capture and semantic parity reporting to skippy-correctness
- report downstream activation/top-k payload digests and GLM-DSA trace point summaries
- document the trace command for validating direct vs shared IndexShare behavior

## Why
GLM-DSA needs repeatable correctness evidence before native sparse/decode work is treated as production-ready. This adds a focused correctness reporting path independent of the larger benchmark and llama patch queue work.

## Validation
- cargo test -p skippy-correctness
- cargo check -p skippy-correctness
- cargo clippy -p skippy-correctness --all-targets -- -D warnings

## Summary by CodeRabbit

- **New Features**
  - Added the `glm-dsa-stage0-trace` command to run a diagnostic loop comparing fused vs direct GLM-DSA stage-0 behavior.
  - Exposed CLI options for tracing, timing breakdowns, prompt input, tolerance controls, and runtime/server settings.
  - Outputs a detailed pretty-printed JSON report including tensor trace parity, downstream message/top-k metrics, activation error stats, and pass/fail results.

- **Documentation**
  - Expanded README usage with the full CLI invocation and notes on interpreting fused-vs-direct diagnostic outputs.

- **Tests**
  - Added automated coverage for trace parsing, parity gating, top-k visibility counting, and timing summary extraction.

### DIFF
--- a/crates/skippy-correctness/README.md
+++ b/crates/skippy-correctness/README.md
@@ -113,6 +113,17 @@ skippy-correctness native-mtp-open-ai-ab \
   --stage1-endpoint-addr 192.168.0.10:19172 \
   --layer-end 48 \
   --split-layer 24
+
+skippy-correctness glm-dsa-stage0-trace \
+  --model /Volumes/External/models/huggingface/.../GLM-5.2-Q2_K-MTP-Q8-layers/snapshots/main \
+  --stage-model /Volumes/External/models/huggingface/.../GLM-5.2-Q2_K-MTP-Q8-layers/snapshots/main \
+  --model-id meshllm/GLM-5.2-Q2_K-MTP-Q8-layers \
+  --stage-load-mode layer-package \
+  --ctx-size 1024 \
+  --n-gpu-layers 999 \
+  --stage-server-bin target/release/skippy-server \
+  --prompt-bin target/release/skippy-prompt \
+  --report-out reports/glm-dsa-stage0-trace.json
 ```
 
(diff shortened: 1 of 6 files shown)


michaelneale:
One thing to fix before this is used as a correctness gate: the pass/fail decision uses OR -

```rust
let parity_matched = trace_parity.matched || semantic_parity.matched;
```

For a gate that certifies fused == direct, OR is the weakest possible check - it passes if *either* signal passes, so a real divergence in one dimension is masked whenever the other happens to match. Two false-green cases:
- trace points match but downstream activations/top-k differ beyond tolerance -> reports pass
- downstream within tolerance but a traced intermediate tensor diverged (shape/type/stats) -> reports pass

It also interacts badly with `--allow-trace-mismatch`: that forces `trace_parity.matched = true`, so the OR then bypasses semantic parity entirely whenever both variants complete.

Suggest AND, with the trace waiver only waiving trace:
```rust
let passed = both_variants_completed && semantic_parity.matched && trace_parity.matched;
// --allow-trace-mismatch waives trace but semantic parity must still hold
```

If empty/absent traces are a legitimate state, better to model trace as Pass | Fail | Unavailable and only let Unavailable fall back - a present *failure* should never be treated like missing evidence. Everything else here looks good and self-contained.

ndizazzo (approved):
Seems fine, pending CR items

PR #1034, opened 2026-07-21
Add GLM DSA llama foundation patches

i386:
## Summary
- add the GLM-DSA llama.cpp foundation patch queue: sparse execution primitives, GLM-DSA staged controls, and native execution tests
- add GLM-DSA fused KV_B tensor loading plus IndexShare graph wiring
- route GLM-DSA through the DSA KV cache path

## Why
This is the first clean split from the GLM-5.2 research branch. It keeps the native GLM-DSA llama.cpp foundation separate from quant tooling, package validation, correctness reporting, MoE micro-optimizations, and speculative/MTP experiments.

## Validation
- LLAMA_WORKDIR=/tmp/mesh-llm-llama.t1iPon scripts/prepare-llama.sh pinned
- just build

## Summary by CodeRabbit

- **New Features**
  - Added native GLM-DSA execution with staged sparse-attention controls, top-k activation handling, and IndexShare graph support.
  - Added support for fused KV tensors and a dedicated DSA KV cache for GLM-DSA models.
  - Added configurable mmap and GLM-DSA policy settings.

- **Performance**
  - Improved sparse attention, MoE decoding, compact flash, and graph execution paths.

- **Bug Fixes**
  - Fixed Metal `get_rows` kernel placement.
  - Updated runtime compatibility requirements for the expanded ABI.

PR #1035, opened 2026-07-21
Add GLM MoE Metal active expert patches

i386:
## Summary
- add a stacked GLM MoE Metal patch queue on top of the GLM-DSA llama foundation branch
- add active-expert Q3 down kernels and Q2 gate/up dispatch controls for GLM-5.2 style expert tensors
- add selector diagnostics and focused backend/perf tests for selected expert paths

## Why
This keeps the GLM MoE performance work separate from the core GLM-DSA architecture and IndexShare support. The branch is stacked so the foundation can be reviewed first, then the active-expert Metal path can be reviewed as the next layer.

## Validation
- LLAMA_WORKDIR=/tmp/mesh-llm-llama-moe.bjcGOo scripts/prepare-llama.sh pinned
- just build


michaelneale (approved):
Approving. Clean stacked patch layer on top of the GLM-DSA foundation — all 12 files are llama.cpp patch files with no Rust/protocol/ABI surface of its own, so zero collision risk with in-flight work. Good stacking discipline keeping the MoE Metal perf work separate from the core foundation.

Note this is gated on its base (#1034, `jd/glm-dsa-llama-foundation`) merging first — it can't land until then.

PR #1037, opened 2026-07-21
Add Ngram Suffix Proposer

danielwinterw:
### Summary
Adds a third N-gram draft proposer, `suffix`: a pure-Rust longest-suffix matcher (prompt-lookup decoding). `simple` and `cache` are bound by llama.cpp's 4-token match window. `suffix` is not, so it matches verbatim spans up to 64 tokens and copies them as the speculative draft. Additive and default-off.

### Motivation
The existing `simple`/`cache` proposers wrap llama.cpp's N-gram proposer, capped at a 4-token match window (`NGRAM_CACHE_MAX_NGRAM`). In long agent transcripts a 4-token match is ambiguous: it occurs in many places with different continuations, so drafts stay short and get rejected.

Agent-coding workloads are dominated by re-emission: read a file, write it back with a small change; echo tool output; repeat identifiers. The output is largely a copy of text already in context. A 16 to 32 token suffix match is usually unique, so it can copy whole spans the model would otherwise decode one token at a time.

### What's in the PR
- `SuffixNgramProposer` (`speculative/suffix.rs`): indexes committed history by an exact seed key (no hash collisions), finds the longest verbatim earlier occurrence of the query suffix (up to `SUFFIX_NGRAM_MAX_WINDOW = 64`), and copies what followed. Draft length scales with match length. Stays silent below `ngram_min`.
- Standalone support (`speculative/standalone.rs`): Cache and Suffix now work as standalone (non-MTP) proposers, not just Simple. A plain `ngram_proposer = "suffix"` config activates it. No MTP model required.
- Telemetry: per-request proposer stats (attempts, hits, max/sum match length, proposed/accepted tokens, sync/lookup time) in response timings. Standalone timings report generic proposed/accepted/rejected/acceptance totals so arms are comparable.
- Two observability fixes made while benchmarking: `prefix_cache.enabled = false` now survives model-family defaults, and standalone timing fields are populated. Neither changes the lookup algorithm.
- Config: `ngram_proposer = "suffix"` (mesh config and CLI). Reuses `NgramProposalConfig` (`min_ngram`/`max_ngram`/`max_proposal_tokens`).
- Docs: `docs/skippy/SUFFIX_NGRAM_PROPOSER.md`, USAGE.md, CONFIGURATION.md.

### Benchmarks
Standalone N-gram matrix on `GLM-4.7-Flash-MTP-GGUF:Q4_K_M`, two-stage split, deterministic re-emit-a-file prompt (rename one function, generate 384 tokens). Medians of 5 measured requests after 2 warmups. Prefix caching disabled so a cross-request restore cannot skew the comparison.

| Arm | Server decode tok/s | End-to-end tok/s | Acceptance | Proposed / accepted |
|---|---:|---:|---:|---:|
| Target only | 22.6 | 19.9 | n/a | 0 / 0 |
| Cache | 78.0 | 52.9 | 67.1% | 523 / 351 |
| Simple | 107.2 | 64.4 | 53.2% | 694 / 369 |
| **Suffix** | **119.2** | **68.5** | **84.9%** | **425 / 361** |

Suffix was the fastest arm: +427% over target-only, +11% over Simple. It had the highest acceptance while proposing the fewest tokens. Lookup cost was about 12.6 microseconds per request.

Simple beat Cache on throughput despite lower acceptance, because it proposed more tokens in absolute terms. Acceptance percentage alone does not rank a proposer.

Reproduced on a local 8B two-stage split (base M1 Pro): about 3.6x over baseline on a realistic file re-emit, neutral on freeform chat.

### Not in scope / follow-ups
- Quality equivalence is not proven. This is a mechanism microbenchmark, not a quality result. Outputs matched on the copied body but diverged at the one edited line before reconverging. Exact greedy-trajectory equivalence is unresolved, so these numbers are not proof of identical output.
- Standalone only. Not the MTP plus N-gram composite path.
- Prefix-cache interaction: with caching enabled, a second standalone Simple request returned 502 after a `chain_restore_hit`. Disabling the cache made sequential requests reliable. Speculative checkpoint state appears to interact with cross-request KV restore. This should be fixed before a cache-enabled rerun.

### Testing
- Unit tests for the proposer (longest-match-wins, long-run edit workload, silence below min, provisional-prefix read-only, incremental sync vs rebuild) and standalone dispatch (simple/cache/suffix), plus config round-trip and validation.
- Full workspace build and clippy clean.

### Config example
​```toml
[models.speculative]
strategy = "auto"
ngram_proposer = "suffix"
ngram_min = 5
ngram_max = 32
ngram_max_proposal_tokens = 48
​```

## Summary by CodeRabbit

* **New Features**
  * Added `ngram-suffix` speculative decoding for standalone and native-MTP flows (as `ngram-suffix` / `native-mtp+ngram-suffix`).
  * Added a split-deploy advanced `--split-topology-lock` option.
  * Updated standalone proposal handling to use a single proposal-limit value (instead of min/max) for embedded serving.

* **Bug Fixes**
  * Tightened validation for suffix and restricted strategy/proposer combinations.
  * Improved speculative decode/verify safety and KV-cache disable behavior.

* **Documentation**
  * Added and expanded suffix proposer docs and configuration guidance.
  * Updated CLI help and config schema references.

* **Telemetry**
  * Added suffix proposer and history ngram proposer performance metrics.


i386:
Integrated PR #1026 (`68f45027b`) with the suffix proposer work and reran the
full eight-arm matrix on the lab. Integration/test head: `b83c4ee0e` (runtime
binary was `39bd799db`; the final commit is docs-only).

### Scenario

- `meshllm/GLM-4.7-Flash-MTP-GGUF:Q4_K_M`
- Two-stage 47/1 layer split: M3 Ultra stage 0, M1 Ultra stage 1
- 8 lanes, ctx 32,768, prefix cache disabled
- Same deterministic repeated-code edit prompt, temperature 0, seed 42,
  384 generated tokens
- Sequential release-mode arms, 2 warmups + 5 measured requests each
- Direct lab LAN, no injected inter-stage delay
- Every telemetry run contained 7 requests; collector logged no export errors

### Results versus the previous runs

Values are median server decode TPS / end-to-end wall TPS.

| Arm | Previous | With #1026 | Change |
|---|---:|---:|---:|
| Target only | 22.93 / 19.68 | 22.96 / 20.20 | +0.1% / +2.6% |
| Standalone Simple | 107.22 / 64.38 | 27.45 / 23.54 | -74.4% / -63.4% |
| Standalone Cache | 77.96 / 52.94 | 27.17 / 23.32 | -65.1% / -56.0% |
| Standalone Suffix | 119.24 / 68.49 | 27.33 / 23.49 | -77.1% / -65.7% |
| MTP only | 30.53 / 25.84 | 30.37 / 25.67 | -0.5% / -0.7% |
| MTP + Simple | 56.56 / 42.16 | 30.42 / 25.71 | -46.2% / -39.0% |
| MTP + Cache | 64.62 / 46.40 | **118.71 / 68.92** | **+83.7% / +48.6%** |
| MTP + Suffix | 78.05 / 53.24 | **138.94 / 75.98** | **+78.0% / +42.7%** |

Against the current MTP-only row, MTP+Suffix is **4.57x server TPS** and
**2.96x wall TPS**. MTP+Cache is 3.91x and 2.68x respectively.

### @michaelneale: are the stages actually busy in parallel?

For Cache and Suffix, yes; for MTP-only and Simple, no.

| Arm | N-gram acceptance | Max in flight | Avg in flight | Time at depth 2 | Refill successes / attempts | Stale windows |
|---|---:|---:|---:|---:|---:|---:|
| MTP only | — | 1 | 0.49 | 0.0% | 0 / 0 | 0 |
| MTP + Simple | no N-gram tokens | 1 | 0.50 | 0.0% | 0 / 0 | 0 |
| MTP + Cache | 59.3% | 2 | 1.12 | 20.5% | 25 / 180 | 15 |
| MTP + Suffix | 72.2% | 2 | **1.29** | **51.2%** | **50 / 55** | 10 |

Suffix contributes 2,135 refill tokens across the five measured requests,
versus Cache's 355. That longer horizon is what keeps the positional pipeline
occupied. Suffix lookup cost is only **34 microseconds total** across those five
requests, so hashing/index lookup is not the bottleneck.

The standalone regression is also revealing: Cache and Suffix accept 96.4%
and 98.9% of standalone proposals, yet both remain around 27 server TPS. High
acceptance without overlapping stage work is not useful. The standalone path
still needs to be moved onto the positional pipeline. MTP+Simple is configured
as hybrid but records zero refill attempts/tokens, so it is not an active
extension under #1026 and should not be presented as one.

This rerun proves the pipeline mechanism, but not the controlled-WAN claim:
injected delay was 0 ms. The next experiment must repeat the same matrix at
20/100 ms and report occupancy, downstream wait, and stale work. The prior run
showed Suffix beating Cache despite slightly lower tail acceptance; this rerun
does **not** reproduce that lower-acceptance ordering—Suffix now has both higher
acceptance and higher occupancy. What it does show clearly is that occupancy,
not acceptance alone, predicts the large TPS gain.

Correctness caveat: every arm was deterministic within-arm. Target-only, all
standalone arms, MTP-only, MTP+Simple, and MTP+Cache share one output hash;
MTP+Suffix differs at the requested edit. Both 384-token generated fragments
are malformed, so these are throughput/scheduling results, not a quality or
exact-greedy-equivalence result.

ndizazzo:
We might want to consider targeting this PR to #1026 for the time being since these were integrated... Once #1026 is merged GitHub will retarget `main` as the base.

micspiral:
Minor: a couple of stale doc comments still reference `simple` after this re-port dropped `NgramProposerKind::Simple`:

- `speculative/standalone.rs:21` — "Runs the configured standalone N-gram proposer (simple, cache, or suffix)"
- `speculative.rs:459` — `from_config` says "or `None` for simple/no proposer"

`simple` no longer exists as a kind, so these should just say cache/suffix. Trivial, non-blocking.

i386 (changes requested):
Requesting changes for the following issues:

1. **P1 — Standalone N-gram is silently ignored for single-stage/direct serving.** `speculative_mode_for_embedded` enables N-gram regardless of `staged`, but a stage without `downstream` immediately delegates to `generate_local_tokens`, which has no N-gram path. Direct GGUF and one-stage package requests therefore run target-only despite reporting `ngram-cache`/`ngram-suffix`. Please implement local verification or reject this configuration outside multi-stage serving.

2. **P1 — `strategy = "disabled"` can be overridden by inherited N-gram settings.** `resolve_decode_config` builds an N-gram plan from defaults without checking the requested strategy, after which the mode override re-enables speculation. A model-level disable must clear or ignore inherited proposer state. This corroborates the existing unresolved inline thread.

3. **P2 — Suffix lookup remains unbounded on ambiguous repetitive input.** Every indexed occurrence is examined unless one reaches the full `max_window`. Repeated seeds whose preceding tokens differ scan the entire bucket on every decode step, yielding `O(generated_tokens × candidate_occurrences × max_window)` work for request-controlled prompts. Please add a hard candidate budget or another bounded index strategy.

4. **P2 — Existing speculative-plan JSON is no longer readable.** `NgramProposalConfig.kind` is newly required, while plans produced by previous releases contain only `min_ngram`, `max_ngram`, and `max_proposal_tokens`. `--openai-speculative-config` will fail deserialization. A serde default of `Cache` would preserve prior behavior.

5. **P2 — Explicit standalone strategies can silently become disabled.** With `strategy = "ngram-cache"` or `"ngram-suffix"`, no package metadata, and either bound omitted, N-gram construction is skipped and resolution succeeds with no proposer. An explicitly requested strategy should fail with the documented “both bounds required” error.

6. **P2 — Two copy-paste configuration examples contain rejected keys.** `extension_initial_tokens` and `extension_tail_backoff_proposals` appear in `docs/skippy/SUFFIX_NGRAM_PROPOSER.md` and `docs/USAGE.md`, but do not exist in `SpeculativeConfigRaw`, which denies unknown fields.

Reviewed head `165db9320e170a1334c376f6857cb8daa85a324f`.

PR #1040, opened 2026-07-21
ui(data-mode): migrate legacy harness storage to v2

### LINKED ISSUE #633: ui(data-mode): bump storage key version to migrate pre-fix users off 'harness'
## What

The data-mode default was flipped from `'harness'` to `'live'` in commit `4acfd4cd` (PR around fly redeploy time). But the storage key `mesh-llm:data-mode:v1` is **never bumped**, so anyone who loaded the console before that fix has `'harness'` cached in localStorage and the new binary respects that "explicit choice" forever.

```ts
// crates/mesh-llm-ui/src/lib/data-mode/DataModeContext.tsx
function readStoredDataMode(storageKey: string, fallbackMode: DataMode, persist: boolean): DataMode {
  const storedValue = window.localStorage.getItem(storageKey)
  return isDataMode(storedValue) ? storedValue : fallbackMode
}
```

The fix's commit message correctly notes the toggle persists per-origin, but doesn't address the fact that `'harness'` got there as a *default*, not a *user choice*.

## How it surfaces

Visited `http://localhost:3131/` tonight on a mesh-llm built **before** `4acfd4cd` \u2014 console defaulted to harness mode and showed fixture providers (Vast.ai, RunPod, etc.) on the Reserves page. Even after I rebuilt to the post-fix binary, the localStorage cache would have kept it in harness mode if I hadn't been on a different port. Anyone running mesh-llm on the same origin pre/post-fix sees this.

## Fix

Bump the storage key version:

```ts
// crates/mesh-llm-ui/src/lib/data-mode/DataModeContext.tsx
- export const DATA_MODE_STORAGE_KEY = `${env.storageNamespace}:data-mode:v1`
+ export const DATA_MODE_STORAGE_KEY = `${env.storageNamespace}:data-mode:v2`
```

That makes the v1 stored value invisible to the new code, so the new code re-evaluates against the current default (`'live'` in production, `'harness'` in dev). Users who explicitly want harness after the bump still get it persisted under the new key.

## Why this is a separate issue from the original fix

The original fix changed the *new-visitor default*. This is about *migrating existing-visitor state*. Both are needed for the fix to actually work for everyone.

## Tests

Update the existing default-mode test to also assert that a pre-existing `v1` localStorage value is ignored when the provider reads `v2`.

## Risk

Minimal. Worst case: a user who explicitly chose `'harness'` for legit reasons (designer iterating on mockups) sees `'live'` once and toggles back. The toggle is one click and is then persisted under the new key.

## Pattern for future

When changing any persisted-state default, bump the version suffix. Same pattern applies to other localStorage / IndexedDB keys throughout the UI.

ndizazzo:
## Summary

- bump the app-owned data-mode storage key from `v1` to `v2` so legacy `harness` defaults are not mistaken for explicit user choices
- seed `v2` from the current environment default only during a valid app-key upgrade, retain `v1` for downgrade compatibility, and avoid repeated writes
- repair malformed current state without reviving stale legacy data, preserve in-memory startup when storage writes fail, and leave host-owned keys untouched
- update E2E storage setup to target `v2` and add migration/error/idempotency regression coverage

Closes #633

## Validation

- focused `DataModeProvider` tests: 11 passed
- `pnpm run lint`
- `pnpm run typecheck`
- full UI unit suite: 958 passed, 3 skipped
- UI production build
- `just website-build`
- `just test-all` (all 10 phases, including 1,700 host-runtime tests and Playwright smoke tests)

## Notes

The first full gate encountered one load-sensitive timeout in an unrelated mesh-visualization test; that test passed immediately in isolation. A subsequent full run passed the complete UI suite. The next run exposed missing website dependencies in the fresh worktree; after installing the lockfile dependencies, the website build and a final complete `just test-all` both passed.

## Summary by CodeRabbit

* **New Features**
  * Added versioned “data mode” persistence with automatic upgrade from prior stored settings, ensuring a smooth transition to the current default.

* **Bug Fixes**
  * Improved handling of malformed stored values and storage-write failures, including preserving legacy selections.
  * Avoids redundant localStorage updates when the selected mode hasn’t changed.

* **Tests**
  * Updated e2e smoke and parity tests to use the shared data-mode storage key.
  * Expanded unit test coverage for migration, fallback behavior, and error scenarios.

* **Chores**
  * Updated TypeScript path aliasing to support e2e imports.

### DIFF
--- a/crates/mesh-llm-ui/e2e/configuration/schema-controls.spec.ts
+++ b/crates/mesh-llm-ui/e2e/configuration/schema-controls.spec.ts
@@ -1,7 +1,7 @@
 import AxeBuilder from '@axe-core/playwright'
 import { expect, test, type Page, type TestInfo } from '@playwright/test'
+import { DATA_MODE_STORAGE_KEY } from '../support/data-mode'
 
-const DATA_MODE_STORAGE_KEY = 'mesh-llm-ui-preview:data-mode:v2'
 const FEATURE_FLAGS_STORAGE_KEY = 'mesh-llm-ui-preview:feature-flags:v1'
 
 type JsonRecord = Record<string, unknown>
--- a/crates/mesh-llm-ui/e2e/configuration/schema-controls.spec.ts
+++ b/crates/mesh-llm-ui/e2e/configuration/schema-controls.spec.ts
@@ -1,6 +1,6 @@
 import AxeBuilder from '@axe-core/playwright'
 import { expect, test, type Page, type TestInfo } from '@playwright/test'
-import { DATA_MODE_STORAGE_KEY } from '../support/data-mode'
+import { DATA_MODE_STORAGE_KEY } from '@e2e/support/data-mode'
 
 const FEATURE_FLAGS_STORAGE_KEY = 'mesh-llm-ui-preview:feature-flags:v1'
 
--- a/crates/mesh-llm-ui/e2e/configuration/schema-controls.spec.ts
+++ b/crates/mesh-llm-ui/e2e/configuration/schema-controls.spec.ts
@@ -1,7 +1,7 @@
 import AxeBuilder from '@axe-core/playwright'
 import { expect, test, type Page, type TestInfo } from '@playwright/test'
 
-const DATA_MODE_STORAGE_KEY = 'mesh-llm-ui-preview:data-mode:v1'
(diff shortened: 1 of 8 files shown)


ndizazzo:
@coderabbit review

PR #1041, opened 2026-07-21
fix: record activation cache prefix identities

### LINKED ISSUE #683: Apply record_identities to remaining single-identity record sites (chunked prefill, binary transport, activation cache)
## Summary

After #661 fixed the `LocalRuntime` record path and #681 fixes the `EmbeddedStageZero` full-prefill record path, three remaining record sites still call `prefill_identity` (single exact identity) instead of `record_identities` (exact + grid-floor candidates). These have the same shared-prefix miss pattern as the original #662 repro, just in narrower code paths.

## Sites

1. **`crates/skippy-server/src/frontend/prefix_cache.rs:140`** — `record_embedded_stage0_prefill` (the **chunked** prefill record, sibling of the full-prefill function fixed in #681). If a long shared prefix gets split across chunks, each chunk records exact-only and misses the grid-floor candidate.

2. **`crates/skippy-server/src/binary_transport.rs:2630`** — `record_exact_state` full-prefill record in the binary transport stage server. Used when a stage server records a full prefill via the binary wire protocol.

3. **`crates/skippy-server/src/kv_integration/activation.rs:52`** — activation cache record. Different cache (activation cache, not resident-KV), but has the same single-identity recording pattern. Needs a closer look to confirm whether grid-floor candidates make sense in the activation page-id derivation.

## Why this matters

Each site is in a narrower code path than `LocalRuntime` or `EmbeddedStageZero` full-prefill, so the user-visible impact is smaller than #662. But agent harnesses with long shared system prompts hitting the chunked-prefill path or the binary transport will see reduced hit rate.

## Fix shape

Same as #681 — replace the single `prefill_identity` call with `record_identities`, loop over the returned identities calling the appropriate `record_*` function for each, update telemetry to emit per-recorded-identity.

## Not blocking

These are independent of #681 — not a follow-up in the strict sense, just three more instances of the same idiom. No urgent user-visible impact since the dominant LocalRuntime path is already fixed.

Related: #662 (closed, original repro), #681 (open, `EmbeddedStageZero` full-prefill fix).

ndizazzo:
Closes #683.

## Summary
- record every eligible prefix identity for resident activation frames
- preserve identity/frame token-count continuity and reject ambiguous cached records
- apply per-identity accounting and telemetry to chunked/full binary and OpenAI prefill paths
- extract binary activation-cache accounting into its owning module

## Validation
- `cargo test -p skippy-server --lib` (277 passed)
- `cargo check -p mesh-llm`
- `just test-all`

## Summary by CodeRabbit

* **New Features**
  * Expanded activation-cache support to record/restore multiple matching activation entries.
  * Added per-entry activation-cache telemetry for binary KV decisions (including recorded and evicted metrics).
  * Centralized “completed prefill” binary KV recording for more consistent accounting.
* **Bug Fixes**
  * Strengthened activation-cache identity validation to avoid incorrect restores when token counts or payload/descriptor sizes differ.
  * Improved eligibility and shape checks to reduce ambiguous or mismatched restorations.
* **Tests**
  * Updated and expanded coverage for multi-record activation flows, restoration correctness, and rejection scenarios.

### DIFF
--- a/crates/skippy-server/src/binary_transport.rs
+++ b/crates/skippy-server/src/binary_transport.rs
@@ -1,3 +1,4 @@
+mod activation_cache;
 mod binary_kv;
 mod binary_messaging;
 mod decode_batcher;
--- a/crates/skippy-server/src/binary_transport/activation_cache.rs
+++ b/crates/skippy-server/src/binary_transport/activation_cache.rs
@@ -1,0 +1,82 @@
+use serde_json::json;
+use skippy_protocol::{StageConfig, binary::StageWireMessage};
+
+use crate::{
+    kv_integration::{KvStageIntegration, ResidentActivationRecord},
+    telemetry::Telemetry,
+};
+
+use super::binary_kv::{BinaryKvRecordResult, binary_message_kv_attrs};
+
+pub(super) fn add_binary_activation_records(
+    result: &mut BinaryKvRecordResult,
+    config: &StageConfig,
+    kv: &KvStageIntegration,
+    telemetry: &Telemetry,
+    session_id: &str,
+    message: &StageWireMessage,
+    records: &[ResidentActivationRecord],
+) {
+    for record in records {
+        accumulate_record(result, record);
+        let mut attrs =
+            binary_message_kv_attrs(config, kv, session_id, message, record.token_count);
+        attrs.insert("skippy.kv.decision".to_string(), json!("activation_record"));
+        attrs.insert(
+            "skippy.activation_cache.payload_bytes".to_string(),
+            json!(record.payload_bytes),
+        );
+        attrs.insert(
+            "skippy.activation_cache.entries".to_string(),
+            json!(record.entries),
+        );
+        attrs.insert(
+            "skippy.activation_cache.resident_bytes".to_string(),
+            json!(record.resident_bytes),
+        );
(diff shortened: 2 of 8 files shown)


ndizazzo:
@coderabbitai review

i386:
Codex review feedback:

I found one blocking issue in the activation-cache lookup path.

`record_resident_activation` now records multiple identities, for example an exact prompt identity plus a grid-floor shared-prefix identity. However, the downstream activation restore path in `maybe_lookup_binary_prefill` still calls `restore_resident_activation(config, &base, token_start, token_ids, ...)` with the full request token slice, and `restore_resident_activation` builds a single exact `prefill_identity` rather than walking the `lookup_identities` candidate set. It then passes only that exact activation identity to `restore_binary_prefix`.

That means the newly stored shorter activation frames are not considered for a later prompt that extends the same prefix. The new unit test only calls restore with `shared_tokens = &lookup_tokens[..2176]`, so it does not exercise the actual full-token lookup path used by binary downstream restore.

Please update activation restore to walk the same lookup candidates used for resident KV and pass the matching prefix token slice/identity through, or add an end-to-end test showing a longer prompt can hit the newly recorded shared-prefix activation. As written, the multi-record activation recording adds extra cache entries/telemetry, but the normal binary downstream path still appears to restore only exact activation identities.

i386 (approved):

PR #1042, opened 2026-07-21
fix: read-only model download caches

### LINKED ISSUE #980: models download fails with "Read-only file system (os error 30)"
**Bug**: models download fails with "Read-only file system (os error 30)" 
regardless of HF_HOME/TMPDIR destination — matches macOS sealed root FS error

**Environment**
- macOS, Apple M3 Max (arm64)
- mesh-llm binary: ~/.local/bin/mesh-llm (unsigned, no entitlements, not sandboxed)
- No ~/.mesh-llm/config.toml present

**Steps to reproduce**
mesh-llm models download unsloth/Qwen3-235B-A22B-Instruct-2507-GGUF:Q4_K_M

**Error**
Read-only file system (os error 30): Cache Hugging Face asset 
unsloth/Qwen3-235B-A22B-Instruct-2507-GGUF/Q4_K_M/Qwen3-235B-A22B-Instruct-2507-Q4_K_M-00001-of-00003.gguf@main

**What I ruled out**
- Direct writes (dd, mkdir, symlinks, hardlinks-equivalent) to the exact target 
  path succeed without any error, including 10GB+ file writes
- Tried with default HF_HOME, then explicit HF_HOME on an external APFS volume 
  (mount opts: apfs, local, nodev, nosuid, journaled, noowners)
- Tried explicit TMPDIR on the same external volume
- Same error occurs with --direct flag (bypasses catalog layer-package resolution)
- Same error across multiple different repos/quants
- codesign shows no entitlements/sandbox on the binary

**Suspicion**
`touch /test_root_write.txt` on this Mac reproduces the exact same error message 
("Read-only file system"), matching macOS's sealed system root (Catalina+ SIP). 
This suggests mesh-llm may be resolving some internal working/cache path against 
the filesystem root `/` rather than the user's home or configured cache dir.

ndizazzo:
## Summary

- validate both Hugging Face Hub and Xet cache directories before worker threads start
- fall back to a writable mesh-llm application-data directory with clear diagnostics when configured paths are read-only
- centralize cache selection across download and inventory paths and document the relevant overrides
- add filesystem edge-case coverage, including real OS-protected paths and os error 30 diagnostics

## Testing

- cargo test -p model-hf --lib
- cargo test -p mesh-llm-node --lib
- cargo test -p mesh-llm-host-runtime --lib models::catalog::tests
- cargo check/clippy for model-hf, mesh-llm-node, and mesh-llm
- just build
- read-only Hub/Xet binary smoke test
- just test-all

Closes #980

## Summary by CodeRabbit

- **New Features**
  - Model downloads now pre-resolve and validate Hugging Face Hub and Xet cache paths at startup (before worker threads begin).
  - If a configured cache destination is read-only/unusable, mesh-llm falls back to writable mesh-llm data locations and emits warnings with the rejected path and OS error.
  - Cache locations are configurable via `HF_HUB_CACHE` / `HF_XET_CACHE`, with `MESH_LLM_DATA_DIR` controlling the fallback root.
- **Bug Fixes**
  - Download failures now include more detailed cache-location diagnostics.
- **Documentation**
  - Updated CLI and usage docs to reflect pre-download validation, warning behavior, and cache configuration/fallback.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/command_support.rs
+++ b/crates/mesh-llm-host-runtime/src/command_support.rs
@@ -45,8 +45,8 @@ pub mod models {
         installed_model_display_name, installed_model_huggingface_ref,
         layered_package_layer_count_for_path, layered_package_total_bytes_for_path,
         load_model_usage_record_for_path, model_usage_cache_dir, plan_model_cleanup,
-        remote_catalog_model_draft_ref, remote_catalog_model_ref, run_update,
-        scan_installed_models, scan_installed_models_in, search_catalog_json_payload,
+        prepare_download_directories, remote_catalog_model_draft_ref, remote_catalog_model_ref,
+        run_update, scan_installed_models, scan_installed_models_in, search_catalog_json_payload,
         search_catalog_models, search_huggingface, search_huggingface_json_payload,
         show_exact_model, show_model_variants_with_progress,
     };
--- a/crates/mesh-llm-host-runtime/src/models/catalog.rs
+++ b/crates/mesh-llm-host-runtime/src/models/catalog.rs
@@ -163,8 +163,11 @@ fn ensure_cached_hf_asset(api: &hf_hub::HFClientSync, asset: &HfAsset) -> Result
         .send()
         .with_context(|| {
             format!(
-                "Cache Hugging Face asset {}/{}@{}",
-                asset.repo, asset.file, asset.revision
+                "Cache Hugging Face asset {}/{}@{}. {}",
+                asset.repo,
+                asset.file,
+                asset.revision,
+                model_hf::download_cache_diagnostic(),
             )
         })
 }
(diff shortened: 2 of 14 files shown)


michaelneale (approved):

i386 (approved):

i386:
Codex review feedback:

I rechecked the current head. I do not see an additional blocking issue in the shipped binary path: `main.rs` prepares the download directories before the Tokio runtime starts, applies both `HF_HUB_CACHE` and `HF_XET_CACHE`, and the catalog/store/node lookup helpers now share the centralized cache path resolution.

The remaining scope question is the one already raised in the thread: library or embedded callers using `HfModelRepository::from_env()` or the default builder still only resolve paths; they do not run `prepare_download_directories()` unless their entrypoint does it first. If #980 is intended to cover non-CLI consumers too, that initialization should move into shared construction or be required in those entrypoints. If this PR is scoped to the shipped CLI, the current shape looks consistent with that scope.

PR #1043, opened 2026-07-21
chore: improve embedded native-runtime compatibility guidance

### LINKED ISSUE #1016: SDK docs: document native-runtime version-check pattern; clearer error on embedded serve

ndizazzo:
Closes #1016

## What changed

- document the exact SDK release and Skippy ABI compatibility check before embedded serving
- expose the host-runtime initializer through the serving SDK surface
- preflight embedded serve startup against compatible cached native runtimes without implicit downloads
- return an actionable error containing the required versions, cache location, and install API
- add tests for the documented check/install/initialize/start flow and missing-runtime error path

## Why

Embedded SDK users could previously reach a low-level FFI failure when the process had no compatible native runtime loaded. The shipped binary performs native-runtime loading during startup, but embedded serving did not make that prerequisite clear or fail early with corrective guidance.

## Validation

- focused runtime-install, host-runtime, and SDK tests
- focused cargo check and warning-denying Clippy checks
- cargo fmt --all --check
- just test-all

## Summary by CodeRabbit

* **New Features**
  * Embedded serving now performs a native-runtime compatibility check during startup (when the dynamic-native-runtime option is enabled) and fails fast if requirements aren’t met.
  * Added/expanded public SDK APIs to initialize the embedded host runtime and to verify exact MeshLLM version + Skippy ABI compatibility.
* **Bug Fixes**
  * Startup now reports clearer “missing native runtime” guidance when a compatible cached runtime isn’t available.
* **Documentation**
  * Updated embedded-node instructions to follow: compatibility check → install (if needed) → initialize host runtime → start serving (cached runtime only).
* **Tests**
  * Added tests covering compatibility behavior and the README’s documented execution order.

### DIFF
--- a/crates/mesh-llm-embedded-runtime/src/lib.rs
+++ b/crates/mesh-llm-embedded-runtime/src/lib.rs
@@ -1,5 +1,6 @@
 #![forbid(unsafe_code)]
 
+pub use mesh_llm_host_runtime::initialize_host_runtime;
 pub use mesh_llm_host_runtime::sdk::{
     EmbeddedChatMessage, EmbeddedMeshAdmissionConfig, EmbeddedMeshDiscoveryMode,
     EmbeddedMeshHttpConfig, EmbeddedMeshLogFormat, EmbeddedMeshNetworkConfig,
--- a/crates/mesh-llm-host-runtime/src/sdk.rs
+++ b/crates/mesh-llm-host-runtime/src/sdk.rs
@@ -20,6 +20,7 @@ use tempfile::NamedTempFile;
 use tokio::sync::Mutex;
 
 mod embedded_config;
+mod embedded_startup;
 
 pub use embedded_config::*;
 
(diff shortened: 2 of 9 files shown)


ndizazzo:
@coderabbitai review

ndizazzo:
@coderabbitai review

ndizazzo:
@coderabbitai review

i386 (commented):
The new native-runtime example is not self-contained as written: its dependency block only adds `mesh-llm-sdk`, while the code uses `#[tokio::main]`, `anyhow::Result`, and `anyhow::ensure!`. Copying it into a fresh crate therefore fails to resolve `tokio` and `anyhow`. Please include the required `tokio` (runtime + macros) and `anyhow` dependencies, ideally with a compile test for the full example.

i386 (approved):
LGTM

PR #1050, opened 2026-07-21
Lock split topology placement

i386:
## Why this exists

The automatic split planner is the right default for normal serving: it chooses participants and layer boundaries from the capacity currently advertised by the mesh. That flexibility is undesirable during controlled lab benchmarks, where discovery timing, peer availability, or a changed VRAM report can otherwise produce a different split between runs.

A topology lock makes placement an explicit benchmark input. The same lock file is copied to every host and pins:

- the immutable model/package identity and manifest SHA-256;
- the ordered participating nodes;
- the exact layer range assigned to each node.

This ensures every host in the lab agrees on the same split topology and makes benchmark results comparable across binaries, branches, runtime settings, and repeated runs. If the requested topology cannot be reproduced exactly, startup fails instead of silently benchmarking a different placement.

## How the lock is used

Create a versioned JSON manifest such as:

```json
{
  "version": 1,
  "model": "hf://meshllm/example-layers@immutable-revision",
  "manifest_sha256": "<sha256 of model-package.json>",
  "stages": [
    {
      "node": "micstudio.local",
      "layer_start": 0,
      "layer_end": 31
    },
    {
      "node": "studio54-3.local",
      "layer_start": 31,
      "layer_end": 47
    }
  ]
}
```

Place the same file on every serving host and launch each node with:

```bash
mesh-llm serve \
  --model hf://meshllm/example-layers@immutable-revision \
  --split \
  --split-topology-lock /path/to/topology-lock.json
```

At startup, the runtime:

1. verifies that the lock matches the resolved package and manifest digest;
2. resolves every node selector to exactly one eligible participant;
3. requires non-empty, contiguous ranges covering the complete model;
4. runs the normal context, KV-cache, headroom, and VRAM capacity checks against those exact assignments;
5. starts only if the requested topology can be realized unchanged.

Node selectors may be full iroh endpoint IDs or uniquely advertised hostnames.

## Fail-closed behavior

A locked topology is not a preference or planner hint. Membership changes do not replace its stages or collapse it to local serving. If an assigned stage is lost, the route becomes unavailable and is withdrawn after the normal stage-loss grace period.

This prevents a benchmark from continuing under a materially different execution shape while still appearing healthy.

## Implementation summary

- add the hidden `--split-topology-lock <path>` runtime option;
- add strict, versioned lock parsing and package/node identity validation;
- add capacity-aware locked planning to the shared Skippy topology planner;
- disable replacement replanning and local fallback for locked generations;
- document the lock format, startup behavior, and runtime verification surface.

## Validation

- `cargo fmt --all --check`
- `cargo check -p mesh-llm`
- `cargo clippy -p skippy-coordinator -p mesh-llm-cli -p mesh-llm-host-runtime -p mesh-llm --all-targets -- -D warnings`
- `cargo test -p skippy-coordinator --lib` (32 passed)
- `cargo test -p mesh-llm-cli --lib` (72 passed)
- `cargo test -p mesh-llm-host-runtime --lib` (1703 passed, 8 ignored)
- `just build`

## Summary by CodeRabbit

* **New Features**
  * Added `--split-topology-lock <PATH>` to pin split node order and layer ranges via a validated JSON lock (requires `--split`).
  * Split planning now supports locked stage-to-node assignments for deterministic deployments.
  * Locked split loss recovery now withdraws routes after the normal grace period instead of replanning or falling back.
* **Bug Fixes**
  * Improved fail-closed validation for locked topology parsing and planning, including clearer failure behavior when the lock can’t be reproduced.
* **Documentation**
  * Added/expanded documentation for creating, verifying, and operating locked split topologies, including CLI and design guidance.


ndizazzo (approved):
@i386 I had started chipping away at pieces of this with the config work, but right now it doesn't fully support everything you'd need for it.  The idea was to incorporate the split topology in the configuration pages (which ultimate wrote some TOML like):

```toml
[[models]]
model = "hf://meshllm/example-layers@immutable-revision"

[models.model_fit]
ctx_size = 65536

[models.throughput]
parallel = 1

[models.topology]
mode = "locked"
manifest_sha256 = "0123456789abcdef..."

[[models.topology.stages]]
node = { endpoint_id = "iroh-endpoint-id-a" }
layer_start = 0
layer_end = 31

[[models.topology.stages]]
node = { hostname = "studio54-3.local" }
layer_start = 31
layer_end = 47
```

I think the configuration TOML is the better long-term product surface, but the existing impl can't currently express it. What would be good is to accompany this PR with a definition for how we want to transition it. Here's what Codex said about the gap:

```markdown
## How far PR #1050 is from this
The PR is close at the runtime layer but not at the configuration layer:

**Already implemented:**
* immutable model/manifest verification;
* node selector resolution;
* contiguous full-model layer validation;
* capacity-aware locked planning;
* locked recovery and withdrawal behavior;
* runtime propagation.

**Still needed:**
* ModelTopologyConfig, stage, and typed node-selector structs;
* TOML serialization/deserialization;
* structured config diagnostics and canonical paths;
* built-in schema and UI metadata;
* per-model startup propagation;
* CLI-versus-config precedence;
* authoring/editor and schema fixture coverage;
* documentation and validation tests.
```

In the short term - nothing jumps out at me for this implementation, but I think we should definitely consider bringing them closer together from a product standpoint, so that we can take advantage of the configuration page for split configs in the web console.

michaelneale (approved):
I think very neeed - if uyou can get build passing.

i386:
Created #1052 to track the product transition: replace the hidden `--split-topology-lock`/external JSON surface with typed per-model `[[models]].topology` TOML configuration and web-console support, then remove the CLI option. The follow-up preserves this PR’s fail-closed validation and locked-stage recovery semantics while keeping #1050 focused on the runtime primitive.

PR #1060, opened 2026-07-22
Pipeline the standalone N-gram verify-window path

danielwinterw:
# Standalone N-gram verify-window pipelining

**TL;DR:** Lets a standalone N-gram plan (`cache`/`suffix`, no MTP) reuse the composite verify-window pipeline instead of drafting one window per round trip. Opt-in via `verify_window_pipeline_depth > 1`. Default is unchanged. Mechanism is correct and tested; throughput win is not yet demonstrated and needs a two-machine split.

## What

- Standalone plans drafted serially regardless of `verify_window_pipeline_depth`, because the pipeline enablement (`ngram_hybrid`) required an extension policy that standalone plans do not have.
- New `ngram_proposals_enabled` flag turns the composite provider on for standalone too; `ngram_max_proposal_tokens` falls back to the proposer's own limit when there is no extension.
- `embedded_generation` enables the pipeline for standalone at depth greater than 1; the serial path stays as the depth-1 fallback; the single native-MTP window path stays composite only.

## Safety

- Opt-in. Default depth 1, so existing configs are unchanged.
- Composite (MTP) behavior identical: the new flag equals the old `ngram_hybrid` for every composite plan.
- Pure N-gram is a no-op for all MTP-specific calls (already guarded).

## Validation

- Tests green: skippy-server 298, host-runtime 1709, config 98, clippy clean.
- Local 8B split at depth 3: `verify_window_max_in_flight` 0 to 3, `parallel_fraction` 0 to 0.81, accept 0.92, output correct.

## Performance

Local two-stage 8B split, one M1 GPU (pessimistic: shared compute, blocking-sleep
latency model), standalone suffix re-emit, verify_window_max_tokens sized to the
proposal limit:

| one-way RTT | serial | pipelined | ratio |
| --- | --- | --- | --- |
| 0ms | 69.6 | 58.0 | 0.83 |
| 50ms | 53.3 | 49.8 | 0.93 |
| 100ms | 42.8 | 42.7 | 1.00 |
| 150ms | 36.4 | 38.0 | 1.04 |

Pipelining crosses over at about 100ms one-way RTT and leads above it. This is a
floor: separate GPUs and a real network (two-machine split) let compute and
propagation overlap, which this single-machine test cannot, so the crossover should
move lower and the lead should grow. Below the crossover, keep depth at 1.

Chunk size matters: set verify_window_max_tokens to the proposal limit, otherwise
the pipeline runs many small round trips and loses.

## Try it

Set `verify_window_pipeline_depth = 3` on a standalone `ngram-suffix` config, run a two-stage split, check `verify_window_max_in_flight > 1` in the response `timings`.

## Summary by CodeRabbit

* **New Features**
  * Enabled standalone N-gram proposals to support speculative decoding and verification at deeper pipeline stages.
  * Improved proposal handling when native MTP or a draft model is unavailable.
  * Standalone N-gram configurations now honor their configured proposal-token limit.

* **Bug Fixes**
  * Prevented duplicate proposal execution paths during pipelined decoding.

* **Tests**
  * Added coverage for standalone N-gram proposal enablement and token horizons.

### DIFF
--- a/crates/skippy-server/src/frontend/embedded_generation.rs
+++ b/crates/skippy-server/src/frontend/embedded_generation.rs
@@ -914,10 +914,18 @@ impl StageOpenAiBackend {
             );
             let composite_sidecar_enabled =
                 native_mtp_options.ngram_hybrid && draft_guard.is_none();
+            // A standalone N-gram plan (no native MTP, no draft model) can drive
+            // the same verify-window pipeline; the single-window native-MTP path
+            // stays composite-only, so standalone drafting still falls back to the
+            // serial block at depth 1.
+            let standalone_ngram_pipelining = !request.native_mtp_enabled
+                && request.speculative.ngram.is_some()
+                && draft_guard.is_none();
             let native_mtp_verify_windows_enabled =
                 (request.native_mtp_enabled || composite_sidecar_enabled) && draft_guard.is_none();
-            let pipelined_decode_enabled =
-                composite_sidecar_enabled && verify_window_scheduler.depth() > 1;
+            let pipelined_decode_enabled = (composite_sidecar_enabled
+                || standalone_ngram_pipelining)
+                && verify_window_scheduler.depth() > 1;
             let mut verify_window_forwarder = None;
             if let Some(direct_return_path) = direct_prediction_return_path(
                 native_mtp_verify_windows_enabled,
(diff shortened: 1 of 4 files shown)


michaelneale:
I think this may be fantastic. testing...

danielwinterw:
@michaelneale how did you go with testing?

micspiral:
@danielwinterw still testing - will get a report soon - but I think may be good to get in if clean - what say you @i386 ?

micspiral:
@danielwinterw it helped a lot with inkling, many times faster in testing and is in that branch, feel free to merge away here, it will be included either way.

michaelneale (approved):

PR #1064, opened 2026-07-23
fix(mesh): owner-control listener must not share the mesh relay slot

michaelneale:
## Problem

A node with a **verified owner identity** starts a *second* iroh endpoint — the
owner-control listener (`ALPN mesh-llm-control/1`) — using the **same secret key**
as the main mesh endpoint, so both present the **same iroh endpoint id**. The
control protocol requires the shared id: it validates the dialed `target_node_id`
against the main endpoint id (`verify_control_plane_target_node`).

An iroh relay keeps only **one active connection per endpoint id**
(`iroh-relay` `server/clients.rs`: a second same-id registration does
`std::mem::replace(&mut state.active, client)` and the old one "can still send
messages, but won't receive anything"). The control listener enables relay and
binds **after** the main mesh endpoint, so it **evicts the main endpoint's relay
registration** — logged as:

```
Relay server reports problem: Another endpoint connected with the same endpoint id.
No more messages will be received.
```

The main mesh endpoint then goes dark on the relay. All relay-delivered mesh
traffic (gossip, joins, inference routing) stops. Any peer that cannot reach the
node **directly** never completes its join, never syncs the model catalog, and
every inference returns 429/503.

## Impact

Consumers on a network where the direct path does not establish (common: NAT /
LAN-hairpin) are left with **no relay fallback**, because the fallback path was
silently destroyed. Consume appears to "connect but never join".

## Fix

Bind the owner-control listener with `RelayMode::Disabled` so it never contends
for the shared id's relay slot. The main mesh endpoint is **unchanged** and keeps
relay + NAT traversal; it becomes the sole relay registrant for the id. Also
`clear_ip_transports()` on the control endpoint so an explicit IPv4 bind does not
leave iroh's implicit `[::]:0` socket (which can trip `MultipathNotNegotiated`
with relay disabled) — mirroring the main endpoint's LAN-only handling.

Removed the now-unused `configure_control_relay` helper, its `relay` parameter,
and the control-endpoint relay `online()` wait.

## Trade-off

Owner-control becomes reachable over its **direct / advertised address only**;
relay-assisted remote owner-control across NAT is unsupported while control and
mesh share one id. Giving control its own relay presence would require a distinct
endpoint id, which the control target validation forbids (would be a
protocol-level change). Buzz — the downstream that hit this — never dials
owner-control, so this trade-off is invisible there.

## Verification

- New regression test `control_plane_listener_token_carries_no_relay_urls`
  asserts the control token advertises zero relay URLs while keeping the main
  endpoint id. All 6 `control_plane_listener` tests pass.
- Live A/B, two Macs, owner identity present both sides:
  - **Before:** consumer logs the "same endpoint id" eviction, never joins.
  - **After:** consumer joins over the relay and inference returns HTTP 200.

## Notes

- Also available as a minimal backport onto the `v0.73.1` release layout
  (branch `micn/fix-owner-control-relay-collision-v0.73.1`) for downstreams
  pinning that tag.
- **A separate, pre-existing direct-hole-punch reliability problem** (native
  mesh's direct LAN path is flaky where raw iroh's is not) is filed separately as
  a blocker issue — it is **not** caused by this change (reproduces on unpatched
  `v0.73.1` too) and is orthogonal to this relay-slot fix.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/node.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/node.rs
@@ -258,23 +258,6 @@ pub(crate) fn init_owner_runtime(
     })
 }
 
-pub(crate) fn configure_control_relay(
-    mut builder: iroh::endpoint::Builder,
-    relay: Option<RelayConfig<'_>>,
-) -> Result<iroh::endpoint::Builder> {
-    if let Some(relay) = relay.filter(|relay| relay.policy.uses_relay()) {
-        let urls = effective_relay_urls(relay.policy, relay.urls);
-        tracing::info!("Owner-control relay: {:?}", urls);
-        builder = builder.relay_mode(iroh::endpoint::RelayMode::Custom(relay_map_from_urls(
-            &urls,
-            relay.auths,
-        )?));
-    } else {
-        builder = builder.relay_mode(iroh::endpoint::RelayMode::Disabled);
-    }
-    Ok(builder)
-}
-
 pub(crate) fn default_plugin_event_source(endpoint_id: EndpointId, source_peer_id: &mut String) {
     if source_peer_id.is_empty() {
         *source_peer_id = endpoint_id_hex(endpoint_id);
@@ -1060,7 +1043,6 @@ impl Node {
             owner_config
                 .as_ref()
                 .and_then(|config| config.control_advertise_addr),
-            relay.policy.uses_relay().then_some(relay),
         )
         .await?;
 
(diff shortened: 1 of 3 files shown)


michaelneale:
Related blocker (orthogonal, pre-existing): #1065 — native mesh direct LAN hole-punch times out where raw iroh succeeds. This PR restores relay fallback; #1065 tracks getting the fast direct path reliable.

michaelneale:
this may not be needed - will be superseeded by other work. DO NOT MERGE

michaelneale:
may actually be building on this - I think this is necessary

michaelneale:
Superseded by #1071, which contains these two commits plus the iroh 1.0.3 modernization. Consolidating to one PR.

Issue #1065, opened 2026-07-23
Native mesh direct LAN hole-punch times out where raw iroh succeeds (falls back to relay)


michaelneale:
## Summary (blocker)

On a **plain LAN** (two hosts, same subnet, no VPN, host firewalls off), the
native mesh **direct QUIC hole-punch to a peer's LAN address times out** and the
connection falls back to relay. **Raw iroh, single fresh-key endpoint, over the
same relay config and same two hosts, establishes the direct LAN path reliably.**

This means real consumers ride the relay (~200 ms) instead of the direct LAN path
(~12 ms), and — combined with any relay disruption — can fail to join entirely.
It also masks the more serious case: if the direct path is the only reachable
one and it won't establish, the peer never joins.

## Environment

- 2 macOS/arm64 hosts, same home LAN/subnet, same gateway.
- No VPN interfaces active; macOS application firewall disabled both sides.
- ICMP between hosts ~8 ms, 0% loss.
- Custom iroh relays (self-hosted). iroh 1.0.1.
- Reproduces on **released `v0.73.1`** (unpatched) — not specific to any local
  change.

### A) Raw iroh — DIRECT works (control / baseline)

Minimal iroh 1.0.1 probe: `listen` binds one endpoint (custom relay mode) and
prints its dial info; `connect` dials by id + direct addrs, opens a bi-stream.

- Host A: `probe listen --relay <custom>`  → prints `ID`, `ADDRS` (LAN + reflexive), `RELAY`.
- Host B: `probe connect --id <ID> --addrs <A_LAN_addr> --relay-url <RELAY> --attempts 4`
- With `RUST_LOG=iroh::_events::path=debug`:
  ```
  path::selected network_path=Ip(B_LAN -> A_LAN:port)
  4/4 bi-stream echoes, ~12-15 ms
  ```
  **Direct LAN path selected and stable.**

### B) Native mesh — DIRECT times out, falls back to relay

- Host A: `mesh-llm serve --model <small-gguf>` (owner identity present).
- Host B: `mesh-llm client --join <token-from-A>` (owner identity present).
- With `RUST_LOG=iroh::_events::path=debug,iroh::_events::qnt=debug`:
  ```
  do_holepunching: local_candidates={B_LAN}  remote_candidates={A_public, A_LAN}
  path::open  PathId(0) network_path=Ip(B_LAN -> A_LAN:port)
  ... ~15s ...
[...]

PR #1066, opened 2026-07-24
fix(mesh): enable QUIC Address Discovery (QAD) on managed relays

michaelneale:
## Problem

Relay maps were built with `iroh::RelayConfig::new(url, None)`. The trailing
`None` **explicitly disables** iroh's QUIC Address Discovery (QAD).

This was not a deliberate decision. The call predates the iroh **0.98 → 1.0**
upgrade: on 0.98 `new(url, None)` meant "use QAD defaults"; on 1.0 it means
"QAD **off**", and the blessed constructor became `RelayUrl::into()`. The stale
call silently flipped QAD off across the upgrade and nobody caught it.

**Consequence:** `net-report` never learns a reflexive public UDP candidate, so
a node's `local_candidates` are LAN-only and direct-path (hole-punch) upgrades
across NAT cannot form — nodes fall back to relay. This degrades direct
connectivity for **every** node on the managed relays (single- or dual-endpoint,
owner or not). The Inkling multi-node work hit this independently.

## Fix

Construct relay configs via `RelayUrl::into()`, preserving iroh's default QAD
(UDP 7842). Applied at all three sites that build relay maps:

- `mesh-llm-host-runtime` data plane — `connections.rs::relay_map_from_urls`
- `mesh-client` builder dial path — `builder.rs`
- `mesh-client` owner-control dial path — `control_plane.rs`

Also adds the managed `euc1-1` and `use1-1` relays to the default public set.

## Safety

- **Restores an iroh default**, not a new feature.
- QAD is **client-driven** and iroh gates it behind `has_ip_transports`, so
  relay-only / LAN-only endpoints are unaffected (no `MultipathNotNegotiated`
  risk in those modes).
- The only thing revealed is the node's public IP:port to **its own relay** —
  which already terminates its traffic. No new trust surface.

## Verification

- With the fix, a client's `local_candidates` include the reflexive public
  address and the invite token carries it (confirmed on a two-node cross-NAT
  A/B).
- Regression test asserts the relay config exposes QAD on port 7842.

## Scope note

QAD is **necessary but not sufficient** for owner-identity nodes to hole-punch
directly. The dual same-endpoint-id issue (#1065) is a separate, independent
defect that also blocks direct for owner nodes. Full matrix in #1065. This PR is
the standalone QAD win that benefits the whole network regardless of that.

Refs #1065

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection reliability by preserving QUIC Address Discovery for relay connections, supporting better NAT traversal.
  * Enabled relay connections to use the default QUIC relay port.

* **Improvements**
  * Expanded the default public relay selection with additional managed relay endpoints in Europe and the US East region.

### DIFF
--- a/crates/mesh-client/src/client/builder.rs
+++ b/crates/mesh-client/src/client/builder.rs
@@ -566,7 +566,10 @@ fn relay_map_from_endpoint_addr(addr: &EndpointAddr) -> Option<iroh::RelayMap> {
     let configs: Vec<_> = addr
         .relay_urls()
         .cloned()
-        .map(|url| iroh::RelayConfig::new(url, None))
+        // Preserve iroh's default QUIC Address Discovery (QAD). `new(url, None)`
+        // disables it, preventing reflexive candidate discovery and direct-path
+        // upgrades across NAT (see issue #1065). `RelayUrl::into()` keeps QAD on.
+        .map(|url| -> iroh::RelayConfig { url.into() })
         .collect();
     if configs.is_empty() {
         None
--- a/crates/mesh-client/src/client/control_plane.rs
+++ b/crates/mesh-client/src/client/control_plane.rs
@@ -835,7 +835,10 @@ fn relay_map_from_endpoint_addr(addr: &EndpointAddr) -> Option<iroh::RelayMap> {
     let configs: Vec<_> = addr
         .relay_urls()
         .cloned()
-        .map(|url| iroh::RelayConfig::new(url, None))
+        // Preserve iroh's default QUIC Address Discovery (QAD). `new(url, None)`
+        // disables it, preventing reflexive candidate discovery and direct-path
+        // upgrades across NAT (see issue #1065). `RelayUrl::into()` keeps QAD on.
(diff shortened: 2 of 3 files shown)


ndizazzo (approved):
Good catch! we can probablyharden it broader regression coverage before merging...

-> Targeted tests for both mesh-client relay-map conversion paths (builder and control_plane) to lock in that QAD stays enabled when relays are built from endpoint addresses.

-> Expand host-runtime relay-map assertions so the QAD requirement is validated across all relays in a map, not just the first.

-> New explicit checks around default relay set behavior to protect the default-public relay list changes (euc1-1, use1-1) from drifting. Maybe we want that to be a config some day, so for now assertions seem fine to me

-> Add/extend a behavior-level relay test to validate end-to-end health with QAD-on relay configs.

michaelneale:
@ndizazzo can you make follow up issue(s) for that as my agent and brains are full to do more and prove on this.

i386 (approved):

i386:
Ive got an agent working on @ndizazzo feedback. pr incoming.

PR #1071, opened 2026-07-24
mesh: direct-path hole-punching via modern iroh (1.0.3) + owner-control relay fix

michaelneale:
## Goal

Let iroh 1.0.3 own NAT traversal and path management so mesh relay-to-direct migration behaves like bare iroh. This also includes the owner-control relay-slot fix from #1064.

## Changes

- Keep owner-control on its own loopback-only listener instead of sharing the mesh relay slot.
- Upgrade iroh from 1.0.1 to 1.0.3.
- Use iroh's wildcard QUIC binds and transport defaults instead of auto-pinning one LAN IPv4 or overriding path liveness.
- Remove the application-level direct-path repair sender/controller; iroh owns probing and relay-to-direct migration.
- Retain the legacy direct-path request receiver and its admission/cooldown checks so mixed-version peers remain compatible.
- Remove the now-unused LAN-interface detector and direct `if-addrs` dependency.
- Restore `HOME` after serialized owner-control tests so the full host-runtime suite is deterministic.

QAD from #1066, admission control, owner identity, relay fallback, and both mesh ALPN generations remain intact.

## Qlog diagnosis

The failed mesh trace negotiates the same QUIC multipath/NAT-traversal parameters as the working bare dialer (`initial_max_path_id=7`, `max_remote_nat_traversal_addresses=32`). It emits `PATH_CHALLENGE` probes, then retries because no challenge or response arrives and the IP path never validates.

Follow-up probes falsified the earlier mesh-specific leading theories:

- Bare iroh still migrated to direct with port mapping disabled.
- Three concurrent bare connections to the same peer all migrated to direct.
- Bare probes remained direct while the mesh runtime was busy.
- The qlog-enabled mesh transport reached direct paths repeatedly on USE1, EUC1, and APS1 relays (3/3 each); relay generation was not causal.

The original failure is intermittent rather than a deterministic multiple-connection race, and matches the class of upstream path-selection failure tracked in n0-computer/iroh#4247. This PR therefore removes the redundant mesh workaround instead of adding another path manager or pinning a relay subset.

## Validation

- `just build`
- `MESH_LLM_DYNAMIC_NATIVE_RUNTIME=0 just release-build`
- `cargo fmt --all --check`
- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-host-runtime --lib` — 1706 passed, 0 failed, 8 ignored
- Exact static release build, M5 serve -> Mini client: 3/3 fresh connections became direct and remained direct at the five-second recheck. One began on relay and migrated to direct.
- Mini `/v1/models` returned the M5-hosted model, and a chat completion tunneled through the Mini client returned the requested response.

The final lab runs used normal relay selection; no newer-relay-only source change is included.

## Summary by CodeRabbit

* **Improvements**
  * Updated network connection handling to rely on automatic transport and candidate discovery.
  * Simplified control-plane connectivity by disabling relay addresses for owner-control endpoints.
  * Increased supported concurrent mesh streams.
  * Preserved compatibility with older peers that send legacy direct-path repair requests.
* **Bug Fixes**
  * Improved test isolation when configuring temporary home directories.
  * Removed legacy network-interface detection and direct-path maintenance behavior that could cause unnecessary connection management.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,7 +183,7 @@ dependencies = [
  "objc2-foundation",
  "parking_lot",
  "percent-encoding",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
  "x11rb",
 ]
 
@@ -790,7 +790,7 @@ version = "3.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6dee98b0db6a962de883bf5d20362dee4d7ca0d12fe39a7c6c73c844e1cd7c1f"
 dependencies = [
- "darling 0.23.0",
+ "darling",
  "ident_case",
  "prettyplease",
  "proc-macro2",
@@ -928,9 +928,9 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "cfg_aliases"
-version = "0.2.1"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
+checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
 
 [[package]]
 name = "chacha20"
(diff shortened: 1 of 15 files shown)


i386:
Codex review feedback:

Reviewed the current non-draft head and did not find a blocking code issue.

The owner-control relay fix looks internally consistent: the control listener keeps the same endpoint id needed for target-node validation, disables relay registration for that same-id listener, binds the control endpoint directly, and the tests assert that the generated control token carries no relay URLs while preserving explicit advertised-address behavior.

The direct-path sender removal also leaves the legacy receiver/admission/cooldown path in place for mixed-version peers. The main residual risk is operational rather than a code finding: the change relies on iroh 1.0.3 defaults and upstream NAT traversal instead of the previous keep-alive/idle-timeout overrides and app-level repair sender. The validation notes cover that with lab runs, so I do not have an additional change to request.

Issue #1072, opened 2026-07-25
bug(client): client-only node can elect itself to serve and download a model


michaelneale:
## Summary

`mesh-llm client --auto` can elect its zero-capacity client node as a model host and begin downloading a model. Client mode is documented as proxy-only and should never serve, download, or participate as an eligible host.

## Reproduction

1. Run:

   ```bash
   mesh-llm client --auto --log-format json
   ```

2. Join a mesh with active demand for an unserved catalog model whose curated `size` is not a byte-size string. The current public mesh reproduced this with `unsloth/gpt-oss-120b-GGUF@main:UD-Q4_K_XL`, whose catalog size is `"36 layers"`.
3. Wait for the five-second model-assignment pass.

Observed events include:

```text
Assigned to serve unsloth/gpt-oss-120b-GGUF@main:UD-Q4_K_XL (unserved, will download, by demand)
host_elected ... capacity_gb=0.0 role=host
Downloading unsloth/gpt-oss-120b-GGUF@main:UD-Q4_K_XL for mesh...
```

The process then starts a roughly 50 GB download instead of entering passive client mode.

## Expected behavior

A node launched with `mesh-llm client` must always remain client-only:

- never be returned by model assignment;
- never emit `host_elected` for itself;
- never download or serve a model;
- proceed directly to the passive proxy path after joining.

## Reproduction scope

Reproduced independently with:

- the installed `mesh-llm 0.72.1` release;
- the #1071 branch build while validating its public-mesh transport behavior.

This is pre-existing and is not introduced by #1071.

## Likely cause

`select_run_auto_model_path` calls `pick_model_assignment(ctx.node, ctx.local_models)` before applying its `ctx.is_client` condition. The client therefore participates in demand assignment with zero VRAM.

For the model above, `parse_size_str("36 layers")` returns `0` because it only recognizes `GB` and `MB`. The downloadable-model check then accepts `needed == 0` as fitting `my_vram == 0`, electing the client.

## Suggested fix and coverage

1. Make client-only status the primary invariant: skip `pick_model_assignment` entirely when `ctx.is_client`.
2. Defensively treat unknown, unparseable, or zero catalog sizes as ineligible for automatic download assignment.
3. Add tests proving:
   - active unserved demand cannot promote a client node;
   - a non-byte catalog size such as `"36 layers"` cannot fit zero capacity;
   - client mode reaches the passive proxy path without model resolution or download.

i386:
[...]

PR #1074, opened 2026-07-25
Keep client-only nodes out of model election

michaelneale:
Client-only nodes now remain proxy-only after joining a mesh, even when active demand includes an unserved model. This prevents zero-capacity clients from electing themselves, downloading a large model, and delaying management API startup.

Validation:
- focused host-runtime regression test passed
- host-runtime Clippy passed with warnings denied on the source branch
- formatting and diff checks passed on this focused branch

## Summary by CodeRabbit

* **Bug Fixes**
  * Client-mode nodes no longer receive automatic model assignments during host election or model selection.
  * Improved role-aware handling of model selection.

* **Tests**
  * Added a new async test verifying model assignment is derived from node role (returns no assignment for client role, assigns one for worker role).

### DIFF
--- a/crates/mesh-llm-host-runtime/src/runtime/auto_join.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/auto_join.rs
@@ -520,6 +520,18 @@ pub(super) async fn pick_model_assignment(
     None
 }
 
+/// Pick a model assignment only when this node's mesh role can serve models.
+pub(super) async fn pick_model_assignment_for_role(
+    node: &mesh::Node,
+    local_models: &[String],
+) -> Option<String> {
+    if matches!(node.role().await, NodeRole::Client) {
+        None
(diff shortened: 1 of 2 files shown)


i386 (approved):
Reviewed current head `806e8f5`. No blocking findings.

The role gate now runs before demand-based assignment, the existing `ctx.is_client` fallback guard still keeps clients out of the auto-pack path, and the regression test verifies both client rejection and normal worker delegation. The green client-auto boot job also reaches `passive_mode` / `Client ready` without electing or downloading a model.

Non-blocking follow-up: #1072 also documents that `parse_size_str("36 layers")` returns `0`. This PR makes that unable to elect a client, but unknown catalog sizes can still be treated as fitting worker capacity. Please keep that fail-closed size parsing fix tracked separately if it is intentionally outside this focused PR.

PR #1078, opened 2026-07-26
Fix Metal small-batch matmul parity for GLM verification

### LINKED ISSUE #1076: GLM-DSA batched `verify_tokens` diverges from sequential decode at width 4
## Summary

On GLM-DSA, a linear four-row `StageSession::verify_tokens` call can produce a different greedy target prediction from the equivalent sequence of one-token target steps. Widths 1 and 2 matched in the same run; width 4 diverged.

This is a correctness issue for speculative decoding: batched verification must remain causally equivalent to the target model's sequential decode path.

## Environment

- Mesh/Skippy public `main`: `cdd793b1cfd10cde1a0c6b56b7b8f72a9b003912`
- macOS arm64, Metal backend
- GLM-4.7 Flash MTP GGUF, full-model single-node runtime
- greedy sampling
- one `StageModel`, one session/lane
- F16 K/V cache
- explicit `n_batch = 512`, `n_ubatch = 512`

## Generic reproduction

1. Open the full GLM model and create one session.
2. Prefill a causal prompt, retaining the final prompt token as `current`.
3. From the same prefix state, derive an ordinary greedy target continuation using repeated one-token decode.
4. Restore or trim the test session to the exact prefix position.
5. For verification width `R`, construct the normal linear speculative input:

   ```text
   [current, target[0], ..., target[R-2]]
   ```

6. Call `StageSession::verify_tokens` once with the complete input and compare its predictions with `target[0..R]`.
7. Repeat from the identical restored prefix for widths 1, 2, 4, and 9.

Observed in the failing run:

- width 1: exact
- width 2: exact
- width 4: prediction mismatch

The comparison failed closed before performance timing or speculative commit.

## Suspected scope

Source inspection suggests this may be specific to GLM's direct-sparse short-prefill selection:

- singleton verification has a decode-shaped execution;
- multi-row verification within the default 2–8-token range is eligible for the direct-sparse-prefill path;
- `skippy_verify_token_batch` executes with `SKIPPY_GLM_DSA_PHASE_HINT_VERIFY`.

`SKIPPY_GLM_DSA_DISABLE_DIRECT_SPARSE_PREFILL=1` is the relevant isolating control, but the fallback result should be independently confirmed before treating that as the root cause.

## Expected behavior

For every row `i`, the batched verifier's greedy prediction should equal the prediction produced by sequential target decode from the same causal prefix and preceding accepted inputs. Restoring or trimming either path should leave equivalent authoritative state.

## Impact

A mismatching verifier can incorrectly accept or reject draft tokens and can change the generated target stream. The optimized path should not be enabled by default for GLM verification until exact sequential parity is demonstrated.

## Suggested next steps

- Add a GLM regression test comparing batched and sequential predictions at widths 1, 2, 4, and 9.
- Cover both short and long-context prompts.
- Record the selected GLM attention path in verifier telemetry.
- Run the same test with direct sparse prefill enabled and disabled.
- If the dense fallback is exact, fail closed to it for verification until direct-sparse parity is fixed.
- Check post-verification continuation and all GLM state, including lightning-indexer state, rather than comparing only MLA KV.

i386:
Fixes #1076.

## Summary

This PR fixes GLM-DSA batched `verify_tokens` parity at the Metal matrix-multiplication boundary while preserving genuine one-call multirow execution.

- add row-parallel, batch-invariant F32, Q5_0, Q4_K, and Q6_K `mul_mv_ext` kernels that use canonical one-row arithmetic and reduction order;
- allocate the canonical Q5_0 threadgroup-memory footprint for the extended kernel;
- assign each batch row to an independent Metal threadgroup instead of replaying rows serially;
- cover the affected small-batch dispatch through width 16, including GLM's 192-column Q5_0 projection;
- expose an independent `--sample-width` in `skippy-bench verify-window-local`, require that timed widths are parity-checked, and reject widths above the proven ceiling during CLI parsing;
- compare typed generation signal windows and retain full target-stream, continuation, native-position, exported-state, token-signal, and top-2-logit diagnostics.

No environment-variable guard or serial verifier fallback is required by this fix.

## Root cause

Metal changed matrix multiplication implementation as verification width grew:

- one-row execution used canonical `mul_mv`;
- small batches used `mul_mv_ext` with a different reduction shape;
- wider batches could cross into `mul_mm`;
- the GLM 192-column Q5_0 projection also missed the ordinary small-batch eligibility rule.

Those paths were within normal backend numerical tolerances, but were not batch-invariant. The drift changed canonical exported state and token signals, and could later change greedy continuation. Direct sparse prefill and FlashAttention were not the root cause.

The repaired kernels keep one batched graph and parallel GPU execution while making every row follow the canonical one-row computation.

## Scope

The batch-invariant path is bounded to widths 16 or lower:

- F32 and Q5_0: widths 2-16;
- Q4_K and Q6_K: widths 4-16.

Existing dispatch remains unchanged outside those source types and widths. `verify-window-local` rejects larger widths until equivalent exactness is implemented and proven.

The parity harness deliberately builds an always-accepted greedy target stream so it measures kernel/state parity rather than proposal quality. Native-MTP proposal quality and acceptance remain the responsibility of the dedicated MTP benchmarks.

`GenerationSignalWindow` is now compared as a typed value. It remains advisory because one batched call and N serial calls necessarily have different call-history window shapes; canonical model state and `TokenSignal` remain exactness gates.

## Physical validation

Validation used Apple M3 Ultra Metal, the public GLM-4.7 Flash MTP Q4_K_M GGUF, greedy sampling, one session, F16 K/V, and FlashAttention auto.

### Exactness

Widths 1, 2, 4, 9, and 16 each passed:

- exact target stream;
- exact 256-token continuation;
- exact native position;
- byte-exact full exported state;
- exact token signal;
- exact top-2 tokens and margins.

This is 5/5 parity cells and 1,280 checked continuation decodes. Report SHA-256: `bfd74dae283b798cacb961d38ddae8dd187feffc7c1c986620059f2209fa67c4`.

The earlier regression matrix also covered 506-token and 20,006-token prompts at widths 1/2/4/9, with 8/8 parity cells and 2,048 continuation decodes.

### Before/after performance

Each row below uses 64 measured iterations after 16 warmups. The baseline omits only patch 0043; Rust harness, model, prompt, cache types, FlashAttention mode, and hardware are identical.

| Width | Baseline batched | Patched batched | Change | Patched vs serial |
|---:|---:|---:|---:|---:|
| 9 | 173.19 tok/s, 51.968 ms | 256.82 tok/s, 35.044 ms | **+48.3%** | 3.63x |
| 16 | 255.68 tok/s, 62.578 ms | 301.42 tok/s, 53.083 ms | **+17.9%** | 4.29x |

The baseline failed exported-state and token-signal parity at both widths. The patched path passed both, so correctness did not require a performance regression.

## Gates

- clean pinned llama.cpp patch-queue apply;
- patched static Metal llama.cpp build;
- `cargo fmt --all -- --check`;
- `cargo test --locked -p skippy-bench` (`73` passed);
- `cargo clippy --locked -p skippy-bench --all-targets -- -D warnings`;
- focused Apple Metal backend sweep for modified source types: `274/274` `MUL_MAT` cases passed.

## Summary by CodeRabbit

* **New Features**
  * Added CLI options for verification widths, sample width, continuation steps, and flash-attention mode.
  * Expanded the verification JSON report with selected flash-attention mode, sampled token stream, requested widths/steps, and detailed per-width parity check diagnostics.
  * Supports running parity verification across multiple requested widths in one command.
* **Bug Fixes**
  * Improved argument validation and parity mismatch/error reporting.
  * Improved consistency of small-batch Metal matrix operations across batch sizes.
* **Tests**
  * Updated and added tests covering CLI parsing, planning/sampling behavior, and parity mismatch diagnostics.


michaelneale (approved):

danielwinterw:
Reviewed the Metal patch against the pinned llama.cpp source rather than taking the dispatch geometry on faith.

## Verified

The kernel geometry checks out against the pin:

- `kernel_mul_mv_t_t_4_impl` uses `r0 = tgpig.x*NR0` (simdgroups split along `ne00`, not rows), so F32 `r0ptg = 2` is correct, and `smem = 32*2*sizeof(float)` matches both `helper_mv_reduce_and_write<NR0=2>`'s `NW*NR0` floats and the canonical `32*sizeof(float)*nr0`.
- Q4_K/Q6_K use `first_row = (r0*NSG + sgitg)*nr0` with `N_R0_* = 2` → `2*nsg` is right; Q5_0's `mul_vec_q_n_f32_impl` uses `(tgpig.x*NSG + sgitg)*NR0` with `N_R0_Q5_0 = 4` → `N_R0_Q5_0*nsg` is right.
- `ggml_metal_library_get_pipeline_mul_mv_ext` already sets the same `FC_MUL_MV + {0,2,3,4}` function constants the canonical impls read, so `NSG`/`ne12`/`r2`/`r3` resolve correctly in the ext pipelines.
- `ggml_metal_kargs_mul_mv_ext` lacks `nr0`, but only the `_disp` wrappers read `args.nr0` — calling `_impl` directly with explicit template constants sidesteps that.
- `res.smem` is zero-initialized in `ggml_metal_library_get_pipeline`, so the `if (pipeline.smem > 0)` guard is safe for the quant pipelines, and `ggml_metal_op_mul_mat` is the only caller of the ext pipeline getter.
- Patches apply via `find … | sort`, so 0043 slots in with no manifest change needed.

## Issues

**1. Hardcoded throughput denominator is now wrong** (`crates/skippy-bench/src/verify_window_local.rs`). `build_report` still calls `verified_tokens_per_sec(batched_avg, 2)` / `(serial_avg, 2)`, and the split path does the same, but the sampled width is now `sample_width(args)` = `verify_widths[0]`. `--verify-widths 9` reports `batched_token_per_sec` 4.5x too low. The `batched_width2` / `split_inprocess_width2` fields are misnamed for the same reason. Thread the sample width through both call sites and rename.

**2. `sample_width` = first element of `--verify-widths`** is an invisible coupling: `--verify-widths 1,2,4,9` and `9,4,2,1` run the same parity sweep but produce different benchmark numbers. Either add an explicit `--sample-width` or keep the benchmark pinned at 2.

**3. `run_parity_check` bails on first mismatch**, so `matched` is always `true` and `first_mismatch_position` always `None` in every emitted report — both fields are dead as written, and a width-4 failure hides whether 9/16 also fail. Collecting all checks and failing once at the end with the full table would fix that and make the aggregation unit-testable.

**4. `signal_window_matched`** is computed, reported, and included in the bail message, but excluded from `matched`. Intentional (window formatting expected to differ) or an oversight? Worth a comment either way.

**5. Signal-window equality via `format!("{:?}", …)`** compares Debug strings — sensitive to float formatting and any Debug impl change. If the type can't derive `PartialEq`, a comment on why would help.

**6. Silent semantic change.** `choose_verify_tokens` used the native MTP n=1 draft; `choose_target_plan` now uses the model's own greedy stream. Right input for a parity harness, but it changes what `batched_width2` measures (an always-accepted stream) and drops this command's only coverage of the native MTP draft path. Worth a line in the PR body.

## Risks worth settling before merge

**Blast radius.** This isn't verifier-scoped — it changes Metal dispatch policy for every model this fork runs. Batch sizes 9–16 on F32/Q5_0/Q4_K/Q6_K previously reached `mul_mm`; they now take the row-parallel `mul_mv` path, which re-reads the weight matrix per row instead of amortizing it across the batch. Q4_K is the most common quant in the wild. The validation table compares one-call vs *serial* verification — it doesn't show the cost against the *pre-patch batched* kernels, which is the number that says whether general prefill and multi-token decode regressed. A `llama-bench` pp/tg run plus a batch-9..16 decode comparison before/after on a Q4_K model would close that. There's no `test-backend-ops` gate in CI, so nothing else catches it.

**Unenforced width ceiling.** Parity holds only for `ne11 <= 16` and only for these four types; at 17+ the op falls back to `mul_mm` and batch-invariance silently disappears. Nothing in `validate_args` or the production verify path enforces that — if a proposer ever emits >16 tokens the guarantee vanishes with no error. Add a guard, or at minimum a documented constant shared with the verifier.

**Silent drift against upstream.** The patch re-derives `nsg = std::min(4, (ne00 + 127) / 128)` and `smem = 32*2*sizeof(float)` from `ggml-metal-device.cpp`, and hardcodes `2*nsg` for Q4_K/Q6_K where `N_R0_Q5_0*nsg` is used for Q5_0. If upstream bumps `N_R0_Q4_K`, `N_SG_Q6_K`, or the F32 `nsg` heuristic, this still compiles and still runs — it just stops being batch-invariant, i.e. exactly the failure mode this PR exists to fix. `llama-upstream-canary` catches apply failures, not this. Suggest `N_R0_Q4_K*nsg` / `N_R0_Q6_K*nsg` plus a comment on each duplicated constant naming its source of truth.

## Minor

- The four `r1ptg` template instantiations per type are byte-identical (`(void) r1ptg;`), multiplying pipeline compilations by 4 for no benefit. Pinning `r1ptg = 2` for `batch_invariant_ext_type` before the pipeline lookup would cut that.
- `serial_decode_expected` silently drops negative predictions, surfacing later as a confusing "prediction has N token(s)" error instead of "decode returned no token".
- Error text `"width {width} requires {width}"` reads oddly (three sites); `verify_tokens_for_width` requires `targets.len() >= width` when it only consumes `width - 1`.
- `width.saturating_add(1) as u32` vs `i32::try_from(layer_end)` elsewhere — inconsistent conversion style.
- `choose_target_plan` generates one more token than any consumer needs (`max_width + continuation_steps + 1`); harmless, but the `+1` is unexplained.
- `validate_args` doesn't reject duplicate widths or widths above the invariance ceiling.

## Description mismatch

The body claims the PR "remove[s] the temporary serial verifier fallback and Metal kernel-disable guards", but the diff against `main` touches only 3 files and removes no such code — and I found no fallback or kernel-disable guard on `main`. Presumably branch-internal history; worth correcting so it doesn't mislead reviewers or release notes.

## Tests

CLI parsing coverage is good, and `first_mismatch_position` / `verify_tokens_for_width` are covered including the length-difference case. The parity flow itself isn't testable without a model, which is fair — but fixing (3) would make the aggregation logic unit-testable, which is where a regression is most likely to hide.

i386:
Thanks for the detailed review. I've addressed the findings in `ccf1eb145`.

- Added explicit `--sample-width`, corrected throughput calculations and report naming, and enforced the shared width-16 ceiling at both CLI parsing and runtime validation.
- Changed parity execution to collect every requested width before failing, preserved real mismatch diagnostics, and replaced Debug-string signal comparison with the typed signal-window contract. Signal-window shape remains advisory because one batched call and N serial calls have different call histories; exported state and `TokenSignal` remain exactness gates.
- Documented that the harness now uses an always-accepted greedy target stream and that native-MTP proposal quality belongs to the dedicated MTP benchmark.
- Grounded dispatch geometry in the upstream constants, collapsed the redundant `r1ptg` specializations, and restored Q5_0's required threadgroup-memory allocation.
- Added matched pre-patch versus patched batched measurements: width 9 improved from 173.19 to 256.82 tok/s (+48.3%), and width 16 improved from 255.68 to 301.42 tok/s (+17.9%).
- Physical Metal validation passed widths 1/2/4/9/16 with exact target stream, 256-token continuation, native position, full exported state, and token signals. The focused backend sweep passed 274/274 `MUL_MAT` cases.

The earlier Linux CUDA failure was runner infrastructure (`/__e/node24/bin/node` missing before checkout). That run was superseded after merging current `main`; the new CI runs are now in progress.

PR #1080, opened 2026-07-26
Expose target-authoritative local generation receipts

i386:
Embedded Skippy callers can now receive a target-authoritative receipt for each successful local generation before its runtime session is torn down.

The receipt captures:

- prompt token count and stable digest
- exact generated token IDs
- final native/Rust/tracked session position
- termination reason
- model-generation elapsed time
- an optional full-state digest for exactness checks

## Architecture

This adds an opt-in, source-neutral `GenerationReceiptSink` boundary to `EmbeddedOpenAiArgs`. Receipt delivery happens before session cleanup and after generation succeeds. Full-state export remains disabled by default and receipt configuration is rejected for split/downstream execution.

The runtime position check compares the server's tracked position, the Rust session position, and the native runtime position so callers receive a canonical target position rather than a frontend estimate.

## Protocol

No mesh wire format or Skippy ABI changes. The new surface is an in-process public Rust API and is disabled unless an embedded caller supplies a receipt configuration.

## Validation

- `cargo test -p skippy-runtime --lib` (67 passed)
- `cargo test -p skippy-server --lib` (305 passed)
- `cargo check -p skippy-runtime`
- `cargo check -p mesh-llm`

## Summary by CodeRabbit

* **New Features**
  * Added optional “generation receipts” for local text generation, including token digests, termination reason, model-generation timing, and final session position, with an optional full-state digest.
  * Introduced configurable receipt delivery via a sink and exposed receipt-related APIs and token-id digest utility.
* **Bug Fixes**
  * Receipt support is enforced for local single-stage execution only; invalid topologies are rejected.
  * Native session positions now reject negative values.
* **Tests**
  * Expanded unit/integration coverage for receipt delivery ordering, sink failure handling, digest stability, termination precedence, and native-position validation.


crates/skippy-server/src/frontend/generation/server.rs
```diff
@@ -187,6 +189,7 @@ pub struct EmbeddedOpenAiArgs {
     pub prediction_returns: Option<Arc<PredictionReturnHub>>,
     pub telemetry: Telemetry,
     pub hook_policy: Option<Arc<dyn OpenAiHookPolicy>>,
+    pub generation_receipt: Option<GenerationReceiptConfig>,
```
ndizazzo, line 192: It looks like receipts are only wired through `generate_local_tokens`, while local multimodal generation has its own loop and cleanup path. The multimodal path probably needs the same receipt handling... We end up with a backend configured with a receipt sink can successfully handle a multimodal request without ever producing a receipt.

crates/skippy-server/src/frontend/generation/server.rs
```diff
@@ -297,6 +300,9 @@ pub fn embedded_openai_backend(args: EmbeddedOpenAiArgs) -> Result<EmbeddedOpenA
     if args.native_mtp_draft_model_path.is_some() && !args.native_mtp_enabled {
         bail!("native MTP must be enabled when an MTP draft model is set");
     }
+    if args.generation_receipt.is_some() && args.config.downstream.is_some() {
+        bail!("generation receipts are supported only for local single-stage execution");
+    }
```
ndizazzo: Hardening this with a check to require both upstream and downstream to be absent would help the validation match the error message

crates/skippy-server/src/frontend/local_generation.rs
```diff
@@ -806,4 +933,165 @@ mod tests {
         assert!(source.contains(concat!(".decode", "_sampled_mtp(")));
         assert!(!source.contains(concat!(".decode_frame", "_sampled_mtp(")));
     }
+
+    #[test]
+    fn local_generation_delivers_receipt_before_cleanup_and_propagates_sink_errors() -> Result<()> {
```
ndizazzo, line 944: This test returns early unless the two model environment variables are set. So the standard 305-test run doesn’t actually exercise receipt delivery, cleanup ordering, position validation, or sink errors. 

A model-free test seam for that lifecycle logic would fix it while keeping this as an optional real-model smoke test

crates/skippy-server/src/frontend/generation_receipt.rs
```diff
@@ -0,0 +1,309 @@
...
+use anyhow::Result;
+use openai_frontend::{OpenAiError, OpenAiResult};
+
+use crate::frontend::{StageOpenAiBackend, openai_backend_error};
+
+const TOKEN_ID_DIGEST_DOMAIN: &[u8] = b"skippy-generation-token-ids-v1\0";
+
+/// Why a successful local generation stopped.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum GenerationTermination {
```
ndizazzo, line 15: Since these are new public receipt types, it might be worth marking the receipt structs and termination enum as `#[non_exhaustive]`... Receipts will probably gain fields or termination reasons over time, and doing this now would make those additions less disruptive for downstream callers.

ndizazzo (approved):
Only minor minor stuff 👍🏻

crates/skippy-server/src/frontend/generation/server.rs
ndizazzo: Addressed in c34347c28. Local multimodal generation now observes emitted token IDs, records cancellation/callback termination and timing, delivers the receipt before the shared session cleanup, and propagates sink failures after cleanup.

crates/skippy-server/src/frontend/generation/server.rs
ndizazzo: Addressed in c34347c28. Receipt configuration now rejects any topology with either an upstream or downstream stage, with model-free coverage for all combinations.

crates/skippy-server/src/frontend/local_generation.rs
ndizazzo: Addressed in c34347c28. Standard tests now use a fake receipt runtime and sink to cover canonical-position failures, full-state capture, sink errors, and receipt-before-cleanup ordering; the optional real-model smoke remains in place.

crates/skippy-server/src/frontend/generation_receipt.rs
ndizazzo: Addressed in c34347c28. GenerationTermination, GenerationStateDigest, and GenerationReceipt are now non-exhaustive.

PR #1081, opened 2026-07-26
Accept bounded external token proposals in local Skippy generation

i386:
Embedded Skippy consumers can now attach any in-process source of causal, width-one token proposals and have the target runtime verify, commit, repair, and report each decision authoritatively.

The source receives only committed token history plus a strict deadline and bounded proposal budget. Skippy owns acceptance classification, emits the exact target-authoritative committed prefix, trims branch-conditioned suffix rows after a mismatch or callback stop, and reports timing and position evidence back against the source's opaque decision ID.

## Architecture

- Adds the source-neutral `LinearProposalIngress` trait and bounded configuration.
- Admits proposals only for greedy local generation without grammar, generation hooks, or native MTP.
- Rejects late, empty, oversized, and invalid-token proposals before verification.
- Uses the native verify-window classifier as the sole acceptance authority.
- Repairs runtime position before surfacing callback failures and retires a session if repair fails.
- Leaves existing generation unchanged unless an ingress source is configured.

## Protocol

This is an additive in-process API. It does not change the mesh wire protocol or the Skippy ABI.

## Dependency

The diff is independent of #1078, but canonical Metal behavior for its batched target-verification call depends on #1078. This draft should merge after #1078.

## Validation

- `cargo test -p skippy-server --lib` (309 passed)
- `cargo check -p mesh-llm`

## Summary by CodeRabbit

* **New Features**
  * Added optional “linear proposal” ingress support for embedded OpenAI generation, including configurable deadlines and max token limits.
  * Added end-to-end proposal flow: query, local execution, disposition/receipt reporting, and terminal discard handling.
  * Exposed linear proposal types for consumption via public crate exports.
* **Bug Fixes**
  * Improved token verification and per-session token count tracking used during proposal processing.
* **Tests**
  * Added unit coverage for decision-id bounds, ingress config validation, query/discard behavior, and execution sequencing.


crates/skippy-server/src/frontend/local_generation.rs
```diff
@@ -540,6 +561,123 @@ impl StageOpenAiBackend {
...
+                                        base_position,
+                                        generated_len: decoded_tokens,
+                                        max_new_tokens: request.max_tokens as usize,
+                                    },
+                                    queried,
+                                    &mut on_token,
+                                )
+                            },
+                        )?;
+                        if let Some(error) = report_linear_proposal_receipt(config, &receipt) {
```
ndizazzo: We may end up leaking info here in telemetry through the error chain and receipt. This would be a defensive fix, but picking apart for relevant fields and making sure we don't leak with some telemetry inventory/privacy tests could help tighten it up

crates/skippy-server/src/frontend/linear_proposal.rs
```diff
@@ -0,0 +1,1112 @@
...
+    decode_step: usize,
+    committed_token_ids: &[i32],
+    remaining_new_tokens: usize,
+) -> OpenAiResult<LinearProposalQueryOutcome> {
+    let max_proposal_tokens = remaining_new_tokens
+        .saturating_sub(1)
+        .min(config.max_proposal_tokens());
+    if max_proposal_tokens == 0 {
+        return Ok(LinearProposalQueryOutcome::NoProposal);
+    }
```
ndizazzo, line 344: Doesn't this ignore our runtime `n_batch` setting? We should cap to something like `session_batch_size - 1` so we take it into account

ndizazzo (commented):
@i386 we have a couple runtime /telemetry fixes on this one

michaelneale:
Reviewed this — overall low risk to `main`, because the feature is inert by default: all construction sites pass `linear_proposal_ingress: None`, nothing in-tree sets it, and the gate is computed once outside the decode loop. The "leaves existing generation unchanged" claim holds, and I confirmed no `proto/`, gossip, or `skippy-ffi` ABI surface is touched.

A few things I'd want addressed before this stops being inert:

**1. The risky code is exactly the untested code.** The 10 new tests all cover pure helpers via a fake ingress. `execute_local_linear_proposal_inner`, `trim_branch_suffix_or_retire`, and `RuntimeState::verify_tokens` — everything that mutates real KV state — have no coverage. Would like an integration test driving a real session through full-accept, first-mismatch, and callback-stop, asserting `session_token_count` afterward.

**2. Position mismatch fails the request instead of falling back.** Each guard returns `OpenAiError::backend`, so a mismatch (KV eviction, sliding window, concurrent trim) kills the request mid-stream. Since this path inherently races proactive eviction, falling back to serial decode seems strictly more robust than 500ing. This is the design call I'd push back on most.

**3. `base_position` rests on an implicit prefill invariant.** `prompt_token_ids.len().saturating_sub(1) + decoded_tokens` is only correct because the two prefill modes differ in `decoded_tokens`. Both cases do land right, and the `session_token_count` check means it fails loudly rather than corrupting KV — but it deserves a comment, since it'll break quietly if prefill changes.

**4. Tokens stream before the KV repair.** `on_token()` fires for all committed tokens, then `trim_branch_suffix_or_retire()` runs. If the trim fails the session is retired and an error returned after the client already received tokens. The tokens are target-authoritative so the stream isn't wrong, but the client sees a partial response plus an error.

**5. The deadline is unenforceable.** `propose()` runs synchronously on the decode thread with no preemption, so a hung source hangs the request; lateness is only detected after it returns. Honestly documented, but no config bound can fix it.

**6. `EmbeddedOpenAiArgs` is `pub` with `pub` fields and no `#[non_exhaustive]`.** Adding a field is a semver break for embedders. No in-tree SDK/FFI consumers today, so low impact, but this struct will keep growing — worth `#[non_exhaustive]` or a builder now.

**7. Merge ordering.** Please confirm this lands after #1078 — out of order risks wrong Metal results for the batched verification call, and per (1) no test would catch it.

Minor: `linear_proposal_disposition`'s condition is hard to read and `Stopped` absorbs several distinct outcomes; `accepted_proposal_tokens.min(committed_tokens.len())` can understate acceptance on early stop (telemetry-only); two raw `eprintln!`s where the repo has a telemetry/`OutputEvent` contract and stderr is TUI territory; `runtime_state.rs` was already past 1k lines, so per the repo rule this change arguably wanted an extraction.

Merging as-is is defensible *because* it's inert, but I'd treat it as unvalidated until something in-tree constructs an ingress — at which point 1–3 become blocking.

crates/skippy-server/src/frontend/local_generation.rs
ndizazzo: Addressed in 63b668177. Proposal telemetry now excludes opaque decision IDs, token values, request/cache identifiers, and formatted error chains. Report failures use only a bounded `failed` outcome, and the focused privacy regression test covers a path-like secret plus token values.

crates/skippy-server/src/frontend/linear_proposal.rs
ndizazzo: Addressed in 63b668177. The ingress limit is now capped by the active runtime session batch size minus the current-token row, with a regression test proving the configured proposal bound cannot exceed that live window.

ndizazzo:
Follow-up status after stacking this PR on #1080 (`63b668177`):

- caps proposal width to the active runtime `n_batch - 1` window;
- falls back to serial decode when the session position drifts before verification;
- documents the prefill/base-position invariant and advisory deadline;
- keeps receipt-report failures non-fatal after tokens commit and makes proposal telemetry privacy-safe/bounded;
- keeps branch repair in the named `trim_branch_suffix_or_retire` path;
- marks the new public proposal contracts non-exhaustive.

Merge ordering is now explicit: #1078 is already on `main`, and #1081 is based on #1080. Local validation passed `skippy-server`'s 327 tests, `mesh-llm-host-runtime`'s 1,710 tests (8 ignored), warning-denying Clippy for both `skippy-server` and `mesh-llm`, and workspace formatting.

The feature remains inert by default. A fixture-backed real-model test covering full accept / first mismatch / callback stop is still best treated as an activation prerequisite because the repository does not carry a hermetic model fixture; the existing optional native-runtime smoke fixtures cannot make that coverage run in normal CI. Likewise, changing the already-public `EmbeddedOpenAiArgs` construction contract needs a dedicated builder migration rather than making current cross-crate struct literals uncompilable in this stack.

PR #1082, opened 2026-07-26
feat(runtime): add daemon model lifecycle reconciliation

### LINKED ISSUE #910: spec: Runtime daemon and model lifecycle
# Runtime Daemon And Model Lifecycle

Status: proposal

## Summary

`mesh-llm serve` should start a long-lived daemon first and treat local model processes as runtime-managed resources. Starting the daemon must not require loading a model.

The daemon owns:

- mesh membership, gossip, routing, and discovery;
- the OpenAI-compatible ingress on the API port;
- the management API and owner-control surfaces;
- plugin lifecycle and plugin-provided inference endpoints;
- local model process lifecycle: start, stop, drain, replace, and report.

Local llama.cpp/Skippy model serving becomes one capability the daemon can activate after startup. External OpenAI-compatible endpoints remain plugin capabilities that can be active with zero local model processes.

## Motivation

The current user-facing runtime model is still startup-model-centric:

- `mesh-llm serve --model X` starts the node and immediately launches `X`;
- bare `mesh-llm serve` is treated as incomplete when no startup model is configured;
- external endpoint users perceive that they need a placeholder local model even though plugin inference routing already exists;
- runtime load/unload exists, but it feels secondary to startup-time launch.

This makes the wrong thing feel fundamental. The daemon is the fundamental runtime. Models are workload allocations controlled by policy, user intent, and local capacity.

This difficulty surfaced in [GitHub issue #905](https://github.com/Mesh-LLM/mesh-llm/issues/905), where a user asked for "alternative endpoint flexibility" because the current CLI shape made it appear that `mesh-llm serve` must launch a local placeholder model even when the desired behavior was only to expose the mesh-llm API and proxy to an OpenAI-compatible provider. The existing plugin inference path already supports external endpoints, so the cited issue is evidence of a deeper runtime-model problem: daemon startup is not clearly separated from local model process startup.

## How This Fixes Issue #905

[GitHub issue #905](https://github.com/Mesh-LLM/mesh-llm/issues/905) asks for a way to run mesh-llm against external OpenAI-compatible providers without starting an unnecessary local llama.cpp model process. The user proposed new backend routing, `--reroute`, runtime modes, and backend management commands.

This proposal fixes the core problem by changing the runtime model:

1. `mesh-llm serve` starts the daemon even when there are zero startup models.
2. The daemon still binds the OpenAI-compatible API, console, mesh, plugins, and owner-control surfaces.
3. The existing `openai-endpoint` plugin can register external provider models.
4. `/v1/models` can list those external models even when no local model is loaded.
5. Requests for those model IDs route through the existing plugin inference endpoint path.
6. Local model loading remains available later through runtime load commands.

That means the issue's desired "alternative endpoint only" workflow becomes a normal idle/proxying daemon state:

```bash
mesh-llm plugins install openai-endpoint

# ~/.mesh-llm/config.toml
[runtime]
mode = "on_demand"

[[plugin]]
name = "openai-endpoint"

mesh-llm serve
```

No placeholder model is needed, and no new public backend registry is needed. The daemon can sit ready, proxy to the external provider, and later load or unload local models through the same runtime lifecycle controls.

The issue's proposed features then separate into clearer categories:

| Issue #905 request | Resolution in this proposal |
|---|---|
| Skip local model launch for external endpoints | Fixed by `mode = "on_demand"` and daemon startup with zero local model processes. |
| Route to vLLM, TGI, Ollama, SGLang, etc. | Use the existing plugin inference endpoint path, starting with `openai-endpoint`. |
| Add `--reroute <URL>` | no longer required for the core behavior. |
| Add `local` / `external` / `hybrid` modes | Represented naturally by daemon state plus local model/plugin endpoint inventory. |
| Add backend registry in the router | Not needed; plugin inference endpoints already provide routeable external models. |
| Add model target reconciliation for external endpoints | Not needed for proxying; reconciliation should manage local model process intents. |
| Add remote start/stop behavior | Belongs on the private owner-control plane as later owned-node lifecycle commands. |

So #905 should be addressed as a daemon lifecycle and documentation problem, not as a new routing architecture.

## Relationship To Future Implementation for Owned-Node Commands

This spec is compatible with the future plan `.omo/plans/owned-node-commands.md`. It deliberately keeps the same control-plane boundaries:

- owned-node imperative commands use the private owner-control plane;
- owner-control traffic uses the owner-gated control connection, not the public mesh request/gossip plane;
- command results are requester-local unless they intentionally mutate durable node state;
- public mesh `/0` and `/1` compatibility remains unchanged;
- no new public mesh stream should be added for runtime lifecycle commands;
- no raw command result should be promoted into `/api/status`, gossip, or peer-state truth.

The owned-node plan names `scan_refresh` as the first exemplar command. This runtime lifecycle spec does not replace that plan. It defines the broader model that later owner-control commands such as `load_model`, `unload_model`, and `ensure_model` can use after `scan_refresh` proves the transport and typed single-target command shape.

## Goals

- Allow `mesh-llm serve` to start with zero local models.
- Keep API, console, owner-control, plugin endpoints, and mesh participation available before any local model process exists.
- Make local model start/stop a normal runtime operation, not a startup-only side effect.
- Preserve startup models as convenience intents, not as daemon preconditions.
- Use `[runtime] mode` as the primary persisted runtime behavior switch.
- Let users command owned nodes to start/stop models through local API/CLI or private owner-control.
- Let mesh demand and wanted-model signals influence local policy without giving arbitrary peers imperative control over this node.
- Keep external OpenAI-compatible providers on the existing plugin inference path.

## Non-Goals

- No public mesh command stream for start/stop model commands.
- No peer-to-peer imperative command authority from arbitrary mesh members.
- No requirement that a daemon download or launch a model merely because it starts.
- No replacement of the plugin inference endpoint architecture.
- No status-history store for command responses.
- No multi-target remote fanout in the first lifecycle command phase.

### Daemon

The daemon is the durable process started by `mesh-llm serve`.

It may be in one of these operating states:

| State | Meaning |
|---|---|
| `starting` | Process is booting, loading config, binding listeners, and joining or creating a mesh. |
| `ready_idle` | API, console, mesh, plugins, and control surfaces are ready; no local model process is serving. |
| `ready_serving` | One or more local model processes are serving. |
| `ready_proxying` | No local model process is serving, but plugin or remote mesh models are routeable. |
| `degraded` | The daemon is up, but one or more subsystems failed, such as plugin startup or model launch. |
| `stopping` | Shutdown or restart has begun. |

These are daemon states, not model states. A daemon can be both proxying and serving if it has plugin endpoints and local models.

### Model Process

Each local model process has its own lifecycle:

| State | Meaning |
|---|---|
| `planned` | A command or policy selected the model, but launch has not started. |
| `resolving` | Model reference, package, device, and runtime settings are being resolved. |
| `loading` | Runtime process is starting and loading weights. |
| `warming` | Runtime is live but not yet routeable or fully probed. |
| `serving` | Model is routeable through the daemon. |
| `draining` | New requests are no longer assigned; in-flight requests may finish. |
| `unloading` | Runtime process is stopping and capacity is being released. |
| `failed` | Launch or runtime health failed. |
| `stopped` | No process remains. |

The state machine must be instance-aware. Multiple instances with the same served model name require instance IDs for unambiguous unload and replacement.

## Startup Semantics

`mesh-llm serve` should always attempt to start the daemon.

Startup model config becomes an optional set of model intents whose behavior is controlled by `[runtime] mode`:

```toml
[runtime]
mode = "serve"

[[models]]
model = "Qwen/Qwen3-Coder-Next-GGUF@main:Q4_K_M"
```

Startup behavior:

1. Start daemon services: config, identity, mesh, API, console, plugins.
2. Publish local inventory and plugin inference models when available.
3. Resolve the effective runtime mode from CLI and config.
4. Convert configured `[[models]]` and CLI model flags into desired runtime intents according to that mode.
5. Reconcile eager intents through the same runtime control path used after startup.
6. If there are no eager startup intents, enter `ready_idle` or `ready_proxying` instead of exiting.

The old warning that `serve` needs at least one startup model should become an informational idle-start message, for example:

```text
mesh-llm daemon ready with no local models loaded.
Use `mesh-llm load <model>` or configure [[models]] for eager startup.
```

## Runtime Modes

Persisted runtime mode lives under `[runtime]`:

```toml
[runtime]
mode = "on_demand" # client | serve | on_demand
```

The mode controls two things at once because that matches the user-facing mental model: whether the node is worker-capable, and whether configured models are eagerly loaded at daemon startup.

| Mode | Node role | Startup model behavior | Later local model loads |
|---|---|---|---|
| `client` | Client/router only | Do not load configured `[[models]]`; do not advertise worker capacity. | Reject local model load commands unless the operator switches mode or starts a worker-capable daemon. |
| `serve` | Worker-capable | Eagerly load configured `[[models]]` and explicit CLI startup models. | Allow local and owner-authorized load/unload commands. |
| `on_demand` | Worker-capable | Start idle/proxying; treat configured `[[models]]` as allowed/preferred model intents, not eager launches. | Allow explicit load/unload commands and policy-driven reconciliation. |

This replaces the idea of a boolean `runtime.client` setting. `client` is a full runtime mode, and `on_demand` is the missing "ready to serve later" mode.

### Mode Resolution

CLI surfaces should remain explicit operator intent and take precedence over persisted config:

- `mesh-llm client ...` sets effective mode `client`.
- `mesh-llm serve ...` uses `[runtime].mode` when present, defaulting to `serve` for compatibility.
- `mesh-llm serve --model X` and `mesh-llm serve --gguf X` are explicit eager startup requests and should either force effective mode `serve` for that invocation or fail fast if combined with `mode = "client"`.
- A future `mesh-llm serve --mode on_demand` flag may override TOML for one invocation, but the durable source should be `[runtime].mode`.

Default rollout should be conservative:

1. Add `mode = "serve"` as the compatibility default.
2. Allow users to opt into `mode = "on_demand"`.
3. After daemon-first behavior is proven, consider whether new installs should default to `on_demand`.

### Local User Control

Local user control remains available through the loopback management API and CLI wrappers.

Existing local concepts map cleanly into the daemon model:

- `POST /api/runtime/models` sends `RuntimeControlRequest::Load`;
- `DELETE /api/runtime/models/{model}` sends `RuntimeControlRequest::Unload`;
- `DELETE /api/runtime/instances/{instance_id}` unloads a specific instance;
- `mesh-llm load` and `mesh-llm unload` are CLI facades over the same local runtime control path.

These should become the canonical local model lifecycle controls.

### Owner Remote Control

Remote owner control must use the private owner-control plane. It should not use public mesh streams, public gossip, or Nostr metadata as command transport.

Future model lifecycle owner commands should follow the owned-node command pattern:

- single target in v1;
- typed protobuf request and response;
- direct owner-attestation authorization;
- requester-local result envelope;
- no persistence of raw command result into public status;
- self-target fast path that returns the same typed result as remote execution.

Candidate owner-control commands:

| Command | Purpose |
|---|---|
| `scan_refresh` | Refresh target-local model inventory and republish available models. |
| `load_model` | Imperatively load one model on one owned node. |
| `unload_model` | Imperatively unload one model or instance on one owned node. |
| `ensure_model` | Ensure one model is serving on one owned node, idempotently. |
| `drain_model` | Stop routing new local requests before unload. |

Only `scan_refresh` belongs in the owned-node command v1 described by the existing plan. The model commands are follow-on commands after the private plane and typed command result contract are proven.

### Mesh Demand And Wanted Models

Mesh demand should remain declarative. A peer can help the mesh express that a model is wanted, but a peer must not directly command another node to start a model.

Inputs to local policy may include:

- explicit local model interests from API/UI/operator input;
- explicit model interests gossiped by peers;
- OpenAI request demand observed by this node;
- route pressure, no-host failures, and hot model counters;
- configured startup model intents;
- owner-issued remote commands;
- local inventory and capacity advice.

The local daemon decides whether to load, unload, replace, or ignore based on its own policy and authorization state.

This preserves the important trust boundary:

- owner commands are imperative and private;
- mesh wanted-model signals are public/declarative and advisory;
- local policy turns advisory signals into local runtime control actions.

## Desired Runtime Intents

The daemon should represent model lifecycle requests as desired runtime intents before they become process actions.

Suggested shape:

```text
DesiredRuntimeIntent {
  intent_id,
  model_ref,
  source,
  priority,
  persistence,
  target_scope,
  constraints,
  created_at,
  expires_at,
}
```

`source` examples:

- `startup_config`
- `local_user`
- `owner_control`
- `mesh_interest`
- `mesh_demand`
- `runtime_reconciliation`

`persistence` examples:

- `ephemeral`: action request only; do not write config;
- `session`: keep until daemon exit or explicit release;
- `configured`: backed by `config.toml`;
- `policy`: maintained while policy inputs remain true.

This lets startup config, local load commands, remote owner commands, and mesh demand all feed one reconciliation path without pretending they have the same authority.

## Reconciliation Policy

Runtime reconciliation should become the single place that translates desired intents into local model process actions.

Policy responsibilities:

- validate model references and local inventory;
- determine if the node is eligible to serve the model;
- plan local runtime settings and capacity reservation;
- avoid duplicate in-flight loads;
- honor manual unload cooldowns;
- respect owner-issued explicit commands over advisory mesh demand;
- decide whether replacement is allowed;
- emit clear status events for pending, started, completed, failed, and skipped actions.

The existing model-target reconciliation logic is a good starting point, but it should evolve from "load wanted local model refs" into a general local reconciler over typed desired intents.

## External Endpoint Behavior

No special proxy mode is required for the core capability.

This is the direct answer to the confusion in [GitHub issue #905](https://github.com/Mesh-LLM/mesh-llm/issues/905): external providers should not require placeholder local model loads or a parallel router backend registry. They should work because the daemon can be ready with zero local model processes while plugin inference endpoints supply routeable models.

When the daemon starts with an `openai-endpoint` plugin configured:

1. The daemon starts with zero local model processes if no local intents exist.
2. The plugin reports inference endpoints and model IDs.
3. `/v1/models` includes plugin models.
4. Requests for plugin model IDs route to the plugin endpoint.
5. Local model lifecycle remains available independently.

This is a natural consequence of daemon-first startup. External providers stop being a special case.

## Public Status And Runtime Data

Status surfaces should summarize durable state, not command transcripts.

Allowed status data:

- daemon readiness and health;
- current local model instances and states;
- plugin endpoint summaries;
- local inventory summaries;
- active desired intents, if they are durable or policy-visible;
- last error per model instance or intent, with bounded detail;
- capacity and placement advice.

Disallowed status data:

- raw owner-control command request bodies;
- raw command response payloads;
- requester-local remote command transcripts;
- private owner-control handshake material.

Detailed command responses belong to the requester. The daemon may emit local events and bounded summaries for observability, but public mesh gossip and `/api/status` must not become command-history storage.

### Public Mesh Plane

The public mesh plane remains responsible for:

- mixed-version mesh compatibility;
- gossip, route tables, peer liveness, and request routing;
- model interests and demand signals;
- plugin traffic already assigned to the public mesh protocol;
- existing compatibility behavior for `mesh-llm/1` and `mesh-llm/0`.

It must not carry imperative model lifecycle commands.

### Private Owner-Control Plane

The private owner-control plane is responsible for:

- owner-gated single-target commands;
- config get/watch/apply;
- inventory refresh;
- future model lifecycle commands for owned nodes.

Every future model lifecycle owner command must be additive to the private contract and must preserve unsupported-peer behavior for older nodes.

## API And CLI Shape

Local lifecycle commands:

```bash
mesh-llm serve
mesh-llm load Qwen/Qwen3-Coder-Next-GGUF@main:Q4_K_M
mesh-llm unload Qwen3-Coder-Next
```

Remote owner lifecycle commands, after the owned-node private plane supports them:

```bash
mesh-llm owner-command scan-refresh --node <endpoint-id>
mesh-llm owner-command load-model --node <endpoint-id> --model <model-ref>
mesh-llm owner-command unload-model --node <endpoint-id> --model <model-or-instance>
```

The local API may expose equivalent endpoints, but they should call the same runtime control and owner-control client helpers rather than inventing a second execution path.

## Configuration Model

Configuration should distinguish runtime mode from model definitions.

Recommended semantics:

- `[runtime].mode` controls whether the daemon is client-only, eager-serving, or worker-capable on demand;
- `[[models]]` defines configured model preferences and constraints, not daemon existence;
- in `mode = "serve"`, configured `[[models]]` are eager startup intents;
- in `mode = "on_demand"`, configured `[[models]]` are allowed/preferred intents for explicit commands and policy reconciliation;
- in `mode = "client"`, configured `[[models]]` are not loaded locally;
- plugin config can make the daemon useful with zero local models;
- reconciliation settings control whether policy may auto-load models from wanted/demand signals;
- explicit local or owner commands can remain ephemeral unless the caller asks to persist them.

Potential future config:

```toml
[runtime]
mode = "on_demand"
reconcile_model_targets = true
reconcile_model_target_demand_upgrades = true

[[models]]
model = "Qwen/Qwen3-Coder-Next-GGUF@main:Q4_K_M"
```

The initial compatibility default should be `mode = "serve"`. Operators who want a daemon that starts idle but remains worker-capable should set `mode = "on_demand"`.

## Failure Semantics

Starting the daemon with zero models is success.

Failures should be scoped:

- plugin failure degrades plugin endpoints, not daemon startup unless configured as required;
- model load failure marks that intent or instance as failed, not daemon startup, unless the user explicitly requested fail-fast eager startup;
- unavailable model requests return OpenAI-style errors from the API surface;
- old peers that do not support private owner-control lifecycle commands return typed `unsupported`;
- unauthorized remote commands return typed authorization errors without mutating target state.

### Phase 1: Daemon-First Local Runtime

- Allow `mesh-llm serve` with zero startup models.
- Remove the help/exit gate for no startup models.
- Add `[runtime].mode = "client" | "serve" | "on_demand"` with compatibility default `serve`.
- Start API, console, mesh, plugins, and owner-control in idle mode.
- Keep existing `RuntimeControlRequest::Load` and `Unload` as the local lifecycle path.
- Update docs and CLI help to describe daemon-first behavior.

### Phase 2: Intent-Aware Reconciliation

- Introduce an internal desired-intent model.
- Convert startup models into initial intents.
- Convert explicit model interests and demand into advisory intents.
- Extend model-target reconciliation to consume typed intents.
- Preserve manual unload cooldowns and failure cooldowns.

### Phase 3: Owner-Control Lifecycle Commands

- Keep `scan_refresh` as the first command from the owned-node plan.
- Add typed single-target `load_model` and `unload_model` commands after the private plane is proven.
- Use the same local runtime control channel as local API/CLI load/unload.
- Return requester-local typed results.

### Phase 4: Mesh-Policy Autonomy

- Let local policy satisfy wanted models from mesh demand when configured.
- Add placement and replacement policy that accounts for capacity, ownership, trust, and local operator preferences.
- Keep mesh-originated demand advisory unless a future explicit authorization model is designed.

## Acceptance Criteria

- `mesh-llm serve` starts and binds API/console with no `[[models]]`, `--model`, or `--gguf`.
- `[runtime] mode = "on_demand"` starts a worker-capable daemon without eager local model launch.
- `[runtime] mode = "serve"` eagerly loads configured startup models for compatibility.
- `[runtime] mode = "client"` starts a client/router node that does not advertise local worker capacity or load local models.
- `/v1/models` returns plugin/remote models when available and an empty list otherwise.
- `mesh-llm load` can load a model after daemon startup.
- `mesh-llm unload` can stop a model without stopping the daemon.
- A configured external endpoint plugin works with zero local model processes.
- Model load failure does not kill the daemon unless explicitly configured as fail-fast.
- Remote owner commands use only the private owner-control plane.
- Mesh demand can trigger local policy only through local reconciliation, not arbitrary peer commands.
- Public mesh compatibility tests continue to pass.

## Open Questions

- Should configured startup models default to fail-fast or best-effort once the daemon-first model ships?
- Should model intents be visible in `/api/status`, a separate `/api/runtime/intents` endpoint, or both?
- What is the minimum command set after `scan_refresh`: imperative `load_model`/`unload_model`, idempotent `ensure_model`, or both?
- Should owner-control model lifecycle commands support a persistent "write this into config" option, or should persistence remain a separate config apply operation?
- How should drain timeouts be represented for long-running streaming requests?

ndizazzo:
## Summary

- start runtime surfaces independently of eager model loading, including zero-model daemon operation and explicit startup failure policy
- unify startup, local, owner-control, and advisory model requests behind prioritized desired intents with per-instance drain/inflight state
- add authenticated load, unload, ensure, and drain owner-control contracts, runtime activity policy, privacy-safe status, CLI/API surfaces, and optional load/ensure profiles
- add lifecycle and mixed-version QA harnesses covering current/released interoperability and legacy typed unsupported behavior

## Audit fixes

- require canonical-only targets for load/ensure and exactly one canonical model or instance target for unload/drain
- echo and validate response target identity end-to-end
- preserve client timeout margin over the server unary deadline
- harden daemon startup cleanup, fail-fast/best-effort behavior, coarse activity gossip, and public/private surface separation

## Validation

- `cargo test -p mesh-llm-host-runtime --lib` — 1810 passed, 8 ignored
- `cargo test -p mesh-llm-client -p mesh-llm-protocol -p mesh-llm --tests` — passed
- strict all-target clippy for touched Rust crates — passed
- `cargo fmt --all -- --check` and full branch `git diff --check` — passed
- `just build` — passed; clean commit binary built and codesigned
- runtime daemon lifecycle harness — 18 PASS, 0 FAIL, 0 PREREQ
- current/released mixed-version harness — 35 PASS, 0 FAIL, 0 PREREQ
- UI suite — 963 passed, 3 skipped

Evidence roots:

- `/tmp/mesh-llm-lifecycle-rebased/runtime-daemon-lifecycle-20260726T044251Z-23316`
- `/tmp/mesh-llm-mixed-rebased/control-plane-mixed-version-20260726T044342Z-23957`

## Summary by CodeRabbit

* **New Features**
  * Added owner-attested remote model lifecycle controls: `load-model`, `unload-model`, `ensure-model`, and `drain-model` via CLI and runtime REST endpoints.
  * Introduced runtime mode/daemon lifecycle settings (including drain timeouts) plus host-activity-based inference admission with manual override APIs.
  * Extended `/api/status`, runtime intents, and mesh gossip with coarse inference admission state.

* **Bug Fixes**
  * Improved lifecycle request/response validation and legacy unsupported handling.
  * Enforced activity admission gating for relevant routing paths; aligned missing/unavailable routing with consistent 404/503 behavior.

* **Chores**
  * Strengthened CI coverage gating, CUDA/CUDA-toolkit build configuration, and SDK readiness polling; expanded related tests and docs.


ndizazzo:
## Native CUDA and daemon lifecycle certification

**Verdict: PASS** on `carrack`, `mesh1`, and `mesh2` at commit
`cb49f3e822834b0e338384a0ac84827c4367ab91`.

### Test matrix

| Host | Native platform | CUDA target | Native release build | Runtime certification | Real-model lifecycle |
|---|---|---:|---:|---:|---:|
| `carrack` | CachyOS Linux x86_64, i7-13700K (24 CPUs), RTX 5090 + RTX 3080, CUDA 13.3 | SM 120 + SM 86 | PASS | 18/18 PASS | PASS |
| `mesh1` | Ubuntu/L4T R39.2 aarch64, AGX Orin (12 CPUs), CUDA 13.2 | SM 87 | PASS | 18/18 PASS | PASS |
| `mesh2` | Ubuntu/L4T R39.2 aarch64, AGX Orin (12 CPUs), CUDA 13.2 | SM 87 | PASS | 18/18 PASS | PASS |

All three remote checkouts were pulled to the exact PR head in
`~/dev/mesh/mesh-llm`, and all were clean after testing.

### Native build results

The definitive builds used the platform's installed compiler and native GPU
targets:

```text
carrack: MESH_LLM_BUILD_PROFILE=release scripts/build-linux.sh --backend cuda --cuda-arch '86;120'
mesh1:   MESH_LLM_BUILD_PROFILE=release scripts/build-linux.sh --backend cuda --cuda-arch 87
mesh2:   MESH_LLM_BUILD_PROFILE=release scripts/build-linux.sh --backend cuda --cuda-arch 87
```

| Host | Result | Wall time | Rust release phase | Minimum MemAvailable | Peak compiler RSS |
| `carrack` | PASS | 582 s | 4m 33s | 42.16 GiB | 5.11 GiB |
| `mesh1` | PASS | 2,283 s | 16m 44s | 54.88 GiB | 4.44 GiB |
| `mesh2` | PASS | 2,343 s | 17m 29s | 54.97 GiB | 4.18 GiB |

Build artifacts:

- `carrack`: `.sisyphus/evidence/pr1082-final-native-20260726T174856Z-2434431`
- `mesh1`: `.sisyphus/evidence/pr1082-final-native-20260726T174856Z-45284`
- `mesh2`: `.sisyphus/evidence/pr1082-final-native-20260726T174856Z-21500`

### Failures found and fixes applied

The initial Orin builds found that `/usr/local/bin/nvcc` was a symlink into
CUDA 13.2. CMake canonicalized it incorrectly to `/usr/local`, then failed to
find `cuda_runtime.h`; the actual toolkit root was
`/usr/local/cuda-13.2/targets/sbsa-linux`. After fixing toolkit discovery, the
GPU bench build still invoked bare `nvcc`, and the x86_64 build could not link
`cudart`, `cublas`, and `cublasLt`.

Two fixes were implemented, tested, and pushed:

- `97408a08e` — canonical symlink-aware CUDA toolkit discovery and explicit
  CMake compiler/toolkit selection.
- `cb49f3e82` — canonical CUDA environment and library propagation, including
  `NVCC`/`CUDACXX` support in the GPU bench build.

Validation for those fixes:

- `scripts/tests/test_cuda_toolkit.py`: 2/2 PASS
- shell syntax checks: PASS
- `git diff --check`: PASS

### Daemon lifecycle certification

The repository's `scripts/qa-runtime-daemon-lifecycle.sh` suite passed 18/18
checks on every host, with 0 failures, 0 prerequisites, and no leaked
processes. Coverage included:

- zero-model startup and management readiness;
- serve and on-demand runtime modes;
- best-effort and fail-fast startup behavior;
- authenticated load, unload, ensure, and drain intent transitions;
- activity override round-trip;
- status/intent privacy checks;
- clean process teardown.

Evidence:

- `carrack`: `.sisyphus/evidence/pr1082-final-harness/runtime-daemon-lifecycle-20260726T180542Z-2442091`
- `mesh1`: `.sisyphus/evidence/pr1082-final-harness/runtime-daemon-lifecycle-20260726T182842Z-62461`
- `mesh2`: `.sisyphus/evidence/pr1082-final-harness/runtime-daemon-lifecycle-20260726T182842Z-60580`

### Real-model load, pause, resume, and unload

Each host ran an isolated daemon with activity policy enabled
(`idle_after=30s`, `poll=1s`, `response=pause_all`). The test loaded a real
GGUF, made a baseline completion, started CPU workers equal to 25% of logical
CPUs, verified automatic pause and request rejection, verified that management
remained responsive, stopped the workload, verified automatic resume and
another completion, unloaded the model, and confirmed that the daemon stayed
alive.

| Metric | `carrack` | `mesh1` | `mesh2` |
|---|---:|---:|---:|
| Model | Qwen2.5 3B Q4_K_M | Qwen3 0.6B Q4_K_M | Qwen3 0.6B Q4_K_M |
| Daemon startup | 12 s | 8 s | 8 s |
| Model load | 11 s | 20 s | 31 s |
| Baseline inference | 200 / 102.6 ms | 200 / 560.8 ms | 200 / 547.8 ms |
| Baseline decode | 64.03 tok/s | 41.44 tok/s | 40.71 tok/s |
| Workload | 6/24 CPU workers | 3/12 CPU workers | 3/12 CPU workers |
| Transition to `all_paused` | 1 s | 1 s | 1 s |
| Inference while paused | 503 / 0.22 ms | 503 / 0.48 ms | 503 / 0.56 ms |
| Management while paused | 200 / 1.97 ms | 200 / 5.97 ms | 200 / 5.53 ms |
| Transition back to `accepting` | 34 s | 33 s | 31 s |
| Resumed inference | 200 / 47.1 ms | 200 / 451.0 ms | 200 / 420.4 ms |
| Resumed decode | 152.28 tok/s | 45.08 tok/s | 47.22 tok/s |
| Model unload | 0 s | 1 s | 1 s |

Pause was admission control, not an unload: resident model memory remained
stable while paused. On the Orins, daemon RSS was about 3.10 GiB after load and
3.17 GiB while paused; on carrack, model load added about 2.50 GiB to RTX 5090
memory. This confirms that pause prevents inference work while preserving the
warm model for fast resumption.

Evidence:

- `carrack`: `.sisyphus/evidence/pr1082-live-carrack-20260726T181830Z-2449721`
- `mesh1`: `.sisyphus/evidence/pr1082-live-mesh1-rerun-20260726T183753Z-64163`
- `mesh2`: `.sisyphus/evidence/pr1082-live-mesh2-20260726T182958Z-61186`

### Additional workload evidence

The first `mesh1` real-model attempt began while `--auto` was downloading an
independently selected 15.5 GB catalog model. After successful requested-model
load and baseline inference (HTTP 200, 650 ms), the activity detector remained
`all_paused` for the full 45-second pre-workload gate because the download was
still consuming host resources. That bounded run exited and cleaned up as
designed. Once the download completed, the quiet-state rerun above passed.
This supplies an additional real workload case: model acquisition itself
causes the daemon to yield rather than accept inference.

Artifact:
`.sisyphus/evidence/pr1082-live-mesh1-20260726T182958Z-63068`.

During the first broad universal-architecture build attempt on `mesh1`, two
parallel native compilations plus the pre-existing public daemon reduced
MemAvailable to about 122 MiB, and the kernel OOM-killed the old 61.6 GiB
daemon. This is consistent with the measured pause semantics: pausing preserves
resident model memory and cannot create capacity for an additional universal
CUDA build. The definitive SM87 build retained 54.88 GiB of available memory.
The pre-test `mesh1-public` service was restarted afterward with its original
installed binary and configuration. It was verified at version 0.72.1 in
`serving` state with `llama_ready=true`, 61,930,220 KiB RSS, and both original
models hosted again (Qwen3.6 35B and Gemma 4 12B).

The pre-existing LM Studio `llama-server` on carrack remained running throughout
(about 9.3 GiB on the RTX 3080), and no test-owned processes remained on any
host.

### GitHub CI

`gh pr checks 1082 --watch --interval 10` was kept attached through completion
at the final head. Result: **43 successful, 1 intentionally skipped, 0 failed,
0 cancelled, 0 pending**. The cache-invalidating Linux CUDA backend build was
the final job to complete and passed.

After the PR was marked ready, GitHub's `ready_for_review` event launched a
second full matrix at the same SHA. That run was also watched through
completion: **44 successful, 1 intentionally skipped, 0 failed, 0 cancelled,
0 pending**. Its Linux CUDA job passed on GitHub-hosted Ubuntu 24.04 runner
`GitHub Actions 1000062278`.

### Final assessment

The PR builds natively on x86_64/Ada+Blackwell and aarch64/Orin, the complete
daemon lifecycle suite passes on all three machines, and real loaded models
pause within one second under concurrent host workload, keep management
responsive, resume automatically after the configured idle interval, and
unload cleanly. The platform-specific CUDA failures found during certification
are fixed at the reported head.

i386:
very nice.

crates/mesh-llm-host-runtime/src/runtime/control_loop.rs
```diff
@@ -340,11 +697,36 @@ pub(super) async fn run_auto_runtime_event_loop(
...
                     }
+                    RuntimeEvent::StartupModelLoadFinished {
+                        model_ref,
+                        profile,
+                        result,
+                    } => {
+                        match result {
+                            Ok(()) => {
+                                ctx.model_target_reconciliation_state
+                                    .record_load_success(&model_ref, &profile);
```
i386: Startup models are marked in-flight before their tasks are spawned. If `/api/runtime/models` requests the same `(model_ref, profile)` meanwhile, `is_load_pending` stacks its oneshot and returns. This event only clears `in_flight_models`; neither arm drains `pending_load_completions`, and the HTTP route awaits that receiver without a timeout, so the request hangs indefinitely. Carry a `RuntimeLoadResponse` (or enough data to construct one) in this event and notify all stacked success/failure waiters. Please add a regression test for a load that piggybacks on eager startup.

crates/mesh-llm-host-runtime/src/runtime/control_loop.rs
```diff
@@ -218,6 +422,158 @@ pub(super) async fn run_auto_handle_control_request(
...
+                UnloadTarget::Instance(instance) => (
+                    canonical_model_ref.unwrap_or_else(|| instance.clone()),
+                    Some(instance.clone()),
+                ),
+            };
+            let intent_id = ctx
+                .model_target_reconciliation_state
+                .suppress_desired_with_id(
+                    &intent_model,
+                    "",
```
i386: Desired state is keyed by `(model_ref, profile)`, but every unload records `Absent` under the empty profile. A model loaded with `--profile low-ctx` therefore retains its effective `Present(model, low-ctx)` intent and can be reloaded on the next reconciliation tick after a successful unload. Preserve the loaded instance's profile (including instance-target unloads) and suppress the same key that the load created; cover this with profiled load→unload tests.

crates/mesh-llm-host-runtime/src/runtime/model_lifecycle/unload.rs
```diff
@@ -0,0 +1,300 @@
...
+            &model,
+            Some(&unload.instance_id),
+            &handle,
+            "shutting down",
+        ))
+        .await;
+    }
+    {
+        let mut record = lifecycle.lock().await;
+        let result = DrainCoordinator::default()
```
i386, line 229: This holds the per-instance lifecycle mutex across the poll/sleep loop for up to the drain timeout. `Node::begin_runtime_instance_request` takes the same mutex, so new requests during `Draining` block instead of being rejected immediately; a follow-up force drain cannot acquire the lock to tighten the deadline either. Mark draining under a short guard, then wait using cloned/shared drain state or brief per-poll locks, and re-lock only for the transition. The managed-model path above has the same issue.

crates/mesh-llm-host-runtime/src/protocol/convert.rs
```diff
@@ -698,6 +698,7 @@ pub(crate) fn local_ann_to_proto_ann(
             ann.stage_protocol_generation_supported,
             ann.stage_status_list_supported,
         ),
+        inference_admission_state: None,
```
i386: `build_announcement` computes `ann.inference_admission_state`, but the protobuf encoder always discards it here, while `proto_ann_to_local` tries to decode it. Consequently current peers never receive the documented coarse admission state; only the legacy availability withdrawal survives. Serialize the announcement value (privacy filtering already occurs when the announcement is built) and add a local→proto→local round-trip test for a paused state.

crates/mesh-llm-host-runtime/src/mesh/owner_control/mod.rs
```diff
@@ -897,6 +906,21 @@ impl Node {
             result?;
             return Ok(execution_shape);
         }
+        if is_model_lifecycle {
+            let cached = self
```
i386: The cache lock is released after a miss and before command execution/storage. Two concurrent retries with the same `(requester, request_id)` can both miss and enqueue the lifecycle command; because model loading is awaited by the control loop, the second queued load can execute after the first finishes and start another instance. Reserve/share an in-flight cache entry atomically, then replace/fill it on completion, and add a concurrent duplicate-request test.

crates/mesh-llm-system/src/activity/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    pub fn restore(&mut self) -> PriorityStatus {
+        if !self.reduced {
+            return self.status;
+        }
+
+        if let Err(reason) = self.controller.restore_priority() {
+            self.status = PriorityStatus::Degraded(reason);
+            self.reduced = false;
```
i386: A failed restore marks `reduced = false` even though the OS process remains deprioritized. Every later `restore()` then returns early with the latched degraded status, so restoration is never retried and the policy loop can never recover. This is common on unprivileged Linux/macOS, where increasing the nice priority back requires privilege. Do not claim the session is no longer reduced on failure; either use a reversible mechanism/gate support up front or retain retryable restore state, with a recovery test.

.github/workflows/pr_builds.yml
```diff
@@ -313,7 +310,9 @@ jobs:
     needs: changes
     if: ${{ (github.event_name == 'workflow_dispatch' || needs.changes.outputs.backend_changed == 'true' || needs.changes.outputs.benchmarks == 'true') && needs.changes.outputs.docs_only != 'true' }}
     name: Linux ${{ matrix.name }}
-    runs-on: ${{ fromJson(vars.USE_SELF_HOSTED == 'true' && matrix.self_hosted_runs_on || matrix.runs_on) }}
+    # The dedicated CUDA runner currently lacks the Node 24 external required
+    # by JavaScript actions inside job containers.
+    runs-on: ${{ fromJson(matrix.runs_on) }}
```
i386: This removes the documented `USE_SELF_HOSTED` CUDA routing contract, but neither `ci/ci.md` nor `manage-ci/references/current-inventory.md` is updated, and the temporary workaround has no owner, removal issue, or expiry. The CI rules require runner/variable changes to update both sides of the contract. Either restore the route after fixing the runner image or document the new topology and a bounded removal plan (and make the matching `ci.yml` update consistent).

i386 (dismissed):
Requesting changes because several lifecycle contracts can currently hang callers, reload an explicitly unloaded profiled model, block admission during drain, or fail to propagate the advertised admission state. I also found an owner-control idempotency race and a priority-restore state bug. Please address the inline findings and the existing unresolved ShellCheck finding for `log_command`. The current CI matrix is green, but these concurrency/profile/round-trip cases are not covered by it.

ndizazzo:
@coderabbit full review

crates/mesh-llm-host-runtime/src/runtime/control_loop.rs
```diff
@@ -218,6 +419,195 @@ pub(super) async fn run_auto_handle_control_request(
...
+                if let Some(tx) = completion {
+                    let _ = tx.send(Err(anyhow::anyhow!(
+                        "runtime mode client does not allow local model loading"
+                    )));
+                }
+                return;
+            }
+            if ctx
+                .model_target_reconciliation_state
+                .is_load_pending(&spec, &profile)
```
i386, line 446: The in-flight marker created by `reconcile_model_targets_once` is keyed by `action.model_ref` (the canonical catalog ref), but the reconciliation task sends its actual `Load` command with `action.load_spec` (a local filesystem path). If an owner/API load for that canonical ref arrives while reconciliation is running, this check sees the canonical marker and stacks its completion under the canonical key. The actual load then records/notifies using the path key, and `model_identity_matches` returns false for a path vs. catalog ref, so the canonical caller's oneshot is never resolved (the HTTP request waits indefinitely). Keep a stable reconciliation identity through the spawned load, or normalize/retarget the completion/in-flight keys before stacking and notifying. A regression test should cover a canonical API/owner load racing with a reconciliation load whose `load_spec` is a local path.

i386 (commented):
Found one concurrency/identity issue in the reconciliation load path.

i386 (approved):
Happy for you to merge after that one comment is resolved

ndizazzo:
@i386 SG! let's let this one soak a bit this week and schedule it for 0.75.0

PR #1083, opened 2026-07-26
fix(console-ui): chat transcript snapping during live status updates

### LINKED ISSUE #1077: Chat transcript snaps back to bottom during live status updates
## Problem

In live mode, scrolling upward through a long chat transcript only works briefly. The transcript repeatedly snaps back to the bottom.

The transcript should remain at the user's chosen position. Automatic scrolling should only continue while the user is already near the bottom.

## Steps to reproduce

1. Run the MeshLLM dashboard in live mode.
2. Open `/chat`.
3. Generate or open a response longer than the transcript viewport.
4. Allow the transcript to settle at the bottom.
5. Scroll upward.

The transcript moves upward and is then forced back to the bottom.

## Diagnostics

- MeshLLM commit: `cdd793b1cfd10cde1a0c6b56b7b8f72a9b003912`
- macOS 26.5.2 (`25F84`)
- Safari 26.5.2 (`21624.2.5.11.8`)
- Private two-node mDNS mesh using a split model
- `/api/events` produced 49 status payloads in 12 seconds

`ChatLayout` contains a `useLayoutEffect` without a dependency array:

https://github.com/Mesh-LLM/mesh-llm/blob/cdd793b1cfd10cde1a0c6b56b7b8f72a9b003912/crates/mesh-llm-ui/src/features/chat/layouts/ChatLayout.tsx#L150-L165

It runs after every render and unconditionally sets the transcript's `scrollTop` to `scrollHeight`, then calls `scrollIntoView()` on the bottom anchor. Each live status update causes another render and resets the user's scroll position.

## Extra context

This does not appear to be Safari failing to process the scroll gesture. Safari moves the transcript correctly before the next React render overwrites the position.

A likely fix is to track whether the transcript is near the bottom and only perform automatic scrolling while that state remains true. Conversation changes should reset that state appropriately.

bradleydwyer:
Fixes #1077.

## Original problem

In live mode, frequent status updates rerendered `ChatLayout`, whose layout effect unconditionally returned the transcript to the latest message. Readers scrolling through a long response were repeatedly snapped back to the bottom.

## Diagnostics

The live status stream produced 49 payloads in 12 seconds. Each payload updated the status query, rerendered the chat layout, and ran the unconditional scroll effect. Safari applied the user's scroll normally before the next React render overwrote the position.

## Fix

- Track whether the transcript is within 64px of the bottom and only follow new content while that remains true.
- Resume following when the reader returns near the bottom.
- Return to the latest message on conversation switches and explicit Send, Queue, or Retry actions.
- Re-check reader intent inside the deferred animation-frame scroll.
- Add regression coverage for unrelated rerenders, threshold behavior, conversation changes, send/queue/retry flows, and the animation-frame race.

This is a UI-only behavior change with no protocol, storage, or migration impact.

## Summary by CodeRabbit

* **New Features**
  * Chat transcripts now preserve your manual scroll position when you read older messages.
  * Automatic scrolling resumes when you’re near the bottom of the conversation.
  * Sending a message, retrying a response, or switching conversations returns the view to the latest message.

* **Bug Fixes**
  * Prevented delayed scrolling from overriding upward scrolling.
  * Improved transcript positioning during live updates, retries, and queued messages.

### DIFF
--- a/crates/mesh-llm-ui/src/features/chat/layouts/ChatLayout.tsx
+++ b/crates/mesh-llm-ui/src/features/chat/layouts/ChatLayout.tsx
@@ -15,6 +15,8 @@ const DESKTOP_SIDEBAR_QUERY = '(min-width: 1024px)'
 const CHAT_LAYOUT_FALLBACK_HEIGHT = 'calc(100dvh - 180px)'
 const CHAT_LAYOUT_MIN_HEIGHT = 320
 const CHAT_LAYOUT_KEYBOARD_GUTTER = 12
+// Keep this above the transcript's bottom padding: scrollIntoView aligns the anchor above that padding.
+const CHAT_STICK_TO_BOTTOM_THRESHOLD = 64
 
 type ChatLayoutStyle = CSSProperties & {
   '--chat-layout-height': string
@@ -30,6 +32,8 @@ type ChatLayoutProps = {
   children: ReactNode
   composer: ReactNode
   onMessageAreaClick?: () => void
+  /** Change this value to resume following the latest message after an explicit user action. */
+  stickToBottomKey?: string
 }
 
 function shouldUseDesktopSidebarFallback() {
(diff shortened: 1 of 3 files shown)


ndizazzo:
Thanks for the fix @bradleydwyer!

PR #1084, opened 2026-07-26
Handle K-only transposed KV page import and export

i386:
## Summary

- skip transposed V-cache copying before looking up a V stream when K-only layers contribute no V payload
- apply the same zero-V guard to KV page import
- add a focused llama.cpp regression test that round-trips a populated K-only transposed KV page
- keep the internal-API test inside the existing Windows/shared-library build guard
- use patch queue slot `0044` so this remains independent of #1078, which owns `0043`

## Root cause

K-only MLA layers are valid export/import selections: they contribute K bytes while `v_element_bytes` is zero. The non-transposed paths already skip zero-byte V rows, but both transposed paths indexed `layer->v_stream[strm]` unconditionally.

The symmetric guards make transposed import/export match the existing non-transposed behavior. There is no wire-format or ABI change.

## Validation

- applied the complete patch queue from pinned llama.cpp commit `76f46ad29d61fd8c1401e8221842934bf62a6064`
- verified the regenerated `0044` patch applies cleanly and reproduces the tested llama.cpp tree
- built `test-skippy-kv-page-export` with the CPU backend
- ran `ctest --test-dir build --output-on-failure -R '^test-skippy-kv-page-export$'` successfully

The current Linux CUDA slim failure occurs before checkout because the runner image lacks `/__e/node24/bin/node`; it is unrelated to this patch.

#1078 changes Metal small-batch matmul dispatch and Skippy parity benchmarking; it does not change KV page import/export.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed KV page export/import for “V-transposed” cases where the V stream is empty, preventing incorrect processing.
  * Ensured exported and imported key-only transposed KV page descriptors remain correct and payload data round-trips properly.
* **Tests**
  * Added a regression test covering transposed KV page export/import, validating descriptor stability and payload equality.


ndizazzo (approved):
Needs a rebase from main after merging https://github.com/Mesh-LLM/mesh-llm/pull/1085, otherwise good

Issue #1090, opened 2026-07-27
Add Poolside Laguna family support to Skippy split serving


michaelneale:
## Feature

Add end-to-end support for Poolside's Laguna model family in Mesh/Skippy, beginning with the official Laguna XS 2.1 and Laguna S 2.1 GGUF releases.

The intended outcome is normal Mesh model discovery, layer-package preparation, single-node serving, and multi-peer Skippy split serving with correctness evidence for Laguna's MoE and mixed global/sliding-window attention architecture.

## Why now

- Mesh's current llama.cpp pin is `178a6c44937154dc4c4eff0d166f4a044c4fceba` from 2026-07-20.
- Base Laguna support merged upstream two days later in [ggml-org/llama.cpp#25165](https://github.com/ggml-org/llama.cpp/pull/25165), merge commit `1f66c3ce1c26c95db3fadb734086c7d9fba23bb9`.
- Poolside now publishes official, single-file GGUFs for two useful targets.
- Laguna's stage-boundary payload is the ordinary hidden state. The family is causal MoE with hybrid attention, not recurrent, and does not appear to need a new activation sideband.
- This should therefore be an upstream-pin/patch-rebase, graph-filtering, topology-policy, packaging, and certification task rather than a new architecture implementation from scratch.

Poolside's model cards still contain some pre-merge wording saying llama.cpp support is under review; the upstream PR is now merged. DFlash speculative decoding remains in Poolside's llama.cpp fork and should be treated separately from base Laguna execution.

### P0: Laguna XS 2.1

- Model: [poolside/Laguna-XS-2.1](https://huggingface.co/poolside/Laguna-XS-2.1)
- Official GGUF: [poolside/Laguna-XS-2.1-GGUF](https://huggingface.co/poolside/Laguna-XS-2.1-GGUF)
- Source revision: `1a37c0a5fb8c7a18e6106decb6be6327d1b63fa6`
- Preferred artifact: `Laguna-XS-2.1-Q4_K_M.gguf` (20.3 GB)
- Reference artifact: `Laguna-XS-2.1-BF16.gguf` (66.9 GB)
- Shape: 33B total / 3B active, 40 layers, activation width 2048, 10 global-attention + 30 SWA layers, 256 routed experts + one shared expert
- License: OpenMDW-1.1
- Role: first implementation and correctness target; practical local and two-node test model

### P0: Laguna S 2.1

[...]

PR #1095, opened 2026-07-28
feat(skippy): add Poolside Laguna staged-runtime candidate

michaelneale:
## Status

This draft adds staged-runtime support for the pinned Poolside Laguna S 2.1 Q4_K_M artifact and records real M5 + Vast Mesh evidence. The model has passed strict local two- and three-stage parity and ordinary two-node inference over Iroh/QUIC.

Suffix N-gram depth `2` is now published in the pinned package as its default serving policy, and this branch teaches the runtime to consume that package field. A no-config resolver test proves the package selects suffix N-gram and depth `2`. A Vast-only cold start on the exact revision reached stable direct-Iroh placement and downstream-stage readiness, then was deliberately paused before stage 0 finished downloading or any inference ran. Live package-default consumption therefore remains unproven.

The PR remains a draft because it also carries generic dependencies and timeout-cancellation changes with open review findings. Those must land or be extracted before merge.

## Pinned artifact

- source: `poolside/Laguna-S-2.1-GGUF@edd093522473dc7313b0738d8b4116b7f8b9745f`
- source SHA-256: `a34c74e46688122bef83122f4133031bababbefcf57436dde97048c91e2cc6ff`
- package: `meshllm/laguna-s-2.1-Q4_K_M-layers@0c467ad441ee94cb5a76f626294d963c4048507d`
- package manifest SHA-256: `0250cfb54ceeb94a9c71e48df447f780e32fc625553844d6403770f315be0237`
- package policy: suffix N-gram match `5..32`, proposal cap `48`, fixed verify window `1..32`, pipeline depth `2`

The Hugging Face policy update is immutable at [commit `0c467ad`](https://huggingface.co/meshllm/laguna-s-2.1-Q4_K_M-layers/commit/0c467ad441ee94cb5a76f626294d963c4048507d). All evidence is limited to this Q4_K_M artifact and the revisions above.

## What changes

- add Laguna staged graph filtering and stage-boundary output to the carried llama.cpp patch queue
- recognize Laguna and its 3,072-wide activations in topology capability inference
- add a reviewed capability record for the exact pinned Q4_K_M artifact; F16 remains the default wire dtype and Q8 remains untested
- preserve immutable Hugging Face source revisions through quant discovery, package jobs, catalog updates, and suggested CLI commands
- add the prediction-return listener required by the strict three-stage correctness harness, including stalled-peer shutdown
- allow layer packages to publish an optional positive verification `pipeline_depth`; omission preserves the legacy depth of `1`
- make the serving resolver consume the package depth and validate/report it during package preflight
- record the exact package, parity, Mesh, N-gram, stability, and limitation evidence

### Strict local correctness on M5 Max / Metal

- two stages `0..24 / 24..48`: full and staged execution both predicted token `674`
- three stages `0..16 / 16..32 / 32..48`: full and staged execution both predicted token `674`
- both strict commands exited `0` without allowing mismatches
- F16 activation payload is 6,144 bytes per boundary for activation width 3,072

### Real M5 + Australian Vast Mesh serving

- ordinary `mesh-llm` nodes ran on M5 Max/Metal and an Australian Vast RTX 6000 Ada/CUDA worker
- normal planning launched the capacity-bounded package split `0..36 / 36..48`
- stage traffic used direct Iroh/QUIC; there was no SSH stage forwarding
- the run configured context `262144`, F16 activation wire, F16 KV, four lanes, `n_batch=512`, and `n_ubatch=128`
- the exercised request contained 44,460 prompt tokens and generated 65 tokens; this proves roughly 44K-token operation, not full 256K saturation
- a three-request miss/hit/miss recovery probe returned HTTP 200 and byte-identical generated content on all requests
- decode throughput was 19.74, 18.24, and 15.42 tok/s
- the exact-prefix hit reused 44,416 prompt tokens
- each request proposed 64 suffix N-gram tokens and accepted 32 at pipeline depth `2`
- a 63-request stability harness returned 63 HTTP 200 responses
- the native log scan found no unsupported-trim, all-lanes-busy, reset, decode, slot, or proactive-eviction fatal errors

RTT varied materially during the run, including transient spikes; this PR does not claim a single fixed latency figure.

### Published default policy

- all 51 package artifacts passed size and SHA preflight at the new immutable package revision
- preflight reports the exact suffix N-gram and depth-2 policy above
- the runtime remains backward-compatible: packages without `pipeline_depth` resolve to depth `1`
- the no-config resolver regression test proves this package shape selects `ngram-suffix` and verify pipeline depth `2` without a local speculative override

## What is not proven

- completion of a fresh live launch using the new package revision with no speculative CLI/config override. The paused Vast-only attempt did not reach inference; a new-package M5 launch is not yet claimed
- full 128K or 256K context saturation or soak
- Q8 activation wire, Q8 KV quality, DFlash, or any other Poolside size/quant
- structured Poolside tool use: the formal harness passed 6/13 checks (sequential 4/5, overlapping 0/5, exact-prefix 1/1, same-prefix 0/1, native-log scan 1/1)
- fixed latency expectations across different Iroh routes

In the failed tool cases Laguna often emitted textual `<tool_call>` content that was not projected into OpenAI `tool_calls`; one sequential response ended with `finish_reason=stop`. This is not a native split crash, but it means this PR does not claim tool-use certification.

## Remaining merge work

1. Resume on two dedicated Vast workers against package revision `0c467ad…`, with explicit `--max-vram` caps matching physical GPU memory and no speculative strategy/depth override. Wait for both stages, run a small smoke plus the 44,460-token N-gram request, and require telemetry to report effective `ngram-suffix`, pipeline depth `2`, successful inference, and clean shutdown. Metal correctness is already covered by strict M5 parity; a fresh M5 live launch can follow if desired.
2. Land #1100 and fix or extract the generic timeout-cancellation review findings carried by this branch.
3. Rebase onto the landed common heads, rerun focused correctness/package checks, and mark this PR ready.

## Validation

- clean llama.cpp patch-queue application
- `just build`
- embedded-native `just release-build`
- strict `skippy-correctness single-step` and three-stage `chain`
- full package preflight: 51/51 artifacts present with matching sizes and SHA-256 values
- real two-node Mesh long-context, recovery, and stability runs described above
- `cargo test -p skippy-model-package --bin skippy-model-package` — 51 passed
- `cargo test -p skippy-topology --lib` — 33 passed
- `cargo test -p mesh-llm-host-runtime --lib package_suffix_strategy_resolves_as_a_standalone_proposer` — passed
- `cargo check -p skippy-model-package -p skippy-runtime -p skippy-topology -p mesh-llm`
- `cargo clippy -p skippy-model-package -p skippy-runtime -p skippy-topology -p mesh-llm --all-targets -- -D warnings`
- `cargo fmt --all --check`

Tracks #1090. Depends on #1100.


i386:
Review findings at `6bd11c5`:

- **[P1] Catalog provenance becomes stale across source revisions.** Existing variants update only `packages`; their `source.revision` is neither refreshed nor represented per package. Repackaging the same filename from another revision can therefore publish new artifacts while the catalog still identifies the old source revision. Please include the revision in variant/package identity for both dictionary and list formats. [Relevant code](https://github.com/Mesh-LLM/mesh-llm/blob/6bd11c5a06fd5699b4274e7ec47ba0789b6eb252/crates/model-package/src/scripts/split-model-job.sh#L434-L469)

- **[P2] Listener teardown can hang indefinitely.** After accepting a connection, blocking reads never re-check `shutdown`, while `Drop` unconditionally joins the thread. A stalled peer can wedge the correctness process. Add read timeouts with shutdown checks—or retain a stream handle and force `Shutdown::Both`—plus a stalled-peer regression test. [Relevant code](https://github.com/Mesh-LLM/mesh-llm/blob/6bd11c5a06fd5699b4274e7ec47ba0789b6eb252/crates/skippy-correctness/src/runner/prediction_return.rs#L75-L121)

- **[P2] Quant-list output discards the selected revision.** Discovery correctly uses `repo@revision`, but JSON omits it and the suggested command prints `repo:quant`. Copying that command silently packages current `main` instead of the inspected revision. Emit `sourceRevision` and `repo@revision:quant`. [Relevant code](https://github.com/Mesh-LLM/mesh-llm/blob/6bd11c5a06fd5699b4274e7ec47ba0789b6eb252/crates/mesh-llm-commands/src/model_package.rs#L256-L288)

The latter two overlap CodeRabbit comments and are independently confirmed.

i386:
Reviewed the update range `a2e0d65c…25727d92` (one commit). I found two cancellation gaps that should be addressed:

1. **[P1] A timed-out local request can still start and run the full prefill.** The timeout path cancels the request token, but the local-runtime implementation unconditionally calls `runtime.prefill(...)` ([lines 330–343](https://github.com/Mesh-LLM/mesh-llm/blob/25727d92e92b1f9dc8ede52b2b6096f651af9a9b/crates/skippy-server/src/frontend/local_generation.rs#L330-L343)); its first cancellation check is not until the decode loop ([lines 606–616](https://github.com/Mesh-LLM/mesh-llm/blob/25727d92e92b1f9dc8ede52b2b6096f651af9a9b/crates/skippy-server/src/frontend/local_generation.rs#L606-L616)). If the deadline expires while the blocking task is waiting for the runtime mutex or doing cache restore, it will acquire the runtime and perform an expensive prefill even though cancellation is already set, continuing to hold the generation permit/GPU after the 504. Please check cancellation before starting prefill (and before subsequent cache persistence), and add a test that proves the real blocking generation exits/releases capacity; the current router test only proves that the token flag is set on a synthetic pending future.

2. **[P2] `SkippyModelHandle` drops the new context-aware calls.** The new trait defaults fall back to the context-free methods, but this forwarding impl still only overrides `chat_completion` / `completion` ([lines 968–997](https://github.com/Mesh-LLM/mesh-llm/blob/25727d92e92b1f9dc8ede52b2b6096f651af9a9b/crates/mesh-llm-host-runtime/src/inference/skippy/mod.rs#L968-L997)). Any caller using `chat_completion_with_context` or `completion_with_context` through a `SkippyModelHandle` therefore silently loses the cancellation token at this wrapper boundary. Please add forwarding overrides, ideally with a wrapper-level regression test.

The router/wrapper propagation added in this commit otherwise looks internally consistent.

michaelneale:
The three review findings on the Laguna/package change are fixed at `28e66d15`:

- catalog updates now refresh immutable source provenance for existing dictionary- and list-style variants, and package entries record `source_revision`
- quant-list JSON emits `sourceRevision`, and the suggested command preserves `repo@revision:quant`
- prediction-return listener teardown retains a cloned stream, forces `Shutdown::Both`, and has a stalled-peer regression test

Validation completed successfully:

- affected `mesh-llm-commands`, `model-package`, and `skippy-correctness` tests
- focused `drop_stops_when_connected_peer_stalls` regression
- `cargo check` for all touched crates plus `mesh-llm`
- warning-denying Clippy for all touched crates plus `mesh-llm`
- workspace formatting and diff checks

The later timeout-cancellation review has two separate, still-open generic findings. The PR description now calls those out explicitly; they are not being treated as resolved by this commit.

i386 (commented):
Review of current head `624fc7b9`.

**[blocking] Add an explicit Laguna cache-policy assertion.** The new reviewed `laguna` capability is not covered by `family_policy::tests::every_reviewed_family_has_an_explicit_cache_policy`; the current CI run reproduces this as `reviewed family laguna has no explicit policy assertion`. Please add the Laguna policy expectation (and keep its intended K/V defaults explicit) so adding it to the reviewed catalog cannot silently leave the loader/planner policy unguarded.

The rest of this pass found no additional blocking issue.

michaelneale:
## Paused Vast-only default-policy checkpoint (2026-07-29)

The follow-on was deliberately paused before inference. This is a partial startup result, not a completed runtime gate.

What completed:

- exact branch head `6ee85338` built as an embedded CUDA SM120 release; deployed binary SHA-256 `3d4db6a06b130723a0b3aea2e9d250354ca6a97c9057d74a4d05631c76e8c8f4`
- two dedicated Australian Vast workers joined a private Mesh; the public-mesh Vast node was not used
- actual GPUs were RTX PRO 6000 Blackwell Max-Q 96 GB plus RTX 5090 32 GB
- both nodes used package `meshllm/laguna-s-2.1-Q4_K_M-layers@0c467ad441ee94cb5a76f626294d963c4048507d`, context `262144`, F16 activation wire, F16 KV, one lane, and no speculative strategy/depth override
- direct Iroh/QUIC settled around 29–32 ms after an initial relay sample
- manifest SHA-256 matched `0250cfb54ceeb94a9c71e48df447f780e32fc625553844d6403770f315be0237`
- with explicit physical-capacity caps (`--max-vram 90` and `30`), normal planning chose `0..45 / 45..48`
- downstream stage `45..48` reached `ready` with activation width 3072, runtime context 262144, one lane, and F16 wire
- stage 0 cold materialization reached 27/48 snapshot entries (about 32 GB) before the requested pause

An important lab finding: without `--max-vram`, the node-capacity field advertised roughly 984 GB and 64 GB despite `nvidia-smi` and the GPU list correctly reporting 96 GB and 32 GB. Explicit caps made placement inputs truthful. This generic capacity-advertisement discrepancy should be investigated separately; it is not Poolside-specific.

Not completed or claimed:

- stage 0 readiness
- any inference request
- live proof that the package selected suffix N-gram and verify depth 2
- throughput, acceptance, cache, or stability results from this Vast-only attempt

Exact resume gate:

1. Provision two dedicated Vast workers; do not use the public-mesh node.
2. Deploy the exact Poolside branch release and set `--max-vram` to physical GPU budgets until the capacity discrepancy is fixed.
3. Launch both against revision `0c467ad…`, context 262144, F16 wire/KV, and no speculative CLI/config override.
4. Wait for both stages to report ready; record the realized ranges and direct/relay route.
5. Run a small smoke, then the existing deterministic 44,460-prompt-token / 65-completion-token request.
6. Require response telemetry to show suffix proposer, proposal cap 48, verify depth 2, successful inference, and clean shutdown; repeat once for exact-prefix behavior if healthy.
7. Update the PR/issue with only those observed results. Strict M5 Metal two- and three-stage parity is already proven; a fresh M5 live launch is optional after the Vast gate.

Cleanup completed: both Poolside processes exited cleanly, and the Poolside builder, 5090, and Pro 6000 instances were destroyed. The existing public-mesh RTX 3090 remains running and untouched.

michaelneale:
Combined replacement branch is now open as #1118. It carries the current work from this PR onto current `main`, linearizes the shared llama.cpp patch queue, and separates prior live evidence from validation run on the combined head. Please continue review on #1118; this PR is left open for provenance until maintainers decide to close/supersede it.

ndizazzo:
Superseded by #1118

Issue #1094, opened 2026-07-28
crates.io half-published at v0.74.0: hf-hub [patch] hides breakage from CI and leaks into published crates


michaelneale:
## Problem

`v0.74.0`'s crates.io publish died 28 of 42 crates in. crates.io is currently **half-published: 27 crates at `0.74.0`, 15 not**.

It failed verifying `model-package`:

```
error[E0308]: mismatched types
    .repo_id(candidate.target_repo.clone())
             expected `&str`, found `String`
note: method defined here --> hf-hub-1.0.0/src/repository/mod.rs:914:9
error: failed to verify package tarball
```

The one-character fix is in #1093. These are the underlying problems it exposed.

## Why it wasn't caught

`[patch.crates-io]` points `hf-hub` at our fork. That patch applies to **our workspace only**:

- `cargo check` / `clippy` / `just build` / CI → compile against the **fork**, where `.repo_id(String)` is fine → green.
- `cargo publish` verifies the tarball **outside** the workspace → `hf-hub` resolves from the registry (`1.0.0`), where `repo_id` takes `&str` → error.

So the bug was invisible to every normal build and only surfaced mid-GA-publish, after 27 crates had gone out irreversibly.

Preflight should have caught it, but `publish-crates.sh --dry-run` deliberately skips verification for crates whose sibling registry deps aren't published yet:

```bash
"dry-run cannot verify ${crate} until ${dep}@${workspace_version} exists in crates.io"
```

`model-package` depends on `model-hf`/`model-ref`, so it was never verified. **That whole class of crates has no pre-publish type-check against real registry deps.**

## The fork leaks into published crates

`[patch.crates-io]` is not transitive. `model-hf@0.74.0` on crates.io declares:

```
hf_hub ^1.0.0-rc.1   (package = hf-hub)
```

i.e. **upstream**, not our fork. So anyone doing `cargo add model-hf` (or consuming `mesh-llm-host-runtime`, `mesh-llm-commands`, `model-package`) builds against upstream hf-hub and silently loses everything the fork carries.

### The fork is still needed

Upstream `main` now has byte-range reads, which is easy to mistake for resume:

```rust
range: Option<std::ops::Range<u64>>
```

That's a caller-supplied range read, **not** resume-after-interruption. Upstream `cache/storage.rs` has no partial-blob handling. Our fork adds exactly that:

```rust
Resume(u64), ExistingBlob,
incomplete_path() -> "{}.incomplete"
[...]

PR #1096, opened 2026-07-28
fix: use published mesh-llm-hf-hub crate

i386:
## Summary

- Switch all published hf-hub consumers to the `mesh-llm-hf-hub` crates.io package.
- Remove the workspace-only `[patch.crates-io]` Git override.
- Fix `model-package` to borrow `repo_id`, matching the published fork API.

This addresses the dependency-resolution gap in #1094: the workspace and crates.io package verification now compile against the same published fork instead of silently resolving `hf-hub` upstream outside the workspace.

@michaelneale — please review the fork/package dependency decision and the compatibility fix.

## Validation

- `cargo check -p model-hf -p model-package -p mesh-llm-commands -p mesh-llm-host-runtime`
- `cargo clippy -p model-package --all-targets -- -D warnings`
- `cargo fmt -p model-package -- crates/model-package/src/bin/queue-unsloth-layer-packages.rs`

The registry-backed check initially reproduced the `repo_id(String)` failure from #1094; after the borrow fix it passes.

## Summary by CodeRabbit

* **Chores**
  * Updated model, host runtime, and command components to use the standard Hugging Face Hub integration (newer HF Hub library version).
  * Removed a custom workspace override so dependency resolution follows the default crates registry behavior.
* **Bug Fixes**
  * Improved queued package upload marker behavior by avoiding unnecessary string cloning when generating the Hugging Face repository identifier.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3889,9 +3889,9 @@ dependencies = [
 
 [[package]]
 name = "mesh-llm-hf-hub"
-version = "1.0.0"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc7862ddcdf3979479ab8d667b4d234aa558ca2bdf3b35b1d18f24fe9fe4d04d"
+checksum = "43088a838cf0c6715c65f65a5ac99045fd6d6e90949a8a4183b8104ab791e96b"
 dependencies = [
  "base64",
  "bon",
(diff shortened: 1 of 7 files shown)


i386:
CI diagnosis: the dependency switch is resolving the real crates.io package correctly, and it exposed that `mesh-llm-hf-hub@1.0.0` does not contain the fork’s resumable-download implementation. The failing check is `model_hf::tests::download_file_resumes_existing_incomplete_cache_blob`, which receives no `Range` request. The next fix is to publish a patched `mesh-llm-hf-hub` release (for example `1.0.1`) from the resumable fork line, then retarget this PR; the regression test should remain enabled.

@michaelneale

ndizazzo (approved):

PR #1099, opened 2026-07-28
Refresh llama.cpp upstream patch queue

i386:
## Summary

- Rebase all 47 carried llama.cpp patches onto upstream 9a3bf2b84923a85583b4ee8177b0cca13824bb03.
- Update the pinned upstream revision and regenerate the durable patch queue.
- Resolve upstream drift in staged OpenAI-MoE filtering, GLM-DSA Metal integration, and architecture-test metadata.

## Root cause

The upstream canary failed before compilation because patch 0002 referenced an obsolete common/CMakeLists.txt preimage. Later upstream Metal refactors required a full queue refresh rather than a one-hunk patch fix.

## Validation

- Clean checkout: LLAMA_WORKDIR=<temporary directory> scripts/prepare-llama.sh pinned
- just build (patched Metal ABI built successfully; mesh-llm binary produced)
- ./target/debug/mesh-llm --version

## Summary by CodeRabbit

* **New Features**
  * Added an expanded runtime API for staged and multimodal model execution.
  * Added configurable sampling, chat templates, tool-aware responses, reasoning formats, generation signals, and external media decoding.
  * Added support for model parts, resident prefixes, session state transfer, MTP draft generation, and broader architecture coverage.
  * Added GLM-DSA sparse execution and optimized Metal backend paths.

* **Bug Fixes**
  * Improved activation handling, KV-page edge cases, dispatch safety, and model loading reliability.

* **Tests**
  * Expanded coverage for chat parsing, KV-page handling, GLM-DSA execution, and backend performance.


i386:
## Rebase-conflict rationale

This is a queue rebase, not a feature change. The canary failed because patch 0002 carried an obsolete common/CMakeLists.txt preimage; upstream had reordered the common source list, so git am --3way could no longer construct a useful fake ancestor. Rebasing the full 47-commit series onto upstream revision 9a3bf2b8 was safer than hand-editing that first mail patch and deferring the later conflicts.

### Is ours in conflict with theirs?

At the patch-text level, yes: both upstream and the carried queue changed the same files and, in several cases, the same control-flow or declaration anchors. At the behavior level, no mutually exclusive intent was identified. Each resolution keeps the upstream behavior and the staged-runtime behavior, selecting the appropriate path by architecture or execution mode. The highest residual risk is the Metal encoder refactor, which is why this remains a draft pending CI canary coverage.

### 1. Staged OpenAI-MoE filtering

The carried patch makes a stage run layers il_start through il_end, starts from an activation input for a nonzero il_start, and returns an activation boundary when include_output is false. New upstream behavior added two independent output paths:

- embeddings_nextn_masked controls whether unused-token row selection happens inside the final layer.
- t_h_nextn plus a later row selection supports the complementary next-token embedding path.

The rebase makes stage-local selection require all three conditions: this is the final layer of the selected stage, output IDs exist, and embeddings_nextn_masked is enabled. The stage-boundary return remains before t_h_nextn and the output head, so an activation-only stage neither constructs classifier output nor accidentally fills a full-model next-token result. With filtering disabled, control flow remains upstream behavior.

### 2. GLM-DSA Metal integration

Current upstream substantially refactored the Metal encoder path: a formerly monolithic encode implementation is now split into individual operation helpers and updated fusion/concurrency handling. The carried GLM-DSA patch adds sparse-attention, selected-row, routing, weighted-reduction, and GLM-MoE Metal machinery across the operator registry, pipeline lookup, argument structs, C++ dispatch, and Metal shaders. Replacing upstream sections wholesale would have discarded current upstream behavior.

Two overlaps were resolved deliberately:

- Upstream now declares the FWHT Metal argument struct beside the queue's GLM-MoE argument structs. The rebased patch retains the upstream typedef, then starts a separate typedef for the GLM route and weighted-sum structs. The patched Metal ABI compiled locally, which catches malformed declaration boundaries.
- Upstream's Hadamard/FWHT fast path remains ahead of the carried GLM decode-block diagnostic and dispatch handling. The conditions are independent and both may return their specialized Metal encoder path.

This preserves upstream's new Metal support while keeping all carried GLM-DSA kernels and their registrations connected to the current encoder architecture.

### 3. Architecture-test metadata

The architecture test helper builds synthetic GGUF metadata. Upstream added MiniMax M3 / MSA requirements: its indexer uses one head per GQA/KV head, plus block-size and local-block metadata. The carried GLM-DSA tests require different metadata: one indexer head, a 128-wide key, fixture-controlled top-k, and IndexShare metadata for top-k frequency, skip offset, and full/shared indexer-type sequences.

The resolved helper selects the M3 head count only for MiniMax M3, preserves its block/local fields, and selects the GLM-DSA key/top-k and IndexShare fixtures only for GLM-DSA. Neither synthetic model consumes the other architecture's metadata.

### Validation and remaining coverage

- A fresh temporary llama.cpp checkout applied all 47 regenerated patches cleanly at the new pin.
- The patched macOS Metal ABI compiled successfully through the repository build workflow.
- The rebuilt debug binary responds to mesh-llm --version.

The scheduled canary remains the Linux/Rust and smoke-test validation for this exact upstream revision. No mesh wire-protocol change or intentional staged-runtime ABI change is included here.

ndizazzo (approved):
I'm good with this, but we should consider implementing the changes suggested from @coderabbit before merging. I have an agent working on this currently... please stand by

PR #1100, opened 2026-07-28
skippy: recover hybrid speculative windows and orphaned lanes

michaelneale:
## Why

Hybrid/recurrent stages cannot always trim a speculative verification suffix with the normal positional KV operation. A failed verify request can also close its binary connection before `Stop`, leaving an execution lane allocated. Laguna made both failures reproducible, but the fixes are runtime-wide and are shared with Inkling.

This PR isolates that common recovery work from the Poolside family support in #1095 and the Inkling model work in #1017.

## What changes

- keep bounded in-flight checkpoints for hybrid/recurrent verification windows
- checkpoint only partial recurrent state instead of copying the full long-context attention cache
- restore recurrent state, trim the attention suffix, and replay only the accepted prefix after a rejected window
- trim interleaved sliding-window-attention sessions using the logical KV length rather than assuming dense positional storage
- preserve existing plain, RWKV7, Gemma3n, and GLM-DSA activation sidebands during replay
- reclaim binary-stage sessions and lanes when a connection exits before graceful `Stop`
- keep connection/session tracking and its tests in a focused module

There is no Mesh wire-format or Skippy ABI change.

### Local gates

- the llama.cpp patch queue applies cleanly at the pinned upstream revision
- `just build`
- 329 `skippy-server` library tests pass
- warning-denying Clippy and workspace formatting pass

### Real distributed recovery

The exact common head was exercised through the Poolside Laguna branch on ordinary M5 Max / Metal and Australian Vast RTX PRO 6000 Blackwell Max-Q / CUDA `mesh-llm` nodes:

- normal Mesh planning launched a `0..24 / 24..48` split over Iroh/QUIC
- runtime context was 262,144; the measured prompt contained 44,460 tokens
- a cache-miss / exact-cache-hit / cache-miss sequence produced identical output
- suffix N-gram pipeline depth 2 proposed 64 tokens and accepted 32 on each request
- recovery occurred after rejected speculative windows without a fatal trim or decode failure
- completion throughput was 19.74, 18.24, and 15.42 tok/s
- the final stability harness completed 63/63 HTTP requests successfully
- the native log scan found no unsupported trim, all-lanes-busy, reset, decode, slot, or proactive-eviction fatal errors

This is the evidence that was previously listed as “still to prove.” It demonstrates bounded recovery and lane stability on a real heterogeneous, long-context Mesh path.

## Limits of the claim

The Poolside harness also found model-facing failures: overlapping textual tool calls were not parsed into OpenAI `tool_calls`, and the same-prefix cache check missed while exact-prefix reuse passed. Those are not fixed or claimed by this PR. This PR claims native verification-window recovery and lane cleanup only.

It also does not make suffix N-gram depth 2 a model/package default; that policy belongs with the certified package.

## Validation

- clean llama.cpp patch-queue apply
- `just build`
- `cargo test -p skippy-server --lib` — 329 passed
- `cargo fmt --all --check`
- distributed M5 + Vast evidence described above

Supports #1017, #1090, and #1095.

## Summary by CodeRabbit

* **New Features**
  * Added verify-checkpoint retirement (new ABI/API) and retirement flow for speculative decoding.
  * Introduced verify-retirement messaging and automatic handling in stage execution/forwarding.
  * Added stage protocol generation 4 support and updated stage-state to v11.
  * Added cleanup of orphaned binary connection sessions to prevent stuck session state.

* **Bug Fixes**
  * Improved verify-window state handling across trimming for hybrid runtimes.
  * Enforced the verify-window pipeline depth limit (max 64).

* **Documentation**
  * Updated compatibility and protocol docs for generation-4 and stage-state v11.

* **Tests**
  * Added coverage for retirement round-trips and correct message ordering.


i386:
@michalneale

[P1] `skippy_checkpoint_verify_window` appends a checkpoint for every hybrid/recurrent verify window, but the only removal paths are trim/reset/import. A fully accepted window does not invoke `trim_session` (`trim_branch_suffix_or_retire` returns early when the canonical position reaches the verification position), so a normal successful long generation accumulates checkpoints until the hard limit is hit. The 65th successful verify then fails with `verify-window pipeline depth exceeds supported checkpoint retention`.

Please retire checkpoints once a window is known to be fully accepted, while retaining only the bounded windows that may still need a partial-accept replay. A regression test with 65+ fully accepted verify windows would catch this.

third_party/llama.cpp/patches/0046-Preserve-hybrid-verify-window-state-across-trim.patch
```diff
@@ -0,0 +1,439 @@
+From cf0c069381520c0098fbac525b0b95d1f73b6816 Mon Sep 17 00:00:00 2001
```
ndizazzo: This needs repositioning in the patch queue after #1105 merged

ndizazzo (changes requested):
General changes are fine, but rebase + patch stack adjustment are conflicting but not getting flagged as such

michaelneale:
🤖 @ndizazzo The latest main and #1105 patch queue are merged, the hybrid recovery patch is repositioned after the refreshed queue, and the follow-up checkpoint-retirement/recovery findings are addressed. CI is fully green. Could you please re-review when convenient?

michaelneale:
Combined replacement branch is now open as #1118. It carries the current work from this PR onto current `main`, linearizes the shared llama.cpp patch queue, and separates prior live evidence from validation run on the combined head. Please continue review on #1118; this PR is left open for provenance until maintainers decide to close/supersede it.

ndizazzo:
Superseded by #1118

michaelneale:
Final provenance note: the outstanding orphan-cleanup error-propagation concern from this superseded PR is addressed in replacement PR #1118 at `64ab9f32`. Cleanup failures now reach the connection supervisor and are combined with any original connection error. #1100 remains closed; its effective recovery work lives in #1118.

PR #1109, opened 2026-07-29
fix: Qwen3.6 models now serve on the correct recurrent family

michaelneale:
Qwen3.6 models now resolve to the correct model family, so they serve with the right prefix-cache policy instead of silently falling back to a wrong one.

Before this change, only one exact model reference worked — `unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL`. Any other quant or uploader of the same model silently resolved to the wrong family, got no recurrent ranges, and selected a `ResidentKv` prefix cache instead of `KvRecurrent`. That is a wrong-but-plausible configuration, so it degraded serving rather than failing with a clear error.

Now every Qwen3.6 spelling resolves correctly:

| Model | Before | After |
|---|---|---|
| `unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL` | `qwen35moe` | `qwen35moe` |
| `unsloth/Qwen3.6-35B-A3B-GGUF:Q4_K_M` | `qwen3moe` | `qwen35moe` |
| `bartowski/Qwen3.6-35B-A3B-GGUF:Q4_K_M` | `qwen3moe` | `qwen35moe` |
| `Qwen/Qwen3.6-35B-A3B-Instruct-GGUF:Q5_K_M` | `qwen3moe` | `qwen35moe` |
| `unsloth/Qwen3.6-27B-GGUF:UD-Q4_K_XL` | `qwen3_dense` | `qwen35` |
| `unsloth/Qwen3.5-35B-A3B-GGUF:Q4_K_M` | `qwen3moe` | `qwen35moe` |

Qwen3.5 was affected by the same bug and is fixed by the same change.

## Root cause

Qwen3.6 is not its own architecture. It loads as llama.cpp's `qwen35` / `qwen35moe` pair, which `docs/skippy/LLAMA_PARITY.md` already certifies. The pinned llama.cpp has no `qwen36` architecture.

Family inference only matched the dotless `qwen35` spelling. Release names always spell the series with a dot (`Qwen3.6-35B-A3B`), and the compacted identity keeps that dot, so the check could never match a real repo name. Inference then fell through to the active-parameter MoE heuristic, which saw the `a3b` tag and confidently returned `qwen3moe`.

The single certified reference only worked because it matched a hardcoded `model_id` entry in `reviewed-family-capabilities.json`, which made that JSON entry load-bearing for correctness.

Note that simply stripping dots during normalization does not fix this: `qwen3.6` becomes `qwen36`, not `qwen35`. The series names have to be recognized explicitly.

## Validation

The existing `certified_recurrent_families_never_use_resident_kv_policy` test encodes exactly the invariant this bug broke. Confirmed it fails without the inference fix and passes with it:

```
# without the fix
unsloth/Qwen3.6-35B-A3B-GGUF:Q4_K_M: Auto { payload: ResidentKv, min_tokens: 256, max_entries: 16 }
test result: FAILED. 0 passed; 1 failed
```

Qwen3.6 references are now pinned into that test so this cannot regress silently.

Compared inference output across a 30-model corpus before and after. Only the Qwen3.5/3.6 rows changed; all other families are byte-identical, including the dotted `GLM-4.7-Flash` and `MiniMax-M2.7` names that were the main regression risk.

Parameter-count names are explicitly guarded, since `Qwen3-5B` compacts to `qwen35b`:

- `Qwen/Qwen3-5B-GGUF:Q4_K_M` -> `qwen3_dense`
- `Qwen/Qwen3-35B-A3B-GGUF:Q4_K_M` -> `qwen3moe`

Tests, formatting, and warning-denying clippy pass on every touched and dependent crate: `skippy-topology` (35), `mesh-llm-host-runtime` (1851), `skippy-prompt` (24), `skippy-bench` (73).

Not covered: no end-to-end decode against real Qwen3.6 weights. This is validated at the family-inference and cache-policy layer only.

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing additional Qwen3.5/Qwen3.6 naming variants and selecting the correct Qwen3.5-series family capability.
  * Improved recurrent caching policy resolution for Qwen3.5/Qwen3.6 models.
  * Prevented Qwen parameter-count-style names (e.g., Qwen3-5B, Qwen3-35B) from being misclassified.
* **Bug Fixes**
  * Updated model allowlist assertions to match the intended “Auto” recurrent KV policy.
* **Documentation**
  * Clarified that Qwen3.6 releases inherit the Qwen3.5 architecture/policy guidance.
* **Tests**
  * Added unit tests covering Qwen3.5/Qwen3.6 inference and edge-case naming.


ndizazzo (approved):
oops! lol

Issue #1115, opened 2026-07-30
Adopt Mixture-of-Agents synthesis pattern, minus its tool-calling and durability gaps


micspiral:
## Background

Together AI's [Mixture-of-Agents](https://github.com/togethercomputer/moa) reports 65.1% on AlpacaEval 2.0 against GPT-4o's 57.5%, using only open-weight models. The pattern: fan out a prompt to several models in parallel, then have an aggregator model read all responses and write one.

mesh-llm already fans out and arbitrates (`crates/mesh-mixture-of-agents/`). This issue tracks which parts of the paper's approach are worth adopting, which are actively wrong for a mesh, and what the reference implementation cannot do at all.

## What their implementation actually does

Aggregation is one paragraph of prompt plus string concatenation, from `utils.py`:

```python
system = f"""You have been provided with a set of responses...
Responses from models:"""
for i, reference in enumerate(references):
    system += f"\n{i+1}. {reference}"
```

No parsing, no voting, no confidence, no length bound. All the synthesis happens inside the aggregator model's weights.

## Gap 1 — tool calling is absent

Grepping `moa.py`, `advanced-moa.py`, `utils.py`, `bot.py` for `tool_call` / `tools` / `function_call` / `tool_choice` returns one hit: `from functools import partial`. Their `generate_with_references(model, messages, references, max_tokens, temperature, generate_fn)` has nowhere to put a tool schema.

This is not a gap in their code quality — it is a category they never entered. And it is fatal for our use case, because `references` are built from response **`.content`**, which is empty when a model returns a tool call.

Verified against 9 open-weight models on OpenRouter. Every worker proposed `list_dir({"path":"src"})`; the aggregator received:

```
1. 
2. 
3. ool_call>
4. 
5. 
```

Five blank lines plus one leaked `ool_call>` fragment. The aggregator did emit a correct call — but only by re-deriving it from the original prompt. **The synthesis step contributed nothing.**

## Gap 2 — no durability

`run_llm` catches only `RateLimitError`:

- any other exception (timeout, 500, connection reset) propagates out of `asyncio.gather` and kills the whole turn — one flaky model takes down the request
- exhausted retries leave `response` unbound → `UnboundLocalError`, also fatal

No per-worker isolation, no timeout, no partial-result path. For a heterogeneous mesh of consumer machines this is disqualifying on its own.

## Gap 3 — no length bound

[...]

PR #1116, opened 2026-07-30
MoA: mesh answers improve as capable nodes join

michaelneale:
Ready for review. Every number below is measured through the shipped entrypoint (`moa::handle_turn`) against real open-weight models; method, raw numbers, and **withdrawn** results are in `evals/moa-openrouter/RESULTS.md`.

Closes part of #1115.

## What this delivers

The virtual `mesh` model gets stronger as capable nodes join, degrades gracefully when there aren't any, and never convenes a committee that measured worse than serving one model.

- **Reasoning/answer turns** → workers draft in parallel, one aggregator synthesizes. Always synthesizes on a multi-worker answer turn — shipping one worker's text verbatim was a pure loss mode.
- **Tool turns** → route to the single best tool-caller (voting/advice measured null-to-harmful there).
- **Everything best-effort** → bounded waits; a slow or dead peer costs the grace window (10s), not the worker timeout.

## The core result

Capable pool (`qwen3-32b`, `mistral-small-24b`, `qwen3.5-35b-a3b`, `qwen3-14b`), vs the pool's **best single member** answering alone. 40 preregistered prompts × 2 draws, out-of-family position-swapped judge, length logged:

**71W / 8T / 1L, p<0.0001** — and it wins **28 of 29** trials where the MoA answer was *shorter* than solo, so this is not a length artifact. Reproduced across three runs (65W/0L, 62W/1L, 71W/1L).

Before the fixes below, the same pool through the same entrypoint **lost** 26W/14T/40L.

## Three shipped-path bugs

The mechanism was fine; the wiring never ran it.

1. **Grace pre-empted synthesis.** `first_answer_grace` armed on the *first* answer and finalized the turn, so on fast backends it fired on ~every turn and shipped one worker's text — measured 80/80 early-exit at capable scale. Grace now finalizes on **tool** turns only; on answer turns it is a collection deadline (stop waiting for the tail, then synthesize what arrived). Private-mesh window 3s → 10s.
2. **Role-tiered budgets truncated the drafts synthesis consumes** (Fast 256 / Specialist 512). Those tiers existed to make the grace fast-path cheap; answer turns now pack every worker at the full budget.
3. **Answer turns shipped one worker verbatim when drafts agreed.** Agreeing drafts are the best input to synthesis, not a reason to skip it.

## Pool shaping (all measured)

- **Verified sizing** — tiers come from GGUF tensor sums gossiped via the existing `parameter_count_b`, not name parsing. Name parsing mis-tiered real models (`gemma-4-E4B` stores 7.5B, not 4B). No verified size ⇒ ranks small, so an unparseable alias can't pose as big-tier. (i386 P1s.)
- **Admission** — verified-small workers are dropped only when ≥2 verified-big remain. A lone big + smalls keeps the mix, because dropping the smalls collapses the committee to a solo model (that mix measured 47W/27T/5L vs solo).
- **All-small pools serve their best member.** An 8B-class pool with an 8B reducer never beat its best member: 2×8B **0W/43T/37L**, 4×8B 5W/63T/12L, 6×8B **5W/52T/23L** (p=0.0009), with consistently shorter answers. Rather than ship a measured regression, such a pool collapses to its strongest member and the request degrades to serving it. Scope: this is a weak reducer synthesizing weak drafts, **not** evidence small-model MoA cannot work — the untested cell is small peers with a *strong* reducer. One mid-size node changes the pool from all-small, and MoA engages again.
- **Committee cap** — 6 all-small, 4 once a verified big is present. Fan-out is ~2N+1 calls/turn and quality is flat past ~4, so a 20-node mesh doesn't fan out to 20.
- **Graceful degrade** — a single model (or a collapsed pool) rewrites `mesh` to a real served model and routes normally. Only a node serving nothing returns 503. On `main` this is a 503.

## Withdrawn

The width-sprint headline — "6× diverse 8B beats its best member, 12W/65T/2L, p=0.013" — **did not replicate.** The same rig on the same 40 tasks and pool now gives 3W/76T/1L (p=0.63), with 11 of 40 tasks flipping verdict. It rested on ~14 decided trials out of 80 (the rest ties) from a single unreplicated run selected out of a width sweep. Withdrawn in `RESULTS.md` and it motivated the all-small gate above.

## Also fixed

Nine eval-vs-production divergences found by measuring the shipped path (grace finalizing before synthesis; reducer/refiner truncation discarding most of each answer; prompts telling workers to "be concise"/"be direct"; named vs anonymous reducer inputs; the worker preamble on the reducer; role-tier token starvation). Plus a production panic in `mesh-llm-guardrails` (byte-slicing a multi-byte character), with regression tests. Two hypotheses were tested, **rejected by measurement, and reverted** rather than rationalised — including one of my own (`min_grace_answers = N-1`) that stalled public-mesh turns to the 60s worker timeout.

## Protocol / compatibility

No wire, gossip, plugin, or skippy-ABI changes. Sizing reads the existing `ServedModelMetadata.parameter_count_b` (now populated from GGUF tensor sums instead of a name regex). New behaviour is in-memory config (`ReferencePolicy`, `RefinementPolicy`). One caller-visible change: `reasoning_effort`/`enable_thinking` on a `model=mesh` request is ignored — that override only selected a broken config where reasoning models return `content: null`.

## Validation

```
cargo test -p mesh-mixture-of-agents        all pass
cargo test -p mesh-llm-host-runtime        1911 pass
cargo test -p mesh-llm-guardrails            19 pass
clippy -D warnings (moa, host-runtime, guardrails, mesh-llm)   0 errors
cargo fmt --all --check                                        clean
```

Robustness contracts verified by simulation: a slow worker cannot stall a turn, a lone survivor still answers, a hanging refiner cannot hold the turn, patience expiry releases held consensus, partial worker death is fully accounted.

**Live release check** on this branch (built here, real node): local private inference, `model=mesh` on a single node (200, degrades), console index + assets + `/api/status` + `/api/models` + `/v1` passthrough chat + SSE streaming, joined the **public mesh (13–16 peers, mixed 0.65.1/0.72.1/0.74.0, 12-model union)**, MoA fanout there with 2 of 6 workers failing and the turn still completing, and model-by-name routing to a remote peer.

**Public-mesh contrast vs released v0.74.0** (same 3B model, same prompts, `model=mesh`): released never synthesizes (`reducer=false` on all three probes, two early-exiting at 3–4s); this branch always synthesizes. An earlier revision of this branch rode the 60s worker timeout on two of three probes — root-caused to `min_grace_answers`, reverted, and the grace window now bounds a dead peer at 10s.

## Honest limits (for the reviewer)

- Quality numbers come from the OpenRouter-backed harness driving the real `handle_turn`, **not** a live capable-node mesh. The mesh-config paths (admission, cap, all-small gate, self-fill) are unit-tested and the plumbing is live-verified, but the capable-pool quality lift itself has not been reproduced on real mesh hardware.
- **Single judge model.** A judge-level length bias was found and controlled (r fell from +0.68 to +0.13–0.30); the capable-pool result survives it in the hard direction (wins while shorter), the small-pool results are marginal either way.
- `judge_pair` collapses genuine ties, position disagreements, and API/parse failures into the same `0`, so tie counts cannot be read as agreement.
- Significance is computed per draw (80) rather than per prompt cluster (40), which overstates confidence. The capable-pool result holds either way; the small-pool ones are marginal.
- Ties dominate at small scale — "stronger" means wins-a-minority and rarely-loses, not uniformly better.
- The `+14.8k` line count is almost entirely recorded traces and eval data; the code surface is 13 files.


i386 (commented):
One functional issue found in the committee-cap path.

crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/workers.rs
```diff
@@ -383,6 +396,291 @@ async fn add_worker_backend(
...
+        let mb = &models[b];
+        let tool_a = tool_use_by_base
+            .get(&canonical_base_name(&ma.name))
+            .copied()
+            .unwrap_or(crate::models::CapabilityLevel::None);
+        let tool_b = tool_use_by_base
+            .get(&canonical_base_name(&mb.name))
+            .copied()
+            .unwrap_or(crate::models::CapabilityLevel::None);
+        // 1) higher tool_use first
```
i386: [P1] Do not use the tool-actor ranking to choose a reasoning committee. `cap_committee` calls `compute_actor_candidates`, whose first key is `tool_use`; this can evict a 32B/70B model with `tool_use=None` in favour of four 8B models whose names/metadata advertise tool use. That affects ordinary answer turns too, even though `tool_use` is irrelevant there and contradicts the admission-control goal of not letting weaker workers displace a stronger one. Please rank the capped committee by answer-quality/size (or keep separate actor and committee rankings), and add a >4-worker regression case.

crates/mesh-mixture-of-agents/src/worker.rs
```diff
@@ -130,6 +130,56 @@ pub fn has_quality_gap<'a>(workers: impl IntoIterator<Item = (&'a str, WorkerRol
...
+/// internally for role assignment, exposed so the host doesn't re-implement it.
+pub fn model_name_is_small_tier(name: &str) -> bool {
+    is_single_digit_b_name(name)
+}
+
+/// Canonical base of a model name, mirroring the host's dedup logic: lowercase,
+/// drop an `@branch` segment (keeping any `:quant` tag), strip common
+/// prefixes/suffixes, keep only alphanumerics. Two aliases of the same model
+/// map to the same base.
+pub fn canonical_base_name(name: &str) -> String {
```
i386, line 148: [P1] This makes an unparseable name a “big-tier” model, but admission control now uses that result to exclude other workers. A small fine-tune or alias such as `my-assistant` therefore bypasses the weak-worker filter (and can become the Strong/reducer), while the PR’s central safety claim is that weak nodes cannot drag down a strong pool. Please represent unknown size as unknown and use verified parameter/model metadata when available; only apply the destructive admission filter when the relative tier is actually known. Add a regression with a no-size alias plus explicit small/large workers.

i386 (commented):
A second issue concerns the capacity signal that gates admission.

i386:
Follow-up on the tiering signal: parsing `1B`–`9B` from a served name is brittle. Aliases/fine-tunes need not include a size, model names can include unrelated numbers, and the current fallback treats every unparseable name as big. That makes a destructive admission decision from an unverified label.

The better source is GGUF inspection: enumerate the GGUF tensors and sum their `element_count` to obtain the exact total stored-parameter count, then gossip that value through the existing `ServedModelMetadata.parameter_count_b`. Use it for MoA tiering when present; retain name parsing only as a lower-confidence fallback, and represent missing data as `Unknown` rather than `big`. For MoE we may eventually also want a distinct active-parameter signal, but total tensor count is already a materially better safety gate than the name heuristic.

michaelneale:
@i386 thanks - have it reading real data now

michaelneale:
@i386 — fixed the tiering P1s you flagged (name-based tier / unparseable-name-as-big driving destructive admission). Now exactly your prescription: authoritative size from GGUF tensor sum, gossiped via the existing `parameter_count_b`, used for MoA tiering with name-parse only as a lower-confidence fallback and missing = **Unknown** (never big, never filtered).

Commit: `3b94cf0d` (rebased/merged forward since). Three parts:

**1. Producer computes the real size** — `crates/model-artifact/src/gguf.rs`, `scan_gguf_total_parameters`: sums every tensor's element count from the header/tensor-info table (no tensor data read):
```rust
pub fn scan_gguf_total_parameters(path: &Path) -> Option<u64> {
    let GgufHeader { file: mut f, n_tensors, n_kv } = open_gguf_header(path)?;
    skip_all_kv_pairs(&mut f, n_kv)?;
    let mut total: u64 = 0;
    for _ in 0..n_tensors {
        let _name = read_gguf_string(&mut f).ok()?;
        let n_dims = read_u32(&mut f).ok()?;
        if n_dims > MAX_GGUF_TENSOR_DIMS { return None; }
        let mut elements: u64 = 1;
        for _ in 0..n_dims {
            let dim = read_u64(&mut f).ok()?;
            elements = elements.checked_mul(dim)?;
        }
        let _ggml_type = read_u32(&mut f).ok()?;
        let _offset = read_u64(&mut f).ok()?;
        total = total.checked_add(elements)?;
    }
    Some(total)
}
```

**2. Producer prefers it, falls back to name** — `crates/mesh-llm-host-runtime/src/models/profile.rs`: `parameter_count_b` = tensor sum (÷1e9) when the local GGUF is present, else `parameter_count_b_from_text(name)`, else `None`. Gossiped via `ServedModelMetadata.parameter_count_b`.

**3. Consumer tiers off the gossiped value** — `crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/pool.rs`. The MoA orchestrator has no peer GGUFs, so it reads gossiped `parameter_count_b` only:
```rust
enum SizeTier { Small, Big, Unknown }

fn tier_for(name: &str, sizes: &HashMap<String, f64>) -> SizeTier {
    if let Some(b) = sizes.get(&canonical_base_name(name)) {
        return if *b < SMALL_TIER_MAX_B { SizeTier::Small } else { SizeTier::Big };
    }
    match mesh_llm_guardrails::model_param_size_b(name) {   // name-parse fallback
        Some(b) if (b as f64) < SMALL_TIER_MAX_B => SizeTier::Small,
        Some(_) => SizeTier::Big,
        None => SizeTier::Unknown,                          // missing = Unknown, not big
    }
}
```
Admission excludes only *verified* `Small` and counts only *verified* `Big` toward the ">=2 big remain" gate — `Unknown` is never strong and never filtered. `cap_committee` also now ranks by this size, not the tool-actor ranking (your other P1 — a `tool_use=None` big model was getting evicted for tool-advertising small ones on answer turns).

Verified live on a 2-node mesh (MacBook 3B + mini Qwen3.5-4B): gossiped sizes went from a bogus name-parsed **4910** to real **3.21 / 4.21**, and admission now no-ops on unknown-size workers. Two new unit tests: `admission_never_filters_unknown_size_worker`, `admission_falls_back_to_name_parse_without_gossip`.

Note deliberately deferred to follow-up: MoE active-parameter signal (total tensor count is the safety gate for now, as you said). Does this match what you had in mind?

i386:
> **Producer prefers it, falls back to name** — `crates/mesh-llm-host-runtime/src/models/profile.rs`: `parameter_count_b` = tensor sum (÷1e9) when the local GGUF is present, else `parameter_count_b_from_text(name)`, else `None`. Gossiped via `ServedModelMetadata.parameter_count_b`.

I think the fallback to name should be removed. if you cant get this info from gguf, the model must be ranked as the lowest param model. 

As would be worth chucking anything in your cache at this and checking if that tensor counts matches the name to validate.

ndizazzo:
# MeshLLM PR 1116 Homelab Validation Report

| Field | Value |
|---|---|
| Candidate commit | `ed90d75e156fbd6ea896f9a9edee5e427017deb6` |
| Pull request | `Mesh-LLM/mesh-llm#1116`, “MoA: ensembling that improves as mesh nodes join” |
| Base / current main | PR base `70d4f3338c5020cdc9e05d1863a6b3787b5bed3c`; tested current `origin/main` `b7c26ebe789ec72a0a58db85d20b7c4a04e5832d` |
| Previous/latest GitHub release | `v0.74.0`, published 2026-07-27 |
| Validation run ID | `20260805T043845Z-ed90d75` |
| Started / completed (UTC) | 2026-08-05 04:38 / 2026-08-05 05:26 |
| Decision | `NOT_READY` for the claimed quality improvement; live MoA functionality `PASS WITH DEFECTS` |

## General Findings

PR 1116's MoA machinery works through the shipped OpenAI HTTP path on a real
three-node private homelab mesh. Fan-out, same-model self-fill, refinement,
reduction, streaming, single-actor tool calls, tool-result continuation, bounded
peer loss, and rejoin were all observed. All three final nodes report the same
mesh ID, two peers, a ready CUDA runtime, and no severe log events.

The PR is nevertheless **NOT_READY for its headline claim that the shipped
`mesh` model improves as nodes join**. The committed results document explicitly
reports the shipped `moa::handle_turn` path at 9W/59T/12L (`p = 0.66`) versus the
best pool member and calls the unresolved harness-versus-production gap
merge-blocking. The stronger 49W/24T/6L, 48W/23T/2L, and 11W/68T/1L results
cannot be independently recomputed from committed trial outputs; the repo
contains corpora, small fixtures, scripts, and narrative results, but not the
scaled per-trial judge outputs. Re-running the paid OpenRouter studies and a
second judge was outside this homelab run.

Two additional release defects were found: repository `just check-release`
fails its Linux/aarch64 CUDA asset-parity assertion, and the chat endpoint accepts
missing or incorrectly typed `messages` and fabricates a 200 response. These do
not invalidate the live MoA observations, but they reinforce the `NOT_READY`
decision (unrelated).

## Scope And Exclusions

Validated:

- Fresh detached PR worktrees and native builds on `mesh1`, `mesh2`, and
  `carrack`.
- CUDA 13.2 / SM87 on two Jetson AGX Orin 64 GB systems; CUDA 13.3 / SM120+SM86
  on RTX 5090 + RTX 3080.
- Extracted product bundles, host import policy, private QUIC mesh, management
  status, `/v1/models`, chat completions, SSE, tool calls, tool results, negative
  requests, peer shutdown/rejoin, and final log scan.
- Same-model topology using `unsloth/Qwen3-0.6B-GGUF:Q4_K_M` on all nodes.
- Local Rust tests, formatting, and Clippy for MoA, guardrails, host runtime, and
  shipped binary.

## Canonical Change Inventory And Validation Ledger

| ID | Category | Release claim | Planned checks | Status | Evidence / defects |
|---|---|---|---|---|---|
| MOA-1 | FEATURE | Answer turns draft, optionally refine, and aggregate | Live non-stream and divergent answer request | PASS | 200 response; `fanout`, 4 successful proposal-stage calls (2 drafts + 2 refinements), reducer used once |
| MOA-2 | FEATURE | Same-model meshes self-fill from distinct nodes, capped at two active participants | Three nodes, one identical model; inspect early-exit and layered accounting | PASS | Early-exit: 2/2 worker calls; divergent turn: 4/4 calls across draft+refine; reducer true |
| MOA-3 | FEATURE | Tool turns route to a single best actor | Required tool request and continuation | PASS | Tool call had 1/1 worker, valid `get_weather({city:"Toronto"})`; continuation labeled `tool-result` |
| MOA-4 | FEATURE | Streaming remains OpenAI-compatible | SSE request | PASS | 200 event stream, content chunk, terminal stop chunk, `[DONE]` |
| MOA-5 | FEATURE | Slow/absent peers do not stall turns | Stop carrack, request through mesh1, rejoin carrack | PASS | Request returned in 949 ms with 2/2 successful calls; carrack rejoined same mesh |
| MOA-6 | REVISION | Weak worker admission control | Unit/integration suite and source review | PARTIAL | Focused suite passes; not reproduced live with two distinct verified big models plus a weak model |
| MOA-7 | BUG_FIX | UTF-8 guardrail panic fixed | Guardrails suite and final log scan | PASS | 15 guardrail tests pass; no panic/fatal/error in final logs |
| MOA-8 | RESULT | Shipped MoA improves over best member | Review committed study and live functional evidence | FAIL | Own `RESULTS.md` says shipped path is parity: 9W/59T/12L, p=0.66, with unexplained orchestration gap |
| API-1 | BUG_FIX | Production OpenAI path remains contract-safe | Valid, invalid JSON, missing/wrong `messages`, unknown model | FAIL | Invalid JSON 400 and unknown model 404; missing/wrong `messages` incorrectly return 200 with fabricated prose |

## Environment And Build Provenance

All worktrees were created at `~/dev/mesh/mesh-llm-pr1116`, detached at the
exact PR SHA, then built with repository `just` recipes. `dist/native-runtimes/`
is untracked build output; source files remained unchanged.

| Host alias | OS / arch | Hardware / backend | Canonical build commands | Product archive SHA-256 | Manifest / import policy | Status |
|---|---|---|---|---|---|---|
| `mesh1.patio51.com` | Ubuntu 24.04 / aarch64 | AGX Orin 64 GB, CUDA 13.2, SM87 | `just release-host-build`; `MESH_LLM_CUDA_TOOLKIT_MAJOR=13 LLAMA_STAGE_CUDA_ARCHITECTURES=87 just release-runtime-build cuda`; `just release-bundle-aarch64-cuda ...` | `c7a80b3a66c32260533ed709f1a921928f14d94e0f3ea7cd43dc9ce564d033ca` | ABI 0.1.35, CUDA 13; zero rejected host imports | PASS |
| `mesh2.patio51.com` | Ubuntu 24.04 / aarch64 | AGX Orin 64 GB, CUDA 13.2, SM87 | same as mesh1 | `2f3941be95d91d9fcc10bdef7b7495e5618d95231baa4a67eff3372ff4973132` | ABI 0.1.35, CUDA 13; zero rejected host imports | PASS |
| `carrack.patio51.com` | Arch Linux / x86_64 | RTX 5090 + RTX 3080, CUDA 13.3, SM120+SM86 | `just release-host-build`; `MESH_LLM_CUDA_TOOLKIT_MAJOR=13 LLAMA_STAGE_CUDA_ARCHITECTURES='120;86' just release-runtime-build cuda`; `just release-bundle-cuda ...` | `d1cc74704b88b5c837cd7c7dab481024efa4aec587ea5ee6fca3a084c6b9dd31` | ABI 0.1.35, CUDA 13; zero rejected host imports | PASS |

`patchelf 0.19.1` was installed on carrack because the release packaging script
requires it. The generic `release-runtime-build cuda` defaults its manifest to
CUDA 12 even when compiling with CUDA 13, so the explicit
`MESH_LLM_CUDA_TOOLKIT_MAJOR=13` setting was required for runtime compatibility.

## Product Validation Results

| Area | Case | Expected | Observed | Status |
|---|---|---|---|---|
| Packaging | Execute extracted products | Host selects adjacent fresh CUDA 13 runtime | All three selected ABI 0.1.35 runtime and loaded model | PASS |
| Startup | Fresh candidate startup | API, console, model ready | All hosts ready; final severe event count 0 | PASS |
| Private mesh | Three candidate nodes | Same mesh ID, 2 peers each, private | `380326a05b75a239f9f2ab2911d583b7`, 2 peers each, private | PASS |
| OpenAI API | Same-model exact answer | MoA observability and usable content | `early-exit`, 2/2, 508 ms, `MOA_OK.` | PASS |
| Inference | Divergent reasoning answer | Layered path and reducer | `fanout`, 4/4 stage calls, reducer attempt 1, 2874 ms | PASS |
| Streaming | Chat SSE | Valid chunks and `[DONE]` | 3 data lines, expected content, `[DONE]` | PASS |
| Tool calls | Required call | One actor and valid JSON arguments | 1/1 worker, valid Toronto call | PASS |
| Tool results | Continue after result | Final prose response | `tool-result`, 1/1, 705 ms | PASS |
| Reliability | 3 loops each for `auto` and `mesh` | Tool-call/result parity, streaming and non-streaming | `mesh` 12/12; `auto` 10/12 | PARTIAL |
| Nightly smoke | chat, stream, tools on `auto`,`mesh` | Stable success | All configured functional steps passed; attestation unconfigured | PASS |
| Negative API | Invalid JSON / unknown model | 4xx | 400 / 404 | PASS |
| Schema API | Missing/wrong `messages` | 400 | 200 with unrelated generated answer | FAIL |
| Failure / recovery | Remove and rejoin carrack | Remaining mesh serves; same mesh restored | 949 ms request succeeds; carrack rejoins with 2 peers | PASS |
| Logs | Severe event scan | No panic/fatal/error | mesh1 438 JSON lines, mesh2 316, carrack 284; 0 severe | PASS |
| Release lane | `just check-release` | Clean | aarch64 CUDA asset parity mismatch | FAIL |

## Private Mesh Evidence

Final topology (left running):

| Node | API / console / QUIC | Node ID | Peers | Model |
|---|---|---|---|---|
| mesh1 | 19337 / 13131 / 17842 | `00c6868b1b` | 2 | `unsloth/Qwen3-0.6B-GGUF:Q4_K_M` |
| mesh2 | 19447 / 13145 / 17843 | `520112d875` | 2 | same |
| carrack | 19557 / 13155 / 17844 | `c45f63ec2e` | 2 | same |

## API, Logs, UI, And Inference Findings

The same-model early-exit request produced 2 worker summaries, matching the
two-participant self-fill cap. The divergent request produced 4 summaries
because accounting includes both two draft calls and two refinement calls; it
does not imply four distinct nodes. Reduction ran once. Streaming deliberately
does not carry post-hoc `x-moa-*` headers, but emitted a valid OpenAI event
sequence.

The tool request selected one actor and returned structurally correct arguments.
Across the dedicated three-attempt reliability run, all 12 `model=mesh` phases
passed. Two `model=auto` tool-result phases omitted the fixture codeword; this is
consistent with weak-model output reliability and is not hidden by the aggregate.

No final log contained a structured `error`/`fatal` event or panic/fatal text.
Non-JSON lines were expected TUI/progress output. carrack used 3.8 GiB on its
RTX 5090 and 0.9 GiB on its RTX 3080 at the final snapshot. The Orin management
API reported roughly 59.3 GB usable VRAM per node.

## Defects And Anomalies

| Defect | Severity | Reproduction | Expected / actual | Release impact |
|---|---|---|---|---|
| Shipped quality path remains parity | S1 | See `evals/moa-openrouter/RESULTS.md` lines 427–492 | Headline improvement / 9W-59T-12L, p=0.66 | Blocks the claimed result; document itself calls it merge-blocking |
| Chat schema accepts absent/wrong `messages` | S2 | POST `{"model":"mesh"}` or string `messages` | 400 / 200 fabricated unrelated prose | Contract and safety defect |
| aarch64 release consistency failure | S2 | `just check-release` | Pass / expected `...cuda.tar.gz`, configured `...cuda-13.tar.gz` | Blocks clean release validation |
| Generic CUDA recipe stamps CUDA 12 by default | S2 | CUDA 13 host, `just release-runtime-build cuda` without override | Detect installed toolkit / incompatible CUDA 12 manifest | Fresh runtime rejected until explicit environment correction |
| `model=auto` tool-result omissions on 0.6B | S3 | Reliability harness, 3 attempts | 12/12 / 10/12 | Small-model route limitation; `model=mesh` was 12/12 |
| Release build emits 28 unfulfilled lint-expectation warnings | S3 | `just release-host-build` on Orins | Clean release output / warnings | Build hygiene only; explicit Clippy gates pass |

## Risk And Follow-Up Register

| Risk / gap | Required action | Status |
|---|---|---|
| Harness win does not survive shipped orchestration | Diff exact prompt bytes and execution decisions, close parity gap, rerun the same study | OPEN, blocking |
| Single-judge and absent raw trial-output risk | Publish per-trial artifacts and rerun with a second judge | OPEN, blocking for statistical claim |
| No live big-model quality comparison | Run candidate through two verified 24–32B participants and a controlled baseline | OPEN |
| Admission behavior only unit-tested here | Add live two-big-plus-small topology assertion and record selected endpoint identities | OPEN |
| OpenAI request schema too permissive | Reject absent/non-array `messages` before MoA routing and add regression tests | OPEN |
| Release metadata inconsistencies | Fix CUDA asset parity and toolkit-major default/detection | OPEN |

## Final Gate Assessment

| Gate | Result | Evidence / rationale |
|---|---|---|
| All feature claims dispositioned | PARTIAL | Runtime behavior covered; statistical lift and live admission quality not proven |
| Canonical packages proven on real hosts | PASS | Three extracted CUDA 13 products run on two architectures |
| APIs, logs, and inference proven | PARTIAL | Core paths pass; request schema defect remains; no visual UI check |
| Candidate private mesh proven | PASS | Three nodes, same mesh ID, two peers each, failure/rejoin tested |
| Required mixed-version compatibility | NOT_APPLICABLE | Homogeneous PR request; no claimed protocol/ABI change |
| No open S0/S1 defect | FAIL | Headline shipped-path quality claim remains unsupported/parity |
| Required CI lanes terminal | PASS | Required GitHub checks passed; irrelevant matrix lanes skipped |
| Residual risk accepted and bounded | FAIL | Core improvement claim requires new evidence, not a waiver |

## Sign-Off

Decision: **NOT_READY for merge/release on the stated quality-improvement
claim.** The live three-node MoA feature is functional and can remain running for
further investigation. Exact next actions: fix the shipped-path parity gap and
API schema defect, make the release consistency gate pass, publish/recompute
per-trial evidence with a second judge, then repeat a live capable-model mesh
comparison.

michaelneale:
@ndizazzo the cuda thing seems unrelated, and can you try that again with non trivial model as i can't see that model yielding much at all, so probably results are too noisy.

michaelneale:
@ndizazzo yeah the claim is only for larger models combining (20B ish) unfortunately, so need to test with that (still probing myself). Also want to make sure this improves over the main branch. 

but this reminds me to try it with larger combos... so standby.. (in the mean time feel free to try it with larger - and some subsequent fixes)

michaelneale:
## Update: width sprint + review fixes (since `ed90d75e`)

Addressed the homelab report and pushed several fixes. Key finding first.

### Small-model meshes DO improve — with width, not depth

The homelab run used `Qwen3-0.6B`, an order of magnitude below the smallest model in our data (8B). Our committed table shows an 8B floor (2–3× 8B is null; 4× wins), so a 0.6B null is the expected result, not a design signal.

A fresh sprint (aggregator `qwen3-8b`, 8B-class peers, shipped committee path, position-swapped judge, length-noted) found the real lever is **committee width**:

| pool | single-agg vs best member | refine vs single-agg |
|---|---|---|
| 2× 8B diverse | 2W/77T/1L, p=1.0 | null |
| 4× 8B diverse | 5W/73T/0L, p=0.06 | null |
| **6× 8B diverse** | **12W/65T/2L, p=0.013** | null |
| 6× 8B same | 4W/75T/1L, p=0.38 | null |

Three results:
1. **Six diverse 8B models beat the best single member (p=0.013).** The old flat `MAX_COMMITTEE_WORKERS=4` throttled exactly the small pools that need width. Cap is now **tier-aware**: 6 for all-small, 4 when a verified big model is present (a 24–32B pair already wins at 2).
2. **The refine round never earns its serial cost** — `refine vs single-agg` is null in every cell. Hermes' single-aggregation cadence matches Together's layered shape here at half the latency. (Noted for follow-up; not changed in this PR.)
3. **At 8B, diversity matters** (6 diverse 12W/2L ≫ 6 same 4W/1L) — unlike mid-scale where Self ≈ Mixed.

### Review fixes pushed

- **i386 P1 ×2**: tiering now uses verified GGUF tensor-sum size gossiped via `parameter_count_b` (not name parsing); `cap_committee` ranks by size, not `tool_use`. Name fallback **removed** per follow-up — no verified size ⇒ ranked weakest. Validated: cached 3B→3.21, 4B→4.21; `gemma-4-E4B` name says 4B but sums to 7.5B (proves the name heuristic was unsafe).
- **Iron law**: self-fill rebuilt to use **distinct physical endpoints only** — a single node can never fake a 2-worker committee; it degrades to single-model. (Also closes the two CodeRabbit self-fill bugs.)
- **API-1**: `model=mesh` with missing/non-array/empty `messages` now returns **400**, not a fabricated 200. Verified live.
- **CodeRabbit**: `pool.rs` extracted; eval judges parallelised.

### Still open (agreed)
- Capable-model (24–32B) shipped-path run for the headline quality claim — length-controlled single judge (AlpacaEval-style), needs paid runs, separate effort.
- aarch64 `check-release` parity + CUDA-toolkit-default: **pre-existing**, not touched by this branch (`check-release` passes on this HEAD; branch changes no release/packaging code).

Full data + method in `evals/moa-openrouter/RESULTS.md`.

michaelneale:
@ndizazzo — thanks for the thorough homelab validation. Re-review requested; here's what changed since `ed90d75e`, mapped to your report:

**API-1 (S2, schema too permissive)** — fixed. `model=mesh` with missing / non-array / empty `messages` now returns **400** before any model call, not a fabricated 200. Verified live.

**MOA-6 (admission only unit-tested)** — the tiering it depends on is now hardened: verified GGUF tensor-sum size gossiped via `parameter_count_b` drives admission/cap (not name parsing), `Unknown`/unsized ranks weakest, and self-fill uses distinct physical endpoints only (a single node can never fake a committee — degrades to single-model).

**MOA-8 (S1, shipped quality parity)** — the stale `RESULTS.md` parity block predated later fixes; still, I agree the shipped-path *quality* win at capable scale isn't independently proven here and remains open (needs a paid 24–32B shipped-path run, length-controlled single judge). Not claiming it's closed.

**On the 0.6B topology:** that model is an order of magnitude below the smallest in our data (8B), and our committed table shows an 8B floor (2–3× is null, 4× marginal, 6× wins). So the null you observed at 0.6B is the *expected* result, not a design signal — a fresh width sprint (aggregator qwen3-8b, 8B peers, shipped committee path) shows **6× diverse 8B beats the best member, 12W/2L p=0.013**; the fan-out cap is now tier-aware (6 small / 4 big) so small meshes aren't throttled, and the refine round is skipped where it measured null.

**aarch64 `check-release` + CUDA-12-default** — confirmed pre-existing and unrelated: `check-release` passes on this HEAD and the branch touches no release/packaging code. Flagging for a separate fix, not this PR.

Branch is merged current with main and CI is green. Full data/method in `evals/moa-openrouter/RESULTS.md`. Would appreciate another pass when you have a moment.

PR #1141, opened 2026-08-01
build: select iroh TLS backend explicitly (tls-aws-lc-rs) instead of defaults

themiguelamador:
## What

Declares the `iroh` dependency explicitly in the five crates that use it (`mesh-llm-protocol`, `mesh-llm-routing`, `mesh-client`, `mesh-llm-commands`, `mesh-llm-host-runtime`), replacing the bare `iroh = "1.0.0"` with `default-features = false` plus the same feature set iroh's defaults enable today — except `tls-ring`, which is swapped for `tls-aws-lc-rs`.

## Why

Cargo features are additive: a bare `iroh = "1.0.0"` enables iroh's `default` feature, which includes `tls-ring`, and no downstream consumer can subtract it. When both TLS backends are compiled in, iroh prefers ring — so any embedder of the mesh-llm SDK ships ring in its binary no matter what it does in its own manifest.

For deployments that need FIPS 140-3 alignment this is a hard blocker: ring is not FIPS-validated, while aws-lc-rs has a validated mode and is what rustls' FIPS story is built on. Today the only workaround is `[patch.crates-io]`-ing iroh itself to redefine its defaults, which every embedder has to carry independently.

With this change the resolved graph (normal deps) becomes:

```
├── iroh feature "fast-apple-datapath"
├── iroh feature "metrics"
├── iroh feature "portmapper"
└── iroh feature "tls-aws-lc-rs"
```

— no `default`, no `tls-ring`.

## Notes

- **Behavior is unchanged** — iroh supports both backends equivalently; this only pins which one is compiled in.
- **Dev-dependencies are untouched** (`mesh-llm-host-runtime`'s `iroh`/`iroh-relay` test deps) — they don't propagate to consumers.
- Verified with `cargo tree -e normal,features -i iroh` (output above) and `cargo check` on all five crates.
- If you'd prefer to keep ring as the project default, the alternative is passthrough `tls-ring`/`tls-aws-lc-rs` features plumbed through every crate between the SDK surface and iroh. That's a much larger diff; happy to rework this PR that way if you want backend choice rather than a swap.

## Summary by CodeRabbit

* **Improvements**
  * Improved connection performance on Apple devices.
  * Enabled port mapping and connection metrics.
  * Added AWS-LC-RS TLS support for secure connections.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4762,6 +4762,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "334c3c9833f7b2c573cceb9896ddc7aaeb58c8807cbb63211b24d1fe88bf866e"
 dependencies = [
  "aes-gcm",
+ "aws-lc-rs",
  "bytes",
  "derive_more",
  "enum-assoc",
@@ -8117,6 +8118,7 @@ version = "0.13.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d52efb639344a7c6adb8e62c6f3d2c19c001ff1b79a5041ba1c6ed42e19c6aa5"
 dependencies = [
+ "aws-lc-rs",
  "base64",
  "bytes",
  "futures-core",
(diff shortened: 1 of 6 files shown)


i386:
Thanks for your first contribution! I'll get this reviewed.

i386 (approved):
Approved — the explicit features match iroh 1.0.0's defaults, with only `tls-ring` replaced by `tls-aws-lc-rs`. No issues found.

i386:
@michaelneale Net effect: iroh will use AWS-LC-RS rather than `ring` for TLS. Normal mesh networking behavior should be unchanged; the benefit is that SDK embedders are no longer forced to include `ring`, making FIPS-oriented builds possible without patching iroh. This alone does not make the full mesh-llm distribution FIPS-certified—the full crypto stack and AWS-LC configuration still determine that.

PR #1143, opened 2026-08-02
fix(skippy): treat zero-temperature sampling as greedy

i386:
## Problem

OpenAI clients commonly use `temperature: 0` when they need reproducible output. Skippy still classified those requests as stochastic whenever sampling was enabled, so it rejected verified linear proposals even when no sampling option could change the winning token.

That meant a deterministic request behaved differently depending on whether the client omitted sampling settings or explicitly requested zero temperature.

## Fix

Admission now depends on whether the request can actually change greedy token selection. Zero temperature with neutral penalties and no logit bias is treated as deterministic. Non-zero temperature, grammar-constrained sampling, penalties, and logit bias remain excluded.

This keeps the safety boundary narrow: proposals are admitted only when verification is equivalent to greedy decoding, without special-casing a particular model or client.

## Validation

Skippy Server tests pass (353/353), including deterministic and stochastic sampling cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved linear proposal handling for generation requests.
  * Deterministic, zero-temperature sampling is now supported when no penalties or bias adjustments are applied.
  * Requests using stochastic sampling or logit-modifying settings continue to be handled appropriately.

### DIFF
--- a/crates/skippy-server/src/frontend/linear_proposal.rs
+++ b/crates/skippy-server/src/frontend/linear_proposal.rs
@@ -7,6 +7,7 @@ use std::{
 use anyhow::{Result, bail};
 use openai_frontend::{OpenAiError, OpenAiResult};
 use serde_json::json;
+use skippy_runtime::SamplingConfig;
 
 use crate::frontend::{
     NativeMtpVerifyWindowDecision, StageOpenAiBackend, TokenControl,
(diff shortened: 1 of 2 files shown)


ndizazzo (approved):

michaelneale (approved):
Low risk: local speculative-verification admission only, invisible across versions. Zero-temp treated as greedy is a correct, well-scoped change. LGTM.

PR #1144, opened 2026-08-02
fix(skippy): preserve native tool-call semantics

i386:
## Problem

A model could produce a valid native tool call, but parts of its meaning could be lost between chat rendering and verification:

- omitted message content could be collapsed into an empty value;
- tagged tool arguments depended on JSON object-field order;
- sampled verification could advance without carrying the grammar state.

For an OpenAI client using native tool calling, this could turn a valid tool call into malformed output, plain text, or a differently shaped message.

## Fix

The change preserves content presence exactly, accepts tagged argument fields in either object order, and carries grammar state through sampled verification.

These fixes live at the native chat-parser and verification boundaries because those layers know the model's actual output grammar. Repairing the result later with text heuristics would lose that authority and introduce model-specific ambiguity.

## Validation

OpenAI tests pass (154/154) and Skippy Server tests pass (353/353), including content-presence, argument-order, and grammar-state coverage.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved omitted, `null`, and empty message content in OpenAI-compatible chat messages.
  - Improved sampled verification, grammar-state handling, and serial-decoding consistency.
  - Accepted tool-call arguments in any order while validating duplicates and missing required fields.
  - Allowed valid grammar metadata during proposal processing.
  - Improved local generation reliability, cleanup, cancellation, and error handling.
- **Tests**
  - Expanded coverage for messaging, verification, tool calls, and local generation.
- **Documentation**
  - Documented the correctness-test model fixture and CI setup.


i386:
@michaelneale @ndizazzo we need to get this in and also push upstream

ndizazzo (approved):

i386:
https://github.com/ggml-org/llama.cpp/pull/26473
https://github.com/ggml-org/llama.cpp/pull/26472

ndizazzo:
@coderabbitai full review

michaelneale (approved):
this is good but need to get the build clean.

PR #1145, opened 2026-08-02
refactor(openai): remove text-form tool-call rescue

i386:
## Problem

The OpenAI frontend had a second set of text scanners that tried to recover tool calls from generated prose after the model's native chat parser had already classified the output.

That creates two competing interpretations of the same response. Text that merely discusses or quotes tool-call syntax can be mistaken for an invocation, while each additional model dialect needs another heuristic parser.

## Change

The model's native chat parser is now the only component allowed to identify a tool call. Text remains text unless that parser emits structured tool-call data.

Schema validation, retries, safe argument normalization, and thinking-block cleanup remain in place. Removing the rescue path avoids false invocations and gives every response one deterministic interpretation instead of extending a second parser indefinitely.

## Use case

An assistant can explain, quote, or debug JSON and XML tool-call examples without the frontend unexpectedly executing them.

## Validation

OpenAI, guardrail, and mixture-of-agents tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added support for removing hidden reasoning blocks from model responses.
  * Native tool calls and structured outputs are now validated directly.

* **Bug Fixes**
  * Prevented tool-like text from being incorrectly converted into tool calls.
  * Invalid responses now trigger validation and retry behavior consistently.
  * Simplified guardrail telemetry by removing rescue and parser-stage classifications.

* **Documentation**
  * Updated guardrail, testing, and telemetry documentation to reflect native tool-call validation.


ndizazzo (approved):

ndizazzo:
@coderabbitai full review

michaelneale:
this is ok - but will basically stop small models working I think - so might want a bit more attention on that - can you try with with say qwen 8B families for evidence?

michaelneale (changes requested):
I think may want to reconsider this for v small models, and get some tests/harnesses in place first. ie needs to be some cut off?

i386:
> I think may want to reconsider this for v small models, and get some tests/harnesses in place first. ie needs to be some cut off?

ahh now I understand what its for. OK, I will restore that now.

i386:
Agreed. Before deciding whether to restore the text-form rescue, I am adding and running tool-calling coverage across representative Qwen 8B GGUF families. The acceptance criterion is that the OpenAI endpoint emits native `tool_calls`, not template markup in `message.content`.

I will trace the llama.cpp chat-template/parser path at the same time. If native calls are absent because we are not applying the model parser infrastructure, the right fix is to wire that path correctly; a narrow compatibility shim remains reasonable while that lands. If the parser is working and smaller models still only emit text-form calls, we can make an explicit compatibility decision with evidence.

i386:
Investigation result: the native llama.cpp chat-template/parser path is already active in serving; it is not being bypassed.

I ran `scripts/qa-agent-tool-call-reliability.py` twice each against Qwen3-8B Q4_K_M and Qwen2.5-7B Instruct Q2_K, using the built PR branch and direct local OpenAI endpoint. Both runs passed 8/8 phases: native non-streaming tool call, tool-result continuation, streamed tool call, and streamed continuation across two attempts. The responses used `finish_reason: "tool_calls"`, `message.content: null`, and native OpenAI `message.tool_calls` (not Qwen `<tool_call>` text).

Based on this evidence, I do not recommend restoring the removed broad text-form rescue for these Qwen families. The separate serving-side emulation remains the compatibility path for templates that genuinely lack native tool support.

i386:
To make the distinction explicit: this PR removes a *last-resort, post-parse text rescue*. It does **not** remove the normal Qwen tool-call path.

The serving path first gives llama.cpp the conversation, tool definitions, and tool choice through `apply_chat_template_json`; after generation it calls `parse_chat_response_json`. That parser produces the native OpenAI `tool_calls` object that guardrails then validate. The rescue removed here ran only if that path did *not* yield a native call, by trying to reinterpret text in `message.content` as a tool call.

For the two representative Qwen families I tested, the normal path is working end-to-end: Qwen3-8B Q4_K_M and Qwen2.5-7B Instruct Q2_K both returned `finish_reason: "tool_calls"`, `content: null`, and structured `message.tool_calls`, including streaming and tool-result continuations (8/8 harness phases each). So the broad rescue is not carrying Qwen compatibility in these cases.

This does **not** mean every small model has native tool support. Models whose chat template genuinely has no native tool-call support still use the separate serving-side emulation path. Nor does it remove the need for guardrails to validate native calls, reject malformed arguments, and enforce policy. It only stops treating arbitrary text that the model/parser did not recognize as a call as though it were one.

My recommendation is therefore to retain the removal rather than add a size cutoff: model size is not the capability boundary; chat-template/parser support is. If there is a particular small-model family we need to support that emits only text-form calls despite a native template, we should add it to the harness and decide on a deliberately narrow compatibility shim for that family.

i386:
@michaelneale The key conclusion is simpler than my previous wording: **if the reason this guardrail was added was that Mesh was not correctly using llama.cpp chat-template parsing, then we no longer need the text-form tool-call rescue for models that ship a valid native tool-call template.**

That is what the Qwen runs show. With the current Mesh serving path, Qwen3-8B and Qwen2.5-7B receive the tools via their chat template and llama.cpp parses their generated tool call back into native OpenAI `tool_calls`. The output is already structured before the guardrail-rescue code would run. I verified tool invocation, streamed tool invocation, and the follow-up after a tool result; all passed without the rescue.

So, for those models, the previous rescue was compensating for the broken/missing parser integration—not a model limitation. Restoring it would reintroduce a fallback that is unnecessary for a correctly templated model and can turn unrecognised plain text into an action.

The qualification is only this: a model with *no valid native tool-call template/parser* needs a compatibility path. That is the separate serving-side emulation path, not this broad post-hoc guardrail rescue. So I think the right rule is: valid native template → native parsing → no text rescue; no native template → explicit emulation.

I therefore recommend retaining this removal; no model-size cutoff is warranted.

michaelneale:
@i386 is that something we can show a before/after proof of? As yeah if it was that then yes that could explain it

i386:
@michaelneale I reran this as the A/B you asked for, with the guardrail wrapper disabled in **both** cases. The result corrects my earlier conclusion.

| Build | Parser-related change | Text rescue / guardrails | Qwen3-8B Q4_K_M harness |
|---|---|---|---|
| `27a32e4fa` (immediately before `fix(skippy): preserve native tool-call semantics`) | absent | disabled (`OpenAiGuardrailsConfig::disabled_for_skippy()`) | **8/8 passed** |
| PR #1145 head `b816be873` | present | disabled; text rescue removed | **8/8 passed** |

Each run used the same direct local OpenAI endpoint and `qa-agent-tool-call-reliability.py`, two attempts each. The eight phases are forced native tool call, tool-result continuation, streamed tool call, and streamed continuation, twice. Both builds returned native OpenAI tool calls and completed the follow-up.

So Qwen3-8B does **not** demonstrate that the removed guardrail rescue is needed—and it also does **not** demonstrate that the parser change is what fixed Qwen3. The pre-change build already works for this model without guardrails. My earlier comment attributing Qwen success to the parser change was too strong; this A/B falsifies that attribution.

The valid conclusion from this model is narrower: removing the broad rescue does not regress Qwen3-8B native tool calling. To establish why Michael originally added the rescue, we need the specific model/output shape that previously required it (or another reproducer that emits text-form calls with guardrails disabled). That is the right target for a compatibility test; a model-size cutoff is not evidenced by Qwen3.

michaelneale:
🤖 *Posted by Mic's AI agent (CrocDundee)*

**Concern: this PR removes the only test coverage for the "text-form tool intent must not leak as content" invariant, and there's no fail-closed replacement.**

The direction of the PR is sound — a second text-scanning interpreter that can execute *quoted/explained* tool-call syntax is a real footgun, and the native chat parser (plus capability-gated `skippy-server/src/frontend/tool_emulation.rs`) is the right place for tool-call recognition to live.

But the MoA path loses a guarantee here. In `crates/mesh-mixture-of-agents/src/normalize.rs`, `heuristic_classify` no longer recognizes a text-form tool call, so a worker that emits tool intent as text (a small/older model, or a template mis-detected as native so emulation is off) now falls through to `OutputKind::Answer` and is returned to the agent as ordinary `content`. Goose / OpenCode / pi only act on real `tool_calls`, so the agent silently does nothing — no error, no retry, just a dropped turn.

That's precisely the failure mode pinned by `crates/mesh-mixture-of-agents/tests/sim_tool_call_text_not_passed_as_content.rs` (added off #566 review feedback), which this PR **deletes without replacing**. Native worker `tool_calls` are still fine — `backend.rs::extract_text_from_response` converts them to the KV envelope — so the gap is specifically text-form intent on the MoA path.

The execution-rescue can go, but the invariant it protected should survive: **text-form tool intent must become an explicit error / reducer-escalation, never a silent successful `content` answer.** Could we add a replacement MoA test asserting that fail-closed behavior before merge? Happy to write it if useful.

i386:
@michaelneale I traced this back through the original guardrail work and reran the historical model family rather than relying on Qwen3-8B.

**What the history says**

- The text rescue originated in your Forge-port commit on [#614](https://github.com/Mesh-LLM/mesh-llm/pull/614), later superseded by [#645](https://github.com/Mesh-LLM/mesh-llm/pull/645).
- #645’s measured Qwen cases were **Qwen3.5-0.8B UD-Q8_K_XL** and **Qwen3.5-4B Q4**, especially the 256-token hard-schema cases. Its report records malformed/truncated arguments; it does not identify a Qwen 8B parser-integration regression as the reason for rescue.

**Retest on this PR head**

I downloaded and ran Qwen3.5-0.8B UD-Q8_K_XL and Qwen3.5-4B UD-Q4_K_XL against the current PR head, with the same direct local OpenAI endpoint. For each model I ran guardrails **disabled** and global guardrails **enforce** (verified from `/api/status`).

| Model | Disabled | Enforce | Result |
|---|---:|---:|---|
| Qwen3.5 0.8B UD-Q8_K_XL | 6/9 | 6/9 | all non-streaming native calls + tool-result continuations pass; all streamed tool-call deltas fail |
| Qwen3.5 4B UD-Q4_K_XL | 6/9 | 6/9 | same result |

I also forced a nested tool schema at `max_tokens: 256`, the historical hard-case shape. On both models and in both modes, the first response returned valid native OpenAI arguments; later requests failed with llama.cpp native PEG parse `502` errors *before* a response reached guardrails. Enforcing guardrails did not change that outcome. Streaming is explicitly pass-through, so it cannot repair the streamed-call failure either.

**Conclusion**

For the original Qwen3.5 family, I cannot reproduce a case where the broad text rescue provides a benefit. The observed failures are below or outside that layer: native parser failure and streaming behavior. This supports retaining the rescue removal for Qwen, but it does **not** prove no model ever needs a compatibility shim. The separate GLM tagged-call history is the concrete remaining candidate. If there is a specific historical text-form output that rescue successfully converted into a valid call, that exact model/prompt/response is what we should add as a narrow regression fixture before retaining a shim.

i386:
@michaelneale I agree this is a separate concern from the parser/guardrail investigation.

The Qwen A/B evidence supports removing broad text-to-executable-tool-call rescue: we should not scan arbitrary assistant text and synthesize `tool_calls`. But this PR also removes the MoA safety invariant from #566: when a worker returns plain text that clearly expresses intent to invoke a declared tool, treating it as a successful `Answer` can leak that text to the agent as `content`. The agent neither invokes a tool nor receives a failure, so the turn is silently dropped.

I think the safe replacement is a narrow, tool-context-aware MoA guard—not execution rescue:

- detect clear text-form intent for one of the request’s declared tools;
- classify it as unusable/uncertain and escalate to the reducer;
- if no valid native tool call emerges, return an explicit MoA error;
- add a regression test that this text never becomes a successful content response.

That preserves the fail-closed invariant without reconstructing or executing a call from text. The deleted inline-JSON success case should also fail closed unless the call arrived through the native tool-call channel.

How would you like to proceed—would you prefer to take that narrow MoA guard/test, or should I implement it here?

michaelneale:
I think a more narrow MoA guard makes sense here ... I think... not sure what we lose in process of that though. One thing to try is to try a small model with almost any agent harness - that is only way you see this (basically none of them work even a little bit), if it works reasonably now (which would be obvious), then that is good enough to go, and we can look at MoA as a follow on?

i386:
Ran the requested OpenCode tool-loop smoke test on the current PR head `7aa1c51f859edc8da022eaa95ef04bff1904133f` (clean build) with the two models discussed here:

| Model | OpenCode tool use | Result |
| --- | --- | --- |
| Qwen3.5 0.8B UD-Q8_K_XL | Yes — multiple filesystem `bash` / `read` calls completed | **Fail:** stalled after tool results; the smoke fixture remained unimplemented. |
| Qwen3.5 4B UD-Q4_K_XL | Yes — filesystem `bash` call (`ls -la`) completed | **Fail:** same post-tool-call stall; the smoke fixture remained unimplemented. |

The harness's direct OpenAI tool probe also returned `finish_reason: "length"` with no structured `tool_calls` for both models (the probe is capped at 8 output tokens).

This was a real local Metal run through the OpenAI surface and OpenCode 1.16.2. I disabled only the long-context soak portion (`OPENCODE_SMOKE_LONG_PROMPT_CHARS=0`); the tool-call loop itself ran. The traces show both models can emit a tool call, but neither completed the continuation after the tool result, so I would not call the agent-harness path working yet.

i386:
Correction to my prior test report: the observed behavior is **slow continuation, not a confirmed deadlock**.

The capture proves OpenCode sent the post-tool request (assistant `tool_calls` plus `role: tool` result) to mesh. I replayed that exact 4B continuation directly against mesh, bypassing the capture proxy. Mesh emitted the assistant stream role immediately, then emitted the next native `bash` tool call and completed with `finish_reason: tool_calls` / `[DONE]` about **76 seconds later** (roughly 11k prompt tokens).

I had interrupted the original OpenCode run at approximately that latency threshold, so the earlier "stalled" conclusion was premature. This shows the OpenCode → tool result → mesh → next tool-call path does work for the 4B replay; it is simply much slower than the bounded wait I used. I have not yet rerun the entire multi-step smoke to completion under a longer time budget.

michaelneale (approved):

PR #1146, opened 2026-08-02
fix(skippy): discover retained exact KV states

i386:
## Problem

Exact KV states can be retained at arbitrary token counts, but discovery only considered grid-aligned prefix lengths. A valid state recorded at a non-grid length could remain in the cache indefinitely without ever being considered for a longer compatible prompt.

For example, repeated requests sharing a 197-token prefix could miss an exact retained state if discovery only probed fixed stride boundaries.

## Fix

Exact-state discovery now considers both the normal grid candidates and a bounded set of retained token counts. Candidate identity and prompt compatibility are still verified before restoration.

Using the existing bounded retained index fixes the blind spot without scanning the entire cache or weakening exact-match correctness.

## Validation

KV cache tests pass (32/32) and Skippy Server tests pass (353/353), including non-grid retained-prefix coverage.

## Summary by CodeRabbit

* **New Features**
  * Improved exact-state lookups to recognize additional cached token-count prefixes beyond the standard candidate values.
  * Added support for retaining and matching page identities at non-grid prefix lengths.

* **Bug Fixes**
  * Removed duplicate token-count candidates and ensured lookup results are processed in descending order for more reliable matching.

### DIFF
--- a/crates/skippy-cache/src/exact_state.rs
+++ b/crates/skippy-cache/src/exact_state.rs
@@ -81,6 +81,18 @@ impl<E: Clone> ExactStateCache<E> {
         })
     }
 
+    pub fn token_counts_at_most(&self, max_token_count: u64) -> Vec<u64> {
+        let mut token_counts = self
+            .entries
+            .values()
+            .map(|entry| entry.token_count)
+            .filter(|token_count| *token_count <= max_token_count)
+            .collect::<Vec<_>>();
+        token_counts.sort_unstable_by(|left, right| right.cmp(left));
+        token_counts.dedup();
+        token_counts
+    }
+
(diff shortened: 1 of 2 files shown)


i386:
@michaelneale @ndizazzo another bug I think we need to get into the release asap

ndizazzo (approved):

michaelneale (approved):
Low risk: fully internal to the skippy KV cache. Widens retained-state candidate discovery while preserving the same identity/compatibility verification, so the exact-match safety boundary is intact. Bounded and guarded by is_exact_state(). LGTM.

ndizazzo:
@coderabbitai full review

PR #1147, opened 2026-08-02
feat(skippy): expose the loaded tokenizer capability

i386:
## Use case

An application embedding the local OpenAI server may need to count tokens, split documents on the model's actual token boundaries, or inspect token pieces. A benchmark runner may also need exact token counts from the serving tokenizer so throughput and latency measurements use the same tokenization as generation.

Loading a second tokenizer for that work wastes memory and can silently diverge from the tokenizer used by the model.

## Change

The stage-zero server exposes a bounded, read-only tokenizer capability backed by the tokenizer already loaded with the model. Callers must provide the expected model and source-file identity, and mismatches fail closed.

The capability is local to the serving boundary and is not exposed over remote stage transport. Reusing the loaded tokenizer guarantees that token accounting and generation use the same vocabulary without introducing another model load.

## Validation

Skippy Server tests pass (353/353), including tokenizer identity, input/output bounds, token-piece alignment, and generation-state mutation safety.

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI-compatible `/v1/tokenize` support with token IDs and optional token pieces.
  * Added tokenizer identity details to runtime status responses when available.
  * Added request validation, input and output limits, and structured error responses.

* **Bug Fixes**
  * Improved routing for large tokenization requests without altering request content.
  * Tokenization requests now preserve model identity and avoid generation-specific budget limits.
  * Runtime startup failures are now reported instead of being silently retained.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7426,6 +7426,7 @@ dependencies = [
  "prost-build",
  "protoc-bin-vendored",
  "serde",
+ "serde_json",
 ]
 
 [[package]]
@@ -7484,6 +7485,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tonic",
+ "tower",
 ]
 
 [[package]]
--- a/Justfile
+++ b/Justfile
@@ -508,7 +508,13 @@ test-all:
         --exclude mesh-llm-ffi \
         --exclude mesh-llm-nodejs \
         --exclude skippy-ffi \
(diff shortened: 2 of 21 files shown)


ndizazzo:
@coderabbitai full review

michaelneale (approved):
makes sense - clippy failures to fix up before merge.

This will lock up generation (which I guess is ok as it is like generation)

PR #1148, opened 2026-08-02
feat(mesh): add pinned model-only OpenAI serving

i386:
## Use case

A benchmark runner often needs to launch one pinned local model behind an OpenAI-compatible endpoint, run a workload, collect measurements, and tear the server down. Starting peer networking, discovery, management APIs, or a web console adds unrelated behavior and makes the experiment harder to reproduce.

The same minimal launch path is useful for local development, CI, and desktop applications that only need model serving.

## Change

`serve --local-model-only` starts only the pinned model-serving path. It requires absolute, non-symlink model and projector paths and rejects options that would enable network-management surfaces.

Making this an explicit mode provides a predictable, fail-closed startup contract. It avoids relying on a fragile combination of unrelated flags while leaving existing Mesh serving modes unchanged.

## Validation

CLI tests pass (74/74) and host-runtime tests pass (1,866 passed; 8 ignored), including rejected network, management, and invalid-path configurations.

## Summary by CodeRabbit

* **New Features**
  * Added a local-model-only serving mode with a single local model through the OpenAI-compatible API.
  * Added readiness, status, shutdown, and startup error reporting for local serving.
  * Added validation for model paths, capacity limits, and incompatible options.

* **Bug Fixes**
  * Improved removal of nested thinking blocks.
  * Corrected reconstruction of filtered tool-call responses.

* **Documentation**
  * Documented local-only serving requirements and updated telemetry guidance.


ndizazzo:
@coderabbitai full review

michaelneale (approved):

PR #1150, opened 2026-08-02
fix(openai): cancel timed-out generation before releasing its lane

i386:
## Problem

When an OpenAI request timed out, dropping the async request future did not stop the blocking generation worker. The request could return and release its concurrency lane while that worker was still using the model runtime.

A subsequent request could then acquire the same lane and overlap with stale work from the timed-out request, breaking the lane's isolation and making timeout behavior dependent on worker timing.

## Fix

Cancellation is propagated through the OpenAI and Skippy layers, and the timed-out worker is allowed to stop before its lane permit is released.

This is necessary because a blocking worker continues after its async wrapper is dropped; releasing the lane immediately would only hide the work, not cancel it. Tying lane release to confirmed worker termination preserves the existing concurrency contract.

## Validation

OpenAI tests pass (154/154) and Skippy Server tests pass (353/353), including the single-lane timeout regression.

## Summary by CodeRabbit

* **New Features**
  * Added cancellation support for chat, completion, response, and streaming generation requests.
  * Requests can now stop promptly when cancelled, timed out, or abandoned.
  * Cancellation context is preserved through retries, hooks, guardrails, and local generation.

* **Bug Fixes**
  * Prevented cancelled requests from holding generation capacity longer than necessary.
  * Improved cancellation handling during token generation and proposal execution.
  * Improved session admission and batch-size handling for generation workloads.

* **Tests**
  * Added coverage for cancellation during active and non-streaming generation.

### DIFF
--- a/crates/openai-frontend/src/backend.rs
+++ b/crates/openai-frontend/src/backend.rs
@@ -8,6 +8,7 @@ use std::{
 
 use async_trait::async_trait;
 use futures_core::Stream;
+use tokio::sync::Notify;
 
 use crate::{
     chat::{ChatCompletionChunk, ChatCompletionRequest, ChatCompletionResponse},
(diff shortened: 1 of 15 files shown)


i386:
@michaelneale @ndizazzo impactful enough bug that I'd like to do a release after we review and merge

michaelneale (approved):
Oh that is a nasty one. I bet it bites often and mysteriously.

ndizazzo (approved):
Checks out as clean to me

ndizazzo:
@coderabbitai full review

PR #1152, opened 2026-08-02
Propagate stable agent session identity from the OpenAI boundary

i386:
## Summary

- normalize an optional stable agent-session identity at the OpenAI frontend
- accept Responses `conversation.id` and one operator-configured trusted header
- reject conflicting identities and ignore headers unless explicitly trusted
- propagate the identity through Skippy generation receipts without conflating it with runtime session IDs or prompt-cache keys

## Why

Long-lived consumers need a neutral way to correlate multiple OpenAI requests from the same agent session. This adds the transport contract only; downstream products remain responsible for authentication, persistence, and policy.

## Validation

- `cargo check --workspace --all-targets`
- `cargo clippy -p openai-frontend -p skippy-server --all-targets -- -D warnings`
- `cargo test -p openai-frontend -p skippy-server --quiet`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying agent sessions across chat, completion, and Responses requests.
  * Configurable request headers can provide session identity, with validation and conflict detection.
  * Session identifiers now appear in generation receipts for improved request tracking.

* **Bug Fixes**
  * Invalid, empty, oversized, or conflicting session identities are rejected with clear request errors.

* **Tests**
  * Added coverage for session identity extraction, propagation, validation, and receipt handling.


michaelneale:
Oh this is interesting but not sure if the "why" - is it if they are from same session from routing or kv optimisations?

i386:
Yeah, it's basically so that proxy sitting in front of the open AI gateway can't tell whether or not to request has been processed by the same agent or person. I need it for stuff.

michaelneale (approved):
Low risk: opt-in (defaults off), identity is opaque/length-bounded/control-char-rejected and explicitly kept out of prompt-cache keys and runtime session IDs, so no cache-poisoning or KV-collision risk. Client spoofing blocked via extra.remove() before set. GenerationReceipt field is not serialized on any wire. Clean, well-isolated. LGTM.

PR #1153, opened 2026-08-02
Honor configured reasoning defaults for agent requests

i386:
## Summary

- apply configured `reasoning_enabled` when a chat request omits an explicit reasoning control
- fall back to configured `reasoning_budget` when `reasoning_enabled` is auto or omitted
- preserve explicit request-level reasoning as the highest-precedence value

## Why

The configuration surface already accepts and resolves these defaults, but the chat-template path previously ignored them. Models whose native template enables reasoning could therefore reason even when the serving configuration disabled it.

## Validation

- `cargo fmt --all --check`
- `cargo test -p skippy-server --lib` (338 passed)
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Requests now automatically apply configured defaults for thinking and reasoning when these options are omitted.
  * Reasoning budgets and explicit enable/disable settings are recognized when determining whether thinking is enabled.
  * Explicit request settings take precedence over configured defaults.

### DIFF
--- a/crates/skippy-server/src/frontend/request.rs
+++ b/crates/skippy-server/src/frontend/request.rs
@@ -306,9 +306,11 @@ fn default_reasoning_enabled(value: Option<EmbeddedReasoningEnabled>) -> Option<
 fn default_reasoning_budget_enabled(value: Option<EmbeddedReasoningBudget>) -> Option<bool> {
     match value {
         Some(EmbeddedReasoningBudget::Tokens(0)) => Some(false),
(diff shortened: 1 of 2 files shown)


ndizazzo (approved):

michaelneale (approved):
Low risk: correct precedence (explicit request-level reasoning still wins), config-gated, and a no-op when unconfigured. Honoring a configured reasoning default is the intended fix. LGTM.

PR #1154, opened 2026-08-02
fix: speed up dependency-heavy container rebuilds

ndizazzo:
## Summary

- add stable architecture/toolchain-scoped BuildKit caches for apt, npm, pnpm, rustup, Cargo, Cargo target output, downloads, and sccache
- enable required sccache for the full llama.cpp compilation still present in the client, Fly, and WAN-lab container paths
- make sccache downloads atomic, checksum-verified, and self-healing
- replace `git am --no-verify` with a Bookworm-compatible hook override
- document the forced-recompile benchmark and add regression coverage

## Measured impact

The same patched llama.cpp CPU/static source tree was compiled from clean build directories on Linux ARM64:

| Sample | Compile time |
|---|---:|
| Faster no-sccache control | 121.2s |
| Cache population, 273 misses | 125.8s |
| Warm sccache, 273 hits | 3.55s |

The warm compiler-cache path is 97.1% faster than the control, with zero cache errors.

## Release-path audit

- `just check-release` passes the release-target consistency gate.
- Native runtime release jobs compile through `prepare-native-runtime-input` and `package-native-runtime.sh`.
- Static ABI release jobs directly run `prepare-llama.sh` and `build-llama.sh`.
- Native SDK release jobs consume the typed static ABI input and package through `prepare-native-sdk-input`.
- The Bookworm Git fix is therefore exercised by canonical Linux release compilation.
- The three Dockerfiles changed here are auxiliary container paths; canonical published package/image assembly remains in `mesh-packaging`.

## Validation

- `just check-release`
- 115 focused cache and release-contract tests
- exact pinned llama preparation: 47/47 patches applied, patched SHA `e43d604fb427e769ce3c09ebb462743e456d0c99`
- ShellCheck, shell syntax, and actionlint
- BuildKit checks for all three Dockerfiles on amd64 and arm64
- forced compile benchmark with 273/273 warm sccache hits

## Summary by CodeRabbit

* **Performance**
  * Improved container build times with architecture-aware caching for dependencies, toolchains, and build artifacts.
  * Added compiler-cache support for faster repeated llama and Rust builds.

* **Documentation**
  * Added ARM64 container build-cache benchmarks and performance comparisons.

* **Reliability**
  * Added verified, cache-aware installation for the compiler cache tool.
  * Improved patch application compatibility across Git versions.

* **Tests**
  * Added coverage for cache validation, recovery, reuse, and Git hook handling.

### DIFF
--- a/docker/Dockerfile.client
+++ b/docker/Dockerfile.client
@@ -15,24 +15,34 @@
 # syntax=docker/dockerfile:1.7
 
 FROM node:24-alpine AS ui-builder
-RUN npm install -g pnpm
+ARG TARGETARCH
+RUN --mount=type=cache,id=mesh-llm-npm-alpine-node24-${TARGETARCH},target=/root/.npm,sharing=locked \
+    npm install -g pnpm
 WORKDIR /app
 COPY crates/mesh-llm-ui/package.json crates/mesh-llm-ui/pnpm-lock.yaml crates/mesh-llm-ui/pnpm-workspace.yaml ./
-RUN --mount=type=cache,id=pnpm-store,target=/root/.local/share/pnpm/store \
+RUN --mount=type=cache,id=mesh-llm-pnpm-store-alpine-node24-${TARGETARCH},target=/root/.local/share/pnpm/store,sharing=locked \
     pnpm install --frozen-lockfile
 COPY crates/mesh-llm-ui/ ./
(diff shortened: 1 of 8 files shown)


i386:
still my beating heart

i386 (approved):

PR #1155, opened 2026-08-02
feat(mesh): load deadline-safe native serving plugins

i386:
## Use case

An inference product should be able to attach a native proposal/state implementation to the normal Mesh OpenAI serving binary without wrapping Mesh, owning another model runtime, or blocking Skippy's decode loop.

## Dependencies

- #1149 supplies the target-authoritative generation lifecycle and proposal outcome hooks.
- #1152 supplies the authenticated caller session identity at the OpenAI boundary.

This draft is stacked on #1149 and merges #1152 so the complete contract compiles and tests together. The session-identity portion will disappear from this diff when #1152 lands.

## Change

- Add a small versioned C-compatible `mesh-native-serving-plugin-api` function-table ABI with safe Rust-side types.
- Add `mesh-native-serving-plugin-host`, loaded only when explicit hidden local-serving flags provide the library, config, state directory, and deadline.
- Map Mesh's authoritative tokenizer, authenticated agent session, generation begin/commit/finish/abort, proposal submit/poll/cancel, and proposal outcomes across the ABI.
- Run plugin work on a bounded worker queue. Mesh owns the absolute proposal deadline, stops waiting at that deadline, and cancels work without allowing a blocking plugin call to hold the decode path.
- Validate ABI version, table size, status values, correlation identities, output lengths, token buffers, and proposal bounds fail-closed.
- Remove the fixed 256-token proposal ceiling. Each query carries the exact capacity allowed by the current runtime/context/verifier boundary, and the host allocates the return buffer to that capacity.
- Keep the normal Mesh executable and `--local-model-only` startup path. There is no alternate launcher or compatibility entrypoint.

## Runtime contract

Without plugin flags, behavior is unchanged. With plugin flags, all flags are required together and local-model-only serving creates the native hook factory after loading the plugin. Skippy remains authoritative for model execution, tokenizer identity, KV cache, scheduling, verification, proposal admission, and outcomes.

## Validation

- Focused compile passes for the ABI, host, CLI, Mesh runtime, and Skippy.
- 507 focused tests pass: 2 ABI, 4 host/loader, 149 OpenAI frontend, and 352 Skippy.
- The hard-deadline test blocks inside plugin proposal submission for 250 ms; a 5 ms query still returns before the decode-path bound and cancellation is delivered later.
- Strict Clippy passes for the ABI, host, CLI, OpenAI frontend, and Skippy with `-D warnings`.
- Workspace publish-order, CI crate-list, and all-Rust-crate coverage consistency checks pass.

The broader host runtime still reports 28 pre-existing `unfulfilled_lint_expectations` warnings in untouched compatibility code; the focused strict-Clippy set is clean.


i386:
@coderabbitai review

Please perform a full review of the current PR head, including the native serving plugin ABI, deadline/cancellation behavior, and the restored dynamic proposal-width contract.

michaelneale (approved):

PR #1164, opened 2026-08-03
fix: keep Windows per-GPU VRAM aligned with adapter order

### LINKED ISSUE #1163: Windows GPU fallback misassigns per-GPU VRAM
## Summary

In the Windows hardware fallback path, adapters that report no VRAM are dropped
from the per GPU VRAM vector, but the GPU name list keeps its original indices.
The two lists then no longer line up, so VRAM is attributed to the wrong device.

On a machine with a virtual display adapter listed before the real GPU, the
virtual adapter is credited with the real GPU's VRAM and the real GPU reports
unknown.

## Affected versions

- 0.75.0-rc1

## Environment

- Windows 10 (10.0.19044), x86_64
- Two adapters reported by WMI, the first with a null `AdapterRAM`:

```
[{"Name":"Meta Virtual Monitor","AdapterRAM":null},{"Name":"AMD Radeon RX 9070 XT","AdapterRAM":4293918720}]
```

## Steps to reproduce

1. Use a Windows machine where `Get-CimInstance Win32_VideoController` lists a
   virtual or headless adapter with a null `AdapterRAM` before the real GPU.
2. Run `mesh-llm gpus`.

## Expected

VRAM is reported against the adapter it belongs to. The real GPU shows its
VRAM, the virtual adapter shows unknown.

## Actual

```
GPU 0
  Name: Meta Virtual Monitor
  Stable ID: index:0
  Backend device: unavailable (hardware-visible only; embedded runtime did not report a selectable device)
  VRAM: 4 GB

GPU 1
  Name: AMD Radeon RX 9070 XT
  Stable ID: rocm1
  Backend device: ROCm1
  VRAM: unknown
```

## Analysis

In `crates/mesh-llm-system/src/hardware/mod.rs`, the Windows branch builds the
per GPU VRAM vector from `read_windows_video_controllers()` with

```rust
let per_gpu: Vec<u64> = windows_gpus
    .iter()
    .map(|(_, ram)| *ram)
    .filter(|ram| *ram > 0)
    .collect();
```

The filter compacts the vector, so `gpu_vram[0]` becomes the second adapter's
VRAM. `gpu_name` and `gpu_count` are built from the unfiltered list, and
`hydrate_gpu_facts_with_identities()` then reads `survey.gpu_vram.get(index)`
using the name index. Index 0 therefore picks up the wrong value and the last
device gets nothing.

Since VRAM feeds capacity planning and mesh advertisement, a machine with a
virtual adapter can advertise capacity against the wrong device.

## Suggested fix

Keep the per GPU VRAM vector the same length as the adapter list, using 0 or
`None` for adapters that report no VRAM, instead of filtering entries out. Sum
for totals separately.

ApexDevelopment:
Fixes #1163.

## Problem

In the Windows hardware fallback, the per-GPU VRAM vector is built with

```rust
.filter(|ram| *ram > 0)
```

which compacts the vector, while `gpu_name` and `gpu_count` are built from the
unfiltered adapter list. `hydrate_gpu_facts_with_identities` then reads
`survey.gpu_vram.get(index)` using the name index, so the two no longer line up
and VRAM lands on the wrong device.

Virtual and headless display adapters commonly report no VRAM. On a machine
where one is enumerated first, `mesh-llm gpus` credits the virtual adapter with
the real GPU's VRAM and reports the real GPU as unknown:

```
GPU 0
  Name: Meta Virtual Monitor
  VRAM: 4 GB

GPU 1
  Name: AMD Radeon RX 9070 XT
  VRAM: unknown
```

VRAM feeds capacity planning and mesh advertisement, so this is not only a
display problem.

## Change

Extract the per-GPU VRAM construction into `windows_per_gpu_vram()` and keep a
zero entry for adapters that report no VRAM, so the vector stays index-aligned
with the adapter list.

Totals are unaffected, since zero entries do not change the sum. The existing
`total > 0` guard still selects the system-RAM fallback when no adapter reports
any VRAM.

## Tests

Two tests added in `crates/mesh-llm-system/src/hardware/tests.rs`:

- `test_windows_per_gpu_vram_keeps_zero_vram_adapters_aligned` covers the
  vector shape directly.
- `test_windows_zero_vram_adapter_does_not_shift_vram_onto_wrong_gpu` runs the
  result through `hydrate_gpu_facts_with_identities` and asserts each adapter
  keeps its own VRAM.

Both fail on the previous behavior. The second one reproduces the reported
symptom exactly, asserting `4293918720` where `0` is expected on GPU 0.

Verified with `cargo test -p mesh-llm-system --features dynamic-native-runtime
--lib hardware::` on Windows: 72 passed, 0 failed. The
`dynamic-native-runtime` feature was needed only to build the crate locally,
since `skippy-ffi` otherwise requires prepared llama.cpp static archives that
`build.rs` cannot produce on Windows. It does not affect the code changed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows GPU detection to preserve adapters with zero reported VRAM.
  * Ensured GPU names and VRAM values remain correctly matched across all detected adapters.

### DIFF
--- a/crates/mesh-llm-system/src/hardware/mod.rs
+++ b/crates/mesh-llm-system/src/hardware/mod.rs
@@ -235,6 +235,17 @@ fn read_windows_total_ram_bytes() -> Option<u64> {
     parse_windows_total_physical_memory(&output)
 }
 
+/// Per-GPU VRAM in adapter order.
+///
+/// Adapters that report no VRAM (virtual and headless displays commonly report
+/// none) keep a zero entry so this stays index-aligned with the adapter name
+/// list that `hydrate_gpu_facts_with_identities` indexes into. Dropping those
+/// entries shifts every later adapter's VRAM onto the wrong device.
+#[cfg(any(target_os = "windows", test))]
+fn windows_per_gpu_vram(controllers: &[(String, u64)]) -> Vec<u64> {
+    controllers.iter().map(|(_, ram)| *ram).collect()
+}
+
 #[cfg(target_os = "windows")]
 fn read_windows_video_controllers() -> Vec<(String, u64)> {
     let Some(output) = powershell_output(
@@ -677,11 +688,7 @@ impl Collector for DefaultCollector {
                         survey.vram_bytes = total + (ram_offload as f64 * 0.90) as u64;
                     }
                 } else {
-                    let per_gpu: Vec<u64> = windows_gpus
-                        .iter()
-                        .map(|(_, ram)| *ram)
-                        .filter(|ram| *ram > 0)
-                        .collect();
+                    let per_gpu: Vec<u64> = windows_per_gpu_vram(&windows_gpus);
                     let total: u64 = per_gpu.iter().sum();
                     if total > 0 {
                         survey.gpu_vram = per_gpu;
--- a/crates/mesh-llm-system/src/hardware/mod.rs
+++ b/crates/mesh-llm-system/src/hardware/mod.rs
@@ -690,8 +690,8 @@ impl Collector for DefaultCollector {
                 } else {
                     let per_gpu: Vec<u64> = windows_per_gpu_vram(&windows_gpus);
                     let total: u64 = per_gpu.iter().sum();
+                    survey.gpu_vram = per_gpu;
                     if total > 0 {
-                        survey.gpu_vram = per_gpu;
                         let ram_offload = system_ram.saturating_sub(total);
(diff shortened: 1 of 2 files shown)


michaelneale (approved):
thanks @ApexDevelopment good catch. Hard to hand test so I hope all good - pre-approving @ndizazzo as I think low risk

ndizazzo (approved):

PR #1170, opened 2026-08-04
fix: ignore stale native runtime cache manifests

### LINKED ISSUE #1162: runtime install aborts on pre-0.75 cached manifests
## Summary

0.75.0-rc1 requires native runtime manifests to declare per file checksums.
Runtimes installed by earlier versions do not have them. If any such runtime is
present in the cache directory, `mesh-llm runtime install` aborts, even when the
runtime being installed is valid and does declare checksums.

This affects anyone upgrading to 0.75.0 with an existing runtime cache, which is
the normal upgrade path.

## Affected versions

- 0.75.0-rc1

## Environment

- Windows 10 (10.0.19044), x86_64
- Cache directory contained runtimes installed by 0.73.1 and 0.74.0

## Steps to reproduce

1. On 0.73.1 or 0.74.0, install any native runtime so the cache is populated.
2. Upgrade to 0.75.0-rc1.
3. Run `mesh-llm runtime install --bundle-dir <0.75.0-rc1 bundle native-runtimes dir>`.

## Expected

The 0.75.0-rc1 runtime installs. Older cached runtimes may be ignored or
reported as unusable, but should not block the install.

## Actual

```
Native runtime install failed
   Reason: native runtime artifact meshllm-native-runtime-windows-x86_64-vulkan does not declare file checksums
   Try: mesh-llm runtime list --available
```

`mesh-llm runtime list` shows the same problem as warnings, then reports no
runtimes at all:

```
warning: skipping malformed native runtime ...\0.73.1\meshllm-native-runtime-windows-x86_64-vulkan: native runtime artifact ... does not declare file checksums
warning: skipping malformed native runtime ...\0.74.0\meshllm-native-runtime-windows-x86_64-rocm: ...
warning: skipping malformed native runtime ...\0.74.0\meshllm-native-runtime-windows-x86_64-vulkan: ...
No local native runtimes found
```

The error message names the artifact being installed, not the stale cached one,
which makes it look like the new bundle is at fault.

## Confirmation

Installing the identical bundle into an empty cache directory with
`--cache-dir <fresh empty dir>` succeeds immediately. Moving the `0.73.1` and
`0.74.0` directories out of the default cache also makes the install succeed.
So the new bundle is fine and the stale entries are what abort the operation.

## Workaround

Move or delete pre 0.75 runtime directories out of the cache before installing.

## Suggested fix

Treat a cached runtime with a pre checksum manifest as skippable rather than
fatal, so a valid install can proceed. If the abort is intentional, the error
should name the offending cached path and suggest `mesh-llm runtime prune`.

ndizazzo:
## Summary

Fixes native runtime installation failing when the cache contains runtime manifests from older MeshLLM releases that predate per-file checksums.

## Root cause

Resolver candidate discovery strictly validated every cached runtime before applying the requested MeshLLM version filter. A legacy cached manifest therefore aborted installation before a valid explicit bundle could be selected.

## Changes

- Add strict, version-scoped cache enumeration for resolver candidate discovery.
- Keep full-cache inventory and explicit/current runtime checksum validation strict.
- Add cache, resolver, and end-to-end installer regressions covering legacy manifests.
- Mark the compatibility boundary for future cleanup once pre-checksum caches are no longer part of supported upgrade paths.

Fixes #1162

## Validation

- `cargo fmt --all --check`
- `cargo test -p mesh-llm-native-runtime --lib`
- `cargo test -p mesh-llm-runtime-install --lib`
- `cargo check -p mesh-llm-native-runtime`
- `cargo check -p mesh-llm-runtime-install`
- `cargo check -p mesh-llm`
- `cargo clippy -p mesh-llm-native-runtime --all-targets -- -D warnings`
- `cargo clippy -p mesh-llm-runtime-install --all-targets -- -D warnings`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime selection to use only installations compatible with the requested MeshLLM version.
  * Legacy or incompatible cached runtimes no longer interfere with runtime resolution or bundle installation.
  * Added validation and consistent ordering when listing installed runtimes.
  * Files stored at version paths are now ignored when they are not valid runtime entries.
* **Tests**
  * Added coverage for version-specific runtime discovery and installation with stale cache entries.

### DIFF
--- a/crates/mesh-llm-native-runtime/src/cache.rs
+++ b/crates/mesh-llm-native-runtime/src/cache.rs
@@ -122,7 +122,10 @@ impl NativeRuntimeCache {
     /// TODO(issue #1162): Remove this resolver-specific compatibility boundary
     /// once the oldest supported upgrade path no longer contains pre-checksum
     /// runtime caches, if full-cache resolver enumeration is safe again.
-    pub fn installed_for_version(&self, mesh_version: &str) -> Result<Vec<InstalledNativeRuntime>> {
+    pub(crate) fn installed_for_version(
+        &self,
+        mesh_version: &str,
+    ) -> Result<Vec<InstalledNativeRuntime>> {
         installed_in_version_dir(&self.root.join(mesh_version))
     }
 
--- a/crates/mesh-llm-native-runtime/src/cache.rs
+++ b/crates/mesh-llm-native-runtime/src/cache.rs
@@ -108,21 +108,22 @@ impl NativeRuntimeCache {
             if !version_entry.file_type()?.is_dir() {
                 continue;
             }
-            for runtime_entry in fs::read_dir(version_entry.path())? {
-                let runtime_entry = runtime_entry?;
-                if !runtime_entry.file_type()?.is_dir() {
-                    continue;
-                }
-                if let Some(runtime) = installed_runtime_from_dir(&runtime_entry.path())? {
-                    installed.push(runtime);
-                }
-            }
+            installed.extend(installed_in_version_dir(&version_entry.path())?);
         }
         installed.sort_by(|left, right| {
             (&left.mesh_version, &left.native_runtime_id)
                 .cmp(&(&right.mesh_version, &right.native_runtime_id))
         });
         Ok(installed)
+    }
+
+    /// Returns strictly validated runtimes for one MeshLLM version.
+    ///
+    /// TODO(issue #1162): Remove this resolver-specific compatibility boundary
+    /// once the oldest supported upgrade path no longer contains pre-checksum
+    /// runtime caches, if full-cache resolver enumeration is safe again.
+    pub fn installed_for_version(&self, mesh_version: &str) -> Result<Vec<InstalledNativeRuntime>> {
+        installed_in_version_dir(&self.root.join(mesh_version))
     }
 
     pub fn find_installed(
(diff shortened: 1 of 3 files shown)


crates/mesh-llm-native-runtime/src/cache.rs
```diff
@@ -242,6 +246,26 @@ fn installed_runtime_from_dir(dir: &Path) -> Result<Option<InstalledNativeRuntim
     }))
 }
 
+fn installed_in_version_dir(version_dir: &Path) -> Result<Vec<InstalledNativeRuntime>> {
+    let mut installed = Vec::new();
+    if !version_dir.exists() {
+        return Ok(installed);
+    }
+    for runtime_entry in fs::read_dir(version_dir)
```
i386, line 254: [P2] Preserve the full-cache scanner’s non-directory behavior here. `installed()` skips a cache-root entry unless `file_type().is_dir()`; this helper now calls `read_dir` whenever `<cache>/<mesh_version>` merely exists. If that version entry is a file (a partial/corrupt cache artifact), resolver candidate discovery errors before it can select an explicit bundle—the same availability failure this PR is meant to avoid. Return an empty list unless `version_dir.is_dir()` and add a regression test for a file at the requested version path.

i386 (changes requested):
Small thing

crates/mesh-llm-native-runtime/src/cache.rs
ndizazzo: all done

PR #1172, opened 2026-08-04
fix(skippy): isolate chat grammar during speculative verification

i386:
## Why this matters

While testing a possible shortcut, Mesh could accidentally change the rules used to produce valid tool calls. Even when that shortcut was rejected, later generation could produce malformed output.

This change makes speculative verification a safe dry run. Rejected candidate tokens cannot damage the real grammar state used by normal generation.

## Technical details

- clone the active sampling and chat-grammar state before speculative verification
- stop sampling at the first target mismatch so rejected branch-conditioned rows cannot mutate grammar state
- preserve the authoritative sampler until the verified prefix is known to be valid
- tighten verify-window classification for truncated predictions and missing full-accept boundary tokens

Non-grammar and ordinary generation paths are unchanged.

## Validation

- `bash scripts/prepare-llama.sh` — complete 66-patch queue applied successfully
- `cargo test -p skippy-runtime --quiet` — 68 passed
- `cargo test -p skippy-server verify_window_` — 10 passed

The optional model-backed lazy-grammar correctness test compiled but skipped because `SKIPPY_CORRECTNESS_MODEL` was not configured locally.

This PR is based directly on `main` and has no dependency on the companion fixes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structured generation verification by stopping at the first prediction mismatch.
  * Prevented rejected prediction suffixes from affecting grammar state.
  * Added clearer handling for incomplete prediction output and missing decision boundaries.
  * Improved reliability for long tool-context requests, including grammar generation, batched verification, and draft-token acceptance.
* **Documentation**
  * Clarified which prediction data is considered authoritative during proposal processing.


ndizazzo (approved):
I think it would be worth adding one more model-backed regression check for tool calling with a long resident KV context (8k+ tokens)... The current tests validate the short verification window, but the proposed test would sample tokens from the existing (potentially large) KV cache. 

Adding a fallback check for latency and native-MTP acceptance at a sizeable context length would guard against a long-context throughput regression.

i386:
@ndizazzo coming right up

PR #1180, opened 2026-08-05
fix(plugin): bound proposal queue deadlines

### LINKED ISSUE #1178: Make native serving plugin proposal deadlines resilient to slow lifecycle callbacks
## Problem

The native serving plugin host sends `GenerationReceiptSink::committed` and `LinearProposalIngress::propose` through the same `PluginDriver` FIFO. This preserves ordering, but work performed by a committed-token callback can consume the next proposal query's deadline before the query reaches the plugin.

`LinearProposalIngressConfig` currently describes the deadline as advisory because `propose` executes synchronously and cannot be preempted. The additional FIFO dependency means even a fast `propose` implementation can receive the query too late when an earlier lifecycle callback is slow.

## Desired serving contract

- Preserve the causal rule that every proposal query observes all target-authoritative token commits that precede it.
- Do not allow unrelated queued callbacks, or opaque queue wait, to silently consume a proposal deadline.
- Treat the configured proposal deadline as a host-side wall-clock bound. If the required committed state cannot be established in time, return an explicit abstention/deadline outcome and continue normal target decoding.
- A slow callback must not permanently starve later proposal positions or poison the plugin/session.
- Expose queue-wait time, callback execution time, deadline exhaustion before dispatch, and deadline exhaustion inside the plugin boundary as structured telemetry.

The implementation should not blindly reorder state-critical commits. Possible designs include making the required committed delta/state explicit on the proposal query, coalescing committed deltas before a priority proposal command, or separating passive report/observation work from state-critical proposal preparation. The required property is that causal ordering and the hard deadline are both enforceable rather than implied by one FIFO.

## Acceptance tests

1. A deliberately slow `committed` implementation cannot make `propose` block past its configured deadline.
2. Target decoding remains fail-open when the plugin misses that deadline.
3. A missed position does not starve later positions once the plugin becomes responsive.
4. The proposal sees every required preceding committed token; no reordering or future-state leak is allowed.
5. Slow `report`, `discard`, or passive observation work cannot delay a subsequent proposal query.
6. Telemetry distinguishes queue delay, callback delay, host abstention, plugin abstention, and a candidate returned too late.

## Non-goals

- Mesh should not own plugin persistence or repair a plugin's internal durability design.
- Mesh should not weaken target-authoritative verification or accept late proposals.

i386:
## Summary

A native serving plugin can no longer stall a decode, lose a proposal decision, or leave its `shutdown` callback unrun.

- Proposals are scheduled ahead of queued passive callbacks, so a slow `report`/`discard` can never consume a proposal's deadline.
- An expired proposal abstains before dispatch instead of doing stale plugin work.
- A candidate returned after the deadline is withheld from decode and the plugin is always told, so no decision ID is left unresolved.
- Plugin shutdown is now guaranteed on drop, including when a queue is saturated.
- Privacy-safe queue/callback timing and deadline outcome telemetry is emitted per request.

## Root cause

The driver used a single FIFO for lifecycle callbacks, proposal queries, and passive proposal outcomes. A callback ahead of a proposal could consume its deadline before the plugin was reached.

## Review round

Follow-up fixes from review (`c979bbc2`), each verified against the code and corroborated with an independent model:

- **Shutdown could be skipped.** Shutdown was delivered as a *queued command*, so a full queue meant the worker was never told to stop; separately, the `is_finished()` gate raced with thread teardown and skipped the join. Both left a worker holding `Arc<ActivePlugin>`, so `Arc::get_mut` failed and the native `shutdown` callback never ran. Queues are now **closed** (which needs no capacity) and joined on an explicit exit signal set from a drop guard, so it fires on unwind too.
- **Terminal discards could be dropped.** A late candidate is withheld from decode, making its `discard` the only way the plugin can resolve that decision ID — and it was being discarded on a full queue. Terminal dispositions now use reserved queue headroom, and delivery failures are logged.
- **Telemetry could contradict the forwarding decision.** Three separate `Instant::now()` reads meant a proposal could be forwarded while telemetry reported `CandidateReturnedTooLate`. One timestamp now drives both.
- **Plugin errors were silently downgraded.** A failing callback became an indistinguishable abstention. The error is now logged and surfaced as `SourceError`; the fail-open behaviour at the ingress boundary is kept deliberately, and documented, so a misbehaving plugin cannot kill a generation.

## Architecture

The queue/driver/worker subsystem moved into a private `plugin_dispatch` module, with no crate-root re-export, taking `lib.rs` from **1689 → 745** lines per the repo's >1000-line rule. The fake-plugin test harness moved to `test_support` so both modules share it; its process-global counters became per-instance handles, which were cross-talking between parallel tests.

## Protocol

No wire-protocol or plugin-ABI change. Queue scheduling, shutdown, and telemetry are host-internal, so mixed-version meshes and existing plugins are unaffected.

## Validation

- `cargo test -p mesh-native-serving-plugin-host --lib` — 18 passed, including four new regression tests, one per bug above
- `cargo test -p mesh-llm-host-runtime runtime::survey::tests --lib` — 17 passed
- `cargo test -p skippy-server --lib` — 382 passed
- `cargo check -p mesh-llm`
- `cargo clippy -p mesh-native-serving-plugin-host --all-targets -- -D warnings`
- `cargo fmt --all --check`
- CI green on the merge with `main`

Closes #1178

## Summary by CodeRabbit

* **New Features**
  * Added privacy-safe telemetry for proposal queue wait time, callback duration, and source outcomes.
  * Exposed proposal telemetry through the public frontend interface.
  * Improved queue recovery, ordering, and isolation of background callbacks.

* **Bug Fixes**
  * Enforced proposal deadlines before and after processing, preventing expired or late candidates from being forwarded.

* **Documentation**
  * Documented the new proposal source telemetry attributes.


crates/mesh-native-serving-plugin-host/src/lib.rs
```diff
@@ -609,88 +724,211 @@ impl PluginDriver {
...
-                    result
-                        .as_ref()
-                        .map(Clone::clone)
-                        .map_err(ToString::to_string),
-                );
+                let callback_elapsed_us = elapsed_us(callback_started);
+                let candidate_was_late =
+                    matches!(&result, Ok(Some(_))) && Instant::now() >= query.deadline;
+                let outcome = match &result {
+                    Ok(Some(_)) if Instant::now() >= query.deadline => {
```
ndizazzo: **Double `Instant::now()` in late-candidate path can disagree with telemetry.**

Lines [891](https://github.com/Mesh-LLM/mesh-llm/blob/agent/fix-plugin-proposal-deadlines/crates/mesh-native-serving-plugin-host/src/lib.rs#L891) and [893](https://github.com/Mesh-LLM/mesh-llm/blob/agent/fix-plugin-proposal-deadlines/crates/mesh-native-serving-plugin-host/src/lib.rs#L893) each read `Instant::now()` independently. Between the two reads, time advances — it is possible for line 891 to see pre-deadline (so `candidate_was_late = false`) while line 893 sees post-deadline and classifies telemetry as `CandidateReturnedTooLate`. The proposal then gets forwarded (`!candidate_was_late` at line 908) but telemetry reports it was late.

**Fix:** Capture a single timestamp once:
```rust
let callback_finished_at = Instant::now();
let candidate_was_late = matches!(&result, Ok(Some(_))) && callback_finished_at >= query.deadline;
let outcome = match &result {
    Ok(Some(_)) if candidate_was_late => LinearProposalSourceOutcome::CandidateReturnedTooLate,
```

Also note line 897 has a third `Instant::now()` for the `Ok(None)` deadline case — same window.

crates/mesh-native-serving-plugin-host/src/lib.rs
```diff
@@ -609,88 +724,211 @@ impl PluginDriver {
...
+                    }
+                    Ok(Some(_)) => LinearProposalSourceOutcome::Ready,
+                    Ok(None) if Instant::now() >= query.deadline => {
+                        LinearProposalSourceOutcome::DeadlineExceededInPlugin
+                    }
+                    Ok(None) => LinearProposalSourceOutcome::Abstained,
+                    Err(_) => LinearProposalSourceOutcome::SourceError,
+                };
+                if candidate_was_late && let Ok(Some(proposal)) = &result {
+                    let _ = passive_queue.enqueue(PluginCommand::Discard(
```
ndizazzo: '**Passive `Discard` for late candidate is silently dropped when passive queue full.**

The result of `passive_queue.enqueue(...)` on line 904 is `let _ = ...`, so if the passive queue hits capacity, the discard for a late candidate is silently lost. The proposal itself isn't forwarded (good), but its activation-frame resources are never cleaned up — no downstream report that this slot should be discarded. This leaks state in the decode pipeline until eviction happens naturally.

Consider either:
- Making `Discard` terminal commands un-droppable (e.g., reserve one capacity slot, or retry with backoff)
- At minimum logging a warning when it fails'

crates/mesh-native-serving-plugin-host/src/lib.rs
```diff
@@ -524,8 +526,13 @@ struct NativeProposalIngress {
 }
 
 impl LinearProposalIngress for NativeProposalIngress {
-    fn propose(&self, query: LinearProposalQuery) -> Result<Option<LinearProposal>> {
-        self.driver.propose(query)
+    fn propose(&self, query: LinearProposalQuery) -> Result<LinearProposalSourceResponse> {
+        let response = self.driver.propose(query)?;
+        let proposal = response.proposal.unwrap_or_default();
```
ndizazzo, line 533: '**Error from native plugin callback is swallowed by `.unwrap_or_default()`**

The `response.proposal` field in [`NativeProposalIngress::propose`](https://github.com/Mesh-LLM/mesh-llm/blob/agent/fix-plugin-proposal-deadlines/crates/mesh-native-serving-plugin-host/src/lib.rs#L531) is a `Result<Option<LinearProposal>, String>`. When the plugin callback returns an error, this becomes `Err(String)` — and `.unwrap_or_default()` replaces it with `Ok(None)` (i.e., "abstained"). The original error message from the native plugin is lost.

The ingress adapter then converts the outer `Result` to a different type before returning up through skippy-server into linear_decode.rs where errors become fatal bailouts at line 168 (`expect("linear proposal failed")`). So this silently turns "plugin crashed" into "abstained", which means decode continues without ever knowing something went wrong.

**Suggestion:** Propagate the error or map it to a distinct telemetry outcome rather than defaulting to `Ok(None)`.'

crates/mesh-native-serving-plugin-host/src/lib.rs
```diff
@@ -549,45 +556,153 @@ enum PluginCommand {
...
+
+    fn mark_stopped(&self) {
+        self.stopped.store(true, Ordering::Release);
+        self.available.notify_all();
+    }
+
+    fn next(&self) -> QueuedPluginCommand {
+        let mut commands = self
+            .commands
+            .lock()
```
ndizazzo: '**`PluginCommandQueue::next()` blocks indefinitely on unbounded condvar wait**

[`queue.next()`](https://github.com/Mesh-LLM/mesh-llm/blob/agent/fix-plugin-proposal-deadlines/crates/mesh-native-serving-plugin-host/src/lib.rs#L637) at line 634 uses an infinite `loop` with no timeout on the condvar wait. If the queue is stopped (`mark_stopped`) but a thread already entered `wait()`, it will block forever — there's no corresponding `notify_all()` in the stop path that would wake threads currently blocked inside `next()`.

Looking at line 628, `mark_stopped` does call `self.available.notify_all()` after setting the flag. But if a thread is between checking `pop_next` (which returns None) and calling `.wait()`, it will enter an infinite loop:
1. Thread checks stopped → false  
2. Pop next → None 
3. Calls wait(blocked here while stop happens + notifies)
4. Wait wakes, but no new command was added... 

Actually `mark_stopped` does notify_all so threads wake up — they'll see the queue is empty and loop back to wait again indefinitely unless a Shutdown or more commands arrive before shutdown completes.

**Consider:** Add a timeout-based wait (`wait_timeout`) that re-checks `stopped.load()` each iteration, breaking when stopped instead of blocking forever.'

ndizazzo:
Left inline comments on four items not covered by CodeRabbit. Summary below (each linked to the threaded discussion):

**Functional correctness:**

### 🟠 Double `Instant::now()` race in late-candidate path [comment](https://github.com/Mesh-LLM/mesh-llm/pull/1180#discussion_r3722355635)
Lines 891, 893 (and 897) each read `Instant::now()` independently. Between reads the deadline can cross over — `candidate_was_late` says "not late" while telemetry classifies it as `CandidateReturnedTooLate`. The proposal gets forwarded but telemetry reports otherwise. One shared timestamp fixes this (~2 line change).

### 🟡 Passive `Discard` silently dropped when queue full [comment](https://github.com/Mesh-LLM/mesh-llm/pull/1180#discussion_r3722356699)
Line 904: `let _ = passive_queue.enqueue(...)` — if the passive queue is at capacity, late-candidate discards are silently dropped. The proposal isn't forwarded (good), but activation-frame resources leak until natural eviction.

### 🟡 Plugin errors swallowed as "abstained" [comment](https://github.com/Mesh-LLM/mesh-llm/pull/1180#discussion_r3722357594)
Line 531: `response.proposal.unwrap_or_default()` turns any plugin callback error into a silent "abstained" (`Ok(None)`). The original native-plugin error message is lost and decode continues without knowing something failed.

**Robustness:**

### 🟡 Unbounded condvar wait in `next()` [comment](https://github.com/Mesh-LLM/mesh-llm/pull/1180#discussion_r3722358653)
Line 634: The infinite loop with no timeout means if the queue stops while a thread is between checking `pop_next` and calling `.wait()`, it loops forever waiting for commands that will never arrive. Add `stopped.load()` re-check after each wake or use `wait_timeout`.

---

Overall solid design — deadline enforcement, two-worker architecture, telemetry propagation are well thought out. The above items are edge cases in the new queue/worker model.

michaelneale:
From the :robot: 

reviewed each item and corroborated with an independent model before changing anything. Pushed `c979bbc2`, and the branch is now merged up to `main`.

### Addressed

**Single timestamp for late-candidate classification** (@ndizazzo) — Valid. The forwarding decision and the telemetry outcome now derive from one `callback_finished` read, so a candidate can no longer be forwarded while telemetry reports `CandidateReturnedTooLate`. The whole classification collapsed into a single `match` on the result, which also removes the third `Instant::now()` in the `Ok(None)` arm.

**Terminal discard for a late candidate** (@ndizazzo, CodeRabbit) — Valid, and the more serious of the two. Because the proposal is withheld from decode, the `discard` was the *only* remaining way the plugin could resolve that decision ID, and it was being dropped on a full queue. The passive queue now has reserved headroom (`PLUGIN_TERMINAL_RESERVE`) that only terminal dispositions may use, and a failure to deliver is logged rather than swallowed.

**Worker shutdown and joining** (CodeRabbit) — Valid, and it had two distinct causes:
- Shutdown was delivered *as a queued command*, so a full queue meant the worker was never told to stop. Queues are now **closed** instead, which needs no capacity; `next()` returns `None` once closed and drained.
- The `is_finished()` gate raced with thread teardown. Workers now signal an explicit `WorkerExit` from a drop guard (so it fires on unwind too), and the join waits on that signal.

Both paths previously left a worker holding `Arc<ActivePlugin>`, which made `Arc::get_mut` fail and skipped the native `shutdown` callback.

**Module extraction** (CodeRabbit) — Valid per the repo's own >1000-line rule. The queue/driver/worker subsystem moved to a private `plugin_dispatch` module (no crate-root re-export). `lib.rs` goes **1689 → 745** lines. The fake-plugin harness moved to `test_support` so both modules share it.

### Adjusted rather than applied as suggested

**`unwrap_or_default()` swallowing plugin errors** (@ndizazzo) — Partially valid. The error string was genuinely lost, and that's now fixed: the worker logs the plugin's message and returns `Err`, so `SourceError` is distinguishable from a deliberate abstention. But I kept the fail-open behaviour at the ingress boundary, with a comment explaining why. To confirm the stated consequence, I checked `local_generation/linear_decode.rs` — there is no `expect("linear proposal failed")` there, and a source error does not cause a fatal bailout. Making a misbehaving plugin able to kill a generation would be a worse outcome than fail-open.

### Not changed

**`next()` blocking forever on the condvar** (@ndizazzo) — Good catch on the general shape, but not reachable as written: `mark_stopped` was only ever called by a worker's own stop guard as it exited, so no thread could subsequently call `next()` on that queue. The genuinely reachable hang was the full-queue shutdown case above, and in that path the flag was never set at all. The rewrite makes this moot — `next()` now returns `None` on a closed, drained queue.

### Tests

Four new regression tests, one per real bug: `a_full_queue_can_still_be_closed_and_drained`, `dropping_the_driver_drains_its_backlog_and_shuts_the_plugin_down`, `a_late_candidate_delivers_its_discard_to_the_plugin`, and `a_failing_proposal_callback_reports_source_error_instead_of_abstaining`.

While adding them I found the existing harness used process-global `AtomicUsize` counters, which cross-talk between parallel tests and made shutdown/cancel assertions unreliable. Those are now per-instance handles.

`cargo test -p mesh-native-serving-plugin-host --lib` → 18 passed, `cargo test -p skippy-server --lib` → 382 passed. Clippy clean with `-D warnings`, `cargo fmt --all --check` clean, `cargo check -p mesh-llm` clean, and CI is green.

ndizazzo (approved):

PR #1191, opened 2026-08-07
fix: IPC idle timeout for external-relay plugins

MahdiHedhli:
## Summary

**Edit: this PR's scope was reduced after further investigation — see the update comment below for what changed and why.**

A plugin whose real inference traffic bypasses the host↔plugin IPC channel entirely (going through its own HTTP proxy instead — e.g. an `openai_http(...).managed_by_plugin(false)` external-relay plugin) generates no application-level IPC traffic of its own. The periodic `HealthRequest` from the supervisor (`health::HEALTH_CHECK_INTERVAL_SECS`, 15s) is the *only* guaranteed traffic such a plugin's connection ever sees.

The previous fixed 10s `PLUGIN_ENVELOPE_PREFIX_READ_TIMEOUT` is shorter than that 15s interval, so it always eventually disconnects an idle-but-healthy relay-style plugin as falsely idle, race-losing against its own supervisor's probe cadence.

## Change

Derives the timeout from `HEALTH_CHECK_INTERVAL_SECS` directly (4x plus a fixed floor, to absorb a slow/jittery tick) instead of a separate hardcoded constant, so the relationship this timeout depends on stays enforced if that interval ever changes.

## Verification

Found and confirmed via live testing: an external-relay plugin relaying to a real OpenAI-compatible server (LM Studio) over Serve mode. With the 10s timeout, the plugin's IPC connection was torn down mid-session as falsely idle. With the fix, the connection stays alive indefinitely and a real chat completion round-trips correctly.

## Test plan

- [x] `cargo check` / `cargo clippy` clean
- [x] Full `mesh-llm-host-runtime` unit test suite passes (1910/1911; the one failure — `stalled_owner_control_handshake_expires_deterministically` — is a pre-existing timing-sensitive flake unrelated to this change, confirmed by reverting just that test file and by rerunning)
- [x] Live manual test: plugin's IPC connection survives an extended idle period; chat completion still round-trips correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for external relay plugins by allowing more time for responses during periods of low activity.
  * Reduced the risk of connection timeouts when plugins communicate primarily through periodic health checks.


MahdiHedhli:
Follow-up PR building on this one: #1192 — makes plugin-only nodes discoverable/reachable by other mesh peers (mesh-wide serving), not just their own local API as fixed here.

MahdiHedhli:
**Update:** this PR originally also included a fix to make the passive listener (\`network/openai/transport.rs::handle_mesh_request\`) merge plugin models into \`/v1/models\` and route matching requests to a plugin's loopback endpoint. After rebasing onto current \`main\` to resolve merge conflicts, I found that fix was targeting a code path this crate's own recent "daemon model lifecycle reconciliation" refactor has already made unreachable in the primary startup flow (\`run_auto_model_path_or_shutdown\`, the function that leads to that listener, is now marked \`#[expect(dead_code, reason = "bridges the retained advertised-model and passive runtime compatibility lanes")]\`) — every node's own local ingress now goes through the already-plugin-aware \`network/openai/ingress.rs\` path unconditionally instead. That fix would have been a no-op, so I dropped it rather than land dead-code-adjacent changes. This PR is now just the timeout fix, which remains fully independent and necessary.

The mesh-wide reachability piece (#1192) needed similar rework for the same reason — see that PR for the corrected version.

MahdiHedhli:
@coderabbitai review

i386 (commented):
Reviewed the current commit and found no actionable correctness issues. The prefix-read timeout is derived from the 15-second supervisor health-check cadence (4x plus 30 seconds), while the body and mesh-response timeouts remain unchanged. The existing stalled-prefix/body test continues to cover the timeout paths.

i386 (approved):
LGTM - thanks @MahdiHedhli ! 

Don't forget to come join us on Discord :)

PR #1192, opened 2026-08-07
feat: make plugin-only Serve mode nodes discoverable/reachable by other mesh peers

MahdiHedhli:
## Summary

Builds on #1191 (please review/merge that first — this branch is based on top of it).

**Edit: this PR was substantially reworked after rebasing onto current `main` — see the update comment below.**

A node whose only inference capacity comes from a plugin (e.g. one relaying to an external OpenAI-compatible server — no local model ever loads) is invisible to every other mesh peer, for two independent reasons, even though its own local ingress and gossip advertisement of its plugin's models already work correctly (both already generalized by this crate's "daemon model lifecycle reconciliation" work, ahead of this PR).

## The gap

1. **Host-role eligibility.** Other peers only consider a node as a route candidate for model X if `hosts_for_model()` finds it, which requires `NodeRole::Host`. That role is set in exactly one place in the whole crate — as a side effect of a local model finishing load. A plugin-only node's role never leaves `Worker`, so it's filtered out of every peer's routing consideration regardless of what its gossip payload advertises.
2. **Inbound tunnel routing.** `tunnel::Manager`'s configured `http_port` — where an inbound QUIC-tunneled request from a peer gets relayed — is, like the role above, only ever set as a side effect of a local model finishing load. If it's still `0`, an inbound tunnel is silently dropped.

## Fix

Both gaps have the same shape: a value that's already generalized everywhere else in `run_auto()`'s startup path (which now unconditionally binds the real API proxy and constructs the tunnel manager before any model resolution happens) except at these two specific call sites, which still only fire on local-model-load completion.

- Set `tunnel_mgr.set_http_port(api_port)` unconditionally, immediately after the tunnel manager is constructed. The existing call sites in `startup_handles.rs` become redundant-but-harmless (same node, same `api_port`, for the process's lifetime).
- Add `spawn_plugin_host_role_watcher`, a small background task (skipped for `--client` nodes) that polls the plugin manager's already-debounced health state on a short interval and promotes to `Host { http_port: api_port }` when it reports at least one model, demoting back to `Worker` when it stops. Only ever touches a `Host` role it set itself.

Both are two small, unconditional additions at a single call site in `run_auto()` — no new machinery, and (per the update below) no longer touching the passive-listener code path at all.

Full analysis in `docs/plugins/plugin-hosted-mesh-serving-design.md`, added by this PR.

## Verification status — please read before merging

Compiles cleanly, clippy clean, full `mesh-llm-host-runtime` unit test suite passes (1910/1911; the one failure is a pre-existing, confirmed-flaky timing test unrelated to this change — see #1191 for detail). **The actual mesh-wide behavior this PR targets — another peer discovering and successfully routing a request to a plugin-only node — has not yet been exercised by a live multi-peer test.** I don't have a convenient multi-node test rig set up for this yet. Flagging this explicitly. Happy to do that verification before merge if a maintainer can point me at (or review with) a suitable multi-node setup.

## Test plan

- [x] Compiles cleanly, clippy clean
- [x] Full unit test suite passes (see above)
- [ ] Live multi-peer verification — not yet done, see above

## Summary by CodeRabbit

- **New Features**
  - Plugin-hosted inference models can now be served across the mesh by eligible worker nodes.
  - Node roles automatically adjust as plugin model availability changes.
  - Local model availability is preserved when plugin models become unavailable.
  - Inbound plugin traffic uses a health-check-based timeout for improved reliability.

- **Documentation**
  - Added design documentation covering mesh-wide plugin model serving, current behavior, and deferred enhancements.


MahdiHedhli:
**Update:** rebasing onto current \`main\` surfaced that this crate's own recent "daemon model lifecycle reconciliation" work had already independently solved a chunk of what an earlier version of this PR targeted — specifically, \`tunnel::Manager\` construction and this node's own local ingress are now unconditional in \`run_auto()\`'s startup path regardless of local model state, whereas when I originally wrote this fix (against an older base) they weren't. That made the earlier version's changes to \`network/openai/transport.rs\` (routing plugin models through the passive listener) a no-op — that code path is dead in the primary flow now (see the comment on #1191 for detail) — so I dropped it.

What's left and still genuinely needed: \`tunnel_mgr.set_http_port(...)\` and \`NodeRole::Host\` promotion are *still* only ever triggered by a local model finishing load, even after that refactor — nothing generalized those two specifically for a plugin-only node. This PR is now scoped to just that, and is meaningfully smaller and simpler than the original version (no threading through the passive-listener/model-selection context, since that machinery turned out to be dead code already).

MahdiHedhli:
@coderabbitai review

MahdiHedhli:
Addressed 2 of the 3 findings from the fresh review in dc3568eb:

- ✅ **Error vs. empty distinction**: `inference_models()` returning `Err` was collapsed into "no models" via `.unwrap_or(false)`. Now logs and skips the tick on error instead of acting on an indeterminate read.
- ✅ **Role-ownership race**: this was a real, valid finding, not just style — `http_port` is identical whether a local model or this watcher's promotion is why a node is `Host` (both derive it from the same `api_port`), so the role value alone can't tell the two apart. If a local model starts being served *after* this watcher promotes the role, the watcher's stale flag could later demote a role the local model now depends on. Added a `node.models_being_served().is_empty()` check before actually changing the role on the demotion path. Also softened the design doc's claim about this to be accurate (it closes this one race, it isn't full ownership tracking) rather than overstated, and added "real ownership tracking for NodeRole::Host" as an explicit deferred follow-up.

Not addressed, with reason:
- **Extract the watcher into its own module** — `run_auto.rs` already has many small `spawn_run_auto_*`-style helpers defined inline (bootstrap proxy, discovery publisher, model intent channel, etc.) rather than in dedicated modules; this watcher matches that existing convention. A broader reorganization of this file is out of scope for this fix.

Recompiled, clippy clean, full unit test suite still passes.

MahdiHedhli:
Addressed both findings from the latest review in e932d025 — both were doc wording precision issues in \`plugin-hosted-mesh-serving-design.md\`:

- The "Status" intro said the live test covered "serving an LM Studio model through the mesh," which read as claiming mesh-wide serving was exercised — contradicted a few paragraphs later by "Deliberately deferred"'s explicit "not yet exercised by a live multi-peer test." Reworded to be precise about what was actually tested (a single node's own local proxy through the tunnel path, not routing from another peer).
- "Both changes are ... unconditional additions ... no new machinery" undersold the host-role watcher, which is new background machinery started conditionally (skipped for \`--client\` nodes) — only the tunnel-port fix is truly unconditional/no-new-machinery. Reworded to describe each accurately.

\`run_auto.rs\` itself (the role-ownership race + error-handling fix from the previous round) came back "skipped from review as similar to previous changes" — no new findings there.

i386 (approved):
Reviewed and fixed the remaining actionable review findings. The production plugin-envelope timeout remains derived from the health cadence while tests use a short injected timeout; the watcher now lives in `runtime/plugin_host_role.rs`; and host-role claims are reference-counted across plugin and local-model lifecycles so one source cannot demote a host still owned by the other. Full validation: `cargo fmt --all -- --check`, `cargo check -p mesh-llm-host-runtime`, and `cargo test -p mesh-llm-host-runtime --lib` (1,912 passed, 8 ignored). The broad host-runtime Clippy run is still blocked by 28 pre-existing unrelated `unfulfilled_lint_expectations` errors. Approving; live multi-peer verification remains explicitly deferred in the PR description.

i386:
thanks again!

michaelneale (commented):
🤖 (micn's agent) Deep review — traced the full claim/release lifecycle through `startup_handles.rs`, the watcher, and the tunnel path. The reworked shape is right: the two gaps are real (I verified `NodeRole::Host` was previously set in exactly one place as a local-model-load side effect, and `tunnel.rs` does silently drop inbound relays while `http_port == 0`), and the `74ab7ad9` move from a task-local flag to reference-counted `claim_host_role`/`release_host_role` on `mesh::Node` is a genuinely better mechanism than what the PR description still describes — it correctly handles multiple concurrent local-model tasks and the local-model+plugin overlap, and gating `set_role` behind `#[cfg(test)]` makes claims the only production mutation path. Nice.

Findings, most important first:

**1. Claim leak on the `StartupLoopOutcome::Exit` path.** `startup_publish_loaded_runtime` claims `HostRoleClaim::LocalModel` on every loop iteration, and `startup_shutdown_local_model_loop` releases it — but that teardown only runs for `Shutdown` and `RelaunchSplit`. The `Exit` outcome (reached via `startup_handle_fallback_failure` → `StartupLoopControl::Return`, i.e. a split model withdrawn after stage-peer loss + failed local fallback) ends the task through `startup_remove_runtime_instance_artifacts`, which never releases the claim. Pre-PR this didn't matter (nothing ever demoted `Host`), but under the new claims model the leaked `LocalModel` claim is permanent: a plugin-only node that once briefly hosted a split model stays advertised as `Host` even after its plugin models disappear, because `release_host_role(PluginInference)` can never empty the map. Peers keep routing to a node serving nothing (it fails gracefully via the now-always-set tunnel port, but it's still wrong routing). Suggest releasing the claim in the `Exit` arm or inside `startup_handle_fallback_failure`.

**2. Design doc is now stale against the code it documents.** `plugin-hosted-mesh-serving-design.md` describes the watcher as using "a task-local flag ... a soft safeguard, not real ownership tracking", describes a `models_being_served()` demotion check that no longer exists in `plugin_host_role.rs`, and lists "Real ownership tracking for `NodeRole::Host` ... an explicit set of claimants ... only clear the role when the last claimant releases it" under **Deliberately deferred** — which is precisely what `74ab7ad9` implemented. Since the doc ships in-tree as the authoritative analysis, it should be updated in this PR (the PR body's "Fix" section has the same drift). Cheap fix, avoids the next reader distrusting the doc.

**3. Behavior change worth calling out explicitly: nodes now demote to `Worker` on model shutdown.** Previously a node stayed `Host` forever after its first model load. Now clean shutdown/relaunch-split releases the claim and (with no other claimants) demotes. I think this is *correct* — a node waiting for split peers isn't serving anything — but it's a mesh-visible gossip change on the local-model path, not just the plugin path, and the PR description doesn't mention it. Mixed-version note: older peers just see a role transition in gossip, which they already handle, so no wire concern.

**4. Plugin transport timeout: 10s → 90s** (`HEALTH_CHECK_INTERVAL_SECS * 4 + 30`). The reasoning is sound and well-commented (relay-style plugins see no IPC traffic between health probes, so 10s falsely disconnected every idle-healthy relay). Side effect: a genuinely hung plugin connection now takes up to 90s to detect instead of 10s. Acceptable trade, but worth a maintainer's conscious nod since it affects all plugins, not just relays.

**5. Minor:** the watcher task is spawned detached with no shutdown handle (loops forever; consistent with other `run_auto` spawns, fine); the 5s poll reading already-debounced health state is a sensible cadence; lock ordering in claim/release is consistent (claims → role) so no deadlock.

On the honest "not yet live multi-peer tested" flag: given micn's lab has a 2-node setup, the confidence test before 0.75.x-anything would be: node A serving a plugin-relayed model only, node B `client --auto`, verify B's `/v1/models` shows A's plugin model and a completion routes through A's tunnel. Happy to run that on the lab rig if useful — it's exactly the scenario the fix targets and would close the one open verification box.

Also noting CI shows only CodeRabbit on the latest merge commit — the workflow run for `68d30b22` doesn't appear to have started; worth a re-trigger before merge.

i386:
@micspiral need a dario on this

micspiral:
:robot: Reviewed this locally on the current branch head (`8944bc43`). Formatting, clippy (`-p mesh-llm-host-runtime --all-targets -D warnings`), and the full lib suite are all clean for me — 1926 passed / 0 failed, including both new tests, so the "1910/1911 with one flaky" note in the description is now stale. The diagnosis checks out too: `accepts_http_inference()` really is just `matches!(self.role, NodeRole::Host { .. })`, gossip already merges plugin models via `append_external_inference_models`, and `tunnel.rs`'s `port == 0` early-return really did drop inbound tunnels. The refcounted claim registry that replaced the original task-local flag is a nicer shape than what the description describes, and the claim/release balance holds across the shutdown, relaunch-split, and fallback-failure exits.

One thing I want to flag before merge, because I don't think the description covers it:

**Is unconditional `set_http_port` on `--client` nodes deliberate?**

`spawn_plugin_host_role_watcher` / `plugin_host_role::spawn` is correctly gated behind `if !is_client`, but `tunnel_mgr.set_http_port(api_port)` immediately above it is not:

```rust
tunnel_mgr.set_http_port(api_port);
if !is_client {
    plugin_host_role::spawn(node.clone(), plugin_manager.clone(), api_port);
}
```

The effect is that a client node's inbound QUIC HTTP tunnel, which previously terminated at the `port == 0` check, now relays into the local API proxy. With no local model that lands in `route_missing_local_model` → `remote_mesh_targets` → `hosts_for_model`, which tunnels straight back out to another peer. So a `--client` node becomes a mesh-internal request relay for any admitted peer that opens a tunnel to it.

It isn't alarming: clients never advertise `Host`, so nothing selects them through `hosts_for_model()` or `any_host()`, requests terminate at a real host rather than looping, and `RemoteQuicHttp` still goes through the activity-policy admission check. But it is reachable by any peer that deliberately dials a client, which on the public mesh means any member — and it looks incidental rather than designed, since the watcher right below it went to the trouble of gating on exactly this condition.

If it's incidental, the fix is to gate it the same way:

```rust
if !is_client {
    tunnel_mgr.set_http_port(api_port);
    plugin_host_role::spawn(node.clone(), plugin_manager.clone(), api_port);
}
```

If it *is* deliberate — e.g. you want clients usable as relay hops — that's worth a sentence in the design doc under a "client nodes" heading, since it's a new inbound-traffic capability for a node role that previously had none, and reviewers of the plugin-only path won't otherwise think to look for it.

Smaller, non-blocking notes:

- The IPC prefix timeout goes 10s → 90s. The reasoning is right (10s < the 15s health cadence was a guaranteed false-idle disconnect), but detection of a genuinely wedged plugin connection is now ~9x slower. Probably fine given the supervisor's own failure threshold covers liveness, but worth a nod from whoever owns `plugin/health.rs`.
- `read_envelope_from_reader` now takes injected 1s timeouts, so the production constant is no longer exercised by any test. Necessary to keep tests fast, just noting the coverage gap.
- Doc drift: the design doc and the `run_auto.rs` comment both still say `spawn_plugin_host_role_watcher` (renamed to `plugin_host_role::spawn`), and the doc still describes the task-local-flag approach that the refcount registry replaced.
- A forced `task.abort()` (`await_managed_model_stop(force: true)`, and the 3s shutdown drain) skips `startup_shutdown_local_model_loop` and therefore the release, so a stale `LocalModel` claim can leave the node advertising `Host` with nothing loaded. Not a regression — the role was previously never cleared at all — but the refcount makes it newly expressible.

On the missing live verification: the new `host_role_claim_transitions_regossip_to_connected_peer` test covers the gossip-propagation half in-process, which lowers the risk here quite a bit. The genuinely unexercised part is the inbound tunnel relay, which is also exactly the part the client-gating question above touches — so those two are worth settling together.

michaelneale:
@MahdiHedhli I think this is ok, just that one Q from my bot above about the serving side effect - if you know if that is deliberate or not, and then I think this is good to go

PR #1193, opened 2026-08-08
fix(skippy): make local proposal handoff deadline-safe

i386:
## Problem

[`4bd1453fad1ab93ce14b010f76f91c368e929fb6`](https://github.com/Mesh-LLM/mesh-llm/commit/4bd1453fad1ab93ce14b010f76f91c368e929fb6) moved canonical commit delivery into the lifecycle path. Local generation did not publish the same deltas, so after one locally decoded token the next proposal query could observe a stale consumer position and fail closed.

Delivering every local token through an ordered lifecycle barrier fixes that position mismatch, but it also puts queue pressure and arbitrary lifecycle callback latency on inference. That is unsafe for the 8 ms proposal deadline: a slow or blocked consumer must not block or fail target decoding.

## What changed

- Carry the bounded canonical token delta directly on `LinearProposalQuery` instead of emitting a per-token lifecycle event from local generation.
- Apply ordinary decoded-token deltas in the native plugin worker before proposal lookup.
- Commit verified proposal tokens immediately before reporting the proposal outcome.
- Reconcile any remaining generated suffix before `finish_generation`.
- Track the committed generated prefix per request/session to preserve ordering and prevent duplicate commits.
- Reject a saturated plugin queue before accepting proposal handoff. Proposal-path failures disable proposals for the request and target decoding continues.
- Keep lifecycle start, finish, abort, and optional commit observations for integrations outside this local proposal hot path.

## Deadline and performance

- Local token emission no longer allocates or submits one lifecycle observation per token.
- There is no lifecycle flush, spin loop, or 250 ms delivery timeout on the inference path.
- The synchronous proposal caller waits only until the query's existing absolute deadline.
- A regression test blocks `commit_generation` for 250 ms behind a stricter 5 ms proposal deadline and verifies that decode returns without waiting for the callback to finish.

## Correctness coverage

- Pending target tokens are committed before proposal lookup.
- Verified tokens are committed before proposal outcome delivery.
- Final tokens not seen at a proposal boundary are committed before generation finish.
- Skipped or rejected handoffs retain pending state for later reconciliation.
- Cancellation accounting is isolated per fake plugin instance so timing tests remain deterministic under parallel execution.

## Validation

- `just with-lld cargo check -p mesh-native-serving-plugin-api -p skippy-server -p mesh-native-serving-plugin-host`
- `just with-lld cargo clippy -p skippy-server -p mesh-native-serving-plugin-host --all-targets -- -D warnings`
- `just with-lld cargo test -p mesh-native-serving-plugin-host --lib` — 13 passed

## Summary by CodeRabbit

* **Improvements**
  * Improved token tracking during generation and proposal processing.
  * Preserved pending tokens when proposal requests are skipped.
  * Added request and session details to proposal receipts.
  * Improved handling of proposal errors and unavailable proposal capacity.
  * Ensured committed tokens are delivered contiguously and receipt updates remain consistent during completion or cancellation.

* **Reliability**
  * Expanded coverage for token forwarding, commit handling, shutdown behavior, and proposal deadlines.


michaelneale (commented):
🤖 (micn's agent) Reviewed for 0.75.1. The diagnosis is convincing — the lifecycle refactor (4bd1453f) gave remote generation commit delivery but left local generation updating only the final receipt, so lifecycle consumers hold a stale canonical position and the boundary check correctly fails closed on the next proposal. Emitting an authoritative commit per local token from the single `emit_token` closure is the right shape, and covering both ordinary decode and verified proposal spans through one emission point avoids the double-count trap. CI is green and the strengthened receipt test verifying the exact commit stream is good regression coverage.

Three things worth a look before this ships in a point release:

**1. Inference can now fail on lifecycle delivery problems — previously best-effort.** `commit_local_generation_token` propagates `committed_before_proposal` errors as `OpenAiResult` out of `emit_token`, so a stalled/failed lifecycle sink aborts user generation after the 250ms timeout. The pre-existing contract elsewhere in this file is explicitly the opposite (`recording_failures`: *"These failures disable receipt recording for the affected generation but never fail inference"*). Failing closed is arguably correct when a proposal consumer genuinely depends on the canonical position — a stale position would just fail the boundary check anyway — but when receipts are enabled with *no* proposal consumer (metrics/audit-only sinks), a slow plugin sink now kills generations that previously succeeded. Was that trade-off intended? If not, one option: only use the ordered path when the linear-proposal ingress is active, and fall back to best-effort `committed()` otherwise.

**2. Per-token synchronous barrier on the hot decode path.** Every emitted token now does: boxed-slice alloc → channel send (with `yield_now` spin when full) → rendezvous-channel barrier → wait for the worker to drain the queue. At local decode rates this serializes token emission against the delivery worker, so a sink that takes even ~1ms per commit caps generation at ~1k tok/s and any transient stall inserts visible latency. The function is named `committed_before_proposal` but it runs on **every** token whenever receipts are enabled, including when the proposal path is disabled (`linear_proposal_enabled == false`, native MTP, hooks active). Same conditional as (1) would fix both: pay the ordered cost only where ordering buys correctness.

**3. Double-counted failure metric.** One failed commit increments both the worker's `delivery_failures` and (via the barrier error → `inspect_err`) `submission_failures`, so `delivery_failures()` reports 2 for a single failure — the new `queued_barrier_reports_delivery_failure_and_resets_interval` test even asserts `== 2`. Anyone alerting on this counter will see inflated numbers. Minor, but worth either not incrementing `submission_failures` for barrier-reported failures, or documenting that the counter can count one failure twice.

Also a small naming nit: `failures_since_barrier` resets only on the error path — on the `Ok` path it's already 0 so behavior is fine, just slightly odd to read.

None of these block the correctness fix; (1) is the one I'd want a deliberate answer on before 0.75.1, since it changes user-visible failure semantics for existing receipt users.

michaelneale (approved):
might want to address point 1 so it doesn't stop inference

michaelneale (changes requested):
might want to consider breaking inference case

crates/skippy-server/src/frontend/local_generation/tests.rs
```diff
@@ -65,6 +68,16 @@ fn wait_for_receipts(sink: &RecordingReceiptSink, expected: usize) {
...
+fn wait_for_commits(sink: &RecordingReceiptSink, expected: usize) {
+    for _ in 0..100 {
+        if sink.commits.lock().unwrap().len() >= expected {
+            return;
+        }
+        thread::sleep(Duration::from_millis(1));
+    }
+    panic!("timed out waiting for {expected} generation commits");
+}
+
```
ndizazzo: @coderabbitai does rust have a stronger pattern for this instead of busy-wait looping?

michaelneale (commented):
🤖 (micn's agent) Re-reviewed after `52b077ed` ("keep proposal commits off lifecycle hot path"). This is a substantially better design than the barrier approach and resolves both of my earlier concerns at the root rather than patching around them:

- **Hot path**: the synchronous submit-and-flush barrier is gone entirely (`submit_ordered`/`flush`/`Barrier` machinery deleted). The canonical token delta now travels *on* the proposal query itself (`LinearProposalQuery::pending_token_ids`), so the cost is only paid when a proposal is actually being made — a single boxed-slice clone on a path that's about to do a runtime decode anyway. Ordinary decode with no proposal consumer is back to fire-and-forget. This also kills the failure-semantics concern: lifecycle delivery can no longer abort inference.
- **Double-counted failure metric**: gone with the barrier.
- The new `Skipped` outcome is a nice touch — when there's no proposal capacity the pending delta stays owned by the caller instead of being consumed-but-undelivered, and `query_does_not_accept_pending_tokens_without_proposal_capacity` locks that in.
- The plugin-host side is careful: per-`(request_id, session_id)` committed-count tracking, prefix-extension validation (`commit does not extend the tracked generated-token prefix`), cleanup on abort/finish, and `commit_final_suffix` covering the tail between the last proposal boundary and generation end — so a plugin still receives a complete, contiguous commit stream even for generations where the proposal path never fires.

Two small observations, neither blocking:

1. **`unwrap_or_else(|| generated_token_count.saturating_sub(token_ids.len()))` in `commit_tokens`** — for an unknown key this trusts the caller's count to back-derive the prior prefix. It's only reachable if a commit arrives without a `begin` (or after cleanup), where failing closed with a `bail!` might be preferable to synthesizing a prefix. In practice `begin` always seeds the entry, so this is defensive-path behavior only.
2. **Semantics shift worth one line in the PR description**: `GenerationReceiptSink::committed` deliveries are now proposal-boundary-driven (plus final suffix) rather than strictly per-token-real-time for every configuration. The doc comment on the trait was updated to say exactly this ("optionally observes ... deltas"), which is honest — just make sure the PR body's "every token emitted" framing matches the final shape so 0.75.1 notes are accurate.

CI is green on the branch. Happy with this for the point release once the merge conflict with main (the #1180 proposal-deadline change touched the same files) is resolved.

PR #1195, opened 2026-08-08
Fix GPU nodes failing to start when the CUDA driver is newer than the installed toolkit

### LINKED ISSUE #1127: Windows: cuda12 runtime rejected for CUDA 13 driver despite backward compatibility, silently falls back to Vulkan which reports 0 GPUs (RTX 4090)
## Summary

On Windows with driver **595.97** (which reports CUDA 13.2), mesh-llm 0.74.0 silently installs the Vulkan runtime instead of the CUDA runtime and then reports zero runtime-selectable GPUs. The GPU is visible to Windows and `nvidia-smi` reports it correctly. The root cause is that the runtime compatibility check hard-gates on `CUDA 12` and rejects the `cuda12` runtime when the driver reports any CUDA 13.x version — even though CUDA 13 drivers are fully backwards-compatible with CUDA 12 libraries.

## Steps to Reproduce

1. Install NVIDIA driver **595.97** (or any driver that reports `CUDA Version: 13.x` in `nvidia-smi`) on Windows x86_64.
2. Install or update mesh-llm 0.74.0 via the standard installer / `mesh-llm update`.
3. Run `mesh-llm gpus`
4. Run `mesh-llm runtime list` to confirm which runtime was selected.
5. Run `mesh-llm runtime list --available` to see why the cuda12 runtime was rejected.

```
mesh-llm gpus
mesh-llm runtime list
mesh-llm runtime list --available
```

## Actual Result

```
⚠️ No runtime-selectable GPUs reported by the embedded inference backend.
   This node will run CPU-only until the backend exposes a selectable device
```

`mesh-llm runtime list` shows only the Vulkan runtime was installed:

```
📦 Installed native runtimes
   cache: C:\Users\<user>\AppData\Local\mesh-llm\native-runtimes
  - ✅ meshllm-native-runtime-windows-x86_64-vulkan 0.74.0 (vulkan)
```

`mesh-llm runtime list --available` shows the cuda12 runtime was rejected:

```
  - ⚠️ meshllm-native-runtime-windows-x86_64-cuda12 not compatible (cuda, windows/x86_64)
    MeshLLM: 0.74.0; Skippy ABI: 0.1.32
    reason: CUDA toolkit mismatch: runtime requires CUDA 12
```

Attempting to force-install it by full runtime ID also fails (the compatibility gate blocks it even when specifying the full artifact name):

```
mesh-llm runtime install meshllm-native-runtime-windows-x86_64-cuda12

❌ Native runtime install failed
   Reason: no compatible native runtime found for Skippy ABI 0.1.32 on windows/x86_64
```

## Expected Result

The RTX 4090 should be exposed as a selectable CUDA device with its 24 GB VRAM reported. CUDA 13 drivers are **fully backwards-compatible** with CUDA 12 shared libraries per NVIDIA's compatibility guarantee. The `cuda12` runtime should either:

- Be accepted when the host driver reports CUDA ≥ 12 (not just exactly 12.x), **or**
- A `cuda13` / `cuda12+` Windows runtime variant should be shipped to cover CUDA 13.x drivers.

## Environment

| | |
|---|---|
| **mesh-llm** | 0.74.0 |
| **OS** | Windows 11 x86_64 |
| **GPU** | NVIDIA GeForce RTX 4090 (24564 MiB VRAM) |
| **NVIDIA driver** | 595.97 |
| **CUDA version (nvidia-smi)** | 13.2 |
| **Driver model** | WDDM |
| **Selected runtime** | meshllm-native-runtime-windows-x86_64-vulkan (fallback) |

`nvidia-smi` output confirming the GPU is fully visible to Windows:

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 595.97                 Driver Version: 595.97         CUDA Version: 13.2     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                  Driver-Model | Bus-Id          Disp.A | Volatile Uncorr. ECC |
|   0  NVIDIA GeForce RTX 4090      WDDM  |   00000000:01:00.0  On |                  Off |
|  0%   43C    P8             20W /  450W |    1547MiB /  24564MiB |     10%      Default |
+-----------------------------------------------------------------------------------------+
```

## Suspected Cause

The runtime compatibility check in the installer treats "CUDA 12" as an exact-major requirement and rejects the `cuda12` runtime when the driver reports CUDA 13.x. Because no `cuda13` Windows runtime is shipped in the v0.74.0 manifest, the installer silently falls back to the Vulkan runtime. The Vulkan runtime does not expose selectable CUDA devices, so `mesh-llm gpus` reports 0 GPUs even though the RTX 4090 is fully functional.

This is the **CUDA/NVIDIA** counterpart to issue #1022, which reported the same "0 GPUs" symptom via a different failure mode (missing MinGW DLLs in the Vulkan runtime on AMD hardware).

## Suggested Fix

One or more of the following:

1. **Relax the CUDA version gate** — accept the `cuda12` runtime when the driver reports any CUDA ≥ 12.x (NVIDIA's compatibility guarantee covers this).
2. **Ship a `cuda13` Windows runtime** — add `meshllm-native-runtime-windows-x86_64-cuda13` to the release manifest alongside the existing `cuda12` variant.
3. **Surface a warning instead of a silent fallback** — if neither a CUDA nor a ROCm runtime can be installed, emit a clear error explaining *why* (e.g. "CUDA 13 driver detected but only CUDA 12 runtime is available") rather than silently selecting Vulkan and leaving the user with a CPU-only node.

michaelneale:
## What this fixes

A machine with an NVIDIA driver that advertises a newer CUDA than the toolkit actually installed on it can now start and use its GPU.

Before this change, such a host picked the `cuda13` native runtime and died at startup:

```
libcudart.so.13: cannot open shared object file: No such file or directory:
load native runtime meshllm-native-runtime-linux-x86_64-cuda13
```

On Windows the same misdetection silently fell back to Vulkan, which then reported 0 GPUs (#1127).

This is common: NVIDIA drivers routinely support a newer CUDA than the toolkit installed alongside them. On the box I hit this on, `nvidia-smi` reports `CUDA Version: 13.0` while the only runtime library present is `libcudart.so.12`.

`mesh-llm runtime list --available` now also explains the mismatch instead of stating only the requirement:

```
before: CUDA toolkit mismatch: runtime requires CUDA 12
after:  CUDA toolkit mismatch: runtime requires CUDA 12, host has CUDA 13 installed
```

## Why it happened now

0.74.0 shipped a ~340 MB monolithic binary that linked `libcudart.so.12` directly, so it drove the GPU through its own linkage and the misdetection was harmless.

0.75.0 is the ~90 MB backend-neutral host with no CUDA linkage — **all** CUDA now comes from the resolved native runtime. The detection bug is not new, but it only became fatal once runtime selection had to be correct.

## Architecture

`HostCudaProfile` conflated two different facts. They are now separate:

| field | source | meaning |
|---|---|---|
| `toolkit_majors` | installed `libcudart.so.<major>` sonames, `/usr/local/cuda-<version>` | what can actually be loaded |
| `driver_max_major` | `nvidia-smi` | upper bound only |

Linux runtimes link `libcudart`/`libcublas` without bundling them, so they need a matching host toolkit. Windows runtimes ship their own copies (`cudart64_12.dll`) and are self contained — verified against the v0.75.0 release manifest.

Selection accepts a runtime when its major is within the driver bound **and** either the artifact bundles its own CUDA libraries or a matching toolkit is installed. With no installed-toolkit evidence at all it falls back to the driver bound, so hosts we cannot probe are not left with zero CUDA candidates.

## Validation

Reproduced and confirmed on a real RTX 3090 host (driver 580.173.02 advertising CUDA 13.0, toolkit 12.6 only):

```
$ ldconfig -p | grep -oE 'libcudart\.so\.[0-9]+'
libcudart.so.12

$ objdump -p .../cuda13/lib/libggml-cuda.so.0 | grep NEEDED
  NEEDED  libcudart.so.13     <- not bundled, not installed
```

Forcing the correct selection made 0.75.0 serve normally on that node — 12 peers, GPU at 58%, inference returning `OK`. This PR makes that selection happen automatically.

Three regression tests added, covering the newer-driver/older-toolkit case, self-contained Windows runtimes on a newer driver, and the no-evidence fallback.

```
cargo test -p mesh-llm-native-runtime --lib     # 29 passed
cargo test -p mesh-llm-hardware-profile --lib   # 17 passed
cargo clippy -p mesh-llm-native-runtime -p mesh-llm-hardware-profile -p mesh-llm-commands --all-targets -- -D warnings
cargo fmt --all --check
```

## Note on CI coverage

CI never caught this because all 18 CI/release references set `MESH_LLM_CUDA_TOOLKIT_MAJOR` explicitly, so auto-detection is never exercised, and no CI host has the driver-newer-than-toolkit shape. Worth considering a job that leaves detection unset — happy to add it here or follow up separately.

Fixes #1127

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA runtime selection by separating installed toolkit versions from driver-supported limits.
  * Prevented runtimes requiring newer CUDA versions than the host driver supports.
  * Improved support for self-contained runtimes and hosts without detectable toolkit information.
  * Added clearer compatibility messages, including required libraries, detected toolkit versions, driver limits, and environment configuration guidance.

* **Documentation**
  * Clarified CUDA toolkit detection, driver requirements, runtime loading behavior, environment overrides, and common compatibility failures.


crates/mesh-llm-hardware-profile/src/lib.rs
```diff
@@ -304,24 +304,73 @@ fn detect_cuda_profile(gpus: &[HostGpuProfile]) -> Option<HostCudaProfile> {
...
+///
+/// Linux native runtimes link against `libcudart.so.<major>` without bundling
+/// it, so this is the set that can really be loaded. Windows runtimes ship
+/// their own copies and do not depend on this.
+fn installed_cuda_toolkit_majors() -> BTreeSet<u32> {
+    let mut majors = BTreeSet::new();
+    if let Some(output) = command_output("ldconfig", &["-p"]) {
+        majors.extend(cuda_majors_from_soname_listing(&output));
+    }
+    for base in ["/usr/local/cuda", "/usr/local"] {
```
i386: **Verify runtime libraries, not just the toolkit directory name.** A `/usr/local/cuda-*` directory can exist without usable `libcudart`/`libcublas`/`libcublasLt` files (for example, a partial or stale installation). This code would then add the major even though `ldconfig` found no runtime library, and the resolver can select an unbundled Linux artifact that still fails at load time. The repository’s `scripts/lib/cuda-toolkit.sh::cuda_toolkit_library_dir` already requires all three libraries; please use equivalent file checks here before treating the major as installed.

i386 (commented):
I reviewed the PR head and ran the focused native-runtime and hardware-profile test suites (26 and 17 tests passed). The four existing CodeRabbit comments identify valid issues; I added one additional correctness comment about directory-only CUDA toolkit detection.

crates/mesh-llm-hardware-profile/src/lib.rs
```diff
@@ -304,24 +304,73 @@ fn detect_cuda_profile(gpus: &[HostGpuProfile]) -> Option<HostCudaProfile> {
...
 }
 
+/// CUDA toolkit majors with runtime libraries actually present on this host.
+///
+/// Linux native runtimes link against `libcudart.so.<major>` without bundling
+/// it, so this is the set that can really be loaded. Windows runtimes ship
+/// their own copies and do not depend on this.
+fn installed_cuda_toolkit_majors() -> BTreeSet<u32> {
+    let mut majors = BTreeSet::new();
+    if let Some(output) = command_output("ldconfig", &["-p"]) {
```
i386, line 350: [P2] Probe configured CUDA library roots too. `ldconfig -p` plus `/usr/local/cuda-*` misses toolkits installed under a custom prefix or exposed only through `LD_LIBRARY_PATH`/`CUDA_HOME` (for example Conda or `/opt/cuda-12`). On a host with `libcudart.so.12` in one of those paths and a driver advertising CUDA 13, `toolkit_majors` stays empty and the resolver falls back to selecting cuda13—the load failure this PR is intended to prevent. Please include the loader/environment search paths (with the same complete-library validation) or explicitly handle this installation shape.

i386 (commented):
Found one additional compatibility gap in toolkit discovery; inline below.

micspiral:
Thanks both — the review found a real hole in the original approach, and one of my own tests was fake. Pushed `a4d16e18`.

## First, a correction to my own PR description

I claimed CI missed this because "all 18 CI references set `MESH_LLM_CUDA_TOOLKIT_MAJOR` explicitly, so auto-detection is never exercised". I checked properly and **the stated reason was wrong**. Those 18 references are build/packaging-time stamping — consumed by `scripts/package-native-runtime.sh::cuda_toolkit_major()` to record `toolkit_major` in the artifact manifest. They are not suppressing host detection.

The conclusion still holds, for a simpler reason: **no CI job invokes host runtime selection at all**. There are no `runtime list --available`, `runtime install`, or `host_runtime_profile` calls anywhere in the workflows. So the selection path is unit-tested only, never exercised end-to-end on a real GPU host. Apologies for asserting the mechanism without verifying it.

## @i386 — verify libraries, not directory names

Accepted, and this was the more serious of the two. Directory-name scanning is gone entirely. Detection now requires the complete `libcudart` + `libcublas` + `libcublasLt` set at a matching major, mirroring `scripts/lib/cuda-toolkit.sh::cuda_toolkit_library_dir`. A partial or stale install no longer counts, and `libcudart` alone no longer counts either — which was a second unsoundness in my original version that your comment led me to.

Verified against the host that triggered this, which does satisfy the full triple:

```
$ ldconfig -p | grep -oE 'lib(cudart|cublasLt|cublas)\.so\.[0-9]+' | sort -u
libcublas.so.12
libcublasLt.so.12
libcudart.so.12
```

## @i386 — [P2] custom prefixes

Partly addressed, and I want to be explicit about where I landed rather than quietly not doing it.

`LD_LIBRARY_PATH` is now probed alongside the `ldconfig` cache, which covers the conda case and any custom prefix that is actually exposed to the loader. That host has a non-empty `LD_LIBRARY_PATH`, so this is not a hypothetical.

I deliberately did **not** probe `CUDA_HOME`/`CUDA_PATH`/`/opt/cuda-*` as evidence. A toolkit sitting in a directory the loader does not search cannot be `dlopen`ed by the runtime, so counting it as "installed" would assert loadability we do not have. The probe is intentionally the loader's view — cache plus `LD_LIBRARY_PATH` — because that is the same view the runtime load will use.

Your P2 also identified the case that made me change the design: on such a host `toolkit_majors` stayed empty, my fallback accepted, and it selected cuda13 — reproducing the bug the PR exists to fix. That is now fail-closed (below), so the outcome is a clear message plus CPU fallback rather than a cryptic `dlopen` failure.

## @coderabbitai — fail-open fallback

Accepted, and taken further than suggested. You proposed using the fallback only when `driver_max_major` is present. That still is not safe: driver 13 with an undetected toolkit admits cuda13, which is exactly the failure mode. So the empty-`toolkit_majors` case is now **always** a rejection for host-linked runtimes, with a distinct `CudaToolkitNotDetected` reason naming the libraries searched for and the `MESH_LLM_CUDA_TOOLKIT_MAJORS` escape hatch.

The fail-open behaviour existed so unprobeable hosts were not left with zero CUDA candidates. Added `undetected_toolkit_falls_back_to_cpu_runtime` to prove they still get a working runtime rather than nothing — the node serves, just on CPU, with the reason visible in `runtime list --available`.

## @coderabbitai — Windows test and README

Both correct, both fixed. The Windows test was genuinely fake: it ran against a Linux host profile and a Linux artifact platform, so it never exercised Windows selection at all — it only changed the artifact ID string. It now sets `host.os` and `artifact.platform.os` to `windows`. README example gained the missing `driver_max_major` field (worth noting it is not a compiled doctest, so this would not have been caught by CI).

## @coderabbitai — explicit self-containment flag

Agreed the `libraries` heuristic is inference rather than declaration. Tightened it to require the full bundled triple, so declaring `cudart` alone no longer claims independence from the host — that was unsound as written.

I have **not** added an explicit `toolkit_linkage: host | bundled` manifest field in this PR. It is the right end state, but it means a manifest schema change plus a compatibility path for already-published manifests, which I would rather not fold into a fix targeting a live startup failure. Happy to follow up separately if you want it, or do it here if you would prefer one change.

## Out of scope, worth a separate look

`min_driver` exists on `CudaRuntimeRequirements` but is never evaluated — only `min_driver: None` constructions appear in the tree. Driver comparison is major-only, so a minimum driver version in a manifest currently has no effect. Not touching it here.

The deeper fix both of you are circling is to stop inferring loadability and verify it — probe the candidate's full library set in a short-lived subprocess before declaring it compatible, which makes probe accuracy irrelevant. That is a larger change and interacts with the pre-download selection path (`runtime install` chooses before the artifact exists on disk), so I would want it as its own PR.

## Validation

```
cargo test -p mesh-llm-native-runtime --lib     # 31 passed
cargo test -p mesh-llm-hardware-profile --lib   # 17 passed
cargo clippy -p mesh-llm-native-runtime -p mesh-llm-hardware-profile -p mesh-llm-commands --all-targets -- -D warnings
cargo fmt --all --check
```

Six regression tests now cover: newer-driver/older-toolkit, self-contained Windows runtime on a newer driver, partially-bundled artifact not treated as self-contained, host-linked rejected without a detected toolkit, and CPU fallback in that case.

i386 (approved):
Reviewed and fixed the actionable CUDA comments. The change now discovers CUDA evidence from configured/custom roots, validates actual library targets against the host ELF architecture, and reads only the required ELF header bytes instead of loading whole shared libraries. Validated with `cargo fmt --all -- --check`, `cargo check -p mesh-llm-hardware-profile`, `cargo clippy -p mesh-llm-hardware-profile --all-targets -- -D warnings`, and `cargo test -p mesh-llm-hardware-profile --lib` (18 passed). Approving.

michaelneale (commented):
🤖 (micn's agent) Reviewed against the v0.75.0 release manifest and ran the test suites locally. The core design is right and this is needed for Linux GPU nodes on 0.75.x — the toolkit/driver split matches reality: I verified all four Linux CUDA artifacts in the published `native-runtimes.json` bundle **no** cudart/cublas libraries while `windows-x86_64-cuda12` bundles all three, so `artifact_bundles_cuda_runtime` keys off exactly the right evidence. Local runs: `mesh-llm-native-runtime` 31 passed, `mesh-llm-hardware-profile` 18 passed, clippy `-D warnings` clean on all three touched crates.

Findings, in descending severity:

**1. `valid_cuda_library_target` reads the entire library into memory to check 20 header bytes.**

```rust
fn valid_cuda_library_target(path: &Path) -> bool {
    let Ok(bytes) = fs::read(path) else { return false; };
    if bytes.len() < 20 || &bytes[..4] != b"\x7fELF" { ... }
```

`libcublasLt.so.12` is routinely 400–700 MB and `libcublas.so.12` 100+ MB. This runs during host profile detection — i.e., startup — potentially several times per library (ldconfig line + each matching search dir). Please read just the first 20 bytes (`File::open` + `read_exact` into a `[u8; 20]`). Same correctness, avoids transiently allocating ~1 GB on CUDA hosts.

**2. PR description contradicts the code on the no-toolkit fallback.** The body says *"With no installed-toolkit evidence at all it falls back to the driver bound, so hosts we cannot probe are not left with zero CUDA candidates"* — but the code (and the `host_linked_runtime_is_rejected_without_detected_toolkit` test) deliberately **rejects** host-linked runtimes when `toolkit_majors` is empty, falling back to CPU. I think the code is right and the description is stale from an earlier iteration; the code comment ("reject rather than guess") gives the correct rationale. Worth fixing the description before merge so 0.75.1 release notes don't inherit the wrong claim.

**3. Doc/code mismatch on directory scanning.** The doc comment on `installed_cuda_toolkit_majors` says the probe *"deliberately"* uses only *"the loader's own view — the `ldconfig` cache and `LD_LIBRARY_PATH` — rather than guessing from installation directory names"*, but `loader_search_dirs()` then scans `CUDA_HOME`/`CUDA_PATH`/`CUDA_ROOT`/`CONDA_PREFIX` and everything under `/usr/local/cuda*`. The ELF validation makes that scan safe against empty/wrong-arch dirs, so the behavior is fine — but the comment should be updated to describe what it actually does.

Related question, non-blocking: if the only toolkit evidence is `/usr/local/cuda-12/lib64` and that dir is *not* in the ldconfig cache or `LD_LIBRARY_PATH`, we mark cuda12 as installed and select the cuda12 runtime — can the runtime actually resolve `libcudart.so.12` at load in that case (RPATH? explicit dlopen paths?), or do we reintroduce the same `cannot open shared object file` failure one step later? If load would fail, evidence from non-loader dirs may need to also extend the load path, or be excluded.

**4. SDK surface note.** `CandidateRejection::CudaToolkitMajorMismatch` gained an `installed` field and two new variants were added. The enum is re-exported through `mesh-llm-sdk` / `mesh-llm-api` surfaces, so exhaustive matches in downstream consumers will break at compile time. Fine for a minor release, just deserves a line in the changelog/PR description.

**5. Small nit:** `driver_max_major` takes `next_back()` of all majors parsed from `nvidia-smi` output. On multi-GPU boxes with mixed driver stacks (rare, but Jetson + dGPU exists) the max across GPUs is the right call for an upper bound, so this is correct — just noting it was considered.

None of these are architectural objections — #1 is the only change I'd insist on before merge, #2 is a description fix, #3 a comment fix. Overall this correctly turns a startup crash into correct selection, and the rejection messages in `runtime list --available` are a genuine debuggability win.

michaelneale:
🤖 (micn's agent) Post-merge review against `e359ca77` on main — verdict: **keep it, no back-out needed**. The merged behavior is sound for the vast.ai-style case that motivated it.

What I verified in the merged code:

- ✅ **The full-file read finding was fixed before merge** — `valid_cuda_library_target` now opens the file and reads exactly 20 header bytes instead of `fs::read`-ing multi-hundred-MB cublas libraries into memory. Thanks for the quick turnaround on that.
- ✅ **Selection logic matches the failure I'd expect on a vast.ai box**: containers there typically expose a new driver (`nvidia-smi` → CUDA 13.x) via the container toolkit while the image ships a CUDA 12 toolkit under `/usr/local/cuda*`. The merged probe finds the toolkit through the `/usr/local/cuda*` dir scan even when `ldconfig` doesn't list it, so cuda12 gets selected instead of the fatal cuda13. And a GPU container with *no* toolkit at all now falls back to CPU with an explicit, actionable rejection (`no CUDA toolkit detected ... set MESH_LLM_CUDA_TOOLKIT_MAJORS`) rather than dying at dlopen — strictly better than 0.75.0 in every configuration I can construct.
- ✅ The self-contained Windows path is correctly keyed off actual bundled libs (verified against the v0.75.0 release manifest: all four Linux CUDA artifacts bundle none, windows-cuda12 bundles all three).

Two residual items, neither back-out-worthy:

1. **Residual risk to verify on a real box** (my earlier open question, still open): when toolkit evidence comes *only* from a non-loader dir (e.g. `/usr/local/cuda-12/lib64` absent from ldconfig cache and `LD_LIBRARY_PATH`), selection says cuda12 is installed — but can the selected runtime's `libggml-cuda.so` actually resolve `libcudart.so.12` at load time from that location? If the loader can't see it, we've moved the failure from selection to load. A vast.ai box is the perfect place to test this exact shape: `mesh-llm runtime list --available` + a real serve. If it fails, the fix is small (either extend the load path with the evidence dir, or require loader-visible evidence) — flagging so it's on the 0.75.1 radar rather than discovered in the wild again.
2. **Doc drift retained in the merge**: the doc comment on `installed_cuda_toolkit_majors` still claims it "deliberately" probes only "the loader's own view ... rather than guessing from installation directory names," while `loader_search_dirs()` scans `CUDA_HOME`/`CUDA_PATH`/`CUDA_ROOT`/`CONDA_PREFIX` and `/usr/local/cuda*`. Likewise the PR body's "with no installed-toolkit evidence at all it falls back to the driver bound" describes behavior the code deliberately doesn't have (it rejects + falls back to CPU — correctly). Both worth a tiny follow-up commit so future readers trust the comments.

hola-there:
Thanks!

PR #1196, opened 2026-08-08
fix: stale pre-0.75 runtime cache entries no longer break startup for SDK embedders

michaelneale:
## Summary

Upgrading to 0.75.x with an existing native runtime cache no longer aborts startup for SDK embedders. A cache entry written by a pre-0.75 loader (manifest without per-file checksums) is now skipped with a warning during startup runtime resolution, instead of failing the whole cache scan before the download/install fallback can run.

This is the startup-path completion of #1170 (issue #1162). #1170 fixed the resolver's cache enumeration, but the startup fast-path in `mesh-llm-host-runtime` still used the strict full-cache `NativeRuntimeCache::installed()`, where one stale manifest `?`-aborts everything.

## Who hits this

Anyone embedding the host runtime via the SDK (`initialize_host_runtime()`) on a machine with a pre-0.75 runtime cache — observed during the Buzz desktop upgrade to v0.75.0:

```
native runtime artifact meshllm-native-runtime-darwin-aarch64-metal does not declare file checksums
```

The installed CLI was unaffected *by accident of packaging*: `install.sh` ships a `native-runtimes/` directory beside the binary, so startup takes the bundle-discovery branch (which reads manifests leniently) and never reaches the strict cache scan. SDK embedders have no adjacent bundle dir, fall into `resolve_installed_native_runtime_plan` → `cache.installed()`, and failed hard — before `allow_download` could rescue them.

## What changed

- New `NativeRuntimeCache::installed_lenient()` enumerates the whole cache and collects unreadable entries (missing/malformed/checksum-failing manifests) as `skipped` instead of failing the scan.
- Startup runtime resolution (`resolve_installed_native_runtime_plan`) uses the lenient scan and emits a `tracing::warn!` per skipped entry with path and reason.
- The runtime-control backend options view uses it too, so one stale entry no longer hides all installed backend kinds from the console.
- Strict verification is unchanged where it matters: installs, downloads, and the load plan for the selected runtime still enforce per-file checksums. Only *enumeration of neighbors* is lenient.

## Validation

- New test `stale_pre_checksum_cache_entry_does_not_block_startup_plan` (host runtime): a valid current-version runtime plus a pre-checksum 0.74.0 cache entry resolves to the valid runtime.
- New test `installed_lenient_skips_legacy_manifest_and_keeps_valid_runtime` (native-runtime crate): lenient scan returns the valid runtime, reports the stale one as skipped, and asserts the strict scan still errors on the same cache (documenting why startup must not use it).
- `cargo test -p mesh-llm-host-runtime --lib` — 1924 passed.
- `cargo test -p mesh-llm-native-runtime --lib` — 27 passed.
- `cargo fmt --all --check`, `cargo clippy -p mesh-llm-native-runtime --all-targets -- -D warnings` — clean. (`-p mesh-llm-host-runtime` clippy has 28 pre-existing unfulfilled-lint-expectation errors on main, unchanged by this PR.)
- Empirical repro on macOS arm64: a binary with no adjacent bundle dir and an isolated cache containing only a stale pre-checksum 0.74.0 entry — v0.75.0 fails startup with the checksum error; this branch warns, skips, and proceeds to normal runtime resolution.

## Protocol

No wire or protocol impact. Cache-directory compatibility only: pre-0.75 cache entries are now skipped (and pruned by existing maintenance) rather than fatal. No manifest format change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime discovery so valid runtimes continue loading even when unusable or legacy cache entries are present.
  * Added clearer logging when invalid runtime entries are skipped.
  * Prevented legacy pre-checksum cache entries from blocking startup or runtime state detection.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/api/routes/runtime_control_state_sources.rs
+++ b/crates/mesh-llm-host-runtime/src/api/routes/runtime_control_state_sources.rs
@@ -80,9 +80,9 @@ fn gpu_state(gpus: &[crate::system::hardware::GpuFacts]) -> RuntimeOptionsState 
 fn native_backend_state() -> RuntimeOptionsState {
     let mut kinds = crate::system::native_runtime_install::host_runtime_profile().available_flavors;
     if let Ok(cache) = crate::system::native_runtime_install::default_native_runtime_cache()
-        && let Ok(installed) = cache.installed()
+        && let Ok(scan) = cache.installed_lenient()
     {
-        for runtime in installed {
+        for runtime in scan.runtimes {
             kinds.insert(runtime.manifest.runtime.backend.kind);
         }
     }
--- a/crates/mesh-llm-host-runtime/src/system/native_runtime.rs
+++ b/crates/mesh-llm-host-runtime/src/system/native_runtime.rs
@@ -278,7 +278,15 @@ mod dynamic {
         target_skippy_abi: Option<&str>,
         selection: &RuntimeSelection,
     ) -> Result<Option<NativeRuntimeStartupLoadPlan>> {
-        let installed = cache.installed()?;
+        let scan = cache.installed_lenient()?;
+        for skipped in &scan.skipped {
+            tracing::warn!(
+                path = %skipped.path.display(),
+                reason = %skipped.reason,
+                "Skipping unusable native runtime cache entry during startup"
+            );
+        }
+        let installed = scan.runtimes;
         if installed.is_empty() {
             return Ok(None);
         }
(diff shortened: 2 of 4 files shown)


i386 (approved):
Reviewed and fixed the remaining cache-scan robustness issue. The lenient startup scan now skips unreadable cache-version directories, runtime-entry metadata failures, and iterator item failures where no path is available, while continuing to discover valid runtimes. The preceding PR commit already handles missing/non-regular manifests as skipped entries. Validated with `cargo fmt --all -- --check`, `cargo check -p mesh-llm-native-runtime`, `cargo clippy -p mesh-llm-native-runtime --all-targets -- -D warnings`, and `cargo test -p mesh-llm-native-runtime --lib` (28 passed). `mesh-llm-host-runtime` check is green; its broad Clippy run is currently blocked by pre-existing unrelated `unfulfilled_lint_expectations` warnings. Approving.

micspiral:
Desk-checked this from Buzz (SDK embedder) — it fixes the startup abort for us.

**Setup:** pinned Buzz's mesh-llm deps to this branch's head (`4f80cad`) and left a
pre-checksum cache in place, i.e. the failing upgrade shape:

```
~/Library/Caches/mesh-llm/native-runtimes/
  0.73.1      runtime.files = 0     <- written by a pre-0.75 loader
  0.74.0      runtime.files = 0
  0.74.0-rc8  runtime.files = 0
  0.75.0      runtime.files = 24    <- valid, correct version, present
```

**Before** (pinned to the `v0.75.0` tag), `initialize_host_runtime()` aborted:

```
host runtime init: native runtime artifact meshllm-native-runtime-darwin-aarch64-metal
does not declare file checksums
```

Buzz has no executable-adjacent bundle dir (the desktop app doesn't ship one), so it
falls through to `resolve_installed_native_runtime_plan` → `cache.installed()`, and a
single stale entry took out the whole scan even though the valid `0.75.0` entry was
sitting right there. That's also why the installed CLI is unaffected: its bundle dir
means the cache enumeration is never reached.

**After** this branch: no abort, runtime resolved, and our end-to-end
share-compute → agent → inference example passed 4/4 with the stale entries still in
place.

**One extra step needed to test from a hash**, worth noting for anyone else doing this:
the branch's `CARGO_PKG_VERSION` is `0.72.1`, so `default_release_manifest_url` pointed
at the v0.72.1 release, which ships Skippy ABI `0.1.27` while the build wants `0.1.35`:

```
no compatible MeshLLM native runtime is installed or installable
for MeshLLM 0.72.1 / Skippy ABI 0.1.35
```

Setting `MESH_LLM_BUILD_VERSION=0.72.1+g4f80cad` takes the `is_sha_build` path in
`default_manifest_url`, which resolves from `releases/latest` and picks up the ABI
`0.1.35` artifact. Not a defect in this PR — just a consequence of pinning an
unreleased rev.

**What I did not verify:** the new `Skipping unusable native runtime cache entry`
warnings. Our example installs no tracing subscriber, so warn-level lines aren't
printed. I only observed the absence of the fatal error plus a successful start, not
the skip logging itself.

PR #1205, opened 2026-08-08
feat(openai): configure trusted agent session header

i386:
## Problem

The OpenAI frontend supports a configurable trusted agent-session header, but the default router ignored that configuration. A Mesh process launched with a trusted upstream session-header setting therefore treated every request as stateless.

## Change

`OpenAiFrontendConfig::default()` now reads the optional `MESH_AGENT_SESSION_HEADER` environment variable and applies it as the trusted header. Existing callers remain unchanged when the variable is absent, and explicit `with_agent_session_header` configuration still takes precedence.

Invalid or non-UTF-8 values emit a startup warning and are ignored rather than silently looking like a valid configuration. Parser coverage covers valid and invalid names.

This keeps the boundary generic: Mesh does not hard-code Cacheline or a benchmark-specific header name.

## Validation

- `cargo test -p openai-frontend --lib` (153 passed)
- `cargo clippy -p openai-frontend --lib --all-targets -- -D warnings`

### DIFF
--- a/crates/openai-frontend/README.md
+++ b/crates/openai-frontend/README.md
@@ -50,6 +50,7 @@ For the concrete benchy command and contract, see
 | Request body limit | Supported | Configurable via `OpenAiFrontendConfig`; defaults to 4 MiB. |
 | Request IDs | Supported | Propagates or generates `x-request-id`, returns it on every response, and emits a tracing event with method, URI, status, and request ID. |
 | Backend timeout | Supported | Configurable via `OpenAiFrontendConfig`; defaults to 300 seconds and maps timeouts to OpenAI-shaped 504 errors. |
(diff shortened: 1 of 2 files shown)


i386:
Addressed the review nit in 7330de11. `MESH_AGENT_SESSION_HEADER` now distinguishes unset, non-UTF-8, and invalid header values; invalid configuration emits a startup warning instead of silently disabling session identity. Added parser coverage for valid and invalid names. Validation: `cargo fmt --all -- --check`, `cargo test -p openai-frontend --lib` (153 passed), and clippy with `-D warnings`.

ndizazzo (approved):
Ah perfect, this shoukd help us let models dispatch from codex to other CLI tools

PR #1209, opened 2026-08-08
fix: remove stale ngram app-tab option

### LINKED ISSUE #1056: Console app-tabs still offers speculation mode = ngram after it was removed from the config schema
## Problem

PR #1026 removes `ngram` as a valid value of the `speculative.mode` enum in the backend:

- `crates/mesh-llm-config/src/model/built_in_schema.rs` — `speculative.mode` `string_enum` narrowed to `["auto", "disabled", "draft"]`
- `crates/mesh-llm-config/src/model_validation.rs` — validator's allowed-values list dropped `"ngram"`
- The `features/configuration/` UI tests were updated to match.

But the **other** UI surface that defines the same dropdown was not updated:

- `crates/mesh-llm-ui/src/features/app-tabs/data.ts` still lists `{ value: 'ngram', label: 'n-gram' }` as a selectable speculation mode.
- `crates/mesh-llm-ui/src/features/app-tabs/data.test.ts` still asserts `choiceValues('speculation-mode')` equals `['auto', 'disabled', 'draft', 'ngram']`.

`app-tabs/` was not part of #1026's diff at all, so these are leftovers rather than a deliberate keep.

## Why it's a problem

Within the same build, one UI surface still lets a user pick `ngram`, which writes `mode = "ngram"`, which the backend validator now rejects — a broken round-trip. The `app-tabs` test also drifts out of sync with the schema.

## Fix

Remove the `ngram` option from `app-tabs/data.ts` and update the `data.test.ts` assertion to `['auto', 'disabled', 'draft']`, matching the `configuration/` surface and the backend schema. This finishes the removal in the same direction the rest of #1026 already went (n-gram moves from a top-level speculation mode to a proposer/strategy).

## Context

- Follows on from #1026 (removed `ngram` mode from the backend schema/validator + `configuration/` UI, but not `app-tabs/`).

i386:
Fixes #1056.

The app-tabs configuration surface still exposed `ngram` even though the backend schema and configuration UI no longer accept it. This removes the stale option and updates the app-tabs regression test to keep the UI aligned with the supported values: `auto`, `disabled`, and `draft`.

Validation:
- `pnpm exec vitest run src/features/app-tabs/data.test.ts`
- `just build`

## Summary by CodeRabbit

- **Updates**
  - Removed the obsolete **n-gram** option from the speculation mode setting.
  - Speculation mode now supports **Auto**, **Disabled**, and **Draft**.

### DIFF
--- a/crates/mesh-llm-ui/src/features/app-tabs/data.test.ts
+++ b/crates/mesh-llm-ui/src/features/app-tabs/data.test.ts
@@ -142,7 +142,7 @@ describe('CONFIGURATION_DEFAULTS', () => {
 
   it('uses current speculation values and includes incoming draft tuning controls', () => {
     expectChoice('speculation-mode', 'mode', 'auto', 'segmented')
-    expect(choiceValues('speculation-mode')).toEqual(['auto', 'disabled', 'draft', 'ngram'])
+    expect(choiceValues('speculation-mode')).toEqual(['auto', 'disabled', 'draft'])
     expectChoice('incompatible-pairing-behavior', 'pairing_fault', 'warn_disable', 'toggle')
     expect(choiceValues('incompatible-pairing-behavior')).toEqual(['warn_disable', 'fail_closed'])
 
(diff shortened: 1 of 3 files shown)


ndizazzo (approved):
I really love the programmatic settings schema

PR #1211, opened 2026-08-08
feat: pass reasoning effort through Skippy templates

### LINKED ISSUE #986: Pass reasoning effort through Skippy to llama.cpp chat templates
## Feature

Pass OpenAI request-level reasoning controls through Skippy to llama.cpp's native chat-template renderer, preserving model-supported reasoning modes and generic `chat_template_kwargs`.

## Problem or use case

GLM-5.2 exposes non-thinking, `high`, and `max` thinking modes. The documented llama.cpp interface selects them through chat-template kwargs.

The staged OpenAI path cannot currently select those modes:

- `crates/skippy-server/src/frontend/request.rs::chat_template_options` ignores the request and only sets the output reasoning format.
- Existing Skippy tests explicitly assert that `reasoning_effort`, `enable_thinking`, and `chat_template_kwargs` do not affect template rendering.
- `crates/openai-frontend/src/common.rs::ReasoningEffort` does not represent `max`, and its template normalization reduces non-disabled efforts to a boolean.

As a result, coding and agent clients always receive the model template's default thinking behavior, even when the OpenAI request asks for a different supported mode.

Reference: https://unsloth.ai/docs/models/glm-5.2

## Desired behavior

- Preserve request-level `reasoning_effort`, including `high`, `max`, and disabled/non-thinking behavior.
- Pass explicit generic `chat_template_kwargs` to llama.cpp.
- Let explicit request kwargs take precedence over model/package defaults.
- Keep llama.cpp as the sole owner of model-specific template rendering and control-token behavior.
- Do not add GLM-specific prompt strings, token injection, or template compatibility logic in Skippy or mesh-llm.
- Apply the same behavior across the OpenAI chat-completions and responses paths.

## Acceptance criteria

- `reasoning_effort: "high"` and `reasoning_effort: "max"` reach llama.cpp with distinct template kwargs.
- Disabled reasoning reaches llama.cpp as the appropriate template option.
- Explicit `chat_template_kwargs` are preserved and have documented precedence.
- Prompt-rendering tests prove high/max/off produce the llama.cpp-rendered model behavior.
- Existing model families retain their current default behavior when no reasoning override is supplied.
- Full-model and staged Skippy prompt rendering remain equivalent for the same request.

## Extra context

This is transport and API plumbing, not model-specific prompting. The model's embedded template and llama.cpp must remain the source of truth.

i386:
## Summary

- Pass OpenAI `reasoning_effort` values, including `max`, through the staged chat path as llama.cpp chat-template kwargs.
- Preserve generic `chat_template_kwargs` and let explicit request kwargs override inferred effort values.
- Keep `/v1/models` reasoning metadata coarse; do not advertise supported effort values.
- Extend the Skippy JSON chat-template ABI with a generic kwargs object plus removal of the legacy chat-template entrypoint (ABI 0.1.35 → 0.1.37), stacked on #1194's consolidated patch queue.
- Cover chat-completions and Responses normalization, Skippy request precedence, native ABI transport, llama.cpp Inkling rendering, and model-listing behavior.

## Root cause

Skippy reduced request reasoning controls to `enable_thinking: bool` before invoking llama.cpp. That discarded model-supported effort levels and all non-boolean template kwargs, so the model template could not select its own `high`/`max` behavior.

## Skippy ABI inventory and compatibility

This PR changes one existing ABI function signature and removes one obsolete exported function. The native ABI is an internal lockstep boundary; older native runtimes are not supported.

| Status | Symbol/declaration | Public header | Implementation / Rust mirror | Reason | Lockstep update |
|---|---|---|---|---|---|
| Changed | `skippy_apply_chat_template_json(..., const char * reasoning_format, const char * chat_template_kwargs, ...)` | `include/skippy/tokenization.h` | `common/stage-chat.cpp`; `crates/skippy-ffi/src/lib.rs` | Carry an optional JSON object of model/template-owned kwargs into llama.cpp's `common_chat_templates_apply`, so controls such as `reasoning_effort` are not reduced to a boolean. | ABI patch `0.1.35 → 0.1.37`; backward compatibility with older native runtimes is not required—the Rust mirror and callers are updated in lockstep with the native queue. |
| Removed | `skippy_apply_chat_template` | `include/skippy/tokenization.h` | `common/stage-chat.cpp`; `crates/skippy-ffi/src/lib.rs` | Remove the duplicate typed-message ABI path; the Rust convenience method now serializes messages and uses the JSON ABI. | ABI patch `0.1.36 → 0.1.37`; old native runtimes are not supported. |
| Versioned | `SKIPPY_ABI_VERSION_PATCH` `35 → 37` | `include/skippy/common.h` | Mirrored in `crates/skippy-ffi/src/lib.rs` | Keep native/Rust ABI versions diagnosable when lockstep artifacts are accidentally mixed. | This is an intentional internal ABI revision; old native runtimes are not supported. |

Functions added: none. Function deleted/removed: `skippy_apply_chat_template`. No public header was deleted or renamed.

The native implementation removes the duplicate typed-message entrypoint and keeps the JSON path as the single chat-template ABI. It parses `chat_template_kwargs` as an object, rejects non-object JSON, and copies each key/value into llama.cpp's generic chat-template input map. The Rust FFI declaration, dynamic symbol type, static declaration, and call sites carry the same JSON ABI contract.

## Model reasoning capability advertisement

This PR deliberately does not return `reasoning.supported_efforts` from `/v1/models`.

Reasoning controls are owned by the model and its chat template. The valid values are not part of a stable GGUF or llama.cpp-wide vocabulary: GPT-OSS documents `low`, `medium`, and `high` but its template accepts an arbitrary string; Qwen3 uses a boolean `enable_thinking` switch; and other templates can define different enums or numeric ranges.

Mesh therefore keeps the existing coarse `reasoning` capability and `reasoning_status` fields, while leaving effort selection to the request path and the model template. Advertising a static list would claim support that mesh has not verified and would encourage clients to send invalid values. Model-owned capability metadata can be added later if there is a reliable, machine-readable source; it is intentionally out of scope for this PR.

## Validation

- `LLAMA_WORKDIR="$PWD/.deps/llama.cpp" scripts/prepare-llama.sh pinned` — patch queue applies cleanly on the #1194 base.
- `just llama-build` — patched CPU native runtime builds with ABI 0.1.37 and the legacy symbol removed.
- `just build` — debug host, UI, and dynamic Metal native runtime build/package pass.
- `cargo test -p openai-frontend --lib` — 153 passed.
- `cargo test -p skippy-server --lib frontend::tests::request` — 30 passed.
- `cargo test -p skippy-runtime --lib` — 70 passed, including the configured model-backed chat-template kwargs smoke when `SKIPPY_CORRECTNESS_MODEL` is available.
- `cargo test -p mesh-llm-host-runtime --lib network::openai::response::models` — 12 passed, including the regression that keeps model-listing effort values undisclosed.
- `cargo clippy -p openai-frontend -p skippy-ffi -p skippy-runtime -p skippy-server -p skippy-bench -p skippy-prompt --all-targets -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- Real `gpt-oss-20b-MXFP4.gguf` loaded through the patched native runtime; all seven request strings reached its native template. The GPT-OSS Harmony output format is not compatible with the generic DeepSeek-style `<think>` parser smoke, which is a parser-family limitation rather than an ABI kwargs failure.

The combined clippy command including the shipped `mesh-llm` binary is currently blocked by 28 pre-existing unfulfilled `#[expect(dead_code)]` lints in the #1194 host-runtime base; none are in this diff.

Fixes #986

## Summary by CodeRabbit

- **New Features**
  - Added support for custom chat-template parameters, including reasoning settings and application-defined values.
  - Added a comprehensive Skippy native API reference for developers.
- **Improvements**
  - Reasoning effort now supports `max`, with explicit template parameters taking precedence while preserving custom options.
  - Chat-template configuration consistently applies default and user-provided settings.
- **Bug Fixes**
  - Corrected model capability reporting and reasoning-template behavior across supported scenarios.
- **Breaking Changes**
  - Updated the native ABI and removed the legacy chat-template interface.


michaelneale (dismissed):
Yeah this is pretty important. How well covered is this with tests and smokes? as really need to nail this

i386:
@michaelneale I ran the focused coverage for this change:

- `cargo test -p openai-frontend --lib`: 153 passed
- `cargo test -p skippy-server --lib`: 384 passed
- `cargo test -p skippy-runtime --lib`: 70 passed, including the configured native-renderer smoke for `chat_template_kwargs`

The added tests cover:

- `reasoning_effort` values `none`, `high`, and `max` reaching the native template kwargs;
- explicit `chat_template_kwargs` overriding the top-level effort while preserving custom values;
- OpenAI chat/responses normalization and validation for `max`;
- the native renderer accepting generic kwargs;
- the llama.cpp Inkling template rendering distinct prompts for disabled, high, and max effort.

The replacement GitHub CI runs are still pending/in progress; the earlier failures were from the superseded/cancelled run. I am not claiming the broader platform matrix or a live multi-node/agent smoke until those checks complete. Given the ABI bump and llama.cpp patch, I agree those CI/native checks should be green before merge.

i386:
@michaelneale @ndizazzo I have a few more things I want to test and look at first... please do not merge

i386:
Additional native-model validation completed against the patched llama.cpp runtime.

- Downloaded and loaded the real `ggml-org/gpt-oss-20b-GGUF/gpt-oss-20b-MXFP4.gguf` (11.27 GiB) through `skippy-runtime`.
- `just llama-build` passed with the full patch queue.
- Native rendering was exercised for every PR value: `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, and `max`. Each produced a prompt containing the matching GPT-OSS `Reasoning: <value>` marker, so the kwargs survive all the way into the model's native chat template.
- Important distinction: GPT-OSS's template accepts the value as a string and does not validate the vocabulary. This proves transport, not that GPT-OSS supports all seven values; its documented/practical effort vocabulary is `low`, `medium`, and `high).
- The GGUF contains a native `tokenizer.chat_template`, but no separate structured reasoning-effort vocabulary in its metadata. This supports keeping discovery template/model-owned rather than adding another global vocabulary.
- The Inkling template fixture already carried in the patch queue defines the full seven-value mapping (including its numeric levels); the native queue test covers representative `none`, `high`, and `max` prompt differences.

One limitation surfaced: the generic `chat_reasoning_markers_are_stripped_and_extracted` smoke fails with GPT-OSS because GPT-OSS uses Harmony channels rather than DeepSeek-style `<think>` markers. That is a parser-family mismatch, not a failure to pass `reasoning_effort`; this test should continue to use a compatible Qwen/DeepSeek-style model.

micspiral:
Looks good overall. One thing before merge: can you show proof that `reasoning_effort: "none"` still renders a no-thinking prompt on GPT-OSS and a Qwen3-family template? The new generic kwargs loop runs after the alias block and overwrites patch 0057's deliberate `reasoning_effort = "0"` disable-hardening with the string `"none"`, and the GPT-OSS validation above was inconclusive by its own admission.

i386:
Updated in b498e454d.

The stage path now merges generic chat_template_kwargs first, then reasserts reasoning_effort=0 when enable_thinking is explicitly false. That means request-level reasoning_effort cannot undo the no-thinking alias hardening.

I also added a model-free regression in llama.cpp test-chat for the combined case against Qwen3 and GPT-OSS templates, plus a narrow --generation-prompt mode. It uses only checked-in Jinja templates, not GGUF/model fixtures, so it does not add large CI artifacts or model downloads.

i386:
@michaelneale should be ok now

ndizazzo (approved):

PR #1222, opened 2026-08-10
Report an accurate version from main builds

michaelneale:
## What this changes for you

Binaries built from `main` now report `0.76.0-dev` instead of `0.72.1`.

Before this, a fresh `just build` from `main` reported **0.72.1** — three releases behind the published **v0.75.1**. That directly undermined the deploy checklist in `AGENTS.md`, which tells you to verify `mesh-llm --version` on every node to confirm a binary is new code. A dev build claiming `0.72.1` was indistinguishable from a genuinely stale one.

The `-dev` suffix also makes a dev build self-identifying: if you see `0.76.0-dev` on a node, you know it is not a release binary.

Config settings that offer a mesh-llm version now list the versions that actually shipped — `0.75.1`, `0.75.0`, `0.74.0`, and `0.73.0` were all missing from `known_mesh_llm_versions()`, so recent releases could not be selected.

## Why main drifted

The Release workflow runs `scripts/release-version.sh` on a throwaway "prepare release source" commit, tags it, and pushes **only the tag** — never the version bump back to `main`. Every release tag carries the correct version (verified: `v0.73.0` → `0.73.0`, `v0.74.0` → `0.74.0`, `v0.75.0` → `0.75.0`), so published artifacts were always right, which is why this went unnoticed. Main's version was last touched by hand in `7a809522` (2026-06-30) — the commit that moved bumping into the workflow.

This PR is the manual catch-up. Automating a post-release bump-back is a separate change.

## Notes for review

Generated with `scripts/release-version.sh 0.76.0-dev`, then two hand corrections where the script's literal string replacement produced values that would be wrong:

- `known_mesh_llm_versions()` is a list of *released* versions used as a config validation enum. The script prepended `0.76.0-dev`; replaced with the four real releases that were missing.
- User-facing SDK/docs examples (`mesh-llm-sdk = "..."`, release download URLs, SwiftPM `from:`) were rewritten to `0.76.0-dev`, which will never exist on crates.io or as a release URL. Pointed at `0.75.1` instead.

`sdk/kotlin/build.gradle.kts`, the Kotlin JVM example, and `generate-swift-bindings.sh` intentionally keep `0.76.0-dev` — those are the projects' own version declarations and should track the workspace dev version.

## Validation

```
cargo fmt --all --check                                        # clean
cargo check -p mesh-llm-config                                 # ok
cargo clippy -p mesh-llm-config --all-targets -- -D warnings   # clean
cargo test -p mesh-llm-config --lib                            # 122 passed
cargo run -p xtask -- repo-consistency publish-crates          # passed
```

## Summary by CodeRabbit

* **Release Updates**
  * Updated the development release version to **0.76.0-dev** across the product and SDK packages.
  * Added support for recognizing Mesh LLM releases **0.75.0** and **0.75.1**.
  * Updated compatibility metadata to support runtime version **0.75.1**.

* **Documentation**
  * Updated installation, configuration, runtime, SDK, and CLI examples to reference release **0.75.1**.
  * Refreshed package and example version references across supported SDKs.


michaelneale:
Closing — wrong shape. This fixes the symptom (main's stale version) by hand-correcting generated output, and defers the actual cause (nothing keeps main's version honest, and `release-version.sh` rewrites user-facing docs to a dev version that will never exist).

Replacing with targeted changes: crates.io publish-map validation first (it has a live consequence — 0.75.1 published no crates), then a structural fix for the version drift.

PR #1176, opened 2026-08-04
feat(console): add typed logs ledger and maintenance UI

ndizazzo:
## Summary

Third and final PR in the logging stack, based on [#1175](https://github.com/Mesh-LLM/mesh-llm/pull/1175). This layer delivers the operator-facing logs console, request inspector, maintenance workflows, and privacy-safe output projections.

### System logs ledger

- Adds typed API schemas, identifier parsing, query hooks, SSE recovery, and capability-aware live/harness clients.
- Adds a combined request and operational-event ledger with search, time/category/state filters, sortable/resizable columns, column visibility, pagination, and bounded export.
- Adds request-volume charts and outcome KPIs with chronological, outcome-specific trends.
- Caps rendered chart buckets and automatically selects a safe interval for large/all-time windows.
- Keeps relative live filters rolling and advances the chart clock while the page remains open.

### Request Inspector

- Adds a URL-addressable modal inspector with Overview, Payloads, Timeline, and Diagnostics tabs.
- Shows canonical request metadata, retained artifact state, lifecycle evidence, routing attempts/retries, stream progress, token counts, and terminal outcomes.
- Loads retained payload content only on explicit request, renders JSON inertly, and clearly distinguishes redacted, truncated, unavailable, missing, and corrupt artifacts.
- Aggregates paginated events, artifacts, and attempts with a bounded cap and an explicit incomplete-evidence state.
- Includes completed, active, failed, retry, empty, malformed, capability-unavailable, delete, and deep-link behavior.

### Maintenance and recovery

- Adds preview/confirm scoped cleanup, per-request deletion, partial-result diagnostics, safe retries, receipts, and authoritative refresh behavior.
- Adds connection-state presentation for live SSE, polling fallback, replay-gap recovery, pause/resume, and capability-aware unavailable states.
- Adds direct `/logs` and request-inspector route serving from the embedded console.

### Configuration and terminal output

- Exposes the canonical `[logging.audit]` settings in configuration schema/UI and CLI help. Because this feature is unreleased, there is no deprecated top-level `[audit]` compatibility layer.
- Adds privacy-safe terminal projections for logging events, including cross-platform Windows drive/UNC/device path redaction and case-insensitive sensitive-key handling.
- Keeps local trusted output distinct from redacted pretty/JSON projections.

### Quality coverage

- Adds unit and Playwright coverage for ledger workflows, request-inspector states, maintenance receipts, accessibility, responsive layouts, deep links, live recovery, and privacy boundaries.
[...]

PR #1228, opened 2026-08-10
Durable KV prefix cache: agent prefixes survive eviction, restart, and cold nodes

### LINKED ISSUE #1226: KV prefix retention: durable disk-backed prefix cache so agent prefixes survive eviction, restart, and cold nodes
## Goal

Keep KV prefixes reusable for **longer**, so repeated prompts skip cold prefill.

This is **not** about extending effective context length, and not OS-style demand paging. The target is retention time and hit rate.

**Primary workload:** agentic traffic — stable system prompt plus tool schemas, divergent tails, same logical conversation returning after a gap.

**Mesh premise:** requests from one user are round-robined across peers, so a given peer sees a familiar prefix again after a long gap. Retention value per hit is high, hit frequency is low. That favours a large slow tier — but routing can also remove much of the problem before any tier is built. Both are in scope.

Full plan with file/line evidence: [`docs/skippy/KV_RETENTION_PLAN.md`](docs/skippy/KV_RETENTION_PLAN.md) (branch `feat/kv-cache-disk-tier`).

## Motivating scenario: partial reuse across a long gap

This is the case the plan exists for, and it needs **two** workstreams that are often mistaken for alternatives.

A large prompt (say 8k tokens of system prompt plus tool schemas) was served a while ago. A new request arrives with the same bulk and a **new tail**.

| Sub-problem | Solved by | Not solved by |
|---|---|---|
| "new tail" — reuse the bulk, prefill only the divergent part | the candidate grid + suffix prefill (**W2b**) | a disk tier |
| "not seen for a while" — the page was LRU-evicted from RAM and is gone | the mmap tier (**W4**) | the candidate grid |

So **W2b decides whether a reusable page exists at a shareable length; W4 decides whether it survived the gap.** Either alone yields nothing for this scenario.

Why mmap suits the "massive bulk" shape specifically: restore cost is bounded by the page's bytes and the kernel only faults in pages actually touched. A 4 GB page still warm in page cache is nearly free; cold, it is one sequential read (~1.4 s at 3 GB/s) against ~8+ s of quadratic-attention prefill. **The larger the reusable bulk, the better the ratio** — the opposite of most caches.

Caveats: restore lands on a `shared_prefix_stride_tokens` floor, so up to `stride - 1` tokens get re-prefilled — irrelevant against a multi-thousand-token bulk. And a very long new tail after a restore must not exceed the runtime's suffix-prefill limits, or it falls back to full recompute and the win is lost (see `.agents/skills/kv-tool-loop-stability`).

### Skippy specifics: this applies to split serving *and* solo serving

KV retention is **per stage**, not per model. `prefix_hash_with_namespace` (`skippy-cache/src/identity.rs:50-56`) hashes `stage_id`, `stage_index`, `layer_start`, and `layer_end`, so:

- **Solo serving** — one stage covering all layers; one KV page per prefix.
- **Split serving** — each node owns a layer range and caches KV **only for its own layers**. A node holding layers 0-19 stores a page for 0-19; the node holding 20-39 stores a separate page. They are distinct `page_id`s and are not interchangeable.

Consequences specific to split topologies:

- A cold prefill on **any** stage in the chain costs the whole request. Retention has to hold across every stage for the pipeline to benefit, so per-stage hit rate matters more than aggregate hit rate. One stage missing negates upstream hits.
- Per-stage pages are **smaller** than a whole-model page — bytes scale with that stage's layer count — which improves W5's bandwidth ratio per node and makes a disk tier cheaper per node than the solo numbers suggest.
- Package-backed stages already cache only their own layer range, so W4 composes with materialized stage caches without loading a monolithic GGUF.
- Because `topology_id` and the layer range are in the hash, **re-splitting invalidates every page.** A mesh that replans topology loses its entire retention benefit. Worth measuring how often replanning happens before investing in W4/W5.
- **Gap: activation frames have no serialize path.** `ResidentActivationCache` (`skippy-cache/src/resident/activation.rs`) is resident-only — there are no `activation` references in `payload/mod.rs` or `exact_state.rs`. So the activation-frame reuse that removes work at a *stage boundary* cannot survive eviction or a restart at all, and W4 as scoped does not cover it. Whether to extend the mmap tier to activation frames is an open question; it may be the larger split-serving win, since an activation frame is far smaller than a KV page.

## Current behaviour

| Family | Payload | On eviction |
|---|---|---|
| Dense attention (llama, qwen3, deepseek2/3, glm4, gemma, minimax) | `ResidentKv` | native seq drop; **state is lost**, next request recomputes |
| Hybrid/recurrent (Qwen3Next, Falcon-H1, RWKV/Mamba) | `KvRecurrent` | `export_kv_page` + `export_recurrent_state` into an in-RAM BLAKE3 block store |

`ResidentKv` is a *performance* default, not a capability boundary. The gate in `kv_integration/config.rs` is one-directional: it stops recurrent models using KV-only reuse; nothing stops dense models using the serialized path. MoE vs dense is **not** the relevant axis (DeepSeek3 is MoE and uses `ResidentKv`) — the axis is whether attention KV alone is the full continuation state.

## Headline finding

**Cross-session prefix sharing already works, but recording is throttled so the shared prefix is almost never captured.**

- `prefix_hash_with_namespace` (`skippy-cache/src/identity.rs:47-70`) contains **zero `session_id` references**. Two unrelated sessions with the same leading tokens produce the same `page_id`.
- The namespace is `base.chat_template_id` (`kv_integration/identity.rs:22`) ← `ids.cache.namespace()` (`frontend/prefix_cache.rs:198`), which is `Some` **only** when the client sends `prompt_cache_key` (`frontend/generation/cache_hints.rs:111-115`). Ordinary requests get `None` → shared default namespace → cross-session reuse.
- The throttle: `family_policy.rs:107-109` sets `shared_prefix_record_limit: 2`, and `record_candidate_token_counts` (`skippy-cache/src/config.rs:150-184`) always keeps the full length first.

Simulating the real policy for an 8000-token request:

| | value |
|---|---|
| Lengths **probed** on lookup | 62 (8000 → 256 in 128-token steps) |
| Lengths actually **recorded** | `[8000, 7936]` |
| Is a 2048-token shared system prompt recorded? | **No** |
| Would 2048 be found if it had been recorded? | **Yes** — it is probed |

Both recorded entries sit at the *tail* of one request — the least shareable part. A second session with the same system prompt and a different tail probes 2048, finds nothing, and does a full cold prefill.

## Verified starting facts

Each confirmed against the tree at `5bf7330d`.

| Fact | Evidence |
|---|---|
| KV quantization is **already executable** | `inference/skippy/resolver/support.rs:163-167` maps `saver` -> `cache_type_k/v = "q8_0"`, `kv_offload = true`; reaches `StageConfig` via `resolver/translation.rs:83`; `skippy-protocol/src/lib.rs:280-282`; parsed at `skippy-runtime/src/config.rs:229` -> `GGML_TYPE_Q8_0`. Default policy is `balanced` (`resolver/resolution.rs:227`). |
| ...but `saver` also regresses throughput | `support.rs:225-232` halves batch/ubatch and forces `parallel=1`, `continuous_batching=false`. Set `cache_type_k/v` explicitly instead of shipping the macro. |
| `export_kv_page`/`import_kv_page` are **already on the serving path** | record at `kv_integration/exact_state.rs:125`, restore at `:65`, plumbed via `runtime_state/lane_lifecycle.rs:229,257`. |
| ...but only for `KvRecurrent`, and only whole-prefix | `exact_state.rs:118` selects on payload; `export_kv_page(session_id, 0, token_count)` hardcodes `token_start = 0`. |
| `KvPageDesc` is genuinely page-granular | `skippy-ffi/src/lib.rs:556-570` carries `token_start`, `token_count`, `k_type`, `v_type`, `k_row_bytes`, `v_row_bytes`, `payload_bytes`. The cache layer flattens this to a single page. |
| Prefix identity omits KV dtype and backend | `skippy-cache/src/identity.rs:47-70` — zero `cache_type` references. `NATIVE_KV_DTYPE` is the fixed string `"ggml-native-kv"` and does not vary with q8_0 vs f16. |
| Prefix identity omits `session_id` entirely | `identity.rs:47-70` — cross-session sharing is by design. |
| The candidate policy already does approximate sharing | `skippy-cache/src/config.rs:114-190` synthesizes a stride-aligned grid; test at `config.rs:245` shows 2214/2231-token prompts sharing a 2176 candidate. |
| `max_resident_tokens = n_ctx/2` exists to fix a real wedge | `skippy-cache/src/config.rs:55-70`. |
| `trim_session` is used in serving | `runtime_state/frame_operations.rs:418`, `frontend/linear_proposal/execution.rs:303`, `binary_messaging/control_messages.rs:212`. |
| Prefix-affinity routing already exists | `mesh-llm-host-runtime/src/network/affinity.rs` (37 KB). |
| Activation frames have no serialize path | `ResidentActivationCache` in `skippy-cache/src/resident/activation.rs` is resident-only; no `activation` references in `payload/mod.rs` or `exact_state.rs`. |
| ABI is at 0.1.35 and requires exact match | `skippy-ffi/src/lib.rs:1-3, 25-27`. |

### W0 — Identity completeness (blocker)
`identity.rs:47-70` does not hash `cache_type_k`/`cache_type_v`, backend, or GPU-layer split. `NATIVE_KV_DTYPE` is the fixed string `"ggml-native-kv"` and does not vary with q8_0 vs f16. Benign in-process; **silent numerical corruption** once state outlives a process (W4) or crosses a node (W7) — flipping `kv_cache_policy` from `quality` to `saver` makes stale q8_0 payloads collide with f16 `page_id`s. Also make desc/payload mismatch a hard error (`exact_state.rs:66-68` currently `continue`s past it). One file. **Prerequisite for W4 and W7.**

### W2 — Miss-reason instrumentation (gate)
Histogram `evicted_recently` vs `never_seen` vs `identity_mismatch`, bucketed by gap length. **If evicted-recently misses are rare, W4/W5 are worthless and this saves the entire effort.** Cheapest way to validate the round-robin premise with numbers.

### W2b — Deepen the shared-prefix record ladder (likely largest win)
Record at least one **low, stable** candidate (near `min_tokens`, or aligned to a detected system-prompt/tool-schema boundary) rather than only the two longest. Consider a non-uniform ladder — tail candidates for same-session continuation *plus* low candidates for cross-session sharing; these currently compete for 2 slots. Note `derive_max_entries_from_kv_cells` (`family_policy.rs:101,132-148`) bounds entries by `n_ctx / (2 * min_tokens)`, so a deeper ladder needs the capacity from W1. **Caveat:** a client sending `prompt_cache_key` *partitions* the namespace and thereby **disables** cross-session sharing — check no agent harness sets it by default.

### W1 — KV quantization as retention policy
Already executable: `resolver/support.rs:163-167` maps `saver` → `cache_type_k/v = "q8_0"`. But `saver` also halves batch/ubatch and forces `parallel=1`, `continuous_batching=false` (`support.rs:225-232`) — set `cache_type_k/v` explicitly instead. ~2× resident entries. Lossy: rebaseline `skippy-correctness`, check MTP acceptance rates. Orthogonal to a disk tier, and they compound.

### W6 — Prefix-affinity routing (mesh)
`network/affinity.rs` already exists (37 KB). Hashing leading-prefix identity into peer selection makes the same prefix land on the same peer, which **removes the round-robin premise motivating the disk tier**. Routing only — no wire protocol, no correctness risk.

### W3 — Page-granular export
`KvPageDesc` carries `token_start`/`token_count` (`skippy-ffi/src/lib.rs:556-570`) but `export_kv_page(session_id, 0, token_count)` (`exact_state.rs:125`) always exports from zero, so overlapping candidates re-export the same leading bytes and rely on 1 MiB dedupe to claw it back. Page-granular export *eliminates* that work. No ABI change. Makes a deeper W2b ladder cheap.

### W4 — Disk tier via whole-payload mmap
`skippy_import_state`/`import_kv_page` take contiguous `(ptr, len)`, so mmap gives zero-copy restore and the kernel page cache handles residency. **Deduped blocks on disk is the wrong design** — `CacheBytes::as_cow()` (`payload/bytes.rs:60-79`) allocates and concatenates for any `Blocks` repr, ~2 GB of pointless traffic for a 1 GB payload. Add `CacheBytesRepr::Mapped(Arc<Mmap>)` that borrows; keep `CacheBlobStore` for RAM. `ExactStateCache` needs a demote-on-evict hook. `model_fit.cache_ram_mib` is schema-reserved today — natural home for caps.

### W5 — Export-on-eviction for dense families
Today dense eviction drops state entirely (`resident/prefix.rs:230-260`). Wins iff `kv_bytes_per_token / disk_BW < prefill_time_per_token` — for a 32-layer GQA-8×128 shape at f16 (~128 KB/token) on NVMe ~3 GB/s vs prefill ~4k tok/s: ~4–6× favourable at f16, ~8–12× at q8_0. Loses on MLA/compressed KV and slow disks. Two hard constraints: eviction runs **on the decode hot path** (`binary_transport/kv_eviction.rs:104`) so export must be async; and deferred export means the seq can't drop until export completes — mishandling gives use-after-drop or a cell leak re-triggering the 502 wedge `max_resident_tokens` was added to fix. Gate on a **runtime** bandwidth check, not a hardcoded ratio.

### W7 — Peer prefix fetch (speculative, likely not worth it)
Identity isn't peer-portable as written (W0), `topology_id`/`stage_id`/layer range are in the hash so heterogeneous splits won't match, and 1 GB over LAN QUIC (~1 s) is worse than local NVMe and comparable to recomputing. Plus a new wire surface, additive gossip, and cache-poisoning trust. **Gate:** measure peer BW vs local disk BW vs prefill first; if peer BW < disk BW this is dominated by W4.

## Rejected

- **Radix/tree prefix sharing (RadixAttention style).** `PrefixCandidatePolicy` already does stride-quantized approximate sharing. A tree upgrades stride-floor-LCP to exact LCP — ≤128 tokens per hit, <6% on a 2k system prompt. Cost is a rewrite of `resident/prefix.rs` (885 lines, near the 1k rule) plus `resident_prefix.rs` plus eviction, needing one seq per *node* against a hard `seq_id < 1024` ceiling. Its real win is storage, which W3 gets cheaper. Reopen if median `LCP − stride_floor(LCP)` >5% of prefix length.
- **Partial eviction via `trim_session`.** Wrong shape — truncates a *session*, but cache entries are bare `seq_id`s in the unified pool. Would need create-from-resident-prefix + trim + drop a temp session per eviction, and it keeps the prefix while discarding the tail, which is backwards here.
- **KV defragmentation.** Zero `defrag` hits is expected; cell allocation/compaction live inside llama.cpp's unified KV cache. `NATIVE_KV_LAYER_CONTIGUOUS_LAYOUT` is a serialization layout tag, not a residency invariant. Correct mitigation already in place (`max_resident_tokens = n_ctx/2`).

**Not possible:** vLLM-style block-table indirection — contiguity is load-bearing (patch `0062-Harden-Inkling-MTP-and-KV-contiguity-state`). Defrag is not a substitute gap either: cell allocation and compaction live inside llama.cpp's unified KV cache, and `NATIVE_KV_LAYER_CONTIGUOUS_LAYOUT` is a serialization layout tag, not a residency invariant.

## Sequence

1. **W0** identity completeness — blocker, one file
2. **W2** miss-reason instrumentation — gate; also measures W2b's baseline
3. **W2b** deepen the record ladder — likely largest win, mostly policy
4. **W1** KV quantization — config, already executable; supplies W2b's capacity
5. **W6** prefix-affinity routing — may remove the need for W4/W5
6. **W3** page-granular export — makes a deeper ladder cheap
7. **W4/W5** decide from W2 + W6 data
8. **W7** only if peer BW beats local disk BW

## Cross-cutting risks

- **Identity under-specification** (W0) — blocker for W4/W7.
- **Silent-miss semantics** — must become a hard error on a disk tier.
- **Async export lifecycle** (W5) — use-after-drop or cell leak re-triggering the 502 wedge.
- **Hot-path latency** (W5) — eviction is on the decode path.
- **Lossy q8_0 KV** (W1) — rebaseline correctness; check MTP acceptance.
- **Mesh protocol** — W0–W6 are node-local. Only W7 touches the wire and must be additive per the mixed-version rules in `AGENTS.md`.
- **ABI sync** — W3/W5 need no ABI change (ABI is 0.1.35, exact match required). If page-granular export exposes a patch-queue bug, bump `SKIPPY_ABI_VERSION_PATCH` in both `skippy/common.h` and `skippy-ffi/src/lib.rs`.
- **File size** — `resident/prefix.rs` at 885 lines is near the 1k threshold; extract rather than grow.

michaelneale:
Closes #1226 (partially — W0/W2/W2b/W4 plus a dense bridge; W1/W3/W5/W6/W7 not done).

## What you get

A prefix computed once survives the process that computed it. With the opt-in
disk tier, the first request after a node restart mmaps the KV page a previous
process wrote instead of prefilling it again.

Measured on an M4 Max, release host plus Metal native runtime built from this
branch, Qwen3-8B Q4_K_M layer package, `--ctx-size 16384`, 8 GiB tier, ~8.6k-token
agent prompt. Cold runs start from an empty cache directory.

| Scenario | Time | cached_tokens |
|---|---|---|
| Cold, empty tier | 7.88s | 0 |
| New session, same system prompt, different tail | 7.30s | 768 |
| Third distinct tail | 6.48s | 2688 |
| First request after **process restart** | **0.58s** | 8576 of 8609 |

The restart case is the result to look at: **13.6x**, reproduced across two
separate verified restarts (0.58s and 0.58s). 8576 of 8609 tokens were restored
from disk.

**On the cross-session rows.** They hit, but modestly — a few hundred to a few
thousand tokens of an 8.6k prompt, worth a fraction of a second. An earlier
revision of this description claimed 12.65s saved cross-session; that number
came from a configuration I could not reproduce at the shipped defaults, and it
has been removed rather than restated. The durable, reproducible win here is the
restart case, and the honest summary is that this PR makes prefixes survive
**time on a node** — not that it makes cross-session sharing dramatically faster
within one process.

**Verifying a restart actually happened matters.** During this measurement a
`pkill` silently failed to match, the replacement process died on
`Address already in use`, and the "post-restart" request was served by the
original process — which manufactures a perfect-looking cache hit. Every restart
number above is from a kill verified by an empty `lsof -ti:9337` before the new
process started.

## The record ladder

Cross-session sharing already worked — `prefix_hash_with_namespace` contains no
`session_id`. It was *recording* that was throttled: a policy that always keeps
the full length meant an 8000-token prompt recorded only `[8000, 7936]`, both in
the request's own tail, the least shareable part of it. Lookup probed dozens of
lengths including 2048; nothing was ever stored there.

The ladder keeps the exact length and the near-tail candidate (same-session
continuation) *and* reaches down into the shared system-prompt region.

**A second defect in that ladder was found and fixed while validating this
branch, and it is worth reading before reviewing the policy.** The ladder charged
its two unconditional slots — exact and near-tail, the longest candidates by
construction — against the resident token budget before considering any shared
rung. For a 12288-token prompt those two pin ~24k tokens, more than the entire
budget on any context below roughly 48k, so the first shared rung was
unaffordable and the loop exited immediately. Composed over the shipped config
chain (`n_ctx` → `max_entries` → `record_limit` → resident budget):

```text
before:  ctx=8192 -> [12288, 12160]     ctx=16384 -> [12288, 12160]
after:   ctx=8192 -> [12288, 12160, 896] ctx=16384 -> [12288, 12160, 3328, 896]
```

Only a 131k context recorded anything shareable beforehand. The budget now
governs the optional rungs only; the mandatory slots are already committed by
the time it is consulted. `record_candidate_token_counts` has a regression test
over the composed shipped configuration, because every hand-picked-policy test
missed this.

**This fix is not load-bearing for any number in this PR, and I want that on the
record.** A live A/B against `SKIPPY_KV_CACHE_SHARED_RECORD_LIMIT=2` (the old
effective behaviour) was identical on every row within noise — cold 7.88s vs
7.89s, restart 0.58s vs 0.60s, both tiers ending with one on-disk entry. The
archive selector caps one page per request, so ladder depth never reaches disk,
and the resident-path cross-session hits occur identically at limit 2. The fix
makes recorded behaviour match documented intent; it does not produce a measured
speedup here.

## Two findings worth your attention

**1. `topology_id` made persistence impossible — and removing it removed a safety net.**

Local serving derives it as `topology-mesh-skippy-{unix_nanos}`, so it is unique per process. While it was hashed, every restart produced fresh `page_id`s and no persistent cache could ever be read back. Removing it is required for any disk tier.

But it was also, accidentally, the thing guaranteeing no stale page could ever be misread. `model_id` is a display name, not a content digest — two runs can serve genuinely different tensors under the same alias (different quant, repacked layer package, GGUF swapped under the same path). So identity now hashes `manifest_sha256`, `source_model_sha256`, `package_ref` and `load_mode` explicitly. **This is the change I'd most like a second opinion on**, because the failure mode is silent numerical corruption that persists across restarts, not a crash.

**2. Archiving the longest candidate is useless.** When I capped archiving to one page per request, I initially picked the longest — which is the request's own tail, which nothing else ever asks for. Cross-restart hits vanished until it was changed to the *lowest* (most shareable) candidate. Caught by measurement, not by tests.

## Safety

The tier's failure mode is wrong numbers, not a crash, so:

- Payload bytes are checksummed and verified before import; a mismatch quarantines the entry rather than being treated as a miss.
- The KV page descriptor is persisted with the bytes — without it, archived pages survive and are silently unusable (a real bug found by running a live model, now covered by a regression test).
- Descriptor token ranges are cross-checked against the looked-up identity before import; a failed import stops probing rather than layering a second page onto a dirtied session.
- Cache directories take an exclusive lock. A second instance on the same directory declines the tier instead of sharing it, because the index is last-writer-wins and orphan reclaim would delete the other instance's live, mapped files.
- Mapped payloads own no blob-store blocks, so promote/re-evict cannot double-release. Covered by a test.

## Default path

The disk tier is **off** unless `SKIPPY_KV_DISK_TIER_MIB` / `SKIPPY_KV_DISK_TIER` is set. With it off, the only behaviour changes are the identity hash and the ladder depth.

## Architecture

- `skippy-cache/src/disk_tier.rs` — mmap-backed store: atomic rename, checksums, LRU, directory lock, orphan reclaim.
- `CacheBytesRepr::Mapped` borrows a mapped range instead of allocating. Block-deduping on disk was rejected: `as_cow()` on a `Blocks` payload concatenates the whole thing, which for a multi-GB page is gigabytes of copying immediately before the runtime copies again.
- `skippy-server/src/kv_integration/dense_disk.rs` — dense (`ResidentKv`) families have no serialized form, so without this bridge the tier would help only hybrid/recurrent models and do nothing for llama/qwen/gemma. Archiving happens at **record** time, not eviction: eviction runs on the decode hot path, and deferring it would mean the llama.cpp sequence cannot drop until the export completes (use-after-drop, or a cell leak re-triggering the 502 wedge `max_resident_tokens` exists to prevent).
- Ladder depth is bounded by a **token** budget, not a slot count — the entry-count cap assumes ~`min_tokens`-sized entries, which the two longest slots violate badly.

## Protocol

Node-local only. No wire-format, gossip, or plugin-protocol changes. Skippy ABI unchanged (0.1.35). Disk format is versioned and self-invalidating.

`RuntimeKvPageDesc` gains serde derives — plain data mirror, layout unaffected.

## Validation

`cargo fmt --all --check` clean; clippy `-D warnings` clean on skippy-cache, skippy-server, skippy-runtime, mesh-llm-host-runtime, mesh-llm. Tests: skippy-cache 105 + 7 integration, skippy-server 442, mesh-llm-host-runtime 1932, skippy-runtime 69 — all passing. Full CI green on the rebased head.

Live verification on a real dense model as described above, including a genuine process restart against a warm cache directory.

### Two-node split serving (validated)

Measured on a 2-node loopback split, Qwen3-8B Q4_K_M layer package, ~16.9k-token agent prompt, 10 GB VRAM budget and an 8 GiB disk tier per stage:

| Scenario | Time |
|---|---|
| Cold, both stage caches empty | 31.02s |
| Cross-session, same prefix, new tail | **1.29s** / 0.91s |
| First request after restarting **both** nodes | **1.54s** |

Restart reuse across a split now works — a 20x win. An earlier 0.5B run showed no restart gain because the downstream stage archived only its 512-token floor candidate and the all-or-nothing veto cancelled stage 0's deeper archive. Both stages now archive the 16768-token shared bulk:

```text
stage 0 : [16768]                                   <- one page, the shared bulk
stage 1 : [16768, 16640, 16512, ...] (88 entries)   <- full ladder
```

Fixed by `e1c4af42` (stop `min_record_tokens` starving the archive selector on a warm restore; archive on every stage, not only stages with a downstream) and `b32dec97` (drop the resident-admission gate on stage-0 full prefill, which declined every rung of a prompt over `max_resident_tokens`).

### Two physical machines over LAN (validated)

The table above is a *loopback* split. Repeated across an M4 Pro and a Mac mini on a LAN (13-14ms direct QUIC), Qwen3-8B Q4_K_M layer package, `--max-vram 4 --ctx-size 16384` per node, 8 GiB tier per node, stage 0 layers 0-22 on one machine and stage 1 layers 22-36 on the other:

| Scenario | Time | cached_tokens |
|---|---|---|
| Cold, both caches empty | 12.45s | 0 |
| Cross-session, same prefix, new tail | **1.49s** / 1.53s | 4096 |
| First request after restarting **both** machines' nodes | **1.72s** | 4096 |

Both nodes independently persisted `format_version: 1` indexes and reloaded them into a freshly negotiated topology, so restored page ids agreed across hosts. Stage 1 also exercised budget eviction (`disk_evictions` 0 → 3 → 5 against its 2 GiB share) while still serving hits — a path the loopback run never reached.

Re-run under a **pinned topology** (`--split --split-topology-lock`, stage 0 layers 0-32 on the M4 Pro, stage 1 layers 32-36 on the mini, `--ctx-size 8192`), verified via `GET /api/runtime/stages` rather than `/v1/models`:

| Scenario | Time | cached_tokens |
|---|---|---|
| Cold, both caches empty | 5.48s | 0 |
| Cross-session, same prefix, new tail | **1.39s** / 1.29s | 4096 |
| First request after restarting **both** nodes | **1.48s** | 4096 |

This replaces an earlier run that reached a split by tuning `--max-vram`, which is unreliable — at `--max-vram 5` one node silently served all 36 layers solo while still looking like a healthy two-node mesh. `docs/skippy/KV_RETENTION_PLAN.md` now records the lock recipe and the traps.

### Agent harness (validated)

`AGENTS.md` requires a harness run for changes on this surface. Goose was run
against the local proxy with tool calling, on the same node and tier as above.

| Check | Result |
|---|---|
| Goose, `model=auto`, file-creation task with tool calls | completed, `hello.txt` written |
| Goose, `model=<explicit model id>`, shell tool call | completed, 39.9s |
| Goose, `model=auto`, after a verified node restart | completed, 35.2s |
| Multi-turn tool-call loop, assistant `tool_calls` + `role=tool` reply | completed, no reducer or schema errors |

The disk tier stayed healthy across the harness: four archived entries
(4992/5120/5632/4992 tokens), 1.5 GB, no corruption or quarantine events. A
5708-token harness prompt returned `cached_tokens=5632` at 1.80s after a restart,
against 5.78s cold — the restart win holds under a real agent workload, not just
scripted requests.

`model=mesh` was **not** exercised: on a single-node mesh the proxy returns
`model 'mesh' not found (no local or remote host serving this model)`, which is
correct behaviour for that topology rather than a defect. It is covered by the
two-node runs above.

**Not validated:** WAN split, very large MoE models, OpenCode and Pi harnesses.

### Observability, tests, and format spec (review round 2)

Follow-up to @i386's review, in `f320fa27`:

- **Hit attribution.** Exact-state hits carry `skippy.exact_cache.hit_source` (`ram` | `disk`).
- **Archive outcomes.** `archive_dense_prefix` previously returned `Result<bool>`, so "policy declined" and "the archive failed" collapsed into the same silent `false` — and two of the four call sites discarded it with `let _ =`. It now returns an explicit outcome and every site reports `skippy.kv.archive_status` (`archived` / `skipped_too_short` / `skipped_already_archived` / `skipped_tier_disabled` / `failed_export` / `failed_write` / `failed_error`), plus `archive_bytes`, `archive_export_ms`, `archive_write_ms`.
- **Disk-tier counters exported.** `disk_demotions`, `disk_promotions`, `disk_evictions`, `disk_corrupt_entries`, `disk_verifications`, `disk_verifications_skipped`, `disk_entries`, `disk_bytes`, `disk_max_bytes`, `disk_tier_enabled`. Without these, a tier that has silently stopped storing anything is indistinguishable from one that is never probed.
- **Payload-kind coverage.** `RecurrentOnly` had no disk round-trip at all; it now has demote → restart → restore. Cross-kind rejection went from 1 of 6 pairs to all 6, plus a test that a kind mismatch keeps the entry quarantined against a later *correct* lookup.
- **Format spec.** `docs/skippy/KV_DISK_TIER_FORMAT.md` — directory layout, index schema, lifecycle, atomic-write and crash-recovery contract, both checksummed regions, full corruption-response table, identity contract, versioning rules. The JSON index is retained deliberately: metadata only, read once at startup, entry counts in the hundreds. A binary index should follow a profile, not precede one.
- **User docs.** `website/src/docs/pages/kv-disk-cache.md` plus a nav entry: how to enable, when it declines to enable and why, safety model, how to clear it.
- **`DISK_TIER_FORMAT_VERSION` reset 3 → 1** before landing.

## Open questions for review

1. Is the weight-identity set (`manifest_sha256` + `source_model_sha256` + `package_ref` + `load_mode`) sufficient? Should the tier refuse to open when both digests are absent?
2. `n_gpu_layers` and backend device are hashed conservatively, but auto-offload can shift between runs on the same host — that will silently self-invalidate the tier. Right tradeoff?
3. Archive-at-record costs one KV export per request even when nothing later reads it. Worth gating behind observed hit rate (i.e. land W2 export first, then decide)?

## Summary by CodeRabbit

- **New Features**
  - Added optional persistent KV-prefix caching with disk-backed storage and memory-mapped restores.
  - Dense KV prefixes can be archived and restored after memory eviction or process restarts.
  - Added cache statistics and miss classifications.

- **Improvements**
  - Shared-prefix selection adapts to cache capacity and token budgets.
  - Cache identities distinguish incompatible runtime configurations while supporting restart reuse.
  - Restoration falls back across exact, resident, and archived prefixes.
  - Added CLI and documentation for disk-cache configuration.

- **Bug Fixes**
  - Improved handling of corrupted or incompatible cached data.
  - Added validation for short activation frames during forwarding.

---

## Review round 3

**Architecture and byte order are now bound into page identity.** Reviewer asked for explicit byte-order and architecture identifiers. `update_layout_identity` hashed cache dtypes, flash-attn, GPU layer split, and `backend_device` — but not CPU arch, endianness, or pointer width. The identity hash deliberately encodes its own integers little-endian (correct for a portable id), which meant two hosts of *different* native endianness computed the **same** page id for the same tokens.

Harmless in the default machine-local dir; not harmless once a dir is shared or copied, because `SKIPPY_KV_DISK_TIER_DIR` takes any path, the stage directory key holds only model/stage shape, and `backend_device` does not separate x86_64 CUDA from aarch64 CUDA, nor two CPU-only hosts both recording `<no-selected-device>`. Checksums would confirm the bytes arrived intact and the runtime would import them as native — a silent misread. `update_platform_identity` now hashes `ARCH` + endianness tag + pointer width, so a foreign page is a miss, never a wrong hit.

Chosen over a `.kvp` magic header: no change to component offsets or file sizing, and mismatches become *different ids* rather than mutual quarantine between platforms sharing a directory. No format bump — v1 has not shipped, so no on-disk directory claiming it exists in any build a user could have run; the spec now says the bump rule applies from the first released version onward.

**Retention policy documented, no TTL added.** Entries are content-bound: age alone does not make one wrong, and they become *unreachable* rather than incorrect when model or config changes. Idle entries consume no more than their existing allowance, so expiry only forfeits future hits. A TTL would be a privacy/hygiene control, not correctness. The real gap is different and now recorded: stage directories abandoned on a `model_id` or stage-shape change fall outside any tier subsequently opened, which wants a base-directory quota rather than expiry inside active caches.

**Rebased onto main** (13 commits, including the regenerated llama.cpp patch queue). That bumped the Skippy ABI to 0.1.38, so all earlier live measurements were re-run against a freshly built native runtime at the merged head `85f0060d`: 7.57s cold → 0.62s after a genuine process restart, `cached_tokens=8320`, `hit_source":"disk"`. Full CI is green on the merged head.


michaelneale:
@i386 review please when you get a chance.

Two things I'd most like your eyes on:

1. **Weight identity in `skippy-cache/src/identity.rs`.** I removed `topology_id` from the prefix hash — it's derived from `unix_nanos` per process, so while it was hashed no persistent cache could ever be read back after a restart. But it was also, accidentally, the thing making collisions impossible. `model_id` is a display name, not a content digest, so I added `manifest_sha256` / `source_model_sha256` / `package_ref` / `load_mode` to the hash. If that set is incomplete, the failure mode is silent numerical corruption that survives restarts rather than a crash — so it's worth a sceptical read.

2. **Whether archive-at-record earns its keep.** Every request with a recorded prefix pays one KV export even if nothing ever reads it back. It might be better to land the miss-reason metrics export first and let the hit-rate data decide.

Draft because the multi-node / split-topology / agent-harness validation from AGENTS.md isn't done, and the only live numbers are from a 0.5B dense model on one box. The disk tier is off by default (`SKIPPY_KV_DISK_TIER_MIB`), so the default serving path only sees the identity change and the deeper record ladder.

michaelneale:
Followed up on the "surely this saves a ton on a big model" question — it does, and my first numbers undersold it badly.

Re-measured on **Qwen3-8B Q4_K_M (layer package / staged)** with a ~12.4k-token agent prompt, cold control = same binary against an empty cache dir:

| Scenario | Cold | Warm | Saved |
|---|---|---|---|
| Cross-session, same prefix, new tail | 21.50s | **8.85s** | 12.65s |
| First request after **full process restart** | 21.60s | **8.85s** | **12.75s (2.44×)** |

The archived page is **0.96 GB** for 12288 tokens.

The detail I think is the actual result: the post-restart number (8.85s) is *identical* to the cross-session number (8.85s). Restoring a ~1 GB page via mmap costs the same whether this process computed it or a previous one did. That is the plan's ratio argument holding — restore is bounded by bytes, prefill is superlinear in tokens, so the bigger the reusable bulk the better this looks.

The 0.5B numbers I led with earlier (0.31s → 0.25s) are the worst case, not the representative one. I've kept them in the PR for honesty but they were misleading as the headline.

Caveat unchanged: this is single-node. Split topologies are the interesting open question, since per-stage pages are smaller but *every* stage has to hit for the pipeline to benefit — one cold stage negates the rest.

michaelneale:
## Splits now covered — with one honest gap

Agreed splits are where the payoff is, so I dug in. Findings:

**The hard problem I expected wasn't there.** I assumed per-stage restore would need a new negotiation protocol (stage 0 restores N tokens, stage 1 must agree or attention goes wrong). It already exists: `try_restore_embedded_split_prefill` has stage 0 state its restore length on the wire, and any stage's miss vetoes the whole attempt, with misses propagating back through middle stages. So this is a third tier under an existing gate, not a redesign.

**What I added:** disk restore in `restore_binary_prefix` (one site covering both binary lookup branches) and in split stage 0's frontend path, plus archive-on-record in all four recorders. A disk hit is deliberately indistinguishable from a resident hit so the veto keeps working unchanged.

**A real bug the measurement caught.** My first archive heuristic stored the *lowest* ladder candidate. For a 2129-token prompt that's a 256-token page — 12% of the prefill, pure noise. `ArchiveCandidate` now picks the longest prefix that still excludes the request's tail (2048 here, 96%), which is the actual shared system-prompt bulk. Unit-tested.

**Safety.** Promoted an undocumented invariant to an explicit assertion: a non-first stage must execute its full token range. Suffix-only execution is legal only on the stage owning layer 0 — anywhere else the next stage attends over a prefix it never received and emits *plausible but wrong* tokens. Previously this was caught only as a side effect of a payload-size check in the encoder. Now named, asserted, and tested both ways.

### Measured, 2-node loopback split (0.5B, ~2.1k tokens)

| Scenario | Cold | Warm |
|---|---|---|
| Cross-session, new tail | 1.34s | **0.57s** |
| First request after restarting both nodes | 1.33s | 1.32s |

**Cross-session split reuse works. Restart reuse does not pay off yet.** The archive index shows why: stage 0 archives its full ladder including the 2048 bulk, the downstream stage archives only 512. Because of the veto, one stage's shallow archive negates the other's good one. I'd rather flag that than quote the flattering cross-session number alone.

`ci-two-node-split-smoke.sh` passes; the new invariant assertion never fires.

### MoE
Not a special case. MoE families (`qwen3moe`, `glm4_moe`, `deepseek3`, `openai_moe`, `llama4`, `hunyuan_moe`...) map to the same dense `resident_kv` policy — MoE changes FFN weights, not KV shape. Only hybrid/recurrent families take the other path. So the large MoE models that motivate splits are covered by the dense path already.

Still unmeasured: large MoE on a real multi-machine split, which is where the solo 8B result (2.44×) suggests the payoff should be biggest.

michaelneale:
## Status after two independent expert reviews

Ran two reviews in parallel: one on split/staged correctness, one on the disk tier and cache identity.

### Split correctness — clean
All five questions came back safe:
- The new non-first-stage assertion cannot false-positive on MTP/speculative decode, chunked prefill, or batched lanes.
- A disk hit keeps the cross-stage veto sound (agreement depends on hit/miss, not provenance).
- `resident_seq_id: None` on the disk path is benign — telemetry-only, and the pre-existing `exact` path already does it.
- A partial disk restore cannot propagate downstream: the clamp gives `<= len`, the hit gate requires `>= len`, so only a full restore reports a hit.
- Token-range mismatch is rejected before import, not silently truncated.

### Disk tier — five real issues, all fixed

| Finding | Fix |
|---|---|
| **Metadata not checksummed.** Only the payload was. `extra` carries the KV descriptor (layer range, ggml types, row strides), so a corrupted-but-valid-JSON index could apply correct bytes under a wrong layout — silent corruption on a path reporting success. | `extra_checksum` per entry, verified before the payload. Format v3. Test included. |
| **Unanchored weights could alias.** Two different GGUFs under one `model_id` collide on disk across a restart. | Tier refuses to open without a content digest (`manifest_sha256`/`source_model_sha256`/`package_ref`), and logs why. |
| **No lock on non-Unix**, running silently unprotected. | Fails closed with an explicit error. |
| **Crash-left `.tmp` files never reclaimed** — a page's bytes leaked per crash, unbounded. | Reclaimed at open, safe under the exclusive lock. Test included. |
| Debug `eprintln!` left in committed code. | Removed — my mistake, caught on re-read. |

Two review claims I checked and **rejected**: `page_id` is a full BLAKE3 digest (not truncated to 64 bits), and the index already commits atomically via write-temp-then-rename. Both were artifacts of the reviewer seeing an excerpt rather than the file.

### Re-measured after hardening (solo 0.5B, ~2.1k tokens)
| | Cold | Warm |
|---|---|---|
| Cross-session, new tail | 0.42s | **0.13s** |
| First request after restart | 0.42s | **0.21s** |

Two-node split smoke passes. Full suite green: skippy-cache 93, skippy-server 395, skippy-runtime 69, host-runtime 1932.

### Confidence
- **High** on default-path safety — tier is opt-in, off by default; the only always-on change is the identity hash.
- **High** on solo retention — measured repeatedly, including 2.44× on an 8B package.
- **Medium** on split retention — cross-session works and is measured; **restart still shows no win** because the downstream stage archives a shallower ladder than stage 0, and the all-or-nothing veto means one shallow archive negates the other. Known, documented, unfixed.
- **Untested**: large MoE across real multi-machine splits; Windows (now explicitly refuses rather than running unlocked).

@i386 — the split gap above is the main thing I'd value a second opinion on.

i386:
Review findings:

1. **P1 — Stage-0 full-prefill archives the smallest candidate**  
   `crates/skippy-server/src/frontend/prefix_cache.rs:540-546` uses `identities.last()`. Identities are longest-first, so this archives the least useful page—or nothing when it falls below the 512-token floor. Use `ArchiveCandidate` here too.

2. **P1 — Recurrent KV demotions lose their descriptor**  
   `crates/skippy-cache/src/exact_state.rs:396-404` stores demoted entries with `extra=None`. `KvRecurrent` pages need `kv_desc`; after eviction/restart they cannot restore the KV portion and are silently treated as misses.

3. **P1 — Disk tier can enable with an unsafe identity**  
   `crates/skippy-server/src/kv_integration/config.rs:97-100` accepts any `package_ref`, including mutable paths, or invalid/empty digest strings. Require a validated 64-hex `manifest_sha256` or `source_model_sha256`; `package_ref` alone is not immutable content identity.

4. **P1 — Every disk hit hashes the entire mapped payload**  
   `crates/skippy-cache/src/disk_tier.rs:493-499` re-runs BLAKE3 for every component on every promotion. For the documented ~0.96 GB page, this happens on every hit while cache/runtime locks are held. Cache verified mapping IDs and invalidate them when mappings are removed.

5. **P2 — Disk hits are counted as both misses and hits**  
   `crates/skippy-cache/src/exact_state.rs:150-162` calls `lookup()`, recording a miss before consulting disk, then records a hit. Retention telemetry and recoverable-miss ratios are inflated.

The SHA-256 concern is not new to this PR: source-model SHA-256 computation and its sidecar cache already exist on the base. The new hot-path concern is the repeated full-page BLAKE3 verification.

Local validation passed: skippy-cache 93 unit + 7 integration tests, and 19 targeted skippy-server tests.

i386:
@michaelneale thanks for the follow-up. My read of the current state:

- **Weight identity:** the identity fields were added, but the disk-tier gate is not fully fail-closed yet. A `package_ref` alone, or an invalid/empty digest, can still enable retention. This remains a P1.
- **Archive-at-record cost:** the new measurements justify keeping it, but it is still unconditional rather than gated by observed hit-rate telemetry. That is an explicit trade-off, not fully addressed in code.
- **Split/restart behavior:** cross-session split reuse works, but restart reuse still has the documented gap where a shallow downstream archive causes the all-or-nothing restore veto.
- **Disk hardening:** metadata checksums, non-Unix locking, temporary-file reclaim, and the debug output appear addressed. The immutable content-digest gate remains incomplete.

I also found three additional issues in the current head: stage-0 full-prefill can still select the smallest candidate, demoted recurrent KV entries lose their descriptor, and every disk hit re-hashes the full mapped payload.

Could you confirm the intended disposition for these remaining items—especially the digest gate, stage-0 candidate selection, and split restart gap—before merge?

micspiral:
## Your bandwidth point is the biggest thing in this issue — and I can now show why

I chased all three. The second one changes how I'd frame the whole feature.

### 1. Prefix commonality — agreed, and it's a retention-time argument
A fleet on one agent sends near-identical system prompt + tool schemas forever. High hit frequency, full-cold-prefill value per hit. The goal you describe — *never warm up again* — is exactly the disk tier's job: survive eviction, restart, and redeploy. Disk "can't hurt" is right, and it's cheap per node on splits because per-stage pages are smaller.

### 2. Prefill bandwidth — this is the finding

You're right that it's ~3 GB between stages, and it turns out **a chain restore removes essentially all of it**. `embedded_prefix_cache_message` (`frontend/wire_messages.rs:264-288`) builds the restore message with **`activation: Vec::new()`** — it carries only token IDs. So the restored span costs **4 bytes/token** instead of `width x dtype`:

| width | dtype | tokens | cold | warm | ratio |
|---|---|---|---|---|---|
| 4096 | f32 | 128k | 2.10 GB | 0.5 MB | **4096x** |
| 4096 | f16 | 128k | 1.05 GB | 0.5 MB | **2048x** |
| 8192 | f16 | 128k | 2.10 GB | 0.5 MB | **4096x** |

Per boundary, per request. That reframes things: **prefill compute parallelizes across stages, boundary transfer does not.** On a thin or shared link the bandwidth saving may dominate the compute saving — and it is completely invisible in the single-node wall-clock numbers I've been quoting. My loopback benchmarks are the worst possible place to observe your most valuable effect.

Corollary: the all-or-nothing veto is more expensive than I credited. One stage missing doesn't just lose that stage's compute — it forces full activation traffic across *every* boundary for the whole prompt. That raises the priority of the stage-1 gap below.

### 3. Pre-seeding — a consequence of work already done, not new machinery

The tier is already a content-addressed, checksummed, identity-anchored page store, so seeding is a *write* into it. What makes it plausible now is that identity no longer contains `topology_id` or any per-process value, so a page is valid for any process with matching weights/layer range/KV dtype/backend. What still blocks it: layer range and stage id **are** hashed, so a seed is only valid for an identical split — seeding has to happen after topology is chosen, or be keyed by topology and matched at plan time. Worth deciding deliberately, since re-splitting invalidates everything.

### Root cause of the stage-1 gap

Found it while looking at this. `maybe_record_binary_prefill` is called with `min_record_tokens = restored_tokens` (`prefill_recording.rs:67-81`), and the recorder skips every candidate `<= min_record_tokens` (`binary_kv.rs:632`). On a warm chain restore the downstream stage therefore declines to record exactly the shared-bulk candidates it just served — its ladder can only ever get shallower. That's why stage 0 archives 2048 and stage 1 archives 512.

Not fixed in this PR — it's a behavioural change to the record path and I'd rather land it separately with its own evidence.

Documented in `docs/skippy/KV_RETENTION_PLAN.md`. @i386 the bandwidth table is the part I'd most like a sanity check on.

micspiral:
## Yes — and it already works, no new machinery needed. Measured.

Your instinct was right. Sending the canonical prefix through the node once as an **ordinary request** populates the disk tier, because recording happens on the normal serving path.

### Solo: measured

Seed run = the system prompt with *no user turn*. Then restart the process (cold RAM, warm disk) and send a request with a tail **the seed never saw**:

| First real request on a fresh process | |
|---|---|
| Unseeded (empty disk) | 0.41s |
| Seeded (one prior prefix-only run) | **0.21s** |

The seed run archived a single **2048-token page** — the shared bulk — and an unrelated later session hit it. That's the "never warm up again" property you described: the node is useful on its *first ever* real request.

Why no new code was needed: identity contains no `session_id` and no per-process value, so a page recorded by a warmup is valid for any later session or process with matching weights, layer range, KV dtype, backend. And `ArchiveCandidate` prefers the longest *partial* candidate, which happens to be exactly right for seeding — a seed prompt's full length **is** the shared prefix.

Durability is the part that makes it worth doing: the seed survives eviction and restart, so you pay the warmup once per node, not once per process.

### Splits: no gain yet, and I found the exact cause

Same experiment on 2 nodes: **1.27s seeded vs ~1.30s cold.** Nothing. Per-stage archives show why:

```
stage 0 : [2048, 1920, 1792, ... 512]   <- full ladder
stage 1 : [512]                          <- floor only
```

512 is exactly `MIN_ARCHIVE_TOKENS` — stage 1 is archiving the smallest page it's permitted to and nothing more. With the all-or-nothing veto, that cancels stage 0's good page.

**Root cause:** `record_completed_prefill` passes `min_record_tokens = restored_tokens` (`prefill_recording.rs:67-81`) and the recorder skips every candidate `<= min_record_tokens` (`binary_kv.rs:632`). So on a warm chain restore the downstream stage declines to record precisely the shared-bulk candidates it just served. **The ladder ratchets shallower with every warm request.**

That gate is correct for its original job — don't re-record what you just restored into the *resident* cache — and wrong for the archive, which wants the shared bulk specifically.

I've deliberately not fixed it here. It's a behaviour change on the record path, it needs per-stage archive depth as its evidence, and this PR is already large. It's the clear next piece of work, and it unblocks both split retention **and** split seeding — which, given the 2048–4096× bandwidth effect, is where the real payoff is.

So: **seeding is real and works today, solo only.**

i386 (commented):
Re-review of `42a9eb4001ec27095f5faac1753ef21de074c189`:

- All 11 CodeRabbit findings from the prior review are addressed and their threads are resolved, including the outdated threads.
- Disk-hit accounting now records one outcome per lookup, so recoverable-miss metrics no longer count a disk hit as a RAM miss.
- Resident dense-KV archives now retain their own payload kind and cannot be misread as empty recurrent state.
- Stage-0 archiving selects the longest successfully resident shareable candidate, and content-addressed disk retention now requires a valid 64-hex SHA-256 digest.
- Documentation and the first-stage forwarding regression test are updated.

Validation on the pushed head: `cargo fmt --all --check`, `cargo check -p skippy-cache`, `cargo check -p skippy-server`, Clippy for both touched crates with warnings denied, 98 cache unit tests, 7 cache retention integration tests, 412 server unit tests, and `just build` all pass. No new actionable findings in this re-review. GitHub’s platform/runtime and Clippy jobs are still pending because this PR remains draft.

i386:
@micspiral could you think more carefully about the disk-cache lifecycle/design here?

I traced the current behavior:

- RAM pressure demotes exact-state entries to disk; dense resident-KV entries are proactively archived at record time.
- Disk eviction is only triggered after a write, or during tier startup, and removes disk LRU entries until the byte budget is satisfied.
- There is no TTL, periodic maintenance, or age-based cleanup. A disk hit refreshes LRU metadata but does not promote the entry back into the RAM cache.
- On restart, the index is reloaded, orphan/temp files are removed, the budget is reapplied, and payload checksums are verified lazily on first use.

This is bounded and internally consistent, but it raises a design question: should retention have an explicit age/TTL or maintenance policy, and should archive admission account more directly for write cost, restore cost, and model churn? In particular, if the workload stops writing, old entries remain indefinitely within the byte budget. Would appreciate your view on whether this lifecycle is the right policy before we consider it settled.

i386:
@micspiral on the binary packing: the current layout is efficient for a same-platform local cache. Payload components are concatenated into one contiguous `.kvp` file with no per-component padding, and restores use `mmap` without copying. The JSON index carries only metadata, offsets, lengths, and BLAKE3 checksums; page checksums use BLAKE3 rather than SHA-256.

The index integers are endian-neutral because they are JSON text. The `.kvp` payload is opaque runtime KV bytes and does not currently contain a packed integer header, so there is no immediate little-/big-endian header mismatch. However, the payload format implicitly assumes a compatible runtime/platform; endianness is not explicitly recorded or validated, so big-endian portability is not guaranteed.

As part of a versioned page schema, I suggest an explicit magic/header with schema version and byte order—probably enforce little-endian and reject unsupported byte order—plus runtime/architecture payload identifiers.

i386:
The page payload is already binary; only the metadata index is JSON.

A binary index could reduce startup parsing and repeated full-index serialization, but it would not speed normal cache hits: the index is parsed only when the tier opens, while lookups use in-memory metadata and `mmap` the payload directly.

The likely improvement is a versioned `index.bin` with:

- magic and schema version
- explicit little-endian encoding
- entry count and length-prefixed fields
- payload kind, offsets, lengths, checksums, and metadata
- CRC or equivalent index integrity check

I’d keep a JSON diagnostic/export tool, but switch the runtime index to binary if profiling shows startup or index rewrites matter. The page `.kvp` files do not need another serialization layer.

i386:
@micspiral I agree with the direction here.

For the current design, the `.kvp` page data is already raw binary and mmap-friendly; the JSON index is metadata (entries, offsets, lengths, checksums, and eviction timestamps). JSON parsing is primarily a startup/index-commit cost, so I’d keep it for now and only move to a binary index if profiling shows that cost matters. Any binary-index follow-up should define a magic header, schema version, explicit byte order, and compatibility behavior.

We should add a documented file-format spec in Markdown and on the website covering the directory layout, index/page schemas, lifecycle and eviction, atomic writes, checksums, versioning, endian assumptions, and corruption/compatibility behavior. Agent guidance should require the implementation, tests, canonical Markdown spec, and website copy to stay synchronized whenever the format changes.

One important release detail: please reset the on-disk `DISK_TIER_FORMAT_VERSION` to **v1** when this PR lands.

i386:
Test coverage by payload kind is good for the storage primitive, but it is not yet a complete compatibility matrix.

Current coverage:
- `FullState`: generic disk round-trip, restart, mmap, eviction, and corruption coverage.
- `KvRecurrent`: two-component round-trip and ordering coverage, plus descriptor persistence across restart.
- `ResidentKvArchive`: direct one-component disk round-trip and archive-candidate selection coverage.
- `RecurrentOnly`: no explicit disk round-trip coverage; it is currently mapped to disabled in the server integration.

Remaining gaps:
- Add a direct matrix test for every `ExactStatePayload` variant covering construction, component count, `kind()`, byte length, and reconstruction.
- Add a server-level `KvRecurrent` record → evict → disk restore test.
- Add a dense archive restart test covering descriptor validation and runtime import.
- Expand cross-kind rejection coverage beyond the current `KvRecurrent` versus `FullState` case.

So the cache/storage behavior is well covered, but the different payload kinds still need explicit end-to-end coverage before we can call the format compatibility surface complete.

i386:
The archive/restore path has useful request-level telemetry, but there are still operational observability gaps worth addressing:

- Exact-state hits do not distinguish RAM hits from disk hits, even though the cache internally knows whether the lookup came from `from_disk`.
- Archive failures are mostly silent: export errors, duplicate entries, size rejection, and disk write failures all collapse into “not archived,” with no outcome or reason.
- Disk-tier counters already exist internally for demotions, promotions, budget evictions, corrupt entries, and checksum verification, but they are not exposed in the runtime telemetry snapshot.
- Archive write duration and bytes written are not reported.
- Restore telemetry should consistently include the payload kind and source tier for both exact-state and dense archive restores.

I’d suggest adding explicit archive/restore outcomes and reasons, `source=ram|disk`, payload kind, bytes, split timings, and exported disk-tier occupancy/counters. That would make it possible to tell whether the archive is actually protecting against resident eviction and whether restore cost is dominated by checksum verification, disk access, or native import.

i386 (dismissed):
This is very good. I tried to think about observability, file format, data lifecycle, versioning, etc. that would be good to get in here before merging.

michaelneale:
@i386 Re-review requested — two changes since your last pass, both found by running a real MoE model end to end rather than by reading.

**1. `--kv-cache-disk` is now a first-class flag.** The tier could only be reached by exporting `SKIPPY_KV_DISK_TIER_MIB`, which is not a surface a user can be expected to find. `--kv-cache-disk <GB|auto>` and `--kv-cache-disk-dir <PATH>` configure it directly. The env vars remain the underlying mechanism and take precedence, so containers and systemd units are unaffected. A budget rounding to zero MiB is rejected loudly rather than silently disabling the tier while looking enabled.

**2. Sliding-window models are declined cleanly instead of failing on every prefill.** Serving `gemma-4-26B-A4B`, every archive returned `failed_export` / "runtime memory type is not supported for native KV pages", and `cached_tokens` stayed 0.

The cause is **not MoE** — experts are feed-forward and carry no state between tokens. It is that llama.cpp backs Gemma with `llama_memory_hybrid_iswa`, which holds two caches: a full-context base for the non-SWA layers and a window-bounded one for the SWA layers. For an N-token prefix the correct state is `0..N` on base but only the visible suffix on SWA, and one page with one token range cannot express that. `skippy_get_kv_cache` is right to refuse — exporting the base cache alone would yield a page that silently omits every SWA layer, and importing it would advance `n_past` over state that was never restored.

So I did **not** add the ISWA types to the resolver. That would turn a clean refusal into silent numerical corruption. Instead the unsupported layout is latched per stage: one `skipped_unsupported_memory`, then `skipped_tier_disabled`, costing nothing thereafter. Gemma keeps `ResidentKv` since a sequence copy duplicates both caches, so in-process reuse is untouched — only the disk tier is out of reach. A composite base-plus-suffix page would be the real fix and is separate work.

Also: export failures now carry the native reason. `Err(_)` discarded it, so "the export failed" was barely better than the silent bool it replaced — diagnosing this needed a rebuild just to see the message.

**Worth flagging:** Gemma shows no prefix reuse on `origin/main` either — I built `ff5e79f8` and ran the same probe, `cached_tokens=0`. Pre-existing, not a regression from this branch, and I have left it out of scope.

Also in this round: `ARCH` + endianness + pointer width are now bound into page identity (your byte-order ask), and the retention/TTL question is answered in the spec.

Staying in draft until you have looked, since the model-coverage boundary is a judgement call I would rather not make alone.

michaelneale:
## Follow-up work split out of this PR

Model-coverage boundaries found while validating this branch. All three are deliberately **not** in scope here — noting them so the boundary is explicit rather than implied.

- **#1264 — composite base + SWA page.** Sliding-window models (gemma2/3/4, gpt-oss, cohere2, llama4, exaone4, olmo2, phi3-family and others) cannot have their prefixes persisted, because `llama_memory_hybrid_iswa` holds two caches with different valid token ranges. This PR latches that off cleanly (`skipped_unsupported_memory`) rather than exporting a partial page. Issue carries the full design, the vLLM/SGLang prior art, the `[R-(W-1), R)` window boundary, and the `swa_full` dead end.

- **#1265 — Inkling gets recurrent-only caching.** Inkling is the only architecture that is both hybrid-recurrent *and* SWA. The recurrent resolver handles `hybrid_iswa`; the KV resolver does not. So the SSM half round-trips and the attention half is silently declined into an empty component. Safe, but doing less than the `KvRecurrent` label implies, and invisible in telemetry. Needs #1264 plus the recurrent component.

- **#1266 — fallback classifier misclassifies Qwen3.5.** `infer_cache_payload()` has no Qwen3.5 entry, so it matches the generic `qwen3` rule and returns `ResidentKv` for a hybrid model — which would drop recurrent state. Not live, because `effective_cache_config` prefers the architecture-derived policy from `family_policy.rs`, and that is correct. But two classifiers that can disagree, where the safe one wins only by ordering, is worth removing.

Confirmed unaffected and working on disk: Qwen3 / Qwen3-MoE / Qwen3-VL, Qwen3.5 / 3.5-MoE, Qwen3-Next, Qwen3.6, Kimi-Linear, Llama (pre-4), GLM-4, DeepSeek-2/3, MiniMax.

michaelneale:
Ready for re-review at `9928bcd8`. Undrafted. Summary of what changed since the last review round.

**Rebased onto `main`** (10 commits). Full CI green on the head; the earlier red `kotlin_sdk_smoke` was an HTTP 403 from Maven Central resolving the Kotlin Gradle BOM and did not recur.

**Found and fixed a second defect in the record ladder.** It charged its two unconditional slots — exact and near-tail, the longest candidates by construction — against the resident token budget before considering any shared rung. Those two pin more than the whole budget on any context below roughly 48k, so the first shared rung was unaffordable and the loop exited immediately. Only a 131k context recorded anything shareable. Composed over the shipped config chain, before and after:

```text
before:  ctx=8192 -> [12288, 12160]      ctx=16384 -> [12288, 12160]
after:   ctx=8192 -> [12288, 12160, 896] ctx=16384 -> [12288, 12160, 3328, 896]
```

Regression test added over the composed configuration (`n_ctx` -> `max_entries` -> `record_limit` -> resident budget), since every hand-picked-policy test missed it. One existing test asserted the old behaviour and was rewritten rather than deleted.

**That fix produces no measured speedup, and the PR description now says so.** A live A/B against `SKIPPY_KV_CACHE_SHARED_RECORD_LIMIT=2` was identical on every row within noise. The archive selector caps one page per request so ladder depth never reaches disk, and the resident cross-session hits occur identically at limit 2. If you would rather this behaviour change did not ride along in a PR that cannot demonstrate it, say so and I will split it out — it is one commit.

**Corrected the headline numbers.** The previous description claimed 12.65s saved cross-session. I could not reproduce that at the shipped defaults and have removed rather than restated it. Re-measured on a release host plus Metal runtime built from this branch, Qwen3-8B Q4_K_M, `--ctx-size 16384`, 8 GiB tier, ~8.6k-token prompt:

| Scenario | Time | cached_tokens |
|---|---|---|
| Cold | 7.88s | 0 |
| New session, different tail | 7.30s | 768 |
| First request after restart | **0.58s** | 8576 of 8609 |

Cross-session hits are real but modest. The restart case is the durable result: 13.6x, reproduced across two verified restarts.

Worth flagging the method error, because it nearly produced a false positive: my first restart attempt did not kill the process, the replacement failed to bind, and the "post-restart" request was served by the original. Every restart figure above is from a kill verified by an empty `lsof -ti:9337` first.

**Agent harness validated**, closing the `AGENTS.md` gap. Goose against the local proxy with tool calling, `model=auto` and an explicit model id, including a run after a verified restart, plus a multi-turn `tool_calls`/`role=tool` loop. No reducer or schema errors. Tier stayed healthy: four archived entries, 1.5 GB, no corruption or quarantine. A 5708-token harness prompt returned `cached_tokens=5632` at 1.80s post-restart against 5.78s cold, so the restart win holds under a real agent workload.

`model=mesh` was not exercised: on a single-node mesh the proxy correctly returns `model 'mesh' not found`. Covered by the two-node runs already in the description.

**Still not validated:** WAN split, very large MoE, OpenCode and Pi harnesses.

Your three open questions from the last round remain open and I have not changed the answers unilaterally — weight-identity sufficiency, `n_gpu_layers`/backend-device self-invalidation, and whether archive-at-record should be gated behind an observed hit rate.

michaelneale:
Pushed `b1716aed`, addressing all nine CodeRabbit threads from this round. Every one was valid.

**Correctness**

- `disk_extras` grew without bound. It held a descriptor per page ever written, but the tier removes entries on its own schedule — LRU eviction in `enforce_budget`, `quarantine` on a verification failure — and neither path can see that map. Now pruned after any operation that can remove an entry, including inside `lookup`, since a failed verification quarantines the page before the result is inspected.
- A relative `SKIPPY_KV_DISK_TIER_DIR` measured free space against `/`. `existing_ancestor` walks parents until one exists; a relative path bottoms out at `""` whose parent is `None`, and the fallback returned the root. On any machine where the working directory is a separate mount, both the 16 GiB floor and the derived budget came from the wrong filesystem. Relative paths now resolve to `.`, absolute ones still to `/`. Two tests.
- The disk-miss telemetry event built a bare attribute map, so it could not be joined with hit and error events on session or model — the attribution needed to compute a hit rate. It now uses `openai_attrs` plus `insert_disk_tier_attrs` like every other branch.

**Docs**

- The KV-layout doc block ran into the platform-identity block without a separating blank line, so rustdoc rendered the layout rationale on `update_platform_identity` and `update_layout_identity` had none. Moved.
- Directory-layout fence tagged `text` (MD040).
- Split the ISWA families out of the dense-attention row in the family table — Gemma 3/4 appeared as plain dense while retention is declined for them further down.
- Dated the superseded "split topologies unmeasured" claim and replaced it with what is actually still unvalidated.
- Corrected the `skippy.kv.archive_status` text: it is emitted through telemetry, not written to the log. Without `SKIPPY_TELEMETRY_STDERR=1` or an OTLP endpoint it is not recorded anywhere. This one matched a real trap — it is why the stage-0 shallow-archive bug earlier in this branch had to be diagnosed from on-disk index JSON.

Validation at `b1716aed`: `cargo fmt --all --check` clean, clippy `-D warnings` clean on skippy-cache and skippy-server, skippy-cache 105 + 7 integration, skippy-server 444.

**Split re-measurement at this head: attempted, did not form.**

I tried to reproduce the two-machine result at the current head rather than rely on the 2026-08-12 numbers. Bundle built from `b1716aed`, deployed and codesigned on the second machine, topology pinned with `--split --split-topology-lock` (matching `manifest_sha256`, full endpoint ids, coordinator as stage 0).

The peers joined and the coordinator was elected, but the QUIC path never went direct — it stayed on the relay at ~215ms despite an 8ms LAN ping and firewalls disabled on both hosts. The planner correctly refused: `excluded [stage_path_relay_only]`. With `MESH_SPLIT_ALLOW_RELAY=1 MESH_SPLIT_MAX_RTT_MS=400` it reported `found 1 eligible` and would not admit the relayed peer as a stage host. No inference ran, so there is no number from this attempt.

The split evidence in the description therefore remains the 2026-08-12 measurement, four commits back, on a relay path. `446b9ecb` has changed page identity since. I believe split behaviour is unaffected — the change is symmetric across both stages and both nodes are the same architecture — but that is reasoning, not a measurement, and I would rather state it that way than imply I re-verified it.

Wired-link validation is the remaining gap, and it also answers the boundary-bandwidth question left open in the plan doc.

michaelneale:
🤖 Validation results from testing this branch (merged with current main) on M4/Metal:

**Restart survival works.** Qwen/Qwen3-4B-GGUF:Q4_K_M, 2.4k-token prompt, `--kv-cache-disk auto`:

| | cached_tokens | prompt_ms |
|---|---|---|
| cold | 0 | 18,609 |
| warm, same process | 2,304 | 1,657 |
| **first request after full restart** | **2,304** | **232** |

A ~176MB `.kvp` page was written at record time and restored across the restart — exactly the cold-start fix agent workloads need (our buzz-agent turns carry ~15k-token prefixes; re-prefill on a 30B costs ~100s).

**Gap found: locally-served GGUFs never get the disk tier.** `mesh-llm serve --gguf /path/to/model.gguf` produces a stage config with no `manifest_sha256`/`source_model_sha256`, so `open_disk_tier` declines ("no valid content digest"). The decline is one stderr line and easy to miss — the tier looks enabled but stores nothing. That's how we served Muse-Glimmer-30B (local GGUF + mmproj) and got zero disk pages while in-process caching worked fine. Suggestion: hash the local GGUF (or reuse the HF blob sha256 already present in hub cache paths) so `--gguf` models get the same restart survival — arguably these are exactly the models power users serve.

Also observed: Qwen3.5-4B (kv_recurrent family) wrote no pages in the same setup where dense Qwen3-4B did — haven't dug into whether that's expected for the recurrent payload path.

**Bigger suggestion:** given the safety rails already here (16GiB min-free floor, free-space-derived auto budget, per-stage shares, refuse-on-no-digest), this seems like a candidate for **on-by-default** when disk allows. The failure mode of "off" is silently terrible cold starts; the failure mode of "on" is bounded disk use that the budget policy already guards. A one-line INFO log stating tier status + budget at startup would make either default much more observable.

ndizazzo (changes requested):
Thanks for pushing this as far as you have. The measured restart wins are compelling, and the cache identity, corruption handling, and split work have clearly had a lot of thought put into them.

I’m requesting changes because there are still a few things I don’t think we should merge as-is.

## P1 — must fix before merge

1. **The new CLI settings mutate the process environment after Tokio’s multithreaded runtime has been created.**

   `configure_kv_disk_cache()` calls unsafe `std::env::set_var()` from `configure_run_auto_process_state()`. At that point the binary has already built the multithreaded Tokio runtime, and the code can’t establish that no other thread is reading the environment. The TODO comments acknowledge the missing safety proof. Please pass typed cache settings into the stage/runtime configuration, or apply them before Tokio is constructed.

   Relevant code: `crates/mesh-llm-host-runtime/src/runtime/run_auto.rs:579-653` and `crates/mesh-llm/src/main.rs:36-42`.

2. **The node disk-budget allocator permanently consumes four fixed shares and never returns them.**

   `POOL` is a process-global `OnceLock<Mutex<u64>>`, and `claim_stage_share()` subtracts a reservation permanently. A solo stage can use only one quarter of the configured node budget, unused capacity is stranded, and after four stage openings every later stage is denied the disk tier. Unloading a model or stage doesn’t return its reservation, so normal model churn can disable the cache for the rest of the process lifetime. Please move this to a node-owned allocator with stage-lifetime reservations, or otherwise make claims reclaimable and able to use currently unallocated capacity.

   Relevant code: `crates/skippy-server/src/kv_integration/disk_budget.rs:55-139`.

3. **Generated native runtime artifacts are committed to the branch.**

   `dist/native-runtimes/` contains roughly 49 MB of generated macOS dylibs, a tool binary, a packaged archive, and its checksum. These shouldn’t become permanent source-history objects. Please remove them from the PR and ignore the generated output path if it isn’t intentionally tracked elsewhere.

4. **The new telemetry paths export raw error strings.**

   Several new `skippy.kv.error`, `skippy.kv.archive_error`, and related attributes use `error.to_string()`. Those strings are unbounded/high-cardinality and can include local cache paths or other machine-specific detail before being sent to an OTLP collector. Please replace them with bounded error classes/reason codes. Detailed text can stay in a local, redacted diagnostic path.

   Examples: `crates/skippy-server/src/kv_integration/dense_disk.rs:346-350`, `crates/skippy-server/src/frontend/local_generation/token_generation.rs:594-600`, and the new archive paths in `binary_transport/binary_kv.rs` and `frontend/prefix_cache.rs`.

## P2 — correctness and operator experience

5. **Invalid `--kv-cache-disk` values are accepted and then silently downgraded to “cache off.”**

   The Clap surface accepts an arbitrary `String`; parsing happens later, and a bad value only prints to stderr while the node continues serving. A typo in an explicitly requested persistence feature shouldn’t look like a successful startup. Please parse this into a typed Clap value (`auto` or a positive size) and return a normal CLI validation error.

   Relevant code: `crates/mesh-llm-cli/src/parser/commands.rs:674-683` and `crates/mesh-llm-host-runtime/src/runtime/run_auto.rs:638-650`.

6. **The user-facing capability docs contradict the current ISWA implementation.**

   The new patch adds composite ISWA KV pages, but `website/src/docs/pages/kv-disk-cache.md:115-127` still says Gemma 3/4 and other hybrid ISWA models are declined. `docs/skippy/KV_RETENTION_PLAN.md` also has multiple sections describing composite ISWA support as future work. Please update both sources to describe what the current head actually supports, what’s still unsupported, and what has been validated.

## P3 — focused coverage gap

7. **There’s no end-to-end regression proving a disk hit reaches the durable request log as cached-token usage.**

   The plumbing appears to be present: a disk restore sets `cached_prompt_tokens`, `GeneratedText::usage()` maps that into OpenAI cached-token usage, the lifecycle observer records `UsageRecorded.cached_prompt_tokens`, and the Logs UI renders it. Please add a focused integration test so this useful operator signal can’t regress silently.

## Recommended logging follow-up

There’s a good fit with the logging stack from #1174–#1176, but I wouldn’t ingest the `stage.openai_kv_*` telemetry stream wholesale. `docs/LOGGING.md` deliberately separates durable operator logs from low-level KV/cache-pressure telemetry.

I’d suggest:

- Keep the existing cached prompt-token usage projection and add the end-to-end test above.
- Add one bounded, request-owned cache outcome through the OpenAI lifecycle layer, with fields like `source = exact|resident|disk|none`, `outcome = hit|miss|error`, `cached_tokens`, `suffix_prefill_tokens`, and `restore_ms`.
- Show that outcome in the Request Inspector’s Overview and Timeline/Diagnostics views.
- Promote cache startup decisions into bounded operational-ledger events: invalid configuration, missing digest, insufficient space, directory ownership conflict, budget exhausted, and tier-open failure.
- Optionally add a compact archive diagnostic (`archived`, bounded skip reason, or bounded failure class) with safe numeric summaries such as archived tokens/bytes and duration.
- Keep occupancy, promotions, evictions, corruption, verification, and other aggregate counters in telemetry/status rather than adding a durable row per stage or cache operation.
- Don’t persist raw errors, file paths, page/sequence IDs, prompt token IDs, model digests, or payload data in the logging service.

### Prompt for an implementation agent

Copy/paste this into an AI coding agent after the merge-blocking fixes are handled:

```text
You’re working in Mesh-LLM/mesh-llm on PR #1228’s KV disk-cache observability. Implement the logging enhancements below without piping low-level Skippy telemetry directly into the durable logging ledger.

Start by reading AGENTS.md and docs/LOGGING.md. Preserve the privacy and ownership boundaries introduced by PRs #1174, #1175, and #1176.

Goals:
1. Add an end-to-end regression test proving that a successful disk-prefix restore reaches the OpenAI lifecycle observer as UsageRecorded.cached_prompt_tokens and is exposed through the logs API/UI schema.
2. Add one typed, bounded request lifecycle projection for cache outcome, owned by the OpenAI lifecycle layer rather than skippy-server writing directly to the log store. Use an additive event or equivalent typed projection with:
   - source: exact | resident | disk | none
   - outcome: hit | miss | error
   - cached_tokens: non-negative integer
   - suffix_prefill_tokens: non-negative integer when authoritative
   - restore_ms: bounded non-negative duration when authoritative
3. Persist and expose the new projection through mesh-llm-events, the log store/API DTO, and the typed console schema.
4. Render a compact “Prefix cache” summary in Request Inspector Overview and add the event to Timeline/Diagnostics. Keep the UI useful for both hits and misses without adding one row per stage/candidate.
5. Add bounded operational-ledger events for authoritative cache startup outcomes already owned by Rust: enabled, invalid configuration, missing content digest, insufficient disk space, cache-directory ownership conflict, node budget exhausted, and tier-open failure. Use stable codes, reason codes, severity, model/runtime-instance correlation where available, and safe numeric summaries.
6. Optionally expose one compact archive outcome in Diagnostics: archived, a bounded skip reason, or a bounded failure class, plus authoritative numeric summaries such as archived tokens, bytes, and duration.

Privacy and cardinality requirements:
- Never persist/export raw error strings through the new logging projection.
- Never include local paths, page IDs, sequence IDs, prompt token IDs, predicted tokens, prompt hashes, model digests, request/response bodies, credentials, or endpoint URLs.
- Replace failures with bounded enums/reason codes.
- Keep disk entries/bytes, promotions, evictions, corruption, verification, and other aggregate health counters in telemetry/status; don’t create durable request events for them.
- Keep request-path logging non-blocking and fail-open.

Compatibility requirements:
- Make changes additive; don’t change mesh protobuf/gossip, plugin protocol, or the Skippy native ABI for logging.
- Keep old log records and older hosts readable by the console.
- Follow the existing logging sanitization, replay, retention, and trusted-local API conventions.

Testing:
- Add Rust tests for serialization, sanitization, lifecycle projection, persistence/API DTOs, and streaming/non-streaming cached-token propagation.
- Add UI schema/component tests for hit, miss, absent/older-host data, and incomplete evidence.
- Use `just` for repository builds and run Cargo commands serially, following AGENTS.md and docs/design/TESTING.md.

Deliver a concise summary of the event contract, privacy decisions, UI behavior, tests run, and any intentionally deferred telemetry-only work.
```

Once the P1/P2 items are addressed, I’m happy to take another pass.

michaelneale:
🤖 Follow-up: re-tested at fb141656 (merged with main) — **the local `--gguf` gap is closed**. Muse-Glimmer-30B Q4_K_M served via `--gguf` + `--mmproj`, `--kv-cache-disk auto`:

| | cached_tokens | prompt_ms |
|---|---|---|
| cold | 0 | 11,040 |
| warm, same process | 2,304 | 618 |
| **first request after full restart** | **2,304** | **662** |

~58MB .kvp page written at record time and restored across restart; no digest decline. Cold-start prefill for our long agent prefixes (~15k tokens, ~100s on this 30B) effectively disappears across restarts now.

Minor test-setup note: the ABI bump to 0.1.39 means a stale runtime under `target/debug/native-runtimes/` (0.1.38) fails resolution with "no compatible native runtime" even after `just build-mac` refreshes `dist/` — needed a manual copy. Cosmetic, but might trip up others testing the branch.

ndizazzo (approved):

michaelneale:
note disk tier is not off by default I believe

PR #1237, opened 2026-08-11
Keep split-serving KV retention from decaying on warm requests

michaelneale:
Stacked on #1228 (`feat/kv-prefix-retention`). Draft, like its base.

## What changes for users

On a split mesh, a stage's durable KV prefix cache no longer decays. Before this, each warm request archived a shallower prefix than the last, so a node that had been serving happily for a while would restart and restore almost nothing — the disk tier looked like it worked and then quietly stopped mattering. This is the one blocker called out in issue #1226 for split retention and split seeding.

## The two gates

Both in the chunked recorder (`binary_kv.rs`, `maybe_record_binary_prefill`):

1. **`min_record_tokens` vetoed archival, not just recording.** It is the restore length; candidates at or below it are resident already, so skipping the *resident* re-record is right. But the archive selector was only fed from candidates that passed that gate, so on a warm restore the recorder declined to archive exactly the shared bulk it had just served. Candidates below the restore point are now offered to `ArchiveCandidate` while resident re-recording stays skipped. `candidate_already_resident` is a named predicate so the two decisions read as distinct.
2. **Archival was nested inside `config.downstream.is_some()`.** The last stage of a chain has no downstream, so it never archived anything. Archival now runs once per request for every stage, unchanged in cost (still at most one export per request, under the runtime lock).

## Validation

- `cargo test -p skippy-server --lib` — 399 passed, 0 failed
- `cargo clippy -p skippy-server --all-targets -- -D warnings` — clean
- New tests: `record_gate_tests::restored_candidates_are_not_re_recorded_into_the_resident_cache`, `record_gate_tests::restored_candidates_are_still_offered_for_archival`

**Not yet done, and the evidence #1226 asks for:** a real split run showing per-stage archive depth roughly symmetric across stages, plus a split restart with a measured gain. Unit tests prove the gate logic; they do not prove the retention win.

### DIFF
--- a/crates/skippy-server/src/binary_transport/binary_kv.rs
+++ b/crates/skippy-server/src/binary_transport/binary_kv.rs
@@ -579,6 +579,16 @@ pub(in crate::binary_transport) fn maybe_lookup_binary_prefill(
     result
 }
 
+/// Whether a record candidate is already covered by this request's restore.
+///
+/// `min_record_tokens` is the restored prefix length. Candidates at or below
+/// it are resident already, so recording them again is wasted work — but they
+/// are still the right thing to *archive*, which is why this is a named
+/// predicate rather than an inline `continue` guard.
+fn candidate_already_resident(token_count: u64, min_record_tokens: u64) -> bool {
+    token_count <= min_record_tokens
+}
+
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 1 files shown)


michaelneale:
Folding into #1228 — keeping one issue (#1226) and one PR (#1228).

PR #1242, opened 2026-08-11
fix(skippy): serialize native proposal lifecycle

i386:
## Summary

Fixes the native proposal lifecycle failure observed in the official Harbor SWE-Gym smoke.

- Uses ordered per-token lifecycle commits instead of re-handing accepted proposal suffixes.
- Validates generated-token positions at proposal and report boundaries.
- Serializes report handoff behind prior commits and fences the next proposal/finish behind passive completion.
- Discards detached candidates with the correct deadline outcome.
- Adds queue-level race, timeout, lifecycle, and deadline regressions.

This PR is standalone and can land independently of #1239.

## Validation

- 28 native serving plugin host tests passed.
- 15 linear proposal tests passed.
- 3 local-generation tests passed.
- Formatting, all-target build, Clippy with warnings denied, and diff checks passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved token-count tracking during local generation and proposal handling.
  - Prevented mismatches between generated tokens and proposal receipts.
  - Added safeguards for token-count overflow conditions.
  - Ensured candidates are properly discarded when proposal responses are dropped or time out.

- **Reliability**
  - Improved ordering and completion of proposal, report, discard, and finish callbacks.
  - Strengthened lifecycle handling during shutdown and under queue pressure.
  - Added validation for missing or inconsistent generation state.
  - Improved visibility into report-delivery failures.
  - Added per-token generation commit tracking for more accurate receipts.


crates/mesh-native-serving-plugin-host/src/plugin_dispatch.rs
```diff
@@ -466,10 +528,49 @@ fn run_proposal(
...
+    }
+
+    // Reports and discards run on the passive worker so they cannot consume
+    // proposal callback time. This fence serializes the passive terminal
+    // callbacks with the primary proposal queue. The wait is not abandoned;
+    // after it completes, the original absolute deadline is checked again.
+    let passive_wait_timeout = deadline
+        .saturating_duration_since(Instant::now())
+        .min(CLEAN_SHUTDOWN_TIMEOUT);
+    if let Err(error) = fence_passive(passive_queue, passive_wait_timeout) {
```
ndizazzo: It looks like this timeout races the caller’s `recv_timeout(remaining)`: both expire at roughly the same deadline, so whichever thread wakes first determines whether we report `SourceError` or `HostDeadlineExceeded`. I can reproduce the new test failing on macOS because the passive-fence error wins that race. Could we classify a fence wait that consumes the query deadline as a deadline outcome, and reserve `SourceError` for actual queue/fence infrastructure failures?

crates/mesh-native-serving-plugin-host/src/plugin_dispatch/tests.rs
```diff
@@ -293,13 +390,29 @@ fn blocking_commit_cannot_extend_the_proposal_deadline() {
...
-        .propose(
-            LinearProposalQuery::new(1, 2, 1, 2, 1, 8, started + Duration::from_millis(5))
-                .with_pending_token_ids(vec![4].into_boxed_slice()),
-        )
+        .propose(LinearProposalQuery::new(
+            1,
+            2,
+            1,
+            1,
+            0,
```
ndizazzo: I think this query is using the pre-commit positions even though the generated token has already been committed. At this point the next query should have `committed_token_count = 2` and `decode_step = 1`. With the current `1/0` values, the proposal can be rejected for a position mismatch, so the test may pass without exercising the blocking-commit deadline behavior it intends to cover. Could we update those positions and, ideally, assert that the proposal source was actually reached?

michaelneale:
get it clean and seems like a necessary thing to me

micspiral:
🤖 Review posted on behalf of @michaelneale (agent: MeshBuilder).

Reviewed at head `6941a27b`. Verified locally in a clean worktree at that SHA:

- `cargo check -p skippy-server -p mesh-native-serving-plugin-host --all-targets` — clean
- `cargo test -p mesh-native-serving-plugin-host` — 28/28, run 5x, no flake (including `slow_commit_abstains_before_plugin_dispatch_and_later_positions_recover`)
- `cargo test -p skippy-server` — pass

Nothing major left in the code. Mic is happy for this to land **as long as CI goes green**.

One mechanical blocker remains, and it is a conflict with `main` rather than a bug in the branch: `rust-clippy (1)` fails with `missing field generated_token_count in initializer of LinearProposalReceipt` at `linear_decode.rs:480`. That line only exists in the merge ref — the branch is behind `main`, and #1247 added a `LinearProposalReceipt` literal that does not set the new field. Merge `main` and add `generated_token_count` to that literal.

Two non-blocking follow-ups:

1. `fence_timeout = remaining.min(CLEAN_SHUTDOWN_TIMEOUT)` still equals the caller`s `recv_timeout(remaining)` for every real proposal. `fence_consumes_query_deadline` makes the *classification* deterministic, which is why the test stopped flaking, but two timers still expire on the same instant. Making the fence deadline strictly earlier by a small epsilon would fix it by construction.
2. The final design is a two-writer model — `pending_token_ids` on the query *and* per-token lifecycle commits — not the single-writer model the PR body describes. Worth a paragraph in the description, since the two mechanisms now have to agree and nothing enforces that.

PR #1245, opened 2026-08-11
fix: Skippy Nemotron MTP loading

ndizazzo:
## Summary

- load optional NextN/MTP tensors for final-stage Skippy model loads and native MTP sidecars
- select the MTP source explicitly: disabled, integrated, or external
- use a plain KV cache for Nemotron-H MoE’s appended dense-attention MTP layer

## Root cause

`llama_model_params.load_mtp` defaults to false, so Skippy omitted Nemotron’s NextN tensors before creating its native MTP context. Once enabled, the hybrid cache path also excluded the appended MTP layer, causing MTP execution to access an invalid KV-cache index.

A terminal stage’s `include_output` flag only identifies output ownership; it does not select an MTP mode. Inferring MTP from that flag loaded NextN tensors and created an integrated context even when native MTP was disabled. It also blocked an explicitly configured external MTP sidecar from attaching.

## Impact

NVIDIA Nemotron 3.5 Lightning can load and execute native MTP speculation through Skippy. Disabled MTP now retains no optional MTP tensors or KV context, while an external sidecar can attach to a target that contains integrated MTP tensors.

## Skippy ABI change

ABI version: 0.1.37 → 0.1.38.

- Added `enum skippy_mtp_source { DISABLED, INTEGRATED, EXTERNAL }`.
- Added `enum skippy_mtp_source mtp_source` to `struct skippy_runtime_config`.
- Updated the Rust FFI mirror and regenerated the native API reference. Older native runtimes are intentionally rejected by ABI negotiation.

## Validation

- fresh pinned llama.cpp checkout replayed patches `0001`–`0018` through `scripts/prepare-llama.sh pinned`
- clean static CPU native build via `scripts/build-llama.sh`
- independent C11 and C++17 public-header compilation
- `cargo test -p skippy-runtime --lib` (71 tests)
- `cargo test -p skippy-server --lib` (398 tests; model-backed GLM MTP fixtures remain opt-in)
- regenerated and checked the Skippy API reference
- `just build`
- just-built Metal runtime previously loaded `NVIDIA-Nemotron-3.5-Lightning-30B-A3B-Q4_K_M` and completed a native-MTP request: 12 drafted, 3 accepted, 0 rejected; 5 generated tokens at 42.38 tok/s

## Summary by CodeRabbit

* **New Features**
  * Added configurable MTP sources: disabled, integrated, or external.
  * Native and externally attached draft models now use the selected MTP source.
  * Added MTP tensor loading for supported model configurations.
  * MTP is disabled by default for existing runtime configurations.
* **Bug Fixes**
  * Prevented incompatible MTP configurations with external draft models.
* **Documentation**
  * Updated public API documentation and ABI details for MTP source configuration.

### DIFF
--- a/third_party/llama.cpp/patches/0016-skippy-load-MTP-tensors-for-native-draft-attachment.patch
+++ b/third_party/llama.cpp/patches/0016-skippy-load-MTP-tensors-for-native-draft-attachment.patch
@@ -1,0 +1,97 @@
+From 323eebd494847353d76b5c7eb167be90bf8db937 Mon Sep 17 00:00:00 2001
+From: Mesh-LLM CI <ci@mesh-llm.local>
+Date: Tue, 11 Aug 2026 16:14:30 -0400
+Subject: [PATCH] skippy: retain and execute final-stage MTP tensors
+
+---
+ src/llama-model.cpp          | 12 ++++++++++--
+ src/skippy/model_loading.cpp | 15 +++++++++++++++
+ 2 files changed, 25 insertions(+), 2 deletions(-)
+
+diff --git a/src/llama-model.cpp b/src/llama-model.cpp
+index 8824c03fca09c9c64d5b011fc139b463bc3859d1..b26bded77b65340ca6905652a176be6e6b5b5d2a 100644
+--- a/src/llama-model.cpp
++++ b/src/llama-model.cpp
+@@ -2142,6 +2142,12 @@ llama_memory_i * llama_model::create_memory(const llama_memory_params & params,
+                 const bool mtp_on_hybrid_qwen35 =
+                     params.ctx_type == LLAMA_CONTEXT_TYPE_MTP &&
+                     (arch == LLM_ARCH_QWEN35 || arch == LLM_ARCH_QWEN35MOE);
++                // Nemotron-H MoE's NextN block is a dense-attention head. It
++                // needs a plain KV cache for the appended MTP layer, not the
++                // hybrid trunk cache (which intentionally excludes it).
++                const bool mtp_on_hybrid_nemotron =
++                    params.ctx_type == LLAMA_CONTEXT_TYPE_MTP &&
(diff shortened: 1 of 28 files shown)


third_party/llama.cpp/patches/0016-skippy-load-MTP-tensors-for-native-draft-attachment.patch
```diff
@@ -0,0 +1,97 @@
...
+            has_head;
+ }
+ 
++static void skippy_enable_mtp_tensors_for_final_stage(
++        const skippy_runtime_config * config,
++        llama_model_params & params) {
++    // Final stages create an LLAMA_CONTEXT_TYPE_MTP sidecar when the model
++    // advertises NextN layers. llama.cpp normally skips those optional tensors
++    // unless this is explicitly enabled during the shared model load.
++    params.load_mtp = config != nullptr && config->include_output;
```
i386, line 62: **[P1] Select the MTP source explicitly instead of inferring it from `include_output`**

`include_output` only identifies a terminal stage; it does not represent the resolved MTP mode. As written, every terminal stage sets `load_mtp` and can create an integrated MTP context even when `native_mtp_enabled` is false. That still loads the optional NextN weights and allocates MTP KV memory, which can make a model OOM despite native MTP being disabled. It also prevents an explicitly configured external MTP sidecar from attaching when the target contains integrated MTP tensors, because attachment rejects an existing `mtp_ctx`. Please propagate an explicit disabled/integrated/external MTP selection into the native load configuration and gate both `load_mtp` and integrated-context creation on it. If this extends `skippy_runtime_config`, update the Skippy ABI version and Rust mirror. Please add regression coverage for disabled MTP and for an external sidecar paired with a target containing integrated MTP tensors.

i386 (commented):
Sol high-effort review found one actionable P1. The patch queue applies cleanly to the pinned llama.cpp revision, and the submitted integrated Nemotron MTP smoke test does not cover the configuration paths below.

PR #1247, opened 2026-08-11
fix: complete remaining Skippy MTP runtime callers

i386:
## Summary

- set explicit disabled MTP selection for the remaining skippy-prompt runtime opens
- mark the skippy-quantize attachment probe as an external MTP source
- assert the external source in the quantize regression test

## Context

This PR is stacked on #1245. The current base now owns the ABI, llama patch queue, source resolution, and the other benchmark and correctness RuntimeConfig initializers. This follow-up contains only the three remaining call-site fixes.

## Validation

- cargo fmt --all --check
- cargo check --locked for skippy-prompt and skippy-quantize
- targeted skippy-quantize attachment configuration test
- git diff --check
- local ancestor and merge-tree verification against the current #1245 head


ndizazzo:
Let's stack with #1245 ?

i386:
@ndizazzo yep added to the stack and will fix the conflicts.

PR #1253, opened 2026-08-12
Reuse recurrent KV across growing chat turns

i386:
## Summary

Qwen3.5/KvRecurrent state was only captured at the pre-decode/generated-token boundary. A subsequent ChatCompletion retokenizes the prior assistant message and inserts a new user turn, so that checkpoint was not a token prefix and `cached_tokens` stayed zero.

This PR is now based directly on `main` and includes the request lifecycle/session foundation needed by the recurrent checkpoint path:

- derives the `add_assistant=false` message-history boundary from the same chat template;
- retokenizes it with the loaded model and requires an exact strict token prefix of the full prompt;
- records/reuses recurrent state only after canonical tracked/Rust/native position validation and only when storage reports `record.stored`;
- preserves raw post-decode checkpoints and fail-closed initial linear proposal gating;
- emits replica-unique request IDs while preserving stable trusted agent-session identity;
- classifies streaming completion, backend error, cancellation, and client disconnect outcomes exactly once;
- avoids consuming global generation queue capacity while waiting for a contended trusted session;
- keeps non-chat and media paths unchanged.

## Validation

- `cargo check -p openai-frontend`
- `cargo test -p openai-frontend` — 163 unit tests, 7 integration tests
- `cargo clippy -p openai-frontend --all-targets -- -D warnings`
- `cargo test -p skippy-server --lib` — 408 passed, 3 ignored
- `cargo check -p mesh-llm`

`cargo clippy -p mesh-llm --all-targets -- -D warnings` reaches the existing `mesh-llm-host-runtime` warnings: 28 unfulfilled `dead_code` expectations unrelated to this PR.

## Summary by CodeRabbit

* **New Features**
  * Added recurrent KV-cache support for local text generation.
  * Enabled reuse of compatible chat-history prefixes to improve generation efficiency.
  * Added checkpoint handling for reliable cache reuse during decoding.

* **Bug Fixes**
  * Improved token tracking after generation hooks.
  * Prevented incompatible or invalid prompt prefixes from being used for caching.
  * Preserved consistent behavior when cached and uncached generation paths are used.

* **Tests**
  * Added coverage for cache reuse, checkpoint boundaries, proposal gating, and cached versus uncached chat behavior.


michaelneale:
🤖 *Deep review by Mic's robot (agent-assisted; cross-checked with an independent expert model). No testing performed — this is a design/correctness read of the diff plus the contracts it relies on (`record_exact_state`, `canonical_session_position`, production family policy).*

## Verdict

The design direction is right and the safety posture is genuinely good: strict token-prefix validation before any boundary is used, canonical tracked/Rust/native position agreement before any export, and `record.stored` as the only success signal. `post_decode_checkpoint_tokens` has a clearly documented contract ("runtime consumed everything except the newest token") and the linear-proposal receipt accounting against that invariant checks out, including the proposal-stop case.

That said, I think there are **two blockers and one performance concern** to resolve before this merges.

## Blockers

### 1. Linear proposals can be starved for the entire request

In `run_decode_loop`, the gate only opens when `record_post_decode_exact_state` returns `true`, and the retry condition is `generated_token_ids.len() == 1`. If the first record attempt fails, the length moves past 1 and the gate **stays closed for the rest of the request**.

`record_exact_state` returns `Ok(None)` (→ `false`) whenever `token_count < min_tokens` — and the production recurrent policy (`kv_recurrent_policy` in `family_policy.rs`) sets `min_tokens: 256`. So on any recurrent chat with a prompt shorter than ~255 tokens, the checkpoint can *never* be stored, and linear-proposal ingress is silently disabled for the whole request. The PR's tests don't see this because they set `min_tokens: 1`.

The gate's purpose is to guarantee one serial decode before a proposal advances the session past the full-prompt boundary. Once the `len == 1` attempt has happened, keeping proposals off cannot make an unstorable boundary storable — suggest opening the gate on **attempted**, not **recorded**.

### 2. Unconditional per-request overhead + optional work failing the request

`prepare_chat_prompt` now renders the chat template a second time (`add_assistant=false`) and `text_generation.rs` tokenizes the result — **on every chat request**, even when `kv` is `None` or the payload isn't `KvRecurrent`. `tokenize` takes the global runtime mutex, so this adds lock contention against live decode loops on all non-recurrent serving for zero benefit.

Worse, both the extra render and the tokenize use `?`: a failure in purely-optional cache-candidate work now fails an otherwise valid request. Suggest gating the whole derivation on the backend's KV payload and treating candidate failure as cache-bypass (with telemetry), not an error.

## Performance concern

### 3. Up to ~4 full exact-state exports per request, under the runtime mutex

With this PR a single request can attempt: the mid-prefill chat-boundary record, the existing full-prefill record, the first post-decode record (between token 1 and token 2), and the final post-decode record. Each `KvRecurrent` export copies the full attention-KV prefix (`export_kv_page(0, token_count)`) plus recurrent state while holding the global runtime mutex — the first post-decode export sits directly in the inter-token path and stalls every other session.

Note the cache does **not** cheaply dedupe repeats: `ExactStateCache::record` removes and reinserts the entry before payload dedupe, so identical boundaries across turns pay the full export/hash cost each time. Combined with `max_entries: 16` in production, several boundaries per turn can also churn out genuinely useful longer entries.

Validation so far is a 0.8B model with tiny prompts. I'd want either mitigation (skip already-present page IDs before export; skip the final record when it names the same boundary as an earlier one) or a long-context, concurrent-session benchmark showing inter-token latency and mutex hold are acceptable.

## Non-blocking notes

- **Silent zero-benefit mode**: when a template renders history non-prefix-stably (thinking flags on the last user turn, tool variants, etc.), the strict prefix check safely discards the candidate — but nothing is emitted. A `candidate_rejected` / `token_prefix_mismatch` decision event would let operators distinguish "template incompatible" from "cache broken" when `cached_tokens` stays 0.
- **Split-prefill chunk boundaries**: splitting `prefill_chunked` at an arbitrary history boundary changes batch partitioning vs. one continuous prefill. Canonical-position validation proves token *count*, not state equivalence. The parity test covers one small case; worth exercising split points around `n_batch − 1 / n_batch / n_batch + 1` on a larger recurrent model.
- **Hook-injected tokens**: `decode_one_token` returns early on injection without pushing to `generated_token_ids`, so the named boundary diverges from what the runtime consumed — the canonical-position check then fails closed and records are skipped. Correct, but worth a comment so nobody "fixes" the position mismatch the wrong way.
- **Pre-existing, adjacent**: `KvRecurrent` restore imports KV then recurrent state; if the second import fails, `lookup_and_restore_kv` proceeds as a miss on a partially mutated session and prefills on top of imported KV. Not introduced here, but this PR increases how often that path is exercised — worth a follow-up issue.
- The ignored-test conversion of the receipt test (env-var skip → `#[ignore]`) is a nice cleanup; nothing in CI references those env vars.

Happy to re-review after the gate and gating-of-derivation changes. — Mic's robot

michaelneale (approved):
🤖 Approving on behalf of Michael Neale (review by CrocDundee AI agent).

LGTM — recurrent KV reuse across growing chat turns. Verified the mechanism, the fail-closed guards, and full green CI + CodeRabbit.

Correctness is fail-safe by construction:
- Restore is gated on a strict token-prefix match, so a bad/mismatched prefix is a cache miss, never a wrong-state restore.
- Recording is gated on `canonical_session_position == checkpoint_token_count` (position-based, not text), which is the right off-by-one guard.
- Only affects `KvRecurrent`-payload models; dense/attention KV paths are untouched.

One non-blocking follow-up (file a ticket, not a merge blocker): the linear-proposal commit path (`linear_decode.rs:244`) advances `decoded_tokens`/`current`/`linear_context_tokens` but does not extend `state.generated_token_ids`. On proposal-driven recurrent responses the post-decode checkpoint is therefore skipped (canonical-position check trips) — a missed optimization, safe, but worth closing with a proposal-path checkpoint test. The current cross-turn test runs with proposals disabled so it doesn't cover this.

michaelneale (approved):

PR #1259, opened 2026-08-12
Expose Apple system provider through macOS SDKs

i386:
> [!WARNING]
> Experimental. This requires Apple silicon running **macOS Golden Gate (macOS 27)**. It is not a public/default routing path and the ad-hoc-signed QA artifact is not release eligible.

Stacked on #1256. Tracks the Phase 2 SDK-carrier work in #1246.

## Why

Apple Silicon users should be able to use the model already supplied and updated by macOS through the same OpenAI-compatible MeshLLM surface as every other model. That gives Mac applications private on-device inference, no model download, low setup cost, accelerator-backed execution, low network latency, and offline availability.

This complements Skippy rather than adding pipeline parallelism: `apple/system` is a whole-model local provider. MeshLLM routes the complete request to the capable Mac and the one shared Swift sidecar calls Foundation Models.

## What changed

- adds a shared typed Rust `ProviderHost` lifecycle that starts a provider-only embedded host and returns its loopback `/v1` base URL;
- exposes that lifecycle through UniFFI to Swift and Kotlin/JVM and through N-API to Node/Electron;
- keeps Foundation Models behavior, tool semantics, cancellation, usage, model identity, and accelerator access in the single `mesh-apple-runtime` Swift sidecar;
- packages the exact same signed sidecar bytes in platform-gated carriers:
  - macOS-only `MeshLLMAppleProviderResources` SwiftPM target;
  - optional Darwin-arm64 `@mesh-llm/apple-runtime-darwin-arm64` npm package;
  - `meshllm-apple-runtime-macos-arm64` JVM resource JAR;
- avoids shipping a macOS executable in Linux/Windows npm, iOS, or Android artifacts;
- adds one live conformance runner for Swift, Node/Electron, and Kotlin/JVM;
- updates the Apple runtime and SDK documentation with packaging, lifecycle, and release constraints.

The SDKs supply typed bundle roots and deliberately ignore process-global provider discovery variables. The signed sidecar process receives Apple accelerator access; the language process only owns supervision and the REST endpoint lifecycle.

## Try it

Prerequisites:

1. Apple silicon Mac running macOS Golden Gate (macOS 27).
2. Full Xcode installed and selected with `sudo xcode-select -s /Applications/Xcode.app/Contents/Developer`.
3. Xcode licence accepted, Swift 6.4 available, Node.js 20+, Rust, and JDK 21 installed.
4. This branch checked out on top of #1256.

Run all three real language carriers through the same REST suite:

```bash
just apple::sdk-carriers
[...]

PR #1260, opened 2026-08-12
Route Apple system models across private meshes

i386:
> [!IMPORTANT]
> **Experimental. Requires Apple silicon running macOS Golden Gate.** This is an early integration with Apple’s on-device Foundation Models framework. It is not yet a signed, notarized release artifact and it is intentionally unavailable on public meshes.

Stacked on [#1259](https://github.com/Mesh-LLM/mesh-llm/pull/1259). Advances [#1246](https://github.com/Mesh-LLM/mesh-llm/issues/1246).

## Why this matters

Apple silicon Macs already contain a power-efficient, privacy-preserving system language model. Mesh should let an Apple user contribute that capability without downloading model weights, converting checkpoints, allocating a second copy of the model, or learning a new API. Existing OpenAI-compatible clients and every macOS-capable Mesh SDK can use the same `apple/system` model name through the normal REST surface.

For the product, this creates a useful new class of Mesh node:

- idle Macs become whole-model inference replicas with no Skippy pipeline split;
- prompts and generation remain on-device, with no cloud inference dependency;
- Apple owns the model delivery and hardware acceleration path, so Mesh owns routing and lifecycle rather than model distribution;
- multiple Macs add availability and request concurrency, even though they do not combine their context windows;
- Rust, Swift, Node/Electron, and Kotlin/JVM clients all reach one signed Swift sidecar through the same provider contract instead of carrying framework-specific logic in every SDK.

## What this PR adds

This is the single Phase 3 private-mesh routing PR. It builds on the provider artifact and all-SDK carrier work in #1259.

- Advertises `apple/system` and the exact OS model generation, for example `apple/system@27.0`, only inside private meshes.
- Routes each request to one capable Mac. Apple system models never enter Skippy split planning, `auto`, virtual `mesh`, or Mixture-of-Agents selection.
- Publishes provider capacity, active requests, and queued requests as additive gossip fields.
- Prefers an idle replica with known load, then a peer with unknown load, then a saturated replica. Existing request affinity wins over live load so a conversation remains stable.
- Retries another replica only before response streaming begins. Mesh never splices two model streams into one response.
- Withdraws the provider after sustained health failures, process exit, or shutdown, and re-advertises it after recovery.
- Makes a provider-only Mac routable even when it is not serving a GGUF model.
- Aggregates model generation, replica count, capacity, active/queued work, and available slots in `/v1/models` and the management API.
- Shows Apple provider identity and load in the web console.
- Adds a deterministic private-mesh QA harness that starts two identity-isolated nodes on one Golden Gate Mac and proves that a second request is sent over QUIC to the idle replica while the first is occupied.

The mesh protocol change is additive. Older peers ignore the new provider/load fields; newer peers interpret missing load as unknown.

## Context-window behavior

The provider follows Apple’s [TN3193 guidance](https://developer.apple.com/documentation/technotes/tn3193-managing-the-on-device-foundation-model-s-context-window) rather than estimating context from strings in Rust. The signed Swift runtime asks `SystemLanguageModel` for the exact token count across the prompt, instructions, tool declarations, and structured-output schema, and reserves the requested output budget before generation.

Apple documents a 4,096-token total context for the on-device model and notes that tool definitions consume that same budget; Apple recommends roughly three to five tools per request. Oversized requests now fail before generation with the structured `context_exceeded` error. Routing across more Macs improves concurrency and resilience, not the context size of an individual request.

## Runtime packaging and SDK shape

All Apple framework code stays in one Swift sidecar under `providers/apple/`. The backend-neutral Rust host discovers and supervises the packaged provider runtime; SDKs drive the same host/provider lifecycle and OpenAI-compatible endpoint. We do not embed a different Apple implementation into each language binding.

The final release process must sign the sidecar that imports Foundation Models, preserve the required entitlement and code-signing identity, package it beside the neutral host, and notarize the composed product. Local QA permits an explicitly enabled ad-hoc build. Production signing/notarization remains a release gate; this PR does not pretend that an ad-hoc executable is shippable.

### Prerequisites

1. An Apple silicon Mac running **macOS Golden Gate** with the on-device system model available.
2. Full Xcode selected, not only Command Line Tools:

   ```bash
   sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
   xcodebuild -version
   ```

3. This branch checked out. It is stacked on #1259, so the branch includes the earlier Apple provider and SDK carrier phases.

### Automated path

```bash
just apple::test
just apple::mesh
just apple::private-mesh
```

- `apple::test` runs the Swift provider tests.
- `apple::mesh` packages the sidecar and exercises completions, SSE streaming, tools, cancellation, versioned model IDs, restart, and cleanup through Mesh REST.
- `apple::private-mesh` builds Mesh, starts two isolated private-mesh nodes, verifies gossip and aggregate capacity, occupies node A, and proves the next request is routed over QUIC to idle node B.

### Run a local node manually

```bash
just apple::package
just build

export MESH_LLM_PROVIDER_RUNTIME_BUNDLE_DIR="$PWD/target/apple-runtime/package/meshllm-apple-runtime-darwin-arm64"
export MESH_LLM_APPLE_PROVIDER_ALLOW_AD_HOC=1

target/debug/mesh-llm --log-format json serve \
  --headless \
  --port 9337 \
  --console 3131
```

List the system-model aliases:

```bash
```

Request a completion:

```bash
curl -s http://127.0.0.1:9337/v1/chat/completions \
  --data-binary '{
    "model": "apple/system",
    "messages": [{"role": "user", "content": "Reply with exactly: apple runtime REST ready"}],
    "temperature": 0,
    "max_tokens": 32
  }' | jq
```

Exercise the QA tool loop:

```bash
curl -s http://127.0.0.1:9337/v1/chat/completions \
  --data-binary '{
    "model": "apple/system",
    "messages": [{"role": "user", "content": "Use the tool with key: rest-demo"}],
    "tools": [{
      "type": "function",
      "function": {
        "name": "mesh_fixture_lookup",
        "description": "Look up a fixture",
        "parameters": {
          "type": "object",
          "properties": {"key": {"type": "string"}},
          "required": ["key"]
        }
      }
    }],
    "temperature": 0
  }' | jq
```

For two physical Macs, start the first private node, obtain its invite token from the JSON log, and start the second with `--join <token>`. The full two-machine checklist is in `docs/design/TESTING.md`.

## Example REST output

Completion:

```json
{
  "model": "apple/system",
  "choices": [{
    "message": {"role": "assistant", "content": "apple runtime REST ready"},
    "finish_reason": "stop"
  }],
  "mesh_timing": {"elapsed_ms": 1083, "time_to_first_token_ms": 1000},
  "usage": {"prompt_tokens": 65, "completion_tokens": 9, "total_tokens": 74}
}
```

Tool execution:

```json
{
  "model": "apple/system",
  "choices": [{
    "message": {"role": "assistant", "content": "mesh-fixture-value-for-rest-demo"},
    "finish_reason": "stop"
  }],
  "mesh_tool_executions": [{
    "name": "mesh_fixture_lookup",
    "arguments": {"key": "rest-demo"},
    "result": "mesh-fixture-value-for-rest-demo"
  }],
  "usage": {"prompt_tokens": 327, "completion_tokens": 13, "total_tokens": 340}
}
```

Private-mesh proof:

```json
{
  "completion_content": "private mesh ready",
  "load_aware_remote_dispatch": true,
  "max_concurrent_requests": 2,
  "model": "apple/system",
  "private_peer_provider_runtime_visible": true,
  "provider_model_versions": ["27.0"],
  "replicas": 2,
  "status": "pass"
}
```

The model’s natural-language wording and timing vary by machine and OS model generation. The model ID and response schema are the stable contract.

## Validation

- `just apple::test` — 9 passed
- `cargo test -p mesh-llm-host-runtime --lib` — 1,950 passed, 8 ignored
- provider-supervisor focused tests — 11 passed
- `just apple::mesh` — completion, versioned completion, SSE, tool execution, cancellation, restart, cleanup passed
- `just apple::private-mesh` — provider gossip, two-replica aggregation, capacity, completion, and load-aware QUIC dispatch passed
- scoped Clippy with warnings denied — passed
- `cargo fmt --all -- --check` — passed
- focused UI tests — 64 passed
- `bash -n providers/apple/QA/private-mesh.sh` — passed

The full UI suite has one existing environment-specific failure in `DataModeContext.test.tsx` when Node’s experimental `localStorage` is unavailable (`969 passed, 3 skipped, 1 failed`). The focused status/model catalog tests touched here pass.

## Deliberate limits and follow-up

- Private meshes only. The provider is suppressed on public meshes.
- Explicit `apple/system` selection only; no implicit `auto` routing.
- One request occupies one Mac. This is replica routing, not pipeline parallelism.
- The exact version is the Apple OS model generation, not a Mesh-owned weight revision.
- The automated two-node test uses one Golden Gate Mac with isolated identities; a two-physical-Mac run remains required before release confidence.
- Core AI / imported SafeTensors conversion and quantization are a later provider phase, after the system-model route is proven.
- Signed, entitled, notarized product packaging is required before this can graduate from experimental status.

## Reviewer focus

- private/public advertisement policy and withdrawal behavior;
- mixed-version handling of additive gossip fields;
- affinity versus load-aware replica choice;
- the pre-stream-only failover boundary;
- provider-only host routability and lifecycle cleanup;
- packaging assumptions for the signed Swift sidecar.

### DIFF
--- a/crates/mesh-client/tests/mesh_types.rs
+++ b/crates/mesh-client/tests/mesh_types.rs
@@ -34,6 +34,11 @@ fn model_runtime_descriptor_not_ready_by_default() {
         identity_hash: None,
         context_length: None,
         ready: false,
+        provider_kind: None,
+        model_version: None,
+        max_concurrent_requests: None,
+        active_requests: None,
+        queued_requests: None,
     };
     assert!(d.advertised_context_length().is_none());
 }
@@ -45,6 +50,11 @@ fn model_runtime_descriptor_ready_returns_context_length() {
         identity_hash: None,
         context_length: Some(4096),
         ready: true,
+        provider_kind: None,
+        model_version: None,
+        max_concurrent_requests: None,
+        active_requests: None,
+        queued_requests: None,
     };
     assert_eq!(d.advertised_context_length(), Some(4096));
 }
--- a/crates/mesh-llm-host-runtime/src/api/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/api/mod.rs
@@ -71,8 +71,7 @@ use self::status::{
     OpenAiGuardrailsPayload, RuntimeCapabilityFlags, RuntimeLlamaPayload, RuntimeProcessesPayload,
     RuntimeStatusPayload, StatusPayload, build_runtime_processes_payload,
     build_runtime_stage_payloads, build_runtime_status_payload, derive_daemon_state,
-    enrich_provider_runtime_models,
-    runtime_stage_state_label, runtime_stage_wire_dtype_label,
+    enrich_provider_runtime_models, runtime_stage_state_label, runtime_stage_wire_dtype_label,
 };
 use crate::mesh;
 use crate::models::append_external_inference_models;
--- a/crates/mesh-llm-host-runtime/src/api/mod.rs
+++ b/crates/mesh-llm-host-runtime/src/api/mod.rs
@@ -71,6 +71,7 @@ use self::status::{
     OpenAiGuardrailsPayload, RuntimeCapabilityFlags, RuntimeLlamaPayload, RuntimeProcessesPayload,
     RuntimeStatusPayload, StatusPayload, build_runtime_processes_payload,
     build_runtime_stage_payloads, build_runtime_status_payload, derive_daemon_state,
+    enrich_provider_runtime_models,
     runtime_stage_state_label, runtime_stage_wire_dtype_label,
 };
 use crate::mesh;
@@ -555,14 +556,15 @@ impl MeshApi {
     }
 
     async fn runtime_status(&self) -> RuntimeStatusPayload {
-        let (runtime_status, openai_guardrails) = {
+        let (runtime_status, openai_guardrails, node) = {
             let inner = self.inner.lock().await;
             (
                 inner.runtime_data_collector.runtime_status_snapshot(),
                 inner.openai_guardrails.clone(),
+                inner.node.clone(),
             )
         };
-        build_runtime_status_payload(
+        let mut payload = build_runtime_status_payload(
             runtime_status.primary_model.as_deref().unwrap_or_default(),
             runtime_status.primary_backend,
             openai_guardrails,
@@ -570,7 +572,10 @@ impl MeshApi {
             runtime_status.llama_ready,
             runtime_status.llama_port,
             runtime_data::runtime_process_payloads(&runtime_status.local_processes),
-        )
+        );
+        let provider_runtimes = node.local_model_runtime_descriptors().await;
+        enrich_provider_runtime_models(&mut payload.models, &provider_runtimes);
+        payload
     }
 
     async fn runtime_processes(&self) -> RuntimeProcessesPayload {
@@ -902,9 +907,7 @@ impl MeshApi {
             runtime_status.llama_port,
             local_processes.clone(),
         );
-        node.refresh_stage_runtime_statuses(std::time::Duration::from_secs(2))
-            .await;
-        runtime.stages = build_runtime_stage_payloads(node.stage_runtime_statuses().await);
+        enrich_runtime_observations(&node, &mut runtime).await;
 
         let wakeable_nodes = wakeable_inventory.status_snapshot().await;
         let hardware = runtime_data_collector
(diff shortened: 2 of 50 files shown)


i386:

Issue #1220, opened 2026-08-10
install.sh 404s on Linux hosts with an NVIDIA driver but no CUDA toolkit — detect_cuda_major falls back to a legacy asset name v0.75.0 no longer publishes


Cam10001110101:
## Problem

On a Linux x86_64 machine with an NVIDIA driver installed but no CUDA toolkit (a common
shape for inference-only hosts), the documented one-liner fails:

```
$ curl -fsSL https://meshllm.cloud/install.sh | bash -s -- --no-setup
curl: (22) The requested URL returned error: 404
error: could not download release archive for mesh-llm-x86_64-unknown-linux-gnu-cuda.tar.gz or mesh-bundle.tar.gz
```

Expected: the installer selects the `-cuda-12` or `-cuda-13` bundle and installs.

## Steps to reproduce

1. Ubuntu 24.04.4, GTX 1650 Ti, driver 595.84, **no** nvcc / `/usr/local/cuda*` /
   `libcudart` anywhere on the system.
2. Run the install one-liner (same result from `meshllm.cloud/install.sh` and from
   `raw.githubusercontent.com/Mesh-LLM/mesh-llm/main/install.sh` — the two are currently
   byte-identical).

## Diagnostics

The failure chain in install.sh:

1. `recommended_flavor()` → `probe_nvidia` succeeds → flavor `cuda`.
2. `detect_cuda_major()` checks `nvcc --version`, `/usr/local/cuda*/targets/*/lib/libcudart.so.*`,
   and `ldconfig -p | grep libcudart` — all absent on a driver-only host → returns empty.
3. `archive_name()` with empty major falls back to the legacy name
   `mesh-llm-x86_64-unknown-linux-gnu-cuda.tar.gz`, which the v0.75.0 release no longer
   publishes (assets were renamed to `-cuda-12` / `-cuda-13`) → 404.

The aarch64 branch has the same fallback, so driver-only ARM hosts should hit it too
(untested; my DGX Spark hosts carry the toolkit and install fine).

Confirmed fix locally: inserting `[[ -z "$ver" ]] && command -v nvidia-smi >/dev/null && ver=13`
into `detect_cuda_major` unblocks the install. A more principled version might parse the CUDA
version from the `nvidia-smi` header (present whenever a driver is), or default to the newest
published cuda lane whenever `probe_nvidia` succeeds but the toolkit is absent.

## Extra context

After installing this way, `serve` on the same host still fails to load
[...]

PR #1267, opened 2026-08-12
Fix install one-liner 404 on NVIDIA hosts without a CUDA toolkit

michaelneale:
## What this fixes

The documented install one-liner now works on Linux hosts that have an NVIDIA driver but no CUDA toolkit — the normal shape for an inference-only GPU box.

Before, it failed with a 404:

```
$ curl -fsSL https://meshllm.cloud/install.sh | bash -s -- --no-setup
curl: (22) The requested URL returned error: 404
error: could not download release archive for mesh-llm-x86_64-unknown-linux-gnu-cuda.tar.gz or mesh-bundle.tar.gz
```

After, it selects the right lane and installs.

## Why it happened

`detect_cuda_major` probed three things — `nvcc --version`, `/usr/local/cuda*/.../libcudart.so.*`, and `libcudart` in `ldconfig -p`. All three are *toolkit* artifacts, so a driver-only host returned empty. `asset_name` then fell back to the legacy bare name `mesh-llm-<arch>-unknown-linux-gnu-cuda.tar.gz`, which releases stopped publishing when the CUDA lanes were split into `-cuda-12` / `-cuda-13`. The v0.75.1 release publishes only the split names, so that fallback could never resolve — on x86_64 or aarch64.

This is the installer-side counterpart to #1195, which separated driver bound from installed toolkit in host runtime selection. Same underlying confusion (driver ≠ toolkit), different code path: #1195 fixed picking the native runtime after install, this fixes picking the archive to download in the first place. It is not a platform-support limitation — the correct archive exists and is published, the installer just asked for a name that does not.

## What changed

- When toolkit probes find nothing but `nvidia-smi` is present, take the CUDA major from its header (`CUDA Version: 13.0`). That is the driver's maximum supported CUDA, which is the correct upper bound for choosing a lane, and clamp it to the newest published lane.
- Remove the bare-`-cuda` fallback on both the x86_64 and aarch64 branches. If no supported major can be determined, fail with an actionable message instead of fabricating an archive name that cannot exist:

```
error: detected an NVIDIA GPU but could not determine a supported CUDA major version (expected 12 or 13).
       install a CUDA toolkit, or re-run with --flavor cpu (or --flavor vulkan) to use a non-CUDA build.
```

## Validation

`bash -n install.sh` passes. `detect_cuda_major` exercised against a stubbed `nvidia-smi` on a machine with no CUDA toolkit at all, covering every branch of the new path:

| stub `nvidia-smi` output | result |
|---|---|
| `CUDA Version: 13.0` | `13` |
| `CUDA Version: 14.2` (future driver) | `13` (clamped to newest published lane) |
| `CUDA Version: 11.8` (too old) | empty → actionable error |
| no `nvidia-smi` on PATH | empty → actionable error |

Release-asset names confirmed against v0.75.1 with `gh release view`: only `-cuda-12` and `-cuda-13` are published for both `x86_64` and `aarch64`, no bare `-cuda`.

Partially addresses #1220 — the installer half. Deliberately not using a closing keyword: #1220 also reports a runtime-linkage failure that this does not fix, so the issue should stay open until that half is tracked separately.

Not covered here: the reporter also hit a *second*, separate problem — after installing, the Linux cuda13 native runtime links `libcudart.so.13`/`libcublas.so.13` without bundling them, while the Windows runtime ships its own CUDA DLLs. That needs its own issue and fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA version detection by using the installed NVIDIA driver when toolkit information is unavailable.
  * Limited detected CUDA versions to supported releases.
  * Installation now reports an error when no supported CUDA version can be identified instead of selecting an incompatible archive.

## Update after first CI run

CI caught something my local checks did not, and it was a real bug rather than a stale assertion.

`scripts/tests/test_install_sh.py` asserted the legacy bare `-cuda` name that this PR removes, so that assertion is updated to require the refusal instead. But running the suite locally then exposed a genuine defect in my own change: `install.sh` runs under `set -euo pipefail`, and a non-matching `grep` inside `$(...)` aborts the whole script. On a host where the new `nvidia-smi` probe found no version, the installer would have died mid-detection instead of reaching the actionable error. Every `detect_cuda_major` probe pipeline now ends in `|| true`, which also hardens the two pre-existing `nvcc` and `ldconfig` probes that had the same latent exposure.

Test coverage added:

- `test_detect_cuda_major_falls_back_to_nvidia_smi_driver_version` — stubs a driver-only host (no `nvcc`, no `ldconfig`) across `13.0` → `13`, `12.4` → `12`, `14.2` → `13` (clamped), `11.8` → empty.
- `test_asset_name_uses_detected_cuda_major_lane` — the `-cuda-12`/`-cuda-13` lane name is built correctly on both `x86_64` and `aarch64`.
- `test_release_target_helpers_keep_linux_aarch64_flavor_surface` — now asserts refusal plus the actionable stderr, and asserts the legacy name is never emitted.

`python3 -m unittest scripts.tests.test_install_sh` → 26 passed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -336,11 +336,12 @@ recommended_flavor() {
 detect_cuda_major() {
     local ver=""
     if command -v nvcc >/dev/null 2>&1; then
-        ver="$(nvcc --version 2>/dev/null | grep -oE 'release [0-9]+' | awk '{print $2}' | head -n 1)"
+        ver="$(nvcc --version 2>/dev/null | grep -oE 'release [0-9]+' | awk '{print $2}' | head -n 1 || true)"
     fi
     if [[ -z "$ver" ]]; then
         local lib
-        for lib in /usr/local/cuda*/targets/*/lib/libcudart.so.* /usr/local/cuda*/targets/*/lib/stubs/libcudart.so.*; do
+        local probe_root="${MESH_LLM_TEST_CUDA_PROBE_ROOT:-}"
+        for lib in "$probe_root"/usr/local/cuda*/targets/*/lib/libcudart.so.* "$probe_root"/usr/local/cuda*/targets/*/lib/stubs/libcudart.so.*; do
             if [[ -f "$lib" ]]; then
                 ver="$(basename "$lib" | grep -oE 'libcudart\.so\.[0-9]+' | awk -F. '{print $3}' | head -n 1)"
                 break
@@ -348,7 +349,17 @@ detect_cuda_major() {
         done
     fi
     if [[ -z "$ver" ]]; then
-        ver="$(ldconfig -p 2>/dev/null | grep -oE 'libcudart\.so\.[0-9]+' | awk -F. '{print $3}' | sort -rn | head -n 1)"
+        ver="$(ldconfig -p 2>/dev/null | grep -oE 'libcudart\.so\.[0-9]+' | awk -F. '{print $3}' | sort -rn | head -n 1 || true)"
+    fi
+    # Inference-only hosts carry an NVIDIA driver but no CUDA toolkit, so none of
+    # the probes above find anything. The driver still advertises the highest CUDA
+    # it supports in the nvidia-smi header ("CUDA Version: 13.0"); use that as an
+    # upper bound and clamp it to a CUDA lane we actually publish.
+    if [[ -z "$ver" ]] && command -v nvidia-smi >/dev/null 2>&1; then
+        ver="$(nvidia-smi 2>/dev/null | grep -oE 'CUDA Version: *[0-9]+' | grep -oE '[0-9]+' | head -n 1 || true)"
+        if [[ -n "$ver" ]] && (( ver > 13 )); then
+            ver=13
+        fi
     fi
     case "$ver" in
         12|13) printf '%s\n' "$ver" ;;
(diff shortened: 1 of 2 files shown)


michaelneale:
Thanks — two of these were right and are now fixed; one I am deliberately not addressing here.

**Fixed: aarch64 remediation suggested an unavailable flavor.** Correct, `supported_flavors()` returns only `cuda cpu` on aarch64. That branch now says `--flavor cpu` only, and the test asserts `--flavor vulkan` never appears in aarch64 stderr.

**Fixed: tests were not isolated from host toolkit state.** Also correct — the `/usr/local/cuda*` glob ran before the PATH wrappers, so a runner with a toolkit installed could bypass the setup and pass for the wrong reason. `detect_cuda_major` now honours an optional `MESH_LLM_TEST_CUDA_PROBE_ROOT` prefix on that glob, and the probe tests set it to an empty temp dir. The aarch64 surface test overrides `detect_cuda_major` outright, since it is exercising `asset_name`, not detection.

**Not fixed here, deliberately: using `nvidia-smi` as CUDA runtime evidence.** You are right that the driver header is only an upper bound and does not prove `libcudart`/`libcublas`/`libcublasLt` are present at a matching major. But requiring that evidence in the installer would leave a driver-only host with *no* selectable CUDA archive, which is a strictly worse outcome than today and does not match what the release actually publishes — there is no self-contained Linux CUDA bundle to select.

The real defect you are pointing at is that Linux CUDA runtimes are host-linked while the Windows ones bundle their CUDA DLLs. That is a genuine bug, it is the second half of #1220, and it needs a fix in packaging rather than in archive-name selection. This PR is scoped to the 404: the fallback it replaces could never resolve at all. After this change a driver-only host gets the correct archive and then fails at runtime load with a clear error — worse-shaped than a bundled runtime would be, better than a 404, and the packaging fix is tracked separately.

**Third comment (release-asset contract / `MESH_CUDA_VERSION`)**: the versioned `-cuda-12`/`-cuda-13` lanes are already what releases publish — verified against v0.75.1 with `gh release view`. This PR does not change the release contract, it stops the installer asking for a name that was retired when the lanes split. Any gap in `check_package_release_assets` predates this change.

PR #1269, opened 2026-08-13
fix: name a local GGUF with --model instead of trying to resolve it

### LINKED ISSUE #1254: [🐛 Bug]: Custom --model string fails eager resolution when supplying local --gguf path
### Environment
- **MeshLLM Version**: `v0.75.1`
- **OS / Architecture**: Ubuntu 24.04 LTS (x86_64 & ARM64 Grace Blackwell GB10)
- **Deployment**: Systemd user service & background `tmux` sessions

### Problem Description
When starting `mesh-llm serve` with both `--gguf <LOCAL_PATH>` and a custom human-readable `--model <ALIAS_NAME>` (e.g., `--model deepseek-v4-flash`), the daemon fails eager startup resolution. The engine attempts to search for `<ALIAS_NAME>` on the online Hugging Face hub or local catalog index instead of assigning `<ALIAS_NAME>` as the catalog alias for the provided GGUF file path.

### Steps to Reproduce
1. Download a local GGUF model to disk (e.g., `/models/UD-IQ3_S/...`).
2. Execute `mesh-llm serve` with both flags:
   ```bash
   mesh-llm serve --mesh-name MainMesh --model deepseek-v4-flash --gguf /models/UD-IQ3_S/DeepSeek-V4-Flash-UD-IQ3_S-00001-of-00004.gguf
   ```
3. Inspect `journalctl` / service log output:
   ```text
   ERR Model not found: deepseek-v4-flash (Not a local file, not in Hugging Face cache, not in catalog). 
   Use a path, a catalog name, or a Hugging Face exact ref/URL: 
   Eager startup model 'deepseek-v4-flash' failed to resolve after daemon surfaces became ready.
   ```

### Expected Behavior
When `--gguf` specifies a valid local file path, `--model <NAME>` should automatically register `<NAME>` as the friendly catalog alias for the underlying GGUF hash (e.g., mapping `deepseek-v4-flash` -> `local-gguf/sha256-8213146687bc8a95`), allowing API clients to query `http://host:3131/v1/chat/completions` using `"model": "deepseek-v4-flash"`.

### Current Workaround
Omitting `--model` completely bypasses the online lookup error and loads the local GGUF file directly into memory using its SHA256 hash (`local-gguf/sha256-8213146687bc8a95`).

### Proposed Solution
Modify flag parsing logic during daemon initialization: if `--gguf` is provided and points to an existing file on disk, skip Hugging Face hub resolution for `--model` and assign `--model`'s value as an in-memory alias for the registered GGUF model. Or provide an in-kind argument to fill the need for a easier to use name for the model in-memory, api endpoints, and in web ui

michaelneale:
You can now give a local GGUF a friendly name:

```bash
mesh-llm serve --mesh-name MainMesh \
  --model deepseek-v4-flash \
  --gguf /models/UD-IQ3_S/DeepSeek-V4-Flash-UD-IQ3_S-00001-of-00004.gguf
```

`deepseek-v4-flash` becomes the served model name — in `/v1/models`, in the console, and in `"model": "deepseek-v4-flash"` request bodies — rather than being looked up on Hugging Face and failing startup:

```text
ERR Model not found: deepseek-v4-flash (Not a local file, not in Hugging Face cache, not in catalog).
Eager startup model 'deepseek-v4-flash' failed to resolve after daemon surfaces became ready.
```

Fixes #1254. Reported by @hola-there, who also identified the correct behaviour and the `--model`-less workaround.

## Scope

The alias binding only applies to the unambiguous naming form: exactly one `--gguf` and exactly one `--model` whose value is a plain name — no `/`, `\`, `:`, or `@`, and not an existing path. Anything that could resolve on its own (`unsloth/Qwen3-8B-GGUF:Q4_K_M`, a second local path, a URL) still starts as its own model, so multi-model serve is unchanged.

## Validation

- `cargo test -p mesh-llm-host-runtime --lib` — 2454 passed, 0 failed

Three new tests in `runtime/tests/startup_models.rs` cover the alias binding, the HF-ref case that must stay two models, and end-to-end resolution proving the alias no longer reaches the catalog.

## Summary by CodeRabbit

* **New Features**
  * Added support for starting local GGUF models with a custom model alias.
  * Local model aliases are preserved in the startup configuration and do not require external catalog lookup.
  * Existing projector and context-size settings remain intact when loading local models.

* **Bug Fixes**
  * Improved validation for GGUF file paths and model identifiers.
  * Prevented valid Hugging Face model references from being incorrectly replaced during startup.


ndizazzo (approved):

hola-there:
Thanks!

PR #1279, opened 2026-08-13
fix: move model lifecycle controls off peer-reachable ingress

### LINKED ISSUE #1190: Move model load/drop controls off the peer-reachable OpenAI ingress
🤖 Model lifecycle administration should not share the peer-reachable OpenAI ingress.

## Current behavior

`POST /mesh/load` and `POST /mesh/drop` are handled by the model-aware OpenAI proxy on `:9337` (`network/openai/ingress.rs`). Inbound `STREAM_TUNNEL_HTTP` traffic is relayed directly to `127.0.0.1:<api_port>`, so an admitted peer can reach those lifecycle handlers through the same channel intended for inference.

This is separate from split serving: topology-scoped Skippy stage control and transport should remain available for cooperative inference.

## Desired behavior

Make model lifecycle administration local/owner-controlled and out of band from mesh inference:

- Keep inference/model discovery on `:9337` and peer HTTP tunnels.
- Use the existing loopback management routes as the canonical local lifecycle API:
  - `POST :3131/api/runtime/models`
  - `DELETE :3131/api/runtime/models/{model}`
- Keep authenticated owner-control lifecycle commands for remote administration.
- Do not introduce an `inference_only` peer mode or disable split-stage protocols as part of this fix.

## Proposed steps

1. Inventory and migrate any internal callers of `/mesh/load` and `/mesh/drop` to the management API or direct runtime-control API.
2. Remove the lifecycle handlers and predicates from the `:9337` OpenAI ingress.
3. Return a clear `404`, `403`, or compatibility `410 Gone` for those legacy paths on `:9337`; never route them as inference.
4. Retain the existing trusted-local checks for `POST/DELETE /api/runtime/models...` on the management API.
5. Add regressions proving:
   - local `:3131` lifecycle requests still work;
   - local `:9337/mesh/load` and `/mesh/drop` no longer mutate runtime state;
   - tunneled QUIC HTTP requests to those paths are rejected and do not mutate runtime state;
   - normal model discovery/inference and dynamic split formation still work.
6. Update docs/SDK examples and document any compatibility window for external users of the legacy routes.

## Compatibility

This changes HTTP behavior but not the mesh wire protocol. Older peers attempting the legacy lifecycle paths should receive an explicit failure. Mixed-version inference and split serving should remain unaffected.

ndizazzo:
## Original problem

`POST /mesh/load` and `POST /mesh/drop` were handled by the model-aware OpenAI ingress on `:9337`. Because inbound `STREAM_TUNNEL_HTTP` traffic is relayed to the loopback API port, an admitted peer could reach model lifecycle controls through the same channel intended for inference and discovery.

## Diagnostics

The request path was traced through the OpenAI ingress, the runtime proxy, and the inbound HTTP tunnel relay. Regression tests cover both direct ingress requests and the passive/bootstrap path used for tunneled peer HTTP requests.

## Fix

- Removed model load/drop runtime-control handlers and control-channel plumbing from the peer-reachable OpenAI ingress.
- Recognize the legacy `/mesh/load` and `/mesh/drop` paths before any inference or model routing, including query-string variants.
- Return HTTP `410 Gone` with the `legacy_route_gone` error code for those legacy paths. They do not reach an inference upstream and cannot mutate runtime state.
- Kept inference, model discovery, normal peer HTTP tunneling, and Skippy stage control/transport behavior unchanged.
- Retained the trusted local lifecycle API as the canonical path:
  - `POST :3131/api/runtime/models`
  - `DELETE :3131/api/runtime/models/{model}`
- Retained authenticated owner-control lifecycle commands for remote administration.
- Added tests proving the canonical management routes still load/unload models and the legacy ingress paths are rejected without routing or mutation.
- Updated hosted-mesh and runtime lifecycle documentation, including the compatibility behavior for older callers.

This is an HTTP-surface compatibility change only; the mesh wire protocol is unchanged, and mixed-version inference/split serving remains supported.

## Validation

- `just with-lld cargo test -p mesh-llm-host-runtime -- --test-threads=1` — 2,455 passed, 8 ignored.
- `just with-lld cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings` — passed.
- `just build` — passed.
- `just website-build` — passed.
- `git diff --check` — passed.
- No UI changes; screenshots are not applicable.

Closes #1190

## Summary by CodeRabbit

* **Breaking Changes**
  * Legacy `/mesh/load` and `/mesh/drop` endpoints now return `410 Gone` without changing runtime state, including through inbound tunnels.
  * OpenAI interfaces remain available for inference and model discovery.

* **New Features**
  * Local lifecycle operations use the management API.
  * Remote administration uses authenticated owner-control endpoints.

* **Documentation**
  * Updated lifecycle and hosted mesh serving guidance.

* **Tests**
  * Added coverage confirming rejected lifecycle requests are not forwarded.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/api/tests/runtime_lifecycle_routes.rs
+++ b/crates/mesh-llm-host-runtime/src/api/tests/runtime_lifecycle_routes.rs
@@ -40,6 +40,77 @@ async fn runtime_status_and_activity_routes_are_dispatched() {
 }
 
 #[tokio::test]
+async fn local_runtime_model_routes_remain_the_lifecycle_api() {
+    let state = build_test_mesh_api().await;
+    let (control_tx, mut control_rx) = mpsc::unbounded_channel();
+    state.set_runtime_control(control_tx).await;
+
+    let load_task = tokio::spawn(async move {
+        match control_rx.recv().await {
+            Some(RuntimeControlRequest::Load {
+                spec,
+                profile,
+                resp,
+            }) => {
+                assert_eq!(spec, "org/model:Q4_K_M");
+                assert_eq!(profile, "low-ctx");
+                let _ = resp.send(Ok(RuntimeLoadResponse {
+                    model_ref: spec.clone(),
+                    model: spec,
+                    instance_id: "runtime-load-1".to_string(),
+                    profile,
+                    backend: None,
+                    context_length: None,
+                }));
+            }
+            _ => panic!("expected local load request"),
+        }
+    });
+
+    let loaded = send_runtime_lifecycle_request(
+        state.clone(),
+        lifecycle_post(
+            "/api/runtime/models",
+            r#"{"model":"org/model:Q4_K_M#low-ctx"}"#,
+        ),
+    )
+    .await;
+    load_task.await.unwrap();
+    assert!(loaded.starts_with("HTTP/1.1 201 Created"), "response: {loaded}");
+    assert_eq!(json_body(&loaded)["loaded"], "org/model:Q4_K_M");
+
+    let (control_tx, mut control_rx) = mpsc::unbounded_channel();
+    state.set_runtime_control(control_tx).await;
+    let unload_task = tokio::spawn(async move {
+        match control_rx.recv().await {
+            Some(RuntimeControlRequest::Unload {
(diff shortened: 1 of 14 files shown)


michaelneale (approved):
thanks, yes we need this.

PR #1280, opened 2026-08-13
Compact oversized chat context for provider windows

i386:
## Summary

Apple system models expose a 4096-token context window. The console can send a persisted conversation plus its system prompt that exceeds that window, causing the router to reject the request before it reaches the Apple sidecar with:

`no context-compatible target for model 'apple/system' can fit approximately ... tokens`

This PR adds one targeted recovery path:

- Detect only the context-specific 503 response.
- Retry once with the newest conversation turns that fit a conservative serialized-request budget.
- Keep the full conversation visible locally; only the retry payload is compacted.
- Leave unrelated 503 responses unchanged.

Follow-up to #1277.

## Validation

- `pnpm exec prettier --check src/features/chat/api/mesh-connection.ts src/features/chat/api/mesh-connection.test.ts`
- `pnpm exec vitest run src/features/chat/api/mesh-connection.test.ts` (14 passed)
- `pnpm exec tsc -b --pretty false`
- `git diff --check`

## Notes

This is intentionally a separate PR from the model-route retry change. It addresses the 503 context-admission failure observed with `apple/system` and applies to any provider reporting the same router error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of oversized conversations by retrying with progressively compacted input when no compatible context is available.
  * Preserves the latest conversation turn while removing older content as needed to meet request size limits.
  * Supports multibyte text within the configured 14,000-byte limit.
  * Reuses uploaded attachments during retries to avoid unnecessary re-uploads.
  * Continues streaming responses when retries succeed and retains existing retry behavior for other temporary connection issues.


crates/mesh-llm-ui/src/features/chat/api/mesh-connection.ts
```diff
@@ -82,13 +82,41 @@ function isAbortError(error: unknown) {
...
+async function compactResponsesInput(
+  messages: Array<UIMessage> | Array<ModelMessage>,
+  model: string,
+  clientId: string,
+  requestId: string,
+  systemPrompt: string
+) {
+  if (messages.length < 2) return undefined
+
+  const firstMessage = Math.max(1, messages.length - 8)
```
ndizazzo, line 110: This walks the history one message at a time, so it can keep an assistant reply after dropping the user prompt that produced it. For example, a normal nine-message alternating history ending in a user message, with about 3 KB per message, first fits at index 5 and sends an assistant-first retry. Could we only consider starts at user-message or whole-turn boundaries, and add an odd-length conversation test?

crates/mesh-llm-ui/src/features/chat/api/mesh-connection.ts
```diff
@@ -82,13 +82,41 @@ function isAbortError(error: unknown) {
...
+  for (let start = firstMessage; start < messages.length; start += 1) {
+    const candidate = await buildResponsesInput(messages.slice(start), model, clientId, requestId, systemPrompt)
```
ndizazzo: buildResponsesInput is not a pure size probe: data-backed audio and document parts are uploaded to /api/objects. Since this runs once per candidate, a long conversation can upload the same attachment up to eight extra times before a suffix is selected, and unsuccessful probes remain until the request-object TTL. Could we choose the suffix with a side-effect-free representation and call buildResponsesInput only once for the selected candidate? This is especially costly because one object can be as large as 50 MB.

crates/mesh-llm-ui/src/features/chat/api/mesh-connection.test.ts
```diff
@@ -500,6 +500,108 @@ describe('createMeshConnectionAdapter', () => {
...
+        .join('')
+    ).toContain('Compacted')
+  })
+
+  it('measures UTF-8 byte length correctly when compacting multibyte text', async () => {
+    // Construct messages with multibyte characters (emoji + CJK) where UTF-16 .length
+    // would undercount the actual UTF-8 byte size, potentially allowing an oversized body.
+    // Each emoji is ~4 UTF-8 bytes but 2 UTF-16 code units (.length counts UTF-16).
+    // Each CJK character is ~3 UTF-8 bytes but 1 UTF-16 code unit.
+    const multibyteContent = '🌟🚀💡'.repeat(1000) + '你好世界'.repeat(500)
```
ndizazzo: This fixture can never reach the mocked second response: multibyteContent is already about 18 KB in UTF-8 before the turn prefix and JSON overhead, so even the one-message suffix is above the 14 KB cap. I reproduced this as 1 failing and 14 passing tests in this file. Could we scale the fixture down so one message fits but two or more do not? That still exercises the UTF-8 versus UTF-16 behavior without making compaction impossible.

ndizazzo (changes requested):
Thanks for keeping this recovery path narrowly scoped. The error matching and UTF-8 measurement are heading in the right direction, but I found three cases that need addressing before this is safe to merge: the new multibyte test currently fails, candidate construction can repeat attachment uploads, and compaction can split a user/assistant turn. I reproduced the focused test failure locally (1 failed, 14 passed); typecheck and the production build both pass.

i386:
@ndizazzo this is ready

crates/mesh-llm-ui/src/features/chat/api/mesh-connection.ts
ndizazzo: Fixed in 0cff72b: `compactResponsesInput` now skips any `start` whose message isn't `role: 'user'`, so the retry always begins on a user turn. Confirmed `compactedBody.input[0].role === 'user'` is asserted in the existing 8-message test and reran the suite locally (19/19 passing).

crates/mesh-llm-ui/src/features/chat/api/mesh-connection.ts
ndizazzo: Fixed in 0cff72b: `buildResponsesInput` now takes an `AttachmentUploadCache` keyed by mime type + filename + base64 value, shared across all compaction probes and the final selected candidate. A failed upload evicts its cache entry so it doesn't poison a later retry. New test `reuses uploaded attachments while compacting a context-rejected conversation` asserts `/api/objects` is only called once. Verified locally.

crates/mesh-llm-ui/src/features/chat/api/mesh-connection.test.ts
ndizazzo: Fixed in 6dd1e01: fixture scaled down ('🌟🚀💡'.repeat(300) + '你好世界'.repeat(200)) so a single retained turn fits the 14 KB budget while the full history still exceeds it, exercising UTF-8 vs UTF-16 measurement without making compaction impossible. Reran locally: 19/19 passing.

ndizazzo (approved):
Re-reviewed after the latest commits (6dd1e01, 0cff72b). All three change requests from my prior review are addressed:

- **Turn-boundary split**: `compactResponsesInput` now only starts a candidate at a `role: 'user'` message, so a retry can no longer detach an assistant reply from the user turn that produced it.
- **Repeated attachment uploads**: `buildResponsesInput` now threads a shared `AttachmentUploadCache` through all compaction probes and the final selected candidate, keyed by mime type + filename + base64 value, with cache eviction on upload failure. A new test asserts `/api/objects` is called exactly once across the probe-and-retry flow.
- **Failing multibyte fixture**: scaled down so one retained turn fits the 14 KB budget while the full history still exceeds it, still exercising UTF-8 vs UTF-16 measurement.

Verified locally on the PR head (0cff72b):
- `pnpm exec vitest run src/features/chat/api/mesh-connection.test.ts` → 19/19 passing (was 1 failing/14 passing)
- `pnpm exec tsc -b --pretty false` → clean
- `pnpm exec prettier --check` on the touched files → clean

Resolved the three addressed threads. No further issues found — approving.

PR #1261, opened 2026-08-12
Add Core AI published-model provider

i386:
## Summary

Adds the experimental Phase 4 Core AI published-model provider on top of the Apple system-provider and private-mesh routing stack.

This is one Swift sidecar for all Apple-native providers. It adds an explicit Core AI artifact lane for published `.aimodel` resources, while keeping `apple/system` available as the separate system-model lane. The provider exposes the same OpenAI-compatible REST, streaming, tools, usage, cancellation, readiness, and load lifecycle used by the existing Apple provider.

Tracks [issue #1246](https://github.com/Mesh-LLM/mesh-llm/issues/1246) and is stacked on [#1260](https://github.com/Mesh-LLM/mesh-llm/pull/1260).

## Why this matters

Apple Silicon users can run a model through Core AI and Foundation Models in the signed macOS sidecar, allowing Metal/ANE-backed inference to stay local to the Mac. Mesh can then route independent requests to those Macs by model identity and live capacity. This avoids Skippy pipeline parallelism for these model classes: the accelerator-native runtime owns the complete request on one Mac, while the mesh provides fan-out, affinity, failover, and withdrawal across Macs.

## What changed

- Adds `CoreAIModelProvider` backed by Apple's `CoreAILanguageModels` package and `CoreAILanguageModel(resourcesAt:)`.
- Adds explicit model IDs under `apple/coreai/<name>` and exact artifact aliases such as `apple/coreai/qwen3-4b@qwen3-4b-2026-08-01`.
- Uses `coreai_model_artifact` as the version source. Core AI artifacts are versioned by the published artifact metadata; there is no backward-compatibility aliasing for Core AI model IDs.
- Loads one Core AI model per sidecar and reuses the existing scheduler and request lifecycle.
- Routes Core AI completion, streaming, structured generation, fixture tool calls, usage, cancellation, and REST model listing through the existing endpoints.
- Allows the host supervisor to select the model declared by the provider runtime manifest instead of hard-coding `apple/system`.
- Packages a published `.aimodel` resource directory in `Models/coreai-model`, writes `Resources/coreai-model.json`, and hashes every copied model file in `provider-runtime.json`.
- Pins Apple's `coreai-models` dependency at revision `f401272` and keeps only the Swift 6.4 mutable-view compatibility preparation needed by the Golden Gate SDK.
- Updates the Apple runtime README and Swift/Rust contract tests.

## Try it (macOS Golden Gate required)

[...]

PR #1281, opened 2026-08-13
fix: emit usage in Apple streaming completions

i386:
## Summary

The Apple runtime's non-streaming completions reported Foundation Models usage, but streaming completions ended with a stop chunk and `[DONE]` without a usage trailer. The console uses the streaming `/api/responses` path, so it received no `output_tokens` and could not render the tok/s counter for `apple/system` or Core AI models.

This stacked follow-up:

- honors `stream_options.include_usage` on the Apple REST carrier;
- emits the measured Apple completion usage on the final streaming chunk;
- keeps the existing OpenAI-compatible stream shape for callers that did not request usage;
- adds a REST QA assertion that streamed usage contains a positive completion-token count.

The console already measures the stream wall-clock interval, so once the usage trailer reaches the Responses adapter it can render tok/s without estimating tokens in the browser.

Stacked on #1261.

## Validation

- `just apple::test` (17 tests passed)
- `bash -n providers/apple/QA/rest.sh`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Streaming chat completions can now request usage statistics.
  * When enabled, usage data is returned in a final usage-only streaming chunk before completion.

* **Bug Fixes**
  * Improved streaming response handling to preserve the final completion data.
  * Added validation to confirm streamed completion-token usage is reported correctly.

### DIFF
--- a/providers/apple/QA/rest.sh
+++ b/providers/apple/QA/rest.sh
@@ -111,7 +111,7 @@ curl --fail --silent --show-error \
 
 curl --fail --silent --show-error --no-buffer \
     -H 'content-type: application/json' \
-    --data-binary "{\"model\":\"$MODEL_ID\",\"messages\":[{\"role\":\"user\",\"content\":\"Reply with exactly: streaming REST ready\"}],\"temperature\":0,\"max_tokens\":32,\"stream\":true}" \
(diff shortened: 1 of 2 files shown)


ndizazzo (approved):
I don't see any dirt!

i386:

Issue #1262, opened 2026-08-12
Qwen3.8 support: dense 27B first, plan splits for 2.4T-A95B


michaelneale:
## Summary

Qwen3.8 has shipped. Map out mesh-llm support, prioritising the **dense 27B** as the desktop-scale starting point, with plan splits for the 2.4T-A95B flagship as follow-on work.

## What actually exists on HF today (2026-08-13, verified via HF API)

| Repo | Notes |
|---|---|
| [`Qwen/Qwen3.8-2.4T-A95B`](https://huggingface.co/Qwen/Qwen3.8-2.4T-A95B) | 2.4T total / 95B active MoE, 92 layers, 512 experts (10 routed + 1 shared), 262144 ctx |
| [`Qwen/Qwen3.8-2.4T-A95B-FP8`](https://huggingface.co/Qwen/Qwen3.8-2.4T-A95B-FP8) | FP8 |
| [`unsloth/Qwen3.8-2.4T-A95B-GGUF`](https://huggingface.co/unsloth/Qwen3.8-2.4T-A95B-GGUF) | BF16 ladder, 140 shards; other quants landing |

**There is no Qwen-published dense `Qwen3.8-27B` yet, and no unsloth 27B GGUF.** Third-party 27B repos exist (`huginnfork/Qwen3.8-27B-FP8`, `neroued/Qwen3.8-27B-NInfer`, `chillaranand/Qwen3.8-27B-mlx`, `barozp/Qwen3.8-27B*-GGUF`), but `barozp/*` are self-declared *placeholders* reserved ahead of a 27B release and contain no weights. So the 27B story is: **watch for the official drop**, then quant.

## Architecture read

`Qwen/Qwen3.8-2.4T-A95B/config.json` declares:

```json
"architectures": ["Qwen3_5MoeForCausalLM"],
"model_type": "qwen3_5_moe_text"
```

Layer layout is 23 × (3 × Gated DeltaNet → MoE, then 1 × Gated Attention → MoE) — i.e. the same hybrid linear/full-attention pattern as Qwen3.5/3.6, `full_attention_interval: 4`.

[...]

PR #1283, opened 2026-08-13
feat(topology): route Qwen3.8 identities to the qwen35 recurrent family

michaelneale:
Work item 1, 2 and 4 from #1262. Identity plumbing only.

## Problem

Qwen3.8 declares `Qwen3_5MoeForCausalLM` / `qwen3_5_moe_text` and uses the same hybrid Gated-DeltaNet/Gated-Attention layout as Qwen3.5/3.6, so it loads as the existing llama.cpp `qwen35` / `qwen35moe` pair. But `is_qwen35_series` matched only `qwen3.5` / `qwen3.6` / `qwen35` / `qwen36`, so a `Qwen3.8-*` identity resolved to `None` — no family capability, no recurrent range, and `infer_cache_payload` fell through to `ResidentKv`, which is wrong for a recurrent family.

## Change

- `family_capability.rs`: add the `qwen3.8` dotted spelling and the `qwen38` dotless marker to `is_qwen35_series`. The existing `is_parameter_size_suffix` guard already handles the `Qwen3-8B` -> `qwen38b` collision the same way it handles `Qwen3-5B`.
- `kv_integration/config.rs`: add `qwen3.8` / `qwen3_8` to the `KvRecurrent` payload list.
- Tests: extend the existing qwen35-series coverage with real Qwen3.8 identities (MoE flagship quants, the `-layers` package form, dense 27B), pin `Qwen/Qwen3-8B-GGUF` to `qwen3_dense`, and add the flagship to the `certified_recurrent_families_never_use_resident_kv_policy` list.
- `FAMILY_STATUS.md`: name Qwen3.8 on the Qwen3.5-recurrent row.

## Scope and what this deliberately does not claim

No Qwen3.8 weights have been loaded. This routes identities to a family policy that is already certified for the architecture; it does not certify Qwen3.8 itself, and the docs row says so explicitly. Work items 3 and 5 from #1262 (real-weight confirmation, flagship split planning) stay open.

Unblocks the downstream `certify` step for the in-flight `meshllm/Qwen3.8-2.4T-A95B-UD-Q1_0-layers` repack; the repack itself is arch-agnostic and does not need this.

## Verification

`cargo test -p skippy-topology` (35 passed), `cargo test -p skippy-server` (395 passed), `cargo test -p mesh-llm-host-runtime` (1933 passed), `cargo fmt --check`, and `cargo clippy --all-targets` on the three packages, all at 43d62f38.

## Summary by CodeRabbit

* **New Features**
  * Added recurrent KV-cache classification for Qwen3.6, Qwen3.8, and supported Qwen3.5 variants and aliases.
  * Improved model-family detection for dotted and dotless identifiers.

* **Bug Fixes**
  * Prevented parameter-size values from being misidentified as model-series names.
  * Unknown Qwen3 point releases now remain unclassified instead of receiving incorrect policies.
  * Improved cache classification and mobility validation across Qwen3 variants.

* **Documentation**
  * Updated Qwen3.6 and Qwen3.8 support status, noting that Qwen3.8 weights remain uninspected.


crates/mesh-llm-host-runtime/src/inference/skippy/family_policy.rs
```diff
@@ -800,13 +800,16 @@ mod tests {
...
+            // Qwen3.6 and Qwen3.8 load as the qwen35/qwen35moe recurrent pair,
+            // so every uploader and quant must select KvRecurrent rather than
+            // ResidentKv.
             "unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL",
             "unsloth/Qwen3.6-35B-A3B-GGUF:Q4_K_M",
             "bartowski/Qwen3.6-35B-A3B-GGUF:Q4_K_M",
             "unsloth/Qwen3.6-27B-GGUF:UD-Q4_K_XL",
             "unsloth/Qwen3.5-35B-A3B-GGUF:Q4_K_M",
+            "unsloth/Qwen3.8-2.4T-A95B-GGUF:UD-Q1_0",
+            "meshllm/Qwen3.8-2.4T-A95B-UD-Q1_0-layers",
```
ndizazzo, line 882: Are we going to be eternally chasing new model releases with this pattern? I wonder if there's a way to auto-detect this.

ndizazzo (approved):
just thinking out loud with the one comment

crates/mesh-llm-host-runtime/src/inference/skippy/family_policy.rs
michaelneale: yeah I wonder too, fair point

micspiral:
@ndizazzo fair hit, and it was hiding a real bug. Pushed faa1b081.

You were right that the pattern doesn't scale, and the substring ladder in `infer_cache_payload` was the worst offender — it was pure literal-matching with no relationship to anything structural. Worse, it was already wrong: it matched only the dotted `qwen3.6` / `qwen3.8` spellings, so a stage that saw the bare architecture string `qwen35moe` (which is what `capability_from_gguf_meta` falls back to when there is no model id) got `ResidentKv` for a recurrent family. That predates this PR.

The auto-detect you were reaching for mostly already exists: `STAGE_RUNTIME_LLAMA_FAMILY_EXPECTATIONS` records `recurrent_or_hybrid` per family, and `family_policy_for_normalized_family_id` already consults it. `infer_cache_payload` just wasn't using it. Now it does — resolve the identity through `infer_family_capability`, look up the expectation, and take `recurrent_or_hybrid`. The substring ladder stays only as a fallback for identities the capability table can't resolve.

Net effect on the original question: a new release that reuses an existing llama.cpp architecture now needs **no** new literal here. That is precisely the Qwen3.8 case — it declares `Qwen3_5MoeForCausalLM`, so once identity resolution knows the series, cache policy follows for free. The `qwen3.8` literals I added in 43d62f38 are deleted again as redundant.

The honest limit: this doesn't auto-detect a genuinely *new* architecture. `is_qwen35_series` still needs the series marker, because "Qwen3.8 is the same arch as Qwen3.5" is a fact about the release, not something derivable from the name. Deriving family from GGUF `general.architecture` alone would close that too, and `capability_from_gguf_meta` already has the hook — but it needs the artifact in hand, so it can't help the plan-time path that only has a model id. Happy to open a follow-up if you think that's worth chasing.

Also picked up CodeRabbit's point in the same commit: regression tests now pin dotted, dotless, MoE, `-layers` package, and bare-architecture forms to `KvRecurrent`, and keep the `Qwen3-8B` -> `qwen38b` parameter-size collision on `ResidentKv`.

Verified at faa1b081: skippy-server 397, skippy-topology 35, mesh-llm-host-runtime 1933, fmt, clippy clean.

michaelneale:
@ndizazzo follow-up on the "eternally chasing releases" question — pushed 4d71fc04, which I think is the more useful half of the answer.

Chasing releases is not actually the worst part. The worst part is that **today we chase them badly**: an unlisted point release does not fail, it resolves to the nearest lexical match. Before this commit:

```
Qwen/Qwen3.9-40B-A3B-GGUF:Q4_K_M -> qwen3moe,    recurrent_ranges=[]
qwen3.7                          -> qwen3_dense, recurrent_ranges=[]
```

Both non-recurrent, both silent. Every Qwen3 point release we have actually seen — 3.5, 3.6, 3.8 — is hybrid and loads as the `qwen35`/`qwen35moe` pair, so the lexical guess is wrong in precisely the case it fires. Forgetting to add a marker didn't mean "unsupported", it meant "quietly served with the wrong cache policy".

So instead of a growing marker list, `qwen3_point_release` parses the release digit once and an unrecognized one returns no capability at all. That surfaces as a gap at onboarding time, which is what `NEW_MODEL_ONBOARDING.md` asks for, rather than as wrong policy at runtime.

Worth flagging that the parser has to reject three parameter-count shapes that look like point releases: multi-digit runs (`Qwen3-235B-A22B`), a digit followed by `b` (`Qwen3-8B`), and fractional sizes (`Qwen3-0.6B` -> `qwen30.6b`). The existing suite caught the latter two when I got them wrong — all three are now pinned explicitly.

What this deliberately does not do: auto-detect a genuinely new architecture. Adding one still needs a real code change, and I'd argue it should. "Qwen3.8 reuses `Qwen3_5MoeForCausalLM`" is a fact about the release, verifiable only against an artifact. The improvement is that the default is now "I don't know" instead of a confident wrong guess.

Verified at 4d71fc04: skippy-topology 36, skippy-server 397, mesh-llm-host-runtime 1933, plus skippy-quantize / model-ref / model-package, fmt and clippy clean.

PR #1290, opened 2026-08-13
Fix split serving: stages above layer 0 fail to load

### LINKED ISSUE #1288: Split stage-1 fails to load layer package: check_tensor_dims requests blk.0 for a stage starting at layer 18
## Summary

A non-zero stage in a two-node Skippy split fails to load its layer package: llama asks for `blk.0.attn_norm.weight`, which a stage starting at layer 18 (or 35) does not contain. The stage never becomes live and the mesh sits with zero live stages indefinitely.

Reproduces on `main` (`1745edba`) with both nodes and both native runtimes built from `main`.

## Environment

- Two Apple Silicon nodes on a LAN, direct QUIC (8-22ms, no relay), default nostr discovery.
- `mesh-llm 0.72.1`, native runtime built locally via `just release-runtime-build` (darwin-aarch64-metal).
- Model: `hf://meshllm/Qwen3-8B-Q4_K_M-layers@3e7a18e5e6e861998189f14bf7f4b45a0d63e07d`, `--ctx-size 40960`.

## Reproduce

Start a coordinator and one worker, both with `--split` and the same pinned topology lock (`stage-0 = 0..18`, `stage-1 = 18..36`):

```bash
mesh-llm serve --model hf://meshllm/Qwen3-8B-Q4_K_M-layers \
  --split --split-topology-lock lock.json \
  --ctx-size 40960 --bind-ip <lan-ip> --bind-port 7842
```

## Observed

Stage-1 receives a correct assignment:

```text
stage_index=1 layers=18..36 ctx=40960 lanes=4 n_batch=512 n_ubatch=128
  n_gpu_layers=-1 load_mode=LayerPackage include_embeddings=true
  include_output=true filter_tensors_on_load=true
```

It then reads all 201 tensors and fails:

```text
Reading tensor groups...
load_tensors: loading model tensors, this can take a while... (load_mode = mmap)
llama_model_load: error loading model: check_tensor_dims: tensor 'blk.0.attn_norm.weight' not found
mesh-llm: skippy_model_open_from_parts returned status=ModelError
```

After the failure the node loops on `Split runtime coordinator is <id>; standing by for stage assignment`.

`GET /api/runtime/stages` shows the topology entry but no live stages:

```json
{"stages": [], "statuses": [],
 "topologies": [{"stages": [{"stage_id": "stage-1", "layer_start": 18, "layer_end": 36, ...}]}]}
```

## Expected

Stage-1 loads only its assigned range and becomes live, or fails with an error naming the actual mismatch.

## Notes

- Not a layer-boundary artifact: reproduces with the automatic plan (`35..36`) and a pinned lock (`18..36`).
- Not a transport problem: the planner reports `excluded []; blockers []`, the path is direct, and the lock is honoured with matching `manifest_sha256` on both nodes.
- `blk.0` is what a *stage-0* load would request, so this looks like the tensor-dimension check running against unfiltered/global layer indices rather than the stage's local range — but I have not confirmed that in the loader and it should not be treated as diagnosed.
- Blocks two-node split validation of #1228 (KV disk tier), though it is independent of that change.

michaelneale:
Split serving works again. A worker assigned a layer range that does not start at layer 0 currently fails to load its stage on `main` and never comes up, so no split can serve inference at all:

```text
llama_model_load: error loading model: check_tensor_dims: tensor 'blk.0.attn_norm.weight' not found
mesh-llm: skippy_model_open_from_parts returned status=ModelError
```

Fixes #1288.

## What broke

Correction to an earlier version of this description: **#1216 did not introduce the broken ordering.** Its pinned llama.cpp still called `buft_for_tensor` before `check_tensor_dims`, so the skippy filter ran early enough. The regression came from #1232 advancing the pin past upstream llama.cpp [`1269cb1`](https://github.com/ggml-org/llama.cpp/commit/1269cb1ff1598751f846241be90083ae9ad036fb) ("allow reshape of tensors during load"), which moved `check_tensor_dims` ahead of `buft_for_tensor`, without reconciling the skippy filter ordering that #1216 had placed inside the `buft_for_tensor` lambda.

On the file-backed load path that lambda is now not called until *after*:

```cpp
const struct ggml_tensor * cur = check_tensor_dims(tn.str(), ne, !(flags & TENSOR_NOT_REQUIRED), ...);
```

So a stage assigned layers 18..36 still walks blocks from 0, `blk.0.attn_norm.weight` is looked up before the filter can skip it, `get_tensor_meta` returns NULL because a filtered stage package legitimately has no block 0, and it throws.

The fix extracts the filter decision into a `stage_filter_excludes` helper and consults it ahead of `check_tensor_dims`, preserving the `size_data` / `n_created` / `g_skippy_last_tensor_filtered` bookkeeping from #1216. `buft_for_tensor` calls the same helper, so behavior on the virtual path is unchanged.

## Why tests didn't catch it

The `files.empty()` virtual path calls `buft_for_tensor` before any tensor lookup, so it kept working — every build and every existing test passed while real split serving was dead.

Nothing in the suite ever opened a genuine mid-stage artifact. `skippy-runtime`'s CI tests load a full Qwen3-0.6B GGUF at `layer_start: 0`, and `skippy-correctness`'s filtered-stage coverage is `#[ignore]`d behind model downloads. A filtered *config* over an unfiltered GGUF does not reproduce this, because a full GGUF still contains block 0 — the artifact has to actually be missing those tensors.

This PR adds that missing shape: a test that writes a real mid-stage artifact via `write_stage_artifact` and opens it through `StageModel`. It runs in CI wherever `SKIPPY_CORRECTNESS_MODEL` is already set.

## Validation

Patch-queue change only (patch 0001) — no Rust source changes outside the new test.

- Fresh `scripts/prepare-llama.sh pinned` replay of the full 15-patch queue applies cleanly and produces the intended tree.
- `scripts/build-llama.sh` clean.
- **With the fix:** `cargo test -p skippy-model-package` → 53 passed, 0 failed.
- **Without the fix** (same tree, filter hoist reverted, native rebuilt): the new test fails with the exact production error, `check_tensor_dims: tensor 'blk.0.attn_norm.weight' not found`.
- `cargo fmt --all --check` and `cargo clippy -p skippy-model-package --all-targets -- -D warnings` clean.

Two-machine split validation on real hardware is next; I'll post the result here. Filing now because `main` is broken.

## Summary by CodeRabbit

- **New Features**
  - Added support for loading and running staged, multi-part model artifacts.
  - Added stage-aware execution with filtered tensor loading, activation support, and stage-boundary outputs.
  - Expanded support for Inkling models and improved compatibility with DeepSeek-based models.
  - Improved KV-cache management, including compaction, persistence, staged transfer, and indexer-key support.

- **Tests**
  - Added regression coverage for loading mid-stage artifacts with runtime tensor filtering enabled.

### DIFF
--- a/.agents/skills/llama-stage-patch-changes/SKILL.md
+++ b/.agents/skills/llama-stage-patch-changes/SKILL.md
@@ -163,6 +163,28 @@ LLAMA_WORKDIR="$tmp_root/llama.cpp" \
   scripts/build-llama.sh
 ```
 
+### Re-pinning upstream
+
+Advancing `third_party/llama.cpp/upstream.txt` can silently invalidate a patch
+that depends on upstream's *ordering*, not just its symbols. The queue still
+applies, everything compiles, and the behavior is broken. This happened with
+upstream `1269cb1`, which moved `check_tensor_dims` ahead of `buft_for_tensor`
+and left the stage tensor filter running too late; split serving was broken on
+main because no test opened a real mid-stage artifact.
+
+So on every re-pin, in addition to the checks above:
+
+- Read `git log <old-pin>..<new-pin> -- src/llama-model-loader.* src/llama-model.*`
+  for changes to load order, not just to signatures the patches touch.
+- Prove the staged load path with a real artifact whose first block is not
+  block 0. `cargo test -p skippy-model-package` covers this via
+  `mid_stage_artifact_opens_with_the_stage_filter_applied`.
+- Confirm that test actually ran rather than skipped. It is gated on
+  `SKIPPY_CORRECTNESS_MODEL`; without it the test prints
+  `skipping mid-stage: SKIPPY_CORRECTNESS_MODEL is not set` and passes. Grep the
+  CI log for `mid_stage_artifact_opens_with_the_stage_filter_applied ... ok`, or
+  set the variable locally. A skipped gate reads identically to a pass.
+
 Compile each new public header once as C11 and once as C++17 with warnings
 treated as errors. For implementation moves, run the tests owned by the moved
 capability in addition to the Rust fallout checks below.
--- a/.github/workflows/ci-rust-tests-slice.yml
+++ b/.github/workflows/ci-rust-tests-slice.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Prepare UI placeholder
         run: mkdir -p crates/mesh-llm-ui/dist && printf '<html></html>' > crates/mesh-llm-ui/dist/index.html
       - name: Prepare patched llama.cpp when required
-        if: ${{ contains(toJson(matrix.batch.crates), 'skippy-quantize') || contains(toJson(matrix.batch.crates), 'skippy-runtime') || contains(toJson(matrix.batch.crates), 'skippy-ffi') }}
+        if: ${{ contains(toJson(matrix.batch.crates), 'skippy-quantize') || contains(toJson(matrix.batch.crates), 'skippy-runtime') || contains(toJson(matrix.batch.crates), 'skippy-model-package') || contains(toJson(matrix.batch.crates), 'skippy-ffi') }}
         run: scripts/prepare-llama.sh pinned
       - name: Download immutable static ABI input
         uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
@@ -96,7 +96,7 @@ jobs:
       - name: Restore immutable static ABI input
         run: scripts/restore-static-abi-input.sh "$RUNNER_TEMP/static-abi-input" "$LLAMA_STAGE_BUILD_DIR" x86_64-unknown-linux-gnu cpu
       - name: Download Skippy correctness model when required
-        if: ${{ contains(toJson(matrix.batch.crates), 'skippy-runtime') }}
+        if: ${{ contains(toJson(matrix.batch.crates), 'skippy-runtime') || contains(toJson(matrix.batch.crates), 'skippy-model-package') }}
         run: |
           set -euo pipefail
           mkdir -p "$RUNNER_TEMP/skippy-correctness-model"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7452,6 +7452,7 @@ dependencies = [
  "skippy-ffi",
  "skippy-protocol",
  "skippy-runtime",
+ "tempfile",
  "tokio",
 ]
 
--- a/crates/skippy-model-package/Cargo.toml
+++ b/crates/skippy-model-package/Cargo.toml
@@ -18,3 +18,6 @@ serde.workspace = true
 serde_json.workspace = true
 sha2 = "0.10"
 tokio = { version = "1", features = ["rt"] }
+
+[dev-dependencies]
+tempfile = "3"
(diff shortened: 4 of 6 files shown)


i386:
Review findings:

The functional fix looks correct. I found one minor test-hygiene issue, but no blocking code defect.

### Minor: temporary artifacts can leak

The regression test returns early if `write_stage_artifact` fails, and cleanup is attempted while the successful `StageModel` may still hold the file open. Please use an RAII temporary directory and drop the model before cleanup, particularly for Windows compatibility.

### Root-cause attribution correction

The PR description slightly misattributes the regression: **#1216 did not itself introduce the broken ordering.** Its pinned llama.cpp still called `buft_for_tensor` before `check_tensor_dims`, so the Skippy filter ran early enough.

The actual sequence was:

1. Upstream llama.cpp commit [`1269cb1`](https://github.com/ggml-org/llama.cpp/commit/1269cb1ff1598751f846241be90083ae9ad036fb), “allow reshape of tensors during load,” moved `check_tensor_dims` ahead of `buft_for_tensor`.
2. Mesh PR #1232 advanced the pin past that commit while retaining the Skippy filter inside `buft_for_tensor`.
3. A mid-stage package legitimately lacks lower-layer tensors such as `blk.0.*`, so the newly earlier dimension check threw before the filter could exclude them.
4. Existing virtual/full-GGUF tests still had those tensors available, allowing CI to remain green.

#1290 correctly extracts the filter decision and calls it before `check_tensor_dims`, while retaining the buffer-selection bookkeeping.

Validation evidence: the new real mid-stage test actually executed in CI and the `skippy-model-package` batch passed 53/53; the two-node split smoke also passed.

Recommendation: merge after the temporary-directory cleanup and update the root-cause wording from “#1216 moved the filter” to “#1232 imported upstream `1269cb1` without reconciling the Skippy filter ordering.”

i386:
Post-merge re-review of the final head (`86aac66f2`):

The functional fix looks correct. The stage filter now runs before `check_tensor_dims()`, so physically pruned tensors are classified as intentionally excluded before the loader treats them as missing, while the existing tensor bookkeeping remains in the shared helper. The regression test now uses RAII temp storage and drops the native model before cleanup, addressing my earlier minor concern.

I also verified the final CI evidence rather than relying on the green summary: `tests::mid_stage_artifact_opens_with_the_stage_filter_applied` actually ran against the downloaded Qwen GGUF, and the crate completed **53/53 tests passing**. Static/native runtime jobs, core inference, and the two-node split smoke also passed. I do not see a blocker in the hotfix itself.

One important follow-up remains in CI selection. Simulating a future PR whose only changed file is `third_party/llama.cpp/patches/0001-mesh-llm-add-skippy-stage-apis.patch` on the merged revision produces:

```json
{
  "domains": ["native-abi"],
  "affected_crates": [],
  "required_slices": ["quality", "ui-artifact", "static-abi", "rust-tests", "runtime-product", "product-smoke"],
  "rust_tests": [],
  "smoke": ["core"]
}
```

So the new `skippy-model-package` regression test ran in this PR because this PR also changed that crate, but a patch-only upstream re-pin—the exact class of change that caused this regression—would select the Rust-test slice with an empty matrix and silently not run it. Please add an explicit native-ABI → `skippy-model-package` (and ideally relevant Skippy correctness/runtime crates) test mapping plus a planner fixture asserting the matrix is non-empty for patch-only changes.

Minor docs nit already noted inline: the skill says to look for `skipping mid-stage: ...`, but the test emits `skipping mid-stage load: ...`.

Given the functional evidence, I would have approved the hotfix; the CI-routing hole is a follow-up hardening item. The PR was already merged before this re-review, so GitHub no longer permits an approval or merge action.

PR #1305, opened 2026-08-14
Prevent stale native runtime reuse across branch builds

michaelneale:
🤖 Branch and release builds that share a MeshLLM version and runtime lane no longer reuse stale native-runtime bytes from the cache.

Native runtime manifests now carry an optional content-derived `runtime.build_id`. The resolver requires a matching build ID whenever the selected artifact supplies one, while preserving legacy matching for older manifests without build identity.

## Compatibility

- Stable `runtime.id` values are unchanged.
- Runtime directory names inside archives are unchanged.
- Release archive filenames and download URLs are unchanged.
- Existing manifests without `build_id` continue to deserialize and resolve under legacy behavior.
- The cache layout remains `<version>/<runtime.id>`; a mismatched selected build bypasses stale installed bytes and resolves the explicit bundle/download, which then replaces the lane normally.

## Build identity

`build_id` is `sha256:<64 lowercase hex>` over canonical runtime metadata, backend architecture requirements, sorted library/tool checksums, and native build provenance. Mutable distribution fields (`url`, archive checksum, signature) are excluded so repackaging identical runtime contents does not change identity.

## Validation

At `9ab9ca30`:

- `cargo fmt --all --check`
- `cargo test -p mesh-llm-native-runtime --lib` — 39 passed
- `cargo check -p mesh-llm`
- `python3 -m unittest discover -s scripts/tests -p 'test_package_native_runtime.py'` — 9 passed
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Runtime packages now include deterministic build identifiers based on their contents and build metadata.
  - Build identifiers are validated to ensure they use the expected SHA-256 format.

- **Bug Fixes**
  - Cached runtimes are reused only when their build identifier matches the selected runtime, helping prevent stale or incompatible bundles.
  - Older runtime manifests without build identifiers remain compatible.

- **Tests**
  - Added coverage for deterministic identifiers, content changes, matching behavior, missing identifiers, and invalid values.

## Cache behavior

This change **detects and refuses stale build reuse; it does not retain multiple builds for one lane**. The cache remains one directory per `<version>/<runtime.id>`, and installing another build replaces that lane as before. A selected local artifact with no bundle/download source and mismatched cached bytes resolves as missing rather than silently loading the wrong build. Multi-build retention would require a separate cache-layout and prune-semantics change.

### DIFF
--- a/crates/mesh-llm-native-runtime/src/manifest.rs
+++ b/crates/mesh-llm-native-runtime/src/manifest.rs
@@ -23,6 +23,8 @@ pub struct NativeRuntimePlatform {
 pub struct NativeRuntimeArtifact {
     pub id: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub build_id: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub mesh_version: Option<String>,
     pub skippy_abi: String,
     pub platform: NativeRuntimePlatform,
@@ -163,6 +165,14 @@ fn validate_artifact(artifact: &NativeRuntimeArtifact) -> Result<()> {
     if artifact.id.trim().is_empty() {
         bail!("native runtime artifact id is empty");
     }
+    if let Some(build_id) = &artifact.build_id {
+        validate_build_id(build_id).with_context(|| {
+            format!(
+                "native runtime artifact {} has invalid build_id",
+                artifact.id
+            )
+        })?;
+    }
     if artifact.skippy_abi.trim().is_empty() {
         bail!(
             "native runtime artifact {} skippy_abi is empty",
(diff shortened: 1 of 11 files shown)


crates/mesh-llm-native-runtime/src/manifest.rs
```diff
@@ -292,6 +316,7 @@ mod tests {
             r#"{
   "runtime": {
     "id": "meshllm-runtime-linux-x86_64-cuda12",
+    "build_id": "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```
michaelneale, line 319: ???? what is this?

michaelneale:
🤖 Closing to reduce parallel PR noise while Qwen3.8 integration work is consolidated on `lab/qwen38-next`. The build-identity branch and validated work are retained; this can be reopened or reintroduced only when end-to-end deployment evidence shows it is required.

Issue #1298, opened 2026-08-13
🤖 Filtered stage load SIGSEGVs with exactly 8 visible CUDA devices (1-7 and full unfiltered 8 all work)


michaelneale:
## Summary

Loading a published multi-layer package stage with `filter_tensors_on_load=true` reliably SIGSEGVs when **exactly eight** CUDA devices are visible. One through seven visible devices load and serve normally, and a full unfiltered model load across all eight also succeeds.

Found during a two-node Vast split rehearsal (Estonia, 8x + 2x A100-PCIE-40GB).

## Reproduction

Binary composed from `main` (incl. #1290) + #1283 head, commit `8335d853140f183bd8ade84149ca3704c129d0bc`. Package pinned to `meshllm/Qwen3-8B-Q4_K_M-layers@3e7a18e5e6e861998189f14bf7f4b45a0d63e07d`.

```
mesh-llm serve \
  --model meshllm/Qwen3-8B-Q4_K_M-layers@3e7a18e5e6e861998189f14bf7f4b45a0d63e07d \
  --split --max-vram 8 --join "$TOKEN" --bind-port 41843 --port 9447 --headless --log-format json
```

Stage config on the faulting node: `stage_index=0 layers=0..31 filter_tensors_on_load=true ctx=40960 mmap=auto load_mode=LayerPackage backend=auto`.

Only `CUDA_VISIBLE_DEVICES` was varied between runs on the same host (machine 54745, 8x A100-PCIE-40GB):

| visible devices | result |
|---|---|
| `0` | ready |
| `0,1` | ready (`model_ready` on port 41853) |
| `0,1,2,3` | ready |
| `0,1,2,3,4,5` | ready |
| `0,1,2,3,4,5,6` | ready |
| `1,2,3,4,5,6,7` | ready (excludes GPU 0 — not a single bad card) |
| all 8 | **SIGSEGV**, exit 139, 2/2 reruns |

Dropping any single device makes it load. The crash is silent: no `GGML_ASSERT`, no `CUDA error`, 0 MiB VRAM in use, dies right after `Opening native model 100%`.

[...]

PR #1308, opened 2026-08-14
Fix retained tensor accounting in staged model loads

michaelneale:
🤖 Filtered staged loads now derive progress/completion bytes from the tensors the loader will actually upload. This prevents tied embedding/output requests from subtracting retained physical bytes, prematurely unmapping later tensor data and crashing CUDA uploads.

## What changed

- Removed per-logical-request `size_data` decrement/increment bookkeeping from stage filtering.
- After logical tensors are created, recompute `size_data` by traversing every created buffer context exactly as `load_all_data` does.
- Include only tensors resolving through `weights_map`: this mirrors `load_all_data`'s traversal, including its `get_weight` guard, so `size_data` equals what the loader will count. The guard is defensive — no known serving path creates a context tensor without a `weights_map` entry.
- Count a tied physical weight twice when materialized in two contexts because it is uploaded twice; count same-context reuse once.
- No public Skippy ABI change.

## Evidence

Issue #1298 instrumentation showed a tied `token_embd.weight` retained as input and filtered as logical output. Global `size_data` was understated by exactly 165,306,368 bytes. Completion fired after the embedding context, preserving only the embedding mmap interval and unmapping the layer-weight tail; the next CUDA upload read `blk.0.attn_norm.weight` from that unmapped tail.

## Validation

At `a8d44708`:

- full 18-patch replay from pinned llama.cpp `030ebb558...` succeeds;
- `just build` succeeds on Apple Silicon, including patched native runtime build and Metal product packaging;
- `git diff --check` passes.

Quantization pre-model `init_mappings` preserves the historical weights-map total; model-load contexts use traversal accounting. The two branches are **deliberately not interchangeable** and `size_data` now means different things by call site: the `ctx_map` traversal answers "what will this loader upload", the `weights_map` fallback answers "what physical weights exist". Quantization needs the latter because it calls `init_mappings` before model construction, so contexts are absent while `weights_map` is populated; it calls `load_data_for` rather than `load_all_data`, so it never reaches the traversal. Staged filtering is unreachable from quantization: the loader is constructed non-staged (`llama-quant.cpp:905`, `ordered_parts=false`) and the filter is `thread_local` state established only by RAII `skippy_filter_scope` at the three model-open sites in patch 0005. Conflating the branches is the class of assumption that caused #1298.

**Scope of the runtime evidence.** The V100 replay exercised the model-load `ctx_map` branch, including the arm where one physical weight is uploaded twice in a single loader (counted twice, 648,313,856 balances) — a dedup-by-name implementation would undercount by 165,306,368 bytes and reintroduce the bug. The `weights_map` fallback branch added at `a8d44708` has **not** executed in any run; its safety rests on the reachability argument above, not on measurement. No runtime claim is made about the split-expert `nullptr` arm.

Runtime replay of the preserved one-GPU CUDA reproducer and published-package validation are follow-up gates before merge.

## Patch queue

The fix is folded into patch 0001, which owns staged model loading/filtering. The patch file was refreshed from the pinned upstream tree; later patches apply cleanly.

### DIFF
--- a/third_party/llama.cpp/patches/0001-Add-staged-model-graph-and-family-support.patch
+++ b/third_party/llama.cpp/patches/0001-Add-staged-model-graph-and-family-support.patch
@@ -4,14 +4,14 @@ Date: Sat, 8 Aug 2026 16:02:19 +1000
 Subject: [PATCH 01/15] Add staged model graph and family support
 
 
-diff --git a/include/llama.h b/include/llama.h
-index 5a0b66dc2..a7893e5e5 100644
+Index: b/include/llama.h
+===================================================================
 --- a/include/llama.h
 +++ b/include/llama.h
 @@ -517,6 +517,13 @@ extern "C" {
                                   size_t    n_paths,
                struct llama_model_params    params);
- 
+
 +    // Load a model from multiple GGUF parts without requiring native split metadata.
 +    // The first path provides model metadata; tensors are resolved from all paths in order.
 +    LLAMA_API struct llama_model * llama_model_load_from_parts(
@@ -22,11 +22,11 @@ index 5a0b66dc2..a7893e5e5 100644
      LLAMA_API void llama_model_save_to_file(
              const struct llama_model * model,
                          const char * path_model);
-diff --git a/src/llama-arch.cpp b/src/llama-arch.cpp
-index 73fb8b981..0d01f8eb9 100644
+Index: b/src/llama-arch.cpp
+===================================================================
 --- a/src/llama-arch.cpp
 +++ b/src/llama-arch.cpp
-@@ -71,7 +71,6 @@ static const std::map<llm_arch, const char *> LLM_ARCH_NAMES = {
+@@ -71,7 +71,6 @@ static const std::map<llm_arch, const ch
      { LLM_ARCH_OLMO,             "olmo"             },
      { LLM_ARCH_OLMO2,            "olmo2"            },
      { LLM_ARCH_OLMOE,            "olmoe"            },
@@ -34,7 +34,7 @@ index 73fb8b981..0d01f8eb9 100644
      { LLM_ARCH_OPENELM,          "openelm"          },
      { LLM_ARCH_ARCTIC,           "arctic"           },
      { LLM_ARCH_DEEPSEEK,         "deepseek"         },
-@@ -101,7 +100,6 @@ static const std::map<llm_arch, const char *> LLM_ARCH_NAMES = {
+@@ -101,7 +100,6 @@ static const std::map<llm_arch, const ch
      { LLM_ARCH_GRANITE,          "granite"          },
      { LLM_ARCH_GRANITE_MOE,      "granitemoe"       },
      { LLM_ARCH_GRANITE_HYBRID,   "granitehybrid"    },
@@ -42,15 +42,15 @@ index 73fb8b981..0d01f8eb9 100644
      { LLM_ARCH_CHAMELEON,        "chameleon"        },
      { LLM_ARCH_WAVTOKENIZER_DEC, "wavtokenizer-dec" },
      { LLM_ARCH_PLM,              "plm"              },
-@@ -147,6 +145,7 @@ static const std::map<llm_arch, const char *> LLM_ARCH_NAMES = {
+@@ -147,6 +145,7 @@ static const std::map<llm_arch, const ch
      { LLM_ARCH_MELLUM,           "mellum"           },
      { LLM_ARCH_NANBEIGE,         "nanbeige"         },
      { LLM_ARCH_QWEN3TTS,         "qwen3tts"         },
 +    { LLM_ARCH_INKLING,          "inkling"          },
      { LLM_ARCH_UNKNOWN,          "(unknown)"        },
  };
- 
-@@ -222,11 +221,6 @@ static const std::map<llm_kv, const char *> LLM_KV_NAMES = {
+
+@@ -222,11 +221,6 @@ static const std::map<llm_kv, const char
      { LLM_KV_TIME_DECAY_EXTRA_DIM,              "%s.time_decay_extra_dim"              },
      { LLM_KV_RESIDUAL_SCALE,                    "%s.residual_scale"                    },
      { LLM_KV_EMBEDDING_SCALE,                   "%s.embedding_scale"                   },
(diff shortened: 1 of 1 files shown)


michaelneale:
🤖 **Runtime replay PASSES.** Ran at `01dcafa8` on a Tesla V100 (a third distinct GPU model, so not card-specific). Total spend $0.107.

## W9_1 green, and all previously-faulting configs

`GGML_CUDA_DEVICES=1 GGML_CUDA_NO_PINNED=1 --split-layer 9 --layer-end 28 --n-gpu-layers 99` -> **EXIT=0**, both stages loaded, generation reached.

| split | ngl | before | after |
|---|---|---|---|
| 9 | 99 | 139 | **0** |
| 10 | 99 | 139 | **0** |
| 14 | 16 | 139 | **0** |
| 20 | 20 | 139 | **0** |
| 14 | 99 | 139 | **0** |

## Accounting equality

`size_data` is now **315,794,432** — exactly the pre-fix value plus the wrongly-subtracted `ggml_nbytes(token_embd)` (150,488,064 + 165,306,368). Instrumented, summing actual uploads per loader:

```
loader1: size_data=315794432  size_done=315794432  EQUAL=True  progress_max=1.0000
loader2: size_data=483007488  size_done=483007488  EQUAL=True  progress_max=1.0000
```

Exact equality with progress capping at 1.0000 means the accumulator reaches `size_data` precisely at the final upload, so teardown fires once after the last context rather than early.

## Tied-weight arm exercised

Setting both `include_embeddings` and `include_output` on **stage 0** is rejected by a deliberate guard:

```
Error: failed to open stage 0
Caused by: InvalidArgument: only the final runtime slice may include output tensors
```

So the conjunction is only legal on the final slice. Moved to stage 1:

```
token_embd uploads (loader, seq, bytes):
  (1, seq0, 165306368)
  (2, seq0, 165306368)   <- tied weight, first context
  (2, seq1, 165306368)   <- same physical weight, second context

loader1: size_data=315794432  size_done=315794432  EQUAL=True
loader2: size_data=648313856  size_done=648313856  EQUAL=True
```

The same physical weight is uploaded twice within one loader and counted twice; 648,313,856 balances exactly. This is the case that distinguishes the traversal-based formula from a unique-physical-bytes formula, which would have undercounted by 165 MB here. Previously this arm was correct-by-reading only — no run in the earlier dataset uploaded any tensor more than once per loader.

The rejected stage-0 attempt also balanced (`size_data = size_done = 497825792`), so accounting is correct even on the configuration the guard refuses.

## Ordering check done before spending

The new `size_data` recompute iterates `ctx_map`, so ordering matters: `create_tensor` calls begin at `llama-model.cpp:1377` and `init_mappings` is called at `:1538`, so `ctx_map` is fully populated when the recompute runs.

## Provenance

| item | value |
|---|---|
| commit | `01dcafa8a75d3c6ab2428b52846c588a8c389c56`, clean tree |
| patched llama.cpp | `814112dc12d5ff179f77a4ca35910323e06a7834` |
| fixed binary | `0af9a50ef19eb3da5fad063adb281ac8781174e4f015bd38f3ac93b29e38bbb7` |
| instrumented binary | `3091b08178f443dfe95979d6aa6d24e327481bc5e781e0956246c9febc1288ca` |
| model | `Qwen3-0.6B-Q8_0.gguf` sha256 `e150ed544dfe6016930c026a93913a5e3184181ebfe6ab2223ae01dd0491784c` |
| GPU | Tesla V100-SXM2-32GB |

Not covered by this run, by instruction: MoE / split-expert validation and recurrent cross-host behaviour. Both remain open and are tracked separately.

michaelneale:
🤖 Runtime validation passed on the load-accounting implementation at `01dcafa8`: the previously deterministic W9_1 crash and four other faulting shapes now reach generation; per-loader `size_done == size_data`, progress remains ≤1.0, mmap teardown occurs after the final context, and a tied weight materialized in two contexts is counted/uploaded twice. Evidence: `RESEARCH/evidence-qwen38/replay-47674079/`. Total replay spend $0.107; contracts destroyed. Current head `a8d44708` differs only by preserving historical `weights_map` accounting when quantization calls `init_mappings` before model contexts exist; the validated model-load branch is unchanged. Full patch replay and `just build` pass at current head.

michaelneale:
🤖 **Scope note on the runtime evidence relative to the current head.**

The validation above was run at `01dcafa8`. Diffing `01dcafa8..a8d44708`: one file, one hunk, 15+/11-, entirely inside the `init_mappings` recompute. The validated code is wrapped rather than changed:

```diff
+    if (!ctx_map.empty()) {
         size_data = 0;
         for (const auto & [_, ctx] : ctx_map) {
             ... get_weight(...) != nullptr -> size_data += ggml_nbytes(tensor)   // unchanged
         }
+    } else {
+        size_data = 0;
+        for (const auto & [_, weight] : weights_map) {
+            size_data += ggml_nbytes(weight.tensor);
+        }
     }
```

Every loader in the replay had a non-empty `ctx_map` — necessarily, since the observed `size_data` values equal the ctx-map sums. **So the model-load evidence transfers to the current head.**

**What the replay does not cover:** the new `else` branch. `local-split-inprocess` never reaches it (quantization calls `init_mappings` before model construction, and no quantization was run), so that path is correct-by-reading only.

Two review points on it, neither requiring hardware:

1. The fallback restores the original `weights_map` sum, i.e. the pre-fix formula. That is appropriate for quantization, which uses `load_data_for` rather than `load_all_data` and so has no `size_done` accumulator to match — but it means `size_data` semantics now differ by call site. Worth an explicit comment that the branches are deliberately not interchangeable, given the original bug came from one accounting rule being assumed to serve both paths.
2. A `quantize` invocation against a **filtered/staged** loader would take the `else` branch and receive the unfiltered total. If that combination is reachable it reintroduces a mismatch. Whether staged quantization is a supported path is unverified here.

ndizazzo (changes requested):
**Request changes**

The retained-tensor accounting logic looks correct; I found no meaningful performance bottleneck or runtime/build issue.

Issues to address:

- **Blocking patch-queue format issue:** `third_party/llama.cpp/patches/0001-Add-staged-model-graph-and-family-support.patch` was rewritten from Git format-patch form into a generic unified diff, removing all 119 blob-index records. Pinned replay still passes, so this is not a current build or runtime failure. However, it disables `git am --3way` fallback when direct context application fails and diverges from the repository patch-generation workflow. Please regenerate 0001 with `git format-patch` and repeat the pinned replay.
- **Non-blocking cleanup:** Remove the now-unused `<unordered_set>` include from `src/llama-model-loader.h`.

ndizazzo (approved):
The comment was wonky and not really a P1 - I think the findings are actually correct, it's a standard patch now instead of git formatted one. Might wanna keep it consistent. Unblocking so you can merge when fixed

Issue #1289, opened 2026-08-13
Unify `auto` and `mesh` into one `auto` directive (MoA with modes)


michaelneale:
*Tracking issue opened by Dario (agent) on behalf of @michaelneale, per direction in the `feature-core-ai` channel. Origin: the routing/sizing findings in #1246. This body was rewritten to carry the agreed design; earlier comments record how it got here.*

## Decision

**Unify `model=auto` and `model=mesh` into a single `auto` directive that is MoA, operating in other modes as needed.**

We do not want two routing mechanisms. `auto` currently picks one "strong" model; `mesh` convenes an MoA committee. They are not really alternatives — `mesh` already degrades into single-model routing when a committee can't form, so **`auto` is a special case of `mesh` that should not be separately addressable.**

`auto` is the surviving public name: it's what every OpenAI-compatible client already understands, and `mesh` names an implementation (a committee) rather than an intent (route this well). Callers shouldn't have to know whether a committee formed — if they must choose between two names, the mechanism has leaked into the API. **`mesh` becomes a deprecated alias.**

**Explicit model IDs are unaffected.** `model=qwen3-30b` keeps working exactly as today. This retires a *directive*, not direct addressing.

## Target design

```
model=auto  (or omitted — see open questions)
  │
  ├─ media present?       → SingleModel  (vision/audio-capable, capability-filtered)
  ├─ stream requested?    → SingleModel  (true token passthrough)
  ├─ pool collapses to 1? → SingleModel  (the pool's own chosen member)
  └─ otherwise            → Committee    (MoA fan-out)

model=<specific>          → direct routing, unchanged
model=mesh                → deprecated alias for auto
```

Every branch shares **one selector**: capability filters (tools / reasoning / vision), readiness, affinity, and context-budget fit. The branches differ only in how many candidates arrive and why.

The load-bearing change is conceptual: **`SingleModel` becomes a mode, not a fallback.** Today it's reachable only as MoA degradation, so it inherits none of `auto`'s routing quality. After this it is a first-class mode that media, streaming, and small pools all select deliberately.

### Mode signals are declarations, not heuristics

Both gates read facts the client stated, and both are already computed in the codebase:

- **Streaming** — `"stream": true`. MoA already parses it (`moa_gateway/mod.rs:156`), but only to choose the response *encoding*. This promotes an existing signal to a routing input; it is not a new classifier.
[...]

PR #1309, opened 2026-08-14
Consolidate model=auto and model=mesh into one automatic-routing directive

michaelneale:
Closes part of #1289.

## What you can now do

**Send an image to `model=mesh` and get an answer about the image.** Today the mesh silently drops it and answers the text half of your question as if you had sent no image — confidently, so there is nothing in the response to tell you. That has been live for anyone using the console's Auto button since it began pointing at `mesh`. Now the request goes to a model whose runtime actually supports vision, or fails with `422` when the mesh has none.

**Ask `model=mesh` to stream and get real tokens as they are generated.** Committee answers cannot stream — workers are called non-streaming because the arbiter needs whole drafts to compare — so streamed committee responses were synthesised after the answer was already complete. A streaming request is now served by a single model, which streams for real.

**Send `model=mesh` on a single-model node and find it in `/v1/models`.** The name used to require two models and vanish below that, so the one name clients are told to send was missing exactly where a new user starts. It also advertised `vision_status: unsupported`, which was wrong once media requests route to a capable model; it now reports the mesh's combined capabilities.

**Send `model=auto` and nothing breaks.** It is a deprecated alias for `mesh` and logs a warning.

**See the real size of split-GGUF models.** A 4-shard model reported the size of shard 1 alone — about a quarter of the truth — which mis-ranked it against single-file peers everywhere size is used, including `/v1/models`.

| model | before | after | publisher's label |
|---|---|---|---|
| `Qwen3-Coder-Next-Q4_K_M` (4 shards) | 24.57 B | **79.67 B** | `512x2.5B` |
| `command-a-reasoning-Q4_K_M` (2 shards) | 65.86 B | **111.06 B** | `111B` |

**Get sensible committee roles for quantised models.** Role assignment ranked on model names, and a trailing quant tag hijacked the parse — `-4bit` read as "4B". So a 70B model was assigned the `Fast` role, which is packed at 256 max tokens with no tool schemas, while an 8B became the reducer and the arbiter's patience anchor:

```
pool ["Llama-3.3-70B-Instruct-4bit", "Qwen3-8B"]
   before:  70B -> Fast         8B -> Strong + reducer
   after:   8B  -> Fast         70B -> Strong + reducer
```

Explicit model routing (`model=qwen3-30b`) is unchanged throughout.

## Documentation

New user-facing page: **Automatic routing** (`website/src/docs/pages/automatic-routing.md`), because `mesh` and `auto` were essentially absent from the public site — `openai-compatible-api.md` said only "Mesh decides which local or peer model handles the request". It covers what you get for each mesh shape, when you get a single model and why, the streaming-versus-resilience trade, how to pick by size yourself from `/v1/models`, and the `auto` migration. Linked from the API page, the model-choice page, and the docs index.

## Architecture

Mode selection lives in one module, `network/openai/automatic.rs`, which owns the directive names and the `Committee` / `SingleModel` decision. It reads **declarations the client made** — content blocks and `stream` — rather than guessing intent, so both gates are facts about the request rather than heuristics.

Both entry paths share that one policy, which mattered because they were broken differently:

- **Host ingress** resolves the directive through the same selector `auto` already used, so `mesh` now gets the media capability filter, readiness, affinity and context-budget fit. The leak here was `resolve_auto_routed_model` early-returning for any model that was not `"auto"`.
- **Passive transport** hands the stream back with the directive *still in the request* rather than rewriting it to a concrete model, so the auto selector downstream still runs. Rewriting first would have made the request look explicitly addressed and skipped the very media filter a media request needs.

`SingleModel` is a mode, not a fallback: media, streaming and model-less requests select it deliberately.

## Protocol

**Compatible.** No gossip fields, protobuf schemas, or stream types change.

- `auto` keeps working as an alias, so existing clients need no change.
- `/v1/models` gains `mesh` on single-model nodes and reports truthful capabilities for it. Additive; a client that ignores the entry is unaffected.
- Mixed-version meshes are unaffected: this changes how one node routes a request it received, not what nodes tell each other.

The one behaviour change for an existing client is that `model=mesh` with `stream: true` now streams from a single model instead of returning a synthesised committee stream. That is the point of the change — it preserves what streaming clients already expected and previously did not get.

## Validation

| check | result |
|---|---|
| `cargo test -p mesh-llm-host-runtime` | 2492 passed, 0 failed |
| `cargo test -p mesh-mixture-of-agents` | 183 lib + 29 sim/integration passed |
| `cargo test -p model-artifact` | 36 passed |
| `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings` | clean |
| `cargo clippy -p mesh-llm --all-targets -- -D warnings` | clean |
| `cargo fmt --all --check` | clean |
| Eleventy build | new page renders, links resolve |

The three new media/streaming tests were checked against the old behaviour: reverting the directive recognition makes exactly those three fail, so they pin the defect rather than merely passing.

Resilience is unchanged and re-verified, since role assignment feeds the patience gate: `sim_partial_worker_survival`, `sim_all_workers_fail`, `sim_refinement_mesh_conditions`, `sim_strong_patience` all pass.

**Not covered:** a live 2-node mesh confirming end-to-end that a corrected size changes committee selection and that a real vision model receives a real image. Everything above is single-node unit/sim plus direct measurement on real artifacts. Happy to run the 2-node shape before merge if wanted.

## Still open in #1289

Deliberately not in this PR, and the issue tracks them:

- **Retiring pipeline routing and the keyword classifier.** `route_with_pipeline` is dead code, but the live `try_pipeline_route` is still reachable for a two-local-model node, and `router::classify` still feeds it. That is a behaviour removal for anyone relying on plan injection and deserves its own review rather than riding along here.
- **Removing `auto` entirely.** It stays an alias for at least one release; removal wants a mixed-version check first.
- **Committee quality on all-small pools.** `pool.rs` deliberately declines to convene a committee when every member is under ~10B, on measured evidence. The one arm that beat solo at 8B scale was two-round layered, which is not what the guard measured — worth its own eval-gated issue.
- **Sticky/availability-aware single-model selection.** The degrade path still picks first-alphabetically and takes neither `required_tokens` nor affinity.

## Summary by CodeRabbit

* **New Features**
  * Added accurate parameter counting across complete split GGUF model bundles.
  * Model tier detection now uses verified parameter counts when available.
  * Added automatic routing with committee or single-model selection based on request needs, including media and streaming support.
  * Virtual model listings now advertise available mesh capabilities.
* **Bug Fixes**
  * Incomplete bundles no longer report incorrect parameter counts.
  * Model-size information is preserved during worker selection and dispatch.
* **Documentation**
  * Added automatic-routing guidance and model-selection examples.
* **Tests**
  * Expanded coverage for routing, capabilities, bundle scanning, tier boundaries, and worker assignment.


crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -197,6 +205,25 @@ async fn resolve_auto_routed_model(
...
+        // gateway picks the request up.
+        automatic::ServingMode::Committee => {
+            return AutoRouteResolution::Continue {
+                effective_model: Some(automatic::DIRECTIVE.to_string()),
+                classification: None,
+            };
+        }
+        // Single-model mode falls through to the capability-, readiness- and
+        // affinity-aware selection below.
+        automatic::ServingMode::SingleModel(reason) => tracing::debug!(
```
ndizazzo, line 228: The PR description says host ingress should resolve `mesh` through the normal automatic selector and then serve the request from that concrete model. I think the last handoff is missing here: when the original model is `mesh`, the selected model is not written back into `request.raw`, which is what `route_model_request` forwards. Could we include `mesh` in this rewrite path and test that the chosen backend receives its concrete model ID?

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -197,6 +205,25 @@ async fn resolve_auto_routed_model(
         };
     };
 
+    automatic::warn_if_deprecated_alias(request.model_name.as_deref());
+
+    match automatic::serving_mode(request.model_name.as_deref(), body_json) {
+        // Committee mode keeps the directive as the effective model so the MoA
+        // gateway picks the request up.
+        automatic::ServingMode::Committee => {
```
ndizazzo, line 220: The intent here seems to be that `auto` becomes a compatibility alias for `mesh` without breaking existing callers. Does that mean every existing `auto` request should keep working, or only chat-style requests? For example, a non-streaming `/v1/completions` request previously selected one concrete model, but now enters committee mode and is rejected because it has no `messages` array. If preserving existing `auto` behavior is the goal, non-chat request shapes may need to remain on the single-model path.

ndizazzo (commented):
A couple of questions about how the automatic-routing intent carries through the existing request paths.

michaelneale:
Yeah not sure if we want this yet @ndizazzo ? Yiu think not yet?

ndizazzo:
> Yeah not sure if we want this yet @ndizazzo ? Yiu think not yet?

@michaelneale I dont think it's overly problematic... I just had a couple what-if thoughts when lightly skimming through the PR that made the direction seem a little more complicated depending on the intent. Honestly, I haven't used MoA as much as you have - I tend to pick one and stick with it through a session

I also know you've been putting MoA through it's paces, so ultimately I defer to the call you make for it

michaelneale:
@ndizazzo both of your comments were right, and the first one was the more serious of the two — thanks for pushing on it. Pushed `67a3c26c`.

## The missing handoff (your `ingress.rs:221` comment)

You were exactly right that the selected model never made it back into `request.raw`. I printed the forwarded bytes rather than arguing about it:

```
effective_model    = Some("vision-model")
request.model_name = Some("mesh")
forwarded body     = {"messages":[...],"model":"mesh"}
```

The backend validates `model` against its own advertised identity (`skippy-server/src/frontend/generation/parsing.rs:21-32`, `ensure_requested_model` → `model_not_found`), so this is a **404, not a wrong answer**. My PR description claimed image requests now reach a capable model; on the host path they reached nothing at all. The gate in `maybe_enable_auto_route_hooks` matched only a missing model and the literal `"auto"` — it never learned the `mesh` spelling this PR introduced.

Fixed and pinned: `a_single_model_decision_is_written_into_the_forwarded_body` fails on the old gate (12 pass / 1 fail) and passes now. Committee mode is explicitly exempted from the rewrite, since the MoA gateway self-gates on the directive still being in the body.

Worth noting your instinct generalised further than the line you flagged: the transport path already injected `mesh_hooks` for both spellings (`transport.rs:464-491`) while ingress did it for `auto` only, so the same request behaved differently depending on which path it arrived through. The widened gate makes them agree.

## Non-chat request shapes (your `ingress.rs:213` comment)

Also correct, and it is a compatibility break rather than a hypothetical. `/v1/completions` carries `prompt`, not `messages`, so under my change `model=auto` on that endpoint entered committee mode and got `400 "MoA requires a non-empty messages array"` — a shape that previously selected one concrete model.

To answer your question directly: **every existing `auto` request should keep working**, not just chat-style ones. Committee eligibility now depends on the forwarded endpoint. `/v1/responses` is normalised onto chat completions before routing, so it stays committee-eligible; `/v1/completions` is served by a single model. Gating on the client's path instead of the forwarded one would have silently excluded `/v1/responses`, so there is a test pinning that too.

## What a cold review of the fix caught

I had an independent reviewer go over the fix before pushing, and it found a real mistake in it. My first version hoisted *all three* mode checks above the gateway's `messages` contract check, when only the endpoint check needed to move. That silently removed the gateway's precise 400 for malformed chat bodies like `{"model":"mesh","stream":true}` with no `messages` — they still fail, but further downstream with a vaguer error.

Mode selection is now two stages: an **envelope** stage (endpoint shape) before the contract check, and a **content** stage (media, streaming) after it. `a_malformed_chat_body_stays_committee_eligible_at_the_envelope` fails if anyone re-merges them.

## Validation

2488 host-runtime, 12 MoA suites, 36 model-artifact, all exit 0. Clippy `-D warnings` clean on `mesh-llm-host-runtime` and `mesh-llm`, fmt clean. Each of the three new gates was verified to fail without its fix.

Docs updated: the single-model case table is now four cases, with the `/v1/completions` row.

Still outstanding and unchanged: the live 2-node run. That is the last gap between this and a claim I can fully back — @michaelneale, happy to run it before merge if you want it.

michaelneale:
## Live 2-node run: done, and it closes the gap flagged above

The comment saying the live run was outstanding is stale. It was run on this branch at `67a3c26c` and the results are below. Recording here so the PR carries its own evidence.

**Setup:** two Macs (M5 serving `unsloth/Qwen3.8-27B-GGUF:Q4_K_M`, plus M4) over relay, invite-token private mesh, ports 9447/3232. Byte-identical binary on both nodes (`shasum 5f3a80abf648382e`), `0.76.0-rc1+g67A3C2`, tree clean at `67a3c26c` before and after.

| case | expected (stated first) | got |
|---|---|---|
| `mesh` + text | committee convenes | **200**, `x-moa-turn: fanout`, `workers-ok: 2`, reducer engaged |
| `mesh` + image | 422 media-unsupported | **422** |
| `auto` on `/v1/completions` | 200 concrete model, not 400 | **200** from the 27B |
| `mesh`, no `messages` | precise 400 | **400**, exact text |
| `mesh` + stream | 200 SSE, concrete model name | **200**, `model: unsloth/Qwen3.8-27B-GGUF:Q4_K_M` |

### The control is the actual result

Rebuilt with **only** the writeback predicate reverted to its pre-fix form (`is_none() || == "auto"`), everything else identical, and ran the same bytes against the same mesh:

```
pre-fix : HTTP 404 {"message":"model not found: mesh","code":"model_not_found"}
fixed   : HTTP 200 SSE, model=unsloth/Qwen3.8-27B-GGUF:Q4_K_M
```

The defect reproducing and then not reproducing, one predicate apart. The control edit was then reverted and `git status` confirmed clean at `67a3c26c` — the binary measured is the one in this PR.

### Two limits on the above, stated plainly

1. **The first committee attempt was a false pass.** It returned HTTP 200 in 0.14s with **no `x-moa-*` headers** — it had silently degraded to a single model, because both models were 3B and the pool hit the pre-existing all-small collapse (`pool.rs:346-353`, where a committee measures worse than its best member). A 200 is not evidence of a committee; the MoA admission log is. Swapping to the 27B made the pool mixed-tier, a genuine committee formed, and the second node's log shows it serving a draft inside the request window, so the fan-out did cross the network.

2. **No vision model exists on either box**, so image-routing-to-a-capable-model was not tested live. The `422` proves the media *filter* engages, not that a projector-loaded model serves an image. That path remains covered by unit tests only. Vision is tracked separately in #1348.

Both nodes stopped afterwards, temp binaries removed, no state left behind.

### Two `"auto"` literals this PR does not sweep

Not regressions, not user-visible, but the same shape as the two review defects already fixed here — a gate that never learned the new spelling:

- `network/metrics.rs:1159` — `normalized_model_name` filters `"auto"` but not `"mesh"`, so the directive appears as a real model name in metrics labels where `auto` is correctly suppressed.
- `network/openai/request_parse.rs:928` — `rewrite_public_model_alias` early-returns on `requested == "auto"`; a `mesh` request falls through to `internal_model_for_public_id("mesh", ...)`. Believed benign because that lookup misses — benign by miss, not by intent. Reachability not proven.

Routing both through `automatic::is_directive` would take the consolidation from 3 of 5 call sites to 5 of 5. Fine as a follow-up.

*(Live run and results by @Dario; recorded here by me.)*

PR #1313, opened 2026-08-14
fix(ffi): match mtmd_helper_bitmap_init_from_buf's current signature

michaelneale:
## Problem

Every vision request through the OpenAI surface fails with HTTP 502 `multimodal tokenization failed with status 2` / `multimodal prompt evaluation failed with status 1`. The native log shows the mechanism:

```
mtmd_tokenize: error: number of media markers in text (0) does not match number of bitmaps (1)
mtmd_encode_chunk_impl: image tokens batch is placeholder
failed to encode image slice
```

The model itself is fine — the same GGUF + mmproj through `llama-mtmd-cli --jinja` (built from the vendored llama.cpp checkout) describes images perfectly.

## Root cause

The vendored llama.cpp changed `mtmd_helper_bitmap_init_from_buf` when video support landed:

```c
// now: 4 args, returns a wrapper struct
mtmd_helper_bitmap_wrapper mtmd_helper_bitmap_init_from_buf(ctx, buf, len, bool placeholder);
struct mtmd_helper_bitmap_wrapper { mtmd_bitmap * bitmap; mtmd_helper_video * video_ctx; };
```

but `skippy-ffi` still declared the old 3-arg bare-pointer form. On aarch64 the wrapper's first field happens to come back in x0, so the bitmap pointer accidentally survived — but the new `placeholder` argument read **garbage from an unset register**. Whenever it read nonzero, llama.cpp constructed an empty placeholder bitmap, and every downstream encode failed at the placeholder gate. Text-only paths never call this helper, which is why nothing else broke.

## Fix

- Update the FFI declaration to the 4-arg wrapper-returning form (`MtmdHelperBitmapWrapper { bitmap, video_ctx }`), pass `placeholder = false`.
- Add `mtmd_helper_video_free` and free the video context alongside the bitmap in both `eval_media` and `eval_media_frame`.

## Verification

Muse-Glimmer-30B Q4_K_M + mmproj on M4/Metal, `POST /v1/chat/completions` with an `image_url` data-URI part:

- **Before**: 502 at the placeholder gate, native log `image tokens batch is placeholder`
- **After**: image preprocessing and encoding succeed — native log `encoding image slice... image slice encoded in 566 ms`, generation proceeds
- Text-only requests unchanged ("391" sanity check passes)
- Control: `llama-mtmd-cli --jinja` with identical model files works, confirming the break was in the FFI layer, not the model or vendored llama.cpp

Note: with this fix the image encodes and generation runs, but Glimmer's vision output then hits a separate response-parsing issue (`peg-native` chat parser rejects the vision-turn output) — that's an unrelated bug in the chat-format parser, will file separately. This PR fixes the FFI break that blocked all multimodal input at the bitmap layer.

Clippy clean on both crates.

## Summary by CodeRabbit

* **New Features**
  * Improved multimedia bitmap handling with support for placeholder images and associated video context data.
  * Added safer cleanup for video-related resources during media processing.

* **Bug Fixes**
  * Preserved bitmap validation and error handling while improving resource management.

### DIFF
--- a/crates/skippy-ffi/src/lib.rs
+++ b/crates/skippy-ffi/src/lib.rs
@@ -316,6 +316,20 @@ pub struct MtmdInputChunks {
 }
 
 #[repr(C)]
+pub struct MtmdHelperVideo {
+    _private: [u8; 0],
+}
+
+/// Mirrors llama.cpp's `mtmd_helper_bitmap_wrapper`: the decoded bitmap plus
+/// an optional video context (non-null only for video inputs, which own the
+/// frame storage the bitmap points into).
+#[repr(C)]
+pub struct MtmdHelperBitmapWrapper {
+    pub bitmap: *mut MtmdBitmap,
+    pub video_ctx: *mut MtmdHelperVideo,
+}
+
+#[repr(C)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum MtmdInputChunkType {
     Text = 0,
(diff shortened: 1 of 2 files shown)


michaelneale:
🤖 Closing per review discussion: this fix is necessary but not sufficient — with it applied, vision requests still fail one layer deeper (degenerate sampling after image eval, #1314). Rather than landing a fix that doesn't make vision work end-to-end, folding this into #1314 so both layers land together. The branch `fix/mtmd-bitmap-helper-ffi` stays up with the reviewed FFI fix ready to cherry-pick.

PR #1316, opened 2026-08-14
Advertise externally observed UDP ports so port-remapped containers are directly reachable

### LINKED ISSUE #1300: 🤖 Invite advertises locally-enumerated public IP with unverified port, forcing relay-only on port-remapping hosts
## Summary

On a host whose container has a globally routable public IPv4 bound directly to its interface, but whose UDP port is remapped by the provider's NAT, mesh advertises the **container** port in the invite token. The advertised tuple is unreachable, hole punching never happens, and split admission correctly refuses the peer as `stage_path_relay_only` — despite the path being genuinely punchable.

Observed on Vast.ai (Estonia), between two machines 4 ms apart in the same facility, which fell back to a **52 ms relay in Frankfurt**.

## Evidence

Vast mapped container UDP `41842` to external `23555`. The invite token advertised:

```json
{"Relay":"https://euc1-1.relay..."},{"Ip":"172.17.0.3:41842"},{"Ip":"213.5.72.196:41842"}
```

`213.5.72.196:41842` is not reachable from outside. Runtime result:

```
Peer 30ef3a48e5 RTT: 52ms (relay)
Split waiting to retry: found 1 eligible; excluded [30ef3a48e5:stage_path_relay_only];
blockers [stage_path_relay_only=1]; next_step: Fix firewall/NAT/direct-path connectivity
```

**The path is punchable** — verified manually:
- observed public address is globally routable (`213.5.72.x`, not `100.64/10`)
- source port is preserved through the NAT: sending from container `41843`, the peer observed `('213.5.72.211', 41843)`
- full bidirectional punch succeeded: B -> `213.5.72.196:23555`, A replied to the observed tuple, B received `GOT REPLY ('213.5.72.196', 23555)`

So this is neither CGNAT nor symmetric NAT. The only thing preventing direct mode is that mesh advertised the wrong port.

**The discovery log line is decisive:**

```
INFO mesh_llm_host_runtime::mesh::stun: QUIC endpoint discovered public address addr=213.5.72.196:41842
```

Discovery *succeeded* and returned the container port. So `public_addr` was `Some(...)` with the wrong port — this is not a "net report never completed" case, and it is not the invite-token guard at `node_identity.rs:105-112` either (injecting that value would have advertised the same wrong port).

## Cause

`crates/mesh-llm-host-runtime/src/mesh/stun.rs:4-9`:

```rust
fn public_ipv4_addr(endpoint_addr: &iroh::EndpointAddr) -> Option<SocketAddr> {
    endpoint_addr.ip_addrs().copied().find(is_public_ipv4_candidate)
}
```

`endpoint.watch_addr()` returns a set that mixes **locally enumerated** interface addresses with **reflexively observed** ones. This function cannot distinguish them — it takes the first globally routable IP it finds. When the container has a public IP bound directly, that candidate comes from local interface enumeration and its port was never externally verified. The function name `stun_public_addr` overpromises: nothing it returns is necessarily STUN-derived.

This never reproduces on a home LAN, where the local address is private and the reflexive candidate is the only public one.

## Test gap

`stun.rs`'s own test `preserves_port_discovered_by_quic_endpoint` constructs an `EndpointAddr` with a single mapped address, so it passes while never covering the case where a local public candidate **precedes** the reflexive one. A fix needs a test with both candidate kinds present.

## Suggested direction

Split admission should trust only externally verified reachable tuples. A locally enumerated global IP is a *candidate*, not proof that its port is externally preserved. Concretely, distinguish the provenance of address candidates rather than classifying by IP range alone, and prefer a reflexively observed tuple over a same-IP/different-port local one rather than advertising both (advertising both leaves a dead candidate).

## Related: diagnostics gap

`mesh-llm doctor` claims split-readiness diagnosis but covers neither of the two checks that actually discriminated here:

1. does a reply to the **observed** tuple arrive (real punchability)?
2. does the **advertised** candidate equal the externally reachable tuple?

(2) is exactly what failed, and nothing currently checks it. Both are worth adding.

Found during the Vast Estonia two-node split rehearsal on commit `8335d853140f183bd8ade84149ca3704c129d0bc`.

michaelneale:
## What changes for users

A node running inside a container whose UDP port is remapped by the host (Vast.ai and similar cloud GPU providers) is now reachable directly by its peers. Previously such a node advertised the port it bound *inside* the container, so no peer could reach it and every mesh split silently fell back to the relay — adding ~52 ms per boundary crossing and making split throughput numbers meaningless.

When a node's address cannot support a direct path, it now says so instead of advertising something unverified:

```
WARN QUIC endpoint public address varies by probe destination — address-dependent NAT, direct UDP unavailable
```

Fixes #1300.

## Architecture

`mesh::stun::stun_public_addr` previously took the first globally-routable IPv4 from `endpoint.watch_addr()`. That set is **local interface enumeration**, so a locally-enumerated candidate and an externally-observed one are the same type and indistinguishable. On a container holding a public IP on its own interface, the local candidate wins — carrying the container's port, not the host's mapped port.

Three changes:

1. **Observed address, where one exists.** With relays configured, discovery reads `endpoint.net_report()`. `Report::global_v4` is set from a QAD probe reply (`iroh-1.0.3/src/net_report/report.rs:81-99`), i.e. the address a relay observed us from, so it carries the NAT-mapped port by construction rather than by inference.

2. **Explicit rejection instead of silent wrong answers.** `mapping_varies_by_dest_ipv4 == Some(true)` means iroh saw *different* addresses from different relays (`report.rs:88-95`) — address-dependent NAT, not hole-punchable — so it is rejected with a warning rather than advertised. Variance being *unmeasured* (`None`) is not evidence of that and is still accepted.

3. **The source is now in the type.** `PublicAddr { addr, source: Observed | LocallyEnumerated }` replaces a bare `SocketAddr`, because the bug was precisely that the two were indistinguishable. `invite_token` uses it: an `Observed` address *replaces* any enumerated public candidate in the advertised set; a `LocallyEnumerated` one only fills a gap, preserving today's behaviour.

### Relay-disabled hosts keep working

Net reports probe *through relays*. With `RelayMode::Disabled` the relay map is empty, no QAD probe runs, and `global_v4` is never populated — so a net-report-only implementation would have left LAN-only and relay-disabled nodes with **no** public address at all, a regression. Those hosts fall back to interface enumeration and log that the port is unverified. Reviewers should check this reasoning specifically; it is the part most likely to be wrong.

### Cost: an unstable iroh feature

`Endpoint::net_report()` is gated behind iroh's `unstable-net-report`, explicitly outside semver ("may change in any release without a major version bump"). Enabled on `mesh-llm-host-runtime` only. The alternative is keeping a discovery path that cannot distinguish observed from enumerated addresses, which is the defect. **This is a maintainer call, not mine.**

## Protocol

No wire change. The advertised direct address is an existing gossip field; only its derivation changes, and the set now excludes a stale enumerated candidate when a verified one exists. Older peers see a reachable address where they previously saw an unreachable one.

## Dependency commitment — requires an unstable iroh feature

This enables `unstable-net-report` on `mesh-llm-host-runtime` only:

```toml
-iroh = { version = "1.0.3", default-features = false, features = ["metrics", "fast-apple-datapath", "portmapper", "tls-aws-lc-rs"] }
+iroh = { version = "1.0.3", default-features = false, features = [..., "unstable-net-report"] }
```

iroh documents that surface as exempt from semantic versioning (`iroh-1.0.3/src/lib.rs:294-299`):

> *"This API is unstable and gated behind the `unstable-net-report` feature. It is not covered by semantic versioning guarantees and may change in any release without a major version bump."*

`observed_public_ipv4` reads `report.global_v4` and `report.mapping_varies_by_dest_ipv4` directly, so **a patch-level iroh bump can break the build or silently change field semantics.** Any pinned-iroh bump must re-verify both fields.

This is a deliberate maintenance commitment, not an oversight: there is no stable iroh API exposing the externally-observed tuple, so the alternative is not addressing #1300 at all. Flagging it so a reviewer accepts it knowingly rather than discovering it at the next dependency bump. Raised by @Dario in review.

## Validation

- `cargo test -p mesh-llm-host-runtime --lib` — **2475 passed, 0 failed**

Six unit tests cover the decision, including the exact Estonia shape (bound 41842, observed 23555). The previous test could not have caught this bug: it *constructed* the address it then asserted on, so it never exercised the observed-vs-enumerated distinction.

**Not validated on hardware.** Proving the direct path end-to-end needs a cross-host run on a port-remapping provider. Do not read this as field-proven.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved public address detection and NAT-mapped port reporting.
  * Added more reliable handling for relay-enabled and relay-disabled network configurations.
  * Prevented destination-dependent address mappings from being used.
  * Improved invite address selection to avoid conflicting or redundant public IPv4 addresses.
  * Added fallback behavior when remote network observations are unavailable.

### DIFF
--- a/crates/mesh-llm-host-runtime/Cargo.toml
+++ b/crates/mesh-llm-host-runtime/Cargo.toml
@@ -57,7 +57,7 @@ skippy-runtime = { path = "../skippy-runtime", version = "0.72.1"  }
 skippy-ffi = { path = "../skippy-ffi", version = "0.72.1", default-features = false }
 skippy-server = { path = "../skippy-server", version = "0.72.1"  }
 skippy-topology = { path = "../skippy-topology", version = "0.72.1"  }
-iroh = { version = "1.0.3", default-features = false, features = ["metrics", "fast-apple-datapath", "portmapper", "tls-aws-lc-rs"] }
+iroh = { version = "1.0.3", default-features = false, features = ["metrics", "fast-apple-datapath", "portmapper", "tls-aws-lc-rs", "unstable-net-report"] }
 tokio = { version = "1", features = ["full"] }
 clap = { version = "4", features = ["derive"] }
 tracing = "0.1"
--- a/crates/mesh-llm-host-runtime/Cargo.toml
+++ b/crates/mesh-llm-host-runtime/Cargo.toml
@@ -57,6 +57,11 @@ skippy-runtime = { path = "../skippy-runtime", version = "0.72.1"  }
 skippy-ffi = { path = "../skippy-ffi", version = "0.72.1", default-features = false }
 skippy-server = { path = "../skippy-server", version = "0.72.1"  }
 skippy-topology = { path = "../skippy-topology", version = "0.72.1"  }
+# `unstable-net-report` is exempt from iroh's semver guarantees (iroh/src/lib.rs
+# "may change in any release without a major version bump"). `mesh::stun` reads
+# `net_report().global_v4` and `mapping_varies_by_dest_ipv4` directly, so any
+# iroh bump must re-verify both fields. Only this crate enables the feature;
+# the other iroh pins in the workspace deliberately do not.
 iroh = { version = "1.0.3", default-features = false, features = ["metrics", "fast-apple-datapath", "portmapper", "tls-aws-lc-rs", "unstable-net-report"] }
 tokio = { version = "1", features = ["full"] }
 clap = { version = "4", features = ["derive"] }
--- a/crates/mesh-llm-host-runtime/src/mesh/node.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/node.rs
@@ -33,7 +33,7 @@ pub struct RouteEntry {
 pub struct Node {
     pub(crate) endpoint: Endpoint,
     pub(crate) endpoint_secret_key: SecretKey,
-    pub(crate) public_addr: Option<std::net::SocketAddr>,
+    pub(crate) public_addr: Option<self::stun::PublicAddr>,
     pub(crate) quic_bind: QuicBindSelection,
     pub(crate) relay_policy: RelayPolicy,
     pub(crate) owner_keypair: Option<crate::crypto::OwnerKeypair>,
@@ -687,11 +687,12 @@ impl Node {
             .await;
         }
 
-        // Use iroh's net report because it probes through the endpoint's actual QUIC sockets.
-        // Its direct address includes the observed NAT-mapped port; substituting the configured
-        // bind port would advertise an address that was never verified externally.
+        // Take the address a remote probe server *observed* us from, so the advertised
+        // port is the NAT-mapped one rather than whatever port we bound locally. Local
+        // interface enumeration cannot tell those apart, and on hosts that hold a public
+        // IP on the container interface it silently advertises the unmapped port.
         let public_addr = if relay.policy.uses_raw_stun() {
-            stun_public_addr(&endpoint).await
+            stun_public_addr(&endpoint, relay.policy).await
         } else {
             tracing::info!("Raw STUN: disabled by LAN-only discovery mode");
             None
--- a/crates/mesh-llm-host-runtime/src/mesh/node_identity.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/node_identity.rs
@@ -105,10 +105,24 @@ impl Node {
     pub async fn invite_token(&self) -> String {
         let mut addr = self.endpoint_addr_for_advertisement();
         // Inject STUN-discovered public address if relay STUN didn't provide one.
-        if let Some(pub_addr) = self.public_addr
-            && !endpoint_addr_has_public_ipv4(&addr)
-        {
-            addr.addrs.insert(TransportAddr::Ip(pub_addr));
+        // An observed address supersedes any locally-enumerated public candidate:
+        // enumeration carries the port we bound, which is wrong wherever the host
+        // remaps ports, and `endpoint.addr()` mixes both kinds indistinguishably.
+        if let Some(pub_addr) = self.public_addr {
+            match pub_addr.source {
+                stun::PublicAddrSource::Observed => {
+                    addr.addrs.retain(|candidate| match candidate {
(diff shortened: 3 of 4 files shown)


michaelneale:
**Flagging a scope problem in my own PR body, prompted by a review of today's work. The code and tests are unaffected; two claims in the description overreach.**

**1. "Fixes #1300" is too strong — I have withdrawn the field evidence that motivated it.**

The PR body says every mesh split *"silently fell back to the relay"* with a `~52 ms per boundary crossing` cost. That framing came from an Estonia observation plus a later two-node attempt, and today I withdrew the second one entirely: in that run the split path was never entered at all (the plan took the capacity-based `Local` branch), so **transport selection was never in play and the relay fallback I attributed to this bug was confounded.**

What survives is the original mechanism, which is what the diff addresses: `stun_public_addr` derived its candidate from `endpoint.watch_addr()`, i.e. local interface enumeration, so a locally-enumerated address and an externally-observed one were indistinguishable by type. That is a real defect and the fix is correct in itself.

Suggest this becomes **"Addresses #1300"** rather than "Fixes", and #1300 stays open until a direct path is demonstrated on a port-remapping provider.

**2. A related result worth recording, because it complicates the story.**

On two same-facility nodes today I ran with `--bind-port` on both, so QUIC bound the exact UDP ports a STUN probe had verified as preserved and endpoint-independent, and the invite token advertised those ports. **The path still selected relay** (69–71 ms). That is not evidence against this fix — the nodes never reached split election, so I cannot separate cause from effect — but it does mean **advertising a correct port is not sufficient on its own**, and nobody should expect this PR alone to produce a direct path.

**3. The "Not validated on hardware" caveat already in the body is the operative one.** It stays true and should not be softened when this merges. Six unit tests cover the derivation decision including the Estonia shape; none of them prove a punch succeeds in the field.

Unchanged and still accurate: `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings` clean, `cargo test -p mesh-llm-host-runtime --lib` 2475 passed, `cargo fmt --all --check` clean, and the note that the previous test constructed the address it asserted on and so could never have caught this.

Also unchanged: this requires the iroh `unstable-net-report` feature, which is outside semver and needs a human call.

michaelneale:
I think this is junk

PR #1327, opened 2026-08-14
CI: Depot authority sentinel probe

ndizazzo:
Disposable protected CI probe for the Depot authority sentinel protocol.

This PR contains only a comment fixture and must not be merged. Repository sentinel variables will be bound to its exact merge ref only for the bounded seed/PR-write/main-verify sequence, then removed immediately.


ndizazzo:
Sentinel complete: unsafe in both directions. PR Depot restored and exactly validated the trusted-main seed, and trusted main restored and exactly validated the PR poison marker. The protected fail-closed gates failed as designed. Sentinel variables were removed immediately; this disposable PR must not be merged.

PR #1330, opened 2026-08-14
chore: sync source versions during GitHub releases

ndizazzo:
## Problem

Creating a GitHub release can advance the published tag—for example from `v0.72.1` to `v0.75.1`—without advancing the tracked version on `main`. The release workflow previously applied version changes only while preparing the detached release-tag commit, while `just release` maintained a separate local mutation path. That leaves the source tree, package metadata, and reported binary version behind the release developers and users see on GitHub.

This drift was present when the PR was prepared: the codebase still reported `0.72.1`, while the newest published GitHub release was `v0.76.0-rc1` (and the latest stable release was `v0.75.1`). This PR therefore also synchronizes the existing tracked version surface to `0.76.0-rc1`.

GitHub-generated notes also chose their comparison tag automatically. After publishing an RC, the final stable release could therefore compare against the RC and produce only the post-RC delta, potentially leaving the final notes empty.

## Why this work

Release versioning should have one low-maintenance authority regardless of how a developer starts a release. The durable source version should match every published non-canary release, tag-based releases should fail rather than publish drifted source, and RCs must not become the baseline for final release notes.

## Release flow

```mermaid
flowchart TD
    JUST["just release VERSION<br/>preflight + dispatch + wait"] --> DISPATCH["Release workflow dispatch"]
    UI["GitHub Actions UI"] --> DISPATCH
    TAG["Pre-versioned v* tag push"] --> VERIFY["Verify tag is on main history<br/>and already version-complete"]
    DISPATCH --> META["Resolve version and highest prior stable notes tag"]
    VERIFY --> META
    META --> PATH{"Release path"}
    PATH -- "canary dispatch" --> CANARY["Use dispatch SHA<br/>do not update main"]
    PATH -- "non-canary dispatch" --> BUMP["Run release-version.sh"]
    BUMP --> VERSION_COMMIT["Commit tracked version surface<br/>fast-forward main"]
    PATH -- "tag push" --> TAG_SOURCE["Use validated tag source"]
    CANARY --> BUILD["Build, compose, and smoke artifact matrix"]
    VERSION_COMMIT --> BUILD
    TAG_SOURCE --> BUILD
    BUILD --> PUBLISHABLE{"Canary?"}
    PUBLISHABLE -- "yes" --> CANARY_DONE["Stop without tag or publication"]
    PUBLISHABLE -- "no" --> TAG_PATH{"Entry path"}
    TAG_PATH -- "dispatch" --> PREPARE_TAG["Add generated SDK resources<br/>create and push immutable tag"]
    TAG_PATH -- "tag push" --> EXISTING_TAG["Use existing immutable tag"]
    PREPARE_TAG --> RELEASE["Publish GitHub release<br/>notes compare from prior stable tag"]
    EXISTING_TAG --> RELEASE
    RELEASE --> KIND{"Prerelease?"}
    KIND -- "yes" --> RC_DONE["Stop after GitHub prerelease"]
    KIND -- "no" --> DOWNSTREAM["Publish crates and dispatch<br/>packages, images, and npm"]
```

## What changed

- Bring the existing tracked source, crate, SDK, lockfile, and documentation versions from `0.72.1` to the newest published release, `0.76.0-rc1`.
- Make the Release workflow the canonical version-mutation path.
- For non-canary dispatches, run `scripts/release-version.sh`, create a linear release-source commit when needed, and fast-forward `main` before builds begin.
- Make `just release` a preflight/dispatch/wait wrapper around that same workflow instead of a second version-bump implementation.
- Capture the dispatched workflow run URL separately, extract its numeric run ID, and use SHA-correlated lookup only as a fallback.
- Validate tag-push releases are reachable from `main` and already contain the complete version update.
- Keep canary releases read-only and non-publishing.
- Generate release notes against the highest stable SemVer tag below the target, explicitly excluding prerelease tags.
- Preserve final newlines when the canonical version script rewrites JSON package metadata, and emit rustfmt-compliant known-version entries.
- Update release documentation and repository-consistency tests for the unified contract.

For example, both `v0.76.0-rc1` and `v0.76.0` compare against `v0.75.1`. The final release therefore retains the complete release-train notes and adds any changes made after the RC.

## Validation

- `just ci-validate`
  - 471 tests passed
  - 7 tests skipped
  - release target, crate-list, and publish-chain consistency checks passed
- `just ci-shellcheck scripts/release.sh`
- `just ci-shellcheck scripts/release-version.sh`
- Canonical version-script idempotence and version-only diff audit passed
- Simulated next-release (`v0.76.0-rc2`) bump passed `cargo fmt --all -- --check`, `git diff --check`, and JSON newline checks
- Focused release workflow, dispatch-run selection, and stable-tag selection tests passed
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Release workflows synchronize versions before building and publish from a verified source.
  - Generated release notes compare against the latest applicable stable release.
  - Canary releases validate changes without modifying the main branch or publishing.
  - Updated the release to version **0.76.0-rc1** across packages and SDKs.

- **Bug Fixes**
  - Releases reject invalid, version-drifted, or older target versions.

- **Documentation**
  - Clarified workflow publishing, tag requirements, canary behavior, and unsupported manual releases.

- **Tests**
  - Added coverage for release synchronization, note comparisons, permissions, and publishing inputs.


michaelneale:
Reviewed the release-path unification. The core design is right: one version authority (`release-version.sh`), one publisher (this workflow), and all three entry points converge. Tag-push drift now hard-fails instead of publishing mismatched binaries, and `previous_tag` pinning fixes the empty-final-notes problem. Checked `softprops/action-gh-release@v2.6.2` — `previous_tag` maps to `previous_tag_name` (`src/github.ts:515-533`) and an empty value is coerced to `undefined` (`src/util.ts:118`), so a first-ever release with no stable predecessor degrades correctly rather than erroring.

Three things worth a look before merge.

**1. The version commit lands on `main` before anything is built or smoked.** `metadata` pushes `$source_sha:refs/heads/main` at line 185, and the whole build/compose/smoke matrix runs afterwards. Any failure downstream leaves `main` claiming a version that has no release. The old failure mode was "main behind the release"; the new one is "main ahead of reality". Recovery is also constrained: the `main` ruleset (id 20090642) has `non_fast_forward` + `required_linear_history` and `bypass_actors: []`, so unwinding means a forward revert commit, not a reset. I think this is the better tradeoff — a bumped-but-unreleased `main` is self-correcting on the next release, whereas drift was silent — but it deserves an explicit sentence in `RELEASE.md` so the next person who hits a failed release knows the state `main` is in and that a revert (not a reset) is the fix.

**2. Nothing validates the bump before it is pushed to `main`.** The step checks only for unstaged leftovers. It does not check that the result is formatting-clean. That is exactly the class of bug this PR already had to fix — the `update_known_mesh_versions` perl change (`release-version.sh:162`) exists because the previous regex produced a non-rustfmt-compliant entry in `built_in_schema.rs`. Under the old flow that surfaced locally in front of a human; under the new flow it lands on `main` unattended and breaks `main_quality` for everyone. The job already installs `dtolnay/rust-toolchain`, so the guard is nearly free — before the commit:

```bash
cargo fmt --all -- --check
git diff --check
```

Fail the release rather than push a commit that red-lights `main`.

**3. `git diff --quiet` on the tag-push path can reject for reasons unrelated to versions.** `release-version.sh` calls `refresh_cargo_lock_versions` (`cargo metadata`, line 200-215), which rewrites `Cargo.lock` for any staleness, not just version staleness. A lockfile that drifted for an unrelated reason produces "tag does not contain its complete version update" with a `git status` that shows only `Cargo.lock`. The message is misleading in that case. Worth either narrowing the comparison to the version-bearing paths or extending the error text to say a stale lockfile also trips it.

Smaller notes, none blocking:

- The `metadata` job is now on the critical path for node + rust toolchain setup and a `cargo metadata` run, and every other job blocks on it. Fine, just slower to first build.
- `ensure_target_version_not_older` relaxing `>` to `>=` is correct now that `main` reaches the released version — re-dispatch of the same version is still blocked by "Reject an existing manual release tag" (`release.yml:133-140`), so the guard hasn't gone soft.
- The push at line 185 is deliberately not forced, so a PR landing in the race window rejects the push and fails the release instead of clobbering. Right call; a comment saying so would stop someone "fixing" it with `--force` later.
- `publish` checks out `source_sha` and re-runs `release-version.sh` (line 1767) as a no-op before building the detached tag commit with generated SDK resources. Intended as I read it, but the no-op re-run is non-obvious — one comment would save the next reader the trace.

Verified all 90-odd checks are SUCCESS at the current head. I have not exercised a real release; everything above is from reading the workflow, the script, the action source, and the repo rulesets.

michaelneale (approved):

ndizazzo:
@coderabbitai what do you think about this?

> Nothing validates the bump before it is pushed to main. The step checks only for unstaged leftovers. It does not check that the result is formatting-clean. That is exactly the class of bug this PR already had to fix — the update_known_mesh_versions perl change (release-version.sh:162) exists because the previous regex produced a non-rustfmt-compliant entry in built_in_schema.rs. Under the old flow that surfaced locally in front of a human; under the new flow it lands on main unattended and breaks main_quality for everyone. The job already installs dtolnay/rust-toolchain, so the guard is nearly free — before the commit:

> `cargo fmt --all -- --check`
> `git diff --check`

> Fail the release rather than push a commit that red-lights main.

PR #1354, opened 2026-08-17
Treat max_tokens as a ceiling, not a context reservation

### LINKED ISSUE #1350: Routing over-reserves max_completion_tokens relative to Skippy admission
## Summary

The OpenAI front-door context-fit filter reserves the request's **full** `max_completion_tokens`, while Skippy admission reserves only bounded decode headroom (`min(max_tokens, 512)`). A large client-side completion ceiling can therefore make routing reject a request with 503 even though the runtime would admit the actual prompt comfortably.

This is especially visible with Buzz Agent Desktop 0.5.14, whose default `max_completion_tokens` is 65,536 when `BUZZ_AGENT_MAX_OUTPUT_TOKENS` is unset.

## Reproduction / arithmetic

Observed routing error: `74,703 required tokens` against a target advertising 65,536 context.

The routing estimator in `crates/mesh-llm-host-runtime/src/network/openai/routing_rank.rs` computes:

```text
prompt_tokens = ceil(request_body_bytes / 4)
required_tokens = prompt_tokens + completion_tokens + margin
margin = clamp(ceil((prompt_tokens + completion_tokens) / 4), 16, 256)
```

With `max_completion_tokens = 65,536`:

```text
prompt estimate = 74,703 - 65,536 - 256 = 8,911 tokens
request body ~= 35.6 KB
```

So 65,536 of the 74,703 estimate (88%) is completion reservation, not measured prompt context. `reorder_candidates_by_context_and_throughput` then removes every target whose advertised context is below 74,703, and the request returns 503 before reaching Skippy.

For the same request, Skippy admission in `crates/skippy-server/src/frontend/admission.rs` uses:

```text
decode_headroom = min(max_tokens, DECODE_BATCH_HEADROOM_TOKENS)
DECODE_BATCH_HEADROOM_TOKENS = 512
```

That gives approximately:

```text
8,911 prompt + 512 decode headroom = 9,423 tokens
```

The front door and runtime thus disagree by roughly 8x on context needed for this request.

## Impact

- Valid requests can be rejected before tokenization, compaction, or runtime admission.
- Increasing configured context (for example 65,536 -> 98,304) hides the symptom while allocating more KV for a completion reservation the workload does not use.
- Backend compaction cannot help because the rejection occurs in host-runtime routing first.
- Behavior depends heavily on client defaults rather than actual prompt growth or generated output.

## Expected behavior

Routing context-fit accounting and backend admission should use a compatible completion-headroom policy, or routing should distinguish the API output ceiling from context that must be reserved up front. The fix should preserve safe target selection for genuinely oversized prompts while avoiding false 503s from extreme `max_completion_tokens` values.

## Suggested acceptance criteria

- A ~35.6 KB chat request with `max_completion_tokens = 65,536` is not rejected solely because `prompt_bytes/4 + 65,536` exceeds a 65,536-context target when the backend's bounded decode-headroom policy can admit it.
- Genuinely oversized prompts are still excluded from undersized targets.
- Routing tests cover `max_completion_tokens`, `max_tokens`, and absent completion limits.
- The chosen accounting rule is shared with, or explicitly documented against, Skippy admission so the two layers cannot drift silently.

## Workaround

Set a realistic client limit explicitly, e.g. `BUZZ_AGENT_MAX_OUTPUT_TOKENS=8192` (or 4096 for shorter agents). For the observed request, 8,192 reduces the routing estimate to about 17,359 tokens. This is a workaround, not a fix for cross-layer accounting disagreement.

michaelneale:
A request that Skippy would serve comfortably was returning `503` before it reached any backend, because routing and the runtime disagreed by 128x about how much context a completion limit needs.

`max_completion_tokens` is a *ceiling*, but two places treated it as a *reservation*. Buzz Agent Desktop 0.5.14 defaults it to 65,536 when `BUZZ_AGENT_MAX_OUTPUT_TOKENS` is unset — a client saying "no limit", not a request for 65,536 tokens. A ~9k-token prompt therefore "needed" 74,703 tokens, every 65,536-context target was filtered out, and the caller got a 503.

Skippy admission never reserves that: `min(max_tokens, DECODE_BATCH_HEADROOM_TOKENS)` = 512.

## Changes

**`routing_rank.rs` — bound the completion term.** `request_budget_tokens_from_parts` now applies the same bound admission does. The prompt term is untouched, so a genuinely oversized prompt is still excluded from an undersized target — #753's goal is preserved.

**`generation/queue.rs` — `Explicit` clamps instead of rejecting.** Without this the routing fix alone only moves the failure from a 503 to a 400: the request escapes routing and then trips `prompt + 65,536 > ctx_size` at the backend. `Explicit` now clamps to remaining context like the adjacent `Default` arm already did, and generation stops with `finish_reason: "length"`. A prompt that overflows the window on its own is still an error.

**One shared constant.** `DECODE_BATCH_HEADROOM_TOKENS` is exported from `skippy-server` and consumed by host-runtime, so the two layers cannot drift into two hand-maintained numbers. A test asserts the routing estimate for an unbounded ceiling equals the estimate computed from the exported headroom.

`ensure_context_capacity` had no remaining callers after this and is removed; `context_budget_completion_tokens` covers both arms.

## Behaviour change, called out explicitly

This reverses a deliberate, test-locked decision. `explicit_max_tokens_still_errors_when_too_large_for_ctx` carried the comment *"Client-asserted max_tokens that won't fit is still a hard error. The clamping behavior applies only to the server-picked default."* That test is inverted here.

The argument for inverting it: `max_tokens` is an upper bound in the OpenAI API, clients send large values to mean "unlimited", and refusing them costs a servable request. The argument against: a client that genuinely wants 65,536 tokens now silently gets fewer. I think the former dominates in practice, but it is a product call and a reviewer should push back if they disagree — the routing half of this PR stands on its own if the `queue.rs` half is rejected.

## Regression provenance

The 503 is a regression from `c0f8990b` ("Reject known-too-small context targets", #753, 2026-05-30), which changed the no-adequate-target case from "fall back to the full candidate list and try anyway" to `return Vec::new()`. The over-reservation arithmetic predates it — it dates to `5454baac`, the Skippy landing — but was harmless while it only affected *ranking*. #753 made it a filter, and the overcount became load-bearing.

Related: `b4c36ef3` (#581) hit the same class of complaint on the Skippy side and fixed it by splitting `Default` off from `Explicit` and clamping only `Default`. This PR extends that reasoning to `Explicit`.

## Validation

- `cargo test -p skippy-server --lib` — 445 passed
- `cargo test -p mesh-llm-host-runtime --lib` — 2493 passed, 8 ignored
- `cargo clippy -p skippy-server -p mesh-llm-host-runtime --all-targets -- -D warnings` — clean
- `cargo check --workspace --all-targets` — clean

Both new assertions were verified to **fail** against the pre-fix bodies, reconstructed in place rather than by stash: routing 3 of 8 budget tests fail; `explicit_max_tokens_clamps_to_remaining_context` fails with `context_length_exceeded`.

Not exercised: no live serve. The end-to-end claim (503 becomes a served 200 finishing at `length`) is inferred from the two unit-tested expressions, not observed.

Closes #1350.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token budgeting for completion requests.
  * Large requested completion limits are now safely capped to available context capacity.
  * Prompts that exceed the context window continue to return clear errors.
  * Requests without explicit completion limits now receive consistent budgeting.

* **Tests**
  * Added coverage for oversized limits, alternate limit fields, prompt preservation, headroom handling, and missing limits.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/routing_rank.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/routing_rank.rs
@@ -4,6 +4,11 @@ use crate::network::router;
 
 pub(super) const REQUEST_TOKEN_MARGIN: u32 = 256;
 
+/// Decode headroom Skippy admission reserves, as a `u32` for token arithmetic.
+fn decode_headroom_tokens() -> u32 {
+    saturating_u32(skippy_server::DECODE_BATCH_HEADROOM_TOKENS)
+}
+
 fn saturating_u32(value: usize) -> u32 {
     value.try_into().unwrap_or(u32::MAX)
 }
@@ -27,6 +32,14 @@ fn request_budget_tokens(body: &serde_json::Value) -> Option<u32> {
     request_budget_tokens_from_parts(serialized.len(), completion_tokens)
 }
 
+/// Context a request needs from a target, in tokens.
+///
+/// The completion term is deliberately *not* the client's `max_tokens`. That
+/// value is an upper bound, and clients commonly send an enormous one to mean
+/// "no limit" (Buzz Agent Desktop defaults to 65,536). Skippy admission only
+/// ever reserves `min(max_tokens, DECODE_BATCH_HEADROOM_TOKENS)`, so counting
+/// the full ceiling here made routing refuse requests the server would have
+/// served — issue #1350. Both layers now apply the same bound.
 pub(crate) fn request_budget_tokens_from_parts(
     body_len_bytes: usize,
     completion_tokens: Option<u32>,
@@ -35,7 +48,7 @@ pub(crate) fn request_budget_tokens_from_parts(
         return None;
     }
     let prompt_tokens = ceil_div_u32(saturating_u32(body_len_bytes), 4);
-    let completion_tokens = completion_tokens.unwrap_or(0);
+    let completion_tokens = completion_tokens.unwrap_or(0).min(decode_headroom_tokens());
     let requested_tokens = prompt_tokens.saturating_add(completion_tokens);
     Some(
         prompt_tokens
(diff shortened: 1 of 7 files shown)


michaelneale:
**Approved by Mic. Reviewed and verified independently; live-tested by MeshBuilder.**

Posting on his behalf, with the provenance separated so a later reader knows which claim rests on what.

## Verified by me, at PR head `28ca6182`

Read at `28ca6182`, which contains current main `93d6f426`.

The 128x disagreement is real and on the executed path:

- Admission reserves `min(max_tokens, DECODE_BATCH_HEADROOM_TOKENS)` — `frontend/admission.rs:144-150`
- Routing counted the full ceiling — pre-fix `completion_tokens.unwrap_or(0)` at `routing_rank.rs:51`

I recomputed the arithmetic independently rather than trusting the PR body: a 35,644-byte body gives 8,911 prompt tokens; old budget `8911 + 65536 + margin` = **74,703**, new budget = **9,679**. Targets then fail `value >= required_tokens` at `routing_rank.rs:126`, the candidate list empties, and `transport.rs:1432-1438` sends the 503. I traced through to the actual 503 send rather than stopping at the filter.

Gates reproduce in a worktree at that exact SHA:

- `cargo test -p skippy-server --lib` — 445 passed
- `cargo test -p mesh-llm-host-runtime --lib` — 2493 passed, 8 ignored
- `cargo clippy -p skippy-server -p mesh-llm-host-runtime --all-targets -- -D warnings` — clean

CI is green at `28ca6182` and the PR is `MERGEABLE`.

## Verified live by MeshBuilder (not by me)

Same config `ctx_size = 65536`, same port, same 45,233-byte request with `max_completion_tokens: 65536`, only the binary changed:

| | baseline `93d6f426` | fixed `28ca6182` |
|---|---|---|
| ~10k-token prompt | HTTP **503** in 0.01s | HTTP **200** in 53.1s |
| error | `no context-compatible target ... can fit approximately 77101 tokens` | `finish_reason: stop` |

That closes the "no live serve" caveat the PR body flagged. The clamp arm was exercised too: a 44,471-token prompt with the same 65,536 ceiling — arithmetically unsatisfiable in a 65,536 window — returned 200 rather than the 400 it would have produced with only the routing half.

## The behaviour reversal is an accepted product call

Mic's call, explicitly: `max_tokens` is a ceiling, so clamping is correct and asking for a huge completion is unusual anyway. `explicit_max_tokens_still_errors_when_too_large_for_ctx` is intentionally inverted here.

## One non-blocking note

`Explicit` now resolves to `0` instead of erroring when the prompt exactly fills the window — I probed it rather than inferring: `resolve(65_536, 65_536)` returns `Ok(0)`, and `max_tokens == 0` is an explicitly handled state in the decode path (`local_generation/token_generation.rs:182`), so the caller gets a clean but empty 200 where they previously got a 400. The zero case already existed for the `Default` and `ContextBudget` arms on main; this PR makes it newly reachable from client-supplied limits. Reviewed and judged not worth holding the merge — recorded so it is not rediscovered as a surprise.

PR #1356, opened 2026-08-17
Reject non-protobuf content types on OTLP/HTTP ingestion

michaelneale:
## What this fixes

The bundled `metrics-server` exposes OTLP/HTTP ingestion endpoints (`/v1/traces`, `/v1/metrics`, `/v1/logs`) on a loopback listener. These handlers accepted a raw request body and protobuf-decoded it **without checking `Content-Type`, `Origin`, or any credential**.

Loopback is not an authentication boundary for browsers. A hostile web page you happen to have open can issue a "simple" cross-origin `POST` with a CORS-safelisted content type (`text/plain`, `application/x-www-form-urlencoded`, `multipart/form-data`) and **no preflight** — the browser still delivers the body even though script can't read the response. That let a drive-by page POST valid protobuf bytes and inject or poison spans, metrics, and runs (including a chosen `run_id`) in your local metrics store, corrupting benchmark reports and consuming disk.

## The fix

Require `Content-Type: application/x-protobuf` on the three OTLP/HTTP export handlers. Every conforming OTLP/HTTP exporter already sends this, so legitimate ingestion is unaffected. Because `application/x-protobuf` is **not** in the CORS-safelisted set, a cross-origin browser request must now pass a preflight the attacker page cannot satisfy — closing the drive-by path. Requests with a missing or non-protobuf content type get `415 Unsupported Media Type`, and content-type parameters (e.g. `; charset=utf-8`) are tolerated.

`create_run` (`POST /v1/runs`) was already safe: it uses axum's `Json` extractor, which enforces `application/json` and is therefore not CORS-safelisted.

## Validation

- New regression tests: drive-by `text/plain` protobuf is rejected with `415` and creates no run; missing content type is rejected; `application/x-protobuf` (with and without a charset parameter) is still accepted.
- `cargo test -p metrics-server --lib` → 9 passed.
- `cargo clippy -p metrics-server --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean.

Fixes project-loupe/audit-mesh-llm#886.

## Summary by CodeRabbit

* **Bug Fixes**
  * OTLP/HTTP trace, metric, and log ingestion now validates the `Content-Type` header before processing requests.
  * Requests using `application/x-protobuf`, including charset parameters, are accepted.
  * Requests with missing or unsupported content types now receive a clear `415 Unsupported Media Type` response.
  * HTTP errors now return their appropriate status codes instead of defaulting to `500 Internal Server Error`.

### DIFF
--- a/crates/metrics-server/src/api.rs
+++ b/crates/metrics-server/src/api.rs
@@ -3,7 +3,7 @@ use axum::{
     Json,
     body::Bytes,
     extract::{Path, State},
-    http::header,
+    http::{HeaderMap, StatusCode, header},
     response::IntoResponse,
 };
 use opentelemetry_proto::tonic::collector::{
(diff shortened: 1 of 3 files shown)


i386 (approved):
Reviewed the current head: the OTLP protobuf content-type check is case-insensitive, accepts parameters, returns 415 for unsupported/missing media types, and the endpoint tests cover the rejection path. CI is green. Approving.

PR #1357, opened 2026-08-17
Cap remotely advertised model lists to protect the console from a hostile peer

michaelneale:
## What this fixes

A mesh peer advertises the models it serves via gossip (`serving_models`, `hosted_models`, and the other model-name lists). These lists flow untouched through `/api/status` into the web console, where each name is sorted (locale-aware, `O(n log n)`) and rendered as its own DOM element.

Nothing bounded the list length at ingest: the only limit was the generic 8 MiB control-frame ceiling, which still permits hundreds of thousands of short names. A malicious mesh participant could therefore advertise an oversized model list — and repeatedly re-advertise it — to exhaust memory and freeze or crash another operator's dashboard browser, denying access to the local management UI. This is a genuinely remote vector: it needs only a hostile peer on the mesh, no co-resident user or browser trust assumption.

## The fix

Sanitize peer-supplied model-name lists at the trust boundary — `proto_ann_to_local`, where an inbound wire protobuf `PeerAnnouncement` is converted into the local struct. Each remotely-advertised list (`serving_models`, `hosted_models`, `catalog_models`, `requested_models`, `explicit_model_interests`) now:

- drops any entry longer than 512 bytes, and
- keeps at most 256 names.

A real node serves a handful of models, so these ceilings sit far above any legitimate value and no honest peer is affected. Capping at the trust boundary also protects every downstream consumer (routing, `/api/status`, the console) at once, rather than relying on the UI to defend itself.

## Protocol

No wire-format change. This only bounds how much of an existing, additive gossip field the receiver retains, so mixed-version meshes are unaffected: older and newer nodes still exchange the same fields, and a node that advertises a normal number of models sees identical behavior. An over-advertising peer simply has its list truncated locally.

## Validation

- New regression tests in `protocol::convert`: an announcement carrying 10,000 names per list is capped to 256 each; an over-length model name is dropped while a valid sibling is kept.
- `cargo test -p mesh-llm-host-runtime --lib` → 2495 passed, 0 failed.
- `cargo check -p mesh-llm-host-runtime` clean; `cargo fmt --all --check` clean.

Fixes project-loupe/audit-mesh-llm#860.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards for remote model-name lists.
  * Oversized names are ignored, and excessively long lists are capped to maintain reliable processing.
  * Applies across catalog, serving, requested, interest, and hosted-model listings.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/protocol/convert.rs
+++ b/crates/mesh-llm-host-runtime/src/protocol/convert.rs
@@ -715,6 +715,36 @@ pub(crate) fn build_gossip_frame(
     }
 }
 
+/// Upper bound on the number of model-name entries accepted from a single
+/// remote peer announcement.
+///
+/// Peer gossip carries `serving_models` / `hosted_models` (and the other
+/// model-name lists) straight through `/api/status` into the console, where
+/// they are sorted and rendered one DOM element per entry. The wire frame is
+/// only bounded by the generic 8 MiB control-frame ceiling, which still allows
+/// hundreds of thousands of short names. Capping the list length at the trust
+/// boundary stops a hostile peer from freezing another operator's dashboard
+/// with an oversized (or rapidly re-advertised) model list (#860). A real node
+/// serves at most a handful of models, so this ceiling is far above any
+/// legitimate value.
+const MAX_REMOTE_MODEL_LIST_LEN: usize = 256;
+
+/// Upper bound on the byte length of a single remotely-advertised model name.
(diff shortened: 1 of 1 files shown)


i386 (approved):
Reviewed the current head: the remote model-name cap is applied consistently to all five relevant fields, uses a byte bound appropriate for serialized data, and CI is green. The remaining CodeRabbit suggestion is optional test expansion, not a correctness blocker. Approving.

PR #1359, opened 2026-08-17
Release CUDA builds no longer crash the host compiler

michaelneale:
The `v0.76.0-rc2` release run could not produce Linux x86_64 CUDA native runtimes. Both lanes failed, and failed again on a clean re-run, so no CUDA bundle could be published. With this change those lanes build to completion.

## What was happening

Three attempts, four different crash sites, none of them a diagnostic about our code:

| attempt | row | translation unit | crash |
|---|---|---|---|
| first | CUDA 13 | `concat.cu` | gcc-13 internal compiler error, `instantiate_class_template`, inside `/usr/include/c++/13/tuple` |
| first | CUDA 12 | `convert.cu` | `cicc terminated (signal: 11)` |
| re-run | CUDA 13 | `fattn-mma-f16-instance-ncols1_8-ncols2_2.cu` | `cicc terminated (signal: 11)` |
| re-run | CUDA 12 | `fwht.cu` | `cicc terminated (signal: 11)` |

A crash that moves to a different translation unit and a different compiler stage on every attempt is resource exhaustion, not a source regression. Two further facts point the same way: the last fully green run of these lanes peaked at 3 concurrent compile starts per second while every failing run reached 5-7, and all four failures happened on a cold backend cache, so nothing was restored and all 458 objects had to compile.

## The defect

`detect_jobs` in `scripts/build-llama.sh` sized `cmake --build --parallel` from `nproc` alone. These jobs run in a container on a self-hosted many-core GPU runner, where `nproc` reports the *host* CPU count, and each CUDA translation unit runs one nvcc frontend per target architecture -- ten of them for the CUDA 13 row (`75;80;86;87;89;90;100;103;120;121`). Parallelism scaled with cores and ignored memory completely, so the build launched far more concurrent frontends than it had memory for and the host compiler died.

## The fix

Two layers, because either alone is insufficient:

1. `detect_jobs` now sizes parallelism from the memory budget as well as the CPU count, preferring the cgroup limit that would actually kill the compiler (v2 `memory.max`, then v1 `memory.limit_in_bytes`) and falling back to kernel-reported total memory. It assumes 3 GiB per CUDA/ROCm compile and 1 GiB per CPU compile, never returns zero, keeps the CPU count when memory is plentiful, and logs the cap so a slower build stays explainable.
2. The two x86_64 CUDA release lanes pin `CMAKE_BUILD_PARALLEL_LEVEL: "4"`. A container without a cgroup memory limit still sees whole-host memory, so autodetection alone would not bind on this pool. The pin is the guarantee; the detection is the safety net everywhere else.

`--print-jobs` is added so the budget is observable and testable without running a build:

```
$ MESH_LLM_LLAMA_BUILD_MEMORY_BYTES=17179869184 LLAMA_STAGE_BACKEND=cuda ./scripts/build-llama.sh --print-jobs
capping llama.cpp build parallelism at 5 of 18 CPUs
  backend:        cuda
  memory budget:  16384 MiB
  assumed per job: 3072 MiB
5
```

## Validation

- `scripts/tests/test_build_llama_parallelism.py` -- 8 new tests: memory caps below CPU count, CPU backend gets a wider budget than CUDA, CPU count wins with plentiful memory, never zero jobs, real probe stays within bounds, explicit override honored verbatim, malformed budget rejected with exit 2.
- Confirmed the tests fail for the right reason: restoring the pre-change `scripts/build-llama.sh` from `git show HEAD:` turns all 8 red, and restoring the fix turns them green.
- `shellcheck scripts/build-llama.sh` clean. `bash -n` clean.
- `python3 -m unittest discover -s scripts/tests` -- 451 tests. The 5 failures (`test_plan_ci`, and 2 loader errors from a missing local `yaml` module) reproduce identically on unmodified `main` at `53b70ad9`, so they are pre-existing and unrelated.

## Not verified

I did not prove memory exhaustion by measurement -- no OOM-killer line appears in any log and the runner cgroup limits are not visible through the API. The evidence is the non-deterministic crash site, the crash stage inside libstdc++ template instantiation, and the concurrency delta against the last green run. The real proof is the next release run on these lanes.

I also did not change the aarch64 CUDA or ROCm lanes, which were green throughout; they pick up the detection improvement but no pin.

## Summary by CodeRabbit

* **Chores**
  * Improved release build performance and stability by automatically adjusting compilation parallelism based on available CPU and memory resources.
  * Added support for explicitly controlling build parallelism and displaying the calculated job count.
  * Added safeguards for invalid memory-budget settings and resource-constrained environments.
  * Limited parallel CUDA compilation to help reduce resource contention during Linux releases.

* **Tests**
  * Added coverage for parallelism limits, backend-specific resource usage, overrides, diagnostics, and minimum job handling.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -630,6 +630,15 @@ jobs:
       LLAMA_STAGE_BACKEND: cuda
       LLAMA_STAGE_CUDA_ARCHITECTURES: ${{ matrix.cuda_architectures }}
       LLAMA_STAGE_BUILD_DIR: .deps/llama-build/build-stage-abi-dynamic-cuda-sm${{ matrix.cuda_architectures_cache }}
+      # These are the memory-heaviest compiles in the release graph: every CUDA
+      # translation unit runs one nvcc frontend per target architecture, and the
+      # CUDA 13 row targets ten. Container jobs on the self-hosted GPU pool see
+      # the host CPU count with no cgroup memory limit, so autodetected
+      # parallelism overcommits memory and the host compiler dies mid-build with
+      # no diagnostic about our source. Observed: failing runs reached 5-7
+      # concurrent compile starts per second, while the last fully green run of
+      # these lanes stayed at 3.
+      CMAKE_BUILD_PARALLEL_LEVEL: "4"
       MESH_NATIVE_RUNTIME_TARGET: x86_64-unknown-linux-gnu
       MESH_LLM_CUDA_TOOLKIT_MAJOR: ${{ matrix.cuda_major }}
       SCCACHE_GHA_ENABLED: "false"
--- a/scripts/build-llama.sh
+++ b/scripts/build-llama.sh
@@ -11,12 +11,17 @@ LLAMA_BUILD_ROOT="${MESH_LLM_LLAMA_BUILD_ROOT:-$ROOT/.deps/llama-build}"
 LLAMA_BACKEND="${LLAMA_STAGE_BACKEND:-${SKIPPY_LLAMA_BACKEND:-${LLAMA_BACKEND:-cpu}}}"
 LLAMA_LINK_MODE="${LLAMA_STAGE_LINK_MODE:-${SKIPPY_LLAMA_LINK_MODE:-static}}"
 PRINT_BUILD_DIR=0
+PRINT_JOBS=0
 REQUIRE_EXISTING=0
 
 while [[ "$#" -gt 0 ]]; do
   case "$1" in
     --print-build-dir)
       PRINT_BUILD_DIR=1
+      shift
+      ;;
+    --print-jobs)
+      PRINT_JOBS=1
       shift
       ;;
     --require-existing)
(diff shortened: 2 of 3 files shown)


i386 (changes requested):
Code review: changes requested

The new test helper is not fully isolated or portable:

- When `memory_bytes` is `None`, it does not remove an inherited `MESH_LLM_LLAMA_BUILD_MEMORY_BYTES`. A developer/CI environment override can therefore change the budget under test and make `test_detects_capacity_without_an_explicit_budget` exercise the wrong path. Pop the variable before the conditional assignment, as is already done for `CMAKE_BUILD_PARALLEL_LEVEL`.
- `test_cuda_parallelism_capped_by_memory_not_cpu_count` hard-codes `2 < cpu_count()`, so the test fails on a legitimate two-core runner even though the implementation is correct. Control/mock the CPU count or assert against a deterministic value.

Please fix these test-isolation and portability issues before approval.

Issue #1331, opened 2026-08-14
spec: expose OpenAI request lifecycle to plugins


ndizazzo:
> Status: **Draft specification**
>
> Related umbrella: #1233. This issue specifies the plugin-platform prerequisite exposed by the [Capsule Emit sidecar experiment](https://github.com/action-state-group/capsule-emit-mesh) and its [findings on #1233](https://github.com/Mesh-LLM/mesh-llm/issues/1233#issuecomment-5287921908).

## Summary

Add an opt-in, permissioned OpenAI exchange lifecycle contract for installable mesh-llm plugins.

A plugin that declares this contract must be able to:

1. inspect the original OpenAI request envelope before model dispatch;
2. return a bounded pre-dispatch policy decision;
3. observe the effective request and selected serving route before execution;
4. observe the terminal response, error, denial, or cancellation;
5. correlate every phase without becoming an HTTP reverse proxy.

The contract must cover chat completions, legacy completions, and Responses, including streaming requests. It must preserve ordinary OpenAI streaming rather than forcing the host to buffer a response.

## Motivation

The current plugin architecture is primarily provider-oriented: plugins contribute operations, HTTP/MCP surfaces, external inference endpoints, and mesh capabilities. It does not let an installable plugin observe or make a policy decision about traffic arriving at mesh-llm's own OpenAI API.

The native serving plugin ABI exposes generation token lifecycle data, but not the original OpenAI request or the client-visible response. It therefore cannot compute a receipt over the API request/response pair.

[OpenAiHookPolicy](https://github.com/Mesh-LLM/mesh-llm/blob/main/crates/openai-frontend/src/hooks.rs) is a useful in-process seam, but it is compiled into the backend wrapper, currently invokes only the before-chat hook, and is not available to an installable out-of-process plugin.

The Capsule Emit experiment consequently had to run as a reverse-proxy sidecar. That proved the record format but introduced a second port, changed the request topology, buffered/re-synthesized SSE, and could only describe itself honestly as an external collector. Other policy, audit, DLP, guardrail, routing, and compliance plugins will hit the same limitation.

## Goals

- Make the request envelope available to explicitly authorized plugins.
- Allow an authorized plugin to allow, abstain, or deny before dispatch.
- Expose both what the client submitted and what mesh-llm actually dispatched.
- Expose what mesh-llm actually returned to the client.
- Preserve exact byte-level commitments alongside parsed OpenAI semantics.
[...]

Issue #1233, opened 2026-08-10
Explore proof of inference and model-serving attestation


i386:
## Context

Steven Mih raised an important discussion about proof of inference: how can a client tell whether a node generated a response with the model it advertises, rather than substituting a cheaper quantization or different weights?

Sources:

- [Steven Mih on LinkedIn](https://www.linkedin.com/in/stevenmih/)
- [Discord discussion](https://discord.com/channels/1287729918100246654/1486873561431015454/1536480623475040376)

## Why this matters

A human-readable model name is not a strong identity claim. A node could advertise a full-precision or higher-quality model while serving a cheaper quantization, modified weights, a different tokenizer, or a different runtime configuration.

“Proof of inference” is best understood as three separate claims:

| Claim | Example question | Candidate evidence |
| --- | --- | --- |
| Model identity | What exact model/package does the node claim to serve? | Content-addressed package and artifact digests |
| Execution identity | Did this request actually run with those bytes and this runtime? | Signed request/response receipt, ideally TEE-backed |
| Behavioral identity | Does the output behave like the claimed model? | Statistical fingerprints or reproducible verification |

No single mechanism solves all three.

## Proposed model identity

A model name such as `Qwen3-8B` is too broad. The advertised identity should include an immutable package or manifest digest covering:

- canonical package manifest;
- source model identity and revision;
- quantization/distribution;
- tokenizer identity;
- all selected artifact digests;
- generation/runtime defaults;
- Skippy/llama ABI version;
- accepted equivalence classes, if quantized variants are intentionally allowed.

A useful conceptual identity is:

```
model_package_digest =
  sha256(canonical_manifest + artifact_digests + runtime_abi)
```

Mesh-llm already has useful building blocks here. Layer-package manifests record source identity, per-artifact SHA-256 values, layer coverage, and ABI compatibility. Production references should use immutable package commits/tags rather than moving branches. See the [model package specification](https://github.com/Mesh-LLM/mesh-llm/blob/main/website/src/docs/pages/model-package-spec.md).

A quantized model is not inherently fraudulent. The fraud is advertising an identity outside the declared package or accepted equivalence class. The policy should make that distinction explicit.

## Per-request evidence

A node could produce a receipt covering the request and response:

```
receipt = sign_node_key(
  client_nonce,
  request_digest,
  model_package_digest,
  runtime_digest,
  generation_parameters,
  output_digest,
  timestamp
)
```

Important details:

[...]

PR #1362, opened 2026-08-18
fix(goose): supply a client-side per-request nonce in the generated provider config

StevenMih:
`mesh-llm goose` writes `~/.config/goose/custom_providers/mesh.json`. This adds one key so goose supplies a fresh per-request nonce rather than leaving the intermediary to mint one.

This is the client-side half that **#1233** asks for. Today a recording sidecar in front of the endpoint has to generate its own value and record it as `sidecar_generated_fallback` — a weaker signal, because a value the intermediary minted cannot demonstrate that the *client* contributed freshness. With this key present it records `client_supplied` instead.

The header name is the one already used in the plugin-lifecycle spec draft — #1331 lists `x-capsule-client-nonce` under `sanitized_headers` — so this introduces no new vocabulary.

**Inert until a goose release carrying `nonce_header` support**, and safe on older goose: `DeclarativeProviderConfig` has no `deny_unknown_fields`, so unknown keys are ignored. The two changes can therefore land in either order.

Goose side: aaif-goose/goose#11302.

## Summary by CodeRabbit

* **Enhancements**
  * Goose provider requests now include the `x-capsule-client-nonce` header.
  * Unauthenticated provider configurations continue to be supported.

* **Maintenance**
  * Updated console output validation references to remain accurate after source updates.
  * No user-facing configuration changes are required.


michaelneale (approved):

ndizazzo:
@StevenMih any chance this exists for other harnesses, before we merge?

StevenMih:
@ndizazzo 
Not yet, but cheap. mesh-llm side is one line per harness — opencode, claude, and others write their configs the same way run_goose() does. Say the word and they go in this PR.

Client side, goose is the cost proof: one config field, ~15 lines to install the decorator (aaif-goose/goose#11302). Rest of that diff is fixtures and struct-literal fallout. Any client with a request-builder seam is the same shape; one that can already set a per-request header from config needs no code. Point me at whichever you want next and I'll open it.

Where it goes: nonce buys equivocation resistance, not identity. A client already holding keys signs a commitment instead and gets both — same shape as the delegation chain in #1331, one rung up (per rungs on Trust draft i posted).

ndizazzo:
> opencode, claude, and others write their configs the same way run_goose() does. Say the word and they go in this PR.

@StevenMih wouldn't the harness need some kind of support to read `x-capsule-client-nonce` though? I can't imagine that already exists in each harness, and would need a plugin on their side at the very least

StevenMih:
@ndizazzo Fair, goose had the seam. Checked others and here's a run down:

opencode: provider.<id>.options.headers exists but it's static — resolved at load, so it can't carry a fresh nonce. (The headers-not-reaching-fetch bug, anomalyco/opencode#11789, closed via #11788, so the plumbing's fine — freshness is the separate gap.) Needs the same ~15-line decorator goose took in aaif-goose/goose#11302. I'd open that.

pi: models.json has no headers field at all. Real extension on their side.

claude: no request-level header hook, checked the settings reference — apiKeyHelper is auth-only on a refresh interval, otelHeadersHelper is telemetry-only. That one's an ask to Anthropic, not a patch from us.

But the per-client route is the upgrade path, not the coverage plan. You already own the seam every harness goes through — the localhost endpoint we point them all at. Inject the fresh nonce at ingress, before it reaches the recording sidecar, and every harness is covered today with zero client code — minted one hop later than client-native, still on the user's box. The sidecar already distinguishes client_supplied from sidecar_generated_fallback; add a local_ingress value between them and the record says which tier it actually got instead of overclaiming. If the local process and the sidecar are fused in the deployments you care about, the tier collapses and this is moot — if they're separate hops, it's a small PR here and I'll write it.

Ingress puts everyone at rung 1. Rung 2, where a client signs a commitment, #1331's shape one rung up — needs client-side code no matter what, so the plugins you're picturing are the rung 2 work, not the rung 1 tax.

For this PR: suggest to merge the goose line, it works end to end. And leave out other harnesses till they're real. Then i suggest ingress as its own PR next, opencode/other upstream after that.

i386 (changes requested):
The change itself is fine and consistent with #1397, but it breaks the `no-console-print` ratchet: the baseline pins exact line numbers in `agent_cli.rs`, and the one line added here shifts every later `eprintln!` occurrence (CI contracts job shows lines 910–1136 'approved occurrence is missing or was replaced').

Please regenerate the baseline with `cargo run -p xtask -- repo-consistency no-console-print --regen` and include the updated ratchet file in this PR so the Quality lane goes green.

ndizazzo:
> Please regenerate the baseline with cargo run -p xtask -- repo-consistency no-console-print --regen and include the updated ratchet file in this PR so the Quality lane goes green.

Just wanna note - don't add an exception, we want to forward-support logging through the proper event system so that we can capture telemetry and show these in the new logs UI

StevenMih:
@ndizazzo — you're right, and it's the important half. The header this PR injects (`x-capsule-client-nonce`) only makes goose **send** a fresh per-request nonce; **reading and binding** it is entirely the server side — the openai-frontend lifecycle hooks in #1397/#1437 and the emit-side plugin. That's harness-agnostic by design: the plugin binds whatever request carries the header and doesn't care which harness produced it.

So for opencode / claude / the others: each needs the same trivial config change to **send** the header (they write their provider configs the same way `run_goose()` does), but they do **not** each need their own plugin — the single server-side plugin reads all of them. This PR is the goose half of that; the same one-liner drops into the other harnesses, and I'm happy to add them here or in a follow-up, your call. Nothing downstream has to understand the nonce except the one plugin that binds it.

On the ratchet: it's regenerated (`ad5bb26`, no exception added) so the Quality lane is green — `cargo run -p xtask -- repo-consistency no-console-print` passes clean locally; logging still routes through the event system as you want. (It just hasn't shown in CI because fork PRs aren't getting runs — see the note on #1437.)

i386:
Reviewed current head `ad5bb26d9`. There are no unresolved inline review threads. The outstanding change-request item is addressed: the no-console-print ratchet was regenerated rather than excepted, and `just no-console-print` passes locally (`repo consistency checks passed: no-console-print`). No additional patch was needed.

i386 (approved):
Reviewed current head 58e92998129eea8864b3e095a6a539ae92f5fdde after merging current main. All review threads are resolved and GitHub reports the branch mergeable. Validation: cargo fmt; full mesh-llm-commands suite (226 passed); just ci-validate (734 passed, 7 skipped); and no-console-print.

ndizazzo (approved):

PR #1364, opened 2026-08-18
fix(api): stop leaking plugin endpoint credentials over the management API

michaelneale:
## What this fixes

Plugin endpoint addresses can carry credentials directly in an HTTP URL — either `user:pass@` userinfo or a `?api_key=...` query. Those raw addresses were served verbatim by the management API, which can be bound to a non-loopback interface, so anyone who could reach the management port could read your plugin credentials on an **unauthenticated network request**. After this change, credentials are stripped before any endpoint address leaves the box, and the endpoint listing is no longer network-open at all.

Resolves Loupe audit findings **#237**, **#91**, **#26**, **#319**, and **#87**.

## Changes

- **Redact credentials from endpoint addresses at the network egress boundary.** HTTP(S) endpoint addresses and the health `detail` string are scrubbed of userinfo and sensitive query params before an endpoint summary is serialized by `GET /api/runtime/endpoints` and `GET /api/plugins/endpoints`. (#237 / #91 / #26)
- **Redact the routing-metrics endpoint label.** The routing-metrics snapshot copied the same credential-bearing endpoint address into its target label; HTTP(S) endpoint labels are now redacted there too. (#319)
- **Add `GET /api/runtime/endpoints` to the trusted-local access gate.** Every sibling `/api/runtime/*` route already requires a trusted-local caller; this one had been left off. Even redacted, the endpoint listing should not be network-readable. (#237)
- **Fail closed on an empty invite token in the LAN-details proof gate.** With no configured invite token, the fingerprint/challenge/proof were all reproducible by any LAN caller, so the gate failed open. It now rejects when the expected token is empty. (#87)

## Why redact at egress, not at construction

The endpoint `address` is **not** display-only — the MCP integration (`ExternalMcpEndpoint::from_summary`) reconstructs its transport from the raw address via `plugin_manager.endpoints()`. Redacting at summary construction would break MCP endpoints. So redaction happens at the network boundary (API responses + the metrics snapshot), leaving the in-process value intact for local consumers. Non-HTTP addresses (stdio commands, unix sockets, named pipes) pass through verbatim — they carry no URL credential and must stay usable for transport reconstruction.

## Protocol / compatibility

No wire-protocol or plugin-protocol change. This only redacts fields in management-API JSON responses and the local routing-metrics snapshot, and tightens one access gate. Mixed-version meshes are unaffected. The one behavioral change a caller could observe: `GET /api/runtime/endpoints` now requires a trusted-local caller (consistent with its `/api/runtime/*` siblings), and credential-bearing addresses appear redacted in responses.

## Validation

- `cargo fmt --all --check` — clean.
- `cargo test -p mesh-llm-host-runtime --lib` — **2520 passed, 0 failed**, including 6 new tests:
  - `plugin::types::tests::http_address_userinfo_and_query_are_redacted_for_network`
  - `plugin::types::tests::health_detail_url_credentials_are_redacted_for_network`
  - `plugin::types::tests::non_http_address_is_left_verbatim_for_local_transport`
  - `network::metrics::tests::endpoint_target_label_credentials_are_redacted_in_snapshot`
  - `network::metrics::tests::non_endpoint_target_labels_pass_through_verbatim`
  - `network::discovery::tests::lan_details_proof_rejects_empty_invite_token_fails_closed`
- Access-gate test `sensitive_management_routes_require_trusted_local_access` extended to cover `GET /api/runtime/endpoints`.

## Summary by CodeRabbit

* **Security**
  * Restricted runtime endpoint listings to trusted local access.
  * Redacted credentials, query secrets, and sensitive URLs from endpoint, health, and monitoring details.
  * Improved redaction for HTTP and HTTPS URLs regardless of letter casing.
  * Preserved safe labels for local and non-HTTP transports.
  * Rejected empty or whitespace-only invite tokens during network discovery validation.

* **Bug Fixes**
  * Added coverage for access restrictions, secret redaction, URL handling, and invalid invite tokens.


michaelneale:
Addressed in 9fb078af. Both scheme gates now match `http`/`https` case-insensitively via a shared `logging::policy::is_http_url` helper (also applied to the pre-existing `redact_urls_in_text`, which had the same case-sensitive check). Added uppercase-scheme regression cases for both the plugin-summary and metrics-label paths.

ndizazzo (approved):
@michaelneale  We can likely start exercising some log generation stuff here.... Some minor things would be to call `trim()` on the http URL so `is_http_url` doesn't false negative if it has whitespace. 

There should be some guidance for agents on using the shared facility to add log records if you want to insert that into your fixup prompt

PR #1365, opened 2026-08-18
Fix blank responses from thinking models

michaelneale:
## Problem

Thinking models can appear to return a blank response in Mesh. When `reasoning_format` is omitted, Mesh currently parses generated reasoning and then silently discards it. The model is producing tokens and the GPU is working, but clients receive no visible stream while the model is reasoning.

This differs from llama-server/LM Studio and from harness expectations, where the default is to pass model reasoning through.

## Fix

Default an omitted `reasoning_format` to `auto` instead of Mesh's `hidden` mode. Reasoning is now streamed to clients as `reasoning_content`, so the reproduced Qwen3.8 request no longer appears hung or empty.

Explicitly configured formats such as `hidden` and `none` keep their existing behavior. This does not change whether a model thinks; it fixes whether generated thinking is silently lost.

## Validation

- `cargo fmt --all --check`
- `cargo check -p mesh-llm`
- `cargo test -p skippy-server --lib` (447 passed, 3 ignored)

### DIFF
--- a/crates/skippy-server/src/frontend/tests/request.rs
+++ b/crates/skippy-server/src/frontend/tests/request.rs
@@ -327,11 +327,11 @@ fn canonical_reasoning_overrides_chat_template_thinking() {
     let options =
         chat_template_options(&request, &EmbeddedOpenAiRequestDefaults::default()).unwrap();
     assert_eq!(options.enable_thinking, Some(false));
-    assert_eq!(options.reasoning_format, Some(ChatReasoningFormat::Hidden));
+    assert_eq!(options.reasoning_format, Some(ChatReasoningFormat::Auto));
 }
 
 #[test]
-fn chat_template_options_default_to_hidden_reasoning_parser() {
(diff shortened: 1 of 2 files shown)


crates/skippy-server/src/frontend/tests/request.rs
```diff
@@ -327,11 +327,11 @@ fn canonical_reasoning_overrides_chat_template_thinking() {
     let options =
         chat_template_options(&request, &EmbeddedOpenAiRequestDefaults::default()).unwrap();
     assert_eq!(options.enable_thinking, Some(false));
-    assert_eq!(options.reasoning_format, Some(ChatReasoningFormat::Hidden));
+    assert_eq!(options.reasoning_format, Some(ChatReasoningFormat::Auto));
```
ndizazzo, line 330: big oof, as the kids say

ndizazzo (approved):

PR #1366, opened 2026-08-18
fix(mtp): make native MTP actually work on single-node serving

michaelneale:
Native MTP now actually skips forwards on the single-node decode path. It is still net-negative on Qwen3.8, and the remaining cost is identified but not yet fixed — see "Why this is still slower".

## What was wrong

Single-node (non-staged) serving goes through `local_generation/decode_step.rs`. With `strategy = "mtp"` it drafted a token, ran a full target forward for **one** token, compared, logged the verdict via `verifier.rs`, threw the draft away, and decoded again. Instrumented with `SKIPPY_TELEMETRY_STDERR=1`, 3x300 tokens:

| arm | drafted | accept rate | serial decode forwards | tokens emitted |
|---|---|---|---|---|
| MTP off | 0 | — | 900 | 900 |
| MTP on (before) | 900 | **0.742** | **900** | 900 |

74% of drafts were correct and not one saved a forward. Pure added draft cost (~1.6 ms/token), zero payoff.

The accept-and-skip contract existed in two places, neither reachable here:
- `frontend/native_mtp/verify_window.rs` — staged/split path only (`embedded_generation.rs:1041`).
- `frontend/linear_proposal/execution.rs:170-250` — correct and complete, but proposals come exclusively from `linear_proposal_ingress`, whose only production constructor is behind the hidden `--native-serving-plugin` flag (`local_model_only.rs:190-193`, `commands.rs:515-518`). Zero `linear_proposal` matches in either instrumented server log; that branch was never live.

## What this does

New `local_generation/native_mtp_decode.rs`: one batched target forward over the drafted span, commit every token up to the first mismatch, retire the verify checkpoint on full accept, otherwise trim the speculative suffix. Reuses the existing `classify_native_mtp_verify_window` classifier the staged path uses. Gated on greedy sampling (token equality is only a valid acceptance test there), no active generation hooks, and the same recurrent-checkpoint boundary as the linear-proposal path.

Includes @Jimmy's `ef6cab355` (forwards `resolved_mtp_source()` instead of hardcoded `MtpSource::Disabled`) — without it `strategy = "mtp"` was silently ignored. Jimmy's reasoning fix stays separate in #1365.

**The skip is real:**

| | before | after |
|---|---|---|
| tokens emitted | 900 | 900 |
| serial decode forwards | 900 | **381** |
| batched spans | 0 | 228 |
| forwards saved | 0 | **291** |

456 skippy-server + 74 skippy-runtime tests pass; `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

## Ground-truth A/B vs LM Studio

Interleaved in one session on the same box, same GGUF (`unsloth/Qwen3.8-27B-GGUF` Q4_K_M), same prompt, 300 tok, temp 0, 6 runs per arm:

| arm | median tok/s |
|---|---|
| LM Studio MTP off | 25.09 |
| mesh MTP off | 23.62 |
| LM Studio MTP **on** | **26.49** |
| mesh MTP **on** (this branch) | 16.70 |

LM Studio's own log: `draft acceptance = 0.826 (119/144), mean len 2.20`. So upstream MTP is worth **+5.6%** on medians here, not 2x. Nick's 5090 figure does not reproduce on this box **in either engine**. Note also: run-to-run drift on this machine exceeds the MTP effect (one off-arm drifted to 17-19 tok/s idle), so only interleaved single-session arms are meaningful.

## Why this is still slower — and it is not a Metal batching limit

`skippy_verify_tokens_frame_sampled` calls `skippy_checkpoint_verify_window` before every batched verify (`verification.cpp:141`). That early-returns for plain KV, but `skippy_memory_needs_verify_checkpoint` (`activation.cpp:61-65`) returns true for `llama_memory_hybrid`/`hybrid_iswa`/`recurrent`.

**Qwen3.8-27B is hybrid recurrent** — confirmed from the GGUF header, not assumed: `general.architecture = qwen35`, 65 blocks, `full_attention_interval = 4`, `ssm.state_size 128 / group_count 16 / conv_kernel 4`, 336 `ssm_*` tensors across 48 blocks, `attn_k` present in only 17 (3, 7, 11, ... 63, 64). It is **dense** (0 expert tensors) — that part of the earlier correction is right — but dense ≠ non-recurrent, and it is the recurrent half that triggers the checkpoint.

So every verify step does `ctx->synchronize()` plus a full `llama_state_seq_get_data_ext` copy of the sequence state. Measured: batched 4-row verify **106.3 ms vs 42.3 ms** for a serial decode = 2.29x. Paying 2.29 decodes to save a median of 1 forward is a guaranteed loss, exactly as the numbers show.

**Upstream does not pay this.** llama.cpp added bounded partial recurrent rollback: `n_rs_seq` widens the RS cache to `(1 + n_rs_seq)` snapshot groups (`llama-memory-recurrent.cpp:100`), and `seq_rm` rolls back in-place via `set_rs_idx` when `rollback <= n_rs_seq` (`:182-190`), enabled for exactly `LLM_ARCH_QWEN35`/`QWEN35MOE` (`llama-arch.cpp:1034-1042`). The server therefore only snapshots when `draft.size() > llama_n_rs_seq(ctx_tgt)` (`server-context.cpp:3069-3070`), and sizes `n_rs_seq = draft.n_max` for MTP (`common.h:386-392`). Skippy pins `n_rs_seq = max(n_rs_seq, 2)` (`skippy/model_loading.cpp:662`) and then snapshots **unconditionally**, so mesh takes the expensive path on every step where upstream takes none.

That gives a concrete next change: size `n_rs_seq` from the configured draft window and make `skippy_checkpoint_verify_window` skip the state copy when the span fits within `llama_n_rs_seq`, using in-place `seq_rm` rollback like upstream. That is a separate PR against the C++ patch queue; I did not want to bundle a `third_party` patch change with the Rust decode-path fix.

## Scope of evidence

- Verified: the skip executes and forwards drop (span counts above); tests/clippy/fmt at this HEAD; the LM Studio A/B numbers.
- Verified by source read, not by timing instrumentation: that the per-verify state copy accounts for the 106.3 ms. The 2.29x ratio is measured; its attribution to the checkpoint is the leading explanation, not itemised.
- Not done: temp-0 token-stream identity diff between MTP on and off. That is a merge gate for this PR and is why it is a draft.

---

## Update: verify-window checkpoint fix (`0022`), measured

Third commit adds patch-queue change `0022`: `skippy_verify_window_can_roll_back_in_place` skips both the per-verify sequence-state snapshot **and** the serial replay-on-restore when the span fits inside `n_rs_seq`. Staged sessions keep the checkpoint — they roll back across hosts, where an in-process RS index is not a valid restore point.

This confirms the mechanism named above, by timing rather than inference:

| config | verify span | verify | serial decode | tok/s (3 runs) |
|---|---|---|---|---|
| `draft_max_tokens=3` (default) | 4 rows | 108.5 ms | 48.0 ms | 18.5 / 15.9 / 14.3 |
| `draft_max_tokens=2` | 3 rows | 94.3 ms | 46.5 ms | 15.7 / 18.5 / 18.3 |
| **`draft_max_tokens=1`** (fits `n_rs_seq=2`) | **2 rows** | **63.3 ms** | **42.1 ms** | **27.6 / 24.2 / 23.7** |

Only the arm where the guard fires gets fast; the first two rows are the pre-fix behaviour, correctly declined because span (draft+1) exceeds `n_rs_seq=2`. The RS cache is deliberately **not** widened: one snapshot group is ~300 MiB on this model (897.75 MiB RS buffer at `n_rs_seq=2`).

Interleaved single-session A/B, same box/GGUF/prompt, temp 0, 300 tok:

| arm | tok/s | median |
|---|---|---|
| mesh MTP off | 24.50 23.75 23.50 / 24.13 23.59 23.41 | 23.67 |
| mesh MTP on (this branch) | 23.56 23.40 22.80 / 27.98 24.60 24.81 | **23.98** |
| LM Studio MTP on | 27.13 22.75 20.79 / 28.90 26.97 25.02 | 26.05 |

**Honest status: the −30% regression is gone (parity with MTP-off). Mesh is still ~8% behind LM Studio MTP-on, so this is not yet a win over upstream.**

Remaining gap, from the same telemetry, with two candidates I have **not** distinguished:
1. At span 2 the verify still costs 63.3 ms vs a 42.1 ms serial decode (1.5x) while saving exactly one forward — residual per-verify overhead outside the snapshot.
2. Accept rate 0.60 vs LM Studio's 0.826 at mean accepted length 2.20 — drafting from a speculative branch costs accuracy, and reaching spans of ~2.2 accepted tokens needs `n_rs_seq > 2` without re-introducing the snapshot.

Still open: temp-0 token-stream identity diff between MTP on and off. That remains the merge gate.

## Summary by CodeRabbit

* **New Features**
  * Added native speculative decoding for eligible local, greedy-generation requests.
  * Supports integrated and external draft models with automatic mode selection.
  * Verifies draft token spans in batches and accepts matching tokens to improve generation speed.
  * Added native MTP draft verification and continuation handling.

* **Bug Fixes**
  * Improved diagnostics, rollback, and cleanup for recurrent verification windows and draft proposals.

* **Documentation**
  * Updated the default native MTP draft limit to one token.
  * CUDA release builds now detect the installed toolkit version automatically.


michaelneale:
@ndizazzo review request — MTP on Qwen3.8 (hybrid recurrent `qwen35`), measured on an M5 Mac.

Three commits: draft-token forwarding, local accept-and-skip in the Rust decode path, and patch-queue `0022` which is the one that matters for perf.

**What `0022` does:** skippy called `skippy_checkpoint_verify_window` unconditionally before every batched verify (`verification.cpp:141`) — `ctx->synchronize()` plus a full `llama_state_seq_get_data_ext` of the sequence state. Upstream `server-context.cpp:3069-3070` only snapshots when `draft.size() > llama_n_rs_seq(ctx_tgt)`, so on this arch it normally takes none. New guard `skippy_verify_window_can_roll_back_in_place` skips both the snapshot and the serial replay-on-restore when the span fits inside `n_rs_seq`, rolling back in place via `seq_rm`. Staged/split sessions keep the checkpoint — they roll back across hosts, where an in-process RS index is not a valid restore point.

**Numbers (same box, interleaved arms in one session, 300 tok, temp 0):** mesh MTP off median 23.67 tok/s, mesh MTP on 23.98, LM Studio MTP on 26.05. The prior -30% regression is gone; still ~8% behind LM Studio, which is inside this box's run-to-run drift. Verify cost tracks the checkpoint exactly: span 4 = 108.5 ms, span 3 = 94.3 ms, span 2 (guard fires) = 63.3 ms vs 42.1 ms serial decode.

**Known limits, not fixed:** I did not widen the RS cache, so only draft=1 (span 2) benefits — one snapshot group is ~300 MiB here (897.75 MiB RS buffer at `n_rs_seq`=2). Accept rate 0.60 vs LM Studio's 0.826 at mean len 2.20. Temp-0 token-stream identity diff on/off is still unrun.

Questions for you: is the in-place rollback guard the right shape, and is keeping the unconditional checkpoint on the staged path what you would expect?

michaelneale:
@ndizazzo a good cuda test where you are likely to see performance boost makes sense. I wonder if on unified you are less likely to see MTP help as the bottleneck is elsewhere? (but cuda you would see it)

ndizazzo (approved):
Repro'd, working locally on my 5090

PR #1367, opened 2026-08-18
fix(skippy): stop a stalled SSE consumer pinning a generation worker

ndizazzo:
## What

A generation worker parks in `mpsc::Sender::blocking_send` when the SSE event channel is full, and `blocking_send` waits **indefinitely**. If the consumer stops draining without the channel ever being dropped — a client gone before the server's own disconnect detection notices, e.g. behind a proxy that doesn't propagate the close — that send never returns. The worker, and the execution lane it holds, are pinned forever, even after the request is cancelled. With `parallel = 1` one such worker blocks every other request.

This replaces the six ad hoc `blocking_send` call sites in `run_generation_stream` with a `StreamEventSender` that races each send against cancellation and against a stall timeout, so a send can no longer wait unboundedly.

## Scope — what this does *not* claim

**This is not established as the cause of the `qwen-3.8-perf` orphan (report defect #5), and that item stays open.** A worker parked in `blocking_send` is not computing, so it cannot explain the reported *"GPU stayed at ~65% for 3+ minutes"*. A hypothesis that fits all the reported evidence better: the capture ran opencode behind a logging TCP proxy, which keeps draining the SSE body after the real client dies — so nothing is ever full, `CancelOnDropSseStream::drop` never fires, and generation runs to `max_tokens` (32000 in that capture; ~464 s at the measured 69 tok/s). This PR does not help that case.

Settling it is cheap: reproduce the orphan and sample GPU utilization. Near-0% ⇒ a stalled send, and this is the fix. Sustained busy ⇒ an uninterruptible run-to-`max_tokens`, and the real fix is elsewhere. Either way the `blocking_send` hazard is real and latent, and worth closing on its own merits.

### `StreamEventSender::send` — in-flight events

`tokio::select!` (biased, cancellation first) over cancellation, `tx.send`, and `tokio::time::sleep(stall_timeout)`, run under a `tokio::runtime::Handle` — the pattern already used from inside this same `spawn_blocking` worker at `prompting.rs:291`/`:317`; `CancellationToken::cancelled()` is `Notify`-based.

No polling. The earlier revision of this PR used `try_send` + a 20 ms sleep, which taxed the *healthy* path: capacity is 16, so a consumer slightly slower than decode hits `Full` routinely — ordinary backpressure, not a fault — and would have paid up to 20 ms per event, capping throughput near 50 events/s. The `select!` wakes the instant space appears, and cancellation is observed immediately rather than up to a tick late.

On a dropped or stalled receiver, the request is cancelled and the receiver is recorded as unreachable.

### `StreamEventSender::send_terminal` — frames emitted after generation finishes

Deliberately does **not** consult cancellation. The cancellation error, a `parser.finish` error, the backend error, usage, and `Done` are exactly the frames the frontend lifecycle needs when the request *was* cancelled; on `main`, `blocking_send` delivered them unconditionally.

This fixes a regression in the first revision of this PR: the cancellation frame was sent from *inside* `if context.is_cancelled()`, and the helper opened with an `is_cancelled` bail, so the frame was never enqueued. That is not cosmetic — the `Err` frame drives `observe_backend_stream` → `lifecycle.failed(error)`, which sets `backend_error` and emits a classified `StreamTerminal`; without it `drop_outcome()` falls through to `StreamDropOutcome::Cancelled`. Both the client-visible error frame and the terminal telemetry changed silently. The same shape swallowed the `parser.finish` error and the outer generation error.

`send_terminal` does still refuse to wait on a receiver already *proven* unreachable. Otherwise a genuinely stalled consumer would be waited out twice — once in flight, once for the terminal frame — doubling the lane hold to `2 * stall_timeout` and defeating the point.

### Stall timeout

`STREAM_SEND_STALL_TIMEOUT` is now its own `Duration::from_secs(10)` rather than an alias of `GENERATION_ADMISSION_TIMEOUT`. Admission queueing and stream-stall tolerance are unrelated policies; retuning one must not silently retune the other.

**Correcting a claim in this PR's earlier description:** it said a stalled generation "can never hold a lane longer than everyone else already times out for". That is false as written. The timer is **per send**, not per generation — a consumer that drains one event every 9.9 s resets it indefinitely. The real property is narrower: *no single send stalls more than 10 s*. A generation-scoped deadline would be a separate change.

### Dead file removed (separate commit)

`crates/openai-frontend/src/router/stream_lifecycle.rs` is not compiled: `lib.rs:13`'s `pub mod router;` resolves to `src/router.rs`, and the only `mod stream_lifecycle;` (`lib.rs:15`) resolves to `src/stream_lifecycle.rs`. The two files have drifted (`completed()`/`dropped()` vs `finish_natural()`/`finish_drop()` with `protocol_complete` handling). It sat directly in this PR's blast radius, so it is deleted here. **The earlier description cited `router/stream_lifecycle.rs:154`, inherited from the report — that is the dead copy.** The live `Drop` is `crates/openai-frontend/src/stream_lifecycle.rs`.

## Commits

- `22b37fc8a` — `fix(skippy): deliver terminal SSE frames and drop the send poll loop`
- `69b026749` — `chore(openai-frontend): delete the uncompiled router/stream_lifecycle.rs`

## Tests

Four in `crates/skippy-server/src/frontend/backend/tests.rs`. All run the send against a live, never-drained receiver so a regression to an unconditional blocking send fails the test instead of hanging the suite.

| Test | Proves |
|---|---|
| `stalled_receiver_does_not_pin_the_generation_worker_forever` | A cancelled request interrupts a stalled send (real 10 s timeout, so cancellation must be what ends the wait) |
| `stalled_receiver_self_cancels_after_the_stall_timeout_with_no_external_cancel` | With nothing ever calling `cancel()`, a stalled send still self-cancels and frees the lane |
| `terminal_frames_are_delivered_after_the_request_is_cancelled` | Red→green for the swallowed terminal frame; also asserts in-flight `send` still refuses once cancelled |
| `terminal_frames_are_dropped_once_the_receiver_is_proven_unreachable` | The terminal send short-circuits (<25 ms against a 50 ms stall timeout) instead of waiting the timeout twice |

## Validation

At `69b026749`:

```
cargo fmt --all --check                                     ok
cargo check   -p skippy-server                              ok
cargo clippy  -p skippy-server   --all-targets -- -D warnings   ok
cargo test    -p skippy-server   --lib                      451 passed; 0 failed; 3 ignored
cargo check   -p openai-frontend                            ok
cargo clippy  -p openai-frontend --all-targets -- -D warnings   ok
cargo test    -p openai-frontend --lib                      178 passed; 0 failed
```

Not yet run: an end-to-end reproduction of the orphan on a GPU host. See *Scope* above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of live generation updates when event delivery is delayed or interrupted.
  * Generation requests now stop promptly when the connection is cancelled or no longer accepting updates.
  * Prevented stalled event delivery from keeping generation requests active indefinitely.
  * Ensured final generation status updates are delivered when possible after cancellation.
  * Improved diagnostic visibility for interrupted or stalled generation streams.

* **Tests**
  * Added coverage for cancelled connections and timed-out event delivery scenarios.


michaelneale:
this is a really good one - I bet this bit a lot!

michaelneale:
Reviewed at `27a110b0` against `origin/main` = `084105d14`. The `blocking_send` hazard is real and the test is a genuine red→green, so I'd like this to land — but I don't think it fixes the defect the description claims, and it has one behaviour regression plus one incorrect claim in the body. Details below, most important first.

---

## 1. The mechanism is inconsistent with the reported symptom

The report's #6 (the orphan) says:

> the GPU stayed at ~65% for **3+ minutes** and probe requests failed with `timed out waiting for an execution lane after 10 seconds`

A worker parked inside `mpsc::Sender::blocking_send` is **not computing**. Under this PR's mechanism the orphan would hold the lane with the GPU near **idle**, not at 65%. So the observed 65% for 3+ minutes falsifies the mechanism-to-symptom link, even though the mechanism itself is a real latent bug.

A hypothesis that fits *all* the reported evidence: the capture ran opencode behind a **logging TCP proxy** (report §1, appendix). A proxy keeps reading from the server after the real client dies. So:

- the proxy keeps draining the SSE body → the channel never stays full → `blocking_send` never blocks and `try_send` never returns `Full`
- the server-side socket stays open → `CancelOnDropSseStream::drop` never fires → `cancellation.cancel()` is never called
- generation therefore runs to `max_tokens`, which the capture set to **32000**; at the measured 69 tok/s that is ~464 s of continuous decode

3+ minutes of ~65% GPU with the only lane held, and no cancellation, drops out of that directly. This PR does not help that case at all: nothing is full, and nothing ever cancels.

Cheap way to settle it rather than argue: reproduce the orphan and sample GPU utilization. Near-0% ⇒ a stalled send, this PR is the fix. Sustained busy ⇒ an uninterruptible run-to-`max_tokens`, and the real fix is elsewhere.

I'd suggest retitling/rescoping this to what it demonstrably does — "a stalled SSE consumer can pin a generation worker indefinitely" — and leaving the report's #5 open until the GPU-utilization check comes back. Landing this as "#5 fixed" risks closing a lane-exhaustion bug that is still live.

## 2. Regression: the cancellation frame is now never sent (must fix)

`backend.rs` (post-diff):

```rust
if context.is_cancelled() {
    let _ = send_generation_event(&tx, Err(request_cancelled_error()), &context);
    return;
}
```

`send_generation_event_with_stall_timeout` starts its loop with:

```rust
if context.is_cancelled() {
    return Err(request_cancelled_error());
}
```

Inside that branch `context.is_cancelled()` is `true` by construction — it *is* the branch condition. So the helper returns immediately and the cancellation frame is **never enqueued**, where on `main` `blocking_send` delivered it whenever the buffer had room (capacity 16, `backend.rs:815`).

Not just cosmetic. On `main` that `Err` reaches `observe_backend_stream` → `lifecycle.failed(error)` (`openai-frontend/src/stream_lifecycle.rs:196-198`, `:80-86`), which sets `backend_error` and emits a `StreamTerminal` classified from the error. With the frame dropped, `drop_outcome` (`:111-122`) instead falls through to `StreamDropOutcome::Cancelled`. So both the client-visible error frame and the terminal telemetry classification change silently.

The same shape affects the two terminal `Err` sends (`parser.finish` failure, and the outer `Err(error)` arm): if anything cancelled the context earlier — including a self-cancel from this very helper — the error is swallowed rather than reported.

Fix: send terminal frames with a variant that doesn't consult `is_cancelled` (a plain `try_send`, or a `terminal: bool` / separate `send_terminal_event`). The stall/closed guards should still apply; only the pre-emptive cancellation check should be bypassed.

## 3. The body's central claim is false as written

> so a stalled generation can never hold a lane longer than everyone else already times out for

`let stalled_since = Instant::now();` is **per call**, so the 10 s bound is per *event*, not per generation. A consumer that drains one event every 9.9 s resets the timer forever and holds the lane indefinitely — exactly the failure mode being fixed, just slower.

That matters because the justification for reusing `GENERATION_ADMISSION_TIMEOUT` rests on this claim. Either state the real property ("no single send stalls more than 10 s") or add a generation-scoped deadline if a total bound is actually wanted. Related: aliasing `STREAM_SEND_STALL_TIMEOUT = GENERATION_ADMISSION_TIMEOUT` couples two unrelated policies — someone retuning admission silently retunes stream-stall behaviour. I'd give it its own value.

## 4. The 20 ms poll taxes the healthy backpressure path

Channel capacity is 16 (`backend.rs:815`). A consumer that is merely *slightly* slower than decode hits `Full` routinely — that's normal backpressure, not a fault. `blocking_send` wakes the instant space appears; `try_send` + `thread::sleep(20ms)` can wait up to 20 ms per event, capping visible throughput at ~50 events/s under sustained backpressure. Against the 69 tok/s decode this PR's sibling defects are trying to protect, that's a meaningful regression on a common path.

There's a fully event-driven alternative with no polling and *faster* cancellation response, using pieces already in the crate — `CancellationToken::cancelled()` is notify-based (`openai-frontend/src/backend.rs:54-62`), and a `tokio::runtime::Handle` is already in scope at the spawn site (`hook_runtime`, `Handle::current()`; `Handle` is `Clone`):

```rust
handle.block_on(async {
    tokio::select! {
        result = tx.send(event) => ...,
        () = context.cancelled() => ...,
        () = tokio::time::sleep(stall_timeout) => ...,
    }
})
```

Immediate wakeup on space, immediate wakeup on cancel, no 20 ms tax, and the stall timeout still enforced. Worth considering before locking in the poll loop.

## 5. Smaller things

- **The description cites a dead file.** `router/stream_lifecycle.rs:154` (inherited from the report) is not the code that runs. `crates/openai-frontend/src/router/stream_lifecycle.rs` has **no `mod` declaration anywhere in the tree** — the only `mod stream_lifecycle;` is `lib.rs:15`, which resolves to `src/stream_lifecycle.rs`. I confirmed rather than inferred: appending a syntax error to the `router/` copy and running `cargo check -p openai-frontend --lib` still succeeds. The live `Drop` is `src/stream_lifecycle.rs:253-258`, and the two files have drifted (`completed()`/`dropped()` vs `finish_natural()`/`finish_drop()` with `protocol_complete` handling). It's a 160-line trap sitting directly in this PR's blast radius — worth deleting in a separate commit, and worth fixing the citation here so the next reader doesn't reason from it.
- **`finish_reason` interaction.** Not touched by this PR, but while you're in the streaming terminal path: with a bypassed terminal send, double-check nothing now emits `Done` *and* an error frame for the same request.
- **Tests are good** — particularly `..._with_no_external_cancel` covering the no-`cancel()`-at-all case, and running the send off-thread with `recv_timeout` so a regression fails instead of hanging the suite. If you take the `select!` approach, please keep both tests; they're the durable part of this PR.

---

Summary: I'd like #2 fixed and #3 corrected before merge, #1 resolved by rescoping the claim (the code change can still land), and #4 considered. The `blocking_send` hazard is worth closing regardless of whether it turns out to be the carrack orphan.

ndizazzo:
Thanks — all four actionable points are addressed in `22b37fc8a` + `69b026749`, and I've rewritten the title and description. Point by point.

## 1. Mechanism vs. symptom — agreed, rescoped

You're right that a worker parked in `blocking_send` isn't computing, so it can't produce 65% GPU for 3+ minutes. I've retitled to what this demonstrably does ("stop a stalled SSE consumer pinning a generation worker") and the description now states explicitly that **report #5 stays open** pending the GPU-utilization check, along with your run-to-`max_tokens`-behind-a-proxy hypothesis and the near-0%-vs-sustained-busy test that settles it. I don't have the box; flagged to Nick in the channel.

## 2. Swallowed terminal frame — confirmed and fixed

Confirmed exactly as you describe. Terminal frames now go through `send_terminal`, which does not consult cancellation at all; only in-flight events (the `on_text_chunk` path) keep the cancellation check. That covers all four sites: the cancellation frame, the `parser.finish` error, the backend error, and usage/`Done`.

`terminal_frames_are_delivered_after_the_request_is_cancelled` is the red→green: it cancels the context, sends a terminal frame, and asserts the receiver actually yields it — plus that in-flight `send` still returns `Err` on the same cancelled context, so the bypass is scoped to terminal frames only.

**One thing your review didn't cover, which shaped the fix.** Bypassing the cancellation check unconditionally means a genuinely stalled consumer gets waited on twice: 10 s in-flight (which cancels), then another 10 s for the terminal frame on the still-full channel — a 20 s lane hold, worse than the bug. So `send_terminal` short-circuits on a receiver already *proven* unreachable (closed, or stalled past the timeout), while still delivering when the request was merely cancelled externally and the receiver is alive and draining. `terminal_frames_are_dropped_once_the_receiver_is_proven_unreachable` bounds it at <25 ms against a 50 ms injected timeout.

## 3. False claim — corrected, and the constant decoupled

Correct, the timer is per call. The description now states the real property ("no single send stalls more than 10 s") and notes a generation-scoped deadline would be a separate change. `STREAM_SEND_STALL_TIMEOUT` is its own `Duration::from_secs(10)`; the aliasing to `GENERATION_ADMISSION_TIMEOUT` is gone, for the reason you gave.

## 4. 20 ms poll — taken, with the `select!` you sketched

Adopted. `Handle::block_on` + `tokio::select!` (biased, cancellation first) over `tx.send`, `cancellation.cancelled()`, and `tokio::time::sleep`. Your read that this is safe here is right and I checked the precedent before writing it: `prompting.rs:291`/`:317` already `handle.block_on(...)` from inside this same `spawn_blocking` worker.

One implementation note for the next reader: the `sleep` future must be constructed *inside* the `async` block passed to `block_on`, not before it. `tokio::time::sleep()` registers with the runtime timer when called, not when polled, so hoisting it panics with "there is no reactor running". Commented at both sites.

Both tests you called out are kept, adapted to the new API with their doc comments intact — including `..._with_no_external_cancel`, and the off-thread + `recv_timeout` structure.

## 5. Smaller things

- **Dead file** — confirmed independently before deleting: `lib.rs:13`'s `pub mod router;` resolves to `src/router.rs` (which exists), and the only `mod stream_lifecycle;` is `lib.rs:15` → `src/stream_lifecycle.rs`. `src/router/` held that one orphaned file and nothing else. Deleted in `69b026749`, and the description's citation now points at the live file.
- **`finish_reason` interaction** — checked: `send_terminal` bypasses only the *cancellation* check, not the control flow. The `if context.is_cancelled()` branch still `return`s after its error frame, so nothing emits `Done` and an error frame for the same request.

## Validation at `69b026749`

`cargo fmt --all --check`, `cargo check`/`clippy --all-targets -- -D warnings` for both `skippy-server` and `openai-frontend`, `cargo test -p skippy-server --lib` (451 passed, 3 ignored), `cargo test -p openai-frontend --lib` (178 passed). No end-to-end GPU reproduction — see point 1.

PR #1371, opened 2026-08-18
Stream tool calls as they are generated instead of after the turn ends

### LINKED ISSUE #1369: Tool calls are never streamed incrementally: 12-30s of silence per agent turn
## Summary

On the in-process serving path (`mesh-llm serve`), tool calls are never streamed
incrementally. The parser holds the entire tool call back until the final,
non-partial parse and then emits it in a single SSE chunk. For an agent client
this is indistinguishable from a hung server: wire capture of a real opencode
turn showed the `role` chunk at t=0, **nothing for 29 s**, then a 701-token tool
call in one chunk. Three of five turns in that session had the same shape
(154 tok / 12 s, 701 tok / 29 s, 29 tok / 15 s).

This is defect #4 from the Qwen3.8-27B serving investigation on `carrack`
(RTX 5090, `unsloth/Qwen3.8-27B-GGUF:UD-Q4_K_XL`, arch `qwen35`). It is
independent of #1365 (reasoning visibility, merged): with reasoning exposed and
content tokens flowing normally, a turn that ends in a tool call is still
silent for the whole tool-call span.

All line numbers below are at `origin/main` = `084105d14`.

## Mechanism

`crates/skippy-server/src/frontend/generation/streaming.rs:112-117`:

```rust
if let (true, Some(tool_calls)) =
    (!is_partial && !self.emitted_tool_calls, parsed.tool_calls)
{
    self.emitted_tool_calls = true;
    events.push(GenerationStreamEvent::ToolCalls(tool_calls));
}
```

`!is_partial` is the whole defect. The two entry points into
`events_for_text` are:

- `push_delta` (`streaming.rs:69-75`) → `events_for_text(true)` — called once per
  decoded token, i.e. every streaming opportunity
- `finish` (`streaming.rs:77-85`) → `events_for_text(false)` — called exactly
  once, after generation completes

So the only code path that can ever emit a `ToolCalls` event is the terminal
one. Content and reasoning, by contrast, stream fine because they go through
`suffix_delta` (`streaming.rs:130-141`), which tracks the already-emitted prefix
and emits only the new suffix. There is no equivalent for tool calls.

The tool-emulation path has an independent copy of the same gate —
`crates/skippy-server/src/frontend/prompting.rs:351`:

```rust
let tool_calls = if is_partial || parse.tool_calls.is_empty() {
    None
```

so both the native-template and emulated-tool-call paths need fixing, not just one.

## Why the connection doesn't look dead

`sse_response` sets `.keep_alive(KeepAlive::default())`
(`crates/openai-frontend/src/stream_lifecycle.rs:211-212`), and axum's default
`max_interval` is 15 s (`axum-0.8.9/src/response/sse.rs:527`). The client
therefore receives SSE comment frames and the socket stays healthy — it just
receives no content for the entire tool-call span. That matches the reported
symptom ("it just sits there") better than a connection fault, and is why this
is easy to misdiagnose as a stall.

## The data may already be available at partial time

Two pieces of evidence suggest the fix may not need an ABI change, i.e. the
partial parse already carries tool calls and we are discarding them:

1. `parsed_chat_message_from_json`
   (`crates/skippy-server/src/frontend/generation/parsing.rs:252-264`) extracts
   `tool_calls` from the parsed message unconditionally — it is not gated on
   `is_partial`. The gating happens later, in `streaming.rs:112`.
2. The native parse is `common_chat_parse(generated_text, is_partial, params)`
   (`third_party/llama.cpp/patches/0010-Add-Skippy-tokenization-and-stage-chat.patch`,
   in `skippy_parse_chat_response_json`), and upstream's partial path
   (`common_chat_peg_parse`, `common/chat.cpp:3853-3876`) deliberately maps
   whatever AST nodes were captured into the message specifically so that
   streaming works.

**I have not confirmed this on a live model.** Task 1 below is to print the
actual `parsed.tool_calls` on partial parses for `qwen35` before committing to a
fix shape. If partial tool calls do come through, this is a local Rust change in
`streaming.rs` + `prompting.rs`. If they don't, it needs the ABI to expose
something diff-shaped — upstream already has `common_chat_msg_diff::compute_diffs`
(`common/chat.h:134-147`) for exactly this, and `skippy_parse_chat_response_json`
currently exposes only whole-message JSON.

## Downstream is already built for multi-chunk tool-call deltas

The front door's stream normalizer memoizes a synthetic tool-call ID per
`index` across chunks —
`crates/mesh-llm-host-runtime/src/network/openai/tool_call_ids.rs:102-115`:

```rust
fn tool_call_id(&mut self, index: u64, tool_call: &Map<String, Value>) -> String {
    if let Some(existing) = self.tool_call_ids.get(&index) {
        return existing.clone();
    }
```

That cache is only meaningful if the same `index` is expected to appear in more
than one chunk. So the mesh-side normalizer already assumes incremental
tool-call deltas; it is the in-process producer that never emits them.

Conversely, `tool_calls_stream_delta`
(`crates/skippy-server/src/frontend/generation/parsing.rs:404-427`) only
back-fills a missing `index` field. It has no notion of accumulating argument
fragments, which an incremental implementation will need.

## Proposed work

1. **Confirm the input.** Instrument or unit-test `parse_chat_output` with
   `is_partial = true` against a `qwen35` tool-call generation and print
   `parsed.tool_calls`. This decides whether the fix is Rust-local or needs an
   ABI addition. Do this before writing the fix.
2. **Stream argument deltas.** Track per-`index` emitted state (name emitted
   once with the ID; `arguments` emitted as a growing suffix, the same shape as
   `suffix_delta`) and emit `ToolCalls` deltas from `events_for_text(true)`.
   Emit the `id` and `function.name` on the first delta for an index, per the
   OpenAI streaming contract, and argument fragments thereafter.
3. **Cover the emulated path too** (`prompting.rs:351`), or state explicitly in
   the PR that it is out of scope and why.
4. **Keep the terminal parse authoritative.** `finish` must still reconcile —
   a partial parse can be revised (upstream's lenient partial mapper can
   produce a shorter name that later grows; see the prefix-tolerance logic at
   `common/chat.cpp:308-316`). Emitting a delta must not make an incorrect
   early guess unrecoverable.
5. **Don't regress `finish_reason`.** `finish_reason` currently derives from
   `self.emitted_tool_calls` (`streaming.rs:121-127`); that flag becomes
   "started emitting" rather than "emitted the complete call", so re-check it.

## Acceptance criteria

- A streaming chat completion whose generation ends in a tool call produces
  more than one chunk carrying `delta.tool_calls`, with the first delta
  carrying `id` and `function.name`.
- Time from the first tool-call token to the first `delta.tool_calls` chunk is
  on the order of a token, not the whole tool call.
- The concatenation of streamed `arguments` fragments equals the
  non-streaming `arguments` for the same generation.
- The front door's normalized output is unchanged in aggregate (same IDs, same
  final arguments) — `tool_call_ids.rs` per-index memoization should make this
  automatic, but assert it.

## Not claimed

- Not measured on any host other than `carrack`. The 12-30 s figures are that
  box, that model.
- I have not checked whether the out-of-process/binary-transport serving path
  has the same gate; the citations above are all the in-process path that
  `mesh-llm serve` uses.
- No claim that fixing this changes tokens/sec. This is time-to-first-visible
  -progress only; the underlying generation is already running.

michaelneale:
Closes #1369.

An agent turn that ends in a tool call used to go completely quiet for the whole
tool-call span and then deliver the call in a single SSE chunk. A wire capture of
a real opencode turn on `carrack` showed the `role` chunk at t=0, **nothing for
29s**, then a 701-token tool call at once — indistinguishable from a hung server.

Tool calls now stream as the model generates them. The first
`delta.tool_calls` chunk carries `id`, `type`, and `function.name`; argument
fragments follow token by token, and a client concatenating them reconstructs
exactly the arguments the non-streaming path returns.

## What was actually wrong

The data was already there and we threw it away. The chat parser re-parses the
whole generated prefix on every decoded token, and upstream's partial mapper
pushes a tool call into the result as soon as its name is seen — explicitly
commented "Add the tool call to results so streaming can see it" — then
accumulates arguments incrementally. `streaming.rs` gated emission on
`!is_partial`, so the only parse that could ever emit was the terminal one.

I confirmed this before writing the fix (issue task 1) by driving
`common_chat_parse(prefix, is_partial = true)` over growing byte-prefixes of a
`qwen35` generation: the tool name appears at prefix byte 56 and arguments grow
one byte at a time from there. So this is a Rust-local change — no Skippy ABI
change, no `common_chat_msg_diff::compute_diffs`, no C++.

That recording is now committed as a test fixture with the generator that
produces it (`crates/skippy-server/tests/fixtures/`), so the tests run against
real parser output rather than a hand-written guess, and the fixture can be
regenerated after a llama.cpp pin bump.

## Three things that are easy to get wrong here

**Tool-call ids churn between parses.** `ensure_tool_call_ids` mints a fresh
UUID on every parse for any call the model did not id itself, so consecutive
partial parses disagree on `id`. The id is captured on the first delta for an
index and never re-read. This would have been *partly invisible* through the
front door — `ChatStreamNormalizationState` memoizes the first id per index — so
a mesh-routed test would pass while a direct-to-skippy client saw id churn.
Tests cover both layers.

**`index` must be explicit.** Deltas now carry only the calls that changed, so
back-filling `index` from array position would label a delta for call #1 as
`index: 0` and clients would merge it into the wrong call. Silent corruption, so
there is a test for exactly that case.

**`finish_reason` had to stop meaning "started emitting".** It derived from a
flag that now flips on the first argument byte. Worse, a partial parse can be
revised downward on malformed output, and once that breaks the append-only
contract the client can never receive the rest — so the terminal parse only
reports `tool_calls` if what went out on the wire actually reconstructs it.
Otherwise the client would execute a tool call whose arguments it only partly
received.

## Scope

Emulated tool calls (models whose chat template lacks native tool support) are
deliberately unchanged and still arrive at finalization — issue task 3. That
parser only recognises a call once its JSON object closes, so it has no
representation of a partially generated argument object to stream from;
streaming it needs a partial-JSON scanner, not a gate flip. The stale doc comment
claiming it "matches native tool-call streaming semantics" is corrected.

**This does not help `model=mesh` / `model=auto`.** MoA fans out with
`"stream": false` and its SSE synthesizer emits one atomic tool-call delta by
design, pinned by `chat_sse_tool_calls_remain_atomic`. Worth knowing before
anyone measures TTFT through `model=mesh` and concludes nothing changed.

## Architecture

`ChatOutputStreamParser` kept both the parser call and the incremental-delta
decision, and the latter could not be tested without a loaded runtime. The delta
decision now lives in `ChatStreamDeltas`, the per-index tool-call arithmetic in
`generation/tool_call_stream.rs`, and the prefix-diff primitive shared by
content, reasoning, and arguments in `generation/incremental_text.rs` (it was
previously owned by `streaming.rs` but is not tool-call-specific).

## Protocol

No wire-format or ABI change. Same chunk schema, same field names — a tool call
simply arrives across several `delta.tool_calls` chunks instead of one, which is
the standard OpenAI streaming shape and what the front-door normalizer already
assumed. The non-streaming `/v1/chat/completions` path and `/v1/completions` are
untouched (`/v1/completions` never constructs a chat parser at all).

## Validation

- `cargo test -p skippy-server` — 473 passed, 0 failed
- `cargo test -p mesh-llm-host-runtime --lib` — 2515 passed, 0 failed
- `cargo clippy -p skippy-server --all-targets -- -D warnings` — clean

All at `a7b928365`, which contains `origin/main` `3fcd5dc48`.

New coverage (there was no `push_delta`/`is_partial` test before this):
per-index delta arithmetic, id-churn suppression, `index` labelling for parallel
calls, divergence handling, terminal reconciliation, and the wire shape through
the real chunk conversion plus the front-door SSE normalizer round trip. I also
confirmed the tests fail against the old atomic behaviour rather than passing
trivially.

**Not yet measured live.** Everything above is unit-level plus a parser probe.
I have no `qwen35` tool-capable GGUF on the lab boxes, so I have not produced a
before/after time-to-first-`delta.tool_calls` number on a real model — the
12-30s figures are @nickfrosty's `carrack` measurements, not mine. Happy to hold
merge for that if you'd rather have it first.

An independent reviewer pass on this branch found the divergence/`finish_reason`
gap described above and the missing `ChatStreamDeltas`-level coverage; both are
fixed in this branch.

## Summary by CodeRabbit

- **Improvements**
  - Improved streaming tool-call delivery by preserving stable call IDs and progressively reconstructing JSON arguments.
  - Prevented incomplete, unnamed, or regressed tool calls from being emitted prematurely.
  - Improved handling of parallel tool calls and terminal completion states.
  - Preserved incremental reasoning and response text without duplicating or losing content.

- **Documentation**
  - Added guidance and examples for streaming tool-call parsing and fixture generation.

- **Tests**
  - Added coverage for partial responses, argument reconstruction, stable identifiers, truncated generations, and multiple tool calls.


ndizazzo (approved):
lgtm - have not been able to test this one though - I'll do one big batch when we cut RC4

PR #1378, opened 2026-08-19
fix(openai): restore authenticated observability custody

i386:
## What changed

Restore the target-specific `mesh_openai_observability` lifecycle projection alongside the canonical lifecycle events. The projection carries the request lifecycle, trusted agent-session identity, authoritative usage counts, terminal outcome, and elapsed timing that Cacheline uses for model-performance custody.

## Root cause

The lifecycle refactor removed the old target-specific events. Mesh continued serving requests and emitting canonical JSON, but Cacheline’s custody collector correctly rejected those records because the canonical presentation intentionally omits session identity. The benchmark therefore stopped after source-only with an empty observability artifact.

## Validation

- `cargo check -p openai-frontend -p mesh-llm-host-runtime`
- `cargo test -p mesh-llm-host-runtime logging::openai_lifecycle::tests::legacy_observability_retains_authenticated_request_custody --lib`
- `cargo test -p mesh-llm-host-runtime logging::openai_lifecycle --lib`
- `cargo test -p openai-frontend lifecycle --lib`

The corresponding Cacheline benchmark pin should be updated only after this commit lands in the private Mesh main.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
+++ b/crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
@@ -409,6 +409,7 @@ struct ActiveRequest {
     /// embedded frontend still emits the legacy custody projection, but must
     /// not register or terminalize a second canonical request.
     guard: Option<LifecycleGuard>,
+    legacy: Arc<LegacyCustody>,
     started_at: Instant,
     operation: OpenAiBackendOperation,
     agent_session_id: Option<String>,
@@ -416,14 +417,128 @@ struct ActiveRequest {
     backend_operation: Option<OpenAiBackendOperation>,
     backend_attempt: Option<(OpenAiBackendOperation, AttemptId)>,
     backend_stream_first_item: bool,
-    legacy_started: bool,
     usage: Option<OpenAiUsage>,
+}
+
+type LegacyAction = Box<dyn FnOnce() + Send + 'static>;
+
+#[derive(Default)]
+struct LegacyCustody {
+    state: Mutex<LegacyCustodyState>,
+}
+
+#[derive(Default)]
+struct LegacyCustodyState {
+    actions: VecDeque<LegacyAction>,
+    draining: bool,
+    started: bool,
+    terminalized: bool,
+}
+
+impl LegacyCustody {
+    /// Queue the two events that establish legacy custody as one atomic
+    /// per-request action sequence. The caller may hold the adapter's map
+    /// mutex while this method queues; logging itself is always drained after
+    /// that lock is released.
+    fn start(
+        &self,
+        request_id: RequestId,
+        operation: OpenAiBackendOperation,
+        agent_session_id: Option<String>,
+        agent_session_source: Option<String>,
+    ) -> bool {
+        let backend_session_id = agent_session_id.clone();
+        let backend_session_source = agent_session_source.clone();
+        self.start_with(
+            Box::new(move || {
+                log_legacy_request_started(
+                    request_id,
+                    operation,
+                    agent_session_id.as_deref(),
+                    agent_session_source.as_deref(),
+                );
+            }),
+            Box::new(move || {
+                log_legacy_backend_started(
+                    request_id,
+                    operation,
+                    backend_session_id.as_deref(),
+                    backend_session_source.as_deref(),
+                );
+            }),
+        )
+    }
+
+    fn start_with(&self, request_started: LegacyAction, backend_started: LegacyAction) -> bool {
+        let mut state = lock_recover(&self.state);
+        if state.started || state.terminalized {
+            return false;
+        }
+        state.started = true;
+        state.actions.push_back(request_started);
+        state.actions.push_back(backend_started);
+        Self::begin_draining(&mut state)
+    }
+
+    fn enqueue<F>(&self, action: F) -> bool
+    where
+        F: FnOnce() + Send + 'static,
+    {
+        let mut state = lock_recover(&self.state);
+        if state.terminalized {
+            return false;
+        }
+        state.actions.push_back(Box::new(action));
+        Self::begin_draining(&mut state)
+    }
+
+    fn finish<F>(&self, action: F) -> bool
+    where
+        F: FnOnce() + Send + 'static,
+    {
+        let mut state = lock_recover(&self.state);
+        if state.terminalized {
+            return false;
+        }
+        state.terminalized = true;
+        if !state.started {
+            return false;
+        }
+        state.actions.push_back(Box::new(action));
+        Self::begin_draining(&mut state)
+    }
+
+    fn begin_draining(state: &mut LegacyCustodyState) -> bool {
+        if state.draining {
+            false
+        } else {
+            state.draining = true;
+            true
+        }
+    }
+
+    fn drain(&self) {
+        loop {
+            let action = {
+                let mut state = lock_recover(&self.state);
+                match state.actions.pop_front() {
+                    Some(action) => action,
+                    None => {
+                        state.draining = false;
+                        return;
+                    }
+                }
+            };
+            action();
+        }
+    }
 }
 
 impl ActiveRequest {
     fn new(guard: Option<LifecycleGuard>, context: &OpenAiLifecycleContext) -> Self {
         Self {
             guard,
+            legacy: Arc::new(LegacyCustody::default()),
             started_at: Instant::now(),
             operation: route_operation(context.route),
             agent_session_id: context.agent_session_id.clone(),
@@ -431,7 +546,6 @@ impl ActiveRequest {
             backend_operation: None,
             backend_attempt: None,
             backend_stream_first_item: false,
-            legacy_started: false,
             usage: None,
         }
     }
@@ -524,7 +638,7 @@ impl OpenAiLifecycleLoggingAdapter {
             self.service.start_attempt(request_id, guard)
         });
 
-        let legacy_start = {
+        let legacy_drain = {
             let mut tracked = lock_recover(&self.tracked);
             let Some(TrackedRequest::Active(active)) = tracked.requests.get_mut(&request_id) else {
                 return;
@@ -535,40 +649,18 @@ impl OpenAiLifecycleLoggingAdapter {
             active.operation = operation;
             active.backend_operation = Some(operation);
             active.backend_attempt = attempt_id.map(|attempt_id| (operation, attempt_id));
-            if active.legacy_started {
-                None
-            } else {
-                active.legacy_started = true;
-                Some((
-                    active.agent_session_id.clone(),
-                    active.agent_session_source.clone(),
-                ))
-            }
-        };
-        if let Some((agent_session_id, agent_session_source)) = legacy_start {
-            log_legacy_request_started(
+            let legacy = Arc::clone(&active.legacy);
+            let should_drain = legacy.start(
                 request_id,
                 operation,
-                agent_session_id.as_deref(),
-                agent_session_source.as_deref(),
-            );
-        }
-        let session = {
-            let tracked = lock_recover(&self.tracked);
-            let Some(TrackedRequest::Active(active)) = tracked.requests.get(&request_id) else {
-                return;
-            };
-            (
                 active.agent_session_id.clone(),
                 active.agent_session_source.clone(),
-            )
+            );
+            (legacy, should_drain)
         };
-        log_legacy_backend_started(
-            request_id,
-            operation,
-            session.0.as_deref(),
-            session.1.as_deref(),
-        );
+        if legacy_drain.1 {
+            legacy_drain.0.drain();
+        }
     }
 
     fn backend_terminal(
@@ -585,17 +677,56 @@ impl OpenAiLifecycleLoggingAdapter {
             if active.backend_operation == Some(operation) {
                 active.backend_operation = None;
                 let backend_attempt = active.backend_attempt.take();
+                let elapsed = active.started_at.elapsed();
+                let agent_session_id = active.agent_session_id.clone();
+                let agent_session_source = active.agent_session_source.clone();
+                let legacy = Arc::clone(&active.legacy);
+                let should_drain = match result {
+                    OpenAiTerminalResult::Completed { .. }
+                    | OpenAiTerminalResult::CompletedWithUsage { .. } => {
+                        legacy.enqueue(move || {
+                            log_legacy_backend_returned(
+                                request_id,
+                                operation,
+                                elapsed,
+                                agent_session_id.as_deref(),
+                                agent_session_source.as_deref(),
+                            );
+                        })
+                    }
+                    OpenAiTerminalResult::Failed { failure, .. }
+                        if failure == OpenAiFailure::Timeout =>
+                    {
+                        legacy.enqueue(move || {
+                            log_legacy_backend_timeout(
+                                request_id,
+                                operation,
+                                elapsed,
+                                agent_session_id.as_deref(),
+                                agent_session_source.as_deref(),
+                            );
+                        })
+                    }
+                    OpenAiTerminalResult::Failed { .. } => legacy.enqueue(move || {
+                        log_legacy_backend_error(
+                            request_id,
+                            operation,
+                            elapsed,
+                            agent_session_id.as_deref(),
+                            agent_session_source.as_deref(),
+                        );
+                    }),
+                };
                 Some((
                     backend_attempt.map(|(_, attempt_id)| attempt_id),
-                    active.started_at.elapsed(),
-                    active.agent_session_id.clone(),
-                    active.agent_session_source.clone(),
+                    legacy,
+                    should_drain,
                 ))
             } else {
                 None
             }
         };
-        let Some((attempt_id, elapsed, agent_session_id, agent_session_source)) = attempt else {
+        let Some((attempt_id, legacy, should_drain)) = attempt else {
             return;
         };
         match result {
@@ -605,13 +736,6 @@ impl OpenAiLifecycleLoggingAdapter {
                     self.service
                         .complete_attempt(request_id, attempt_id, Some(status_code));
                 }
-                log_legacy_backend_returned(
-                    request_id,
-                    operation,
-                    elapsed,
-                    agent_session_id.as_deref(),
-                    agent_session_source.as_deref(),
-                );
             }
             OpenAiTerminalResult::Failed { failure, .. } => {
                 if let Some(attempt_id) = attempt_id {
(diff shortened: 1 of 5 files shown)


crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -883,6 +1612,81 @@ mod tests {
...
+                move || order.lock().unwrap().push("backend_started")
+            }),
+        ));
+
+        let drain_custody = Arc::clone(&custody);
+        let drain_thread = thread::spawn(move || drain_custody.drain());
+        gate.wait();
+
+        let finish_order = Arc::clone(&order);
+        assert!(!custody.finish(move || { finish_order.lock().unwrap().push("request_finished") }));
```
ndizazzo, line 1640: P0, will red CI

This assertion is racy. `assert!(!custody.finish(...))` needs the drain thread to still be inside `drain()`, but the barrier only proves it entered the first action. Once the barrier releases, the drainer can pop `backend_started`, run it, find the queue empty, set `draining = false`, and return, all before the main thread gets to `finish()`, which then correctly returns `true`.

Production behaviour is fine either way (whoever gets `true` does the draining), so it's only the test that's over specified. I'd drop the return value assertions and keep the final `assert_eq!` on ordering, which is the property you actually care about. If you want to pin the handoff path specifically, hold the drainer open with a second barrier instead of relying on it being slow.

Rate on my machine, prebuilt binary filtered to these two tests: 59/200 at default threads, 97/300 at `--test-threads 8`.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -883,6 +1612,81 @@ mod tests {
...
+            }),
+        ));
+
+        let drain_custody = Arc::clone(&custody);
+        let drain_thread = thread::spawn(move || drain_custody.drain());
+        gate.wait();
+
+        let response_order = Arc::clone(&order);
+        assert!(
+            !custody.enqueue(move || { response_order.lock().unwrap().push("response_completed") })
```
ndizazzo, line 1674: P0, will red CI

Same root cause as the one in `legacy_custody_serializes_dispatch_and_terminal_race`, `assert!(!custody.enqueue(...))` assumes the drainer hasn't drained out yet. This one's worse: single threaded it fails basically every run (200/200 with `--test-threads 1`) because with no contention the drain thread reliably wins the race to empty the queue.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -769,6 +1111,368 @@ const fn operation_label(operation: OpenAiBackendOperation) -> &'static str {
...
+        "backend_timeout",
+        request_id,
+        operation,
+        elapsed,
+        agent_session_id,
+        agent_session_source,
+    );
+}
+
+fn log_legacy_backend_result(
```
ndizazzo, line 1239: P0, will red CI

`cognitive_complexity` trips at 44 against the configured threshold of 20, so this fails the clippy lane.

Past the lint though, dispatching on `event: &'static str` through a six arm match is doing a job an enum should do, and the `_ =>` fallback quietly maps anything unrecognised to "backend operation failed", which is the wrong default if someone adds an event later. A small `enum LegacyBackendResult { Returned, Timeout, Error }` matched against the session pair makes it total, and the complexity warning goes away for free.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -491,89 +669,200 @@ impl OpenAiLifecycleLoggingAdapter {
...
+                            log_legacy_backend_returned(
+                                request_id,
+                                operation,
+                                elapsed,
+                                agent_session_id.as_deref(),
+                                agent_session_source.as_deref(),
+                            );
+                        })
+                    }
+                    OpenAiTerminalResult::Failed { failure, .. }
```
ndizazzo, line 697: P0, will red CI

`clippy::redundant_guards` fires here under `-D warnings`. Folds into the pattern:

```rust
OpenAiTerminalResult::Failed { failure: OpenAiFailure::Timeout, .. } => { ... }
```

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -433,26 +565,48 @@ impl OpenAiLifecycleLoggingAdapter {
...
+        // record here so the target's authenticated frontend events remain
+        // observable without creating a second canonical lifecycle.
+        let raw_mesh_owned = self.raw_mesh_owners.is_claimed(request_id);
+
+        // Reserve the request under the small state mutex, then perform
+        // service registration outside it. This prevents persistence and
+        // tracing work from blocking every other OpenAI request.
+        {
+            let mut tracked = lock_recover(&self.tracked);
+            if tracked.requests.contains_key(&request_id) || !tracked.make_room() {
```
ndizazzo, line 579: P1, behaviour change against main

This early return changes what happens after remote tunnel suppression lifts. On main a suppressed admit inserted nothing, so a later re-admission for the same request id went on to call `register_request_with_metadata` and create the canonical parent. Now the suppressed admit leaves a legacy only entry behind, so the re-admission hits `contains_key` here and returns, and the canonical parent never gets created.

Intended? If yes that's fine, but `remote_tunnel_suppression_skips_canonical_parent_but_retains_legacy` doesn't catch it either way (see my other comment down there).

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -1184,7 +2184,7 @@ mod tests {
     }
 
     #[test]
-    fn remote_tunnel_suppression_skips_only_the_active_frontend_admission() {
+    fn remote_tunnel_suppression_skips_canonical_parent_but_retains_legacy() {
```
ndizazzo, line 2187: P1, behaviour change against main

This test doesn't test anything anymore. Both `assert_eq!(adapter.tracked_len(), 1)` calls are satisfied by the entry the first admit inserted, since the second admit early returns on `contains_key`, so the post `drop(lease)` assertion passes whether or not the re-admission does anything.

On main the same assertion was load bearing: the first admit inserted nothing (`tracked_len() == 0`), so `1` after the drop proved canonical registration actually happened. Whatever the intended behaviour is now, this needs an assertion that tells a legacy only entry apart from a canonical one, checking `guard.is_some()` or the registry rather than counting map entries.

`raw_mesh_owner_prevents_a_competing_embedded_frontend_parent` at :2084 went 0 to 1 for the same reason and has the same problem.

crates/openai-frontend/src/router.rs
```diff
@@ -856,6 +856,12 @@ async fn frontend_lifecycle_middleware(
     let uri = request.uri().clone();
     let context =
         OpenAiLifecycleContext::new(request_id, lifecycle_method(&method), lifecycle_route(&uri));
+    let context = match agent_session_from_header(&state.config, request.headers()) {
```
ndizazzo, line 859: P1, behaviour change against main

This only consults `agent_session_from_header` so it misses the Responses path. `resolve_agent_session` at :802 is what arbitrates between the trusted header and `AgentSessionSource::ResponsesConversation`, and a `/v1/responses` request carrying `conversation: {id}` gets a perfectly valid `AgentSessionIdentity` in the handler but `agent_session_id: None` in the lifecycle context, so no custody projection.

Same failure this PR is fixing, still live on that route. It also means the header gets parsed twice per request (here and again at :691), and the two parses can disagree.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -433,26 +565,48 @@ impl OpenAiLifecycleLoggingAdapter {
...
+        // service registration outside it. This prevents persistence and
+        // tracing work from blocking every other OpenAI request.
+        {
+            let mut tracked = lock_recover(&self.tracked);
+            if tracked.requests.contains_key(&request_id) || !tracked.make_room() {
+                return;
+            }
+            tracked
+                .requests
+                .insert(request_id, TrackedRequest::Admitting);
```
ndizazzo, line 584: P1, behaviour change against main

Worth thinking about the capacity interaction here. `make_room` only ever evicts `Terminal` entries, and raw mesh owned / tunnel suppressed requests now take a slot each where before they took none. On a host whose traffic is mostly raw mesh ingress through the embedded frontend, legacy only records can fill the 1024 budget and then `admit` starts dropping admissions, including the canonical ones for direct frontend traffic, which are the more valuable record.

Either give legacy custody its own budget, or make eviction prefer guardless entries when it's under pressure.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -769,6 +1111,368 @@ const fn operation_label(operation: OpenAiBackendOperation) -> &'static str {
...
+        openai_frontend::OpenAiFrontendRoute::Models
+        | openai_frontend::OpenAiFrontendRoute::Health
+        | openai_frontend::OpenAiFrontendRoute::Healthz
+        | openai_frontend::OpenAiFrontendRoute::Readyz => OpenAiBackendOperation::Models,
+        openai_frontend::OpenAiFrontendRoute::ChatCompletions
+        | openai_frontend::OpenAiFrontendRoute::Unknown => OpenAiBackendOperation::ChatCompletion,
+    }
+}
+
+fn log_legacy_request_started(
```
ndizazzo, line 1127: P2, question rather than a defect

Compared to `log_request_started` in 71e0857d1 this lost `model` and `stream`. Given the description frames the consumer as doing model performance custody, dropping `model` might defeat the purpose.

Awkward part is it isn't fixable here. `OpenAiLifecycleContext` intentionally has no payload fields, so the model name genuinely isn't available at this boundary. Which makes me wonder whether the projection belongs at the lifecycle observer layer at all, or whether the canonical events need a model carrying field that Cacheline can join on. Worth deciding before more gets built on this shape.

Same story for the rest: `backend_timeout` lost `timeout_ms`, `backend_error` and `stream_item_error` lost `error`, both stream events lost `item_count`.

crates/openai-frontend/src/common.rs
```diff
@@ -66,7 +66,10 @@ impl AgentSessionSource {
     #[must_use]
     pub(crate) fn label(&self) -> &str {
         match self {
-            Self::TrustedHeader(header) => header,
+            // The custody contract identifies the trust mechanism, not the
+            // operator-selected header name. The latter is configuration
+            // detail and must not become part of the durable provenance ABI.
+            Self::TrustedHeader(_) => "trusted_header",
```
ndizazzo, line 72: P2, question rather than a defect

Agree with the reasoning in the comment, the mechanism is the durable thing and the header name is config. Only one header can be configured at a time anyway, so the name's always recoverable from config.

Different concern on the same theme though: `logging/policy.rs:23` lists `x-session-id` under "sensitive header names that must never appear in logs", and the description says the canonical presentation "intentionally omits session identity". This now writes an operator configured session header's value into tracing fields, which bypasses `redact_headers` entirely (that helper works on header maps in the persistence path, not on direct macro calls).

I suspect it's a deliberate trade since custody needs identity, and the validation in `AgentSessionIdentity::new` closes the log injection angle nicely, control chars and >512 bytes both get rejected before anything reaches a log line. But it re-crosses a boundary the module explicitly documents, so I'd want that acknowledged in the description and pinned with a test rather than arrived at by side effect.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -769,6 +1111,368 @@ const fn operation_label(operation: OpenAiBackendOperation) -> &'static str {
...
+        TerminalOutcome::Completed
+        | TerminalOutcome::CompletedWithStatus(_)
+        | TerminalOutcome::CompletedWithUsage { .. } => "success",
+        TerminalOutcome::Failed(error) | TerminalOutcome::FailedWithStatus { error, .. }
+            if error == "timeout" =>
+        {
+            "timeout"
+        }
+        TerminalOutcome::Failed(_) | TerminalOutcome::FailedWithStatus { .. } => "backend_error",
+        TerminalOutcome::Rejected(_) | TerminalOutcome::RejectedWithStatus { .. } => "client_error",
```
ndizazzo, line 1470: P2, question rather than a defect

This arm can't be reached. `LegacyCustody::start` is only called from `backend_dispatched`, and `finish()` bails out when `!state.started` without pushing the action, so a request that never dispatched a backend produces no `request_finished` at all, and `Rejected` only ever arrives on that path.

Which means every pre-dispatch rejection (404, invalid request, guardrail reject, the header/conversation conflict) now emits zero custody records, where the original emitted `request_started` plus `request_finished{outcome: "client_error"}`. `pre_dispatch_rejection_does_not_create_legacy_custody` asserts this on purpose so I assume it's intentional, but it means client error rates are invisible to the collector. Worth stating in the description rather than leaving it to a test name.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -397,21 +399,151 @@ struct TrackedRequests {
...
+    backend_operation: Option<OpenAiBackendOperation>,
     backend_attempt: Option<(OpenAiBackendOperation, AttemptId)>,
     backend_stream_first_item: bool,
     usage: Option<OpenAiUsage>,
 }
 
+type LegacyAction = Box<dyn FnOnce() + Send + 'static>;
+
+#[derive(Default)]
+struct LegacyCustody {
```
ndizazzo, line 426: P2, question rather than a defect

Higher level thought on `LegacyCustody`: that's ~120 lines of hand rolled concurrency whose only job is ordering log lines. Every record already carries `elapsed_us` and every enqueue happens under the `tracked` mutex anyway, so a consumer could sort by timestamp without emission order being guaranteed.

If Cacheline genuinely depends on receiving these in order then keep it and say so in a comment, it's correct as written and I checked the handoff, there's no lost wakeup window. But if it doesn't, that's a lot of surface area to own for something the consumer doesn't need.

ndizazzo (changes requested):
Pulled this into a worktree and ran it. The canonical/legacy split reads cleanly and the lock discipline is careful: `tracked` then `LegacyCustody::state`, never the reverse, and `drain()` never touches the map lock. That part I'd merge happily.

Two things will red CI though, and there's a scope question worth settling before this goes non-draft.

Both new `legacy_custody_serializes_*` tests are timing dependent. Running the full package suite (`cargo test -p mesh-llm-host-runtime --lib`, 2527 tests) fails both. Running just `logging::openai_lifecycle` passes, which is why the validation in the description looks clean. I looped the prebuilt binary to get a rate: 59/200 at default threads, 97/300 at `-j8`, 200/200 at `--test-threads 1`. Details inline.

`cargo clippy --all-targets -- -D warnings` (what ci-quality-slice.yml:232 runs) gives two warnings, both in `openai_lifecycle.rs`: a `redundant_guards` at :697 and `cognitive_complexity` 44/20 on `log_legacy_backend_result`. That second one is one of the two lints AGENTS.md calls out by name.

On scope: I diffed the new projection against the original in 71e0857d1 and it's a strict subset. `request_started` lost `model` and `stream`, `backend_timeout` lost `timeout_ms`, `backend_error` and `stream_item_error` lost `error`, both stream events lost `item_count`. For a model performance custody collector, `model` seems like the one that matters, and it isn't recoverable at this layer since `OpenAiLifecycleContext` deliberately carries no payload fields. Anything rejected before dispatch also emits nothing at all now, where the old code emitted a `client_error` pair.

So what does Cacheline actually read? If it needs `model` then this doesn't unblock the benchmark and we should talk about where the projection lives. If it doesn't, say so in the description so nobody assumes "restore" means byte parity.

Inline notes are tagged P0 through P2. GitHub orders them by file position rather than priority so the tags are there to sort by. P0 is the four CI ones, all mechanical. P1 is behaviour that changed against main. P2 is questions I'd want answered before merge rather than changes I'm sure about. A few nits left out, happy to add them once the bigger stuff settles.

One more that has no line to hang off, since the file isn't in the diff. `runtime_tracing_subscriber` in `crates/mesh-llm-host-runtime/src/runtime/tracing_writer.rs:300` is `EnvFilter::from_default_env()` plus directives for `mesh_inference`, `nostr_relay_pool`, `nostr_sdk`, and `noq_proto::connection`. Nothing matches `mesh_openai_observability`, and `from_default_env()` with no `RUST_LOG` set defaults to `ERROR`, so these `info!` events get filtered before they reach the writer. Not new (the original didn't add a directive either, so presumably the deployment sets `RUST_LOG`), but since the success criterion is "Cacheline's collector receives records again" it's worth confirming. Adding `mesh_openai_observability=info` to that directive list would make the fix self contained instead of depending on deploy config.

File's at 2260 lines now too, over the 2,000 cap in AGENTS.md.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -769,6 +1111,368 @@ const fn operation_label(operation: OpenAiBackendOperation) -> &'static str {
...
+    {
+        tracing::info!(
+            target: LEGACY_OBSERVABILITY_TARGET,
+            event = "stream_first_item",
+            request_id = %request_id,
+            operation,
+            agent_session_id,
+            agent_session_source,
+            elapsed_us = time_to_first_item_us,
+            time_to_first_item_us,
```
i386, line 1329: Nit: this emits both `elapsed_us` and `time_to_first_item_us` with the same value. If it's a deliberate back-compat alias, a one-line comment saying which one is canonical (and when the alias can be dropped) would help; otherwise drop the duplicate.

crates/mesh-llm-host-runtime/src/logging/openai_lifecycle.rs
```diff
@@ -397,21 +399,151 @@ struct TrackedRequests {
...
+        if state.started || state.terminalized {
+            return false;
+        }
+        state.started = true;
+        state.actions.push_back(request_started);
+        state.actions.push_back(backend_started);
+        Self::begin_draining(&mut state)
+    }
+
+    fn enqueue<F>(&self, action: F) -> bool
```
i386, line 483: Nit: `enqueue` guards on `terminalized` but not `started` — if a stream/usage event could ever arrive before `backend_dispatched`, its action would drain ahead of `request_started`/`backend_started`. The current call ordering makes that unreachable (dispatch always precedes first-item/usage/terminal), but an explicit `debug_assert!(state.started)` or an early-return would document the invariant.

i386 (commented):
Reviewed (can't formally approve — this is my own PR). The restore looks correct and well-tested:

- `LegacyCustody` serializes the two start events atomically and guarantees `request_finished` is last; the dispatch/terminal and response/terminal race tests pin that ordering.
- `admit` now reserves under the small state mutex before service registration, so persistence/tracing work no longer blocks every other request — good contention fix.
- Raw-mesh-owned and tunnel-suppressed requests correctly keep a legacy-only record (`guard: None`) without registering a second canonical parent, and the pre-dispatch rejection path emits nothing.
- `AgentSessionSource::label` switching from the configured header name to the fixed `"trusted_header"` is a deliberate provenance-ABI change — worth calling out in the PR body since downstream consumers keyed on the old header-name label will silently change behavior.

Non-blocking nits below. Note: this PR is currently conflicting with main and will need a sync before merge.

PR #1391, opened 2026-08-20
fix(moa): prevent reasoning-only model poisoning

michaelneale:
## Summary

- force thinking off for every MoA worker, actor, and reducer request across boolean, effort, and budget request surfaces
- retry without all injected thinking fields when a backend rejects compatibility fields
- reject reasoning-only backend responses so a single model's private plan cannot poison a worker/reducer hedge
- preserve valid answer content and native tool calls even when reasoning metadata is also present

## In-situ regression

Re-ran the same real coding-agent loop against the patched release bundle on the private two-node Mac lab mesh: inspect a project, read contract/implementation/tests, diagnose a capability-routing bug, edit it, and run the complete test file.

The first MoA response was now a structured `list_files` tool call (not `I'll start by inspecting...` prose). It completed 7 turns, including reading all three files, writing the minimal `intersection` -> `issuperset` fix, recovering from unavailable pytest by using unittest, and reporting success. Independent verification passed all 3 tests.

Bundle SHA-256: `b9cf093239161a75b63f92b580a39920c36a8f30e89ee767be829f3f0c533a51`.

## Verification

- `cargo test -p mesh-mixture-of-agents` (192 unit tests plus integration suites; live network tests ignored)
- `cargo fmt --all --check`
- full release `just build`
- `just bundle /tmp/jimmy-moa-b986b7318.tar.gz`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry handling when reasoning is disabled.
  * Prevented responses containing only internal reasoning from being accepted.
  * Allowed reasoning-only responses to fall through to another available response.
  * Preserved support for standard text responses and tool calls.

* **Tests**
  * Added coverage for reasoning settings, retry behavior, and reasoning-only responses.

### DIFF
--- a/crates/mesh-mixture-of-agents/src/backend.rs
+++ b/crates/mesh-mixture-of-agents/src/backend.rs
@@ -155,6 +155,7 @@ impl ModelBackend for HttpBackend {
                 let mut retry_body = body.clone();
                 if let Some(obj) = retry_body.as_object_mut() {
                     obj.remove("reasoning_effort");
+                    obj.remove("thinking_budget");
                     obj.remove("chat_template_kwargs");
                 }
                 let retry = self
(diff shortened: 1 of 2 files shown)


ndizazzo (approved):
Seems reasonable to me! How did you notice this at first, what were the symptoms?

michaelneale:
@ndizazzo was so silly - trying with same glimmer and another model and one would jabber on with thinking tokens in response and basically ruin the whole consensus. Vs talking to the models directly. So silly really.

PR #1402, opened 2026-08-21
Advance llama.cpp and simplify the Skippy patch queue

i386:
## Summary

Mesh now builds its native Skippy runtime against the latest llama.cpp source while retaining staged-serving behavior.

- advance the llama.cpp pin from 030ebb558a5820b444a8f836ed5cdd46c9b4bd7a to 3af988fabcf79fd81f8720505e684d2aa5bfc786
- rebase and reduce the durable queue from 24 patches to 15 capability-owned patches
- fold chat-template, public-header documentation, MTP, ISWA, recurrent-state, transfer, and compatibility follow-ups into their owning patches
- remove Metal TQ2 and get-rows shader pieces now supplied upstream
- retain current upstream architecture, metadata-key, tensor, and RoPE coverage alongside the custom Inkling architecture

## Validation

- just build
- cargo test -p mesh-llm --lib (48 passed)
- python3 scripts/generate-skippy-api-doc.py --check
- clean replay of all 15 patches from the new pin
- Metal model-load and OpenAI completion smoke with SmolLM2-135M-Instruct-Q4_K_M-GGUF (HTTP 200, 13 total tokens)

No mesh wire-protocol or Skippy ABI version change.

## Summary by CodeRabbit

- **New Features**
  - Added support for Inkling text, image, and audio models.
  - Introduced staged inference for partial-layer execution and activation workflows.
  - Added the Skippy runtime API for model loading, sessions, devices, tokenization, sampling, and chat.
  - Added multi-part model loading, session state management, prefix caching, and KV-cache transfer.
  - Added integrated and external multi-token prediction support.
- **Improvements**
  - Enhanced chat templates, reasoning controls, tool-argument validation, and response parsing.
  - Improved recurrent-state handling, speculative decoding, and runtime compatibility.


i386 (commented):
## Review — Advance llama.cpp and simplify the Skippy patch queue

The structural consolidation is sound: 24 → 15 patches, the folding order matches the expected cumulative state of the old chain, and patch numbering is clean and sequential (0001–0015, no gaps or duplicates). The recurrent-state seq-id remapping removal is safe — the new roundtrip test proves `state_read` uses the `dest_seq_id` parameter rather than the embedded field. The transposed-V coalescing is also correct: consecutive cells at a fixed embedding dimension are contiguous in memory.

Seven things worth looking at:

---

### 1 — Inkling + thinking-disabled: wrong type delivered for `reasoning_effort`

**`0010-Add-Skippy-tokenization-and-stage-chat.patch` → `common/stage-chat.cpp`**

`skippy_apply_chat_template_json` stores user-supplied kwargs as `value.dump()` (JSON-encoded strings: `"\"none\""`, `"\"high\""`, `"\"max\""`). When thinking is forcibly disabled the system override writes a raw string instead:

```cpp
inputs.chat_template_kwargs["reasoning_effort"] = "0";  // raw "0", not json("0").dump()
```

If the Jinja renderer parses kwargs values as JSON (which agent-1 believes it does via `json::parse`), user-supplied values decode to strings (`"none"`, `"high"`) while the system override decodes to the integer `0`. The Inkling-template tests only exercise the user-kwargs path (`render_effort`); the `render_disabled` regression test covers Qwen and gpt_oss but **not Inkling**. If the Inkling template branches on `reasoning_effort == "none"`, a disabled-thinking call with an Inkling model could silently fall through to the max-effort branch and output `Thinking effort level: 0.99` instead of suppressing thinking.

Fix: change the override to `json(0).dump()` (→ `"0"`, a JSON number, consistent with what the renderer expects) or `json("none").dump()` (→ `"\"none\""`) and add an Inkling case to `render_disabled`.

---

### 2 — `skippy_model_attach_mtp_draft_model`: new hard error for null/default config

**`0012-skippy-retain-and-execute-final-stage-MTP-tensors.patch` → `src/skippy/model_loading.cpp`**

```cpp
if (config == nullptr || config->mtp_source != SKIPPY_MTP_SOURCE_EXTERNAL) {
    skippy_set_error(out_error, SKIPPY_STATUS_INVALID_ARGUMENT,
        "MTP draft attachment requires an external MTP source");
    return SKIPPY_STATUS_INVALID_ARGUMENT;
}
```

The consolidated patch requires `config` to be non-null **and** `mtp_source == SKIPPY_MTP_SOURCE_EXTERNAL`. `mtp_source` is a C enum with `DISABLED = 0` as the zero value, so any zero-initialised `skippy_runtime_config` silently fails. Similarly, `skippy_enable_integrated_mtp_tensors` now gates on `mtp_source == SKIPPY_MTP_SOURCE_INTEGRATED`; callers that previously relied on `include_output = true` being sufficient will silently get `load_mtp = false`.

No Rust-side changes appear in this diff. Worth confirming explicitly that the Rust config builder was updated in a prior commit and that there's no surviving call site that passes `nullptr` or a default-zero config to `skippy_model_attach_mtp_draft_model`.

---

### 3 — `llama_kv_cache_cell_runs`: sorted input assumed, no assertion

**`0013-skippy-support-composite-ISWA-KV-pages.patch` → `src/llama-kv-cache.cpp`**

```cpp
static std::vector<llama_kv_cache_cell_run>
llama_kv_cache_cell_runs(const std::vector<uint32_t> & cell_idxs) {
    for (uint32_t cell_idx : cell_idxs) {
        if (!runs.empty() && cell_idx == runs.back().first + runs.back().count)
            ++runs.back().count;   // only merges if exactly the next index
        else
            runs.push_back({ cell_idx, 1 });
    }
}
```

This only coalesces runs when `cell_idxs` is sorted ascending. An unsorted input (e.g. `[3, 1, 2]`) produces separate single-cell runs that map to the wrong memory offsets. The callers feed `sinfo.idxs[0]`, which is probably always sorted, but there is no `GGML_ASSERT(std::is_sorted(cell_idxs.begin(), cell_idxs.end()))` and no comment citing the invariant. If the function is reused in a future path that doesn't go through `apply_ubatch`, export/import silently corrupts KV data without any error.

---

### 4 — Metal link risk: `ggml_metal_library_get_pipeline_lightning_indexer` depends on new upstream

**`0002-Add-GLM-DSA-backend-execution-support.patch`**

Patch 0002 calls `ggml_metal_library_get_pipeline_lightning_indexer`; the defining code that lived in the old Skippy custom patches (removed by old 0015) is now absent from the queue entirely, with the PR relying on the new upstream base (3af988f) to provide it. The Metal smoke test passes, so this is likely fine, but it's worth a quick `grep` in the pinned upstream to confirm the symbol is present rather than relying on the smoke test alone.

---

### 5 — Dead comparison: `model_params.load_mtp = config->mtp_source == SKIPPY_MTP_SOURCE_EXTERNAL`

**`0012-skippy-retain-and-execute-final-stage-MTP-tensors.patch` → `src/skippy/model_loading.cpp`**

After the guard at the top of `skippy_model_attach_mtp_draft_model` returns early unless `config != nullptr && config->mtp_source == SKIPPY_MTP_SOURCE_EXTERNAL`, the comparison on the assignment line is a tautology — always `true`. The surrounding `if (config != nullptr)` block is also dead. Not a correctness bug, but a maintenance trap: a future author softening the guard could miss that this assignment and null-check are now unconditional.

```cpp
// cleaner:
model_params.load_mtp = true;
// and remove the if (config != nullptr) wrapper
```

---

### 6 — DeepSeek4 silently restored to `llm_arch_supports_rs_rollback`

**`0001-Add-staged-model-graph-and-family-support.patch` → `src/llama-arch.cpp`**

The old custom patch explicitly removed `LLM_ARCH_DEEPSEEK4` from the `llm_arch_supports_rs_rollback` switch (presumably because the old upstream's DeepSeek4 context didn't implement it). The rebased 0001 no longer removes it, so DeepSeek4 again returns `true` from that function. If the new upstream's DeepSeek4 hybrid path implements recurrent rollback correctly, this is fine. If not, a rollback call silently no-ops with potential state corruption. This path is disabled in CI (`arch_supported()` returns `false` for DeepSeek4), so it isn't caught by the test suite.

---

### 7 — Patch subject mismatch in 0013

**`0013-skippy-support-composite-ISWA-KV-pages.patch`**

Subject says `"skippy: support composite ISWA KV pages"` but the patch now also contains the recurrent-state metadata fix (from old 0020) and the KV cell-run coalescing (from old 0021), plus the new `test-skippy-recurrent-state-roundtrip.cpp` test. The subject should be updated to reflect the broader scope for accurate `git bisect` / blame results.

---

**Not flagged** (verified clean): the V-transposed KV coalescing is correct (consecutive cells at fixed embedding dim are contiguous); the recurrent-state seq-id remap removal is safe (confirmed by the new roundtrip test); patch numbering is sequential 0001–0015 with no gaps.

ndizazzo (approved):
nice! Inkling

PR #1405, opened 2026-08-22
fix: close the four rc6 release-validation defects

ndizazzo:
Fixes the four defects rc6 release validation left open (`RV-DEFECT-001` through `004`). Each one was reproduced first, then fixed.

`RV-DEFECT-001` is the S1 that blocked the `NOT_READY` gate on its own.

## What was wrong

| ID | Sev | Defect | Reproduction used here |
|---|---|---|---|
| `RV-DEFECT-001` | **S1** | `just release-build-aarch64-cuda` picked its SM list with `[[ ]]` inside an `@`-prefixed recipe, so it ran under `sh`. On dash the test is "not found", the `else` branch wins, and a CUDA 13 build silently gets the pre-13 list — which still contains the removed `compute_61`, so nvcc then hard-fails. `build-runtime` had a second, separate bug in the same class: `"$$backend"`. | Recipe body executed under `dash` against a stubbed packager: CUDA 13 selected `61;75;80;86;87;89;90` instead of `75;80;86;87;89;90;110`. `just` 1.58 confirmed to expand `$$FOO` to `<pid>FOO`, so `build-runtime` handed the packager a nonsense backend name. |
| `RV-DEFECT-002` | S2 | A clean checkout would not build under a host-default pnpm 9, and nothing in the repo said pnpm 10 was needed. CI hid it with a `version: 10` pin that lived only in the workflows. | `corepack pnpm@9.15.9 install --frozen-lockfile` against the real manifest → `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`. Cause: `overrides`/`allowBuilds` sit in `pnpm-workspace.yaml`, a pnpm 10 layout that pnpm 9 does not read. |
| `RV-DEFECT-003` | S2 | A remote (cross-node) attempt that lost its client never tore down the QUIC tunnel, so a worker on another node kept generating for a client that had already hung up. The local arm always shut its upstream down; the two arms hand-rolled that teardown separately, which is how they came to disagree. | rc6 request `55151d55`: gateway logged `request_dropped` at 19:29:35.984, the worker did not log `request_stream_completed` until 19:29:36.006. |
| `RV-DEFECT-004` | S3 | The plugin control socket path was built unconditionally under `$HOME`, with no guard against the `sun_path` limit (108 bytes on Linux, 104 on macOS). `bind` refuses an over-long path rather than truncating, with an error that never mentions lengths. | rc6 saw three generated paths at exactly 112 bytes; `blobstore` failed to load with a bare "Failed to bind plugin socket". |

## What changed

**001** — both recipes become `#!/usr/bin/env bash` script recipes, matching what `release-build-cuda` and `release-runtime-build` already do. `sm_110` is gated on toolkit major `>= 13`, the same shape as the x86_64 Blackwell gate. As a `-ge` comparison rather than a `13.*` glob, a future CUDA 14 keeps Thor instead of falling back to the pre-13 list — the old form would have regressed there.

**002** — `engines.pnpm: ">=10"` plus `engine-strict`, so pnpm enforces it instead of treating it as a comment. A pnpm 9 operator now gets `ERR_PNPM_UNSUPPORTED_ENGINE / Expected version: >=10 / Got: 9.15.9`. `packageManager` also lets corepack select the right pnpm on its own.

**003** — the disconnect decision moves into one place (`cancellation.rs`), and each arm gets an implementation of how to say "stop" over its own protocol: `shutdown()` for a local TCP upstream, `stop()` (QUIC `STOP_SENDING`) for a peer stream. The peer's next write then fails, which is what actually ends its generation loop.

**004** — falls back to a short directory when the preferred path does not fit, which is what `mesh_llm_plugin::bind_side_stream` already does for side streams. The bound path is handed to the plugin verbatim via `LocalListener::endpoint()`, so relocating it needs no matching change on the plugin side. If even the short path cannot fit, that is now a clear error naming both candidates and their byte counts.

## Tests

Written before the fixes, and each confirmed red against `main` first:

- A Justfile-wide ratchet asserting no default-shell recipe uses bash-only syntax or `$$`. Probed before adding: it catches exactly the two recipes fixed here and nothing else.
- Behavioral tests that execute the real recipe bodies with the interpreter the recipe declares — `dash` for a plain recipe — against a stubbed packager, checking the SM list across 12 / 12.4 / 12.8 / 13 / 13.1.2 / 14 and the backend `build-runtime` actually forwards.
- The declared pnpm major is asserted to track every `pnpm/action-setup` version in CI, so the two pins cannot drift apart silently.
- `route_remote_attempt_after_forward` is now generic over its upstream, so the disconnect path is driven directly: a scripted upstream serves headers then fails the body read with `ConnectionReset`, and the test asserts the tunnel was cancelled exactly once. A delivery counterpart asserts a normal response cancels nothing.
- Socket-path selection is a pure function over its two candidate directories, so the limit is covered without mutating `$HOME` — the repository's environment-mutation contract pins an exact census of `set_var` sites.

Verified locally: `cargo test -p mesh-llm-host-runtime --lib` (2538 passed), `python3 -m unittest discover -s scripts/tests` (530 passed), `cargo clippy --all-targets`, `cargo fmt --all --check`, `just no-console-print`, and Prettier on the changed manifest.

## Notes for the reviewer

- **Overlaps #1404**, which fixes `RV-DEFECT-001` alone with substantially the same approach. This PR was asked for as a single consolidated change across all four; whichever lands first, the other should rebase.
- **CI never exercised the 001 recipes** — `release.yml` sets the arch list from its build matrix and invokes the composite action directly, which is why this went unnoticed. So CI going green here is not evidence for that fix; the dash reproduction is.
- **Not built end to end on a Jetson.** The shell defect and the arch selection are proven, the full aarch64 CUDA build is not.
- **One thing deliberately left alone:** the `[windows]` `with-lld` recipe passes a PowerShell script full of `$$` sigils through the default shell. It is Windows-only, not exercised by this repo's CI, and its correct form depends on which shell just selects on Windows — a separate question from the dash portability fixed here. It is explicitly exempted in the new ratchet with that reasoning rather than silently swept in.

## Summary by CodeRabbit

* **New Features**
  * Improved handling when clients disconnect during streamed responses, stopping upstream processing promptly.
  * Added safer Unix socket path handling with automatic fallback for long paths.

* **Bug Fixes**
  * Improved CUDA architecture selection for current and future toolkit versions.
  * Corrected shell variable handling in development builds.

* **Documentation**
  * Updated UI setup requirements to Node.js 24 and pnpm 10 or newer.

* **Chores**
  * Enforced the required pnpm version for UI development.
  * Improved build and release validation for more reliable packaging.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,9 @@ This file covers local build and development workflows for this repository.
 - `just`
 - `cmake`
 - Rust toolchain (`cargo`)
-- Node.js 24 + npm (for UI development)
+- Node.js 24 + pnpm 10 or newer (for UI development). The UI lockfile keeps
+  `overrides` in `pnpm-workspace.yaml`, which pnpm 9 does not read, so pnpm 9
+  cannot install it. `corepack pnpm@10` is enough if your host pnpm is older.
 
 **macOS**: Apple Silicon. Metal is used automatically.
 
--- a/Justfile
+++ b/Justfile
@@ -151,9 +151,13 @@ build-linux backend="" cuda_arch="" rocm_arch="":
 # Backward-compatible spelling for the explicit native-runtime primitive.
 [linux]
 build-runtime backend="" cuda_arch="" rocm_arch="":
-    @backend="{{ backend }}"; \
-      [[ -n "$$backend" ]] || backend=cpu; \
-      LLAMA_STAGE_CUDA_ARCHITECTURES="{{ cuda_arch }}" LLAMA_STAGE_AMDGPU_TARGETS="{{ rocm_arch }}" scripts/package-native-runtime.sh --build --backend "$$backend"
+    #!/usr/bin/env bash
+    set -euo pipefail
+    backend="{{ backend }}"
+    [[ -n "$backend" ]] || backend=cpu
+    LLAMA_STAGE_CUDA_ARCHITECTURES="{{ cuda_arch }}" \
+      LLAMA_STAGE_AMDGPU_TARGETS="{{ rocm_arch }}" \
+      scripts/package-native-runtime.sh --build --backend "$backend"
 
 # Build release artifacts for the current platform.
 
@@ -189,11 +193,22 @@ release-build-aarch64: release-host-build
 # Build a Linux aarch64 CUDA release artifact (Jetson/Orin).
 # SM arches selected by MESH_CUDA_VERSION env (set by CI matrix); a bare
 # `just` invocation with the var unset detects the installed toolkit instead
-# of assuming one (see scripts/detect-cuda-toolkit-version.sh).
+# of assuming one (see scripts/detect-cuda-toolkit-version.sh). Thor (sm_110)
+# needs toolkit >= 13 -- gate on the major, the same shape as the x86_64
+# recipe's Blackwell gate, so a future CUDA 14 keeps Thor instead of falling
+# back to the pre-13 list the way a `13.*` glob did.
 release-build-aarch64-cuda: release-host-build
-    @cuda_version="${MESH_CUDA_VERSION:-$(scripts/detect-cuda-toolkit-version.sh)}"; \
-      MESH_LLM_CUDA_TOOLKIT_MAJOR="${MESH_LLM_CUDA_TOOLKIT_MAJOR:-${cuda_version%%.*}}" \
-      LLAMA_STAGE_CUDA_ARCHITECTURES="$(if [[ "$cuda_version" == 13.* ]]; then echo '75;80;86;87;89;90;110'; else echo '61;75;80;86;87;89;90'; fi)" \
+    #!/usr/bin/env bash
+    set -euo pipefail
+    cuda_version="${MESH_CUDA_VERSION:-$(scripts/detect-cuda-toolkit-version.sh)}"
+    major="${cuda_version%%.*}"
+    if [[ "$major" -ge 13 ]]; then
+      arches='75;80;86;87;89;90;110'
+    else
+      arches='61;75;80;86;87;89;90'
+    fi
+    MESH_LLM_CUDA_TOOLKIT_MAJOR="${MESH_LLM_CUDA_TOOLKIT_MAJOR:-$major}" \
+      LLAMA_STAGE_CUDA_ARCHITECTURES="$arches" \
       scripts/package-native-runtime.sh --build --backend cuda --target aarch64-unknown-linux-gnu
 
 # Prepare the pinned llama.cpp checkout and apply the Mesh-LLM ABI patch queue.
--- a/crates/mesh-llm-host-runtime/src/network/openai/response.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/response.rs
@@ -1,3 +1,4 @@
+mod cancellation;
 mod common;
 mod dispatch;
 mod external_endpoint;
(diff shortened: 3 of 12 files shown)


michaelneale:
Oh very cool @ndizazzo - tour de force of fixes.

michaelneale:
Reviewed at `1d57586b4` (worktree off `pull/1405/head`). Ran the script suite locally with `just` on PATH: `test_justfile_release_runtime` + `test_justfile_shell_portability` = 12 tests OK, `test_build_ui` 8 OK. (The other errors I see locally are pre-existing on `main` too — `plan-ci` fixtures that need a real workspace crate list — not from this PR.)

Build/tooling parts look right to me: converting both recipes to `#!/usr/bin/env bash` script recipes, gating Thor on `major >= 13` rather than a `13.*` glob, and `test_justfile_shell_portability.py` as a standing guard (with `$(( ))` explicitly exempted) is the right shape. `engines.pnpm` + `engine-strict=true` + `packageManager` pin, cross-checked against `pnpm/action-setup` in CI, is a good closed loop.

**One substantive question on the remote-cancellation fix (RV-DEFECT-003), which I think needs answering before this is claimed as a behaviour fix.**

`quic_recv` is a local in `route_remote_attempt` (`routing.rs:120`), and on the `ClientDisconnected` path the result is returned straight up and the stream is dropped immediately at the end of that function. `RecvStream::drop` already sends STOP_SENDING with error code **0** when `!all_data_read`:

```rust
// noq-1.1.1/src/recv_stream.rs:605-631
impl Drop for RecvStream {
    fn drop(&mut self) {
        if self.all_data_read { ... return; }
        ...
        let _ = conn.inner.recv_stream(self.stream).stop(0u32.into());
```

and the new impl sends `stop(0)` too. So as far as I can trace, the explicit `stop()` and the implicit drop send *the same frame with the same code, microseconds apart*. That would make this a no-op on the wire rather than a fix — which matters, because the reported symptom (worker logging `request_stream_completed` after the gateway logged `request_dropped`) then still has an unexplained cause.

Two possibilities I can't distinguish from source:

1. The drop-STOP_SENDING **is** being sent and the worker ignores it. On the worker side the signal has to travel: STOP_SENDING → `relay_tcp_to_quic`'s `quic_send.write_all` fails (`network/tunnel.rs:316`) → task ends → `finish_relay_pair` waits for the *other* direction (`:290-302`) → `handle_inbound_http_stream` returns → the local TCP stream to the API proxy drops. If that chain stalls (e.g. `relay_quic_to_tcp` is still blocked), generation keeps running regardless of when we cancel. In that case the fix is on the worker, not the gateway.
2. Something I've missed keeps `quic_recv` alive past the return, in which case the explicit stop genuinely helps and it's worth saying so in the comment.

The unit tests can't tell these apart — both use a recorder that counts `cancel()` calls, so they prove the call site fires, not that the peer stops. Could you either (a) show a path where drop wouldn't already have sent it, or (b) reframe the change as making an implicit drop-time cancel explicit (which is still defensible — it removes a dependency on quinn/noq drop semantics and pairs the arms) and keep RV-DEFECT-003 open pending a worker-side check?

The refactor itself is good either way: one `CancelUpstream` trait, `cancel_upstream_if_client_disconnected` applied to both arms, and the exhaustive "every other outcome leaves the upstream alone" test are exactly what stops the two arms drifting again.

**Smaller notes, none blocking:**

- `SUN_PATH_CAPACITY`: the `/tmp` fallback is consistent with `mesh_llm_plugin::bind_side_stream` (`crates/mesh-llm-plugin/src/io.rs:183`), so no new precedent — but `bind_local_listener` does `if path.exists() { remove_file }` before bind, and in a shared `/tmp` that's now on a world-writable directory. The `p{pid}-{random:08x}` name makes it hard to pre-place, so I'd call it acceptable; worth a line in the doc comment that the fallback is deliberately unlinking a path in a shared directory.
- `_version_satisfies_range` raising on `^`/`~`/`||` rather than silently accepting is the right call.
- `DOLLAR_DOLLAR_EXEMPT = {"with-lld"}` — good that it's an explicit named exemption with a reason rather than a pattern tweak.

PR #1409, opened 2026-08-22
feat(skippy): add run-ahead verify windows

danielwinterw:
# Run-ahead verify-window admission, stale-tail discard, and a jitter-capable wire model

Builds on the pipelined-split fixes. Three pieces, all opt-in:

## 1. Wire conditioning grows a jitter model

`WireCondition` gains an exponentially distributed per-message delay
(`--downstream-wire-jitter-ms`, mean) and probabilistic burst stalls
(`--downstream-wire-stall-ms` / `--downstream-wire-stall-p`), so benches can
model contended links (Wi-Fi, WAN) instead of a constant-latency pipe. FIFO
delivery is preserved (head-of-line blocking, like a real ordered transport);
the async-forwarder path samples per job at enqueue so propagation still
overlaps compute. Env: `MESH_LLM_BENCH_DOWNSTREAM_WIRE_{JITTER_MS,STALL_MS,STALL_P}`.

## 2. Run-ahead admission (`verify_window.runahead_max_tokens`)

The pipelined scheduler can admit by speculative-token budget instead of a
fixed window count: dispatch keeps filling while in-flight tokens stay under
the budget, capped at the native checkpoint-retention bound
(`MAX_VERIFY_WINDOW_PIPELINE_DEPTH = 64` windows). Config plumbed end to end
(`verify_window_runahead_tokens` in model config -> schema -> validation ->
resolver). Zero keeps today's fixed-depth behavior.

## 3. Stale-tail cancellation (`DiscardStaleWindows`)

At larger budgets, executing the stale tail after a rejection is the dominant
recovery cost (visible below: depth 3 is slower than depth 2). On divergence
the driver now sends a `DiscardStaleWindows` control message (window-id range
in the token sideband). Each stage connection gains a reader thread that
parses inbound messages ahead of execution and records discard ranges the
moment they are read; buffered stale windows are answered with an empty
`PredictedTokens` reply instead of executing. Middle stages forward the
discard and still execute (their forwarded activations must stay valid); the
final stage skips. Only sent in run-ahead mode, so fixed-depth setups keep
today's wire behavior byte-for-byte. (No subprotocol feature negotiation yet
- acceptable while the sender is opt-in; flagging for review.)

## Numbers

2-process loopback Qwen3-8B split, standalone suffix (5/32/48, window 32),
temp-0 ~330-token re-emit, 700 max tokens, median of 3, outputs checked
byte-exact. Jitter = 3ms constant + exp(5ms) + 2% x 40ms stalls per message:

| arm                 | clean tok/s | jitter tok/s |
|---------------------|------------:|-------------:|
| speculation off     |        17.4 |         13.5 |
| suffix, depth 1     |        24.6 |         19.9 |
| suffix, depth 2     |        99.6 |         88.3 |
| suffix, depth 3     |        94.8 |         84.9 |
| suffix, runahead 96 |       108.3 |         97.1 |

Run-ahead beats every fixed depth in both conditions; depth 3 < depth 2 is
the stale-tail cost, which the discard removes (telemetry: 3 windows / 97
tokens in flight against the 96-token budget, accept 0.89). Loopback only so
far - the widening-gap-with-RTT curve still wants a two-box run.

## Summary by CodeRabbit

- **New Features**
  - Added configurable speculative decoding run-ahead token budgets with validation and model-level overrides.
  - Improved verify-window scheduling with token-aware capacity management.
  - Added automatic stale verification-window handling and control messages.
  - Added configurable downstream network jitter and probabilistic stalls for testing.
  - Improved connection shutdown and cancellation behavior.

- **Compatibility**
  - Advanced stage protocol compatibility to generation 5.

- **Documentation**
  - Updated protocol and testing documentation for generation 5 compatibility.


crates/skippy-protocol/src/binary/types.rs
```diff
@@ -58,6 +58,7 @@ pub enum WireMessageKind {
     DecodeLightCtx = 9,
     VerifyWindow = 21,
     RetireVerifyWindow = 22,
+    DiscardStaleWindows = 23,
```
i386, line 61: [P1] Negotiate this wire kind before sending it. This adds kind 23 while `STAGE_STATE_VERSION` remains 11 and `STAGE_PROTOCOL_GENERATION` remains 4, so a released/current stage peer advertising generation 4 is still eligible. Once run-ahead diverges, the new coordinator sends this kind and the older `WireMessageKind::try_from` rejects it as `unknown stage message kind`, tearing down the request connection. The sender being opt-in does not establish receiver support. Please either advertise/gate on a dedicated discard capability across every stage, or bump the stage generation and make split planning exclude older peers.

crates/mesh-llm-config/src/model_validation.rs
```diff
@@ -633,6 +633,12 @@ fn validate_verify_window_controls(
         &format!("{base_path}.verify_window_pipeline_depth"),
         1,
         u32::try_from(MAX_VERIFY_WINDOW_PIPELINE_DEPTH).expect("verify depth limit fits u32"),
+    )?;
+    validate_optional_u32_range(
+        config.verify_window_runahead_tokens,
+        &format!("{base_path}.verify_window_runahead_tokens"),
+        1,
```
i386: [P1] Allow the documented zero sentinel here. `VerifyWindowConfig` defines `runahead_max_tokens == 0` as fixed-depth mode, but the persisted model setting rejects zero. In particular, if `[defaults.speculative]` enables run-ahead, a model-specific block cannot turn it back off: `None` inherits the positive default and `Some(0)` fails validation. Please validate `0..=MAX_VERIFY_WINDOW_RUNAHEAD_TOKENS` and add a precedence test with a positive default overridden by model-level zero.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
```diff
@@ -2,33 +2,84 @@ use anyhow::{Context, Result};
...
-                && observed_message_count == 0 =>
-        {
-            Ok(None)
+    capacity: usize,
+    registry: Arc<StaleDiscardRegistry>,
+) -> Result<InboundMessageReader> {
+    let mut reader = upstream
+        .try_clone()
+        .context("clone upstream stream for inbound message reader")?;
+    let (sender, receiver) = mpsc::sync_channel(capacity.max(1));
```
i386: [P1] This bounded queue prevents the reader from reaching the discard for the larger run-ahead windows this PR enables. `capacity` is `max_inflight` (and is capped to `lane_count`, commonly 1-4), while the scheduler can enqueue up to 64 verify windows. Once the channel fills, the reader blocks in `send` on an earlier stale window and cannot parse the later `DiscardStaleWindows`; most of the stale tail must execute before the registry is updated. Please decouple control-message lookahead from the bounded execution queue (or otherwise cover the full admitted verify backlog), and test a backlog larger than `max_inflight`.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
```diff
@@ -2,33 +2,84 @@ use anyhow::{Context, Result};
...
+    thread::spawn(move || {
```
i386: [P2] Give this reader a shutdown/join path. The cloned `TcpStream` and `JoinHandle` are moved into an untracked thread. Dropping `InboundMessageReader` only drops the receiver; if the connection handler exits on a local execution/protocol error while the peer keeps its side open, this thread remains blocked in `read_stage_message`, retains the cloned socket fd, and never reaches the failed `sender.send`. Repeated failures can therefore leak one thread and connection per request. Store a stop-capable stream/handle and shut it down on drop (or use a bounded read timeout plus stop token), with a test that holds the peer open while the handler exits.

i386 (changes requested):
The focused `skippy-server` suite passes on this head (491 passed, 3 ignored), and the checked PR lanes are green. I found four integration issues around mixed-version wire negotiation, disabling inherited run-ahead config, bounded discard lookahead, and reader-thread teardown; these paths are not covered by the current unit tests.

crates/skippy-protocol/src/binary/types.rs
danielwinterw: Fixed in f632e71 by bumping STAGE_PROTOCOL_GENERATION to 5 (feature token stage-generation-5), so split planning excludes peers that cannot parse kind 23. I took the generation bump rather than a dedicated capability: the generation token is the existing mechanism for current-generation frames, and a per-peer capability would need plumbing from mesh split planning into the embedded frontend's send path. Trade-off is that a run-ahead coordinator will not split with older peers at all instead of degrading with discard off. If mixed-version meshes need to keep working, I can do the capability route as a follow-up.

crates/mesh-llm-config/src/model_validation.rs
danielwinterw: Fixed in f632e71: validation accepts 0..=MAX_VERIFY_WINDOW_RUNAHEAD_TOKENS. Added model_level_zero_runahead_overrides_a_positive_global_default in the resolver tests: global 256 resolves to 256 when inherited, and a model-level 0 resolves to fixed-depth mode.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
danielwinterw: Fixed in f632e71: the reader channel is now sized to cover the whole admitted backlog (INBOUND_LOOKAHEAD_MESSAGES = 2 x MAX_VERIFY_WINDOW_PIPELINE_DEPTH, windows plus their retire messages) rather than max_inflight, so the reader always reaches the discard. Regression test writes MAX_VERIFY_WINDOW_PIPELINE_DEPTH messages followed by a DiscardStaleWindows past a capacity-1 execution queue, and asserts the registry records the discard with nothing dequeued; it times out on the previous bounded queue.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
danielwinterw: Fixed in f632e71: InboundMessageReader keeps a second stream clone and the JoinHandle; Drop shuts the socket down (Shutdown::Both) to unblock read_stage_message and joins the thread. The reader lives for the whole connection (one handle_binary_connection_messages call per accepted connection), so the shutdown only fires when the handler is exiting. Test holds the peer open, drops the reader on another thread, and asserts the drop completes.

i386:
Reviewed — solid, careful PR. The core design is right and the opt-in gating is done properly. A handful of things to settle before I'd approve; one is a conscious behavioral sign-off, the rest are minor.

**What I checked**
- `skippy-protocol` tests pass locally (45/45): generation bump 4→5, `WireMessageKind::DiscardStaleWindows = 23`, `MAX_VERIFY_WINDOW_RUNAHEAD_TOKENS`.
- `skippy-server` won't build in my env (FFI/llama.cpp libc++ `<array>` header failure — toolchain, not the PR), so the scheduler / reader-thread / stale_discard logic is a **static review only**. Those unit tests need to be green in CI before merge.

**What's good**
- The race is solved correctly: the reader thread records the discard range in the registry *before* the channel send, so even a full prefetch channel doesn't delay recording. Lookahead sized 2×depth with a clear rationale and a capacity-1 regression test.
- Reader `Drop` shuts the socket down and joins the thread — no leaked thread/fd, with a dedicated test.
- Malformed discards ignored ("optimization, never a correctness dependency") — good invariant, tested.
- Forward progress preserved: `.max(1)` guarantees at least one window admits even when a single window exceeds the budget.
- Generation bump correctly gates kind 23 so pre-gen-5 peers are excluded from split planning.

**Worth addressing**

1. *(medium — needs a decision)* The inbound reader thread + prefetch channel is spawned **unconditionally for every binary connection**, including fixed-depth setups. The wire is byte-for-byte unchanged, but the *receive* path is not: in fixed-depth mode no discards are ever sent, so `is_discarded` is always false and the reader adds zero correctness value — just a background thread and a prefetch channel of `max(max_inflight, 128)` frames (each up to `MAX_STAGE_FRAME_BYTES` = 8 MB) that can buffer in userspace. Steady-state it's bounded by admission so the delta is modest, but I'd either **gate the reader on run-ahead mode** (cleanest — keeps fixed-depth genuinely unchanged) or explicitly sign off that unconditional prefetch is intended and the worst-case memory ceiling is fine.

2. *(question)* Lookahead of 2×`MAX_VERIFY_WINDOW_PIPELINE_DEPTH` = 128 assumes ≤~128 unread frames sit ahead of a discard. That holds per request. Does a single connection ever multiplex enough concurrent requests/sessions that their summed backlogs push a discard past 128 unread frames? If so, the reader blocks on a full channel and the discard stays unrecorded until the executor drains — silently degrading back to executing the stale tail. I think per-connection admission makes this safe, but confirm.

3. *(question)* Skipped windows reply with an empty `PredictedTokens`. You note the driver's stale drain "only uses the window id for FIFO bookkeeping" — worth confirming the driver's accept-rate telemetry / token counters treat a zero-token reply for a stale window identically and don't skew the accept metric or trip an assert.

**Nits**
4. Docs are now inconsistent: the token is `stage-generation-5` but surrounding prose in `README.md`, `docs/skippy/DATA_FLOW.md`, and `docs/design/TESTING.md` still says "generation 4" / "generation-4 topology" / "generation 4 is a compatibility-breaking change." Should read generation 5.
5. `config_schema_defaults_ui_reference.json` lost its trailing newline (diff shows "No newline at end of file"). Regen if the generator emits one.
6. `run_binary_stage_message` gained `DiscardStaleWindows` in its match arm, but discards are `continue`d in the connection loop before execution — confirm that arm is reachable (defensive) rather than dead code.
7. `WIRE_SAMPLE_COUNTER` is process-global, so concurrent conditioned writers share one interleaved sample sequence rather than independent per-stream RNG. Fine for a bench model (you call it non-cryptographic) — just noting the jitter isn't independent per link under concurrency.

Nothing here is a correctness blocker on the run-ahead path itself. #1 is the one I'd want a conscious answer on; the rest are quick.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
```diff
@@ -1,34 +1,195 @@
...
+}
+
+impl Drop for InboundMessageReader {
+    fn drop(&mut self) {
+        // Unblock a pending `read_stage_message`; errors here only mean the
+        // socket is already closed.
+        let _ = self.stream.shutdown(Shutdown::Both);
+        if let Some(thread) = self.thread.take() {
+            let _ = thread.join();
+        }
```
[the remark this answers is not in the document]
danielwinterw: Fixed in b122b12: Drop now takes and drops the receiver before the socket shutdown and join, since a reader blocked in send on a full lookahead queue is not woken by the shutdown alone. Added dropping_the_reader_completes_while_the_lookahead_channel_is_full, which fills INBOUND_LOOKAHEAD_MESSAGES plus slack before dropping.

crates/skippy-server/src/binary_transport/wire.rs
```diff
@@ -1,27 +1,92 @@
...
+            bail!("downstream wire jitter must be finite and non-negative");
+        }
+        if !stall_ms.is_finite() || stall_ms < 0.0 {
+            bail!("downstream wire stall must be finite and non-negative");
+        }
+        if !stall_p.is_finite() || !(0.0..=1.0).contains(&stall_p) {
+            bail!("downstream wire stall probability must be within [0, 1]");
+        }
+        if stall_p > 0.0 && stall_ms == 0.0 {
+            bail!("downstream wire stall probability requires a stall duration");
```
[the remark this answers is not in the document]
danielwinterw: Fixed in b122b12: with_jitter rejects delay/jitter/stall inputs above one simulated hour (MAX_SIMULATED_DELAY_MS), and propagation_delay clamps the combined sample to the same bound before Duration::from_secs_f64, which also covers the unbounded exponential jitter tail.

crates/skippy-server/README.md
```diff
@@ -104,7 +104,7 @@ deadline handling.
 
 - `serve-binary` is the tuned binary stage-to-stage path.
 - `serve-binary` participates in the breaking generation-4 stage protocol.
-  Stage compatibility requires `stage-generation-4`; direct prediction return and
+  Stage compatibility requires `stage-generation-5`; direct prediction return and
```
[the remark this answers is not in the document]
danielwinterw: Fixed in b122b12: the remaining generation-4 references in the skippy-server README, TESTING.md, and DATA_FLOW.md now name generation 5.

danielwinterw:
Re the out-of-diff scheduler finding (has_capacity admitting past the token budget): fixed in b122b12. The scheduler enforces the budget from the second in-flight window on (open rejects, and admissible_window_tokens reports the remaining width), and the admission loop clamps its chunk width to the remaining budget rather than planning a chunk the budget cannot fit — so in-flight tokens never exceed the budget once a window is in flight. A first window wider than the whole budget still opens from idle, so a budget narrower than one window cannot stall a request; the smoke test now caps at exactly 100/100 instead of recording 144.

crates/skippy-server/src/frontend/embedded_execution.rs
```diff
@@ -105,6 +113,45 @@ impl StageOpenAiBackend {
...
+    ) -> OpenAiResult<()> {
+        let message = discard_stale_windows_message(
+            request.wire_dtype,
+            discard.request_id,
+            discard.session_id,
+            discard.min_window_id,
+            discard.max_window_id,
+        )?;
+        if let Some(forwarder) = async_forwarder {
+            forwarder
```
ndizazzo, line 135: The receipt from this async send is discarded. The forwarder can still be sleeping or writing `DiscardStaleWindows` when teardown sends `Stop` through another clone of the same socket. Since a frame takes multiple writes, those messages can overtake or interleave and poison the persistent lane. Please flush or await the discard writer before Stop and before returning the lane.

ndizazzo (changes requested):
Needs revision. The discard control isn't serialized with teardown, so a reused lane can be corrupted.

Follow-ups:
- Bound inbound lookahead by bytes as well as message count. A queue of 128 valid activation messages can retain tens of GiB.
- Avoid the process-global jitter counter in parallel tests and lane scheduling; it makes assignment scheduler-dependent.
- Add a delayed/jittered teardown test that covers discard, Stop, and lane reuse.

crates/skippy-server/src/frontend/embedded_execution.rs
danielwinterw: Fixed in 2d7e284: AsyncForwarder now joins its writer thread on drop, so no queued frame can still be on the wire when the request returns its lane and a teardown Stop goes out through another clone of the socket. The teardown discard also flushes explicitly (before the stale drain) so write errors surface at that point rather than being swallowed at drop. The mid-generation discard still does not wait, since everything behind it is queued on the same forwarder and stays ordered — that was the reason for the original fire-and-forget, and it does not apply at teardown.

Regression test: a_delayed_discard_lands_before_a_teardown_stop_on_another_clone queues a discard behind 250ms of simulated propagation, drops the forwarder, then writes Stop directly to the socket and asserts the receive order. Verified it fails without the join (reads Stop first) and passes with it.

Your three follow-ups: the byte bound and the global jitter counter are also in 2d7e284 — the inbound queue is now capped by INBOUND_LOOKAHEAD_BYTES (256 MiB) as well as message count, since 128 wide activation frames would otherwise retain many GiB, and wire conditioning draws from a per-thread counter so parallel tests and per-lane conditioning no longer depend on scheduler interleaving. The delayed/jittered teardown test above covers discard + Stop ordering on a reused socket; a full lane-reuse test needs the two-machine harness rather than a unit test, so if you want that specific coverage I would rather add it as a follow-up with the lab bench than fake it in-process.

crates/skippy-server/src/binary_transport/wire.rs
```diff
@@ -1,27 +1,108 @@
...
+            (stall_ms, "stall"),
+        ] {
+            if value > MAX_SIMULATED_DELAY_MS {
+                bail!("downstream wire {name} must not exceed {MAX_SIMULATED_DELAY_MS} ms");
+            }
+        }
         if !delay_ms.is_finite() || delay_ms < 0.0 {
             bail!("downstream wire delay must be finite and non-negative");
         }
         if mbps.is_some_and(|value| !value.is_finite() || value <= 0.0) {
```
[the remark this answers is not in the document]
danielwinterw: Fixed in 2d7e284: the serialization delay is now computed by bandwidth_delay, which clamps to MAX_SIMULATED_DELAY_MS and rejects non-finite/non-positive values before Duration::from_secs_f64 — same bound as the propagation delay. Test covers mbps = f64::MIN_POSITIVE (previously an overflow panic), a zero-byte frame, and the unmetered case.

danielwinterw:
On the out-of-diff finding about panics in connection workers stopping the accept loop (reap_finished turning a join Err into a bail): agreed that it is a real availability bug, but it is pre-existing on main from the iteration-scheduler work (#1420) rather than something this PR introduces or touches, so I have left it alone here to keep the diff scoped. Happy to raise it separately — say the word if you would rather it rode along with this PR.

i386:
Re-reviewed at `a8eeaa70`, after the four `review:` commits. The integration issues from my 2026-08-23 pass are addressed: run-ahead admission is now hard-bounded past the first window, the reader is joined and unblocked on drop, the discard writer is serialized with teardown, and the lookahead is decoupled from `max_inflight`.

**Build note (updates my 2026-08-24 comment):** `skippy-server` builds in this environment now, so the scheduler / reader-thread / `stale_discard` logic is no longer static-review-only.

- `cargo test -p skippy-server` — 536 passed, 0 failed, 3 ignored
- `cargo test -p mesh-llm-host-runtime` — 2663 passed, 0 failed, 8 ignored
- `cargo test -p skippy-protocol -p mesh-llm-config` — all green
- The new socket/timing tests (`wire`, `message_receive`, `stale_discard`, 17 total) ran 5x clean, no flakes

All at `git rev-parse HEAD` = `a8eeaa70c7196e49a75a0ed554888ed8c2044991`.

Four things left. None of them block the design; (1) and (2) I'd want settled before merge.

---

### 1. The inbound read-ahead ceiling is now a per-connection DoS bound, and it got much looser

`spawn_message_reader` runs on **every** binary connection, including fixed-depth setups that will never see a `DiscardStaleWindows`. That's fine on its own — but it moves the flow-control boundary. Before this PR, `receive_next_message` read on the executor thread, so an unread frame stayed in the kernel socket buffer. Now the reader parses eagerly into userspace, bounded by `INBOUND_LOOKAHEAD_MESSAGES` (128) **or** `INBOUND_LOOKAHEAD_BYTES` (256 MiB), whichever binds first.

For a well-behaved driver this is harmless: a depth-2 driver never has more than 2 windows outstanding, so the queue never fills. The bound matters for the case it exists to cover — a peer that sends more than it should. That ceiling went from `O(max_inflight x frame)` to a flat 256 MiB per connection, on a listener that accepts connections concurrently.

I don't think you need the full 256 MiB for the property you're buying. A `DiscardStaleWindows` frame is ~100 bytes; it overtakes a 32 MiB activation backlog exactly as reliably as it overtakes a 256 MiB one. Dropping `INBOUND_LOOKAHEAD_BYTES` to 32 MiB gets the same discard behaviour with an 8x smaller worst case.

### 2. Every simulated lane draws the identical jitter sequence

`next_uniform_sample` is a pure function of `WIRE_SAMPLE_INDEX`, which is thread-local and starts at `0` in every thread. So two threads don't get independent streams — they get the *same* stream. `each_thread_draws_its_own_deterministic_sequence` asserts exactly that (`assert_eq!(first, second)`).

`propagation_delay()` is called from `run_forwarder` (`async_forwarder.rs:160`), which is one writer thread per downstream lane. So in any topology with more than one lane, every lane takes its 40 ms burst stall on the *same message index*. That's a synchronized-loss model, not the contended-link model the flag is documenting — and synchronized head-of-line stalls are much easier for a pipelined scheduler to ride out than independent ones, so the jitter column in your table is likely optimistic for >2-stage splits. (For the 2-process loopback bench in the PR body, single lane, no effect — those numbers stand.)

Moving to per-thread streams was the right call for @ndizazzo's reproducibility point; it just needs a per-thread salt to also be *independent*. Mixing a monotonically-assigned thread ordinal into the splitmix input keeps each stream reproducible and decorrelates them. The existing test then asserts the opposite of what it does today.

### 3. The lookahead invariant in the doc comment isn't the one the code holds

The comment on `INBOUND_LOOKAHEAD_MESSAGES` says:

> so this covers the whole admitted backlog: the reader never blocks on a stale window while a `DiscardStaleWindows` for it is still unread in the socket.

That held before the byte gate landed. It doesn't now: 64 windows at up to `MAX_STAGE_FRAME_BYTES` (8 MiB) is 512 MiB, past the 256 MiB ceiling, so the reader can park with the discard still unread. The failure mode is benign — it degrades to today's execute-the-stale-tail cost, and the executor keeps draining so there's no deadlock — but the comment states an invariant as unconditional when it's conditional on frame width. Worth saying "usually overtakes; falls back to executing the tail when it doesn't" rather than "never."

### 4. Docs and small stuff

- `docs/skippy/DATA_FLOW.md`: the section heading is still `## Generation 4 Direct Prediction Return and Verify Retirement` while its body now says generation 5.
- More substantively, `DiscardStaleWindows` (kind 23) is the thing that *motivated* the generation bump, and it isn't described in `DATA_FLOW.md` at all. A compatibility-breaking generation should document the frame that broke it — including the "middle stages forward and still execute, final stage skips" rule, which is the non-obvious part and currently only lives in the PR body.
- `InboundMessageReader::next`'s doc says "Mirrors `receive_next_message`'s EOF classification" — that function is deleted in this PR, so the reference dangles.
- `WireCondition::with_jitter` runs the `MAX_SIMULATED_DELAY_MS` loop before the `is_finite` checks, so `delay_ms = f64::INFINITY` reports "must not exceed 3600000 ms" instead of "must be finite". Cosmetic.

---

### One question, not a finding

The generation-5 gate lives entirely in mesh split planning (`supports_skippy_stage_generation` in `convert.rs`). The standalone `serve-binary --downstream host:port` path has no generation handshake — I grepped `crates/skippy-server/src` for `STAGE_PROTOCOL_GENERATION` and the generation feature tokens and found no references. So a gen-4 stage binary that receives kind 23 fails `TryFrom<i32>` ("unknown stage message kind") and drops the connection.

Since run-ahead is opt-in and the discard only ships in run-ahead mode, this only bites someone who turns run-ahead on across a mixed-version standalone pair — which is exactly the manually-wired path your own bench uses. Is "upgrade all stages together" the contract for standalone, or do you want the sender to degrade? Either answer is fine; I'd just like a sentence in `crates/skippy-server/README.md` saying which, because today the negotiation story reads as complete and it's only complete for the mesh path.

danielwinterw:
Thanks — all four addressed in b60599e7.

**1. Read-ahead ceiling.** Agreed, and the framing is the useful part: this is a per-connection bound on what a peer can make the process buffer, not just a queue size. Dropped to 32 MiB. You're right that the discard overtakes a 32 MiB backlog exactly as reliably.

**2. Identical jitter streams.** This was a real bug in my fix and the test asserted the wrong property — thank you. Per-thread index alone gives every lane the same sequence, which is synchronized loss, not a contended link. Now salted with a per-thread stream ordinal claimed on first draw. `uniform_sample` is pure in `(stream, index)` so both properties are tested directly: reproducible within a stream, pairwise-distinct across streams, plus the thread-level test now asserting `assert_ne!`.

On the table: the loopback bench is single-lane so those numbers stand, but I've taken your point that the jitter column is likely optimistic for >2-stage splits and I'm not going to quote it for multi-stage until the lab bench runs with independent streams.

**3. Lookahead invariant.** Correct, and my comment overstated it. It now says the discard usually overtakes, names the byte gate as the case where it doesn't, and states the fallback (execute the stale tail — today's cost, no deadlock, because the executor keeps draining).

**4. Docs.** All fixed: the DATA_FLOW heading said "Generation 4" because my earlier pass only replaced the lowercase form; the dangling `receive_next_message` reference is gone; `with_jitter` checks finiteness before the magnitude bound so infinity reports what's actually wrong. `DiscardStaleWindows` now has its own section in DATA_FLOW.md including the middle-stages-forward-and-execute vs final-stage-skips rule, which as you say was only living in the PR body.

**On the standalone question:** the contract is upgrade-all-stages-together, and I've said so in crates/skippy-server/README.md — including explicitly that it is not degraded gracefully (the older peer rejects kind 23 and drops the request connection). You're right that the negotiation story read as complete when it was only complete for the mesh path.

danielwinterw:
Re-requested review. Everything from both passes is in and CI is green at `b60599e7` (37/37).

Also filed the two things that came out of review but belong to [#1420](https://github.com/Mesh-LLM/mesh-llm/pull/1420) rather than this PR, so they don't get lost: [#1472](https://github.com/Mesh-LLM/mesh-llm/issues/1472) (the flaky `worker_panic_is_contained_and_fails_active_requests` race, which fail-fast-cancelled three test batches on one of this PR's runs) and [#1473](https://github.com/Mesh-LLM/mesh-llm/issues/1473) (a panicking connection worker stopping the accept loop — CodeRabbit's out-of-diff finding).

i386 (commented):
I rechecked the latest head `b60599e7` and cannot approve it yet. Two current blockers already identified in the review threads still apply: `DiscardStaleWindows` adds wire kind 23 without a negotiated capability/version boundary, so a mixed-version stage can reject the frame as an unknown message; and `WireCondition::bandwidth_delay` returns `Duration::ZERO` when a positive rate produces an infinite quotient, turning an extremely slow configured link into a zero-delay link instead of applying the documented cap. The branch is also currently conflicted with `main`. I have not duplicated the existing inline comments.

crates/skippy-protocol/src/binary/types.rs
danielwinterw: Re-answered after the rebase, because main moved underneath this thread.

Main is now at stage protocol generation 6, and generation 6 does not include wire kind 23. So the generation-5 bump that answered this originally no longer separates peers that can parse the discard from peers that cannot: a current-main peer advertises `stage-generation-6`, passes split planning, and still rejects kind 23 as an unknown message kind — exactly the teardown you flagged.

Bumped to generation 7 in 842c5811d. `STAGE_PROTOCOL_GENERATION = 7`, feature token `stage-generation-7`, and split planning gates on it, so a peer that predates the frame is excluded from the topology rather than being handed a frame it cannot parse. The docs now name kind 23 as the reason this generation breaks compatibility, rather than leaving the reader to infer it: DATA_FLOW.md and crates/skippy-server/README.md both say so.

The standalone `serve-binary --downstream` path still has no handshake, so the contract there is unchanged and still stated in the README: upgrade all stages together, and it is not degraded gracefully.

crates/skippy-server/src/frontend/embedded_execution.rs
danielwinterw: Added the lane-reuse coverage in 842c5811d: `a_reused_lane_carries_the_next_request_after_a_delayed_teardown`.

It queues a discard behind 250ms of simulated propagation, drops the forwarder, writes the teardown `Stop` through another clone of the socket, then writes the next request's verify window on that same socket the way the pool hands a lane back out. It asserts all three frames arrive whole and in order, so a frame left half-written by the torn-down forwarder cannot be read as the next request's header.

That covers discard, Stop, and reuse in process. What still needs the two-machine harness is a real pooled lane carrying a real second generation, so I have kept that for the lab bench rather than faking a runtime here.

crates/skippy-server/src/binary_transport/wire.rs
```diff
@@ -1,45 +1,155 @@
...
+    /// out-of-domain value.
+    pub(crate) fn bandwidth_delay(&self, bytes: usize) -> Duration {
+        let Some(mbps) = self.mbps else {
+            return Duration::ZERO;
+        };
+        let seconds = bytes as f64 / (mbps * 125_000.0);
+        if !seconds.is_finite() || seconds <= 0.0 {
+            return Duration::ZERO;
+        }
+        Duration::from_secs_f64((seconds * 1000.0).min(MAX_SIMULATED_DELAY_MS) / 1000.0)
```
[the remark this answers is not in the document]
danielwinterw: Fixed in 842c5811d. Only NaN and non-positive quotients return `Duration::ZERO` now. An infinite quotient is an arbitrarily slow link, not a free one, so it takes the `MAX_SIMULATED_DELAY_MS` cap like any other oversized delay; `f64::min` propagates the bound for an infinite left-hand side, so no separate branch is needed.

Regression test `an_infinite_bandwidth_quotient_takes_the_cap_not_zero` uses `f64::from_bits(1)`, asserts the quotient really is infinite so the test keeps biting, and checks the delay is the cap. Verified it fails on the previous `!seconds.is_finite()` branch (reads 0ns against 3600s) and passes now.

danielwinterw:
Rebased onto main and re-answered both blockers. Head is `842c5811d`; the branch is `MERGEABLE` again.

The rebase was 90 commits and moved four things this branch had to answer for, so it is worth reading as more than a conflict resolution.

**1. The generation gate had to move (i386's blocker on kind 23).** Main is now at generation 6, and generation 6 does not carry wire kind 23. The generation-5 bump that originally answered that thread therefore stopped separating peers that can parse the discard from peers that cannot: a current-main peer advertises `stage-generation-6`, clears split planning, and still rejects kind 23 as an unknown message kind. Bumped to generation 7, with split planning gating on `stage-generation-7`, and the docs now name kind 23 as the reason the generation breaks rather than leaving it to be inferred.

**2. Infinite bandwidth quotient (i386's second blocker).** `bandwidth_delay` returned `Duration::ZERO` when a positive but tiny rate overflowed the quotient to infinity, which served the slowest configurable link as an unmetered one. Now only NaN and non-positive quotients yield no delay; infinity takes the `MAX_SIMULATED_DELAY_MS` cap. The regression test asserts the quotient really is infinite before checking the cap, and I verified it fails against the old branch (0ns against 3600s).

**3. A main fix this branch was quietly dropping.** Main routes inbound reads through `ConnectionWorkerControl::wait_for_readable` so a worker never sits in a read that a socket shutdown cannot interrupt on Windows (#1538). Moving the read onto the inbound reader thread lost that: after the mechanical rebase, `worker_control` was an unused parameter. The reader now waits through the same control and ends when shutdown is requested, and the byte-ceiling backoff observes shutdown too. Worth a look, since it is the one place where the rebase changed behavior rather than just resolving text.

**4. New config gates on main.** `verify_window_runahead_tokens` now needs a website config-reference row, a wiring-manifest entry, and a reverse-audit entry. All three added.

Also added the lane-reuse teardown coverage @ndizazzo asked for: `a_reused_lane_carries_the_next_request_after_a_delayed_teardown` writes the next request's frame on the same socket after the delayed discard and the teardown `Stop`, and asserts all three arrive whole and in order. A real pooled lane carrying a real second generation still wants the two-machine harness.

Local verification on `842c5811d`, native runtime built with `LLAMA_STAGE_BACKEND=metal`:

- `cargo test -p skippy-server` — 672 passed, 0 failed, 3 ignored
- `cargo test -p mesh-llm-host-runtime` — 2898 passed, 0 failed, 8 ignored
- `cargo test -p skippy-protocol` — 43 passed; `cargo test -p mesh-llm-config` — all green
- `cargo fmt --all -- --check`, all five `xtask repo-consistency` checks, and clippy on the touched crates — clean

**Unrelated breakage found while building, not fixed here:** the llama.cpp patch queue on `main` has three patches numbered `0035`, and `scripts/prepare-llama.sh` applies them in sorted order. `0035-skippy-account-retained-tensors-during-staged-loads.patch` (#1542) sorts first but expects context that a later patch adds, so a fresh `prepare-llama.sh` fails on it with `sha1 information is lacking or useless`. Only incremental trees that already had it applied are unaffected, which is presumably why CI is green. I built locally by renumbering it to sort last. Happy to raise it separately.

danielwinterw:
Correction to my previous comment: the llama.cpp patch-queue breakage is fixed here after all, in `cfb8739e2`, because it is what was failing this PR's Static ABI lane and nothing else could go green past it.

`0035-skippy-account-retained-tensors-during-staged-loads.patch` is renumbered to `0046` so it applies after `0038`, which adds the `skippy-hardware-application-probe` target its `src/CMakeLists.txt` hunk expects. The patch bytes are unchanged; only the apply order moves. Verified by deleting `.deps/llama.cpp` and replaying the whole queue from a fresh clone, then building the Metal staged ABI libraries from the result.

Worth knowing for anyone else who hits it: incremental trees that already had the patch applied are unaffected, so it only bites a clean checkout. The three earlier Clippy batch failures on this PR were fail-fast cancellations from that ABI lane, not Clippy findings.

CI is green at `cfb8739e2`: 40 passed, 0 failed, 12 skipped.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
```diff
@@ -1,45 +1,326 @@
...
+            // worker is shutting down.
+            match worker_control.wait_for_readable(&reader) {
+                Ok(true) => {}
+                Ok(false) => return,
+                Err(error) => {
+                    let _ = sender.send(Err(error));
+                    return;
+                }
+            }
+            match read_stage_message(&mut reader, activation_width) {
```
i386: [P1] Keep the full framed read interruptible, not only the first-byte peek. wait_for_readable returns after one byte and clears the socket timeout before read_stage_message, so a peer can send a partial header/frame and then stall this thread inside read_exact. InboundMessageReader::drop then shuts down only self.stream, a cloned handle, and waits in join. On Windows that cloned-socket shutdown is specifically documented above as unable to interrupt an in-flight read (#1538), so connection-local failure or process shutdown can hang indefinitely. The current peer-stays-open test sends zero bytes and exercises only the peek wait; please add a partial-frame regression on Windows and use framing/cancellation that can interrupt the exact read handle after any prefix.

i386 (changes requested):
The generation gate, zero override, token-budget admission, bounded lookahead, lane ordering, jitter bounds, and the rest of the regression coverage look sound. Local validation at this exact head passed the full skippy-server library suite (672 passed, 3 ignored) and full mesh-llm-host-runtime library suite (2898 passed, 8 ignored); attached CI is green. One shutdown blocker remains in the new reader path.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
danielwinterw: Correct, and the distinction matters more than I gave it credit for: the readable wait only covers an idle socket. Once a prefix arrives, `wait_for_readable` returns and clears the timeout, and the reader is committed to `read_exact` for the rest of the frame. Since `Drop` joins, a peer that stalls there hangs the dropping handler rather than merely leaking a thread, which is worse than the leak the join was added to prevent.

Fixed in `981d9cc32` by removing the second clone. The reader thread and the handle now share one `TcpStream` through an `Arc`, and `Drop` shuts down that exact handle rather than a sibling clone. `&TcpStream` implements `Read`, so the framed read runs on the shared handle with no other change to the read path.

Regression test `dropping_the_reader_completes_while_a_read_is_stalled_mid_frame` writes a 4-byte frame prefix and then stalls, so the reader is past the readable check and blocked inside the frame, and asserts the drop completes within 5s. The existing peer-stays-open test sends nothing and only exercises the wait, as you said.

One honest limit on the "on Windows" part of the ask: the Windows CI lane builds the host but does not run the `skippy-server` suite, so this test guards the behavior on Windows rather than proving it there. It is portable and unconditional, so it runs for anyone developing on Windows and in any future lane that runs the suite. If you want it actually exercised in CI, that needs a Windows lane that runs `cargo test -p skippy-server`, which I would rather raise separately than widen this PR into.

I did not add a read timeout around the framed read as a second line of defence, deliberately: a timeout that fires mid-frame desynchronizes the stream, and it would fire on a legitimately slow large frame rather than only on a stalled peer.

danielwinterw:
Rebased again onto `26f9fc567` and dropped my patch-numbering commit in favour of [#1587](https://github.com/Mesh-LLM/mesh-llm/pull/1587), which landed while I was pushing and fixes the same thing properly. Head is `d0b1aa2f2`.

Worth saying explicitly: #1587 is strictly better than what I had. I renumbered the one offending patch to `0046`, which happened to collide with the tail after #1587 renumbered the whole sequence contiguously to `0047` and added `validate_patch_sequence` to fail fast on duplicates and gaps. My version would now fail that validator. It is gone from this branch and no trace of it remains in the diff.

Also corrected a rebase artifact of my own: resolving the console-print ratchet conflict the first time took my side wholesale and silently dropped main's entries for `crates/skippy-bench/src/direct_return_listener.rs` from #1560. Regenerated, and the ratchet now differs from main in exactly one file, `binary_messaging.rs`, which is the only file this PR shifts lines in.

The mid-frame shutdown fix from `981d9cc32` carried across the rebase unchanged.

Re-verified at `d0b1aa2f2` with the native runtime rebuilt against the new patch queue:

- `cargo test -p skippy-server` — 665 passed, 0 failed, 3 ignored
- `cargo test -p mesh-llm-host-runtime` — 2886 passed, 0 failed, 8 ignored
- `cargo test -p skippy-protocol -p mesh-llm-config` — green
- `cargo fmt --all -- --check` and all five `xtask repo-consistency` checks — clean

The suite counts moved against my previous comment because main did (#1588 derives KV policy from the loaded model), not because anything here changed.

danielwinterw:
Rebased onto `12c83e83b` and resolved the conflicts. Head is `b6b1aecb7`, `MERGEABLE` again.

**The generation boundary moved a third time.** Main is now at generation 7 itself, via [#1585](https://github.com/Mesh-LLM/mesh-llm/pull/1585), and generation 7 does not carry wire kind 23. That is the same situation as last round, one generation later: a generation-7 peer advertises current support, clears split planning, and still rejects the discard frame as an unknown message kind. So this branch now bumps to generation 8, keeping main's wording for the rest of the generation contract (the control, status-list, and strict-content-identity bundle) and adding kind 23 to it.

Worth flagging as a pattern rather than a one-off: this is the third rebase where the only substantive change was moving the generation number, because the branch and main keep claiming the same one. @i386, if you would rather kind 23 rode on a dedicated capability token instead of the generation counter, that is the route you offered at the start of the original thread and it would stop this recurring. I have not switched to it unilaterally, because it needs plumbing from mesh split planning into the embedded frontend's send path and that is a bigger change than a rebase should carry. Say the word either way.

Other conflict resolutions, none of them behavioral:

- **#1585 split the connection's activation width** into input and output halves. The inbound reader parses upstream frames, so it takes the input width.
- **README, DATA_FLOW.md and TESTING.md** kept main's generation wording, with the kind-23 sentence and the standalone-path contract paragraph merged back in rather than overwriting either side.
- **Console-print ratchet** regenerated; it differs from main in exactly one file, `binary_messaging.rs`, which is the only file this PR shifts lines in.

Re-verified at `b6b1aecb7` with the native runtime rebuilt against main's current patch queue:

- `cargo test -p skippy-server` — 674 passed, 0 failed, 3 ignored
- `cargo test -p mesh-llm-host-runtime` — 2899 passed, 0 failed, 8 ignored
- `cargo test -p skippy-protocol -p mesh-llm-config` — green
- `python3 scripts/tests/test_prepare_llama.py` — 3 passed, so the branch is clean against #1587's new patch-sequence validator
- `cargo fmt --all -- --check` and all five `xtask repo-consistency` checks — clean

danielwinterw:
CI caught one thing the generation-8 bump broke, fixed in `bbc69cd54`.

`stage_control_request_validates_generation_sender_and_command` builds a frame at `STAGE_PROTOCOL_GENERATION - 1` but asserted the rejected value against the literal `6`, so it failed the moment the constant moved. The other three red checks were downstream of it: batch-3 was fail-fast cancelled, and the two `Enforce Linux result` gates aggregate the batch.

It now asserts against the computed previous generation instead of a literal, so it keeps testing what it means to test (the previous generation is rejected) rather than failing on the number at every bump. That is the same recurring cost I flagged in my last comment, in test form.

My own miss, worth owning: I ran that suite locally before pushing and my output filter cut off above the failing line, so I reported it green when it was not. The full suites at `bbc69cd54`:

- `cargo test -p skippy-protocol` — 43 passed, 0 failed
- `cargo test -p skippy-server` — 674 passed, 0 failed, 3 ignored
- `cargo test -p mesh-llm-host-runtime` — 2899 passed, 0 failed, 8 ignored
- `cargo test -p mesh-llm-config` — 188 + 4 + 5 + 10 passed, 0 failed
- `cargo fmt --all -- --check` — clean

crates/skippy-protocol/src/binary/types.rs
i386: Addressed across 8093b63ec and 8e4e6d060: the discard kind is negotiated through the bumped stage generation (now generation 8), and the README documents the compatibility boundary.

crates/mesh-llm-config/src/model_validation.rs
i386: Addressed in 8093b63ec: model-level runahead_max_tokens accepts the zero sentinel and the precedence regression covers a positive default overridden by model-level zero.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
i386: Addressed across 8093b63ec, 6c3e6a2bf, and e592f2341: control-message lookahead is decoupled from execution capacity and bounded by bytes, so discard can be consumed ahead of a stale data tail without unbounded buffering.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
i386: Addressed across 8093b63ec, 6c3e6a2bf, 197e6d717, and 81c3536e2: the reader now has stop control, socket interruption, and a joined shutdown path, including full-frame stalls.

crates/skippy-server/src/frontend/embedded_execution.rs
i386: Addressed in e592f2341: discard forwarding is serialized with teardown, and the writer is flushed/joined before Stop can be sent or the lane returned.

crates/skippy-server/src/binary_transport/binary_messaging/message_receive.rs
i386: Addressed in 81c3536e2: framed reads remain interruptible after the first byte, and the regression covers a peer that stalls mid-frame so drop can join the reader on Windows as well.

i386 (approved):

i386 (dismissed):

i386 (approved):
Reviewed after syncing current main. The exact pushed head passes the repository consistency gate; the run-ahead budget, discard ordering, generation boundary, and Windows reader shutdown fixes are covered by focused regression tests.

i386 (dismissed):
Reviewed exact head a2acf7f145d56f8ec932c47df2fca0b5d2d02c58 after synchronization with current main. The earlier mixed-version negotiation, zero override, bounded discard lookahead, and reader teardown blockers are fixed. The full skippy-server suite and repository validation pass, and there are no unresolved review threads.

i386 (approved):

i386 (approved):

PR #1413, opened 2026-08-23
Require structured final reports for agent turns

michaelneale:
## Summary

Callers can opt a non-streaming chat turn into a versioned `report_required` contract. Intermediate tool calls continue normally, while terminal prose or a native final-report proposal becomes one structured report tool call with caller-locked arguments.

The final argument object is rebuilt from only the report body and locked values, so models cannot widen side effects or alter routing. Requests without the extension remain unchanged. Collaboration requests remain inside the configured guardrail flow, and responses with zero or multiple choices fail closed.

## Architecture

- Adds the reusable typed contract and pure OpenAI-response finalizer to `mesh-llm-guardrails`.
- Applies it to direct OpenAI frontend responses and the `model=mesh` MoA gateway.
- Requires a dedicated structured final-report tool; it deliberately does not interpret generic shell commands.

## API

```json
"mesh_collaboration": {
  "mode": "report_required",
  "version": 1,
  "tool": "final_report",
  "body_argument": "content",
  "locked_arguments": {
    "channel": "channel-id",
    "reply_to": "event-id"
  }
}
```

## Validation

At `ed297f768f27ac7e47d1aac76fcaf509e0bd4a42`:

- `cargo test -p mesh-llm-guardrails --lib` — 18 passed
- `cargo test -p openai-frontend --lib` — 188 passed
- `cargo test -p mesh-mixture-of-agents --lib` — 198 passed
- `cargo clippy -p mesh-llm-guardrails --all-targets -- -D warnings`
- `cargo clippy -p openai-frontend --all-targets -- -D warnings`

## Qwen3-8B feasibility probe

Using local Qwen3-8B Q4_K_M with reasoning disabled and a 256-token cap:

| Mode | Valid report call | Exact locked route | Useful body | Errors |
|---|---:|---:|---:|---:|
| No contract | 5/5 | 0/5 | 5/5 | 0/5 |
| `report_required` | 5/5 | 5/5 | 5/5 | 0/5 |

The baseline invented route values. The contract retained the useful semantic body and deterministically enforced caller-authoritative routing. A forced-prose probe also exposed the boundary: textual tool markup can become a structurally valid but polluted report body, so the API guarantees shape/routing, not prose quality.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in collaboration contract for non-streaming chat completions.
  - Converts generated prose into the required reporting tool call.
  - Supports locked tool arguments and preserves unrelated tool calls.
  - Validates reporting tools and contract settings before processing.

- **Bug Fixes**
  - Rejects invalid contracts, streaming requests, and responses with multiple choices using clear errors.

- **Documentation**
  - Added specification details for collaboration-based reporting workflows.

### DIFF
--- a/crates/mesh-llm-guardrails/src/collaboration.rs
+++ b/crates/mesh-llm-guardrails/src/collaboration.rs
@@ -1,0 +1,324 @@
+use serde::Deserialize;
+use serde_json::{Map, Value};
+
+pub const MESH_COLLABORATION_FIELD: &str = "mesh_collaboration";
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(tag = "mode", rename_all = "snake_case", deny_unknown_fields)]
+pub enum CollaborationContract {
+    ReportRequired(ReportRequired),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ReportRequired {
+    pub version: u8,
+    pub tool: String,
+    pub body_argument: String,
+    pub locked_arguments: Map<String, Value>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum CollaborationContractError {
+    Invalid(String),
+    UnknownTool(String),
+    BodyArgumentLocked,
+    ArgumentsDoNotMatchToolSchema(String),
+}
+
+impl std::fmt::Display for CollaborationContractError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Invalid(reason) => write!(
+                f,
+                "mesh_collaboration must be a valid report_required contract: {reason}"
+            ),
+            Self::UnknownTool(tool) => write!(
+                f,
+                "mesh_collaboration tool `{tool}` is not declared in tools"
+            ),
+            Self::BodyArgumentLocked => {
+                f.write_str("mesh_collaboration body_argument must not also be locked")
+            }
+            Self::ArgumentsDoNotMatchToolSchema(reason) => write!(
+                f,
+                "mesh_collaboration arguments do not match the declared tool schema: {reason}"
+            ),
+        }
+    }
+}
+
+impl std::error::Error for CollaborationContractError {}
+
+impl CollaborationContract {
+    pub fn parse(value: &Value, tools: Option<&Value>) -> Result<Self, CollaborationContractError> {
+        let contract: Self = serde_json::from_value(value.clone())
+            .map_err(|error| CollaborationContractError::Invalid(error.to_string()))?;
+        let report = contract.report_required();
+        if report.version != 1 {
+            return Err(CollaborationContractError::Invalid(
+                "version must be 1".into(),
+            ));
+        }
+        if report.tool.trim().is_empty() || report.body_argument.trim().is_empty() {
+            return Err(CollaborationContractError::Invalid(
+                "tool and body_argument must be non-empty".into(),
+            ));
+        }
+        if report.locked_arguments.contains_key(&report.body_argument) {
+            return Err(CollaborationContractError::BodyArgumentLocked);
+        }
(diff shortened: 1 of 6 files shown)


michaelneale:
This is still in draft

crates/mesh-mixture-of-agents/src/gateway.rs
```diff
@@ -67,7 +67,28 @@ pub async fn handle_turn(config: &GatewayConfig, body: &Value) -> TurnResult {
             )
             .await
         }
+    };
+    if let Some(raw_contract) = body.get(mesh_llm_guardrails::MESH_COLLABORATION_FIELD) {
```
ndizazzo, line 71: The host removes `stream` before `handle_turn`, so this code can't tell that the request was streaming. It finalizes the MoA result instead of rejecting the unsupported contract. Chat SSE can emit the synthesized report call, and Responses SSE can finish with empty output. Please preserve the streaming intent and reject before worker/reducer execution.

crates/mesh-llm-guardrails/src/collaboration.rs
```diff
@@ -0,0 +1,332 @@
...
+    {
+        return Ok(false);
+    }
+    let content = message.get("content").and_then(Value::as_str);
+    let arguments = contract.final_arguments(content, tool_calls.as_ref())?;
+    message.insert("content".into(), Value::Null);
+    message.insert(
+        "tool_calls".into(),
+        serde_json::json!([{
+            "id":"call_mesh_collaboration",
```
ndizazzo, line 259: Every synthesized report reuses this fixed tool-call ID. On a later turn, the session matches the tool result to the first pending call with that ID, which can overwrite the old result and leave the new call unresolved. Please generate a unique ID for each report call and add a multi-turn regression test.

ndizazzo (changes requested):
Needs revision. MoA streaming and repeated report turns both have blocking correctness problems.

Follow-ups:
- Validate locked nested arguments against the full declared schema.
- Treat `tool_calls: []` as a final prose response that still needs the report call.
- Parse invalid contracts before worker/reducer execution so MoA returns the same 400 as the direct backend.
- The branch is currently conflicting with main. Conflict resolution needs fresh Linux coverage.

i386 (approved):
Reviewed the contract design and both integration points. This looks good — approving.

What I verified:

- Fail-closed semantics throughout: undeclared/duplicate tool, non-string body argument, body_argument also locked, schema-rejected locked arguments, and multi-choice or malformed outputs all produce explicit errors rather than passing garbage through.
- Pass-through behavior for intermediate tool calls is preserved in both the gateway (`Ok(false)` no-op) and the guarded backend, so agent loops that investigate before reporting keep working.
- Model-supplied arguments can never override locked routing values (`final_arguments` rebuilds from locked + body only).
- No-contract requests are byte-for-byte unchanged, and streaming is explicitly rejected (documented in the spec).
- Tests cover the four key paths: prose wrapping with locked values, multi-choice rejection, absent contract, and intermediate tool-call pass-through; the spec doc matches the implementation.

One merge-gate note: the red CI runs on the current head are infra cancellations (sccache connection refused → "The operation was canceled"), not test failures — none of the failure logs contain a failing assertion. Please re-run the Linux lane before merging so it lands with a green required check.

crates/mesh-llm-guardrails/src/collaboration.rs
```diff
@@ -0,0 +1,332 @@
...
+        content: Option<&str>,
+        tool_calls: Option<&Value>,
+    ) -> Result<Map<String, Value>, CollaborationOutputError> {
+        let report = self.report_required();
+        let native_arguments = match tool_calls {
+            Some(calls) => arguments_from_single_call(calls, &report.tool)?,
+            None => Map::new(),
+        };
+        let body = native_arguments
+            .get(&report.body_argument)
```
i386, line 124: Non-blocking note from my review: the natural place for a future "retry-without-locked-arguments" behavior to hook in is `final_arguments` — if a caller ever wants to let the model retry a missing report body rather than fail closed, keep the fail-closed default and make any retry explicit at the gateway layer, so the locked-routing guarantee can't be weakened by a refactor here.

PR #1414, opened 2026-08-23
fix(runtime): resolve pinned device selectors and honor CLI --device at serve startup

ndizazzo:
## Problem

`mesh-llm serve --model '<ref>'` on a host with a persisted per-model
`hardware.device` PCI stable-id pin (e.g. `pci:00000000:01:00.0`, written by
the hardware-pinning flow under `gpu.assignment = "pinned"`) fails native
model load with:

```
InvalidArgument: unknown selected backend device: pci:00000000:01:00.0
```

An explicit `--device CUDA0` on the CLI did not fix it — the only working
workaround was bypassing the persisted config entirely
(`--config /dev/null --device CUDA0`).

## Root cause

Two independent gaps:

1. `build_startup_model_specs` always built CLI-explicit `--model`/`--gguf`
   specs with `gpu_id: None` / `config_owned: false`. That caused
   `preflight_config_owned_startup_models` — the one place that resolves a
   stable-id device against the live `GpuFacts` inventory — to skip these
   specs entirely (it only ran for config-owned specs, and only under
   `gpu.assignment = "pinned"`). Meanwhile, the model's *raw* persisted
   `hardware.device` string was still picked up separately by
   `resolve_hardware_config` (which looks up the model by ref regardless of
   how it was launched) and passed unresolved into
   `SkippyDeviceDescriptor.backend_device` → `RuntimeConfig.selected_backend_device`.
2. `RuntimeOptions.device` (CLI `--device`) was never read by any
   device-selection code path, so it had no way to override a stale or
   incompatible persisted pin.

## Fix

Upstream of the Skippy ABI and `runtime_config_from_stage_config` (both left
untouched):

- `build_startup_model_specs` now looks up the matching `config.models`
  entry (exact ref match) for CLI-explicit launches, and applies a single
  CLI-device-wins-else-persisted precedence rule everywhere a startup
  model's `gpu_id` is set.
- `preflight_config_owned_startup_models[_with_gpus]` resolves any spec that
  actually has a device to resolve (CLI override or persisted pin),
  regardless of `config_owned`/`gpu.assignment`, while still failing closed
  exactly as before when a config-owned model requires a device under
  `gpu.assignment = "pinned"` and has none.
- Resolution mirrors the two-step shape `mesh-llm gpus tune` already uses
  (`resolve_pinned_with_backend_fallback` in
  `mesh-llm-commands/src/gpus/tune_hardware/evaluate.rs`, left unmodified):
  stable-id selectors resolve via `resolve_pinned_gpu_strict`; bare backend
  names (e.g. `--device CUDA0`) fall back to a direct `backend_device`
  match. An unresolvable device now produces a clear error naming the
  request and the available devices instead of the raw native ABI error.
- `preflight_config_owned_startup_models` (the production wrapper) no longer
  returns early for every non-pinned `gpu.assignment`, which had left the
  relaxed inner resolver unreachable in production: a default `auto` host
  launched with `--device` never resolved the selector. It now skips the
  hardware survey only when there is genuinely nothing to resolve — not
  pinned *and* no startup plan carries a device.
- The persisted lookup mirrors `ResolverContext::new`
  (`inference/skippy/resolver/resolution.rs`), which is what actually picks
  `hardware.device` up downstream: declared ref (the `--gguf <path> --model
  <alias>` alias), then model ref, then any row whose `hardware.model_path`
  names the same file. Previously the alias branch looked its pin up by GGUF
  path while config rows are keyed by alias, and a bare `--gguf <path>`
  could not match an alias-keyed row at all.
- A config pinned only through `[defaults.hardware] device` — a shape
  validation explicitly accepts under `gpu.assignment = "pinned"`, leaving
  every `models[].gpu_id` unset — failed startup preflight closed on a
  missing `gpu_id`. The persisted lookup now inherits the defaults device
  the same way `resolve_hardware_config` does. This was a pre-existing
  defect, not a regression from this PR, but it is the same failure mode
  and sits in the code this PR rewrites.

## Testing

TDD, failing-first: added tests reproducing the reported failure (CLI
`--model` matching a persisted PCI-stable-id pin) and the CLI-override
requirement, plus coverage for auto-assignment CLI overrides, unresolvable
devices, and `--device auto` as a no-op.

Review follow-ups are covered through `preflight_config_owned_startup_models`
— the production wrapper — rather than the inner GPU helper, so the outer
guard is exercised: `--device` under auto assignment must reach the resolver
and fail on an impossible selector, and a launch with no selector at all must
stay inert. Alias, `hardware.model_path`, and defaults-inheritance lookups are
pinned too. All three behavior tests fail against the previous commit and pass
after it.

- `cargo fmt --all --check`
- `cargo check -p mesh-llm-host-runtime` / `-p mesh-llm`
- `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings` / same for `-p mesh-llm`
- `cargo test -p mesh-llm-host-runtime --lib` — 2543 passed, 0 failed
- `cargo test -p mesh-llm-commands --lib gpus::tune_hardware` — unchanged, 10 passed
- `cargo test -p mesh-llm-system --lib --features skippy-devices hardware::skippy_devices` — unchanged, 4 passed
- `pinned_gpu_config_accepts_defaults_hardware_device_for_models` (plugin/config.rs) — unchanged, passes

## Out of scope

`crates/mesh-llm-config/src/model.rs::merge_hardware` (the top-level legacy
`gpu_id` ⇄ `hardware.device` config-file shim) was investigated and is
unrelated to CLI-vs-config precedence — left untouched, no behavior change.

## Summary by CodeRabbit

* **New Features**
  * Added GPU selection through `--device`, including stable device IDs and backend names.
  * Startup models can match by alias, model reference, or configured path.
  * Default hardware settings apply when model-specific settings are unavailable.
  * Supports case-insensitive `auto` selection and preserves explicit CPU selection.

* **Bug Fixes**
  * Explicit device selections now take precedence over saved or pinned settings.
  * Improved diagnostics for unavailable or ambiguous devices.
  * Refined GPU validation for automatically assigned models.


i386 (approved):

crates/mesh-llm-host-runtime/src/runtime/startup_models.rs
```diff
@@ -987,29 +1026,94 @@ pub(super) fn pinned_startup_preflight_metrics() -> &'static [hardware::Metric]
...
         specs.len() == plans.len(),
         "startup model preflight received mismatched specs/plans"
     );
 
     for (spec, plan) in specs.iter().zip(plans.iter_mut()) {
-        if !spec.config_owned {
+        // `gpu.assignment = "pinned"` requires every config-owned model to
+        // resolve to a concrete device (fail closed below when absent, same
+        // as before). CLI-explicit models and models under the default
+        // "auto" assignment only go through resolution when there is
```
ndizazzo, line 1188: This says Auto assignment can resolve an explicit `--device`, but the production wrapper returns before calling this helper whenever assignment isn't `Pinned`. Default Auto plus `--device` still leaves the plan unset. Please move or narrow the outer guard so explicit selectors reach this resolver, and test the real wrapper entry point.

ndizazzo (commented):
Needs revision. The production startup wrapper still bypasses the new Auto-assignment behavior.

Follow-ups:
- The `--gguf <path> --model <alias>` branch looks up the persisted pin by path even though normal config rows are keyed by alias/ref. Please cover that path too.
- Add a regression test through `preflight_config_owned_startup_models`, not only the inner GPU helper.

michaelneale:
@ndizazzo some tough failures it looks like - any ideas?

michaelneale:
🤖 Hint on the Linux smoke failures (posted by Mic's agent):

All three smoke suites fail the same way — the eager startup model dies in the new preflight:

```
startup model 'local-gguf/sha256-…' failed pinned GPU preflight: requested device 'CPU' did not match any detected GPU backend device. Available devices: none
```

**Root cause:** the CI smoke scripts launch with an explicit `--device CPU` (`scripts/ci-smoke-test.sh`, `scripts/ci-two-node-client-serving-smoke.sh`, etc.) on GPU-less runners. Before this PR, `RuntimeOptions.device` was never read on this path, so `--device CPU` was inert and CI worked by accident. Now the CLI override flows into `preflight_config_owned_startup_models_with_gpus` → `resolve_requested_startup_device`, whose name-fallback (`resolve_startup_backend_device_by_name`) only matches against the **GpuFacts inventory** — which never contains a CPU entry, and on these runners is empty (`Available devices: none`).

**Suggested fix:** `CPU` is a valid llama.cpp backend device but not a GPU, so it should short-circuit before GPU resolution. Two options that keep the fail-closed pinned semantics intact:

1. In `resolve_requested_startup_device` (or just before calling it), special-case a requested device that names the CPU backend (`backend_device_names_match(requested, "CPU")`): skip GPU resolution, leave `plan.pinned_gpu = None`, and pass the device straight through to the runtime config — i.e. restore the pre-PR behavior for CPU while keeping the new resolution for real GPU selectors.
2. Alternatively, when a `backend_probe` is available, fall back to resolving the name against `probe.available_devices` (which *does* include `CPU` — your own tests use `vec!["Vulkan0", "CPU"]`) instead of erroring when the GpuFacts match comes up empty. That handles CPU and any other non-GPU backend device uniformly.

Either way, `pinned_gpu` fields (`stable_id`, `vram_bytes`) don't exist for CPU, so option 1 is probably the smaller diff. Worth a unit test: CLI `--device CPU` with an empty GPU inventory must not fail preflight.

i386 (commented):
Summary from agent review (Jian Yang) — details inline.

**One blocking issue:** `--device CPU` is now treated as an unresolvable GPU selector, which breaks `serve` on every CPU-only host and has already reddened all three Linux product-smoke lanes at this head (run 32918059417). Root cause and suggested fix inline at `effective_startup_gpu_id`.

Everything else here is strong: the resolver-mirroring is verified against `ResolverContext::new` and `resolve_hardware_config`, the defaults-inheritance fix is real, and the test matrix (CLI override, auto assignment, alias/path/defaults lookups, fail-closed pinned) is exactly the shape I'd want. After the CPU regression is fixed this looks mergeable.

crates/mesh-llm-host-runtime/src/runtime/startup_models.rs
```diff
@@ -611,6 +612,94 @@ fn is_plain_model_alias(candidate: &str) -> bool {
...
+}
+
+/// Effective device selector for a startup model: an explicit CLI `--device`
+/// always wins over whatever is persisted for that model.
+fn effective_startup_gpu_id(options: &RuntimeOptions, persisted: Option<&str>) -> Option<String> {
+    options
+        .device
+        .as_deref()
+        .map(str::trim)
+        .filter(|device| !device.is_empty() && !device.eq_ignore_ascii_case("auto"))
```
i386, line 698: **Blocking: `--device CPU` no longer works — all three Linux product-smoke lanes are red because of it.**

`CPU` passes this filter (it isn't `auto`), so it becomes a `gpu_id`, the relaxed guard at line 1063 (`plans.iter().all(|plan| plan.gpu_id.is_none())`) no longer short-circuits, and `resolve_startup_backend_device_by_name` then fails against a GPU-only inventory:

```
startup model 'local-gguf/sha256-…' failed pinned GPU preflight:
requested device 'CPU' did not match any detected GPU backend device. Available devices: none
```

That exact error appears in run 32918059417 (this head): *Skippy Inference Smoke Tests*, *Two-node client smoke*, and *Two-node split smoke* all fail this way — the eager startup model never resolves, serve never surfaces a model, and the scripts time out polling `/models`.

`--device CPU` is an established invocation (`scripts/ci-smoke-test.sh:83` and `:232`, `scripts/ci-compat-smoke.sh:70`, `scripts/ci-two-node-client-serving-smoke.sh:98`, and `scripts/ci-two-node-split-smoke.sh` where `DEVICE` defaults to `CPU`), and `backend_device_for_flavor(BinaryFlavor::Cpu)` returns `None` — CPU is a pseudo-device, not a GPU backend name, so it can never match by name.

Suggested direction: treat `CPU` (case-insensitive, like `auto`) as "no GPU pin" here, or have the preflight resolve it to the CPU flavor instead of failing. A regression test that pushes `--device CPU` with a GPU-only (or empty) inventory through `preflight_config_owned_startup_models` would lock this in.

— Jian Yang (agent review)

crates/mesh-llm-host-runtime/src/runtime/startup_models.rs
```diff
@@ -611,6 +612,94 @@ fn is_plain_model_alias(candidate: &str) -> bool {
...
+/// Matching mirrors `ResolverContext::new` in
+/// `crate::inference::skippy::resolver::resolution`, which is what actually
+/// picks the persisted `hardware.device` up further downstream: the declared
+/// ref first (the alias from `--gguf <path> --model <alias>`), then the model
+/// ref string, then any config row whose `hardware.model_path` names the same
+/// file. Keeping the two lookups in step is what makes this preflight see
+/// exactly the pins that would otherwise reach the native ABI unresolved —
+/// a bare-path `--gguf` launch never matches a row keyed by alias, and the
+/// alias branch's `model_ref` is the GGUF path rather than the alias.
+fn matching_config_model_gpu_id<'a>(
```
i386, line 627: This is a careful mirror of `ResolverContext::new` (declared ref → model ref → `hardware.model_path`) and `resolve_hardware_config`'s defaults fallback — the lookup-order claims check out against `inference/skippy/resolver/resolution.rs:81` and `:335`, including the alias-vs-GGUF-path fix.

The risk is that it's a *manual* mirror: the two copies can drift apart again and silently reintroduce exactly the bug this PR fixes (preflight seeing pins the resolver doesn't, or missing ones it does). Consider extracting the "find the config entry for this launch shape" lookup into one shared helper the resolver and preflight both call, or at minimum a property-style test that walks config permutations (alias row, path-keyed row, defaults-only pin) through both paths and asserts they agree.

— Jian Yang (agent review)

i386 (changes requested):
**Review: still blocked, unchanged from the prior review.** The device-selector resolution design here (CLI-device-wins precedence, stable-id + bare-backend-name two-step resolution mirroring `gpus tune`, relaxed preflight reachability from production) is the more complete of the two designs in flight, but:

- 5 checks red at the current head (all three Linux product-smoke lanes, Linux CI, PR/Linux) from the known `--device CPU` preflight bug: a CPU selector passes the non-auto filter, becomes a `gpu_id`, then can't resolve against the GPU-only inventory (`resolve_pinned_gpu_strict` → `NonPinnableConfiguredId`/`NoPinnableGpus`).
- The branch is 3 days stale and now overlaps #1467 (merged-pending), which rewired the same `build_startup_model_specs`/preflight functions with a different structure.

Recommendation: rebase this PR's `--device` handling and bare-backend-name fallback onto #1467 once it lands, and add a `--device CPU` → CPU backend resolution (or explicit rejection with an actionable message) to close the red lanes. The pinned-preflight half of this PR is otherwise subsumed by #1467's exact-match + defaults-inheritance work.

i386 (approved):
Owner direction (James, 2026-08-28): approved with the standing comments remaining subject to major changes — the substantive findings (5 red lanes from the `--device CPU` preflight bug; overlap with #1467 requiring a rebase of the `--device` handling) still need addressing at merge. Approving per direction; recommend not merging until the red lanes are green and the #1467 rebase is done.

---
(prior review, still stands) — 5 checks red at head (all three Linux product-smoke lanes, Linux CI, PR/Linux): `--device CPU` passes the non-auto filter, becomes a `gpu_id`, then fails GPU-only resolution. Branch 3 days stale and now conflicts with #1467, which rewired the same `build_startup_model_specs`/preflight functions. The pinned-preflight half is subsumed by #1467's exact-match + defaults-inheritance work; rebase the `--device` handling and bare-backend-name fallback onto #1467 and add CPU resolution or an actionable rejection.

crates/mesh-llm-host-runtime/src/runtime/startup_models.rs
ndizazzo: Moved alias/ref/path matching into the shared `find_model_entry_for_refs` resolver helper and reused it in startup preflight, including alias, path, and defaults coverage.

crates/mesh-llm-host-runtime/src/runtime/local.rs
```diff
@@ -345,6 +347,9 @@ pub(super) fn resolve_local_openai_skippy_config(
     if let Some(gpu) = spec.pinned_gpu {
         resolved.hardware.device = Some(gpu.backend_device.clone());
     }
+    if let Some(device) = &spec.device_override {
+        resolved.hardware.device = Some(device.clone());
+    }
```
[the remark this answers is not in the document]
ndizazzo: Guarded both direct and layer-package pinned-device assignments when `device_override` is present, so explicit CPU and other overrides retain precedence.

crates/mesh-llm-host-runtime/src/runtime/startup_models.rs
ndizazzo: Rebased onto current main and preserved CPU as a non-GPU backend selector. It bypasses GPU preflight while flowing through runtime launch, with startup coverage for the explicit CPU override.

ndizazzo:
Rebased this branch onto current `main` (`97d056490`) and force-pushed the updated series at `c5cf6ac55`.

Addressed the remaining review feedback:
- preserved explicit CPU device overrides through startup and launch
- centralized config model alias/ref/path matching in the shared resolver
- kept explicit `device_override` precedence in both direct and layer-package pinned-device paths

Local validation from a fresh worktree: `just test-all` passed all 11 stages, including Rust/static Skippy tests, plugin and SDK checks, 1,660 UI unit tests, the website build, and 64 Playwright tests (2 skipped).

ndizazzo:
superseded by #1467

PR #1396, opened 2026-08-20
refactor: complete evidence-backed codebase audit waves

ndizazzo:
## Summary

This stack executes the validated codebase-audit waves commit by commit. It preserves the public API surface (verified: 0 dropped `pub` items, 0 dropped re-exports, 0 dropped UI exports vs `origin/main`), but it is **not** behavior-preserving — see "Behavior changes" below.

- removes compiler- and `cargo-udeps`-proven dead code/dependencies plus four verified UI orphans
- archives only two evidence-backed obsolete scripts; preserves live/manual/research tooling
- consolidates shared request classification, size parsing, request affinity, and bounded OpenAI compatibility helpers while retaining established import paths and consumer-specific policy
- replaces the environment-mutation TODO census with a repo-wide structural contract check
- splits the audited Rust and UI monoliths by semantic ownership, including FFI/protocol facades, logging, configuration, chat, topology, logs, and visualization
- preserves compatibility facades and public exports found by independent review (`mesh_llm_host_runtime::*`, `ModelCacheStatus`, `LogExportItem`, Skippy ABI constants)

The stack is a series of focused commits so individual waves remain reviewable.

## Behavior changes

Six production behavior changes ride along with this stack. Each one **fixes a pre-existing `main` bug** — none is a regression introduced here — but they are user-visible and were not previously declared. Listed for release-notes purposes.

1. **Credential redaction on native-runtime archive downloads** (`crates/mesh-llm-runtime-install/src/install.rs:197,289`). `url_without_query` was already applied to the manifest fetch on `main` but not to the archive download, so signed-URL query strings reached `anyhow` error context and the `NativeRuntimeDownloadProgress.url` event. Now redacted on all three paths, plus `reqwest::Error::without_url`. **Security-relevant; consider cherry-picking for rc6 independently of this stack.**
2. **URL userinfo redaction** (`crates/mesh-llm-runtime-install/src/manifest.rs`). `url_without_query` now also strips `user:pass@` from diagnostic URLs, mirroring `mesh-llm-host-runtime::logging::policy::redact_url_userinfo`. Previously only the query string was stripped.
[...]

PR #1419, opened 2026-08-24
fix(build): link libvendor-hash.a and test Rust on llama.cpp pin bumps

michaelneale:
## What is broken

`main` has not built since #1402 (`33185235`, 2026-08-22 20:47Z) advanced the llama.cpp pin `030ebb55` -> `3af988fa`. Every commit since fails `Rust tests (batch-2)` and `(batch-3)`:

```
rust-lld: error: undefined symbol: hash_sha256_hex[abi:cxx11](void const*, unsigned long)
error: could not compile `skippy-runtime` (lib test)
error: could not compile `skippy-server` (lib test)
```

Upstream moved SHA256 into a new static library `vendor::hash` (`vendor/CMakeLists.txt` -> `vendor/hash`) and links it into `mtmd` as **PRIVATE** (`tools/mtmd/CMakeLists.txt:82`). PRIVATE means the archive is not propagated to consumers. Our static link line is a hand-maintained list, so `libmtmd.a`'s call to `hash_sha256_hex` (`tools/mtmd/mtmd-helper.cpp:368`) has nothing to resolve against.

The C++ build and the inference smoke tests still pass — the symbol resolves inside CMake's own link. Only the Rust link line is short an archive.

## The fix

Add `vendor/hash/libvendor-hash.a` to all four places that enumerate static ABI archives:

- `scripts/build-llama.sh` (`required_static_archives`)
- `crates/skippy-ffi/build.rs` (search dir, rerun-if-changed, and `rustc-link-lib` emitted **after** `mtmd` so archive order resolves)
- `scripts/restore-static-abi-input.sh`
- `.github/actions/prepare-static-abi-input/action.yml`

plus the two contract tests that assert on those lists.

Of the five vendor libraries upstream added, `hash` is the only one producing an archive — `miniaudio`, `stb`, `sheredom`, `nlohmann` are all `add_library(... INTERFACE)`.

## Why CI did not catch it

`scripts/affected-crates.sh:171-181` escalates to a full-workspace test run for `Cargo.lock`, `Cargo.toml`, `rust-toolchain`, `.github/cache-version.txt` and the llama build scripts. It does **not** escalate on the llama.cpp pin or patch queue — there is a comment saying those are routed as backend builds instead.

#1402 changed 26 files, **all** under `third_party/llama.cpp/`. So it built the native libraries, ran the smoke tests green, and reported `Rust tests: skipped`. The one input that can change which archives the native build emits is the one input excluded from the only job that exercises the link line.

This also explains why a PR can be green and main red at the same commit content: a PR run tests change-scoped packages, main tests all of them. PR #1364's head ran 8 packages (not including `skippy-runtime`); the main run of the same tree ran 36 (including it).

Fix: escalate on `third_party/llama.cpp/upstream.txt` and `third_party/llama.cpp/patches/`.

## Verification

- `python3 -m unittest discover -s scripts/tests` — the two touched suites (`test_static_abi_artifacts`, `test_ci_artifact_actions`, 55 tests) pass. Remaining failures in `test_plan_ci` and the `yaml`-importing modules are **pre-existing and environment-local**: identical set with this diff stashed on the same worktree, and `PyYAML` is absent locally.
- Escalation verified by differential, not by reading: `printf 'third_party/llama.cpp/upstream.txt\n' | bash scripts/affected-crates.sh --stdin` now returns `all_rust=true` with 61 crates including `skippy-runtime`; the same input against `git show origin/main:scripts/affected-crates.sh` falls through to the dependency-graph path.
- `rustfmt --edition 2024 --check crates/skippy-ffi/build.rs` clean; `bash -n` clean on both scripts.
- **Not locally link-verified.** I have no CUDA/Linux static ABI build here — CI's `Rust tests` batches on this PR are the real proof, and they will now run because the escalation covers this PR's own changes to `build-llama.sh`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved static builds by ensuring the vendor hash library is included in native linking and ABI archives.
  * Builds and artifact restoration now detect missing vendor hash archives earlier, reducing incomplete-package failures.

* **Tests**
  * Expanded static ABI artifact validation to cover the vendor hash library and incomplete archive sets.
  * Updated change detection so relevant native source and patch updates trigger comprehensive Rust validation.

---

## Second, independent main breakage also fixed here: stale Swift UniFFI checksums

`macOS / Swift SDK input` also fails on `main`, at the step `Verify committed Swift binding source is current`. It is unrelated to archives, but `main` is not green until both are fixed, so it is in this PR rather than a follow-up.

Bisected by check-run history down `main`: green through `65035345a`, failing from `277de2adf` (#1396, "refactor: complete codebase audit / cleanup") onward. That PR moved every `uniffi::export` out of `crates/mesh-llm-ffi/src/lib.rs` and `crates/skippy-ffi/src/lib.rs` into submodules. UniFFI hashes `module_path!()` into each API checksum, so all 49 committed checksums in `sdk/swift/Sources/MeshLLM/Generated/mesh_ffi.swift` went stale and the `git diff --exit-code` gate has failed since 2026-08-21.

Regenerated by running the repo's own `sdk/swift/scripts/build-host-macos-xcframework.sh` end to end on an arm64 Mac — a real Metal llama.cpp build, a real `cargo build --release -p mesh-llm-ffi`, and the script's own `otool -tvV` checksum sync. Not hand-edited and not taken from bindgen alone: running `generate-swift-bindings.sh` on its own produces *different* values (e.g. `create_auto_client` 12117), because the authoritative checksums come from the compiled staticlib. The committed values match what CI computed (`create_auto_client` 12517 -> **16973**, identical to the diff in the failing job log), and re-running the sync against the regenerated file is a no-op, which is exactly what the gate checks.

Side benefit: that full macOS build also exercised the `vendor-hash` change on Metal/macOS, not just Linux.

### DIFF
--- a/.github/actions/prepare-static-abi-input/action.yml
+++ b/.github/actions/prepare-static-abi-input/action.yml
@@ -93,6 +93,7 @@ runs:
           "ggml/src/libggml.a"
           "ggml/src/libggml-base.a"
           "tools/mtmd/libmtmd.a"
+          "vendor/hash/libvendor-hash.a"
         )
         test -s "$build_stamp"
         test -s "$patched_sha_file"
--- a/crates/skippy-ffi/build.rs
+++ b/crates/skippy-ffi/build.rs
@@ -56,6 +56,7 @@ fn main() {
 
     let search_dirs = [
         build_dir.join("tools/mtmd"),
+        build_dir.join("vendor/hash"),
         build_dir.join("common"),
         build_dir.join("src"),
         build_dir.join("ggml/src"),
@@ -78,6 +79,10 @@ fn main() {
     for (unix_archive, msvc_archive) in [
         ("src/libllama.a", "src/llama.lib"),
         ("tools/mtmd/libmtmd.a", "tools/mtmd/mtmd.lib"),
+        (
+            "vendor/hash/libvendor-hash.a",
+            "vendor/hash/vendor-hash.lib",
+        ),
         ("common/libllama-common.a", "common/llama-common.lib"),
         (
             "common/libllama-common-base.a",
@@ -122,6 +127,16 @@ fn main() {
 
     if static_archive_exists(&build_dir, "tools/mtmd/libmtmd.a", "tools/mtmd/mtmd.lib") {
         println!("cargo:rustc-link-lib=static=mtmd");
+        // Upstream links mtmd against vendor::hash PRIVATE, so the archive is
+        // not propagated to consumers. It must follow libmtmd.a on the link
+        // line or hash_sha256_hex stays undefined.
+        if static_archive_exists(
+            &build_dir,
+            "vendor/hash/libvendor-hash.a",
+            "vendor/hash/vendor-hash.lib",
+        ) {
+            println!("cargo:rustc-link-lib=static=vendor-hash");
+        }
     }
     println!("cargo:rustc-link-lib=static=llama-common");
     println!("cargo:rustc-link-lib=static=llama-common-base");
--- a/crates/skippy-ffi/build.rs
+++ b/crates/skippy-ffi/build.rs
@@ -129,14 +129,23 @@ fn main() {
         println!("cargo:rustc-link-lib=static=mtmd");
         // Upstream links mtmd against vendor::hash PRIVATE, so the archive is
         // not propagated to consumers. It must follow libmtmd.a on the link
-        // line or hash_sha256_hex stays undefined.
-        if static_archive_exists(
+        // line or hash_sha256_hex stays undefined. mtmd without vendor-hash is
+        // not a degraded configuration to link anyway -- it is a guaranteed
+        // undefined symbol, so fail here with the cause rather than emitting a
+        // link line we know cannot resolve.
+        if !static_archive_exists(
             &build_dir,
             "vendor/hash/libvendor-hash.a",
             "vendor/hash/vendor-hash.lib",
         ) {
-            println!("cargo:rustc-link-lib=static=vendor-hash");
+            panic!(
+                "libmtmd is present in {} but vendor/hash/libvendor-hash.a is not; \
+                 mtmd requires vendor::hash for hash_sha256_hex. The llama.cpp \
+                 build or the restored static ABI artifact is incomplete.",
+                build_dir.display()
+            );
         }
+        println!("cargo:rustc-link-lib=static=vendor-hash");
     }
     println!("cargo:rustc-link-lib=static=llama-common");
     println!("cargo:rustc-link-lib=static=llama-common-base");
--- a/scripts/affected-crates.sh
+++ b/scripts/affected-crates.sh
@@ -168,10 +168,16 @@ main() {
       continue
     fi
 
-    # Escalation patterns. Native llama.cpp inputs are routed by
+    # Escalation patterns. Most native llama.cpp inputs are routed by
     # .github/actions/compute-changes as backend builds, not as all-Rust crate
-    # test fanout.
-    if [[ "$file" =~ ^Cargo\.lock$ ]] || \
+    # test fanout. The upstream pin and the patch queue are the exception: they
+    # can change which archives the native build emits, and the static link
+    # line in crates/skippy-ffi/build.rs is hand-maintained, so only the Rust
+    # test batches can prove the link still closes. Advancing the pin without
+    # this escalation is how an undefined hash_sha256_hex reached main.
+    if [[ "$file" =~ ^third_party/llama\.cpp/upstream\.txt$ ]] || \
+       [[ "$file" =~ ^third_party/llama\.cpp/patches/ ]] || \
+       [[ "$file" =~ ^Cargo\.lock$ ]] || \
        [[ "$file" =~ ^Cargo\.toml$ ]] || \
             [[ "$file" =~ ^scripts/(build-llama|prepare-llama|build-linux|build-linux-rocm|build-mac|build-windows|skippy-ci-smoke|ci-install-native-runtime|ci-prepare-native-runtime|ci-smoke-test|ci-compat-smoke|ci-client-auto-test|ci-two-node-client-serving-smoke|ci-two-node-split-smoke)\. ]] || \
        [[ "$file" =~ ^\.github/cache-version\.txt$ ]] || \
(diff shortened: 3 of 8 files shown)


i386 (approved):

michaelneale:
## This PR cannot go green until the Swift UniFFI binding is regenerated — that work belongs here

The vendor-hash fix in this PR is **correct and proven**. The blocker is a *separate* pre-existing break on `main` that fail-fast-cancels this PR's lanes before they can report. Per Mic: no broken merges, so the binding regeneration lands **in this PR**, not a follow-up.

### 1. The vendor-hash fix is verified working

At `cd78dbaa4`, the exact two link targets that fail on `main` both linked and ran green:

| Target | Result | Job |
|---|---|---|
| `skippy_runtime-5e01b8382cd03afa` | **74 passed, 0 failed** | `97312405940` (batch-3) |
| `skippy_server-a1c9fe9ca4c693f4` | **478 passed, 0 failed** | `97312405946` (batch-2) |

Control against a false green: `hash_sha256_hex` appears **1×** in the `main` batch-2 log (`97290385441`) and **0×** in this PR's batch-2 log (`97312405946`), with `skippy-runtime`/`skippy-server` compiled in both. The symbol resolves. The `affected-crates.sh` escalation also demonstrably works — it is why the batches ran at all.

### 2. Every red Linux job here is `cancelled`, not `failure`

`gh pr checks` renders both as "fail", which is misleading. Reading `conclusion` from the jobs API: all four Rust batches, both SDK inputs, and Windows CUDA are `cancelled`, step 14 killed mid-run, **no error in any of them**. One real failure at `03:35:23Z` tripped `.github/workflows/pr-cancel-sibling-runs.yml` (`cancel-in-progress: true`) and everything else died at `03:37:0xZ`.

The sole genuine failure:

```
macOS / Swift SDK input (host-only), step 20
"Verify committed Swift binding source is current"
generated Swift UniFFI bindings changed during the protected build
```

### 3. Root cause of that failure — and it is not this PR

It reproduces on `8b1812257` (`97290064899`) and `8c3db3dbc` (`97276272108`), both predating this PR. Last green `Swift SDK input (full)` on main was `65035345a` (2026-08-21). This PR touches 7 files, none of them Rust source, the FFI crate, or the Swift SDK — it cannot move a uniffi checksum.

**Mechanism:** all **49 of 49** checksums changed, not a subset — so this is not API drift. uniffi hashes `module_path!()` into every metadata buffer (`uniffi_macros-0.31.0/src/fnsig.rs:281,296,311,329`). #1396 (`277de2adf`, "complete codebase audit / cleanup") moved every `uniffi::export` out of `lib.rs` into submodules — `lib.rs` went **17 exports → 0**, redistributed across `node.rs` (3), `client.rs` (3), `native_runtime.rs` (6), `native_runtime_types.rs` (2), `console.rs`, `identity.rs`, `public_mesh.rs` (1 each). Symbol *names* stayed stable because they derive from the crate name only (`crate_name()` = `module_path.split("::").next()`, `uniffi_meta-0.31.0/src/lib.rs:652`), but every hashed *value* shifted.

### 4. What to do in this PR

Regenerate and commit `sdk/swift/Sources/MeshLLM/Generated/mesh_ffi.swift`. The generator is already wired up — `sdk/swift/scripts/build-host-macos-xcframework.sh:57-95` disassembles `libmeshllm_ffi.a` with `otool -tvV` and rewrites the checksums in place. Running the host-only script on macOS updates the file; commit the result.

Precedent: `d4091455c` ("fix(swift): synchronize generated binding checksums", #1122) — same failure, same remedy.

Two notes for whoever picks this up:

- **The regeneration will be complete.** I checked for `cfg(feature = ...)` on the `uniffi::export` sites — there are none, so the host-only build sees the same 49-item export surface as the full build. No risk of a partial rewrite that flips half the checksums.
- **Verify by re-running step 20's own check**, not by eyeballing the diff: `git diff --exit-code -- sdk/swift/Sources/MeshLLM/Generated/mesh_ffi.swift` must come back clean *after* a fresh build, since that is literally the gate (`swift-sdk-artifact.yml:233`).

### 5. Correction on `llama-quant-ffi/build.rs`

Right verdict, wrong reason in the thread above. The load-bearing fact is not the dependency graph: `crates/llama-quant-ffi/build.rs` has **zero** occurrences of `mtmd` (vs **5** in `crates/skippy-ffi/build.rs`). It never links the library that calls `hash_sha256_hex`, so it cannot reference the symbol regardless of test reachability. **Do not add vendor-hash to that file.**

---

Verified against `cd78dbaa4` and `774250675` (PR heads) and `8b1812257` (main); uniffi 0.31.0 per `Cargo.lock`. Note `774250675` ("fail fast when mtmd is present without vendor-hash") landed after my first pass and does not touch the Swift binding — the gap above is still open.

michaelneale:
Pushed `94548c98f` — regenerated `sdk/swift/Sources/MeshLLM/Generated/mesh_ffi.swift` (49/49 checksums) so the binding fix lands in this PR rather than a follow-up.

**How it was produced:** the repo's own generator step, verbatim — `otool -tvV` over `target/aarch64-apple-darwin/release/libmeshllm_ffi.a` built at main `8b1812257` with `--no-default-features --features host,embedded-runtime`, then the checksum rewrite from `sdk/swift/scripts/build-host-macos-xcframework.sh:57-95`. Not hand-edited.

**Verification of the actual CI gate, not the diff:** re-running the sync against the regenerated file is a byte-for-byte no-op, so `git diff --exit-code -- sdk/swift/Sources/MeshLLM/Generated/mesh_ffi.swift` (`swift-sdk-artifact.yml:233`) passes. That is the condition step 20 checks.

**Scope:** the commit touches exactly one file. `git diff --name-only 8b1812257..HEAD -- crates/mesh-llm-ffi` is empty across this whole branch, so no API surface moved — only the stale hashes.

**Caveats I am not glossing over:**
- All 49 changed, consistent with @Dario's module-path inference from #1396, but I regenerated rather than reproduced #1396's before/after — the mechanism is still inference, the *values* are now measured.
- Built on Apple Silicon host-only. If the `full` lane's cross-target build ever produced different checksums this would resurface there; symbol values derive from metadata, not target, so I do not expect it, but the full lane is the proof.
- `crates/llama-quant-ffi/build.rs` deliberately untouched — zero `mtmd` references, so it cannot reference `hash_sha256_hex`.

CI on this head is the remaining proof for both fixes.

Issue #1332, opened 2026-08-14
Experiment: Capsule Emit inference receipts as a plugin


ndizazzo:
> Status: **Draft prototype specification**
>
> Depends on #1331. Related umbrella: #1233. Prototype source and prior sidecar evidence: [action-state-group/capsule-emit-mesh](https://github.com/action-state-group/capsule-emit-mesh).

## Summary

Build a native Rust, installable Capsule Emit plugin that produces signed, hash-chained Agent Action Capsules for OpenAI inference exchanges served by mesh-llm.

This is a prototype experiment to validate:

- the plugin lifecycle contract in #1331;
- a concrete inference-receipt field mapping;
- local signing, verification, chaining, retrieval, and optional transparency anchoring;
- non-streaming, streaming, error, denial, and cancellation behavior;
- composition between mesh-llm inference receipts and agent/tool action capsules.

The plugin must consume host-provided request/route/response lifecycle data over plugin IPC. It must not listen as an HTTP reverse proxy, require clients to use a second port, replace an inference backend, or require a custom mesh-llm binary.

## Experiment hypothesis

An installable plugin can create a useful, portable, independently verifiable record that says:

> This collector observed mesh-llm accept this request commitment, select this claimed model/runtime route, and return this response commitment.

The prototype does **not** establish that the claimed model performed the computation correctly. A self-signed capsule can cryptographically seal a false runtime claim. Model artifact verification, native compute receipts, hardware-bound signing, TEE/GPU evidence, statistical fingerprints, and distributed stage receipts remain separate assurance layers.

The experiment succeeds if it proves the integration and evidence-envelope mechanics without overstating the resulting trust level.

## Why a plugin

The existing [Capsule Emit mesh experiment](https://github.com/action-state-group/capsule-emit-mesh) proved the capsule format against a real mesh-llm node and Goose session, but it had to be a sidecar because plugins could not receive mesh-llm's own OpenAI request envelope.

That architecture had unavoidable limitations:

- clients had to target a second proxy port;
- the collector sat outside the mesh-llm trust boundary;
- streaming was buffered and re-synthesized;
- client-compatibility transformations occurred in the sidecar;
- it could not distinguish gateway, route-selection, and serving-host facts;
- it could not prove whether a refusal happened before dispatch.

#1331 is intended to remove those limitations generically. This issue validates that work with a real consumer.

### In scope

- /v1/chat/completions;
- streaming and non-streaming responses;
[...]

PR #1421, opened 2026-08-24
spec(openai-frontend): OpenAI exchange lifecycle hooks for out-of-process plugins (#1331 reference)

StevenMih:
Reference implementation of the OpenAI exchange lifecycle contract specified in #1331 (the plugin-platform prerequisite surfaced by the Capsule Emit sidecar experiment, #1233 / #1332).

## What this adds
Two hooks on `OpenAiHookPolicy`, wired into `HookedOpenAiBackend`, so an **out-of-process plugin** can observe the exchange without becoming an HTTP reverse proxy:
- **`on_effective_chat_completion`** — the effective request + selected serving route, before dispatch.
- **`on_chat_completion_terminal`** — the terminal event (success / error / denial), after execution.

## What building it against the real runtime taught us
- Today `OpenAiHookPolicy` is compiled into the backend wrapper and only the before-chat hook fires; the effective-request and terminal phases weren't reachable by an installable plugin. These two hooks close that gap.
- There's a **second, disjoint dispatch path** (the raw-proxy ingress) that `OpenAiHookPolicy` never touches — this PR covers it too, so a plugin sees both paths with no silent bypass. (Design note in the branch maps #1331's acceptance criteria to the actual seam.)

## Status
- `cargo clippy -D warnings` + `cargo test` green (187 tests, incl. 3 new mutant-verified).
- Non-streaming chat-completions first; streaming/delegation are follow-ups, flagged in the design note.

## How to take it
This is offered as a **reference to adopt or clean-room** — whichever the maintainers prefer. Happy to align the hook/phase names to whatever fits the runtime's own conventions. The consuming plugin (a native Rust capsule producer) lives at `action-state-group/capsule-emit-mesh` and is what #1332 tracks; these hooks are its one platform dependency.

cc @michaelneale @ndizazzo

## New Features
- Added lifecycle tracking for non-streaming chat completions, including effective requests and terminal outcomes.
- Added optional capsule markers and exposed valid capsule IDs through the `X-Capsule-Id` response header.
- Added correlated exchange events across typed frontend and proxy request paths.
- Added plugin message broadcasting to all eligible destinations with consolidated delivery errors.
- Added status reporting for successful, failed, denied, and dropped plugin exchanges.

## Documentation
- Added design documentation covering lifecycle events, capsule propagation, plugin exchanges, and current limitations.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
+++ b/crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
@@ -48,6 +48,12 @@ pub enum OpenAiExchangePhase {
 /// either being forced into the other's type.
 #[derive(Debug, Clone, PartialEq, Serialize)]
 pub struct OpenAiExchangeEnvelope {
+    /// Stable per-exchange id, minted when the dispatch path admits the
+    /// request. Shared by an exchange's `EffectiveRequest` and `Terminal`
+    /// envelopes (and mirrored into the transport `correlation_id`), so a
+    /// plugin can pair the two events for one exchange even when concurrent
+    /// requests on the same model are in flight.
+    pub exchange_id: String,
     pub dispatch_path: OpenAiExchangeDispatchPath,
     pub phase: OpenAiExchangePhase,
     pub model: String,
@@ -65,8 +71,13 @@ pub struct OpenAiExchangeEnvelope {
 }
 
 impl OpenAiExchangeEnvelope {
-    pub fn effective(dispatch_path: OpenAiExchangeDispatchPath, model: impl Into<String>) -> Self {
+    pub fn effective(
+        exchange_id: impl Into<String>,
+        dispatch_path: OpenAiExchangeDispatchPath,
+        model: impl Into<String>,
+    ) -> Self {
         Self {
+            exchange_id: exchange_id.into(),
             dispatch_path,
             phase: OpenAiExchangePhase::EffectiveRequest,
             model: model.into(),
@@ -77,12 +88,14 @@ impl OpenAiExchangeEnvelope {
     }
 
     pub fn terminal(
+        exchange_id: impl Into<String>,
         dispatch_path: OpenAiExchangeDispatchPath,
         model: impl Into<String>,
         status: Option<u16>,
         marker: Option<CapsuleMarker>,
     ) -> Self {
         Self {
+            exchange_id: exchange_id.into(),
             dispatch_path,
             phase: OpenAiExchangePhase::Terminal,
             model: model.into(),
@@ -114,7 +127,12 @@ impl OpenAiExchangeChannel for PluginManager {
             }
         };
         if let Err(error) = self
-            .broadcast_channel_message(OPENAI_EXCHANGE_CHANNEL, "application/json", body)
+            .broadcast_channel_message(
+                OPENAI_EXCHANGE_CHANNEL,
+                "application/json",
+                body,
+                &event.exchange_id,
+            )
             .await
         {
             tracing::warn!(%error, "failed to publish openai exchange event to plugins");
@@ -147,6 +165,7 @@ impl OpenAiHookPolicy for OpenAiExchangeHookBridge {
     ) {
         self.channel
             .publish(&OpenAiExchangeEnvelope::effective(
+                route.exchange_id.clone(),
                 OpenAiExchangeDispatchPath::TypedFrontend,
                 route.model.clone(),
             ))
@@ -156,6 +175,7 @@ impl OpenAiHookPolicy for OpenAiExchangeHookBridge {
     async fn on_chat_completion_terminal(
         &self,
         request: &ChatCompletionRequest,
+        exchange_id: &str,
         outcome: &ChatCompletionOutcome<'_>,
     ) {
         let (status, marker) = match outcome {
@@ -165,6 +185,7 @@ impl OpenAiHookPolicy for OpenAiExchangeHookBridge {
         };
         self.channel
             .publish(&OpenAiExchangeEnvelope::terminal(
+                exchange_id,
                 OpenAiExchangeDispatchPath::TypedFrontend,
                 request.model.clone(),
                 Some(status),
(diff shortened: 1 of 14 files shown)


crates/openai-frontend/src/hooks.rs
```diff
@@ -140,11 +228,53 @@ impl OpenAiBackend for HookedOpenAiBackend {
...
-        self.backend
-            .chat_completion_with_context(request, context)
-            .await
+        let route = ChatExchangeRoute::for_request(&request, exchange_id.clone());
+        self.hooks
+            .on_effective_chat_completion(&request, &route)
+            .await;
+        let mut result = self
+            .backend
+            .chat_completion_with_context(request.clone(), context)
```
i386, line 253: Non-blocking: this `request.clone()` now runs on **every** non-streaming chat completion through `HookedOpenAiBackend`, even when no hook implements the new observer methods — and it clones the full message history (which for long-context clients is the bulk of the payload).

Options if this ever matters in production wiring: (a) restructure so the terminal hook borrows before the request is moved into the backend (e.g. take `&ChatCompletionRequest` and clone inside the backend dispatch), or (b) gate the clone on a `hooks_want_post_dispatch_request()` capability probe defaulting to false. For a reference PR this is fine — but worth a line in the design note's deferred list since the production milestone will inherit it.

crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
```diff
@@ -0,0 +1,453 @@
...
+    async fn publish(&self, event: &OpenAiExchangeEnvelope) {
+        let body = match serde_json::to_vec(event) {
+            Ok(body) => body,
+            Err(error) => {
+                tracing::warn!(%error, "failed to serialize openai exchange event");
+                return;
+            }
+        };
+        if let Err(error) = self
+            .broadcast_channel_message(
```
i386, line 130: Non-blocking, but the doc comment on `OpenAiExchangeChannel` says "fire-and-forget by design ... must never affect whether the client's own request succeeds" — while `PluginManager::publish` awaits `broadcast_channel_message`, which awaits each plugin's `send_unsolicited`, which awaits `outbound_tx.send(...)` on a **bounded** `mpsc::channel(256)` (plugin/runtime.rs:213). A plugin process that stalls while its outbound queue is full will backpressure this publish — and path 1's publish is awaited inline inside `chat_completion_with_context`, so it delays request dispatch, not just the terminal event. The retry loop in `send_unsolicited` (restart attempt ×2) adds more worst-case latency.

Delivery failure is already correctly non-fatal (warn + continue), so the *correctness* claim holds; it's the latency/isolation claim that doesn't, strictly. Suggest either (a) softening the doc to "failure-tolerant" and noting the bounded-channel backpressure case, or (b) making the production publish truly non-blocking (e.g. `try_send` or a spawn to a delivery task with a drop-on-slow-consumer policy). Worth deciding before the main-wiring milestone adopts this on the hot path.

i386 (approved):
Reviewed from a fresh frame against the actual call sites (including `PluginRuntime::send_unsolicited` and the outbound `mpsc::channel(256)`).

**Verdict: approve.** This is a well-scoped reference implementation with an unusually honest design note — the two-dispatch-path discovery, the M1 "can't" verdict plus M2 closing exactly the gaps M1 named, and a plainly stated scope boundary (path 2 call sites compile + are wired but not behaviorally covered; main-wiring composition deferred) are all the right calls for a reference PR.

What I verified:
- Hooks are additive with default no-ops, so existing implementors (`MeshAutoHookPolicy` et al.) compile unchanged.
- Terminal fires exactly once on all three outcomes, including the before-hook `Err` path, and denials skip dispatch.
- `capsule_marker` is `#[serde(skip)]` and the `X-Capsule-Id` header rides the same `Response::extensions()` relay `TerminalUsage` already uses — conditional on the hook minting, with a negative test.
- `aggregate_broadcast_results` collects every recipient's outcome before erroring (no first-failure suppression) and has a regression test for it.
- Exchange-id pairing under concurrency on the same model is tested with `start_paused` — the right test shape.

Two non-blocking comments inline (request clone on the hot path; backpressure semantics of the "fire-and-forget" publish). Neither blocks merge for a reference implementation, but both are worth a sentence in the design note before the production-wiring milestone.

crates/openai-frontend/src/hooks.rs
i386, line 253: Suggestion: this clones the full `ChatCompletionRequest` on every non-streaming completion, even when no hook consumes the effective request. If the snapshot is only needed for observation, consider capturing it lazily (only when a hook is registered) or extracting a lightweight summary — the clone cost is inherited by the production milestone wiring.

crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
```diff
@@ -0,0 +1,453 @@
...
+/// [`OpenAiHookPolicy`]'s own observer methods: exchange delivery to a
+/// plugin must never affect whether the client's own request succeeds.
+#[async_trait]
+pub trait OpenAiExchangeChannel: Send + Sync + 'static {
+    async fn publish(&self, event: &OpenAiExchangeEnvelope);
+}
+
+#[async_trait]
+impl OpenAiExchangeChannel for PluginManager {
+    async fn publish(&self, event: &OpenAiExchangeEnvelope) {
```
i386, line 121: Suggestion: `publish` is documented fire-and-forget, but it awaits `broadcast_channel_message`, which hands off to a bounded (256) outbound channel inline. A stalled plugin can therefore delay request dispatch. Consider `try_send` + spawn (dropping or logging on a full channel), or soften the doc so callers know a slow plugin can back-pressure dispatch — worth resolving before main-wiring.

i386 (commented):
Two suggested (non-blocking) improvements from review:

i386 (changes requested):

StevenMih:
Thanks @i386   
The three data-integrity items are addressed in 6515645: the /v1/responses capsule-marker propagation, the stable exchange correlation ID, and continue-delivery after a single plugin send fails. @i386's two non-blocking notes — the request.clone() on every non-streaming completion, and the fire-and-forget/bounded-channel await — are being tightened now. The 2001-line mod.rs split is real but it's mesh-llm-internal maintainability beyond this capsule-integration PR's scope; happy to file it as its own refactor issue rather than balloon this PR. I'm going to close this and reopen a clean, squashed PR with the full set for an easier re-review, same branch, same #1331/#1332 reference.

PR #1428, opened 2026-08-24
fix(skippy): harden scheduler-only production cutover

i386:
## Outcome

Hardens the iteration-scheduler-only serving cutover for direct production rollout. This PR is stacked on #1420 and does not restore any legacy serving scheduler.

## Changes

- bounds the scheduler command queue and rejects overload without blocking request threads
- caps command draining per turn so sustained control traffic cannot starve inference
- contains scheduler-worker panics and fails active/queued requests deterministically
- maps scheduler overload to HTTP 429 and exposes cumulative failure/overload telemetry
- adds SKIPPY_ITERATION_SCHEDULER_SAFE_MODE=1 to serialize the new scheduler during incidents
- removes the stale deleted-batcher configuration documentation
- adds an architecture guard preventing the three deleted serving-scheduler modules from returning

The remaining frontend/decode_scheduler.rs is the speculative verify-window state machine, not the removed serving scheduler.

## Verification

- skippy-scheduler: 10 passed
- skippy-server: 495 passed, 0 failed, 3 fixture-gated ignored
- skippy-scheduler clippy: green with warnings denied
- skippy-server clippy: green with warnings denied

Model-backed production-path verification is running against exact head 22113332def6e7cc1331b684ba75a0ddc3a29078 before this draft is marked ready.

### DIFF
--- a/crates/skippy-scheduler/src/engine.rs
+++ b/crates/skippy-scheduler/src/engine.rs
@@ -233,6 +233,10 @@ impl Scheduler {
                 .collect(),
             prefix_hits: self.metrics.prefix_hits,
             prefix_misses: self.metrics.prefix_misses,
+            finished: self.metrics.finished,
+            failed: self.metrics.failed,
+            cancelled: self.metrics.cancelled,
+            rejected_overload: self.metrics.rejected_overload,
         }
     }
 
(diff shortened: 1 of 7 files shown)


i386:
Superseded as requested: the hardening commit is now incorporated directly into #1420. The reconciled #1420 head is d02b3e16ba198b52b02e1176b31d02fbc3ae0d80, including the concurrent CodeRabbit fix corrected so queue arbitration does not advance an unexecuted scheduler plan.

Issue #1434, opened 2026-08-24
Family-tier CI battery: parity oracle on llama bumps, hybrid boundary sweeps, tier-3 family removals


i386:
## Context

Split-model serving on mixed-hardware labs shows inconsistent, family- and split-boundary-dependent failures. Root-cause review of #1420 and #1429 shows the failures are bugs, not methodology, and that CI currently proves "the splits we tested still run" rather than "any split we'd ship is equivalent to not splitting at all."

Key findings motivating this issue:

- **B1 (High):** `model_requires_recurrent_state` (`crates/skippy-server/src/kv_integration/config.rs:173-186`) probes only the *first* layer file in `[layer_start, layer_end)`. Interleaved hybrids (Kimi-Linear, Qwen3-Next) whose stage starts on an attention layer misresolve `Auto` → `ResidentKv`, so recurrent state never crosses the boundary. Whether it fails depends on where the planner cuts. Not fixed by #1429; the unified radix tree makes payload selection the single point of failure.
- **B2 (High):** patch 0001 drops `synchronize()` before `embd_seq.clear()` in encode/decode and from `~llama_context()` while async copies may be in flight → nondeterministic corruption scaling with model size.
- **B4/B5 (High):** `stage_export_kv_page` never exports `k_idx` indexer rows (GLM-DSA/MSA silently wrong; blocks context-shift) and requires uniform K/V row sizes across stage layers.
- #1429 adds two HIGH bugs in the radix insert/replace path (concurrent duplicate resident record leaks a native seq; replace resets `active_refs`), plus a regression: serving-KV record ladder removed (partial compensation for KvRecurrent).

This issue covers the **CI and family-matrix work** to keep these bug classes from regressing (bug fixes themselves tracked separately).

## Family tiers

Artifacts are the certification set CI actually runs — not the largest available model per family. File sizes are the GGUF download size per artifact, pulled from the artifact source (HF tree API); `~` marks an estimate for repos behind auth where the tree API is unavailable. Package-backed rows show the full layer-set size; CI materializes only the owned stage range.

### Tier 1 — full battery on every llama bump

Parity oracle (full-model vs split, token-exact) + dtype lanes + boundary sweep for hybrids.

| Family | Cert artifact | File size | Why Tier 1 |
|---|---|---|---|
| Qwen3 dense | Qwen3-0.6B Q8_0 | 0.64 GB | Core prod family; state-size baseline |
[...]

PR #1435, opened 2026-08-24
fix(skippy): repair split-serving correctness bugs (B1/B2/B4/B6/B8 + 0016 ABI)

i386:
Stacked on #1429 (do not merge independently).

## What this fixes

Split-model serving on mixed-hardware labs no longer produces family- and boundary-dependent garbage:

- **Hybrid stages keep their recurrent state (B1):** `model_requires_recurrent_state` now probes every layer file in `[layer_start, layer_end)`, so a Kimi-Linear / Qwen3-Next stage that starts on an attention layer resolves `Auto` → `KvRecurrent` instead of `ResidentKv`. This was the canonical "every attempt is a new science experiment" bug.
- **No more nondeterministic corruption on larger models (B2):** the llama patch queue dropped `synchronize()` before `embd_seq.clear()` in encode/decode and from `~llama_context()` while async tensor-get copies could still be in flight. Both syncs are restored.
- **GLM-DSA / MiniMax-MSA KV pages are correct (B4)::** `stage_export_kv_page` / `stage_import_kv_page` now carry the `k_idx` indexer rows (new `SKIPPY_KV_PAGE_FLAG_HAS_K_IDX` flag + `k_idx_row_bytes` descriptor field). Previously sparse attention silently used stale indexer state and context-shift was blocked.
- **Embedding reorder corruption (B6):** `output_reorder` swaps pooled embedding rows at `n_embd_out` stride again, not `n_embd`.
- **No more size-dependent ggml aborts on unlisted arches (B8):** the scaled `graph_max_nodes` budget (`max(n_tokens*40, 32*n_tensors)`, KIMI_K3 keeps 160) applies to every architecture instead of a hand-picked allowlist.
- **Mixed-iteration batch ABI hardening (0016):** activation inputs wider than the stage model's input embedding are rejected with `UNSUPPORTED` instead of overflowing the batch row write, and the Rust-side session token bookkeeping is computed atomically after a successful batch (a mid-loop overflow error can no longer leave a subset of sessions desynced from native state).

## Protocol / ABI

`skippy_kv_page_desc` and `skippy_kv_page_component_desc` grow `k_idx_row_bytes`; `SKIPPY_ABI_VERSION_PATCH` bumps 40 → 41 with the Rust `skippy-ffi` constants and mirrors in sync. Old and new runtimes reject mixed pairing via the ABI version gate, so no silent interop.

## Validation

- Patch queue regenerated from the pinned upstream and round-tripped through `scripts/prepare-llama.sh pinned` (applies clean, 17 patches).
- `cargo check` / `clippy -D warnings` / `cargo test --lib` green for `skippy-ffi`, `skippy-runtime`, `skippy-server`; `cargo check -p mesh-llm` clean; `cargo fmt --all --check` clean.
- New unit test: layer-package inspection selects every in-range layer file (the B1 shape).
- Skippy API docs regenerated (`SKIPPY_ABI_VERSION_PATCH = 41`).

Closes the bug-fix half of #1434 (CI/family-matrix work follows as a stacked PR on this one).

## Summary by CodeRabbit

- **New Features**
  - Added mixed iteration batches supporting variable-length requests, per-request activations, and optional sampling.
  - Added staged model execution and ordered multi-part model loading.
  - Added session state management, KV-cache transfer, trimming, checkpoint retirement, and resident-prefix handling.
  - Added composite ISWA KV-cache support and improved cache portability.
- **Bug Fixes**
  - Improved session consistency when batch updates encounter errors.
  - Strengthened model, cache, and request validation.
- **Documentation**
  - Updated the documented ABI version to 41.

### DIFF
--- a/third_party/llama.cpp/patches/0001-Add-staged-model-graph-and-family-support.patch
+++ b/third_party/llama.cpp/patches/0001-Add-staged-model-graph-and-family-support.patch
@@ -228,7 +228,7 @@ index 7159e23bf..909e34290 100644
      LLM_TENSOR_VISEXP_ATTN_OUT,
      LLM_TENSOR_VISEXP_FFN_GATE,
 diff --git a/src/llama-context.cpp b/src/llama-context.cpp
-index 52f8d5367..f22aa2532 100644
+index 52f8d5367..9d955ae18 100644
 --- a/src/llama-context.cpp
 +++ b/src/llama-context.cpp
 @@ -10,7 +10,6 @@
@@ -341,17 +341,7 @@ index 52f8d5367..f22aa2532 100644
  
      if (cparams.n_ctx_seq < hparams.n_ctx_train) {
          LLAMA_LOG_INFO("%s: n_ctx_seq (%u) < n_ctx_train (%u) -- the full capacity of the model will not be utilized\n",
-@@ -479,9 +482,6 @@ llama_context::llama_context(
- }
- 
- llama_context::~llama_context() {
--    // wait for any pending asynchronous copies into the output buffers before they are freed
--    synchronize();
--
-     if (!model.hparams.no_alloc) {
-         for (size_t i = 0; i < backend_ptrs.size(); ++i) {
-             ggml_backend_t             backend = backend_ptrs[i];
-@@ -503,6 +503,22 @@ llama_context::~llama_context() {
+@@ -503,6 +506,22 @@ llama_context::~llama_context() {
  
  void llama_context::resolve_fused_ops(const llama_memory_context_i * mctx, uint32_t n_seqs) {
      const char * func = __func__;
@@ -374,7 +364,7 @@ index 52f8d5367..f22aa2532 100644
      auto resolve = [&](const llm_fused_op_probe & probe, bool & enabled) {
          if (!enabled) {
              return;
-@@ -1169,7 +1185,7 @@ void llama_context::set_embeddings_nextn(bool value, bool masked) {
+@@ -1169,7 +1188,7 @@ void llama_context::set_embeddings_nextn(bool value, bool masked) {
  void llama_context::set_embeddings_layer_inp(uint32_t lid, bool enable) {
      LLAMA_LOG_DEBUG("%s: lid = %d, enable = %d\n", __func__, lid, enable);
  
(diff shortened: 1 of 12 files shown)


michaelneale (approved):

michaelneale (approved):
yeah - need this, lots of failures though.

third_party/llama.cpp/patches/0001-Add-staged-model-graph-and-family-support.patch
```diff
@@ -2182,27 +2180,33 @@ index 2e2bd7dc6..117de97f9 100644
 +        }
 +        const uint32_t n_embd_v = hparams.n_embd_v_gqa(il);
 +        const bool has_v_stream = layer.v_stream.size() > strm && layer.v_stream[strm] != nullptr;
++        const bool has_k_idx_stream = layer.k_idx_stream.size() > strm && layer.k_idx_stream[strm] != nullptr;
```
ndizazzo, line 2183: This only exports indexer rows when a `k_idx` stream already exists, but the patch leaves `indexer_kv` false and never wires model metadata to enable it. That keeps `n_embd_k_idx` at zero, so the stream and `HAS_K_IDX` flag are never created. GLM-DSA also keeps its indexer state in `get_lid()`, not the cache selected by this path. Please wire the metadata, use the right cache, and prove the round trip with the affected model families.

third_party/llama.cpp/patches/0011-Wire-staged-runtime-builds-and-tests.patch
```diff
@@ -1205,68 +1205,63 @@ index 9556bfd84..446da8dad 100644
...
  LLAMA_API enum skippy_status skippy_import_recurrent_state(
          struct skippy_session * session,
          const void * input,
-         size_t input_bytes,
-         struct skippy_error ** out_error);
- 
-+/** @brief Trims the session's token history and associated state. */
- LLAMA_API enum skippy_status skippy_trim_session(
-         struct skippy_session * session,
+@@ -81,12 +93,14 @@ LLAMA_API enum skippy_status skippy_trim_session(
```
ndizazzo: Regenerating this patch drops the adjacent `@brief` comments for the public session functions. The prepared queue applies, but `generate-skippy-api-doc.py --check` then fails on `skippy_trim_session` and the website reference can't be regenerated. Please keep those comments in the owning patch and commit the regenerated API page.

ndizazzo (changes requested):
Needs revision. The indexer-state fix is still unreachable, and the regenerated public API patch also drops required documentation.

Follow-ups:
- Add real GLM-DSA and MiniMax-MSA round-trip coverage that proves `HAS_K_IDX` is emitted and restored.
- Regenerate the Skippy API reference and run the generator's `--check` mode after restoring the comments.

i386:
-

i386:
-

i386:
Superseded by #1436, which is now based directly on `main` and carries the B1–B8 fixes plus the certified-family CI battery.

The documentation correction was re-homed into #1436 at `7917eebb`, and the scheduler panic-containment ordering fix was preserved at `a8e1db5c`. Closing this PR to keep the forward line in one place.

PR #1437, opened 2026-08-25
spec(openai-frontend): OpenAI exchange lifecycle hooks for out-of-process plugins (#1331)

StevenMih:
Reference implementation of the OpenAI exchange lifecycle contract in #1331 (the plugin-platform prerequisite surfaced by the Capsule Emit sidecar experiment, #1233 / #1332). **Supersedes #1421** — reopened clean after addressing review.

## What this adds
Two hooks on `OpenAiHookPolicy`, wired into `HookedOpenAiBackend`, so an out-of-process plugin can observe the exchange without becoming an HTTP reverse proxy — plus the rung-ladder response-leg verdict.

## Review addressed (from #1421)
- **/v1/responses capsule marker propagation**, **stable exchange correlation ID**, and **continue-delivery after one plugin send fails** — all fixed (commit `6515645`).
- Two non-blocking polish items from @i386 (the per-request `clone()` guard; genuinely-non-blocking `publish()`) are being added as follow-up commits.
- The 2001-line `mod.rs` split is real but mesh-llm-internal maintainability beyond this capsule-integration PR — happy to file it as its own refactor issue rather than balloon this PR.

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle visibility for OpenAI requests, including effective requests, terminal outcomes, cancellations, and response statuses.
  * Added capsule markers exposed through the `X-Capsule-Id` response header.
  * Added OpenAI exchange event broadcasting across supported request paths, including exchange IDs and nonce-source reporting.
* **Bug Fixes**
  * Improved plugin delivery so unavailable or busy plugins do not delay broadcasts.
* **Tests**
  * Expanded coverage for exchange outcomes, capsule headers, cancellation handling, and plugin delivery.
* **Documentation**
  * Added guidance for the OpenAI exchange lifecycle and plugin integration.


crates/mesh-llm-host-runtime/src/plugin/mod.rs
```diff
@@ -1192,6 +1193,42 @@ impl PluginManager {
...
+    pub async fn broadcast_channel_message(
+        &self,
+        channel: &str,
+        content_type: &str,
+        body: Vec<u8>,
+        correlation_id: &str,
+    ) -> Result<()> {
+        let mut results = Vec::new();
+        for (plugin_id, plugin) in &self.inner.plugins {
+            if !self.plugin_declares_mesh_channel(plugin_id, channel).await {
```
ndizazzo: This declaration check calls through `manifest()`, which starts a stopped plugin and waits for its connection/init timeouts. Raw ingress awaits this broadcast, so a lazy observer can block the user request before anything reaches the bounded queue. Please use non-starting manifest metadata here and only enqueue to plugins that are already running.

crates/openai-frontend/src/hooks.rs
```diff
@@ -140,11 +249,62 @@ impl OpenAiBackend for HookedOpenAiBackend {
...
+            Ok(response) => ChatCompletionOutcome::Success { response },
+            Err(error) => {
+                error_message = error.to_string();
+                ChatCompletionOutcome::Error {
+                    status: error.status().as_u16(),
+                    message: &error_message,
+                }
+            }
+        };
+        self.hooks
```
ndizazzo: The terminal hook only runs after the backend future returns. The outer backend timeout drops this future, and request cancellation does the same, so those exchanges never get a terminal event. This needs a cancellation/drop guard or another handoff that guarantees one terminal callback for every admitted exchange.

ndizazzo (changes requested):
Needs revision. There are two lifecycle contract blockers inline.

Follow-ups:
- Validate capsule IDs before terminal publication so the hook and client can't disagree about the acknowledgement.
- The required `capsule_marker` field is a source break for downstream Rust struct literals. This needs a compatible shape or an explicit version boundary.
- Current-head CI hasn't run because the fork approval gate is still waiting. This needs a green run before re-review.

i386 (approved):
Reviewed across all three layers (typed-frontend hook seam, raw-proxy ingress, plugin broadcast transport). Approving — this is a well-executed M1/M2 slice of #1331.

What I verified:

- **Observer safety.** `OpenAiExchangeHookBridge` only observes and mints markers; dispatch can't be short-circuited, and `broadcast_channel_message` failures are logged, never propagated into the request path. Terminal events fire on all three outcomes (success/error/denied), including the before-hook error path.
- **No backpressure leak into serving.** `try_send_channel_message` skips a full/not-running plugin instead of stalling the request, and the dedicated test pins that property (with a sanity check that the blocking sibling still blocks).
- **Broadcast correctness.** The aggregate function attempts every declaring plugin before summarizing failures — the old `?`-per-send pattern's silent-skip failure mode is called out and regression-tested.
- **Correlation.** One `exchange_id` minted at admission pairs effective/terminal events on both dispatch paths; mirrored into `correlation_id` on the wire.
- **Header plumbing.** `CapsuleMarkerExtension` threads the marker to `frontend_lifecycle_middleware` via response extensions (same relay pattern as `TerminalUsage`), covering both chat and responses endpoints; router tests assert the `X-Capsule-Id` header.
- `observes_dispatched_request` defaulting to clone is the right safe default.

Two non-blocking notes:

1. The fallback nonce `format!("fallback-{response.id}")` is minted in the bridge while the doc references the `client_nonce_source` tri-state — worth exposing the source in the envelope when M3 consumes it, so a plugin can distinguish client-supplied from sidecar-minted without string sniffing.
2. Path 2 (raw proxy) publishes no request body — fine per the design note, but worth remembering when someone later asks a plugin to filter on content for plugin-served models.

crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
```diff
@@ -0,0 +1,453 @@
...
+        &self,
+        request: &ChatCompletionRequest,
+        response: &ChatCompletionResponse,
+    ) -> Option<CapsuleMarker> {
+        let nonce = request
+            .extra
+            .get("client_nonce")
+            .and_then(|value| value.as_str())
+            .map(str::to_string)
+            .unwrap_or_else(|| format!("fallback-{}", response.id));
```
i386: This fallback nonce is minted silently here while the doc comment above references capsule-emit-mesh's `client_nonce_source` tri-state (`client_supplied` / `sidecar_generated_fallback`). When M3 consumes these envelopes, a plugin will need to distinguish client-supplied from sidecar-minted nonces without string-sniffing the `fallback-` prefix — consider adding a `nonce_source` field to the envelope now (it's cheap while the channel is v1 and plugins are few), or at least note the prefix as a stable contract here.

StevenMih:
Rebased/updated: the exactly-once TerminalGuard fix now also covers the two lifecycle gaps raised in review — denied outcomes route through `guard.fire` (so a cancel mid-await still fires exactly one terminal via Drop), and the guard is armed before the pre-backend hooks rather than after. Two window-targeted cancellation tests added; `cargo test -p openai-frontend --lib` is green (199 passed). Ready for a CI run whenever a maintainer can approve it.

StevenMih:
@ndizazzo — a status ping to unstick this one.

This branch hasn't changed shape since your review, but it's now at HEAD `602d2f897`, and all six blockers you raised are closed (commit `fe6ea867f` "closes all six blockers" plus three follow-ups):

- terminal-on-cancellation/drop → `TerminalGuard` RAII + `ChatCompletionOutcome::Cancelled` (fires on `Drop`), enum marked `#[non_exhaustive]`;
- `manifest()` lazily starting a stopped plugin → replaced with a non-starting `manifest_snapshot`;
- capsule-id validation before the terminal publish → `capsule_id_is_valid()` gate;
- the `ChatCompletionResponse` source-break → now `#[non_exhaustive]` with a `from_parts()` constructor, and the skippy-server call sites migrated.

Your `CHANGES_REQUESTED` predates all of that, so I think the only thing between here and a re-review is a green CI run — and the fork-CI checks are still `action_required` (waiting on a maintainer to approve the run rather than a failure). Could a maintainer kick those off? Once they're green I'd appreciate a fresh look against current HEAD. Happy to walk through any of the fixes. Thanks!

StevenMih:
One meta-ask for the maintainers, since it's now gating this PR plus #1397, #1362, and #1: **fork PRs never run CI here** — the jobs fail at setup with `Error response from daemon: manifest unknown` and `Can't find 'action.yml' … capture-sccache-stats … Did you forget to run actions/checkout`, i.e. they die before compiling anything, so none of these PRs can show a green signal.

Why it matters concretely: on #1362 the `no-console-print` ratchet was regenerated on 8/26 (`ad5bb26`) and `cargo run -p xtask -- repo-consistency no-console-print` passes clean locally — but with no CI to show it, the review reasonably asked again to regenerate something that was already done. Same shape on the others.

Could a maintainer either enable CI on fork PRs or approve the pending workflow runs on these four? Happy to rebase any of them onto current `main` first if that helps the runners resolve the local composite actions.

StevenMih:
@ndizazzo Thanks for the thorough pass. Here's where each of the three blockers now stands, checklist-style, on the current head (`d5e01cd3e`):

## Blocker 1 — capsule IDs validated before terminal publication

- [x] **Done.** The terminal-publish path validates the capsule id before the marker is ever attached to the response (and therefore before it can ride on the terminal event a plugin observes), so the hook and the client can never disagree about what was acknowledged.
  - `crates/openai-frontend/src/hooks.rs`, in `HookedOpenAiBackend::chat_completion_with_context`: the minted marker is only attached when `capsule_id_is_valid(&marker.capsule_id)` holds; otherwise it's dropped with a `tracing::warn!` and the terminal reports no marker. `capsule_id_is_valid` (`crates/openai-frontend/src/chat.rs`) rejects both empty ids and any string that can't become an `X-Capsule-Id` `HeaderValue`.
  - The inlined skippy backend lifecycle (`crates/skippy-server/src/frontend/backend.rs`, `chat_completion_with_hooks`) applies the identical guard, so both dispatch paths agree.
  - Introduced in `fe6ea867f` ("address #1437 review blockers … capsule id …").

## Blocker 2 — `capsule_marker` as `Option`

- [x] **Already in place** — no source break. `crates/openai-frontend/src/chat.rs:292`:
  ```rust
  #[serde(skip)]
  pub capsule_marker: Option<CapsuleMarker>,
  ```
  It's `#[serde(skip)]`, so it never touches the OpenAI-shaped JSON body — the real `X-Capsule-Id` is emitted as an HTTP response header by the router's `frontend_lifecycle_middleware`, the same layer that already sets `x-request-id`. The field is a compatible additive shape, not a required-break.

## Blocker 3 — green CI

- [x] **Branch is now conflict-free.** `main` had moved (backend.rs / parsing.rs / guardrails.rs / backend tests + a couple of struct refactors — `iteration_scheduler`, `PluginRuntime`); I resolved it with a merge commit (`9a2295160`, `--signoff`), **not** a rebase, so i386's approval stays anchored to the reviewed commits. `cargo check --tests` is clean for `openai-frontend`, `skippy-server`, and `mesh-llm-host-runtime`, and the full `openai-frontend` suite is green (215 tests, incl. all 27 relocated `hooks::tests::*`).
- [ ] **Needs your click.** As a fork PR, the workflows won't run until a maintainer approves the run. Now that the branch is conflict-free, one "Approve and run" on the pending workflows should light CI up green.

## Also addressed in this push (CodeRabbit)

- **A — file-size rule:** `crates/openai-frontend/src/hooks.rs` was 2070 lines (over the 2000 rule). Split the `#[cfg(test)] mod tests` out to a sibling `hooks_tests.rs` via `#[path]` (matching the `router_tests.rs` / `guardrails` convention). Pure move — hooks.rs is now 855 lines, tests unchanged and green.
- **B — `observes_dispatched_request()` in the skippy wrapper:** the inlined lifecycle in `skippy-server` `chat_completion_with_hooks` was cloning the effective request unconditionally on the exact path the flag exists to protect. It's now gated on `observes_dispatched_request()` (default/empty request when false), mirroring `HookedOpenAiBackend`. `MeshAutoHookPolicy` returns `false`, so this removes a per-completion full-request clone (message content + inline media) on the mesh auto-route path.

crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
i386: Verified on current head d5e01cd3e: `OpenAiExchangeEnvelope` now carries serialized `nonce_source: Option<ClientNonceSource>` with `client_supplied` and `sidecar_generated_fallback` values, so consumers do not need to inspect the fallback prefix. Both source branches pass in the `openai_exchange` regression set (6/6).

crates/openai-frontend/src/hooks.rs
```diff
@@ -735,4 +1258,813 @@ mod tests {
...
+        assert_eq!(error.status().as_u16(), 502);
+
+        let terminals = policy.terminals.lock().unwrap();
+        assert_eq!(terminals.len(), 1);
+        assert!(matches!(
+            &terminals[0],
+            TerminalRecord::Error { status: 502, message }
+                if message.contains("upstream exploded")
+        ));
+    }
```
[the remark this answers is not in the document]
i386: Verified on current head d5e01cd3e: production `hooks.rs` is 855 lines and declares the sibling `hooks_tests.rs` through `#[path = "hooks_tests.rs"]`; the extracted test module is 1,215 lines. The full `openai-frontend` suite passes (215 tests including integration tests).

i386 (approved):
Reviewed current head 456ce3ada102cf524e61b78cfbb91ee5d5b60f69 after resolving the backend conflict against current main. The resolution preserves the hook lifecycle and main's adaptive admission/telemetry and offloaded prompt preparation. Validation: cargo fmt; openai-frontend (215 tests including integrations); skippy-server (658 passed, 3 ignored); mesh-llm-host-runtime (2,910 passed, 8 ignored). GitHub reports the branch mergeable.

ndizazzo (approved):

Issue #1425, opened 2026-08-24
Publish a self-contained Nemotron 3 Super Q4 + MTPv2 + suffix N-gram layer package


i386:
## Goal

Build and publish a self-contained Skippy layer package combining:

- the existing Nemotron 3 Super `UD-Q4_K_XL` target;
- NVIDIA's separately released BF16 MTPv2 head;
- Daniel Winter-Wijntjes' request-local exact-suffix N-gram proposer; and
- package-owned native-MTP, suffix N-gram, and composite MTP+suffix generation strategies.

The intended product strategy is `mtp-suffix`: native MTP predicts the uncertain prefix, suffix N-gram lookup extends it from earlier request-local prompt/tool history, and the target verifies the combined candidate. The package should download, plan, split, and serve like any other Mesh layer package, without an external `draft_model_path` and without requantizing the complete 120B target.

## Background

The current target is:

- source GGUF: `unsloth/NVIDIA-Nemotron-3-Super-120B-A12B-GGUF`;
- quant: `UD-Q4_K_XL`;
- Mesh package: `meshllm/NVIDIA-Nemotron-3-Super-120B-A12B-UD-Q4_K_XL-layers`;
- target size: approximately 84.49 GB; and
- trunk layers: 88.

Inspection of the published GGUF shards confirms that they contain only `blk.0` through `blk.87`: there are no MTP/NextN tensors or `num_nextn_predict_layers` metadata. The current package consequently reports `layer_count: 88` and `generation: null`.

NVIDIA now publishes an updated MTP head separately:

- `nvidia/Nemotron-3-Super-120B-A12B-BF16-MTPv2`;
- approximately 3B parameters / 5.88 GB BF16;
- `num_nextn_predict_layers = 1`; and
- intended to be attached to a compatible Nemotron 3 Super target rather than served as a standalone model.

NVIDIA reports mean accepted-token lengths at draft length 7 of 4.86 for coding and 4.31 overall, compared with 3.78 and 3.45 for the original built-in MTP head. These are acceptance lengths, not direct end-to-end speedup figures, and must be remeasured on the Q4 target and Apple Silicon lab.

## Why MTP plus suffix N-gram

The suffix N-gram proposer contributed in #1037 is aimed directly at coding-agent and tool-loop traffic:

- file re-emission with localized edits;
- repeated structured configuration;
- repeated identifiers;
- echoed tool output; and
- long verbatim prompt regions.

[...]

PR #1439, opened 2026-08-25
skippy-quantize: compose-mtp — splice an MTP draft into a sharded target GGUF

i386:
## What you can do now

`skippy-quantize compose-mtp` can build the self-contained composite GGUF issue #1425 asks for: a quantized sharded target (e.g. Nemotron 3 Super UD-Q4_K_XL) plus an MTP/NextN draft head folded in as the final block — no native ABI change, no re-quantization, target bytes preserved verbatim. `scripts/hf-skippy-mtp-compose-job.py` runs the whole package build (convert head → compose → attach probe → stream-publish) as an HF Job.

## How it works

- Rewrites the last shard's header/tensor table and stream-copies its data; GGUF tensor offsets are data-section-relative, so re-serialization preserves them. MTP tensors append at aligned offsets as `blk.N` with the renamed nextn/per-block names.
- `--metadata-shard/--metadata-output`: llama.cpp splits keep the global KV (block_count, tokenizer) only in shard 1 — verified against the unsloth UD-Q4_K_XL target where shards 2–3 carry only `split.*` KVs. The bump (88→89) and `--set-kv` overrides (`nemotron_h_moe.nextn_predict_layers=1`) are applied there, never double-applied.
- Drafts already positioned at the target block are accepted verbatim — the pinned llama.cpp converter's `--mtp` export for the Nemotron MTPv2 head emits exactly that (verified in `.deps/llama.cpp/conversion/nemotron.py`).
- `GgufKv::Raw` preserves metadata value types without typed variants losslessly.

## Validation

- `cargo test -p skippy-quantize` — 162 passed (5 new: round-trip bytes, KV overrides, rename refusal/acceptance, sharded metadata patch).
- `cargo clippy -p skippy-quantize --all-targets -- -D warnings` clean.
- Target/draft tensor names cross-checked against the actual Hub artifacts (unsloth GGUF shard headers, NVIDIA MTPv2 safetensors index) and the pinned converter mapping.

## Summary by CodeRabbit

* **New Features**
  * Added a unified radix-based cache for resident and recurrent state with prefix matching, deduplication, reference tracking, and LRU eviction.
  * Added MTP-to-GGUF composition tooling, composite KV pages, K-index metadata, and mixed iteration-batch execution.
  * Added certification batteries, automated upstream canary repair, and reproducible cache A/B benchmarks.

* **Bug Fixes**
  * Prevented partial session-state updates when token-count validation fails.
  * Improved MTP composition reliability by staging required tokenizer files.

* **Documentation**
  * Added unified cache and canary repair guidance.


crates/skippy-quantize/src/compose_mtp.rs
```diff
@@ -0,0 +1,925 @@
...
+            block_count = patched;
+            Some(output.display().to_string())
+        }
+        _ => None,
+    };
+    // block_count lives in the metadata shard when one was provided; applying
+    // the plan to the last shard too would double-bump it, and split shards
+    // carry only split.* KV anyway.
+    if metadata_shard.is_none() {
+        block_count = plan_metadata(&mut target.kv, &kv_overrides, args.no_bump_block_count);
```
ndizazzo: Sharded GGUFs store the global tensor total in `split.tensors.count`. This path appends the MTP tensors but only applies the regular metadata plan, so that count stays at the target's old value. llama.cpp then rejects the composite because the metadata no longer matches the loaded tensor names. Please update the split tensor count on the relevant shards and cover it with a real sharded fixture.

ndizazzo (changes requested):
Needs revision. The sharded output metadata has a loader-blocking mismatch called out inline.

Follow-ups:
- Add a sharded fixture that contains `split.tensors.count`, then load the composed output with llama.cpp.
- Compose MTP isn't exercised by current CI or the HF job tests.
- The current Linux run also has a deterministic skippy-server test failure. Please rerun the full matrix after the metadata fix.

i386:
Review fixes pushed in 61799e48 (+ job logging in dacfbdca):

- `split.tensors.count` is now bumped by the appended MTP tensor count when patching the metadata shard (and in the target KV when no metadata shard is given). Covered in `patches_metadata_shard_for_sharded_targets`, whose fixture now carries `split.tensors.count` and asserts the bumped total. The composed output is exercised end-to-end by the job's `validate-mtp-attach` probe (full llama.cpp-side attach path); happy to add an in-repo loader fixture on top if you want the check in CI too.
- `GgufKv::Raw` arrays now preserve the GGUF array header (element type + element count) inside the raw bytes so the round-trip no longer shifts subsequent KV entries (`raw_array_metadata_round_trips_with_array_header`).
- `--set-kv` against an existing non-u32 key now errors instead of silently no-oping (`set_kv_override_errors_on_non_u32_existing_key`).

i386:
## Local repro complete — root causes 1 & 2 fixed, full pipeline verified

The failed job 7 path is now reproduced and passing **locally** against the real checkpoint (nvidia `Nemotron-3-Super-120B-A12B-BF16-MTPv2` draft + unsloth `UD-Q4_K_XL` trunk shards), no remote jobs spent.

**Root cause 1 (draft load):** the upstream MTPv2 repo carries no `token_embd`/`lm_head` (1040 keys, only `mtp.layers.*`); the nemotron loaders created `token_embd`/`output_norm` as **required** tensors for the draft load → `check_tensor_dims: token_embd.weight not found`. Fix (patch 0028): mark shared trunk tensors `TENSOR_NOT_REQUIRED` when `mtp_only`, skip the duplicate-embedding fallback; the MTP graph borrows trunk `tok_embd`/`output` via `ctx_other`.

**Root cause 2 (draft context, found by the local repro):** the `llama_context` constructor gates `ctx_other` propagation by arch (GEMMA4_ASSISTANT, EAGLE3/DFLASH, INKLING) — no `NEMOTRON_H_MOE` clause. The attach path set `ctx_other`, the constructor silently dropped it, and `graph_mtp` aborted at reserve time: `nemotron-h-moe.cpp:31 GGML_ASSERT(tok_embd_w != nullptr ...)`. Fix (patch 0029): add the missing clause, gated on `LLAMA_CONTEXT_TYPE_MTP` like INKLING.

**Verified end-to-end on this machine** (patched queue through 0029, cpu static ABI build):
```
convert --mtp → compose-mtp (--mtp-block 88, 89+1 layers) → validate-mtp-attach
✅ MTP attach valid: parts=3 layers=89 mtp_layers=1 projector=none session_created=true
```
`scripts/tests`: 549 passed, 7 skipped. Commits: 0028 + 0029 on this branch (`2e949a20`).

crates/skippy-quantize/src/compose_mtp.rs
i386: All points from this review are addressed on the current head — the review predates the fix series (landed 2026-08-25/26). Mapping:

**`split.tensors.count` stale (the loader blocker):** fixed in 61799e48 — `bump_split_tensors_count` runs on both the per-shard KV and the metadata plan, bumping by the number of appended MTP tensors. ade86316 keeps the bump in the key's original scalar type. Unit-covered in `compose_mtp.rs` (synthetic sharded fixture with `split.tensors.count` present; asserts 1 → 2 after appending).

**Loading the composed output with llama.cpp:** exercised on a real sharded composite, not just unit fixtures — the published Nemotron MTPv2 composite (3 shards, 89+1 layers) passes `validate-mtp-attach` (parts=3, layers=89, mtp_layers=1, session_created=true), which loads the composite and attaches the draft via the patched llama.cpp. Same pipeline verified locally end-to-end before publishing.

**Raw-array / per-layer metadata:** preserved and extended correctly as of ba54c1c8 + cb5da87f (array headers kept, per-layer arrays extended for the appended layers, non-u32 overrides now error instead of silently dropping).

**Deterministic Linux skippy-server failure:** that was the #1452 KV-admission break on main (429 with `capacity=256, minimum_free=512` on the `--ctx-size 256` fixtures) — fixed on main by #1459. Current Linux lane on 6cae7acb is green; full matrix rerun is on this head (Quality/macOS/Linux/Website already SUCCESS, Windows in progress).

**Compose-mtp in repo CI:** agreed it is not exercised by the CI lanes — today it is exercised by the HF job on real weights. A small in-repo fixture test (tiny sharded GGUF → compose-mtp → load via the prepare-llama build) is a reasonable fast-follow; happy to put it up after this merges.

i386: Follow-up filed as #1461 (in-repo compose-mtp fixture compose + llama.cpp load, and contract tests for the HF job script) — kept out of this PR to hold its scope.

PR #1444, opened 2026-08-25
feat: Apple Core AI provider

i386:
## What this adds

The **Apple Core AI provider** — run inference through Apple's on-device foundation models as a first-class mesh provider on macOS Apple Silicon (experimental).

**Capabilities**
- **Apple system-model inference** — an experimental provider backed by Apple's on-device foundation models, exposed to the mesh like any other backend.
- **Provider-host lifecycle APIs** across every SDK — Rust, Swift, Node.js, and Kotlin — to start, supervise, and stop a provider host (`start_provider_host`, `ProviderHostHandle`, `ProviderRuntimeOptionsNative`).
- **Packaged provider runtimes** — discovery, manifest validation, caching, downloading, and secure (signed) installation of provider executables, with a `providers/apple` package + notarization-ready packaging script.
- **OpenAI-compatible loopback serving** — streaming, structured output, tool calls, and status reporting through the standard OpenAI surface.
- **Provider-aware mesh routing** — load-based target selection, capacity metrics, failover, and unhealthy-route withdrawal; explicit-only provider models are excluded from auto/degrade candidate sets so they're never picked implicitly.
- **Network UI** — Apple model recognition and provider load reporting.

## Which PRs this replaces

Collapses the stacked Apple Core AI patch queue into this single PR against current `main`:
`#1249 → #1250 → #1252 → #1255 → #1256 → #1259 → #1260 → #1261 → #1281`

Those PRs were stacked on a base that `main` has since rewritten (Nick flagged #1252 as unmergeable — 16 conflicting paths). This branch reconciles the full stack onto current `main` as one reviewable change and closes the originals.

Closes #1249, #1250, #1252, #1255, #1256, #1259, #1260, #1261, #1281

## Rebase reconciliation onto the rewritten base

- **Affinity:** `main` relocated target selection into the shared `mesh-llm-routing` crate. Added `affinity_selected` to the shared `TargetSelection` and set it in `select_model_target_from_keys`. Provider-preference routing reads it: a plain load-balanced pick may be redirected to a provider host, but an affinity/sticky decision stays authoritative.
- **Ingress / transport:** reapplied provider routing onto `main`'s refactor — `route_model_request` now lives in `transport_route_model.rs`, and the MoA degrade path moved to `context_selection::select_degrade_model`.
- **FFI:** ported the provider-host bindings into `main`'s modular `mesh-llm-ffi` layout and regenerated the Swift bindings from the UDL.
[...]

PR #1446, opened 2026-08-25
ci: expose macos27 Apple runner policy

i386:
## Summary

- Expose an Apple-capable `runner_macos_apple` output from the centralized runner policy.
- Map that semantic output to the newly available `macos27` runner label.
- Add selector regression coverage across hosted and Depot policy branches.
- Route no existing CI row in this prerequisite.

## Why this is separate from #1444

PR workflows intentionally load `select-ci-runners` from protected `main`. The Apple provider row on #1444 therefore cannot consume a new runner-policy output introduced only by its own branch.

A bounded branch dispatch demonstrated that boundary: the lane loaded the current `main` selector, received no `runner_macos_apple` output, and failed before creating the Apple matrix job. Once this prerequisite lands, #1444 can consume the trusted output and run its dedicated `macos-apple` row on `macos27` without adding a raw-label fallback.

Evidence: https://github.com/Mesh-LLM/mesh-llm/actions/runs/32908562559

## Validation

- `actionlint -config-file .github/actionlint.yaml`
- `python3 -m unittest scripts.tests.test_ci_artifact_actions` — 45 passed
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting the `xcode-27` GitHub-hosted runner label.
  * Exposed the selected Apple macOS runner label for supported workflows.
  * Clarified that `xcode-27` provides a macOS 26.5 host with the macOS 27 SDK.

* **Tests**
  * Updated validation for the new runner-selection output.

* **Documentation**
  * Documented runner capabilities and platform/toolchain requirements.

### DIFF
--- a/.agents/skills/manage-ci/references/current-inventory.md
+++ b/.agents/skills/manage-ci/references/current-inventory.md
@@ -418,9 +418,16 @@ permission.
 ## Providers and variables
 
 GitHub-hosted labels are `ubuntu-24.04`, `ubuntu-24.04-arm`, `macos-15`, and
-`windows-2022`. Central policy also exposes the Xcode 27-capable `macos27`
+`windows-2022`. Central policy also exposes the GitHub-hosted `xcode-27`
 label as `runner_macos_apple` for the follow-on Apple-provider platform row;
-this prerequisite does not route an existing row to it. Depot labels are
+this prerequisite does not route an existing row to it. That image is an
+Xcode 27 toolchain lane: today it runs a macOS 26.5 host with the macOS 27
+SDK (actions/runner-images#14404), so the output names the Xcode capability,
+not a macOS 27 host OS. Rows that need an actual macOS 27 host must wait for
+a host image (GitHub or Depot) that provides one — Depot's current
+`depot-macos-26` tops out at macOS 26/Xcode 26.6. #1444 must assert
+`sw_vers`, `xcodebuild -version`, and the selected SDK before doing work, and
+decide compile-only vs test accordingly. Depot labels are
 selected only by `select-ci-runners`; no workflow accepts a raw provider label.
 Trusted main Linux requires
 `DEPOT_RUNNERS_ENABLED=true`. An exact same-repository PR revision may use the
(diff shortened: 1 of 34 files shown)


ndizazzo (changes requested):
Requesting changes. The semantic runner output is the right shape for the protected-main workflow boundary, but the current head cannot be merged.

1. Commit `3facd3085` added about 990 unrelated docstring lines and changed 13 scripts from executable to non-executable. The Quality job now fails when `restore-native-sdk-input.sh` invokes `scripts/safe-extract-tar.py` and receives "Permission denied": https://github.com/Mesh-LLM/mesh-llm/actions/runs/32911811162/job/98007648490. Please revert the auto-fix commit rather than retain the unrelated edits.

2. `runner_macos_apple` emits `macos27`, but GitHub's hosted Xcode 27 labels are `xcode-27` and `xcode-27-xlarge`: https://github.com/actions/runner-images/issues/14404. Unless `macos27` is a separately provisioned organization runner, this leaves the downstream job waiting for a label that no runner owns. The inventory calls this GitHub-hosted, so the checked-in value and test expectation need correction. If it is an organization alias, document and verify that external runner contract.

3. Correcting the label alone does not establish the capability promised by the output. GitHub's `xcode-27` image currently runs macOS 26.5.2 with the macOS 27 SDK, while PR #1444 declares `.macOS(.v27)` and runs `swift test`. Please decide whether this is a compile-only Xcode 27 lane or a test lane that requires an actual macOS 27 host. Split those roles if necessary, and assert `sw_vers`, `xcodebuild -version`, and the selected SDK before doing work.

Depot does have M4 macOS runners, but its current `depot-macos-26` image provides macOS 26 and Xcode through 26.6, so it is not a drop-in replacement for this row today: https://depot.dev/docs/github-actions/runner-types. Keep the Apple role semantic so it can select Depot when Depot offers the required OS and toolchain. If the provider can target macOS 26, lower that requirement explicitly and route the role through the existing Depot trust and cache gates.

After fixing the patch, run the repository's required `just ci-validate` contract. The narrow artifact-action test listed in the PR did not catch the executable-bit regression.

PR #1448, opened 2026-08-26
fix(proxy): cancel upstream before response headers

### LINKED ISSUE #1406: Proxy: a client that disconnects before the first upstream response byte is never detected
## Summary

Both API-proxy route arms fully `await` the upstream HTTP response probe before they ever look at downstream state. If the client disconnects **before the first upstream response byte arrives**, nothing detects it and nothing cancels the upstream.

This is the pre-TTFT sibling of the defect fixed in #1405 (RV-DEFECT-003). #1405 closes the window *after* the response head arrives; this issue is the window *before* it.

## Where

`crates/mesh-llm-host-runtime/src/network/openai/response/routing.rs`

- Local arm: `route_local_attempt_after_forward` — `probe_http_response_local(upstream).await` is awaited to completion; `tcp_stream` is untouched until `relay_attempted_response`.
- Remote arm: `route_remote_attempt_after_forward` — same shape with `probe_http_response(quic_recv).await`.

`probe_http_response_with_timeout` (`response/probe.rs`) reads only from the upstream reader. There is no downstream watcher anywhere in that path.

## Exposure (bounded, but not trivially)

Exposure ends the moment upstream produces its first response byte, so this is **not** the unbounded full-generation leak that RV-DEFECT-003 was. Concretely:

- Remote first-byte timeout is 5 min (`response_first_byte_timeout`), local is 10 min (`local_response_first_byte_timeout`).
- On the remote arm a first-byte timeout yields `RetryableTimeout`, which `should_retry_uncommitted_remote_attempt` retries once (`REMOTE_UNCOMMITTED_RETRIES = 1`) on a **fresh tunnel**. So a departed client can drive two full prefill attempts, worst case ~10 min of peer GPU time.
- Secondary defect in the same function: only the **first** read is timeout-guarded. Continuation reads go through `read_response_chunk`, documented "without any timeout." An upstream that sends a partial header and then stalls blocks the arm **forever**, with no downstream detection and no cap. `MAX_HEADER_BYTES` bounds size, not time.

## Why this is smaller than it looks

Two things that were open questions are already settled by #1405 and by the existing code:

1. **"What does cancelling mean per transport" is already solved.** #1405 centralised it behind the `CancelUpstream` trait (`response/cancellation.rs`): `shutdown()` for local TCP, QUIC `STOP_SENDING` for the peer stream. A probe-window fix reuses `cancel_upstream_if_client_disconnected` unchanged.
2. **The detector is one detector, not two.** It watches the *downstream* `TcpStream`, which is identical on both arms. There is no per-transport detector work.

Also: both upstreams are created fresh per attempt (`TcpStream::connect` in `acquire_local_attempt_upstream`, `node.open_http_tunnel` in `route_remote_attempt`). Neither is pooled, so cancelling mid-probe cannot poison a reused connection.

## Suggested approach

`tokio::select!` the probe future against a downstream-close watcher. The borrows are disjoint — the probe holds `&mut upstream`, the watcher only needs `&*tcp_stream` — so this needs no `Arc`, no `split()`, and no spawned task.

```rust
/// Resolves when the downstream client is gone. Never consumes request bytes.
async fn wait_for_downstream_close(stream: &TcpStream) {
    let mut byte = [0u8; 1];
    loop {
        match stream.peek(&mut byte).await {
            Ok(0) => return,                        // FIN
            Err(_) => return,                       // RST / socket error
            Ok(_) => std::future::pending().await,  // pipelined bytes: leave them, stop watching
        }
    }
}
```

`tokio::net::TcpStream::peek` takes `&self` and does not consume, which is what keeps a pipelined follow-up request intact.

On trigger, return `RouteAttemptResult::ClientDisconnected` and hand it to the existing `cancel_upstream_if_client_disconnected`. `ClientDisconnected` is already terminal in `should_retry_uncommitted_remote_attempt`, so this correctly stops the retry from re-firing.

## The one genuine design risk

**Read-side EOF is not the same as "the client is gone."** A client that does `shutdown(SHUT_WR)` after sending its request and then waits for the response is legal HTTP/1.1, and the watcher above would abort it. nginx accepts exactly this tradeoff (`proxy_ignore_client_abort off` is the default), and half-close-then-await is not a pattern any OpenAI-style client uses in practice — but it is a real behaviour change and it is why this is not a mechanical fix and did not belong in #1405.

Decide it explicitly before implementing. If we want to be conservative, the `Err(_)` (RST) arm alone is unambiguous and still catches the common Ctrl-C case on most clients; it just misses graceful closes.

## Test shape

Follow the pattern #1405's `1acd40c14` established: drive a real socket rather than asserting on an error *kind*. Hold the upstream silent (no header at all), close the client with `set_zero_linger()` to force an RST, and assert the arm returns `ClientDisconnected` and that the upstream saw its cancel — for both the local TCP and remote QUIC arms.

The no-timeout continuation-read gap should get its own test: send a partial header, stall, assert the arm does not hang.

## Not in scope for #1405

#1405 is a release-unblocking PR for four validated rc6 defects. This is a hot-path concurrency change on both transports with a user-visible behaviour decision attached; it gets its own PR and its own validation.

ndizazzo:
Fixes #1406.

## What changed

- Race local and remote upstream response probes against a non-consuming downstream socket watcher.
- Treat socket errors and resets as terminal client disconnects, canceling the active local TCP or remote QUIC upstream through the existing cancellation path.
- Preserve legal HTTP write-half-close behavior by leaving clean EOF pending, and preserve pipelined request bytes with peek.
- Bound the complete response-header and overflow-preview probe to one configured deadline, including continuation reads after partial data.

## Disconnect semantics

A downstream reset or socket error counts as a disconnect and cancels upstream work immediately. A clean FIN does not: HTTP clients may legally half-close their write side after sending a request, so clean EOF remains pending while the response probe runs. If the upstream also stalls, that case is bounded by the probe deadline (5 minutes for remote routes and 10 minutes for local routes).

## Validation

- `cargo test -p mesh-llm-host-runtime --lib`: 2666 passed, 8 ignored (initial change)
- Response probe tests: 6 passed
- Routing tests: 6 passed
- Concrete API proxy regression tests: 28 passed
- `cargo test -p mesh-llm-host-runtime probe_partial_headers_share_one_timeout_budget --lib`
- `cargo clippy -p mesh-llm-host-runtime --lib --no-deps -- -D warnings`
- `git diff --check`

The broader all-targets Clippy run is currently blocked by pre-existing Rust 1.98 `chunks_exact_to_as_chunks` warnings in `skippy-protocol` and `network/openai/transport_tests/lifecycle.rs`; neither is changed here.

The regression suite exercises real local sockets and the remote cancellation arm, including reset, pipelined-byte, half-close, partial-header timeout, and slow-drip deadline behavior. The original rc6 report has not been reproduced on a separate two-node deployment in this worktree.

## Summary by CodeRabbit

- **Bug Fixes**
  - Upstream response probing now respects the remaining configured timeout, preventing stalled responses from extending request wait times.
  - Disconnected client connections now cancel upstream processing promptly, improving responsiveness and reducing unnecessary work.
  - Pipelined downstream data is preserved while checking connection status.
  - Improved handling of partially received responses, stalled continuation reads, and write-half-closed connections.
  - Requests now respond more reliably when clients disconnect or upstream services stop sending data.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/response/probe.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/response/probe.rs
@@ -225,7 +225,8 @@ pub(in crate::network::openai::response) async fn probe_http_response_with_timeo
         let first_read = buffered.is_empty();
         if first_read {
             let mut chunk = [0u8; 8192];
-            let read_result = tokio::time::timeout(timeout, reader.read(&mut chunk))
+            let read_timeout = timeout.saturating_sub(started.elapsed());
+            let read_result = tokio::time::timeout(read_timeout, reader.read(&mut chunk))
                 .await
                 .map_err(|_| {
                     anyhow!(
@@ -238,7 +239,8 @@ pub(in crate::network::openai::response) async fn probe_http_response_with_timeo
             }
             buffered.extend_from_slice(&chunk[..read_result]);
         } else {
-            read_response_chunk_with_timeout(reader, &mut buffered, timeout).await?;
+            let read_timeout = timeout.saturating_sub(started.elapsed());
+            read_response_chunk_with_timeout(reader, &mut buffered, read_timeout).await?;
         }
         if buffered.len() > MAX_HEADER_BYTES {
             bail!("HTTP response headers exceed {MAX_HEADER_BYTES} bytes");
@@ -254,7 +256,8 @@ pub(in crate::network::openai::response) async fn probe_http_response_with_timeo
         0
     };
     while buffered.len() < parsed.header_end + preview_len {
-        read_response_chunk_with_timeout(reader, &mut buffered, timeout).await?;
+        let read_timeout = timeout.saturating_sub(started.elapsed());
+        read_response_chunk_with_timeout(reader, &mut buffered, read_timeout).await?;
     }
 
     let retryable_context_overflow = parsed.status_code == 400
@@ -369,6 +372,42 @@ mod tests {
 
         assert!(is_timeout_error(&error), "unexpected error: {error:#}");
         assert!(error.to_string().contains("continuation read timeout"));
+    }
+
+    #[tokio::test]
+    async fn probe_partial_headers_share_one_timeout_budget() {
+        let timeout_budget = Duration::from_millis(100);
+        let (mut writer, mut reader) = tokio::io::duplex(4096);
+        let writer_task = tokio::spawn(async move {
+            writer.write_all(b"H").await.unwrap();
+            for byte in b"TTP/1.1 200 OK\r\n" {
+                tokio::time::sleep(Duration::from_millis(20)).await;
+                if writer.write_all(&[*byte]).await.is_err() {
+                    return;
+                }
+            }
+            std::future::pending::<()>().await;
+        });
+
+        let started = Instant::now();
+        let result = tokio::time::timeout(
+            Duration::from_secs(1),
+            probe_http_response_with_timeout(&mut reader, timeout_budget),
+        )
+        .await
+        .expect("partial headers must time out before the outer guard");
+        let elapsed = started.elapsed();
+        writer_task.abort();
+
+        let error = match result {
+            Ok(_) => panic!("a partial header stream must time out"),
+            Err(error) => error,
+        };
+        assert!(is_timeout_error(&error), "unexpected error: {error:#}");
+        assert!(
+            elapsed < Duration::from_millis(300),
+            "partial bytes extended the overall header budget: {elapsed:?}"
+        );
     }
 
     #[tokio::test(start_paused = true)]
--- a/crates/mesh-llm-host-runtime/src/network/openai/response/probe.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/response/probe.rs
@@ -94,7 +94,7 @@ pub(in crate::network::openai::response) fn try_parse_response_headers(
 }
 
 /// Read the next chunk of HTTP response data without any timeout.
-/// Used for continuation reads after the first byte has already arrived.
+/// Used by relay paths whose own body limits provide the lifetime bound.
 pub(in crate::network::openai::response) async fn read_response_chunk<R: AsyncRead + Unpin>(
     reader: &mut R,
     buf: &mut Vec<u8>,
@@ -106,6 +106,16 @@ pub(in crate::network::openai::response) async fn read_response_chunk<R: AsyncRe
     }
     buf.extend_from_slice(&chunk[..read_result]);
     Ok(read_result)
+}
+
+async fn read_response_chunk_with_timeout<R: AsyncRead + Unpin>(
+    reader: &mut R,
+    buf: &mut Vec<u8>,
+    timeout: Duration,
+) -> Result<usize> {
+    tokio::time::timeout(timeout, read_response_chunk(reader, buf))
+        .await
+        .context("upstream response continuation read timeout")?
 }
 
 pub(in crate::network::openai::response) async fn read_transformed_response_body<
@@ -228,7 +238,7 @@ pub(in crate::network::openai::response) async fn probe_http_response_with_timeo
             }
             buffered.extend_from_slice(&chunk[..read_result]);
         } else {
-            read_response_chunk(reader, &mut buffered).await?;
+            read_response_chunk_with_timeout(reader, &mut buffered, timeout).await?;
         }
         if buffered.len() > MAX_HEADER_BYTES {
             bail!("HTTP response headers exceed {MAX_HEADER_BYTES} bytes");
@@ -244,7 +254,7 @@ pub(in crate::network::openai::response) async fn probe_http_response_with_timeo
         0
     };
     while buffered.len() < parsed.header_end + preview_len {
-        read_response_chunk(reader, &mut buffered).await?;
+        read_response_chunk_with_timeout(reader, &mut buffered, timeout).await?;
     }
 
     let retryable_context_overflow = parsed.status_code == 400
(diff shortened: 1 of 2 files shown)


ndizazzo:
@coderabbitai review

crates/mesh-llm-host-runtime/src/network/openai/response/routing.rs
```diff
@@ -104,6 +110,99 @@ async fn route_local_attempt_after_forward(
...
+    F: Future<Output = Result<ResponseProbe>>,
+{
+    tokio::pin!(probe);
+    tokio::select! {
+        biased;
+        downstream = wait_for_downstream_disconnect(tcp_stream) => {
+            match downstream {
+                DownstreamWatchResult::Disconnected => ProbeOutcome::ClientDisconnected,
+                DownstreamWatchResult::PipelinedBytes => ProbeOutcome::Response(probe.await),
+            }
```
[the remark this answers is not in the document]
ndizazzo: This is intentional and follows #1406. Once peek observes unread pipelined data, another non-consuming peek can only observe that same queued byte; it cannot expose a later FIN or RST until the byte is consumed. Consuming it would require a new connection-buffer ownership path, and this proxy has a documented single-request contract that closes the routed connection after the first response. The watcher therefore preserves the byte and stops monitoring, exactly as the issue specifies. The regression test verifies that the byte remains unread.

crates/mesh-llm-host-runtime/src/network/openai/response/routing.rs
```diff
@@ -104,6 +110,99 @@ async fn route_local_attempt_after_forward(
...
+/// readiness spin and lets the upstream response probe win for clients that
+/// legally half-close their write side after sending a request.
+async fn wait_for_downstream_disconnect(tcp_stream: &TcpStream) -> DownstreamWatchResult {
+    if tcp_stream.readable().await.is_err() {
+        return DownstreamWatchResult::Disconnected;
+    }
+
+    let mut peeked = [0u8; 1];
+    match tcp_stream.peek(&mut peeked).await {
+        Ok(0) => std::future::pending::<DownstreamWatchResult>().await,
```
i386, line 125: One subtlety worth pinning in a comment or test name: `Ok(0)` here is ambiguous between "clean FIN (client half-closed after sending the request — legal, keep waiting)" and "FIN after RST" — the reason treating EOF as pending is safe is that the reset case is surfaced by `readable()`/`peek()` returning `Err` on the next readiness event, so you only stay pending *forever* if the peer truly went quiet. The doc comment captures the intent; nice.

One edge I couldn't find covered: a client that half-closes (FIN, no RST) and then never gets a response because the upstream stalls forever. `wait_for_downstream_disconnect` stays pending, the probe's first-byte timeout (5 min remote / 10 min local) fires, and we return `ProbeOutcome::Response(Err(timeout))` → `retryable_route_result_from_error`. So the half-closed-then-stalled case is bounded by the probe timeout, not by the watcher. Fine as shipped — but if the intent is "a vanished client should cancel upstream within ~seconds", a graceful FIN without RST still waits for the probe timeout. Worth a line in the PR description clarifying which signals count as "client disconnected" (RST/error = yes; clean EOF = no).

— Jian Yang (agent review)

crates/mesh-llm-host-runtime/src/network/openai/response/probe.rs
```diff
@@ -108,6 +108,16 @@ pub(in crate::network::openai::response) async fn read_response_chunk<R: AsyncRe
     Ok(read_result)
 }
 
+async fn read_response_chunk_with_timeout<R: AsyncRead + Unpin>(
```
i386, line 111: Applying the *first-byte* timeout budget to each continuation read independently (rather than a deadline since `started`) means a slow-drip upstream can hold the probe open for `timeout × N` reads. In practice `response_first_byte_timeout()` is 5 min and `local_response_first_byte_timeout()` 10 min, so a headers phase can now take up to ~unbounded with repeated partial reads. Two cheap hardening options (either is fine):

1. Make the continuation timeout a **deadline**: `timeout.saturating_sub(started.elapsed())` per read (you already track `started`).
2. Keep per-read budgets but add an overall cap on the loop (e.g. bail once `buffered.len() > MAX_HEADER_BYTES`, which you do — but the *time* dimension is the unbounded one).

The new test (`probe_continuation_read_has_timeout_after_partial_headers`) proves a single stalled read honors the timeout; a drip-feed variant (1 byte per 100ms against a 10ms timeout) would document the intended semantics either way.

— Jian Yang (agent review)

i386 (commented):
Solid fix — reviewed as agent on James's behalf. The race shape is right: non-consuming watcher (`readable` + `peek`), `biased` select so a reset wins over a probe completing, pipelined bytes left for the normal path, and half-close handled as pending rather than disconnect. The test suite upgrade (real sockets, zero-linger resets, the removed `Notify` choreography) is a real improvement — the new tests prove liveness of the actual route arms, not just that a helper returned.

Two non-blocking notes inline: (1) clarify that clean-FIN-without-RST still waits on the probe timeout before cancellation, (2) per-read (not per-deadline) continuation timeouts can stretch the header phase unboundedly under a slow-drip upstream.

Questions before this goes green:
- Lanes currently show skipping — PR is still a draft, so presumably intentional; ready-for-review will light up Linux/Quality/Website/Windows/macOS. Confirm that's the plan.
- Has this been exercised against the rc6 scenario in the description (worker kept generating after `request_dropped`)? The duplex test covers the remote arm; an end-to-end two-node confirmation would close the loop on the observed behavior.

— Jian Yang (agent review)

michaelneale (approved):

crates/mesh-llm-host-runtime/src/network/openai/response/routing.rs
ndizazzo: Clarified this in the PR description under **Disconnect semantics**. Only reset/socket-error signals trigger immediate cancellation. Clean FIN remains pending because clients may legally half-close after sending the request; if the upstream then stalls, the existing response-probe deadline bounds the wait (5 minutes remote, 10 minutes local). The PR is now ready for review and the full CI lanes are running. The regression suite exercises real local sockets and the remote cancellation arm, but I have not claimed a separate two-node rc6 reproduction; that limitation is now explicit in the validation section.

crates/mesh-llm-host-runtime/src/network/openai/response/probe.rs
ndizazzo: Fixed in `93be96592`. The initial read, partial-header continuation reads, and overflow-preview reads now each use the remaining duration from one deadline (`timeout.saturating_sub(started.elapsed())`) rather than resetting the budget per read. Added `probe_partial_headers_share_one_timeout_budget`, which drip-feeds bytes every 20 ms against a 100 ms budget and verifies the probe still times out within the overall deadline. Probe tests, routing tests, 28 concrete proxy regressions, affected-crate lib Clippy, fmt, and diff checks pass.

PR #1454, opened 2026-08-26
feat(topology): add a performance-aware planner and simulator

### LINKED ISSUE #1001: Split planner weights layer allocation by hold-capacity (`vram_bytes`), not throughput — RAM-heavy nodes get the largest shard
## Summary

`skippy-topology`'s split planner weights per-node layer allocation by `vram_bytes`, which is (intentionally, per #219) a **blended weight-holding capacity**, not a measure of compute throughput. Because capacity and throughput are different quantities, a node with large system RAM and weak compute is allocated a **larger** share of the pipeline than a node with a fast GPU.

This is not a restatement of #219 — that issue established the number is intentional and fixed the display labels (PR #222). This is about a *consumer* of that number: using hold-capacity as an allocation weight.

## The premise (from #219, quoting @Bortlesboat)

> The number is intentional but the label is wrong. `vram_bytes` is computed as `gpu_vram + 0.75 * (system_ram - gpu_vram)`. llama.cpp can offload layers to system RAM, so the value represents **how large a model the node can actually serve** — not how much GPU memory you have.

Agreed, and that's the correct semantic for a *fit* question. `model_target_reconciliation_local_fit` (`crates/mesh-llm-host-runtime/src/runtime/mod.rs:4018-4026`) asks "can this node hold the model?" and compares against `vram_bytes` — correct usage.

## The problem

`crates/skippy-topology/src/lib.rs:937` asks a different question with the same number:

```rust
(((request.layers.len() as u128) * (node.vram_bytes as u128)) / (total_weight as u128))
```

Layers allocated ∝ hold-capacity. But layers allocated should be ∝ **how fast the node can execute them**. `NodeSpec` (`lib.rs:31-37`) carries only `node_id`, `cached_slice_bytes`, `vram_bytes` — there is no throughput term. `NodePlacementSignal` adds `rtt_ms`/`availability_score`/`missing_artifact_bytes`, but those are tie-break adjustments in `node_package_score` (`lib.rs:1158-1174`), never multipliers on the span.

Pipeline parallelism is serial per request, so an over-allocated slow node isn't a partial contributor — it's a stall on every token.

## Worked example (CPU-only node)

`crates/mesh-llm-system/src/hardware/mod.rs:191-195`:
```rust
fn apply_cpu_only_runtime_budget(survey: &mut HardwareSurvey, metrics: &[Metric], system_ram: u64) {
    if metrics.contains(&Metric::VramBytes) && system_ram > 0 {
        survey.vram_bytes = (system_ram as f64 * 0.75) as u64;
    }
}
```
and the `not(feature = "skippy-devices")` path at `mod.rs:537` uses `* 0.90`.

So a 128 GB CPU-only box advertises ~115 GB capacity and **outranks a 24 GB RTX 4090 by ~4.8x**, receiving ~4.8x the layers — to execute on CPU. There is no zero-capacity guard: `validate_request` (`lib.rs:1500-1528`) checks only `EmptyLayers`/`EmptyNodes` and contiguity, and `span = span.max(1)` (`lib.rs:941`) floors every node at >=1 layer.

The same directionality applies (more mildly) to any GPU node with large RAM offload, since `mod.rs:330` does `vram_bytes = total + (ram_offload as f64 * 0.90)`.

## The data already exists on the wire

`crates/mesh-llm-protocol/src/proto/node.rs:92-97` already carries `mem_bandwidth_gbps`, `compute_tflops_fp32`, `compute_tflops_fp16`; `node.rs:44` carries `advertised_model_throughput`. As far as I can trace, their only consumers are `/status` and gpu-bench display (`api/status.rs:258-344`, `runtime/mod.rs:5864-6012`) — none reach `plan_weighted_contiguous`.

Notably, **request routing already reasons about throughput** (`network/metrics.rs:479-489` `tps_for_model`, consumed at `network/openai/transport.rs:3468` and `ingress.rs:247`), while the split planner does not — `mesh-llm-routing` has no `skippy` dependency.

## One caveat on any fix

`crates/mesh-llm-gpu-bench/src/runner.rs:31-34` returns `None` when `gpu_count == 0`, so bandwidth/TFLOPs are never measured on precisely the CPU-only nodes this affects (base detection hardcodes `None` at `hardware/mod.rs:1125-1127`). A throughput term would be inert for those nodes until some CPU-side measurement (STREAM-style, or a conservative device-class default) exists. Measurement likely has to land before the weight.

A `min(vram_bytes, required_bytes)` clamp would separately address a node with far more capacity than the model needs still absorbing proportionally more layers.

## Possibly related

`ROADMAP.md:60-63` notes MoE sharding "most results show this doesn't perform as well as one would hope, more research is needed." Speculative, but capacity-weighted placement across heterogeneous nodes seems like it could contribute.

## Question

Is capacity-as-allocation-weight intended (e.g. deliberately co-locating layers where the weights fit, accepting the throughput cost), or is a throughput term a gap worth filing properly? Happy to work up a PR if the latter, but this looks like a design call that should be yours first.

Verified against `2c2e808` (2026-07-15). `skippy-topology` is a pure crate (serde-only, 33 unit tests); note `weighted_contiguous_plan_uses_node_vram_for_layer_spans` (`tests.rs:231-255`) pins the current behavior deliberately, which is partly why I'm asking rather than assuming.

i386:
## What

Hardens and completes the existing performance-aware topology-planner foundation. This keeps the production coordinator planner, its exact capacity fallback, the directed-edge plumbing, observed timing feedback, and the simulator; it does not introduce another planner.

Tracking issue: #1455

Design: `docs/design/PERFORMANCE_AWARE_TOPOLOGY_PLANNER.md`

## Current scope

- Plumbs local and gossiped hardware/link signals into the existing automatic planner.
- Uses deterministic integer-microsecond cost math and contiguous-span DP.
- Models directed RTT/bandwidth edges and passive large-frame observations.
- Feeds fresh steady-decode stage observations back as a floor on analytical estimates.
- Adds a scenario-driven planner/execution simulator and calibration corpus.
- Preserves the exact capacity-only placement path whenever performance evidence is incomplete or the experimental mode is disabled.

## Safety hardening after deep review

- Rebases the PR onto current `main`, preserving the newer cache-affinity and activation-boundary work.
- Makes performance-aware placement explicit opt-in (`MESH_TOPOLOGY_PERF_AWARE=1/true/on/yes`); unset or unknown values remain capacity-only.
- Replaces missing-as-success/missing-as-zero ranking with tri-state estimate completeness. Network-only timing can prove an SLO miss, never success.
- Makes span DP and candidate evaluation optimize the same single-stream serial objective; streamed MoE weight fraction is used consistently.
- Prices the current raw-f32 activation payload instead of the stale f16 assumption.
- Bounds peer GPU metric string length, cardinality, and per-device plausibility before placement consumes it.
- Requires fresh multi-sample stage timing, normalizes batched decode by executed tokens, resets on stage-range changes, evicts stale/unbounded history, and quantizes replan signatures.
- Rejects unknown fields at every simulator-schema level.
- Repairs stale cross-package gossip fixtures exposed by the rebase.

## Deliberate limits / follow-ups

This PR remains a hardened foundation, not the end-state topology optimizer. Default-on promotion is blocked on the follow-up work tracked in #1455:

1. One typed, confidence-aware `PlanEstimate` shared by optimizer, simulator, diagnostics, and replan adoption.
2. Family-legal cut search, node-order search, boundary-specific sideband payloads, and real direct/fallback return paths.
3. Shape-aware prefill/decode measurement windows and held-out Apple/CUDA/ROCm plus LAN/WAN validation.
4. Cost-based adoption hysteresis, minimum dwell, migration/cache-loss budgets, warm cutover, and rollback gates.

## Local verification

- `skippy-coordinator`: 55 passed.
- `skippy-topology-sim`: 13 passed across unit, calibration, and corpus suites.
- `skippy-server --no-default-features --lib`: 669 passed, 3 ignored.
- `mesh-llm-host-runtime`: 2,950 passed across unit/integration suites, 9 ignored.
- Formatting, clippy (`-D warnings`), console-print ratchet, and repository consistency checks are run before branch update.

Closes #1001


i386:
Review fixes pushed as 2c53efe3:

**Fixed (code):**
1. **Inverted µs conversion** (`modeled_stage_time_us`, `hop_transfer_us`) — confirmed real. `bytes × 1_048_576 / (MiB/s × 1_000_000)` overestimated every modeled time by ~10%; correct is `bytes × 1_000_000 / (MiB/s × 1_048_576)`. Placement decisions were unaffected (uniform scale), but absolute modeled TPOT was wrong — this mattered before the execution-sim calibration lands, good catch.
2. **Non-conservative synthesized edge RTT** — `participant_edges` took `min` of the two participants coordinator RTTs; now `max`, so a synthesized pair edge can never underestimate a real hop. Doc comment rewritten to match.
3. **Silent link drops in the sim schema** — confirmed, including in the in-file test constant (exactly the bug class you flagged). `Scenario::from_toml` now rejects unknown top-level keys with a pointed error message; regression test added; test constant fixed.

**Fixed (doc):** DP complexity corrected to `O(layers² × nodes)` with the recurrence note; as-built cost-model section now defines canonical units (MiB/s everywhere, integer µs), the all-or-nothing missing-signal fallback, and flags metric-age decay as designed-not-implemented.

**Declined for now (doc-only, tracked in #1455):** objective-specific scoring beyond decode TPOT (interactive TTFT / throughput-at-concurrency equations) — the current planner optimizes decode TPOT only; widening the objective is phase 3+ work that should follow execution-sim calibration rather than precede it.

i386:
Round-2 review fixes pushed as 3868fdad — all three findings addressed:

1. **Malformed link keys now rejected at parse time.** `parse_link_key` is fallible: exactly one `->` separating two non-empty single-word node ids. `[links."alpha"]`, `alpha -> beta -> gamma`, and blank-side keys all fail with `MalformedLinkKey` before planning. Regression test covers four shapes (sim 11/11).
2. **Doc TPOT equation aligned with implementation.** As-built formula documented as `max_i(stage_time) + Σ_hops(edge_time)` — bottleneck stage plus total network time incl. prediction return — with an explicit note that it differs from the design-time per-stage-max for >2 stages and that the implemented form is authoritative.
3. **Fallback text corrected.** Missing node signals (capacity-greedy spans) is now distinct from absent/disabled edge data (legacy `hop_count × max-RTT`) — edge-aware estimation applies whenever edges exist, which the previous wording conflated.

michaelneale:
🤖 Review of head c18da85f (requested in Buzz #skippy-topology). Three read-only passes: planner core, host-runtime plumbing, sim + design doc. Headline claims spot-verified in source.

## CI first — the macOS red is real, not a cancellation

`PR · macOS` (run 33001333142) fails deterministically: I reran the failed jobs and both smoke lanes (Metal model-load, Swift SDK) failed again identically — the first `/v1/chat/completions` after model load returns **HTTP 429**. This PR doesn't touch admission; main had the same failure from #1452's capacity-aware admission and was fixed by #1459 (merged 21:03Z). This branch's merge-base (75b8866c) predates that fix. **Remedy: rebase onto main and rerun.** Linux/Windows reds are genuinely runner cancellations.

## Blockers / should-fix before merge

1. **The "bit-identical fallback" safety claim is false at fleet scope.** The fallback is per-*subset* (`topology.rs:870-875`), not per-fleet. In fleet `{A(bw), B(bw), C(no bw)}`, subset `{A,B}` gets DP spans plus `modeled_decode_tpot_us` and then competes under the *new* ordering rules — the final plan can differ from the old planner even though a node lacks signals. Separately, non-empty `edges` change the network estimate (and therefore winner selection) even for capacity-greedy candidates. The parity unit test uses exactly 2 nodes with one viable subset, so both holes are invisible to tests. The design doc scopes the claim correctly ("all-or-nothing **per candidate**"); the PR description and the phase-1 safety story don't.

2. **`split_participant_set_hash` is out of sync with `split_participant_signature`** (`local_package.rs:868-881`): the signature tuple grew to 10 fields but the hash covers only indices 0–7. The two new sustained-perf fields are silently omitted from claim identity, while the far noisier passive edge-bandwidth sample (index 5) *is* hashed. That's backwards — and it breaks the design doc's claim that "a measured change re-triggers planning" for node perf signals. Looks accidental; a test pinning hash coverage would have caught it.

3. **Unflagged ordering behavior change**: `latency_candidate_ordering` compares modeled TPOT *before* `target_met` and context length (`topology.rs:626-634`). With full signals, a plan that misses the TPOT target can now beat one that meets it, and a smaller context can beat a larger one. Maybe intended — but it's a placement behavior change with no test and no mention.

4. **`decode_tpot_target_met` ignores the model it sits next to** (`topology.rs:503-506`): the perf branch computes `modeled_decode_tpot_us` but evaluates the target against the network-only estimate, so a plan can report `target_met: Some(true)` while its own modeled decode time grossly exceeds the target.

## Design concerns (fine to defer, but say so)

5. **The planner's own TPOT is uncalibrated and misses the dominant cost term.** The sim calibration had to back-solve `per_hop_overhead_ms = 13.0` to hit the BENCHMARKS.md anchors; the planner charges only RTT + frame transfer (~3.3 ms/hop where calibrated reality is ~16.3 ms — ~5× under). And nothing tests the planner's `modeled_decode_tpot_us` against any anchor; the calibration tests validate the sim's *serial* formula, which intentionally differs (Σ stages vs max stage). Until the planner inherits the calibrated overhead, its cross-candidate ordering leans on a model known to be missing its biggest term.

6. **Upload bandwidth measurement over-estimates** (`stage_artifacts.rs:817-840`): the timed window ends when the last chunk is accepted by `write_all` on the QUIC send stream (buffered, not delivered); `finish()` happens after recording. Combined with latest-wins/no-smoothing (`node.rs:1619-1622`), a single inflated sample can govern placement for up to 30 min — and the simultaneous 30-min age-out of all observations can flip plans with no actual network change. Download side is the safer direction (includes disk writes → underestimates).

## Test-strength gaps (the four headline tests are all weaker than claimed)

- *2× bandwidth → ~2× layers*: asserts only `fast > 2×slow − 2` — a one-sided lower bound; a degenerate 39/1 split passes.
- *Memory ceilings*: enforced only via `debug_assert!` (`topology.rs:477`) — compiled out in release; the test itself asserts only "no empty stages".
- *Stage-0 binding*: uses two identical nodes, so the assertion passes even if the binding logic is deleted.
- *Partial-signal fallback*: pins the partial case (good) but contains dead code (`_spans_signaled`) where a stronger assertion was visibly removed, and never covers the fleet-scope hole in (1).

## Smaller items

- Doc/code contradiction on the kill-switch: doc says "checked per planning attempt, no restart needed"; the code comment at `split_planning.rs:437-439` (this head) says "requires a process restart". The code re-reads the env per attempt; pick one story.
- Complexity claim: the DP is O(nodes × layers²) (`topology.rs:892-936`), not the stated O(layers × nodes²).
- Stale contradictory comment `topology.rs:1069-1071` ("2 + 5 = 7 ms") vs the assertion at 1100 expecting 4 ms.
- Sim link keys referencing nonexistent node ids pass parse-time validation (only key *shape* is checked) and surface late.
- Doc promises a `source = "spec"|"measured"` corpus field and "property tests over synthetic grids" — neither exists; the "execution sim" is a closed-form analytic estimate, not discrete-event.
- 2 of 3 calibration anchors are in-sample for the back-solved knobs (`active_weight_fraction`, `per_hop_overhead_ms`); only the 3-way anchor is genuinely out-of-sample. The "10–25 tok/s @ ~20 ms RTT" anchor has no calibration test.
- Duplicated GB/s→MiB/s conversion in `sustained_perf_signals` vs `from_gossip_csvs` with slightly different guards; overflow drops the signal instead of saturating.

## Verified solid

DP boundary/feasibility mechanics (prefix sums, no empty stages, u128 determinism, zero-bandwidth/overflow → clean fallback), unit conversions (GB/s→MiB/s 953.674, TFLOP→GFLOP), div-zero guards, gossip compatibility (**no wire change** — new code only consumes pre-existing optional CSV fields; old↔new nodes trivially compatible), settle barrier unaffected by noisy signals (`split_participant_settle.rs:274-279` uses only node_id+vram), conservative min-merge edges, 30-min age gate, sim determinism (BTreeMap, no RNG), calibration residuals genuinely ~1.3–7.6%.

Items 1–4 plus the rebase are what I'd want before this merges; 5–6 are fine as tracked follow-ups in #1455 if flagged explicitly.

michaelneale:
Review note on the RTT signal, from a read of the current head plus the measurement path on `main`. Not a blocker — every #1454 use of RTT is candidate *ranking*, not admission — but the doc and the code disagree about what the number is, and the gap is in the optimistic direction.

*(Amended: an earlier version of this comment asked for a jitter/variance estimate. Withdrawn — measuring jitter meaningfully from these samples is not something we're confident we can do, and iroh's path selection is the right layer to trust for link quality. The ask is settle-time only.)*

**What the planner actually receives.** `hop_rtt_ms` (`topology.rs:698-750`) resolves a directed edge, else the endpoint node's coordinator RTT, which traces back to `Peer::rtt_ms`. That field is a **best-seen minimum**: `update_peer_rtt` (`mesh/node.rs:1490-1506`) rejects any sample above the stored value, updating only `display_rtt` for the UI. Samples arrive from a 15 s refresh loop (`heartbeat.rs:410,507-534`) and a one-shot 5 s post-connect recheck (`connections.rs:1800-1849`).

**Min-latching is the right primitive here, and the doc should say so.** The justification is the relay→direct upgrade: first gossip usually routes via relay at inflated RTT before holepunch completes (`connections.rs:1771-1776`), and a later worse sample must not clobber a good direct measurement. `update_peer_rtt` even re-elects for split when the value crosses below the ceiling (`node.rs:1518-1531`). Keep it.

The problem is naming. The cost model is written as `p50_latency(e(i))` (`PERFORMANCE_AWARE_TOPOLOGY_PLANNER.md:132`), with `p50_latency_ms` / `p95_latency_ms` / `jitter_ms` / `sample_age_ms` listed as "new probe" (`:98-100`) and `:159` conceding signals are un-aged. In practice the term is fed a p0. A reader calibrating stage-time estimates against `docs/BENCHMARKS.md` will assume a median and be biased fast on exactly the WAN edges where the error matters. **Ask: state the retention policy (best-seen minimum, relay-inclusive, un-aged) next to the `p50_latency` term, or rename the term.** If jitter/p95 are not going to be probed, prune them from the signal inventory at `:98-100` too, so the doc does not promise inputs the design will not have.

**The one substantive gap: there is no settle-time or corroboration signal.** Nothing records how many samples produced the minimum or how old it is. A single lucky sample and a thousand consistent ones are indistinguishable to the planner. Since the retention rule is deliberately optimistic, the planner is the one consumer that needs to know whether the optimism has had time to be corroborated — a link measured once, 200 ms after connect while still on the relay-to-direct transition, should not yet be priced as a settled edge.

Suggested shape: carry sample count and first/last-sample age alongside the minimum, and have the perf-aware path decline on an unsettled minimum — reusing the existing all-or-nothing fallback that a missing `sustained_mem_bandwidth_mib_per_s` already triggers (`perf_balanced_spans` returns `None`; caller keeps the capacity-greedy walk). That is a second predicate on a mechanism that already exists and is unit-tested, not new machinery. Deliberately *not* asking for a variance estimate: link quality stays iroh's job, and this only needs to answer "has this measurement had a fair chance to reach its floor".

**One relay note.** The gate is unchanged by this PR: relay-only stage paths are rejected as `StagePathRelayOnly` unless `MESH_SPLIT_ALLOW_RELAY=1` (`stage_transport.rs:139-145`). But the *RTT value* is not relay-filtered on the way in — the 15 s refresh loop takes whatever path `is_selected()` returns (`heartbeat.rs:524-531`), so a relay sample can seed `peer.rtt_ms`, while `update_peer_selected_path` (`node.rs:1550-1560`) *does* filter to direct. Two feeds, different rules, same field. Harmless under min retention (a relay sample is slower, so it loses) and it supports the relay-start/direct-upgrade story, but the field reads like "direct-path RTT" and is not. Comment fix.

**Suggested sequencing:** settle-time corroboration becomes a precondition for phase 4 (perf-aware default-on) rather than a phase 3/5 follow-up. Placement is a sticky decision, and the existing re-election path only fires in the *improving* direction (`node.rs:1520-1522`) — so an over-optimistic link never gets re-priced downward. That asymmetry is the argument for wanting the settle signal before the planner leans on the value.

i386:
**Author's verification pass over the review findings** (head `c18da85f`) — I built the sim/calibration side of this PR, so rather than a self-congratulatory self-review, this is a first-hand verification of @Galadriel's and MeshBuilder's findings plus my own fresh read. Every claim below was re-checked against source, not taken from the reviews.

**Confirmed — all four of Galadriel's should-fix items are real:**

1. **Fallback safety story is per-subset, not fleet-wide.** `perf_balanced_spans` returns `None` per candidate subset, and non-empty `edges` changes `candidate_network_ms_per_token` even for capacity-greedy plans (the greedy path at `topology.rs:515+` calls the same edge-based estimator). The parity test covers a single 2-node subset, so neither hole is exercised. Agree: weaken the doc claim and add a mixed-fleet parity test.
2. **`split_participant_set_hash` hashes 7 of the 10 signature fields** (`local_package.rs:754-760`): the tuple grew to include `large_frame_mib_per_s`, `sustained_mem_bandwidth_mib_per_s`, `sustained_compute_gflop_per_s`, but only `large_frame` reaches the hasher. The two node perf signals are silently absent from claim identity. Two-line fix; clearly accidental.
3. **Unflagged ordering change**: `topology.rs:599` (`candidate_better_for_same_shape`) and `:629-633` (`latency_candidate_ordering`) short-circuit on modeled TPOT *before* `decode_tpot_target_met` and context length, so a candidate missing the 33 ms target can now beat one meeting it. If intentional, it needs a doc line + test; if not, move the target-met key above the TPOT key.
4. **`decode_tpot_target_met` evaluates the network-only estimate** (`topology.rs:433`), not the modeled TPOT computed from the same plan. Confirmed by reading `fit_candidate`'s perf-aware early return.

**Confirmed — CI:** the macOS smoke failure reproduces the exact #1459 admission bug in the job log (`429 … "39 token deficit (capacity=256, active=0, pinned=0, request=39, minimum_free=512)"`), and this branch's merge-base `75b8866c` predates the #1459 fix `c383b4118`. Inherited from main, not this PR. (Per James: no rebase during review — noted.)

**Confirmed — MeshBuilder's node-ordering gap** is real: `fit_candidate` still stages nodes VRAM-descending; `order_pipeline_nodes` adoption is phase-2 work per the design doc, but it should be an explicit tracked item because it's the term that dominates on asymmetric/WAN edges.

**My plan (pending James):** items 1–3 are small, mechanical fixes I can land in one commit; item 4 likewise. The calibration-gap concern (planner missing the ~13 ms/hop overhead the sim proved dominant) is correctly tracked for phase 3.5/#1455 — cross-candidate TPOT ordering shouldn't be trusted for default-on until the planner inherits the calibrated overhead term.

i386:
All four review items landed in c90ec982 (51/51 skippy-coordinator, 2666/2666 host-runtime, 11/11 sim tests green locally; fmt + clippy clean):

1. Fallback scope documented per-subset + two new tests pinning mixed-fleet and edge-data behavior.
2. `split_participant_set_hash` now hashes the two node perf signals.
3. Target-met restored above the modeled-TPOT tiebreak in `latency_candidate_ordering`. Note the interaction with item 4: once met is scored against modeled TPOT, met is monotone in TPOT (met ⇔ TPOT ≤ target), so the two keys cannot conflict on the fully-signaled path — the restoration matters for mixed-signal comparisons and preserves legacy key priority.
4. `decode_tpot_target_met` scored against modeled TPOT; single-stage plans no longer trivially meet targets.

Reviewing my own work from a fresh frame while writing tests surfaced one fixture subtlety worth recording: stage order is VRAM-descending, so a stage-0 binding to a smaller node is only reachable when no larger node exists in the feasible subset — the ordering test uses a 3-node fleet for that reason.

i386:
Takeover follow-up on `26b595798` completes the remaining phase-3 signal work and settle-time corroboration:

- records steady-decode stage compute time after warmup, normalizes it to µs/layer, and maintains a bounded mean/EWMA with age
- gossips the timing as additive optional fields and only for currently hosted models
- feeds observed timing into both the span DP and serial TPOT score as a measured floor on the analytical model
- retains RTT minimum semantics while tracking sample count plus first/latest observation age
- withholds all remote performance-aware inputs until the RTT floor has at least 2 samples spanning 5 seconds, with the latest no older than 30 seconds; otherwise the exact capacity-only fallback applies
- includes timing and RTT-confidence state in the participant signature so changed evidence triggers replanning
- updates the design record to reflect phase 3 as built and removes jitter/p95 from the planner signal inventory

Verification on the exact pre-commit tree that produced `26b595798`:

- `cargo test -p mesh-llm-host-runtime --lib`: 2676 passed, 8 ignored, 0 failed
- `cargo test -p skippy-coordinator`: 52 passed, 0 failed
- simulator unit, calibration, and scenario suites: 12 passed, 0 failed
- `cargo check` across server/coordinator/simulator/protocol/host-runtime: passed
- `cargo check -p skippy-server --tests`: passed
- clippy with warnings denied across all touched packages: passed
- `cargo fmt --all -- --check` and `git diff --check`: passed

Local execution of the skippy-server unit binary remains blocked by the checkout's existing native archive link mismatch (the prepared llama archive references OpenMP `__kmpc_*` symbols while this local test link omits libomp). The test target compiles, and the server suite is left to the PR's configured CI environment rather than weakening the native link setup locally.

i386:
Final takeover status: head `2844aac63` is pushed and the full replacement CI matrix is green (90 successful checks, zero pending/failed; CodeRabbit complete).

The head supersedes `26b595798` only to regenerate the checked-in console-print ratchet after the new local-split tests shifted four pre-existing approved `eprintln!` line numbers. `cargo run -p xtask -- repo-consistency no-console-print` passes on the final tree; no new console print was introduced.

This completes the requested items 1–5 for this PR: CI/review closure, inherited macOS admission issue cleared on the current base, live per-stage timing feedback, planner/simulator TPOT consistency, and settle-time RTT corroboration.

Remaining beyond this scope: phase-4 reference-hardware A/B, then phase-5 hysteresis/migration budgeting and broader topology corpus/node-ordering adoption.

crates/mesh-llm-host-runtime/src/runtime/local_package.rs
```diff
@@ -321,12 +372,62 @@ impl SplitParticipant {
...
+            self.rtt_last_sample_age_ms = Some(observation.last_sample_age_ms);
+            let observed_span_ms = observation
+                .first_sample_age_ms
+                .saturating_sub(observation.last_sample_age_ms);
+            self.rtt_corroborated = observation.sample_count >= SPLIT_RTT_CORROBORATION_MIN_SAMPLES
+                && observed_span_ms >= SPLIT_RTT_CORROBORATION_MIN_SPAN_MS
+                && observation.last_sample_age_ms <= SPLIT_RTT_CORROBORATION_MAX_LAST_AGE_MS;
+        }
+        if !self.rtt_corroborated {
+            self.rtt_ms = None;
```
ndizazzo, line 417: `rtt_ms` isn't a perf-aware signal, and nulling it here happens outside the gate.

The doc comment above says this withholds "performance signals" so the planner reuses its capacity-only fallback, but `rtt_ms` is the pre-existing field the legacy planner already consumed. `strip_perf_aware_signals` in `split_planning.rs:442` deliberately keeps it, and its own comment spells out why: "the legacy planner consumed both, so stripping them would change capacity-only placement". On `main`, `peer.rtt_ms` flows straight through `with_package_signals`. Here it gets nulled before the gate ever runs.

With `MESH_TOPOLOGY_PERF_AWARE` unset, that means:

- `stage_transfer_latency_ms` goes `None`, so `latency_aware_planning` at `topology.rs:632` can flip to `false`, and `plan_topology_with_required_stage0` returns the first feasible candidate instead of scanning and ranking. Different plan.
- Corroboration needs `last_sample_age_ms <= 30_000` and samples come off a 15s heartbeat, so any lapse past 30s drops to non-corroborated, as does roughly the first 15-20s after peer discovery (`sample_count >= 2` plus a 5s span). That's time-varying, so it flips the replan signature back and forth.
- `rtt_corroborated` is field 11 of `SplitParticipantSignature` and isn't behind `include_perf` the way fields 5 and 8-10 are, so an ungated bool feeds `split_participant_set_hash` and churns claim identity with the flag off.

Let's move the corroboration predicate behind the gate and scope the nulling to actual perf-aware signals:

```rust
if !self.rtt_corroborated {
    // rtt_ms is a pre-perf-aware signal the legacy planner consumes; leave it.
    self.large_frame_mib_per_s = None;
    self.sustained_mem_bandwidth_mib_per_s = None;
    self.sustained_compute_gflop_per_s = None;
    self.observed_decode_us_per_layer = None;
}
```

and gate `rtt_corroborated` in the signature like the other four. Then a parity test that plans the same fixture with the flag on and off and asserts identical stages for a fleet with uncorroborated RTT would catch this class of thing. `disabled_mode_strip_keeps_pre_perf_aware_fields` only asserts the shape of `strip_perf_aware_signals`, never that the resulting plan matches pre-PR.

crates/skippy-coordinator/src/topology.rs
```diff
@@ -502,58 +650,94 @@ fn candidate_has_required_stage0(
...
-        .plan
-        .estimated_decode_network_ms_per_token
-        .unwrap_or_default();
-    candidate_estimate < current_estimate
-        || (candidate_estimate == current_estimate && candidate.cmp(current) == Ordering::Greater)
+    // This comparison is between different node subsets of the same count,
+    // so signal completeness can differ. Prefer a complete TPOT model, then
+    // compare like-for-like estimates; never turn a missing estimate into
+    // zero latency.
+    estimate_completeness(candidate)
```
ndizazzo, line 657: This makes `estimate_completeness` the primary sort key, and it applies in the capacity-only path too. Same thing in `latency_candidate_ordering` just below.

Before this, a subset whose `estimated_decode_network_ms_per_token` was `None` went through `unwrap_or_default()` to `0` and ranked best. Now it gets completeness `0` and ranks worst. With the flag off, `modeled_decode_tpot_us` is always `None`, so completeness collapses to "does this subset contain any RTT-reporting node", which differs per subset in a mixed fleet. And the `rtt_ms` finding on `local_package.rs` makes mixed fleets the common case rather than the edge case.

Concretely with nodes A (has RTT), B and C (no RTT) and `node_count = 2`: before, `{B,C}` won on an estimate of 0. Now `{A,B}` wins.

To be clear, the new behavior is better. Missing-as-zero was a real bug and you're right to kill it. But it ships to everyone, not just opt-in users, and nothing tests it with the flag off. Either apply the completeness key only when perf signals are present (keying off `!input.edges.is_empty()` would do it), or keep it unconditional and drop the byte-identical parity claim from the description and from `PERFORMANCE_AWARE_TOPOLOGY_PLANNER.md:422`, with a test pinning the new flag-off ordering so it's deliberate and visible.

crates/mesh-llm-host-runtime/src/runtime/split_planning.rs
```diff
@@ -146,11 +160,28 @@ fn topology_planning_input(input: SplitTopologyPlanInput) -> TopologyPlanningInp
...
+                sustained_compute_gflop_per_s: node.sustained_compute_gflop_per_s,
+                observed_decode_us_per_layer: node.observed_decode_us_per_layer,
             })
             .collect(),
         context_length_override: input.context_length_override,
         parallel_lanes_override: input.parallel_lanes_override,
         target_decode_tpot_ms: input.target_decode_tpot_ms,
+        // MoE active-fraction metadata is not yet plumbed from the package
+        // identity; dense (1000 permil) is the conservative default.
+        active_weight_fraction_permil: 1000,
```
ndizazzo, line 173: The comment's reasoning (dense over-estimates TPOT uniformly, so relative ordering is unchanged) holds within a fixed stage count, but not across stage counts. And `latency_candidate_ordering:696` compares `modeled_decode_tpot_us` across candidates with different hop counts. Stage service time scales with the fraction; hop terms (RTT, transfer, `CALIBRATED_PER_HOP_OVERHEAD_US`) don't.

Using your own anchor scenario (`benchmarks_anchor_pair.toml`, 40 layers at 425 MB, m4max at 417000 MiB/s, 3ms RTT, 30 MiB/s links):

- at 340 permil: stage sum around 13.2ms, 2-way hops around 32.5ms, so hop-dominated
- at 1000 permil: stage sum around 38.9ms, hops still 32.5ms, so stage-dominated

The dominance flips. That biases the planner toward more stages, spreading weight onto extra nodes across expensive hops, in exactly the regime where BENCHMARKS.md shows hops dominate. It also makes `decode_tpot_target_met` uniformly pessimistic.

The part that bothers me most is that `planner_model_matches_execution_sim` only ever runs at 340 permil, so the single test locking the planner and simulator cost models together is validating a configuration production never uses.

Either plumb the package's active-expert fraction before this goes live, or at minimum add a calibration scenario at `active_weight_fraction = 1.0` and assert the chosen stage count on the anchor fleet is the same at 340 and 1000. If it isn't, the comment's claim needs to come out.

ndizazzo (changes requested):
The direction here is right, and honestly the engineering quality is above average for a PR this size. The cost model is honest integer arithmetic, the DP is correctly formulated (its objective really is equivalent to the candidate evaluator for a fixed subset and order, since the hop and per-stage-overhead terms are constant with respect to the boundary choice), and the tri-state completeness ranking is a genuine improvement over the old missing-as-zero comparison. `planner_model_matches_execution_sim` is exactly the test I'd want here: two cost models silently drifting apart is the classic failure in this kind of work, and that test is the right defense against it.

Most of the hardening claims in the description check out against the code. `deny_unknown_fields` is on all five deserializable simulator levels. The CSV metrics are length, cardinality, and plausibility bounded before anything consumes them, and the parse is fail-closed on a single bad entry. Stage timing needs 8 fresh samples and resets on a stage-range change. Signature quantization works as described. The capacity-only fallback is the original greedy code rather than a reimplementation, which is the right structure and is why the problems below are fixable without touching the planner core.

But the central claim, that this "preserves the exact capacity-only placement path" when the mode is disabled, isn't true. It's false in two independent places, and one of them varies with time, so it'll change placement for people who never set the env var. The gate's plumbing itself is fine, `perf_aware_placement_enabled` parses correctly and there's only one production reach-point. The problem is that two behavior changes landed upstream and downstream of the gate rather than behind it. I've left both inline.

I'd also flag the shipped cost balance, inline on `split_planning.rs`: we hardcode `active_weight_fraction_permil: 1000` while the only planner/simulator calibration test runs at 340, so the balance we validated isn't the one that ships.

On file size: `topology.rs` goes from 1233 to 2218 lines, 1206 of which is implementation. `local_package.rs` goes 1011 to 1375 and `split_planning.rs` 975 to 1251. Our rule says a change touching an already-over-1k file should split out the separable responsibility as part of the change, and the added code here is about as separable as it gets. The `mod locked;` pattern is already in `topology.rs` as precedent. Roughly: `topology/cost_model.rs` for `streamed_layer_weight_bytes`, `modeled_stage_time_us{,_from}`, `pipeline_network_time_us`, `modeled_serial_decode_tpot_us`, and the calibration constants; `topology/span_dp.rs` for `serial_optimized_spans`; `topology/ranking.rs` for `estimate_completeness`, `target_status_rank`, `lower_option_is_better`, and the two comparators, with their tests moving alongside.

A handful of smaller things:

A peer advertising sub-1-MiB/s bandwidth silently disables perf-aware placement for any subset containing it. `parse_bounded_metric_csv_sum` accepts any `value > 0.0`, so `gpu_mem_bandwidth_gbps = "0.0000001"` yields `Some(0)` at `local_package.rs:478`. `Some(0)` passes the all-or-nothing gate in `serial_optimized_spans` (which only checks `is_none()`), then `modeled_stage_time_us` returns `None` for that node, the DP row is unreachable, and the whole subset falls back to capacity-greedy. Fail-safe, but silent and remotely triggerable. A `.filter(|mib| *mib > 0)` at parse plus tightening the DP gate to `is_none_or(|bw| bw == 0)` closes it.

`record_stage_decode_timing` takes a global mutex and does an O(n) `retain` on every decode step, ungated. The `decode_step < 8` early return is correctly before the lock, but every step after that pays a mutex plus a full scan of up to 128 entries, per token, per stage, on the serving path, regardless of the env var. `stage_decode_timing_hints()` already retains, so the hot-path retain is redundant.

The per-layer normalization has a direction to it. `observed_us_per_layer = compute_us / (layer_count * tokens)` folds fixed per-invocation overhead into a per-layer number, so fewer layers gives a higher us/layer, which prices the node worse, which gets it fewer layers. The range-change reset and the 8-sample minimum damp how fast that can run and there's no evidence it's a problem today, but there's no hysteresis either, and the description already lists adoption hysteresis as a follow-up. Worth measuring before default-on. Carrying (fixed, per-layer) as two coefficients instead of one ratio would remove the bias entirely.

The DP allocates a fresh `O(stages * layers)` table inside the innermost combinatorial loop at `topology.rs:1116`. `fit_candidate` runs once per node subset per (context, lanes, node_count), so per-candidate work goes from `O(k*L)` to `O(k*L^2)` with a ~160 KB heap allocation each time on a 50-stage, 80-layer plan. The combinatorial outer loop is pre-existing, but this multiplies its constant by about `L`. Hoisting the buffers out and reusing them is easy.

`pipeline_network_time_us` and `modeled_serial_decode_tpot_us` each define their own `hop_rtt_ms`/`hop_transfer_us` closures with the same exact-edge, reverse-edge, node-RTT-fallback logic and the same `bytes * 1e6 / (mib * 2^20)` expression, differing only in where the RTT fallback reads from. Two copies of a cost formula is the exact divergence `planner_model_matches_execution_sim` exists to prevent. One `hop_cost_us` taking the fallback as a closure would collapse them.

"Closes #1001" is premature. That issue reports capacity-as-allocation-weight and cites `skippy-topology/src/planning.rs:64` plus its pinning test, and neither is touched here. Line 64 still reads the same. The substance of the issue does belong to the coordinator planner this PR improves, but the improvement is off by default and the description itself says default-on is blocked on #1455. A user on defaults still gets exactly the behavior #1001 describes. Let's make it `Refs #1001` and close it when perf-aware is default-on.

On size, and I want to be direct about this because I think it's the actual lesson: this should have been three or four PRs, and the two gate bugs are the evidence. 3765 added lines across 46 files spanning a protobuf change, mesh gossip ingest, the coordinator planner core, the per-token decode path in `skippy-server`, and a whole new crate. Both blocking findings are cases where a behavior change to the legacy path got smuggled in alongside new opt-in machinery, which is precisely what gets missed when a reviewer has 3765 lines to hold at once. The prior review round also found four real defects in this same ranking and hashing region, which says something.

Seams I'd have used, in dependency order:

1. Protocol and measurement plumbing, no behavior change. `node.proto`, `proto/node.rs`, `protocol/convert.rs`, the `network/metrics.rs` sanitization, `stage_performance.rs`, the passive observation in `stage_artifacts.rs`, the RTT window and large-frame recording in `mesh/node.rs`, plus the mechanical fixture updates. Land it and prove it's inert. The hot-path lock above gets its own focused look.
2. Coordinator cost model and DP, split per the file-size note. Reviewable against the design doc on its own, and the flag-off ranking change becomes impossible to miss.
3. Simulator and corpus. New crate, zero production reach. Honestly this should have gone first, it's the instrument the other two get measured with and it's the easiest thing in the diff to approve.
4. Gate and participant plumbing. The only piece that changes production placement, arriving with the cost model already reviewed and the simulator available to test parity against. Both blocking findings would have been the whole diff.

Other things worth calling out as done well: tie-breaking is genuinely deterministic, `(total, max, previous)` tuple comparison with strict `<` over an ascending loop, on a node list sorted by `(vram desc, node_id asc)`, and no `HashMap` iteration reaches any output. `observed_us_per_layer` is used as a floor via `analytical.max(observed)` rather than a default, so `None` correctly means "no floor" instead of "zero time", which is the one place `unwrap_or_default()` on an `Option` is the right call. The three optional proto fields appended at tags 4-6 keep old and new nodes interoperating. And `link_tables_outside_links_fail_loudly` / `malformed_link_keys_fail_loudly` catch the specific TOML footgun where `["a -> b"]` parses as a quoted top-level key and silently drops edges, which is a nasty one to find later.

CI is fully green across all 90 checks. The design doc and the #1455 follow-up framing are good practice, and should ride along with the cost model piece.

PR #1458, opened 2026-08-26
refactor(just): split root Justfile into semantic imports

ndizazzo:
## Summary

Split the large root `Justfile` into nine flat semantic imports while keeping the root file as the public entrypoint.

## Why

The root recipe file had grown difficult to navigate and maintain. The split gives build, release, Skippy, mesh, packaging, web, CI, client, and utility recipes clear ownership without changing the command surface.

## Compatibility

- Preserves all 79 public recipe names and private `with-lld`.
- Preserves parameters, defaults, variadics, dependencies, platform variants, command bodies, comments, quiet behavior, and shebangs.
- Keeps `default: build` and ordinary flat imports, with no namespaced recipes.
- Adds import-aware source tests and exact summary/list/normalized dump equivalence checks.

## CI and tooling

- Updates recipe readers, xtask checks, ownership, cache invalidation, website paths, and backend classification for imported Just sources.
- Keeps protected PR planning rooted in the protected checkout while consuming validated source manifests as inert data.
- Protects source ownership and slice catalogs from unsafe PR expansion.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 just ci-validate` passed, with expected PowerShell skips.
- Focused Justfile/planner tests passed.
- `just --fmt --check`, actionlint, ShellCheck, formatter checks, diff checks, and dry-run interface checks passed.
- No release/product builds were run; this is a source-organization refactor.

## Summary by CodeRabbit

* **New Features**
  * Added cross-platform build, development, release, and packaging commands for CPU, CUDA, ROCm, and Vulkan.
  * Added mesh client, worker, model download, benchmarking, quantization, inference, and UI development workflows.
  * Added cache management, cleanup, Docker client, and process utilities.
* **Bug Fixes**
  * Improved CI change detection, cache invalidation, and pull request validation.
  * Added scheduled upstream certification checks for release reliability.
* **Documentation**
  * Updated CUDA release instructions and CI guidance.
  * Organized development and release commands into dedicated workflows.


ci/ownership.yml
```diff
@@ -114,6 +114,7 @@
         "Cargo.toml",
         "Cargo.lock",
         "Justfile",
+        "just/**",
```
i386, line 117: **Blocking: this PR's own Plan gates can never pass — the `just/**` ownership rule and the `cmp -s` source-manifest guard are dead on arrival.**

The PR Plan gates (`pr_linux.yml` etc.) check out the **default branch** and run `scripts/plan-ci.py` from that protected checkout. Main's `ci/ownership.yml` has no `just/**` rule, so this branch's changed paths fail with exactly:

```
ERROR: unable to build CI plan: ownership has no rule for changed paths:
just/build.just, just/ci.just, just/mesh-client.just, just/mesh.just, just/release-build.just,
just/release-bundle.just, just/skippy.just, just/utilities.just, just/website-ui.just
```

(run 32970562322, head b1b9292a — all five Plan jobs, `PR / Linux` etc. fail with it.)

Worse, the guard added here at line 188 requires `git show $SOURCE_SHA:ci/ownership.yml` to be **byte-identical** to main's copy — but this PR *must* modify `ci/ownership.yml` (to add `just/**`), so even if the unknown-path failure were bypassed, the branch would then be rejected with "source ownership catalog differs from the protected planner catalog". The branch is unreviewable by its own gate: catch-22. (Same trap for `ci/slices.yml` at line 184 — any PR that legitimately needs to touch these catalogs can never merge.)

Options I can see:

1. **Land the ownership/planner changes on main first, separately** (a small PR that only adds `just/**` to the ownership catalog + the `--manifest-root` plumbing), then rebase this branch on it. Bootstrap problem solved by staging: after the small PR merges, its *own* Plan gate runs from a main that already contains the new rule.

   But note the small PR still contains the `cmp -s` guard change itself — it will hit the same self-reference wall unless the guard lands with an escape hatch or lands *before* the strictness exists. Sequencing: (a) land `--manifest-root` + `just/**` rule, (b) then land the `cmp -s` guards, (c) then this branch just works.

2. **Add a sanctioned escape hatch for catalog changes**: e.g. allow the source manifest to differ when the changed paths *are only* the catalogs themselves plus entries that exist in the protected catalog, or gate the strict compare on a repo-maintainer label.

3. **Pre-register the rule on main now**: a maintainer pushes a one-line commit to main adding `just/**` under `rust` (or wherever this PR assigns it). Then this branch's Plan gates pass (protected catalog already knows the paths), and the rest of the machinery validates. The `cmp -s` guard then only bites *future* branches that change the catalogs without pre-registering.

Option 3 is the smallest immediate unblock and matches the existing "protected planner" philosophy (catalog changes are maintainer-gated by construction); option 1 is the cleanest long-term shape. Either way, this needs a decision before the branch can go green — happy to help sequence it.

— Jian Yang (agent review)

.github/actions/plan-ci/action.yml
```diff
@@ -165,8 +165,33 @@ runs:
...
+            if [[ "$mode" != "100644" || "$type" != "blob" || "$path" != "$manifest" ]]; then
+              echo "source manifest is missing or is not a regular file: $manifest" >&2
+              exit 1
+            fi
+          done
+          if ! git show "$SOURCE_SHA:ci/slices.yml" | cmp -s - ci/slices.yml; then
+            echo "source slice catalog differs from the protected planner catalog" >&2
+            exit 1
+          fi
+          if ! git show "$SOURCE_SHA:ci/ownership.yml" | cmp -s - ci/ownership.yml; then
```
i386, line 188: Nice piece of design — using validated source manifests as inert data (compared against the protected planner copy, then archived out of the source SHA) preserves the "planner is trusted, PR content isn't" property while letting the plan reflect per-PR shape. A couple of sharpening points:

- The `ls-tree` mode/type check requires `100644` blobs — symlinked or submoduled manifests fail closed, which is the right default. Good.
- **Bootstrap**: the first PR that legitimately modifies `ci/ownership.yml` or `ci/slices.yml` can never satisfy the byte-identical compare (see my comment on the guard) — the escape path for catalog evolution needs to be explicit somewhere (docs? a `catalog-change` label? sequenced maintainer merges?). Right now the mechanism for changing these files at all is undefined.
- The `--manifest-root` flag is a clean seam; consider documenting in `scripts/plan-ci.py`'s docstring that `--manifest-root` is the *audit* path (source manifests as data) and the default path is the *authority* (protected catalog), so future readers don't "simplify" it away.

— Jian Yang (agent review)

i386 (commented):
Reviewed as agent on James's behalf. The mechanical split itself is well-executed: 79 recipe names preserved, import-aware source tests, exact summary/list/dump equivalence checks, and the CI tooling (readers, xtask checks, cache invalidation, backend classification) all updated to know about `just/*.just`. The planner integration via validated source manifests is thoughtfully designed.

**One blocker, and it's structural:** this branch can't pass its own CI. The PR Plan gates run the planner from the protected main checkout, where no `just/**` ownership rule exists — so the branch fails "ownership has no rule for changed paths: just/*.just" (run 32970562322, all five Plan jobs). And the new byte-identical manifest guard added in this PR would reject the branch again even if that were fixed, because the branch must modify `ci/ownership.yml` to add the rule. Details and three sequencing options inline.

Recommended unblock: maintainer pre-registers `just/**` on main (one-line commit), or the ownership/planner changes land separately ahead of this branch. After that I'm happy to do a second pass over the recipe-content equivalence (the part CI can't fully prove) if useful.

— Jian Yang (agent review)

.github/actions/plan-ci/action.yml
ndizazzo: Addressed. `ci/ci.md`, the CI inventory, and `scripts/plan-ci.py` now document protected catalog authority, the audit-only PR manifest path, and sequenced maintainer catalog merges. We also exercised that process via #1464 before rebasing this branch; all Plan gates now pass.

ci/ownership.yml
ndizazzo: Resolved through the protected catalog sequence: #1464 added `just/**` to `main`, then this branch was rebased onto merge commit `9e626398e`. All five Plan jobs and stable PR lane summaries now pass.

PR #1459, opened 2026-08-26
fix(skippy): restore small-context admission and state handoff

i386:
## Summary

- normalize impossible resident-KV watermarks so requests that fit a small KV pool are admitted while non-evictable overcommit still fails closed
- derive Skippy's restored session cursor from llama.cpp's native memory position instead of parsing its private sequence-state wire header
- add scheduler regressions for watermarks larger than, and equal to, KV capacity

## Root cause

The main-branch smoke failures exposed by `836aeb0f` were introduced by the earlier capacity-aware admission change. A 256-cell smoke runtime received a 39-token request with `minimum_free=512`; treating that configuration-derived watermark as a hard reserve made the first request mathematically impossible and returned HTTP 429.

The [latest-upstream canary](https://github.com/Mesh-LLM/mesh-llm/actions/runs/32964721554/job/98164615089) was a separate compatibility regression. Skippy inferred the imported cursor by reading fixed offsets in llama.cpp's private state payload. Current upstream prepends a magic value and sequence id, so the full-state path read sequence id `0` as the cell count and reported restored position 0 after native state had correctly restored one token.

## Validation

- `just ci-validate` — 606 passed, 7 skipped
- `cargo fmt --all --check`
- `cargo check -p mesh-llm`
- `cargo test -p skippy-scheduler --lib --features scheduler-lab` — 30 passed
- `cargo test -p skippy-runtime --lib` — 79 passed
- `cargo test -p skippy-server --lib` — 532 passed, 3 ignored
- `cargo test -p mesh-llm --lib` — 48 passed
- complete patch queue applies at pinned llama.cpp `d222767c7a6516559a3f49e7721b6c6b1acc87b4` and current upstream `539f24529bdf99e0baefd54fefff1660034bfe7b`
- static CPU llama.cpp builds pass at both revisions
- model-backed Llama 3.2 1B `state-handoff` passes exact roundtrip, restored output, suffix-prefill, and repeated cache-hit checks on current upstream
- local OpenAI smoke with a 256-cell KV pool and configured 512-token watermark returns HTTP 200

## ABI inventory

No public C ABI declarations, Rust FFI mirrors, or ABI version constants changed. Patch 0023 changes only the internal implementation of existing state-import functions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Capacity planning now accepts requests that fit within available pool space, even when configured free-space thresholds are larger than the pool.
  - Requests are still rejected when non-evictable usage leaves insufficient capacity.
  - Restored sessions now calculate their position from active runtime state, improving accuracy for partial and complete state imports.
  - State restoration now reports clear errors for unavailable memory, empty sequences, or unsupported positions.

### DIFF
--- a/crates/skippy-scheduler/src/capacity.rs
+++ b/crates/skippy-scheduler/src/capacity.rs
@@ -57,15 +57,29 @@ pub fn plan_component_capacity(
     snapshot: &ComponentCapacitySnapshot,
     demand: CapacityDemand,
 ) -> CapacityPlan {
-    let minimum_free = demand.minimum_free_units.min(snapshot.capacity_units);
-    let target_free = demand
-        .target_free_units
-        .max(minimum_free)
-        .min(snapshot.capacity_units);
     let protected_units = snapshot
         .active_units
         .saturating_add(snapshot.pinned_cache_units)
         .saturating_add(demand.request_units);
+    // A decode-batch watermark is a hint derived from the serving
+    // configuration. It may be larger than a small runtime's entire KV pool;
+    // when it fills or exceeds the pool, cap it at the space that can actually
+    // remain after the non-evictable work and this request. Otherwise a request
+    // that fits in the pool is rejected before any evictable cache can be
+    // considered. A smaller configured watermark remains a hard minimum.
+    let available_free = snapshot.capacity_units.saturating_sub(protected_units);
+    let minimum_free = if demand.minimum_free_units >= snapshot.capacity_units {
+        available_free
+    } else {
+        demand.minimum_free_units
+    };
+    let target_free = if demand.target_free_units >= snapshot.capacity_units {
+        available_free
+    } else {
+        demand.target_free_units
+    }
+    .max(minimum_free)
+    .min(snapshot.capacity_units);
     let protected_with_minimum = protected_units.saturating_add(minimum_free);
     if protected_with_minimum > snapshot.capacity_units {
         return CapacityPlan {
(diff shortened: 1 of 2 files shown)


i386 (commented):
Independent review (Jian Yang) — LGTM, safe to merge.

**Scope verified:** single commit `a7dd03fc` off `836aeb0f` (main), 2 files, +233/−5. No bot auto-fix commits on the branch.

**Fix 1 — watermark normalization (`crates/skippy-scheduler/src/capacity.rs`)**
- Root cause confirmed: `minimum_free = demand.min(capacity)` turned a configured 512-token reserve into an impossible requirement on the 256-cell CI smoke pool (39-token request + 256 reserve > 256) → 429 on every request.
- New logic is sound: when a watermark ≥ pool capacity it is capped at `capacity − protected` (saturating), so a request that fits is admitted exactly; genuine non-evictable overcommit still fails closed with a deficit (`still_rejects_when_non_evictable_work_exceeds_the_pool`). Watermarks smaller than the pool remain hard minimums — unchanged.
- Boundary `watermark == capacity` is admitted and covered by a dedicated regression. Ran the module locally: 8/8 capacity tests pass.
- Call path (`token_generation.rs` `restore_or_record_kv_on_runtime`): `minimum_free = n_batch` (default 2048) vs `kv_pool_tokens()` — the mismatch is config-derived, so treating it as a hint is the right product call. The 429 + retry-after path survives for real overcommit.

**Fix 2 — patch 0023 (state.cpp)**
- Removes all private wire-header parsing (the actual canary break: upstream now prepends magic + seq id, so the old offset read seq id 0 as the cell count → position 0 after a correct native restore).
- Position is now derived post-import from the public `llama_memory_seq_pos_max` + 1, after `synchronize()`. Verified both `llama_memory_hybrid` and `llama_memory_hybrid_iswa` expose `get_mem_recr()` and `seq_pos_max()` at the pinned upstream `d222767c` — compiles at both revisions (native build lanes green).
- The removed pre-import "cell count exceeds context" check is covered downstream: `llama_kv_cache::state_read_meta` fails closed when cells don't fit, and the Rust side now hard-asserts `validate_imported_full_state_position(expected == native)` with pass/fail unit tests. Defense in depth improved, not weakened.
- No public ABI surface changed (internal implementation only; static ABI lane green).

**CI:** every lane green on the head, including `ci-smoke-test.sh` which runs `--ctx-size 256` — the exact configuration that was broken. Model-backed state-handoff oracles pass on current upstream.

One note for the record: a configured watermark larger than a small pool is now best-effort rather than a hard reserve. If we ever want strict enforcement on tiny pools, `plan_component_capacity` is the place to revisit — the code comment already flags it.

PR #1463, opened 2026-08-27
Send remote mesh inference directly into the serving pipeline

michaelneale:
Remote mesh inference requests can now enter the OpenAI-compatible serving path directly over the authenticated QUIC stream, avoiding an unnecessary plaintext loopback TCP hop. Local agents continue to use the existing localhost HTTP listener.

## Architecture

- Adds `ClientStream`, an `AsyncRead + AsyncWrite` abstraction over TCP and Iroh QUIC bidirectional streams.
- Reuses the existing byte-oriented OpenAI request, routing, streaming, MoA, response, and error pipeline for both transports.
- Replays request bytes consumed during tunnel inspection through a prefixed QUIC reader, avoiding header/body loss.
- Installs direct ingress during runtime startup while retaining the localhost-port bridge as a compatibility fallback for embedders that do not install a direct handler.
- Keeps local upstream model communication on localhost TCP; only remote mesh ingress bypasses the extra loopback connection.

## Protocol

No wire-format or API compatibility change. Remote peers continue sending HTTP-shaped requests over authenticated mesh tunnel streams, and localhost HTTP ingress remains available for local agents.

## Validation

At commit `2c1498194cd7e965c04be9390bfdcd2b5d109da6`:

- Workspace/all-target checks passed.
- Full `mesh-llm-host-runtime` suite passed: **2663 passed, 0 failed, 8 ignored**.
- Only pre-existing macOS linker-version warnings appeared.

Added coverage for direct ingress, prefixed request handling, and durable architecture artifacts.

## Summary by CodeRabbit

* **New Features**
  * Added support for handling remote OpenAI HTTP requests directly over authenticated QUIC connections.
  * Remote tunneled requests can now reach inference targets without requiring a local API listener.
  * Added request metadata identifying direct HTTP and remote QUIC traffic.

* **Bug Fixes**
  * Preserved prefetched request data when processing tunneled HTTP streams.
  * Requests are rejected promptly with a service-unavailable response when remote inference is paused.
  * Improved cancellation when clients disconnect while responses are being routed.


crates/mesh-llm-host-runtime/src/network/openai/ingress_tests/durable_artifacts.rs
```diff
@@ -27,9 +27,10 @@ async fn parsed_missing_model_error_persists_the_client_visible_response_artifac
         let (stream, _) = listener.accept().await.expect("accept ingress client");
         handle_api_proxy_connection(
             node,
-            stream,
+            stream.into(),
             election::ModelTargets::default(),
             affinity::AffinityRouter::new(),
+            crate::runtime::IngressType::LocalOpenAi,
```
ndizazzo, line 33: @coderabbitai Is this plugin-safe?

crates/mesh-llm-host-runtime/src/network/openai/client_stream.rs
```diff
@@ -0,0 +1,112 @@
...
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
+use tokio::net::TcpStream;
+
+type QuicBiStream = tokio::io::Join<iroh::endpoint::RecvStream, iroh::endpoint::SendStream>;
+
+/// Client-facing byte stream accepted by the OpenAI ingress.
+///
+/// Local callers arrive over TCP. Remote mesh callers already have an
+/// authenticated QUIC bi-stream, which enters the same request path without
+/// opening a second plaintext loopback connection.
```
ndizazzo, line 36: very nice - this is a good abstraction

ndizazzo (approved):
@coderabbitai Does this PR make sufficient use of the new tracing so relevant info shows up in our log viewer?

PR #1465, opened 2026-08-27
fix(ci): validate CUDA runtime toolkit labels

### LINKED ISSUE #1370: CUDA runtime packaging can mislabel CUDA 13 builds as toolkit_major 12
## Summary

A CUDA native runtime built on a CUDA 13-only host can be packaged with `runtime.backend.cuda.toolkit_major: 12`, after which startup correctly rejects the otherwise local runtime as incompatible.

## Reproduction

While validating #1366 on Vast instance 47994050:

- Image: `nvidia/cuda:13.0.0-devel-ubuntu24.04`
- Host exposes only `libcudart.so.13` / `libcublas.so.13`
- GPU: RTX 5090 (SM120)
- Command used the normal release-runtime recipe with `backend=cuda cuda_arch=120`, but did not explicitly set `MESH_CUDA_VERSION` or `MESH_LLM_CUDA_TOOLKIT_MAJOR`
- Produced manifest reported:

```json
"cuda": {
  "gpu_arches": ["120"],
  "toolkit_major": 12
}
```

`mesh-llm runtime list --available --json` then reported:

```json
"cuda_toolkit_major_mismatch": {
  "required": 12,
  "installed": [13]
}
```

All `benchmark tune` trial children exited before readiness, including the speculation-disabled control, with `no compatible native runtime found`.

For diagnostic validation only, changing the manifest compatibility field to `13` made the exact same libraries load and run successfully; `libllama.so` remained SHA256 `33c01561d6a9afa372b8a0cc5207336ffca4fd8905b758839e86a3a66e2d3b27`.

## Likely cause

The `justfile` defaults `MESH_CUDA_VERSION` to `12`, then exports `MESH_LLM_CUDA_TOOLKIT_MAJOR` from that value (`justfile:189-192`, similarly `215-218`). `scripts/package-native-runtime.sh` trusts the supplied/default major. Thus local CUDA builds on a CUDA 13 toolkit are mislabeled unless the caller knows to override the environment.

## Expected

Packaging should derive/validate the manifest toolkit major against the toolkit that actually built the runtime (for example via `nvcc --version`), or fail on a mismatch rather than silently emitting an incompatible manifest.

ndizazzo:
## Summary

- Derive CUDA runtime manifest compatibility from the compiler or toolkit that actually built the runtime.
- Validate explicit CI and release CUDA versions against the selected compiler.
- Fail with an actionable error when CUDA packaging has no trustworthy toolkit evidence.

Fixes #1370.

## Reproduction

On a CUDA 13-only build host, package a CUDA runtime without manually setting the toolkit-major environment variables. Before this change, the packager could stamp `runtime.backend.cuda.toolkit_major` as 12 even though `nvcc` and the linked runtime libraries were CUDA 13. Runtime discovery then rejected the local artifact as incompatible.

## Resolution

- Use one shared detector with compiler precedence `CUDACXX`, `CMAKE_CUDA_COMPILER`, `NVCC`, then `nvcc` on `PATH`.
- Accept toolkit-owned version metadata only when no compiler is available.
- Do not treat `nvidia-smi` driver capability or a hard-coded default as build-toolkit evidence.
- Validate `MESH_CUDA_VERSION` and `MESH_LLM_CUDA_TOOLKIT_MAJOR` against the detected compiler/toolkit before build and packaging.
- Pass release and Windows matrix versions into the same validation path.

## Testing

- Focused CUDA toolkit and packaging contract suite: 57 passed.
- `just ci-shellcheck` for changed shell scripts: passed.
- `just ci-validate`: passed, including actionlint and 642 script tests with 7 expected skips.
- `just build`: passed, including UI, neutral host, patched Metal runtime, and product composition.

Actual CUDA 13, aarch64 CUDA, and Windows runners were not available locally; the PR lanes validate those platform contracts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA toolkit detection across builds and packaging.
  * Builds now provide clear diagnostics when a compatible CUDA compiler or toolkit is unavailable.
  * CUDA versions are validated against the selected compiler to prevent mismatched runtime packages.
  * Removed silent fallbacks that could produce incorrectly labeled CUDA packages.

* **Chores**
  * Updated Windows and Linux release workflows to pass CUDA version information consistently.
  * Added automated coverage for CUDA detection, validation, packaging, and workflow configuration.


i386 (approved):
**Review: approve** — verified against `origin/main...pr/1465` (13 files, scripts + tests only).

The fail-closed redesign is right: compiler precedence (`CUDACXX` → `CMAKE_CUDA_COMPILER` → `NVCC` → PATH `nvcc`), toolkit metadata only when no compiler exists, `nvidia-smi` demoted out of the evidence chain entirely, and explicit-selector failure no longer silently falling back to PATH. `cuda_toolkit_manifest_version` cross-checks `MESH_LLM_CUDA_TOOLKIT_MAJOR` and `MESH_CUDA_VERSION` against the detected compiler before either build or packaging stamps the runtime manifest — exactly the #1370 failure mode (CUDA 13 host stamped as 12).

Checked the blast radius of removing the `echo "12"` fallback from `detect-cuda-toolkit-version.sh`: both callers (`release-build-cuda`, `release-build-aarch64-cuda` in `just/release-build.just`) run it under `set -euo pipefail`, so a missing toolkit now aborts the recipe with the actionable message instead of selecting the pre-Blackwell SM list on a CUDA 13 host. `build-llama.sh` similarly turns `configure_cuda_toolkit_env || true` into a hard exit with a pointer to the env vars. 57 focused tests + the packaging/workflow contract suites cover the matrix.

PR #1466, opened 2026-08-27
fix(ui): render LaTeX in active chat responses

### LINKED ISSUE #1360: Model chat output does not support LaTeX rendering
## Problem

I asked chat a math problem, and the model correctly used LaTeX to represent a series. The chat text does not support LaTeX, however.

Add support to render LaTeX inline with markdown and plain text.

## Steps to reproduce

1. Chat with Qwen 3.8
2. Ask it a question where it uses mathematical symbols (like: "What is the 3x+1 conjecture")
3. Observe it inserts controls like: `[ 4 \rightarrow 2 \rightarrow 1 \rightarrow 4 \rightarrow 2 \rightarrow 1 \rightarrow \cdots ]`

## Extra context

```
The 3x+1 conjecture is also called the Collatz conjecture. It is a famous unsolved problem in mathematics involving a simple sequence of operations on positive integers.

The rule is:

Start with any positive integer ( n ).
If ( n ) is even, divide it by 2.
If ( n ) is odd, multiply it by 3 and add 1.
Repeat the process forever.
The conjecture says that no matter which positive integer you start with, the sequence will eventually reach the cycle:

[ 4 \rightarrow 2 \rightarrow 1 \rightarrow 4 \rightarrow 2 \rightarrow 1 \rightarrow \cdots ]

So the sequence is predicted to always end in the repeating loop 1 → 4 → 2 → 1 → ...

Example: start with 7
[ 7 \rightarrow 22 \rightarrow 11 \rightarrow 34 \rightarrow 17 \rightarrow 52 \rightarrow 26 \rightarrow 13 \rightarrow 40 \rightarrow 20 \rightarrow 10 \rightarrow 5 \rightarrow 16 \rightarrow 8 \rightarrow 4 \rightarrow 2 \rightarrow 1 ]

Then it repeats:

[ 1 \rightarrow 4 \rightarrow 2 \rightarrow 1 \rightarrow \cdots ]

Why it is interesting
The rule is extremely simple, but no one has proved that it works for all positive integers. It has been checked by computer for an enormous range of numbers, and no counterexample is known, but a general mathematical proof remains open.

In short: the 3x+1 conjecture claims that every positive integer eventually falls into the repeating cycle 1 → 4 → 2, but this has not yet been proved.
```

ndizazzo:
## Summary

- Render inline and display LaTeX in the shipped chat route.
- Reuse one Markdown renderer for assistant responses, thinking segments, streamed output, and completed output.
- Preserve literal code, ordinary brackets, escaped currency, links, copy selection, and raw-HTML safety.

Fixes #1360.

## Reproduction

Ask a model such as Qwen 3.8 to explain the Collatz conjecture. The active chat route previously rendered model output such as `[ 4 \\rightarrow 2 \\rightarrow 1 \\rightarrow \\cdots ]` as literal TeX because `MessageRow` used a GFM-only Markdown stack. KaTeX support existed only in an older chat component.

## Resolution

- Route active assistant and thinking content through the shared KaTeX-capable `MarkdownMessage` component.
- Normalize `$...$`, `$$...$$`, `\\(...\\)`, `\\[...\\]`, and model-style bracketed TeX outside inline and fenced code.
- Keep delimiters literal while a response is streaming, then render math when the message completes.
- Keep safe external-link handling, inert raw HTML, native copy selection, and `KaTeX` trust disabled.
- Reset the KaTeX fallback state when the expression changes.

## Testing

- Focused renderer tests: 39 passed.
- `just ui-test`: 202 files, 1,652 passed, 3 skipped.
- UI lint, format, and typecheck: passed.
- `just build`: passed, including UI, neutral host, patched Metal runtime, and native runtime packaging.
- Headless Chromium reproduction rendered 3 KaTeX nodes and 1 display block for the Collatz sample.

## Summary by CodeRabbit

* **New Features**
  * Improved chat message formatting with consistent Markdown rendering for responses and thinking content.
  * Added support for inline and block mathematical notation, including TeX delimiters.
  * Enhanced rendering of code blocks, tables, headings, lists, links, and diagrams.
  * Added configurable link behavior and message presentation styles.

* **Bug Fixes**
  * Preserved mathematical notation while responses stream.
  * Improved fallback handling for unsupported or invalid math.
  * Prevented unsafe raw HTML from rendering in messages.


i386 (approved):
**Review: approve** — read the full renderer diff (`MessageRow.tsx`, `MarkdownMessage.tsx`, `math-delimiters.ts`, `KaTeXBlock.tsx`) and the test additions.

The delimiter normalizer handles the traps I'd expect it to: fenced and inline code ranges are protected before any rewriting; markdown link/definition/shortcut-reference brackets are excluded from bracketed-TeX conversion; escaped `\$` stays literal; `\(`/`\)`/`\[`/`\]`/`$$` are converted to remark-math's dollar syntax with unpaired delimiters passed through untouched. Streaming keeps delimiters literal with backslash doubling so CommonMark doesn't eat them, and math renders on completion only — that avoids mid-stream re-parse flicker. `rehypeHighlight` + `remarkMath` ordering and the `language-math`/`language-mermaid` code-component interception look right.

Currency (`$5 ... $10`) relies on remark-math's no-trailing-space rule for single dollars; the escaped-dollar case is tested. One untested edge if you ever want it: prose like `see $x$ and $y$` where both are intended as math is fine, but two *unintended* adjacent single dollars around non-space text (`$3$4`) would render as math — inherent to single-dollar mode and not worth disabling it for.

Headless Chromium reproduction + 1,652 UI tests passing is good evidence for a rendering change.

Issue #1462, opened 2026-08-27
spec: wire every TOML configuration setting end to end with regression-proof TDD


ndizazzo:
## Problem

The TOML schema currently promises configuration that the runtime does not execute. Some values are discarded, some fail only when a model loads, some reach an intermediate Rust struct but never reach Skippy or llama.cpp, and a few native capabilities have no TOML key at all.

The audit at `836aeb0f0` covered every field in `crates/mesh-llm-config/src/model.rs`, every CLI flag, the Rust resolver and native bridge, and the patched Skippy/llama.cpp consumers. Before filing, this was compared with current `main` at `c383b411`; none of the audited config, Rust bridge, or website source paths changed. The only native-path change was the unrelated restored-position patch. The most serious examples are:

- `model_fit.kv_unified=false|auto` is discarded and `true` aborts model loading. llama.cpp derives the value without reading TOML.
- `hardware.placement`, `hardware.tensor_split`, `hardware.cpu_moe`, and `hardware.n_cpu_moe` pass static validation and then fail at model load.
- `advanced.server.alias` has no production reader.
- several `skippy.*` lifecycle and topology values are parsed but ignored while constants or planner defaults win.
- several sampling and chat defaults fail during resolution even though the schema and website present them as usable.
- speculative-tuning trials vary two values that the runtime ignores, so those benchmark dimensions are no-ops.
- `docs/skippy/CONFIGURATION.md` claims wiring that does not exist, while the website splits configuration across overlapping pages and repeats the same stale assumptions.

This issue replaces those promises with a staged, test-driven implementation plan and a synchronized documentation contract.

## Non-negotiable scope

1. Every supported setting must be configurable through `config.toml` at `[defaults.*]` and, where the schema allows it, per-model overrides.
2. This work must **not add settings to the CLI**. Existing CLI flags may keep working, but no PR in this stack should expand the CLI configuration interface unless a separate issue explicitly authorizes it.
3. A field is not "wired" because TOML deserializes, validates, or reaches an intermediate resolved struct. It is wired only when changing the TOML value changes the required downstream consumer or observable runtime behavior.
4. Every implementation PR uses TDD. The PR must contain a test that fails on its parent commit for the audited reason, then passes after the implementation. A config round-trip test does not satisfy this requirement.
5. Every PR updates the relevant website source and synchronized internal status matrix in the same PR. No implementation PR may leave documentation for a later cleanup.
[...]

PR #1467, opened 2026-08-27
fix(runtime): resolve startup GPU selectors and honor CLI --device

### LINKED ISSUE #1374: mesh-llm serve --model bypasses pinned-GPU resolution, fails with "unknown selected backend device: pci:..."
## Problem

`mesh-llm serve --model <ref>` with `gpu.assignment = "pinned"` and a per-model `hardware.device = "pci:<bdf>"` override in `config.toml` fails to start the model, even though the pinned GPU is present and correctly enumerated:

```
InvalidArgument: unknown selected backend device: pci:00000000:01:00.0
```

I expected the pinned GPU's stable PCI id to resolve to its ggml backend device name (e.g. `CUDA0`), the same way it does for `mesh-llm gpus` and for a bare `mesh-llm serve` invocation with no CLI model override.

Root cause: passing `--model` explicitly on the CLI routes model startup through the "ad-hoc model" code path in `build_startup_model_specs`, which sets `config_owned: false` for every spec built that way:
- https://github.com/Mesh-LLM/mesh-llm/blob/3fcd5dc487129131a84f25beb6632f67a7918a85/crates/mesh-llm-host-runtime/src/runtime/startup_models.rs#L624
- (`config_owned: false` at L639, L661, L678)

`preflight_config_owned_startup_models_with_gpus` — the function that resolves a model's pinned GPU stable id into a real backend device name (`CUDA0`, `ROCm0`, etc.) — explicitly skips any spec where `config_owned` is false:
- https://github.com/Mesh-LLM/mesh-llm/blob/3fcd5dc487129131a84f25beb6632f67a7918a85/crates/mesh-llm-host-runtime/src/runtime/startup_models.rs#L1008

So for a CLI-specified `--model`, `StartupPinnedGpuTarget.backend_device` (and therefore `SkippyRuntimeConfig.selected_backend_device`) never gets populated with the resolved name. Something downstream instead passes the raw `hardware.device` string from `config.toml` (a `pci:<bdf>` stable id) straight through to the native runtime, which calls `ggml_backend_dev_by_name()` on it verbatim — that function only knows device names like `CUDA0`, never `pci:...`, so it returns null and the native layer rejects the model load:
- https://github.com/Mesh-LLM/mesh-llm/blob/3fcd5dc487129131a84f25beb6632f67a7918a85/.deps/llama.cpp/src/skippy/model_load_config.cpp (line ~111, `skippy_apply_selected_backend_device`, on the pinned llama.cpp checkout — resolved via `just llama-prepare`)

Confirmed this is pre-existing on `main` (`3fcd5dc48`), not introduced by PR #1366 — zero diff between `main` and that PR's branch in `crates/mesh-llm-host-runtime/src/runtime/` or `crates/mesh-llm-system/src/hardware/`.

## Steps to reproduce

Repro'd live on a 2-GPU box (RTX 5090 + RTX 3080), CUDA 13.3, freshly built `x86_64-unknown-linux-gnu` CUDA runtime (`just release-build-cuda`).

1. In `~/.mesh-llm/config.toml`, set:
   ```toml
   [gpu]
   assignment = "pinned"

   [[models]]
   model = "unsloth/Qwen3.8-27B-GGUF:UD-Q4_K_XL"

   [models.hardware]
   device = "pci:00000000:01:00.0"
   ```
2. Confirm the GPU enumerates correctly and this stable id is real:
   ```
   $ mesh-llm gpus
   🖥️ GPU 0
     Name: NVIDIA GeForce RTX 5090
     Stable ID: pci:00000000:01:00.0
     Backend device: CUDA0
   ```
3. Start serving with the model named explicitly on the CLI:
   ```
   $ mesh-llm serve --model 'unsloth/Qwen3.8-27B-GGUF:UD-Q4_K_XL' --port 9337 --console 3131
   ```
4. Observe startup failure:
   ```
   failure=model=unsloth/Qwen3.8-27B-GGUF:UD-Q4_K_XL: Failed to start model ...:
   InvalidArgument: unknown selected backend device: pci:00000000:01:00.0
   ```
5. Compare against a **bare** invocation with no `--model` flag (startup models sourced purely from `config.toml`), which succeeds and loads normally:
   ```
   $ mesh-llm serve --port 9337 --console 3131
   ...
   16:39:05 OK    sequence=387 status=Ok emitter=OpenThread: Native runtime selected a backend device
   16:39:05 INFO  Model loaded: unsloth/Qwen3.8-27B-GGUF:UD-Q4_K_XL
   ```

Also reproduces with `--ctx-size` added to the CLI in step 3; ctx-size override is not the trigger, `--model` is.

## Diagnostics

- Command: `mesh-llm serve --model '<ref>' --port 9337 --console 3131` vs bare `mesh-llm serve`
- Platform/backend: Linux (Arch), CUDA 13.3 backend, `x86_64-unknown-linux-gnu` target, dynamic host + adjacent native runtime bundle (`MESH_LLM_NATIVE_RUNTIME_BUNDLE_DIR` set to the packaged `dist/native-runtimes/meshllm-native-runtime-linux-x86_64-cuda13`)
- Not a mesh/network issue: `mesh=partial api=ready console=ready`, failure is scoped to `model readiness=failed`; reproduces with no `--auto`/`--join`/`--publish`
- `gpu.assignment = "pinned"` with a per-model `hardware.device` override is required to trigger this; `gpus` subcommand output confirms the stable id ↔ backend device mapping is correct at the hardware-detection layer, so this is specifically a `serve`-startup resolution gap for CLI-named models, not a detection bug

## Extra context

Found while validating #1366's MTP default-depth change. Not fixed as part of that PR since it's pre-existing, unrelated to that diff, and touches cross-crate startup-planning logic (`mesh-llm-host-runtime::runtime::startup_models`) rather than a scoped bug fix.

Two plausible directions, for whoever picks this up to weigh:
1. Make `preflight_config_owned_startup_models_with_gpus` (or an equivalent pinned-GPU resolution step) also run for CLI-named models that happen to match a `config.toml` catalog entry with a pinned `hardware.device`, instead of only for `config_owned` specs.
2. If `--model` naming a config-catalog model is intentionally meant to fully bypass its `config.toml` entry (matching the "ad-hoc model" precedent set for `--gguf`), then reject `gpu.assignment = "pinned"` + `--model <catalog-ref>` combinations at startup with a clear error, rather than passing an un-resolved stable id into the native runtime and surfacing a confusing FFI-level error.

ndizazzo:
## Summary

Supersedes #1414. This PR keeps #1467's stricter model-selection behavior and incorporates the useful CLI device override and startup preflight behavior from #1414.

Fix pinned GPU and explicit device resolution when starting models through `mesh-llm serve`.

## Behavior

- An exact, unique `--model <ref>` match may inherit that configured model's pinned GPU selector.
- Duplicate configured refs are rejected because the CLI has no profile selector.
- Unmatched refs and explicit `--gguf` artifacts do not inherit per-model identity or tuning, though they may inherit `defaults.hardware.device`.
- Explicit `--device` overrides inherited selectors. Stable GPU IDs and backend names such as `CUDA0` resolve before native startup, including when GPU assignment is automatic.
- `--device CPU` bypasses GPU-only preflight and reaches direct and split runtimes.
- `--device auto` preserves the inherited selector.
- Config identity and the selected device are preserved through local startup and split topology replans.
- Unavailable device errors include the detected backend devices.

## Testing

- Focused startup tests: 74 passed, 1 ignored
- Full host-runtime library suite: 2,825 passed, 8 ignored
- Split runtime resolver test: passed
- Clippy for all host-runtime targets with warnings denied: passed
- `cargo fmt --all --check`: passed
- `git diff --check`: passed
- `just build`: passed, including the composed Metal native-runtime bundle

Fixes #1374
Supersedes #1414

## Summary by CodeRabbit

* **New Features**
  * Added explicit device selection for local models, including CPU, GPU identifiers, backend names, and automatic selection.
  * Improved model startup behavior so exact configured model matches can inherit pinned hardware settings while preserving command-line model options.
  * Added configurable health-check and lifecycle timing for split runtimes.
  * Improved split-runtime topology selection and model-specific configuration handling.

* **Documentation**
  * Clarified model, device, GPU, configuration precedence, and startup behavior in usage, design, and testing guides.

* **Bug Fixes**
  * Improved handling of duplicate, unmatched, and ad-hoc model references during startup.


i386 (approved):
**Review: approve, with one behavior question and one sequencing note.**

The core fix is correct and well-tested: exact-ref CLI `--model` matches now carry the configured model's effective device selector (`hardware.device` → legacy `gpu_id` → `[defaults.hardware]` fallback) into `StartupModelSpec`, `resolve_pinned_gpu` gates pinned preflight, ambiguous duplicate refs bail instead of inheriting an arbitrary entry's hardware, and config metadata stays out of CLI-owned selection (`config_model_id: None` for exact matches). `matching_config_model` correctly excludes paths/aliases from matching.

**Question — ad-hoc CLI models now inherit the global default selector and hard-fail preflight.** With `gpu.assignment = "pinned"` and `[defaults.hardware].device` set, an ad-hoc `--gguf`/non-matching `--model` launch now gets `gpu_id = Some(default)` and the widened gate (`!resolve_pinned_gpu && plan.gpu_id.is_none()`) sends it into `resolve_pinned_gpu_strict`. If the defaults selector is a bare backend name (`CUDA0`), `is_pinnable_gpu_stable_id` rejects it (`NonPinnableConfiguredId`) and startup aborts — previously these specs were skipped and launched unpinned. If that fail-closed behavior for ad-hoc models under pinned assignment is intended (it is coherent with the direction of the #1462 stack), a test pinning it would be worth adding; if not, ad-hoc specs should not inherit the defaults selector.

**Sequencing note:** this overlaps #1414 (same two files, `build_startup_model_specs` + preflight, divergent designs — #1414 also wires `RuntimeOptions.device`/`--device`, which this PR leaves unread). They will conflict hard. This branch is green and current; #1414 is 3 days stale with 5 red checks (the `--device CPU` preflight bug Jian Yang flagged). Suggest landing this first and rebasing #1414's `--device` handling onto it.

PR #1469, opened 2026-08-27
fix(hf): avoid AArch64 Xet TLS illegal instructions

### LINKED ISSUE #874: Release binary can SIGILL on aarch64 during Hugging Face/Xet downloads
## Summary

A release build of `mesh-llm` can crash with `Illegal instruction (core dumped)` during `mesh-llm models download` when downloading a Hugging Face/Xet-backed layer-package file on some aarch64 machines.

The crash was observed while downloading:

```bash
mesh-llm models download unsloth/gemma-4-26B-A4B-it-GGUF
```

Observed output:

```text
   ⏬ meshllm/gemma-4-26B-A4B-it-UD-Q4_K_M-layers/model-package.json   0.0% (9KB/0B)
📦 Downloading layer package meshllm/gemma-4-26B-A4B-it-UD-Q4_K_M-layers (2 file(s))
   ⏬ meshllm/gemma-4-26B-A4B-it-UD-Q4_K_M-layers/shared/metadata.gguf   0.0% (0B/16MB)   📄 files 1/2 completeIllegal instruction (core dumped)
```

This happened with `mesh-llm` built from main at commit:

```text
57a3c24ce2862fffc8e0a7f5c470ec7cd5c66cff
```

## Reproduced Environment

Reproduced on `lemony29.patio51.com` from:

```text
/home/ndizazzo/dev/mesh/mesh-llm
```

Relevant host details:

```text
Linux lemony-29 5.15.0-1047-nvidia-tegra-igx ... aarch64
CPU: Cortex-A78AE
```

The CPU flags include `sha1` and `sha2`, but not `sha512`.

The release binary reproduced the crash with fresh caches:

```bash
HF_HUB_CACHE=/tmp/meshllm-repro-release-hub \
HF_XET_CACHE=/tmp/meshllm-repro-release-xet \
timeout -s INT 180 \
./target/release/mesh-llm models download unsloth/gemma-4-26B-A4B-it-GGUF
```

Result:

```text
Illegal instruction (core dumped)
exit code: 132
```

A debug binary from the same checkout got past `shared/metadata.gguf` and continued into later package downloads, so this appears to be release-build-specific on this host.

## Root Cause Evidence

The failing file is Xet-backed on Hugging Face. A HEAD request for `shared/metadata.gguf` includes Xet metadata such as `x-xet-hash`, `xet-auth`, and `xet-reconstruction-info` links. A plain `curl -fL` download of the same file succeeded, so the artifact itself does not appear to be corrupt.

A `strace` reproduction showed the process receiving SIGILL while in the HTTPS/download path:

```text
--- SIGILL {si_signo=SIGILL, si_code=ILL_ILLOPC, si_addr=...} ---
```

Using same-run `/proc/<pid>/maps` plus `addr2line` on the release binary mapped the faulting address to:

```text
aws_lc_0_39_1_sha512_block_data_order_hw at err_data.c:?
```

Disassembly around the crash site shows ARM SHA-512 instructions:

```text
sha512su0 v16.2d, v17.2d
sha512h   q3, q5, v6.2d
sha512su1 v16.2d, v23.2d, v7.2d
```

On this host, `/proc/cpuinfo` does not advertise the `sha512` CPU feature, so executing `sha512su0` traps with SIGILL.

## Why This Would Happen

This does not look like malformed GGUF parsing, layer-package parsing, or ordinary Rust application logic. The process crashes before Mesh reads the GGUF payload.

The likely dependency path is:

```text
mesh-llm models download
  -> layer package materialization
  -> hf-hub / hf-xet download path
  -> reqwest / rustls
  -> aws-lc-rs / aws-lc native crypto
  -> ARM SHA-512 hardware routine
```

The release binary appears to execute an AWS-LC SHA-512 hardware implementation on an aarch64 CPU that lacks the required `sha512` instruction support. That suggests one of the following:

- AWS-LC runtime CPU-feature dispatch selected the SHA-512 hardware path incorrectly.
- The release build configuration or linked object selected hardware-specific code without a sufficient runtime guard.
- A dependency update/build artifact produced a binary that is not portable across the aarch64 machines we expect to support.

## Suggested Reproduction Test

On an aarch64 machine without the `sha512` CPU flag:

1. Build a release binary from main:

   ```bash
   just release-build
   ```

2. Confirm CPU flags:

   ```bash
   grep -m1 '^Features' /proc/cpuinfo
   ```

   The reproducing host had `sha1` and `sha2`, but no `sha512`.

3. Run the download with fresh caches:

   ```bash
   rm -rf /tmp/meshllm-repro-hub /tmp/meshllm-repro-xet
   HF_HUB_CACHE=/tmp/meshllm-repro-hub \
   HF_XET_CACHE=/tmp/meshllm-repro-xet \
   ./target/release/mesh-llm models download unsloth/gemma-4-26B-A4B-it-GGUF
   ```

4. If it crashes, collect:

   ```bash
   strace -f -o /tmp/meshllm-sigill.strace \
     env HF_HUB_CACHE=/tmp/meshllm-strace-hub \
         HF_XET_CACHE=/tmp/meshllm-strace-xet \
     ./target/release/mesh-llm models download unsloth/gemma-4-26B-A4B-it-GGUF
   ```

5. For symbolization, capture `/proc/<pid>/maps` from the same run before the crash, then map the SIGILL address back into `target/release/mesh-llm` with `addr2line`.

A smaller focused reproducer may also be possible by calling the same `hf-hub` / `hf-xet` download path for:

```text
meshllm/gemma-4-26B-A4B-it-UD-Q4_K_M-layers/shared/metadata.gguf
```

That would avoid a full `mesh-llm` CLI invocation while still exercising the suspected dependency path.

## Proposed Hardening

1. Avoid AWS-LC for portable aarch64 release builds if practical.

   If we can select a TLS/crypto backend that is known to be portable across our supported aarch64 fleet, prefer that for release artifacts. The important requirement is that hardware crypto code must be runtime-gated correctly.

2. Add a release-build HTTPS download smoke test on target hardware.

   Debug builds did not reproduce this on lemony29. A meaningful smoke test should use `just release-build` and then perform a small Hugging Face/Xet-backed download with fresh caches.

3. Add a focused aarch64 CPU-feature smoke.

   On aarch64 CI or lab machines, record `/proc/cpuinfo` features and run a minimal TLS/HF download path. This catches binaries that accidentally require unsupported instructions.

4. Add an operator escape hatch for Xet downloads.

   Consider an environment variable such as:

   ```text
   MESH_LLM_DISABLE_HF_XET=1
   ```

   This would let operators force a plain HTTP/HF download path when hf-xet or one of its native dependencies is broken on a specific host.

5. Prefer plain HTTP fallback for small package metadata files.

   `model-package.json` and `shared/metadata.gguf` are small enough that a direct HTTP fallback is reasonable. If the Xet path fails or is disabled, Mesh could download these files via ordinary redirected HTTPS before invoking the heavier Xet path for large tensors.

6. Document SIGILL triage for release binaries.

   Add a short troubleshooting note for `Illegal instruction` on aarch64 that asks for CPU flags, release/debug distinction, `strace` SIGILL address, and `addr2line` output. This would make future reports much faster to classify.

## Notes

This should probably be tracked in Mesh even if the ultimate fix lands upstream in `aws-lc-rs`, `aws-lc`, `rustls`, `hf-hub`, or `hf-xet`, because Mesh release artifacts need to be portable across the aarch64 machines we advertise or commonly run on.

ndizazzo:
## Summary

Prevent release binaries from selecting AWS-LC SHA-512 instructions on AArch64 CPUs that do not advertise `FEAT_SHA512` during Hugging Face/Xet downloads.

## Reproduction

On an AArch64 Linux host such as Cortex-A78AE with `sha1` and `sha2` but no `sha512` CPU flag:

1. Build the release product with `just release-build`.
2. Use fresh `HF_HUB_CACHE` and `HF_XET_CACHE` directories.
3. Download `meshllm/gemma-4-26B-A4B-it-UD-Q4_K_M-layers/shared/metadata.gguf` or the parent layer package.
4. Before this change, the release process can terminate with SIGILL in `aws_lc_*_sha512_block_data_order_hw`.

## Resolution

- Detect AArch64 SHA-512 capability before constructing Hugging Face clients.
- On AArch64 CPUs without SHA-512, install rustls ring as the process default before reqwest/hf-xet can select AWS-LC.
- Preserve an already-installed crypto provider and leave normal provider selection unchanged on other targets.
- Apply the guard at the shipped binary entrypoint and every reusable Hugging Face client construction boundary.
- Add a bounded, isolated Xet-backed metadata download smoke to the ARM release artifact lane.

The explicit iroh QUIC crypto configuration is unchanged.

## Testing

- Focused model-hf tests: passed
- Debug product build and isolated Xet smoke: passed (15,783,680 bytes)
- Release host build and release-mode Xet smoke: passed
- `just ci-validate`: 634 passed
- Focused Clippy with warnings denied: passed
- Release bundle composition/checksum verification: passed for both archives
- Cross-target scripts reached Linux/Windows compilation boundaries; this macOS host lacks those C toolchains

Fixes #874

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Hugging Face downloads and cache access across ARM64 and other platforms by selecting a compatible TLS provider automatically.
  * Preserved existing TLS provider choices while maintaining certificate and hostname verification.

* **Tests**
  * Added release smoke tests for Hugging Face/Xet downloads using the packaged ARM64 binary.
  * Portability checks now report system details, retry transient failures, enforce time limits, and retain hard failures for illegal-instruction errors.
  * Added coverage for retry, skip, and failure scenarios.


crates/model-hf/src/tls.rs
```diff
@@ -0,0 +1,145 @@
...
+            .map(|()| HfTlsProvider::AwsLc)
+            .unwrap_or(HfTlsProvider::Existing),
+    }
+}
+
+fn provider_action(has_arm_sha512: bool, has_existing_provider: bool) -> HfTlsProvider {
+    if has_existing_provider {
+        HfTlsProvider::Existing
+    } else if has_arm_sha512 {
+        HfTlsProvider::AwsLc
```
i386: **The `AwsLc` branch isn't needed, and it silently flips the process-wide TLS provider for everything else in the binary.**

On a SHA-512-capable CPU, installing AWS-LC changes nothing for Hugging Face: `reqwest` is built with `__rustls-aws-lc-rs` in this graph (confirmed via `cargo tree -e features -i rustls@0.23.41` -> `hyper-rustls feature "aws-lc-rs"` <- `reqwest feature "__rustls-aws-lc-rs"`), so `default_rustls_crypto_provider()` already returns `aws_lc_rs::default_provider()` when no process default exists — and notably it does *not* install it. Doing nothing here yields the identical provider.

What it does change is everyone else. `crates/mesh-llm/src/main.rs:17` now runs this as the first statement of `main()`, so the process default is AWS-LC before anything else gets a chance. Every pre-existing ring install is `let _ = ...` and now silently loses the race:

- `crates/mesh-client/src/network/nostr.rs:79,181,210`
- `crates/mesh-llm-host-runtime/src/network/nostr/discovery.rs:83,167`
- `crates/mesh-llm-host-runtime/src/network/nostr/publish.rs:50`

So nostr/relay TLS moves from ring to AWS-LC across the product on every platform except the aarch64-without-SHA512 one this PR is about. The PR description says the change "leave[s] normal provider selection unchanged on other targets" — that isn't what happens.

I think this branch got added in b3bbccd in response to the CodeRabbit comment claiming `HFClientBuilder::build()` would panic under ambiguous rustls features. That claim doesn't hold for reqwest 0.13.4: reqwest calls `builder_with_provider(...)` with an explicitly chosen provider and never routes through `CryptoProvider::get_default_or_install_from_crate_features()`, which is the only thing that panics on ambiguity (`rustls-0.23.41/src/crypto/mod.rs:243-257`).

Suggested fix: have `provider_action` return a distinct "leave it alone" outcome when SHA-512 is available, so the only behavioral change in this PR is the one it's actually about. If a deliberate process-wide default *is* wanted, that's a reasonable call — but then make it an explicit product decision and delete the six now-dead ring installs above so there's a single owner of the choice.

i386 (commented):
## Review: `fix(hf): avoid AArch64 Xet TLS illegal instructions`

The diagnosis is right and the mechanism works. I verified the load-bearing assumption rather than taking it on trust:

- `reqwest 0.13.4` reads the process default and only falls back if none is installed — `reqwest-0.13.4/src/async_impl/client.rs:719`: `CryptoProvider::get_default().map(...).unwrap_or_else(default_rustls_crypto_provider)`. So installing `ring` before the first `HFClientBuilder::build()` genuinely redirects both hf-hub and hf-xet off AWS-LC. 
- Coverage of construction sites is complete. All seven `HFClientBuilder::new()` call sites in the workspace (`model-hf/src/lib.rs:128,399`, `model-hf/src/store/local.rs:128`, `mesh-llm-host-runtime/src/models/mod.rs:63,99`, `model-package/src/lib.rs:15`, plus the test) are now guarded.
- `libc::HWCAP_SHA512` is defined for aarch64 gnu, musl *and* android in the locked `libc 0.2.189`, so the `target_os = "android"` arm in that `cfg` compiles.

Four things I'd want addressed before merge, plus two nits. The first two are the ones that matter.

**Caveat on my verification:** I could not run `cargo test -p model-hf` locally — my host ran out of disk during the build. Everything below is from source reading plus `cargo tree -e features --locked` on the PR branch; the test-run claims in the PR description are yours, not re-confirmed by me.

crates/model-hf/src/tls.rs
```diff
@@ -0,0 +1,145 @@
...
+        HfTlsProvider::AwsLc => rustls::crypto::aws_lc_rs::default_provider()
+            .install_default()
+            .map(|()| HfTlsProvider::AwsLc)
+            .unwrap_or(HfTlsProvider::Existing),
+    }
+}
+
+fn provider_action(has_arm_sha512: bool, has_existing_provider: bool) -> HfTlsProvider {
+    if has_existing_provider {
+        HfTlsProvider::Existing
```
i386: **The guard silently no-ops when any provider is already installed — including an unsafe one — and every caller discards the result.**

`provider_action(false, true)` returns `Existing`: on the exact CPU class this PR targets, if *anything* installed a provider first, the SIGILL comes back and nothing anywhere says so. All call sites are `let _ = configure_hf_tls_provider();`, and the two install arms above also collapse to `Existing` via `.unwrap_or(...)` when the install loses a race — so "the fallback worked" and "the fallback failed" are the same value.

Today the only in-repo pre-installs are ring (safe), so this is latent rather than live. But the ordering guarantee that makes it safe is invisible, untested, and one refactor away from breaking — and its failure mode is a hard crash in a customer's release binary with no log line to point at.

Cheap fix: when `!arm_sha512_available()` and a provider already exists, emit a `tracing::warn!` (or return a distinct `ExistingUnverified` variant) so a future reordering is diagnosable instead of silent.

crates/model-hf/src/tls.rs
```diff
@@ -0,0 +1,145 @@
...
+        assert_eq!(provider_action(false, true), HfTlsProvider::Existing);
+    }
+
+    #[test]
+    fn selects_ring_when_unsupported_arm_has_no_provider() {
+        assert_eq!(provider_action(false, false), HfTlsProvider::Ring);
+    }
+
+    #[test]
+    fn real_hf_client_builds_after_provider_initialization() {
```
i386, line 158: **The one code path this PR exists to add — the ring install — is never executed by any test or CI job.**

`provider_action` is well covered, but it's a pure function over two bools; the interesting parts are the two things it doesn't touch:

1. `arm_sha512_available()` has no test on any platform.
2. `HfTlsProvider::Ring` never gets installed anywhere in CI. x86_64 takes the `AwsLc` branch, and so does Apple aarch64 — I checked on an M-series host: `hw.optional.armv8_2_sha512: 1`. So this test only ever exercises AWS-LC, and its assertion (`matches!(selected, Existing | Ring | AwsLc)`) accepts every variant of the enum, so it can't fail on provider choice regardless.

The new release smoke does cover the ring path, but only if the ARM runner's CPU happens to lack FEAT_SHA512 — and `runs-on` for that job is `${{ needs.metadata.outputs.runner_arm }}`, a variable that can resolve to a Depot/Graviton runner. If that ever resolves to a SHA-512-capable CPU, the regression coverage disappears with no signal.

Worth adding a test-only seam (env override or `#[cfg(test)]` injection point) that forces the ring branch, so `configure ->  ring -> build a real HF client` is covered on every platform. Having the smoke script print the CPU flags it observed would also make a silent runner swap visible.

Nit while you're here: `serial_test` is already a dev-dependency of this crate, but this test isn't `#[serial]` despite asserting on a process-global `OnceLock`. It passes today only because nothing else in the binary installs a provider.

.github/workflows/release.yml
```diff
@@ -968,6 +968,7 @@ jobs:
           fi
           "$binary" --help | head -5
           "$binary" runtime --help | head -20
+          scripts/ci-hf-xet-portability-smoke.sh "$binary"
```
i386, line 971: **This smoke runs only in the release workflow, so the script is entirely untested by this PR — and it makes a release-blocking hard dependency on a live Hugging Face fetch.**

`smoke_linux_arm64_artifact` is release-only; nothing in PR CI invokes `scripts/ci-hf-xet-portability-smoke.sh`. `gh pr checks 1469` confirms it — the green "HuggingFace download smoke" job is a pre-existing one, not this. So the first time these 95 new lines of bash and embedded Python ever execute will be during an actual release cut.

Combined with `set -euo pipefail` and a single `timeout 180` with no retry, the failure surface is: an unauthenticated HF fetch 429s, the CDN blips, or `meshllm/gemma-4-26B-A4B-it-UD-Q4_K_M-layers` gets renamed — and the release fails on something unrelated to the artifact being validated.

Two suggestions, either is fine:
- Wrap the download in a retry/backoff and treat a network-class failure as a clearly-labelled skip rather than a hard fail, so it can only fail the release on an actual SIGILL.
- Or move the guard's real verification somewhere hermetic (see my note on the test file) and keep this download smoke advisory.

Separately, the CodeRabbit note about `timeout` being absent is worth taking — it currently `exit 2`s, which on this lane means a release failure on a runner-image change.

crates/model-hf/Cargo.toml
```diff
@@ -13,8 +13,10 @@ async-trait = "0.1"
 chrono = { version = "0.4", features = ["serde"] }
 dirs = "6.0.0"
 hf_hub = { package = "mesh-llm-hf-hub", version = "1.0.2", default-features = false, features = ["blocking"] }
+libc = "0.2"
 model-artifact = { path = "../model-artifact", version = "0.76.0-rc7" }
 model-ref = { path = "../model-ref", version = "0.76.0-rc7" }
+rustls = { version = "0.23", default-features = false, features = ["aws_lc_rs", "ring", "std"] }
```
i386: Nit: `default-features = false` drops `tls12` and `logging` from what this crate asks for. It's fine today purely by feature unification — `cargo tree -e features -i rustls@0.23.41` shows `rustls feature "tls12"` supplied by `hyper-rustls`/`tokio-rustls` — but `model-hf` on its own would build TLS-1.3-only. If it's ever extracted or those deps shift, TLS 1.2 disappears silently. Cheapest to just say what you mean: `features = ["aws_lc_rs", "ring", "std", "tls12", "logging"]`.

For the record, this dep is not what makes the rustls feature set ambiguous — `rustls feature "ring"` was already enabled on `main` via `tokio-rustls feature "ring"` <- `async-wsocket`. Adding it here is redundant but harmless, and arguably self-documenting.

i386 (changes requested):

i386 (approved):
**Review: approve** — this head (`df04f61f5`, "preserve TLS provider selection") resolves both blocking findings from the earlier review of `b3bbccd9e`:

1. **No more process-wide provider flips.** The `AwsLc` branch is gone entirely (verified: zero references). Non-aarch64 and SHA-512-capable aarch64 now take `Automatic`/`Existing` — reqwest keeps its normal selection and an already-installed provider is never replaced, so the six nostr `ring` installs keep winning as they do today.
2. **The silent no-op is now observable.** An existing-but-unverifiable provider on a non-SHA512 CPU yields `ExistingUnverified` with a `tracing::warn!` naming the SIGILL risk, and the ring-install race loser also warns instead of returning an indistinguishable "ok".

Also checked the smoke-script concern: `ci-hf-xet-portability-smoke.sh` now retries 3× and exits `::warning`-advisory (0) on non-SIGILL failures — a 429 or transient HF outage no longer hard-fails a release cut; only exit 132 (SIGILL) propagates. The Linux `getauxval(AT_HWCAP)` / Apple `sysctlbyname` probes and the conservative default for other aarch64 targets are right. One residual nit, not blocking: the ring-install path still has no CI execution (x86_64 and M-series both short-circuit to `Automatic`), so its first real exercise remains whatever aarch64 host runs the release workflow.

scripts/ci-hf-xet-portability-smoke.sh
```diff
@@ -0,0 +1,134 @@
...
+  echo "::warning::Hugging Face/Xet advisory smoke skipped: timeout is unavailable" >&2
+  exit 0
+fi
+
+error_file="$smoke_root/download-error"
+download_succeeded=false
+for attempt in 1 2 3; do
+  echo "Hugging Face/Xet advisory smoke attempt $attempt/3" >&2
+  set +e
+  timeout 180 "${command[@]}" >"$output_file" 2>"$error_file"
```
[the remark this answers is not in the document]
ndizazzo: Addressed in 374c48154: the smoke command now uses `timeout --kill-after=10s 180s`, with regression coverage for both timeout arguments. The rebased branch passes `just test-all` locally.

PR #1475, opened 2026-08-27
config: derive static validation from an exhaustive schema-status manifest

ndizazzo:
Part of #1462 (PR 1/8). Stacked on #1474.

This PR adds no native/runtime behavior and no CLI flags. It gives `mesh-llm config validate` a single source of truth for what actually runs today, so operators find out about a wiring gap at `config validate` time instead of at model load or first request.

## What changed

- Added an exhaustive schema-to-status manifest (`crates/mesh-llm-config/src/wiring_status.rs`, `WIRING_MANIFEST`, 237 entries) covering every canonical `config.toml` path, each with its `Wired`/`Partial`/`Unwired`/`Rejected` status, the owning implementation PR from the #1462 stack, and why it has that status today.
- Derived static validation from that manifest (`crates/mesh-llm-config/src/wiring_validation.rs`): `mesh-llm config validate` now emits a `RejectedField` error for permanently-rejected fields, an error for fields that pass static validation today but fail at model load or request resolution, and a warning for fields that are silent no-ops.
- Added a permanent contract test (`crates/mesh-llm-config/tests/wiring_status_contract.rs`) proving static `mesh-llm config validate` flags a field that is known (by audited resolver code) to fail at model load, and a test proving the manifest's status agrees, path for path, with the `Status` column of `website/src/docs/pages/config-reference.md` and the `Wiring status` column of `docs/skippy/CONFIGURATION.md` (reusing the parity helpers landed in PR 0).
- Added precedence and behavior tests: `[defaults.*]` applies, `[[models]]` overrides win, `auto`/omission stays distinct from an explicit value, and rejected/gap fields surface diagnostics without changing any default value or `auto` semantics.

Red: `cargo test -p mesh-llm-config --test wiring_status_contract -- --nocapture` fails on RED commit `f9eb38ec477770eb84bbc1d91a35c1ffa79621a8` (immediate parent `e111cf4b1be6596f9d768c14c4699fd1370ede7c`) with the intended `model_fit.keep_tokens` static-validation assertion, after compiling successfully.
Green: the same command passes 1/1 on `248a648b7401d96b6c9b91634b4e1b8c01c18a2d`.
[...]

PR #1476, opened 2026-08-27
config: wire kv_offload, kv_unified, and swa_full model_fit controls

ndizazzo:
Part of #1462 (PR 2/8). Stacked on #1475.

## What changed

Wires 5 of the 12 `model_fit` paths in scope end to end, and gives concrete
per-field evidence for the 7 that need a larger follow-up:

- Extends the Skippy native ABI with a tri-state `kv_offload` / `kv_unified` /
  `swa_full` surface and wires `model_fit.kv_offload`, `model_fit.kv_unified`,
  `model_fit.swa_full`, and `model_fit.kv_cache_policy`'s KV-offload
  expansion: `config.toml` -> resolver -> `SkippyModelLoadOptions` ->
  `StageConfig` -> `skippy-server::RuntimeConfig::as_raw()` -> the patched
  llama.cpp `skippy_runtime_config` struct -> `llama_context_params`
  (`offload_kqv`, `kv_unified`, `swa_full`).
- Wires `model_fit.cache_idle_slots` with **no native ABI change**: its real
  consumer is `skippy-server::RuntimeState`'s idle-session pool
  (`crates/skippy-server/src/runtime_state/lane_lifecycle.rs`), not
  `llama_context_params`. `capped_target_idle_sessions` bounds
  `prewarm_idle_sessions`'s admission target, and `drop_session_timed`
  discards a returned lane instead of growing the idle pool past the
  configured cap.

`true`/`false` now sets the native value directly for the three
`llama_context_params` fields; unset/`auto` preserves today's derived
default. Recurrent/hybrid architectures still force `kv_unified = true`
natively regardless of the requested value (a correctness invariant, not a
preference). `cache_idle_slots` unset preserves today's unbounded pool
(capped only by `lane_count`).

## Red / Green

Red: `cargo test -p mesh-llm-host-runtime --lib kv_unified_true_resolves_and_reaches_stage_config -- --nocapture` fails on RED commit `9a861049850fec73dea0833c1c586b9f2491cec2` (immediate parent `248a648b7401d96b6c9b91634b4e1b8c01c18a2d`) after compiling, with the intended unsupported-`kv_unified` resolver assertion.
Green: the same command passes 1/1 on implementation commit `bf8a87fd9467f64356a5e976eff4a3368cee4cb2` and final head `bd46a452e4b39366e5e6cb1f91c54ce160d1aae6`.
Full suite: `cargo test -p skippy-runtime --lib` (84 passed), `cargo test -p mesh-llm-config --test wiring_status_contract` (1 passed), `cargo run -q -p xtask -- repo-consistency no-console-print`, `cargo fmt --all --check`, `cargo check -p mesh-llm`, `cargo clippy -p mesh-llm -p mesh-llm-config -p mesh-llm-host-runtime -p skippy-ffi -p skippy-runtime -p skippy-server --all-targets -- -D warnings`, `scripts/prepare-llama.sh pinned`, and `python3 scripts/generate-skippy-api-doc.py --check` pass on `bd46a452e4b39366e5e6cb1f91c54ce160d1aae6`; broader historical host-runtime and workspace suites were not rerun during restack.

## Scope

Included TOML paths (12): `model_fit.kv_offload`, `model_fit.kv_unified`,
`model_fit.cache_ram_mib`, `model_fit.cache_idle_slots`,
`model_fit.keep_tokens`, `model_fit.context_shift`, `model_fit.swa_full`,
`model_fit.checkpoint_interval`, `model_fit.checkpoint_count`,
`model_fit.lookup_cache_static`, `model_fit.lookup_cache_dynamic`,
`model_fit.kv_cache_policy`.

Non-goals: `hardware.*`, `throughput.*`, `skippy.*`, `speculative.*`,
`request_defaults.*` (other PRs in the stack). No CLI flags or CLI settings
added.

### Per-path outcome

| Path | Status | Evidence |
|---|---|---|
| `model_fit.kv_offload` | **Wired** | `crates/skippy-runtime/src/config.rs` capture tests; resolver tests `kv_offload_resolved_reaches_model_load_options_via_kv_cache_policy`, `kv_offload_true_and_false_reach_different_stage_configs` |
| `model_fit.kv_unified` | **Wired** | capture tests in `config.rs`; resolver tests `kv_unified_defaults_layer_and_per_model_override_wins`, `kv_unified_true_resolves_and_reaches_stage_config` |
| `model_fit.swa_full` | **Wired** | capture tests in `config.rs`; resolver tests `swa_full_defaults_layer_and_per_model_override_wins`, `swa_full_true_and_false_reach_different_stage_configs` |
| `model_fit.kv_cache_policy` | **Wired** | existing `resolver_macro_expands_kv_cache_tuning_profile_and_safety_margin` plus `kv_offload_resolved_reaches_model_load_options_via_kv_cache_policy` |
| `model_fit.cache_idle_slots` | **Wired** | `crates/skippy-server/src/runtime_state/lane_lifecycle.rs` unit tests `capped_target_idle_sessions_clamps_to_the_configured_bound`, `capped_target_idle_sessions_is_unbounded_when_unset`; `crates/skippy-server/src/runtime_state.rs::cache_idle_slots_reaches_the_idle_session_pool_bound`; resolver tests `cache_idle_slots_true_resolves_and_reaches_stage_config`, `cache_idle_slots_defaults_layer_and_per_model_override_wins` |
| `model_fit.cache_ram_mib` | Left rejected | Real per-token KV byte size lives in `crates/model-artifact/src/gguf/kv_cache.rs::GgufKvCacheQuant::kv_cache_bytes_per_token(&GgufCompactMeta)`, which needs GGUF metadata `skippy-server` (`crates/skippy-server/Cargo.toml`) does not depend on and `skippy_protocol::StageConfig` does not carry. A MiB budget can only be converted to a session/byte cap with that per-model KV size; without it any conversion is an unverifiable guess. `mesh-llm config validate` still fails it via the PR 1 `BailsDownstream` manifest entry (unchanged). |
| `model_fit.keep_tokens` / `model_fit.context_shift` | Left rejected | `crates/skippy-server/src/runtime_state.rs` has no context-overflow/token-eviction logic at all today — only `session_token_counts: BTreeMap<String, u64>` bookkeeping (line 56). Wiring these means adding a new admission-time retention/eviction policy to the decode loop (`crates/skippy-server/src/frontend/local_generation/linear_decode.rs` or the iteration scheduler), not connecting an existing capability. `BailsDownstream` unchanged. |
| `model_fit.checkpoint_interval` / `model_fit.checkpoint_count` | Left rejected | `crates/skippy-runtime/src/session.rs` has `export_state`/`import_state` primitives, but no periodic-checkpoint scheduler or bounded retained-checkpoint ring exists anywhere in `skippy-server`'s decode loop to call them on an interval. Wiring these means building that scheduler/ring, not connecting an existing one. `BailsDownstream` unchanged. |
| `model_fit.lookup_cache_static` / `model_fit.lookup_cache_dynamic` | Left rejected | `crates/skippy-runtime/src/ngram.rs::Cache` wraps `skippy_ngram_cache_create/reset/append/draft`, but `include/skippy/speculative_decoding.h` (native ABI, lines 33-58) exposes no load/save/merge entrypoints wrapping upstream `common/ngram-cache.h`'s `common_ngram_cache_load`/`common_ngram_cache_save`. This genuinely needs new native ABI functions (unlike the other 4 wired-this-round fields, which reused existing `llama_context_params`/pool-bound consumers), which is a larger, separately reviewable native-ABI change. `BailsDownstream` unchanged. |

## Native ABI inventory

| Status | Symbol/declaration | Public header | Implementation / mirror | Reason | Lockstep update |
|---|---|---|---|---|---|
| Added | `int32_t skippy_runtime_config::kv_offload` | `include/skippy/runtime.h` | `src/skippy/model_loading.cpp` (`skippy_resolve_tristate` applied to `llama_context_params.offload_kqv`); Rust mirror `crates/skippy-ffi/src/abi.rs::RuntimeConfig::kv_offload` | Lets `model_fit.kv_offload` reach the native K/V backend-offload decision instead of being silently dropped | Rust mirror + `skippy-runtime::config::RuntimeConfig::as_raw()` updated in this change; old ABI not supported |
| Added | `int32_t skippy_runtime_config::kv_unified` | `include/skippy/runtime.h` | `src/skippy/model_loading.cpp` (applied to `llama_context_params.kv_unified`, both the primary stage context and the MTP draft attach context); Rust mirror `RuntimeConfig::kv_unified` | Lets `model_fit.kv_unified` override the lane-count-derived default instead of always bailing when set to `true` | Same as above |
| Added | `int32_t skippy_runtime_config::swa_full` | `include/skippy/runtime.h` | `src/skippy/model_loading.cpp` (applied to `llama_context_params.swa_full`); Rust mirror `RuntimeConfig::swa_full` | Lets `model_fit.swa_full` reach the native sliding-window-attention cache mode; previously had no consumer at all | Same as above |
| Added | `#define SKIPPY_TRISTATE_AUTO (-1)` / `SKIPPY_TRISTATE_FALSE (0)` / `SKIPPY_TRISTATE_TRUE (1)` | `include/skippy/runtime.h` | Mirrored as `skippy_ffi::{TRISTATE_AUTO, TRISTATE_FALSE, TRISTATE_TRUE}` | Sentinel values for the new tri-state fields | Rust mirror updated in this change |
| No change | — | — | `model_fit.cache_idle_slots` is a pure Rust `skippy-server::RuntimeState` pool bound; no native ABI field involved | — | — |

ABI version: `SKIPPY_ABI_VERSION_PATCH` `41` -> `42` in
`include/skippy/common.h` (via new trailing patch
`third_party/llama.cpp/patches/0024-skippy-add-kv_offload-kv_unified-and-swa_full-runtim.patch`,
per the "ordinary capability change" flow in the llama-stage-patch-changes
skill — appended after the existing queue rather than editing patch 0003's
struct literally, to avoid invalidating the mail-format `index` hashes of
every downstream patch that also touches `runtime.h`/`common.h`) and
`crates/skippy-ffi/src/lib.rs::ABI_VERSION_PATCH`.

Backward compatibility with older native runtimes is intentionally not
required (internal lockstep ABI, per the llama-stage-patch-changes skill); the
Rust FFI mirror and both native call sites (primary stage load, MTP draft
attach) were updated in the same commit as the struct change.

Public-header compilation: `include/skippy/runtime.h` is a C11/C++17
umbrella-compatible header (unchanged include structure; new fields are
plain `int32_t`). `scripts/prepare-llama.sh pinned` reapplies the full
23-patch + new 0024 queue cleanly (`git am` clean apply, verified against a
fresh pinned checkout), proving the mail-format patch is well-formed.
`python3 scripts/generate-skippy-api-doc.py --check` passes against the
prepared checkout, confirming the regenerated
`website/src/docs/pages/skippy-api.md` matches the new header inventory.

Tests exercising the native ABI boundary: `crates/skippy-runtime/src/config.rs`
capture tests (`runtime_config_raw_defaults_kv_session_controls_to_auto`,
`runtime_config_raw_forces_kv_offload_when_configured`,
`runtime_config_raw_forces_kv_unified_when_configured`,
`runtime_config_raw_forces_swa_full_when_configured`) assert directly on
`RuntimeConfig::as_raw()`'s output using the same production
`skippy-ffi::RuntimeConfig` struct the native ABI consumes — no second
test-only mapper. These are pure-Rust tests (the crate's default
`dynamic-runtime` feature does not statically link the native library), so
they do not require a native rebuild; the native patch itself was validated
by reapplying the full patch queue to a fresh pinned `.deps/llama.cpp`
checkout.

## Docs

- `website/src/docs/pages/config-reference.md`: `model_fit.kv_offload`,
  `model_fit.kv_unified`, `model_fit.swa_full`, `model_fit.cache_idle_slots`,
  `model_fit.kv_cache_policy` rows flipped to `wired`.
- `docs/skippy/CONFIGURATION.md`: matching `Wiring status` rows for KV
  offload / Unified KV cache / SWA full / Cache idle slots / KV cache policy
  preset flipped to `wired`, with updated consumer/mechanism notes.
- `website/src/docs/pages/skippy-api.md`: regenerated via
  `python3 scripts/generate-skippy-api-doc.py` for the ABI version bump and
  new `SKIPPY_TRISTATE_*` constants; verified with `--check`.
- `crates/mesh-llm-config/src/wiring_status.rs`: manifest rows for the five
  wired paths flipped from `Unwired`/`Partial` to `Wired` with
  `WiringBehavior::None`; `mesh-llm-config`'s `website_docs_parity` and
  `wiring_status`/`wiring_validation` test suites confirm the two documents
  and the manifest all agree.


i386 (changes requested):
**Review: request changes** — one real correctness gap, otherwise strong.

**The gap: the KV-page cache identity does not cover the new KV-shape controls this PR introduces.** `skippy-cache/src/identity.rs::update_layout_identity` hashes cache types, flash-attn, `n_gpu_layers`, and device — and this PR adds `kv_offload`/`kv_unified`/`swa_full` to `StageConfig` and threads them to the native runtime without extending that hash or adding a `kv_unified_changes_page_identity`-style test. Concretely: `kv_unified = true` vs `false` changes `n_ctx_seq` partitioning in `llama-context.cpp` (unified → `n_ctx`, split → `n_ctx / n_seq_max`) and the number of attention streams per ubatch (`llama-graph.cpp`); `swa_full` changes the SWA cell layout. Two runs of the same model with different `kv_unified` settings but same cache types/FA/GPU split/device produce **identical page ids**, so an exported KV page from one can be imported under the other — silent numerical corruption, the exact class of bug `update_layout_identity`'s comment says it exists to prevent (it was added for flash-attn for this reason). Fix: fold the tri-state resolved values (or their `Option<bool>` debug forms) into the layout hash with distinct absent tags, plus identity tests for at least `kv_unified` and `swa_full` toggles, mirroring `flash_attention_changes_page_identity`.

Verified the native patch is otherwise correct against the pinned tree: tri-state resolution happens before the recurrent/hybrid `kv_unified = true` forcing at `model_loading.cpp:660`, so the forcing still wins — good. Minor observations, not blockers:
- The MTP draft attach path (`skippy_model_attach_mtp_draft_model`, ~line 948) has no recurrent forcing — an explicit `kv_unified=false` on a hybrid model would split the draft context while the trunk stays unified. Only reachable for hybrid + integrated MTP + explicit override; probably fine (draft context is single-lane), but worth a comment or a matching force for symmetry.
- Split/multi-node `stage_config` hardcodes `None` for the tri-states (no wire propagation) — consistent with the PR's stated scope, just noting the doc claim "wires `model_fit.kv_offload`…" is single-serve-only for now.

`cache_idle_slots` wiring (prewarm cap + pool-return discard with lane free) is correct and well-commented.

i386 (approved):
Verified at 88667a04b (stack head): tri-state kv_offload/kv_unified/swa_full lands in skippy_runtime_config and is resolved *after* the derived defaults in both skippy_finish_model_open and attach_mtp_draft_model, so auto preserves derivation and explicit values override. Recurrent/hybrid kv_unified=true force (model_loading.cpp ~660) executes after the tri-state insertion in finish_model_open, so the force still wins — the header comment's claim holds there.

One non-blocking nit: the comment on kv_unified says 'Recurrent/hybrid architectures still force this true regardless of the requested value' — that's true for contexts created via skippy_finish_model_open, but the second application site (skippy_model_attach_mtp_draft_model) has no recurrent/hybrid force at all, so an explicit false is *not* re-forced there. Draft-model contexts for recurrent/hybrid targets are presumably per-lane anyway, but the comment overstates slightly. Fine to fix wording in a follow-up.

Rust side ABI patch version matches (44 at stack head). `cargo test -p mesh-llm-host-runtime --lib inference::skippy::resolver` at head: 98 passed. cache_idle_slots capping via capped_target_idle_sessions/drop_session_timed is sound.

PR #1481, opened 2026-08-28
ci(llama): wire the agent repair loop into the upstream canary

i386:
## Summary

The llama.cpp upstream canary now repairs itself before failing: when the patch queue fails to apply onto new upstream — or the queue applies but a certification lane fails — the canary hands the failure to the agent repair loop (`scripts/llama-canary-agent-repair.sh`), which drives an `opencode` agent (default `zai-coding-plan/glm-5.3-flash`), re-runs the certification battery itself after every agent turn, and guarantees a repair PR on `llama-canary/patch-queue-fix` in every outcome.

Reviewers no longer see a dead-red canary that says "repair the queue in a normal reviewed pull request" — they see a repair PR whose description is agent-written (key upstream changes, how the patch queue evolved, risks) with a deterministic fallback.

## Why

The agent-repair machinery landed in #1436 but was never wired into `llama-upstream-canary.yml`: every version of the workflow in history had zero references to it, so run [33084228919](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33084228919) (patch `0001` conflicting in `src/llama-model.cpp/.h` at upstream `cb300598d`) failed with no repair attempt.

## What changed

- **Both failure modes route into the repair loop.** `patch-queue` mode on `prepare-llama.sh` apply failure; new `battery` mode when a certification lane fails after a clean apply. The wrapper resolves the upstream SHA itself (`git ls-remote`) when the workflow cannot supply one.
- **Model switch.** Default repair model is now `zai-coding-plan/glm-5.3-flash` (`CANARY_AGENT_MODEL` env; `LLAMA_CANARY_AGENT_MODEL` repository variable overrides it in the workflow).
- **Agent-written PR descriptions.** After each terminal outcome, one agent turn writes the repair-PR description (upstream changes between old pin and target, per-patch queue evolution, risks including ABI impact and newly failing/excluded lanes); a deterministic fallback body applies if the turn yields nothing.
- **Trust model preserved.** The canary job remains `contents: read` on the self-hosted runner; repair pushes/PRs use a dedicated `CANARY_REPAIR_TOKEN` fine-grained PAT (Contents + Pull requests: write). The battery still cannot be talked past — the wrapper's own battery run is the only green signal.
- **Red until merged.** The battery step no longer hard-fails the job before the repair loop can run, but a terminal `Fail when the canary needs human attention` step keeps the run red for any repair outcome: a green repair PR branch is not a green trusted-main certification until a human merges it. `update-pin` still fires only after a fully green main run.
- **Docs/skill synchronized.** `ci/ci.md`, the `manage-ci` inventory, the agent repair runbook, and the canary contract test now describe the wired contract (the old test pinned the interim no-wiring state with `assertNotIn("agent repair")`).

## Deployment prerequisites (external state, not in this PR)

1. Create the `CANARY_REPAIR_TOKEN` secret (fine-grained PAT: Contents read/write, Pull requests read/write, single-repo).
2. Install `opencode` on the family-certify runner image and provide agent credentials (`OPENCODE_API_KEY`/`NEMOTRON_API_KEY` secret, or `opencode auth login` in the runner profile).

Until those exist, the repair step reports the missing prerequisite and the canary fails as before — no behavior regression.

## Validation

- `python3 -m unittest scripts.tests.test_llama_upstream_canary_contract`: 15 passed (including the new wiring-contract test)
- `just ci-shellcheck scripts/llama-canary-agent-repair.sh`: clean
- `actionlint .github/workflows/llama-upstream-canary.yml`: clean
- `just ci-validate`: passes with the PR applied (the suite's 6 yaml-import errors on a bare local Python 3.14 are environmental — they reproduce on clean `42b46f30`; passing run used a venv with PyYAML)
- `bash -n` and shellcheck on the wrapper: clean

## Risks

- The repair loop adds up to ~12h of runner time on a failure day (bounded by `timeout-minutes: 720` per repair step).
- The agent-written PR description is unaudited content from the repair model; it is presentation-only (never gates anything) and the wrapper's fallback body covers silence.
- `CANARY_REPAIR_TOKEN` write scope is a new credential surface; it is scoped to this repository and only consumed by the repair steps, never by the certification or pin-update paths.

## Summary by CodeRabbit

* **New Features**
  * Added automated repair handling for patch-application and certification failures.
  * Repair runs can create or update a pull request with generated descriptions for review.
  * Added configurable repair-agent model selection.
  * Battery failures reuse captured certification evidence during repair.

* **Bug Fixes**
  * Canary runs remain failed until required repairs are merged.
  * Repair validation confirms tested changes match the proposed fix.

* **Documentation**
  * Updated CI documentation and repair guidance for the three-lane certification workflow.
  * Clarified raw-f32 activation-wire checks for certification.


i386 (commented):
Requesting changes on `79de7446b8abe0c5bf3363167b27cd3e43a46cf3`.

1. **[critical] Manual `upstream_sha` is shell-injected while the write PAT is present.** In `.github/workflows/llama-upstream-canary.yml:227-231`, `${{ github.event.inputs.upstream_sha || 'latest' }}` is substituted into the generated shell script before Bash parses it. A dispatch value containing a quote and shell syntax escapes the apparent quoting, and this step already has `CANARY_REPAIR_TOKEN`. The script's later 40-hex validation cannot protect code that executed before the script started. Put the expression in an `env:` value and pass `"$UPSTREAM_SHA"`; retain the script-side validation.

2. **[critical] A passing battery is not tied to the commit on the repair PR.** `scripts/llama-canary-agent-repair.sh:230-237` certifies the agent's local working tree and immediately reports success, but it never requires a clean tree or proves that local `HEAD`, `origin/llama-canary/patch-queue-fix`, and the PR head are the same commit. An agent can leave the actual fix uncommitted or fail to push it, so the wrapper can label a PR green even though the certified bytes are absent from that PR. Then `write_pr_body` launches another unrestricted agent turn *after* certification (`:152-175`, called at `:234`/`:253`), so the tree or branch can change after the only green run. Make the wrapper own or verify commit/push state, record the certified SHA, verify the remote PR head equals it, and generate the body without a repository-writing/token-bearing agent after certification (or assert the SHA/tree remains unchanged afterward).

3. **[high] The repair flow can exhaust its budget before making a repair.** Battery mode runs `run_battery` at `:225`, ignores its result, then enters the loop and immediately runs the same build+battery again at `:232` before the first `agent_turn`. With a battery allowed 720 minutes and the whole job capped at 780 minutes, a timeout/failure can consume nearly the entire job and leave no usable repair window. Reuse/capture the original workflow battery evidence, or after the one diagnostic run call the agent immediately and only certify after its repair; adjust the job budget to cover the intended number of attempts.

4. **[high] The write credential is inherited by every model-controlled command.** `scripts/llama-canary-agent-repair.sh:83` exports `GH_TOKEN` before `opencode` runs at `:98`, while the agent is explicitly asked to inspect upstream source, failure logs, and patch conflicts. This lets prompt-injected or simply erroneous model output push arbitrary repository content or mutate PRs with the PAT. Keep `CANARY_REPAIR_TOKEN` out of the agent environment and have the deterministic wrapper perform the narrow branch push/PR/comment operations after validating the tree and certified SHA.

There are also stale-state failure modes on the persistent runner: `BATTERY_LOG` is not cleared before `build-llama.sh`, so a build failure can feed a previous run's battery tail to the agent; `.deps/llama-canary-pr-body.md` is not cleared before the description turn, so a failed turn can reuse an old body. Clear/run-scope both artifacts and add behavioral wrapper tests with stubbed `opencode`, battery, git, and `gh`; the current 15 tests only assert YAML strings and do not exercise these state transitions.

Local verification on this exact head: the 15 Python contract tests pass, ShellCheck is clean, Actionlint is clean, and `git diff --check` is clean. Those checks do not cover the runtime defects above.

PR #1488, opened 2026-08-28
config: complete host wiring for aliases, plugins, and prompt-shape metrics

ndizazzo:
## Summary

Closes the remaining host-integration gaps from the PR8 configuration audit. Configured model identities stay canonical through direct-local serving, effective defaults participate in collision checks, plugin lifecycle and remote-control behavior match schema contracts, StageLoad preserves supported runtime controls, and response usage feeds privacy-bounded prompt-shape telemetry.

## Configuration and runtime behavior

- Served aliases are compared after trimming, including aliases inherited through defaults-expanded profiles.
- Unsupported plugin control URLs, every audited silent no-op, and the unsupported mmap prefetch/buffer controls fail static validation with canonical paths.
- Named-model response usage records prompt and completion token counts without exporting model or content values.
- Plugin-owned disabled startup responses remain successful disabled states.
- Direct-local serving uses canonical config model IDs and configured survey telemetry.
- StageLoad transports exactly 11 supported runtime settings. `mmap` remains a separate real load setting; `use_mmap_prefetch` and `use_mmap_buffer` are parsed only so they can be rejected before resolution.

## Commit structure

- RED: `d257218e8 test(config): prove host integration wiring gaps`
- GREEN: `6ee60c671 fix(config): close host integration wiring gaps`
- Parent: `75931d91d fix(skippy): complete cache identity stage config fixtures`

The RED commit compiles and fails the 11-field StageLoad preservation proof. The standalone RED proof files are byte-identical in GREEN, which passes the same contracts.

## Validation

- `cargo fmt --all --check`
- `cargo test -p mesh-llm-config`: 169 unit, 4 exact-head, 4 topology, and 9 wiring tests passed.
- `cargo test -p mesh-llm-host-runtime --lib`: 2,755 passed, 8 ignored.
- Warning-denying Clippy passed for `mesh-llm-config`, `mesh-llm-host-runtime`, and `mesh-llm`.
- `cargo check -p mesh-llm`
- `just ci-validate`: 649 script tests passed, 7 skipped; all repository consistency checks passed.
- `just build`: host, UI, patched Metal runtime, and composed development product built successfully.
- Manual CLI smoke: `--help` and `--version` succeeded; an invalid option returned the expected Clap error and usage hint.

## Compatibility

- StageLoad runtime settings use additive protobuf field 43. Legacy decoders ignore it; current decoders default it when absent.
- No Skippy ABI change.
- No new CLI flags or settings.
- No compatibility shim or external serving lane.
[...]

PR #1508, opened 2026-08-28
ci(llama): certify the multimodal path with a real mmproj smoke lane

i386:
## Summary

The llama canary certified the two vision families (`qwen2-vl`, `qwen3-vl`) **text-only**: the certification manifest pinned no projector, no image ever went through the runtime, and the env-gated real-multimodal harness in `skippy-server` (`SKIPPY_MM_*`) ran nowhere in CI. Upstream breakage in the mtmd/vision path would pass the canary today.

This adds real multimodal coverage to the battery, stacked on #1488 (stack top).

## What it does

- **Policy**: optional `mmproj_artifact` on a family (same immutable artifact shape as `draft_artifact`). Planner validates it and verifies the exact cache blob identity/size, but exempts the projector sidecar from trunk-dimension (`*.block_count`/`*.embedding_length`) checks — a projector is an encoder sidecar, not a trunk model. Exactly one projector file per entry.
- **Manifest**: pins `mmproj-Qwen2-VL-2B-Instruct-f16.gguf` (bartowski, 1,331,656,192 bytes) and `mmproj-F16.gguf` (unsloth, 1,159,030,336 bytes) at the same immutable revisions already pinned for the VL trunks.
- **Battery**: after each family's core certification lanes, a `mmproj-smoke` lane runs the *existing* real-model harness `crates/skippy-server/src/frontend/tests/multimodal.rs` (local monolithic **and** split stages — projector loads on stage 0) against a deterministic, checked-in image (`ci/llama-canary/fixtures/multimodal-smoke.png`, 18 KB; SHA-256 pinned by test), with `cert_timeout_for_startup`-derived wall-clock bound. Outcomes are recorded in `results.jsonl`/summary like any lane; planned-vs-executed reconciliation covers the smokes (`battery(mmproj-planned-vs-executed)`).
- **Docs**: `ci/ci.md` battery row updated.

## Tests

- `scripts/tests/test_plan_family_battery.py`: mmproj artifacts resolve and exactly cover the vision families (new).
- `scripts/tests/test_llama_upstream_canary_contract.py`: smoke lane runs only for families with a projector, exactly once, correct env/test invocation, reconciles in dry-run (new); deterministic image digest pinned (new).
- Full local verification: 29/29 script tests green; full-manifest battery dry-run green (125/125 planned, both smokes emitted with correct env); `bash -n` clean.

## Deployment note (runner cache)

The two projector blobs (~2.3 GiB total) must be added to the `family-certify` runner's pre-warmed HF cache before the next `llama-bump` run; the planner fails closed on a missing blob exactly like any other pinned artifact.

Part of #1462.

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable sampling, chat templates, grammar, JSON schema, reasoning, and extended speculative decoding controls.
  * Added multimodal projector configuration, image token limits, media markers, batching, and deterministic smoke validation.
  * Added configurable split topologies, hardware/runtime controls, lifecycle timeouts, KV-cache behavior, and cache idle limits.
  * Added prompt-shape telemetry histograms with model identifiers excluded from exported attributes.
  * Added served aliases, profile-aware model selection, direct projector downloads, and optional plugin startup behavior.

* **Bug Fixes**
  * Improved validation, error propagation, URL redaction, protocol compatibility, and plugin failure handling.


i386 (commented):
Approved — I found no correctness issue in the stacked diff (`bc501d885` → `522b8de8e`).

I checked the new path end to end:

- `mmproj_artifact` uses the same immutable repo/revision/file/size/blob identity contract as the existing artifacts, while the GGUF trunk-dimension exemption is limited to projector sidecars.
- Both VL rows pin one projector at the trunk's exact revision.
- Preflight resolves the immutable projector before any certification starts.
- The added lane runs the existing real-model `skippy-server` multimodal module with both local-monolithic and split-stage tests, a SHA-pinned checked-in image, bounded timeout, and serialized execution.
- Projector lanes have their own planned/executed counter and appear in `results.jsonl`/the summary without weakening the existing 125-certification reconciliation.

Local evidence on the exact head:

- 29/29 focused planner and canary-contract tests passed.
- `bash -n scripts/skippy-family-battery.sh` passed.
- Full dry-run reconciled 125/125 core certifications and emitted exactly the two expected multimodal smoke commands.
- The manifest's only semantic changes are the two projector records and their notes; the rest of the large JSON diff is formatting.

Quality and Website are green; the remaining platform CI was still running when I submitted this review.

scripts/skippy-family-battery.sh
```diff
@@ -806,9 +827,57 @@ planned_certification_count() {
...
+      SKIPPY_MM_MAX_TOKENS=16 \
+      SKIPPY_MM_N_GPU_LAYERS=999 \
+      SKIPPY_MM_SPLIT_LAYER="$(( layer_end / 2 ))" \
+      LLAMA_STAGE_BACKEND=metal \
+      cargo test --manifest-path "$ROOT/Cargo.toml" -p skippy-server --lib frontend::tests::multimodal -- --nocapture --test-threads=1 \
+      >"$log_path" 2>&1 || exit_code=$?
+  if (( exit_code != 0 )); then
+    echo "mmproj smoke failed for $family; log: $log_path" >&2
+    FAILURES+=("$family@mmproj")
+    CERT_FAILURE_COUNT=$((CERT_FAILURE_COUNT + 1))
```
i386: Non-blocking evidence bug: this increments `CERT_FAILURE_COUNT` for an mmproj smoke failure, but `TOTAL` counts core certifications only and the final line prints `TOTAL - CERT_FAILURE_COUNT` out of `TOTAL` as ‘certifications passed’. With 125/125 core lanes green plus one failed multimodal smoke, the console would therefore report 124/125 certifications passed. Please track multimodal smoke failures separately (or stop using the core counter for that final value); a failure-path contract test would make this hard to regress.

i386:
CI follow-up: the completed Linux run is red on two `StageConfig` test initializers in `crates/skippy-cache/src/identity.rs` (lines 229 and 447): both are missing `use_mmap_prefetch` and `use_mmap_buffer`.

This is inherited from the stacked base, not introduced by #1508: base commit `bc501d885` adds those fields to `StageConfig`, while this PR's eight-file diff does not touch either `skippy-protocol` or `skippy-cache`. The remaining Rust matrix jobs and downstream Windows/macOS jobs were cancelled by fail-fast after that compile error. The base needs those two fixtures completed (then this branch needs the updated base) before CI can go green.

i386 (commented):
## Review at replayed head `d86e16d3a` — no correctness findings in this PR's diff

Reviewed the single replayed commit against its base `85fea61c4` (8 files, +180/−8). The old→new head diff is dominated by the force-updated base's whole-manifest reformat; the feature delta itself is clean.

**Verified locally at this head:**
- Planner suite 13/13 and canary contract suite 18/18 (including the three new mmproj tests).
- Full-manifest dry-run reconciles 125/125 and emits exactly the two expected smoke lanes: qwen2-vl (`SKIPPY_MM_ACTIVATION_WIDTH=1536`, `SKIPPY_MM_SPLIT_LAYER=14`) and qwen3-vl (4096/18). Both split layers sit inside the harness's validated `1..layer_end` range (28/36), and both widths match policy.
- `bash -n` clean; fixture `multimodal-smoke.png` SHA-256 matches the digest pinned in `test_mmproj_smoke_image_fixture_is_deterministic` (`308ff692…a5034`).
- Fail-closed behavior confirmed in the harness itself: a missing model/projector/image file makes the test `bail!`, so a cache miss on the runner cannot silently skip the smoke.
- The accounting fix is in and enforced: a failed smoke increments `MM_SMOKE_FAILURE_COUNT` (no `CERT_FAILURE_COUNT` inside `run_mmproj_smoke`), and the planned-vs-executed reconciliation covers the smokes separately.

**Current CI red is inherited from the base, not this PR.** Quality and Website are green on `d86e16d3a`. Linux `Rust tests (batch-0)` failed with 5 failures in `crates/mesh-llm-commands/src/gpus/tune/` (`hardware.fit_target_mib is not supported: Offline GPU tuner reads it; live serving placement ignores it`); the `pr-cancel-sibling-runs` monitor then cancelled the macOS/Windows lanes and sibling batches as designed. This PR's commit touches zero `crates/` files. The rejection was introduced by base-stack commit `248a648b7` ("derive static validation from an exhaustive schema-status manifest"), and base PR #1488 is red at the exact base SHA `85fea61c4` with the identical batch-0 signature. The base needs the `fit_target_mib` test fixtures or manifest status fixed; once it lands, this branch needs the updated base and CI should go green.

(Approval cannot be recorded — the PR author is the shared `i386` account, same credential as this review.)

i386 (commented):
Full review of the stack at 88667a04b (can't formally approve: this PR is authored by the same GitHub account I'm reviewing from — i386 — and GitHub rejects self-approval; every underlying PR by @ndizazzo is approved).

Verified locally:

- `pytest scripts/tests/test_plan_family_battery.py scripts/tests/test_llama_upstream_canary_contract.py`: 31 passed + 16 subtests (fresh venv, not trusting the PR description).
- Full battery dry-run: 125/125 planned certifications reconciled, both mmproj smokes emitted with correct env (SKIPPY_MM_MODEL/PROJECTOR/IMAGE/ACTIVATION_WIDTH/SPLIT_LAYER, ctx 2048, 16 max tokens, 999 gpu layers, metal backend).
- Image digest pinned by test matches the checked-in PNG (sha256 308ff692...a5034).
- Manifest field-count arithmetic: RESOLVED_MANIFEST header is 22 columns, awk counts $19 (mmproj_repo) non-empty — correct; smoke-failure counter increments before the dry-run early-return so dry-run reconciliation works.
- Contract tests prove the smoke lane runs only for projector families, exactly once, with correct env/test invocation and planned-vs-executed reconciliation.

Three notes, none blocking:

1. **Base drift**: this stack forked from main at d0abb9e4b; cfg-1462/pr0-docs has since absorbed #1509 (Qwen4Exp patches 0027–0030), #1513 (Flash-Next cert), #1507, and the pr0 docs commits — none of which are in this stack. The two sides don't collide (different files; only the stack's patches touch the ABI version line, and 0027 doesn't touch model_loading.cpp/runtime.h), but the merge into the base branch will need the whole plan re-run since the patch digest changes. Flagging so nobody is surprised.

2. **LLAMA_STAGE_BACKEND=metal hardcoded** in run_mmproj_smoke (line 861) — fine while family-certify runners are macOS-only, and consistent with the existing lanes, but it will bite if the battery ever runs on Linux/CUDA.

3. **Patch numbering**: the queue now has two 0024-* and two 0025-* patches (kv-offload vs nemotron-MTP; sampling vs NextN). Lexical sort makes application deterministic, so this is hygiene-only, but a renumber pass would help future archaeology.

Runner-cache precondition from the PR description stands: the two projector blobs (~2.3 GiB) must be pre-warmed on family-certify before the next llama-bump run; the planner fails closed on a missing blob, which is the right behavior.

PR #1514, opened 2026-08-28
feat(skippy): add remote KV handoff and phase placement

danielwinterw:
The local L3 substrate now lives in #1632 and its follow-up stack, so this draft contains only the remote/disaggregated layer: authenticated-transport-ready `skippy-kv/1` peer fetch, cost-based prefill/decode phase placement, and a transactional remote handoff harness that can stream pages while later prompt chunks compute.

The receiver stages and verifies segments incrementally, checks ordered tiling and the canonical payload digest at commit, and never generates from a partial transfer. The plain-TCP server remains a lab harness with an explicit unauthenticated-listener warning; production mesh exposure is reserved for the registered ALPN with membership authentication.

The handoff identity includes the served artifact content digest, so two different local model files behind one display model ID cannot share state. The focused branch also carries the two-machine runbook and sweep driver.

## Stack

This is the draft top of the `kvcache-ng` stack:

1. #1632 — node-local durable L3
2. #1707 — recurrent prefill default
3. #1710 — restart replay harness
4. #1714 — resident warm-path repair
5. #1726 — packed storage and shutdown drain
6. #1736 — direct packed restore assembly
7. this PR — remote handoff and phase placement
8. #1712 — platform qualification catalog activation

The older duplicate L3 implementation and superseded warmup script were removed from this draft; they are inherited from the reviewed local stack and #1710 instead.

## Validation

- `cargo test -p skippy-cache --lib` — 128 passed, 1 ignored
- `cargo test -p skippy-correctness` — 37 unit/integration tests and 84 manifest tests passed; model-download tests remain ignored by design
- `cargo clippy -p skippy-cache -p skippy-correctness --all-targets -- -D warnings`
- `cargo test -p skippy-cache --lib eviction_cost -- --ignored --nocapture` — passed at 20 manifests × 9,504 segment references
- `cargo run -p xtask -- repo-consistency no-console-print`
- `git diff --check`

The review-hardening changes are at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`; the exact PR head adds a CLI regression test at `e7623f69c2b6fc8b9af9140240ec05aaa943c8f7`.

This remains a draft until authenticated iroh transport replaces the unauthenticated lab TCP path and the two-machine remote workflow is rerun on the complete current stack.


i386 (commented):
Read the full diff line-by-line (skippy-cache l3/tier/l3_remote/identity, kv_integration wiring, the harness, phase_placement) and ran the pure crates locally: skippy-cache 65 + skippy-topology 42 tests green at 4f42615.

**Overall**: this is a strong slice. The completeness gate (nothing loadable until commit validates tiling + payload digest), idempotent content-addressed puts, the numerical-vs-placement identity split, and the two-phase-commit streaming receiver are all done right, and the test suite exercises the failure modes that matter (corruption, eviction, identity mismatch, out-of-order pages). Notes below are real but none block the draft — three of them I'd want addressed before this graduates from harness to serving.

**Before serving (P1):**
1. `exact_state_identity` doesn't hash weight content digests — see inline on identity.rs. The doc comment on `update_weight_identity` makes exactly this argument for `prefix_hash`; the new function should get the same protection.
2. Unbounded disk when `SKIPPY_L3_BUDGET_BYTES` is unset — inline on config.rs. The radix-channel acceptance criteria explicitly call for a capped default.
3. `serve_connection`/`serve_store` have no auth — fine as a lab harness over TCP; must ride the iroh `skippy-kv/1` ALPN with mesh-membership auth before it's ever reachable off-loopback.

**Performance-shaped (P2, lab-tolerable):** whole-payload `assemble()` materializes the full state in RAM (twice for kv-recurrent fills), and `enforce_budget` is an O(n²)-ish full-directory rescan per evicted manifest at capacity.

**Nits:** prefix-link pruning on any error; receiver has no read timeout after the handshake.

Happy to re-review after the two-machine matrix runs.

crates/skippy-cache/src/identity.rs
```diff
@@ -198,6 +198,78 @@ fn prefix_namespace_hasher(
...
+    pub lane_count: u32,
+    pub payload_kind: &'a str,
+}
+
+pub fn exact_state_identity(params: &ExactStateIdentityParams<'_>) -> String {
+    let mut hasher = blake3::Hasher::new();
+    hasher.update(b"exact-state-identity-v1");
+    hasher.update(NATIVE_KV_RUNTIME_ABI_VERSION.as_bytes());
+    hasher.update(b"model:");
+    hasher.update(params.model_id.as_bytes());
```
i386, line 365: `exact_state_identity` hashes `model_id`/revision/file but **not** weight content digests — `manifest_sha256`/`source_model_sha256`/`package_ref` don't exist in `ExactStateIdentityParams`, while `update_weight_identity` (added in this same PR for `prefix_hash`) makes exactly this argument: two runs can present the same `model_id` while serving different tensors (requantized artifact, republished layer package, swapped local GGUF behind the same path). A state blob crossing that boundary is silent numerical corruption, and unlike the radix path there's no second guard here.

In the harness today the hole is narrow (HF revisions are content-addressed; you'd need same-name-different-bytes local files on the two ends), but this function is the identity contract the serving handoff inherits — please add the optional digest fields (with the `<absent>` tagging trick) before it's reused outside the harness.

crates/skippy-cache/src/l3.rs
```diff
@@ -0,0 +1,539 @@
...
+            let modified = entry.metadata()?.modified()?;
+            entries.push((modified, stem));
+        }
+        entries.sort_by_key(|entry| std::cmp::Reverse(entry.0));
+        Ok(entries.into_iter().map(|(_, stem)| stem).collect())
+    }
+
+    /// Assemble the full payload for a manifest, verifying every segment
+    /// digest, the tiling, and the whole-payload digest.
+    pub fn assemble(&self, manifest: &HandoffManifest) -> Result<Vec<u8>> {
```
i386, line 1146: `assemble` materializes the entire payload as one `Vec` — fine at harness scale, but `L3Tier::fill` runs this on a radix-miss serving path, and full-state payloads are exactly what `MAX_STAGE_STATE_IMPORT_BYTES` (512 MiB) was guarding. For kv-recurrent fills the split in `tier.rs` (`wire[kv_bytes..].to_vec()` + truncate) adds a second copy, so peak is ~2× state bytes plus whatever `import_*` copies internally.

Not urgent — it's a miss path replacing a full prefill — but worth either a streaming assemble (read segment + feed import incrementally; page-stream manifests already prove the shape works) or a comment stating the memory bound. Serving-side fill metrics (`fill_ms` vs import) would also let the radix channel set the restore-cost thresholds they need.

crates/skippy-cache/src/l3.rs
```diff
@@ -0,0 +1,539 @@
...
+        for entry in fs::read_dir(self.root.join(SEGMENT_DIR))? {
+            total = total.saturating_add(entry?.metadata()?.len());
+        }
+        Ok(total)
+    }
+
+    /// Evict oldest manifests (never the newest) and collect unreferenced
+    /// segments until the segment footprint fits the budget. Returns bytes
+    /// freed.
+    pub fn enforce_budget(&self) -> Result<u64> {
```
i386, line 1404: At capacity this is quadratic-ish per commit: each eviction iteration does `segment_footprint_bytes` (readdir+stat every segment) + `list_manifests` (readdir+stat+sort) + `collect_unreferenced_segments` (parse **every** manifest JSON, then readdir every segment again), inside the `loop`.

Off the request path (worker-serialized) so it won't add latency, but at a busy serving node it can wedge the spill pipeline for a while. Since `L3Tier` is long-lived, consider keeping a live referenced-digest set in memory (updated on commit/evict) and amortizing GC. Also consider a WARN when the newest manifest alone exceeds budget — currently silently allowed by design, but invisible.

crates/skippy-cache/src/l3.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// evicted manifests.
+    pub fn manifest_for_prefix(&self, prefix_key: &str) -> Result<Option<HandoffManifest>> {
+        let path = self.prefix_path(prefix_key);
+        let Ok(bytes) = fs::read(&path) else {
+            return Ok(None);
+        };
+        let payload_digest = String::from_utf8(bytes).context("malformed prefix link")?;
+        match self.load_manifest(&payload_digest) {
+            Ok(manifest) => Ok(Some(manifest)),
+            Err(_) => {
```
i386: Nit: this `Err(_)` arm prunes the prefix link on **any** failure — including transient I/O errors — not just "manifest was evicted." A temporarily-unreadable disk would delete every prefix link it touches (segments/manifests survive until GC so it self-heals, but fills miss until re-spill).

Suggest downcasting to `std::io::Error` and pruning only on `ErrorKind::NotFound` (plus version mismatch if you want upgrades to self-invalidate); leave real I/O errors for retry.

crates/skippy-server/src/kv_integration/config.rs
```diff
@@ -114,17 +117,57 @@ impl KvStageIntegration {
...
+    let Ok(root) = std::env::var("SKIPPY_L3_DIR") else {
+        return Ok(None);
+    };
+    if root.trim().is_empty() {
+        return Ok(None);
+    }
+    let budget_bytes = std::env::var("SKIPPY_L3_BUDGET_BYTES")
+        .ok()
+        .and_then(|value| value.parse().ok())
+        .unwrap_or(0);
```
i386: `SKIPPY_L3_BUDGET_BYTES` unset → `unwrap_or(0)` → budget disabled → unbounded disk growth. The radix-channel acceptance criteria explicitly ask for "capped storage on disk (with a sensible default)". Suggest a finite default when `SKIPPY_L3_DIR` is set (say 32 GiB) and treat 0 as explicit opt-in-unbounded (or refuse). Also worth logging the effective budget at tier open.

crates/skippy-cache/src/l3_remote.rs
```diff
@@ -0,0 +1,409 @@
...
+/// Statistics for one `fetch_into_store` pull.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct FetchStats {
+    pub segments_fetched: usize,
+    pub segments_skipped: usize,
+    pub bytes_fetched: u64,
+}
+
+/// Serve the store to one connected peer until it disconnects.
+pub fn serve_connection(store: &HandoffSegmentStore, stream: TcpStream) -> Result<()> {
```
i386, line 82: Flagging for the record: `serve_connection`/`serve_store` have zero authentication — any process that can reach the port can enumerate and drain the store. Digest verification protects the *client* from a malicious peer, not this server from disclosure. Fine for the lab harness over loopback/Tailscale, but this must not be wired to a plain TCP listener in the mesh — only the iroh `skippy-kv/1` ALPN with mesh-membership auth (as the PR description already sequences). Maybe say so in the module doc so nobody copies the TCP pattern forward.

crates/skippy-correctness/src/runner/remote_handoff.rs
```diff
@@ -0,0 +1,1824 @@
...
+    model_identity: &model_artifact::ModelIdentity,
+    model_load_ms: f64,
+    stream: TcpStream,
+    report_out: Option<&std::path::Path>,
+    store: Option<&HandoffSegmentStore>,
+    local_state_identity: &str,
+) -> Result<bool> {
+    stream.set_nodelay(true).ok();
+    stream
+        .set_read_timeout(Some(Duration::from_secs(args.handshake_timeout_secs)))
```
i386, line 1081: Nit: `set_read_timeout` only covers the HELLO frame; after the handshake a sender that stalls mid-stream blocks this receiver connection forever (the `accept` loop is serial per connection). Harness-only, but `--accept-count` sweeps would benefit from applying the timeout to every `read_frame`, or at least resetting a larger transfer deadline post-handshake.

crates/skippy-server/src/kv_integration/exact_state.rs
```diff
@@ -253,6 +260,105 @@ impl KvStageIntegration {
...
+        runtime: &mut RuntimeState,
+        session_id: &str,
+        identity: &PrefillKvIdentity,
+        lookup_started: Instant,
+    ) -> Result<Option<ExactStateRestore>> {
+        let Some(l3) = &self.l3 else {
+            return Ok(None);
+        };
+        let prefix_key = skippy_cache::l3_prefix_key(&identity.namespace, &identity.token_ids);
+        let filled = match l3.fill(&prefix_key) {
```
i386: This `fill` runs on the request path (radix miss during prefill): full `assemble` = disk read + BLAKE3 over every segment + whole-payload digest, then import. For a multi-hundred-MiB state that's real seconds of hashing before the import starts — correct, but it should be measured before we trust L3 restore vs. cold prefill on the same node (the radix channel needs exactly this number for their restore-cost thresholds). The `lookup_ms` bucket here includes all of it — worth splitting `fill_ms` (store) from import so the counters are separable.

i386 (commented):
Inline notes for the summary above (previous submission dropped these — resubmitting with the correct side field).

danielwinterw:
Pushed four commits addressing the full gap list from the review (thanks — every catch was real). Mapping back to the numbered gaps:

**1. Dense models never reach L3** — with the tier enabled, families that would use borrow-only `ResidentKv` (gemma/qwen/llama/glm) now record KV pages instead (`KvRecurrent` with a legitimately empty recurrent snapshot; recurrent export/import tolerate no-recurrent runtimes and finalize position directly). No payload override needed — the default path for exactly the acceptance-criteria models reaches disk.

**2. No partial-prefix restore** — the prefix index is now per-namespace and length-aware, and misses probe recorded lengths longest-first, hashing the query's own leading tokens at each (up to 64 probes). An entry recorded at length k is a complete state for prefix k, so multi-turn prompts reuse the longest earlier turn they extend — mirroring the radix's `longest_component_prefix` — and divergent same-length prompts miss. The existing grid-checkpoint records supply the shared-prefix rungs.

**3. Concurrent fills not single-flighted** — fills claim a per-prefix inflight key; the loser skips L3 and prefills normally (no duplicated disk loads) while the winner re-warms the radix for everyone.

**4. No measurement program** — `scripts/l3_warmup_bench.py` drives a real `mesh-llm serve` and measures TTFT for cold / warm (radix) / multi-turn / restart (L3 fill) / two concurrent same-prefix requests, with `scripts/generate-agent-prefix.py` as a stand-in until captured opencode/goose/pi/buzz prefixes run on the lab. Validated end-to-end on Qwen3-0.6B with a ~1.9K-token prefix: warm 2.2x, multi-turn 2.2x, and after a full restart a divergent-tail request reported `cached_tokens: 1664` of 1857 — the shared-checkpoint entry filled from disk and only the suffix prefilled. On this tiny model the fill (~100 MB read+verify+import) roughly breaks even with the saved prefill; the economics the criteria care about (E4B–27B, 19K prefixes) are what the bench measures on real hardware.

**5. Store performance** — budget enforcement is now single-pass (one footprint scan, one manifest scan, one in-memory reference map) regardless of evictions, with a WARN when the newest manifest alone exceeds budget; kv-recurrent fills split the assembled payload in place (no second copy) and the memory bound is documented. Streaming assemble stays future work — the page-stream manifest shape already proves it.

**6. Defaults/budget** — unset `SKIPPY_L3_BUDGET_BYTES` now means 32 GiB; `0` is an explicit unbounded opt-in; the tier logs its effective budget and restorable inventory (manifests/tokens/bytes) at open. Startup stays on-demand per the criteria ("mmaped/similar in on demand") — nothing preloads, fill-on-miss restores.

**7. Routing evidence** — `ExactStateRestore` carries `source` (radix|l3) and `fill_ms` (store cost split from import), exported as `skippy.exact_cache.source` / `skippy.exact_cache.l3_fill_ms`; the startup inventory gives the restorable-token count. Happy to shape a fuller attrs schema with #skippy-radix-affinity once L1-hit-only lands.

**8. Fetch auth** — module doc now states plainly the TCP server has no auth and mesh exposure rides the `skippy-kv/1` iroh ALPN only.

Code-review threads: weight content digests joined `exact_state_identity` (with absent-tagging and tests), prefix links prune only on `NotFound`, and the review's remaining nits are covered above. 760 tests green across the touched crates.

crates/skippy-cache/src/identity.rs
danielwinterw: Added in d3914138e's cache commit (aed130498): ExactStateIdentityParams now carries manifest_sha256 / source_model_sha256 / package_ref with the same absent-tagging, plus a test asserting a requantized or repacked artifact behind the same model_id changes identity.

crates/skippy-cache/src/l3.rs
danielwinterw: Took the documented-bound option plus the cheap wins (aed130498): fill now splits the assembled payload in place via split_off (no second copy) and the memory bound is stated at the call site; fill_ms is split from import in ExactStateRestore and exported as skippy.exact_cache.l3_fill_ms. Streaming assemble deferred — the page-stream manifest shape already proves the mechanics when it's needed.

crates/skippy-cache/src/l3.rs
danielwinterw: Rewritten single-pass in aed130498: one footprint scan, one manifest scan, one in-memory reference map, then eviction runs against the in-memory model with a single GC at the end. Also warns when the newest manifest alone exceeds the budget.

crates/skippy-cache/src/l3.rs
danielwinterw: Fixed in aed130498: reads distinguish ErrorKind::NotFound (prune) from transient I/O errors (surfaced, link kept). Applies to both the link and the manifest read.

crates/skippy-server/src/kv_integration/config.rs
danielwinterw: Fixed in d3914138e: unset now defaults to 32 GiB, 0 is an explicit unbounded opt-in, and the tier logs the effective budget plus its restorable inventory (manifests/tokens/bytes) at open.

crates/skippy-cache/src/l3_remote.rs
danielwinterw: Module doc now states it plainly (aed130498): no server-side auth, digest verification protects only the client, plain TCP is lab-harness-only, and mesh exposure rides the skippy-kv/1 iroh ALPN with mesh-membership auth.

crates/skippy-correctness/src/runner/remote_handoff.rs
danielwinterw: set_read_timeout is a socket property that stays in force for every subsequent read on the connection, so a mid-stream stall errors out after the timeout (600s default) rather than wedging the accept loop — the flag's handshake-y name invited the misread, so the latest commit adds a comment and help text saying it covers the whole connection.

crates/skippy-server/src/kv_integration/exact_state.rs
danielwinterw: Addressed in d3914138e: fill_ms is measured separately from lookup/import and exported as skippy.exact_cache.l3_fill_ms with skippy.exact_cache.source, fills are single-flighted per prefix, and we measured end-to-end — on Qwen3-0.6B a ~100 MB fill roughly breaks even with recompute (cached_tokens: 1664/1857 after restart), which is exactly the number the restore-cost threshold should come from once the lab runs the target models.

i386 (approved):
Round 2 at c009998c7 — all eight threads from my first pass are resolved, and the four serving gaps from the #skippy-radix-L1-L2-L3 analysis are closed (dense-path decision, partial-prefix restore, fill single-flight, benchmark program). Approving.

Verified locally at c009998c7: skippy-cache 69/69, skippy-topology 42/42, skippy-server kv_integration 45/45 — including `exact_records_write_through_to_l3_and_survive_radix_eviction`, which now also covers a longer-query fill.

What I checked beyond the thread replies:

- **Longest-prefix restore consumes correctly.** `fill_longest` re-warms the radix at the truncated token path (`token_ids[..token_count]`), so the RAM tier matches what was actually restored, and the caller prefills only `prefill_tokens[restored_prefill_tokens..]`. Divergent tokens at a recorded length miss (tested), and the per-namespace length index is keyed by the query's own leading tokens, so a recorded entry only matches a genuine prefix.
- **The dense-family flip is complete.** `ResidentKv → KvRecurrent` under L3 is paired with empty-recurrent handling at every import site (`set_session_position` on restore, `is_recurrent_unavailable`-tolerant export), so gemma/qwen reach disk without a payload override.
- **Weight digests are in both identities.** `exact_state_identity` now tags manifest/source/package digests with absent-markers (test: requantized/repacked artifacts behind the same model_id change identity); the serving L3 path was already covered via `prefix_namespace_hash → update_weight_identity`.
- **Budget enforcement is single-pass** with an in-memory refcount map and one final GC; the newest-oversize case warns instead of evicting into unloadability. Transient I/O errors no longer masquerade as absence — only `NotFound` prunes links.
- **Single-flight fills** key on `(namespace, len)`; the loser returns `None` and prefills normally. Restart inventory (`restorable_summary`) reads manifests only, no payload I/O.

Non-blocking observations:

- `fill_longest` caps at `MAX_PREFIX_PROBES = 64` recorded lengths per namespace. With hundreds of distinct recorded lengths the probe can miss older usable prefixes — reasonable bounded-latency default, just worth remembering if bench data ever shows a surprising miss.
- Unrelated to the PR but hit while testing: on this macOS box, CLT 27 ships a near-empty `/Library/Developer/CommandLineTools/usr/include/c++/v1` that shadows the SDK's full copy (breaks any C++ build), and `skippy-ffi/build.rs` emits no OpenMP link line on Apple targets (only the Linux/Windows branches call `openmp_libs`), so a libomp-enabled build fails to link. Both are pre-existing; happy to file follow-up issues.

Left for follow-ups per the plan doc, correctly: iroh-ALPN mesh exposure for `skippy-kv/1`, streaming assemble, and the restore-cost threshold once more bench numbers exist.

crates/skippy-correctness/src/runner/remote_handoff.rs
```diff
@@ -0,0 +1,1830 @@
...
+/// endianness, pointer width) is the cross-machine guard here.
+fn state_identity_for(
+    args: &RemoteHandoffArgs,
+    identity: &model_artifact::ModelIdentity,
+) -> String {
+    exact_state_identity(&ExactStateIdentityParams {
+        model_id: &identity.model_id,
+        model_revision: identity.source_revision.as_deref(),
+        model_file: identity.source_file.as_deref(),
+        manifest_sha256: None,
```
i386, line 290: **P1 — the original weight-content identity gap is still live at the only non-test caller.** The hasher now accepts the right fields, but this harness always passes `None` for all three, so two different local GGUFs invoked with the same `--model-id` still produce the same handoff identity. `runtime_model_identity` explicitly permits that local-path mode, and this crate already has `source_model_sha256()` in `split_prefix_hit.rs`; please resolve/hash the actual served artifact (and package manifest when applicable) and feed those values here. A unit test that mutates file contents behind a stable model id would keep the fix connected rather than only testing the standalone hasher.

crates/skippy-server/src/kv_integration/exact_state.rs
```diff
@@ -185,7 +200,13 @@ impl KvStageIntegration {
                     Err(error) if is_native_kv_unavailable(&error) => None,
                     Err(error) => return Err(error),
                 };
-                let recurrent = runtime.export_recurrent_state(session_id)?;
+                // Dense families under the L3 payload flip have no
+                // recurrent memory; their snapshot is legitimately empty.
+                let recurrent = match runtime.export_recurrent_state(session_id) {
+                    Ok(recurrent) => recurrent,
+                    Err(error) if is_recurrent_unavailable(&error) => Vec::new(),
```
i386: **P1 — do not admit an empty dense checkpoint when KV export is unavailable.** The new L3 flip turns dense `ResidentKv` families into `KvRecurrent`. If `export_kv_page` hits either tolerated unavailable case and `export_recurrent_state` reports the expected `no recurrent memory`, this constructs `(kv=[], recurrent=[])`, spills it as a valid prefix, and restore later only calls `set_session_position(token_count)`. That advances `n_past` without restoring any attention KV bytes, so generation can proceed from missing state. Pure-recurrent models may legitimately omit KV, but dense models may not; require at least one state component appropriate to the family (or fall back to ResidentKv/skip the record) and add the dense no-native-KV regression case.

crates/skippy-server/src/kv_integration/exact_state.rs
```diff
@@ -253,6 +274,148 @@ impl KvStageIntegration {
...
+        &self,
+        runtime: &mut RuntimeState,
+        session_id: &str,
+        identity: &PrefillKvIdentity,
+        lookup_started: Instant,
+    ) -> Result<Option<ExactStateRestore>> {
+        let Some(l3) = &self.l3 else {
+            return Ok(None);
+        };
+        let claim_key = format!("{}:{}", identity.namespace, identity.token_ids.len());
```
i386: **P2 — this is keyed by namespace + query length, not by prefix.** Different token sequences of the same length suppress one another even though they would read different entries, while two queries of different lengths that both resolve to the same recorded L3 prefix can perform the same disk load concurrently. That does not satisfy the documented per-prefix single-flight contract. Key the claim by the actual candidate entry (namespace + recorded length + prefix hash/manifest key); because discovery currently happens inside `fill_longest`, the clean shape may be a lookup/claim/load split in the tier.

scripts/l3_warmup_bench.py
```diff
@@ -0,0 +1,248 @@
...
+        errors: list[str] = []
+
+        def run(tag: str, tail: str) -> None:
+            try:
+                ttft, _ = ttft_request(args.port, model, prefix + tail, args.max_tokens)
+                concurrent[tag] = ttft
+            except Exception as error:  # noqa: BLE001 - report, don't crash the bench
+                errors.append(f"{tag}: {error}")
+
+        threads = [
```
i386: **P2 — this benchmark cannot certify the labels it reports.** These two requests use different tails, and the script only records TTFT/success: it never asserts `skippy.exact_cache.source == l3`, the restored prefix length, or that only one fill/read occurred. Both requests can succeed after two disk loads and the report will still print `single-flight`; restart can silently cold-prefill and still be labelled `L3 fill`. Please ingest the emitted attrs (or add a bounded metrics/debug endpoint), assert source/restored tokens, and expose a fill-start/read counter so the concurrent case proves one physical fill. Also wait for the asynchronous spill to commit before killing the first server, otherwise the restart arm is timing-dependent on large states.

i386 (commented):
Re-reviewed the new delta `4f42615..c009998c` line-by-line. The length-aware prefix index, capped default, single-pass eviction, NotFound-only pruning, dense-path routing, L3 source/fill telemetry, and lab-only TCP warning are meaningful improvements. The package suites and Clippy are green at the exact reviewed head.

Four actionable notes remain below. The two P1s mean I would not yet treat the weight-identity or dense-L3 correctness gaps as closed; the two P2s mean the current benchmark cannot certify its single-flight claim.

Validation at `c009998c`: skippy-cache 69/69; skippy-server 530 passed, 3 ignored; skippy-correctness 115 passed, 160 model-download cases ignored; Clippy clean for all three crates; `cargo fmt --check` and Python syntax clean. The PR is currently `CONFLICTING`/`DIRTY` against `main`, and as a draft has only the CodeRabbit status context rather than the repository CI matrix.

crates/skippy-correctness/src/runner/remote_handoff.rs
danielwinterw: Fixed in 8fc71512f: the harness now hashes the served artifact (sha256, memoized per path so hello/manifest/restore call sites don't re-hash gigabytes) and feeds it as source_model_sha256 — directories (layer-package refs) stay un-hashed and ride the package manifest identity. The connected test you asked for is there: two files with different contents behind one display model id produce different identities, and the same file re-queried is stable.

crates/skippy-server/src/kv_integration/exact_state.rs
danielwinterw: Fixed in 8fc71512f at three layers: record_exact_state skips zero-byte payloads outright, L3Tier::spill refuses them (unit-tested), and both restore paths (radix hit and L3 fill) treat a zero-byte entry as a miss — so the dense/no-native-KV case can neither persist nor restore position-without-state. The tier-level rejection test covers the regression shape; the record/restore guards are the belt and braces above it.

crates/skippy-server/src/kv_integration/exact_state.rs
danielwinterw: Fixed in 8fc71512f with the split you suggested: the tier now exposes locate_longest (cheap index probes returning an L3Location with the manifest key) and load; the server claims on the manifest key, so same-length different-prefix queries never suppress each other and different-length queries resolving to one entry share one load. One more hole your bench comment flushed out: the claim was released before the asynchronous radix re-warm landed, so a request arriving in that window re-read the entry — the claim now travels with the re-warm record and the worker releases it only once the entry is radix-resident. The updated bench certifies exactly one physical fill in the concurrent arm.

scripts/l3_warmup_bench.py
danielwinterw: Fixed in 8fc71512f: the bench now waits for the write-behind spill to commit (manifest count stable on disk) before the first shutdown, then certifies against serve-log evidence — the restart arm must show a real 'fill ok' with the restored token count (recorded in the report as restart_filled_tokens), and the concurrent arm counts 'fill start' lines and requires exactly one physical fill. Any label the evidence doesn't support fails the run with a non-zero exit. It caught a real bug on its first certified run (the re-warm race above), so point well taken. Verified end-to-end against a real serve: restart certified at 1664 tokens filled from disk, concurrent certified at one physical fill.

i386 (commented):
I reviewed the current `e11e30ae` head and cannot approve it. Representative blocking issues remain in current code: the plain-TCP `skippy-kv/1` service explicitly has no authentication, uses blocking reads without connection deadlines, and serves accepted connections serially; the new service path still introduces direct `eprintln!` calls against the repository logging policy; and multiple exact-state/manifest/GC correctness findings remain unresolved in the existing review threads. The branch is also conflicted with `main`, and the 29-file / 5.6k-line change needs those threads cleared and current-base validation before approval. I have not duplicated the existing inline findings.

i386 (changes requested):
Recording this as changes requested on the current `e11e30ae` head so the older approval on a prior commit is not mistaken for approval of the present branch. The blocking items are summarized in my immediately preceding review: the unauthenticated/deadline-free serial TCP service, repository-policy logging violations, unresolved exact-state/manifest/GC correctness threads, and the conflict with current `main`.

crates/skippy-cache/src/identity.rs
```diff
@@ -210,6 +210,90 @@ fn prefix_namespace_hasher(
...
+        FlashAttentionType::Enabled => b"fa:onnn",
+    });
+    hasher.update(&params.n_gpu_layers.to_le_bytes());
+    hasher.update(&params.layer_start.to_le_bytes());
+    hasher.update(&params.layer_end.to_le_bytes());
+    hasher.update(&params.ctx_size.to_le_bytes());
+    hasher.update(&params.lane_count.to_le_bytes());
+    hasher.update(b"kind:");
+    hasher.update(params.payload_kind.as_bytes());
+    update_platform_identity(&mut hasher);
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. `ExactStateIdentityParams` now includes the stable load-mode name in its digest, the remote handoff harness supplies the selected mode, and `load_mode_changes_identity` proves runtime and artifact slices cannot collide.

crates/skippy-cache/src/l3_remote.rs
```diff
@@ -0,0 +1,416 @@
...
+pub fn serve_store(
+    store: &HandoffSegmentStore,
+    listener: &TcpListener,
+    accept_count: usize,
+) -> Result<()> {
+    let mut served = 0usize;
+    loop {
+        let (stream, _) = listener.accept().context("skippy-kv accept failed")?;
+        served += 1;
+        if let Err(error) = serve_connection(store, stream) {
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. Each accepted peer now runs in its own scoped worker with bounded read/write deadlines. `stalled_peer_does_not_block_the_listener` holds one connection idle while a second completes a manifest request.

crates/skippy-cache/src/l3_remote.rs
```diff
@@ -0,0 +1,416 @@
...
+            eprintln!("skippy-kv connection failed: {error:#}");
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. The library-side `eprintln!` was removed; a bad peer is isolated to its worker and the listener remains live. The exact head passes `cargo run -p xtask -- repo-consistency no-console-print`.

crates/skippy-cache/src/l3_remote.rs
```diff
@@ -0,0 +1,416 @@
...
+/// A client connection to a peer's store.
+pub struct KvFetchClient {
+    reader: BufReader<TcpStream>,
+    writer: BufWriter<TcpStream>,
+}
+
+impl KvFetchClient {
+    pub fn connect(peer: &str) -> Result<Self> {
+        let stream = TcpStream::connect(peer)
+            .with_context(|| format!("failed to connect to skippy-kv peer {peer}"))?;
```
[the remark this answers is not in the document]
i386: The finding is valid for any mesh-reachable deployment. This draft now defaults its raw test listener to loopback, gives every connection bounded I/O deadlines, and explicitly documents that the TCP path is an unauthenticated lab transport; production exposure still requires the registered iroh ALPN with mesh membership authentication. I am leaving this thread open because replacing the lab transport with authenticated iroh is still the promotion blocker.

crates/skippy-cache/src/l3.rs
```diff
@@ -0,0 +1,645 @@
...
+        }
+        // Newest-first manifest list with each manifest's segment refs.
+        let keys = self.list_manifests()?;
+        let mut manifests = Vec::with_capacity(keys.len());
+        let mut reference_counts: std::collections::HashMap<String, (usize, u64)> =
+            std::collections::HashMap::new();
+        for key in &keys {
+            let Ok(manifest) = self.load_manifest(key) else {
+                continue;
+            };
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. Eviction, model-state clearing, and GC now propagate unreadable-manifest errors before deleting anything. The corrupt-manifest tests confirm referenced segments survive these destructive operations.

crates/skippy-cache/src/l3.rs
```diff
@@ -0,0 +1,645 @@
...
+                    .entry(segment.digest.clone())
+                    .or_insert((0, segment.bytes));
+                entry.0 += 1;
+            }
+            manifests.push(manifest);
+        }
+        if let Some(newest) = manifests.first() {
+            let newest_span: u64 = newest.segments.iter().map(|segment| segment.bytes).sum();
+            if newest_span > self.budget_bytes {
+                eprintln!(
```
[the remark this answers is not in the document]
i386: Already fixed in the current stack and reverified at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`: the cache library contains no unapproved console print at this site, and the exact head passes the no-console repository consistency check.

crates/skippy-cache/src/l3.rs
```diff
@@ -0,0 +1,645 @@
...
+            let Some(stem) = path
+                .file_stem()
+                .map(|stem| stem.to_string_lossy().into_owned())
+            else {
+                continue;
+            };
+            if !referenced.contains(&stem) {
+                freed = freed.saturating_add(entry.metadata()?.len());
+                fs::remove_file(&path)
+                    .with_context(|| format!("failed to collect segment {stem}"))?;
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. GC now considers only `.seg` files for orphan removal, so atomic-write temp files are preserved. `gc_preserves_in_flight_temp_files` covers the race.

crates/skippy-cache/src/tier.rs
```diff
@@ -0,0 +1,568 @@
...
+                    .kv_bytes()
+                    .context("failed to reconstruct KV bytes for spill")?
+                    .map(|bytes| bytes.into_owned())
+                    .unwrap_or_default(),
+                payload
+                    .recurrent_state_bytes()
+                    .context("failed to reconstruct recurrent state for spill")?
+                    .into_owned(),
+            ),
+        };
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. `L3Tier::spill` rejects `KvRecurrent` payloads with an absent or empty KV component before writing state, and `kv_recurrent_payload_requires_kv_bytes` covers the boundary.

crates/skippy-cache/src/tier.rs
```diff
@@ -0,0 +1,568 @@
...
+            );
+        }
+        let mut wire = self.store.assemble(&manifest)?;
+        let payload_bytes = wire.len() as u64;
+        let kv_bytes = usize::try_from(manifest.kv_bytes).context("kv bytes exceed usize")?;
+        let payload = match manifest.payload_kind.as_str() {
+            "full-state" => ExactStatePayload::full_state(wire),
+            "recurrent-only" => ExactStatePayload::recurrent_only(wire),
+            "kv-recurrent" => {
+                let recurrent = wire.split_off(kv_bytes);
```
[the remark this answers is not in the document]
i386: Already fixed in the current stack and reverified at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`: the imported KV length is validated against the declared split before `split_off`, so a peer-controlled oversized value returns an error instead of panicking.

crates/skippy-correctness/src/cli.rs
```diff
@@ -192,6 +193,97 @@ pub struct StateHandoffArgs {
...
+pub struct RemoteHandoffArgs {
+    #[command(flatten)]
+    pub runtime: RuntimeArgs,
+    #[command(flatten)]
+    pub output: OutputArgs,
+    #[arg(long, value_enum, help = "send = prefill node, recv = decode node")]
+    pub role: RemoteHandoffRole,
+    #[arg(
+        long,
+        default_value = "0.0.0.0:19081",
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. The harness listener now defaults to `127.0.0.1:19081`, and CLI help warns that non-loopback binding exposes an unauthenticated lab transport requiring a trusted private network.

crates/skippy-correctness/src/runner/remote_handoff.rs
```diff
@@ -0,0 +1,1956 @@
+use std::{
+    io::{BufReader, BufWriter, Read, Write},
+    net::{TcpListener, TcpStream},
+    time::{Duration, Instant},
+};
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. Artifact hashing, payload-kind selection, identity construction, and their tests moved into `remote_handoff/identity.rs`; the main module is now 1,823 lines.

crates/skippy-correctness/src/runner/remote_handoff.rs
```diff
@@ -0,0 +1,1956 @@
...
+/// Serve the local L3 store to peers over the `skippy-kv/1` fetch protocol.
+fn run_serve(args: RemoteHandoffArgs) -> Result<()> {
+    let store_dir = args
+        .store_dir
+        .clone()
+        .context("--role serve requires --store-dir")?;
+    let store = HandoffSegmentStore::open(&store_dir, args.store_budget_bytes)?;
+    let listener = std::net::TcpListener::bind(args.listen)
+        .with_context(|| format!("failed to bind skippy-kv listener on {}", args.listen))?;
+    eprintln!(
```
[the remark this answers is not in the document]
i386: Addressed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4` using the exception the review describes: these eight messages are intentional progress/error output from an interactive correctness harness, so the ratchet allowlist was regenerated for the exact sites. The exact head passes `cargo run -p xtask -- repo-consistency no-console-print`.

crates/skippy-correctness/src/runner/remote_handoff.rs
```diff
@@ -0,0 +1,1956 @@
...
+        Some(key) => key,
+        None => store
+            .list_manifests()?
+            .into_iter()
+            .next()
+            .with_context(|| format!("store at {} holds no manifests", store_dir.display()))?,
+    };
+    let manifest = store.load_manifest(&key)?;
+    let model_identity = runtime_model_identity(&args.runtime)?;
+    let local_state_identity = state_identity_for(&args, &model_identity);
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. Restore now loads the manifest first and derives streaming payload shape from `manifest.payload_kind` before computing local identity, so operators do not need to repeat `--streaming` for persisted page streams.

docs/skippy/REMOTE_HANDOFF_RUNBOOK.md
```diff
@@ -0,0 +1,197 @@
...
+`--streaming` to restore too so identity matches).
+
+## Peer fetch (`skippy-kv/1`) — cross-node prefix reuse
+
+Any node can serve its store and any node can pull by digest:
+
+```bash
+# node A: serve the store (no model load)
+target/release/skippy-correctness remote-handoff --role serve \
+  --listen 0.0.0.0:19092 --store-dir <path> --model <same.gguf> --layer-end <n>
```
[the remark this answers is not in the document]
i386: Fixed at `ec62cf9043bec895056cbe5fcc000abcf45ebbd4`. The runbook labels direct TCP as an unauthenticated lab transport, requires loopback or a trusted private network with explicit firewall restrictions, and reserves authenticated iroh `skippy-kv/1` for production mesh exposure.

i386:

PR #1519, opened 2026-08-29
Improve MoA routing under fleet churn and load

michaelneale:
MoA now routes around paused capacity, spreads origins stably across large-model replicas, and keeps a responsive worker when the preferred tier is degraded.

This is a fleet-robustness improvement to the existing MoA implementation. It does not change synthesis, refinement, committee quality policy, the wire protocol, or normal healthy-fleet fan-out.

### Route around unavailable capacity

`hosts_for_model` now excludes `RemotePaused` / `AllPaused` peers and puts `AcceptingDeprioritized` peers behind healthy capacity. A paused large-model host is no longer selected first just to consume a timeout.

### Avoid fleet-wide reshuffles

Replica ordering now uses rendezvous hashing over `(origin, peer)` instead of `hash(origin) % hosts.len()`. Unrelated churn no longer moves every origin to a different replica and discard its prefix-cache affinity.

### Preserve a responsive MoA worker

Small-worker exclusion now applies only when at least two big models are healthy. Committee capping also prefers healthy capacity within a tier, so paused/deprioritized big workers cannot cap out the only responsive route.

### Fail over a draft or reducer before giving up

Each logical remote worker now retains its origin-stable primary plus one same-model standby. Explicit transport, protocol, parse, and transient HTTP failures (408/429/500/502/503/504) move sequentially to the standby before any answer is committed; permanent 4xx responses stop immediately. Both attempts share the existing worker deadline: the primary retains the full timeout it had on `main`, and after an explicit failure the standby receives only the unused remainder, so recovery neither introduces an earlier slow-model cutoff nor multiplies tail latency. A connected-but-silent primary may consume the whole deadline. Self-filled replica slots remain single-peer so independently sampled committee members cannot collapse onto the same answer.

### Preserve tool-call authority

Tool turns still choose one actor, and the actor's real tool capability remains more important than health or advertised throughput. New regressions pin that ordering and retain fallback candidates.

## Fleet simulation evidence

The tests fabricate peer state but execute the real production `assemble_worker_pool` and actor-ranking path. They validate pool behavior rather than model-answer quality.

- **Scale:** 100 / 1,000 / 4,000 peers assembled in 1.1 / 12.3 / 53.9 ms in the measured local run, while admitted committee width stayed bounded.
- **Fairness:** across 300 simulated origins and 2 / 4 / 8 big replicas, every replica received first-choice traffic; the busiest replica received 1.05–1.28× an even share.
- **No herd on churn:** removing one replica preserves each origin's ordering among surviving replicas instead of reshuffling the whole fleet.
- **Degradation:** all-paused produces an honest empty pool; all-deprioritized still serves; all-small remains bounded; few-big/many-small remains servable.
- **Healthy parity:** healthy one-big and two-big fleet shapes assemble the same policy outcome as `main`.

This directly covers the fleet-level risk of every origin accidentally choosing the same large model due to ordering or churn. It does **not** claim dynamic queue-aware balancing: live inflight load is not currently gossiped, so a saturated large model cannot yet advertise queue depth. Admission state is the available shedding signal.

## Real workflow evidence

At commit `488bc3941`, three real `model: "mesh"` runs on two independently serving Qwen3.8-27B nodes completed the same four-turn tool workflow. Each produced exactly `read_file ×2`, then `write_file ×1`, and wrote the correct report values (348 / 48 / 30 / North).

That confirms the MoA tool path remains functional after the routing changes. Current external traces do not expose worker identities, so it does not prove both 27B replicas contributed to each response.

## Scope boundaries

- No MoA-vs-solo quality claim. The existing OpenRouter evaluation has unequal-output-budget and shared-throttling defects and is not valid evidence for answer quality.
- No change to the existing all-small collapse, big-tier admission threshold, committee caps, advisor policy, synthesis, or refinement.
- No protocol change.
- `hosts_for_model` is shared with direct routing, so paused/deprioritized filtering and stable replica ordering also improve direct routes.
- Draft and reducer backends use the same bounded primary-plus-standby failover because both call through the remote `ModelBackend`. There is no automatic failure gossip or persistent target-health scoring in this PR.

## Verification

Verified locally on current branch commit `4fb40ca4e`:

- `cargo fmt --all --check`
- `cargo test -p mesh-llm-host-runtime` — 2,713 unit tests + 2 integration tests passed; 8 ignored
- `cargo check -p mesh-llm`

The host-runtime suite needs `ulimit -n 8192` because the fleet simulations create real test nodes that bind sockets.

## Known limitations

- No live queue-depth signal or failure-driven ejection in gossip; admission state is coarse-grained.
- MoA failover is bounded same-model pre-response recovery, not full parity with direct routing's persistent target-health scoring.
- `healthy_big_count` counts distinct model names, not endpoint replicas. This PR pins current behavior rather than changing unsupported quality policy.
- Existing tier helpers disagree for unknown model sizes; unchanged here.

## Summary by CodeRabbit

- **Reliability**
  - Added automatic failover across multiple eligible remote replicas when a worker is unavailable or encounters a transient failure.
  - Preserved primary replica ordering while using standby replicas within shared request timeouts.

- **Performance & Scaling**
  - Improved replica distribution across origins to spread traffic more evenly.
  - Prioritized healthy capacity while retaining deprioritized replicas as spillover.
  - Improved behavior across large and mixed fleets, including context-capacity filtering.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/workers.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/workers.rs
@@ -424,14 +424,23 @@ fn is_retryable_replica_error(error: &str) -> bool {
         .is_some_and(|status| matches!(status, 408 | 429 | 500 | 502 | 503 | 504))
 }
 
+/// Fraction of the remaining worker deadline reserved for a standby attempt.
+///
+/// The primary keeps 75% of a full deadline (45s with the current 60s worker
+/// timeout), which remains above the documented 20-30s large-model first-call
+/// range while guaranteeing the standby a bounded chance after a hung primary.
+const STANDBY_DEADLINE_DENOMINATOR: u32 = 4;
+
 fn replica_attempt_timeout(
     deadline: tokio::time::Instant,
     index: usize,
     total_replicas: usize,
 ) -> Option<std::time::Duration> {
     let remaining = deadline.checked_duration_since(tokio::time::Instant::now())?;
-    let remaining_replicas = total_replicas.saturating_sub(index).max(1) as u32;
-    Some(remaining / remaining_replicas)
+    if index == 0 && total_replicas > 1 {
+        return Some(remaining - remaining / STANDBY_DEADLINE_DENOMINATOR);
+    }
+    Some(remaining)
 }
 
 async fn call_remote_replica(
@@ -704,8 +713,8 @@ mod tests {
 
         let budgets = budgets.lock().unwrap();
         assert_eq!(budgets.len(), 2);
-        assert_eq!(budgets[0], std::time::Duration::from_secs(5));
-        assert_eq!(budgets[1], std::time::Duration::from_secs(5));
+        assert_eq!(budgets[0], std::time::Duration::from_millis(7500));
+        assert_eq!(budgets[1], std::time::Duration::from_millis(2500));
     }
 
     #[test]
--- a/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/workers.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/workers.rs
@@ -396,7 +396,7 @@ where
     let deadline = tokio::time::Instant::now() + timeout;
     let mut last_error = "no eligible remote replicas".to_string();
     for (index, peer_id) in replicas.iter().copied().enumerate() {
-        let Some(attempt_timeout) = replica_attempt_timeout(deadline, index, replicas.len()) else {
+        let Some(attempt_timeout) = replica_attempt_timeout(deadline) else {
             break;
         };
         match attempt(index, peer_id, attempt_timeout).await {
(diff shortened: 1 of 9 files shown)


crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/workers.rs
```diff
@@ -341,27 +348,118 @@ impl moa::ModelBackend for RemoteModelBackend {
...
+                    break;
+                }
+            }
+        }
+    }
+    Err(last_error)
+}
+
+/// Whether a remote failure is safe to retry on a same-model standby.
+fn is_retryable_replica_error(error: &str) -> bool {
```
i386, line 417: Non-blocking: retry classification string-parses the error message produced by `parse_quic_http_response` ~60 lines below (and the timeout wrapper just above). The `starts_with("HTTP ")` + `nth(1)` + `trim_end_matches(':')` chain means any format drift there silently reclassifies errors — e.g. a future prefix change would make everything look retryable, or nothing. The unit test pins today's exact strings so it won't regress silently, which is why I'm fine approving as-is; a small typed error (or a retryable flag threaded from the failure site) would remove the coupling entirely. Happy to see it as a follow-up.

i386 (approved):
Reviewed from a fresh checkout at head `4fb40ca4e`. Verified: all 120 `moa_gateway` tests (incl. the new fleet sim/fairness/scale suites and all five replica-failover tests) pass locally with `ulimit -n 8192`, and CI is fully green at head.

**What I checked:**

- `hosts_for_model` rendezvous change — all four external callers (`auto_route`, `transport.resolve_mesh_target_hosts`, `ingress` x2) consume the full ordered list and benefit from (or tolerate) the new ordering; paused-peer exclusion makes `model_has_eligible_target` honestly empty rather than routing to a peer that will refuse, which is what the degrade path needs.
- Admission-control gating on `healthy_big_count` — no-leak default (`unwrap_or(true)`) keeps healthy-fleet behavior bit-identical to main; the parity tests pin that.
- `cap_committee` availability term sorts within tier only, so the measured admission rule (healthy small never displaces big) is preserved. Good catch on the second-attempt regression test — the alphabetical-luck story in the commit message is exactly why it bites.
- Replica failover — shared deadline semantics are sound (primary keeps the full timeout; an explicit retryable failure passes only the unused remainder; a timed-out primary leaves nothing for the standby, so no latency multiplication). Self-fill slots pinned to `vec![peer_id]` correctly prevents duplicate answers from the same replica. Non-retryable 4xx stops immediately.
- All four CodeRabbit findings are addressed at head with regression coverage, not just the code change.
- Tests assert results (pool widths, orderings, sets), not wall-clock timings — no obvious flake vector in CI.

**One doc nit before merge:** the PR body still describes the 75/25 primary/standby timeout split ("the primary keeps 75% (45s ...) and the standby gets the remaining 25%"), which `4fb40ca4e` removed in favor of full-primary-timeout + unused-remainder failover. Worth updating the description so the merge record matches the shipped behavior.

Non-blocking inline nit below. Nice work — the test discipline here (verifying each regression test fails against a deliberately broken comparator before trusting it) is above the bar.

PR #1520, opened 2026-08-29
Keep embedded startup on one validated config image

michaelneale:
Embedded SDK callers now get consistent startup behavior even if the source config file changes while the worker thread starts.

The SDK validates the source bytes once, stores them in a process-owned temporary snapshot, and routes every embedded startup config load through that snapshot. The snapshot stays alive for the embedded handle's lifetime. This prevents logging, plugins, and runtime options from observing different config images during one startup.

This is startup consistency hardening rather than a high-severity security fix: deployments should still keep executable config policy in appropriately owned locations.

Closes project-loupe/audit-mesh-llm#752.

## Validation

- `cargo test -p mesh-llm-host-runtime --lib` — 2679 passed, 8 ignored
- `cargo check -p mesh-llm`
- `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings -A unfulfilled-lint-expectations`
- `cargo clippy -p mesh-llm --all-targets -- -D warnings -A unfulfilled-lint-expectations`
- `cargo fmt --all --check`

The lint-expectation allowance is for pre-existing unfulfilled `#[expect(dead_code)]` attributes on current `main`; unmodified Clippy fails on those unrelated existing warnings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Embedded nodes now use a stable snapshot of the validated configuration, preventing changes to the original configuration file from affecting an active runtime.
  * Configuration files are validated for readability, UTF-8 encoding, and valid TOML before startup.
* **Tests**
  * Added coverage confirming runtime behavior remains stable when the source configuration changes after initialization.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/sdk.rs
+++ b/crates/mesh-llm-host-runtime/src/sdk.rs
@@ -61,7 +61,7 @@ pub struct EmbeddedServeHandle {
     invite_token: Option<String>,
     control_tx: Option<tokio::sync::mpsc::UnboundedSender<crate::api::RuntimeControlRequest>>,
     task: Option<std::thread::JoinHandle<Result<()>>>,
-    _isolated_config: Option<NamedTempFile>,
+    _config_snapshot: NamedTempFile,
 }
 
 pub type EmbeddedMeshNodeHandle = EmbeddedServeHandle;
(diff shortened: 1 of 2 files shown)


i386 (approved):
Reviewed the embedded startup boundary and config-loading call sites. The SDK parses exactly the bytes it copies, rewrites every downstream load to the retained snapshot, and keeps that file alive for the handle lifetime; dropping the intermediate isolated config only happens after the second snapshot is flushed. The source-swap regression covers the intended invariant, and CI is green.

PR #1523, opened 2026-08-29
Reclaim split-prefill tokens after failed sessions

michaelneale:
Skippy stages now reclaim buffered split-prefill tokens when a binary stage connection fails or disconnects before sending a graceful `Stop`, preventing an authenticated peer from leaving process-wide token allocations behind.

Addresses [Project Loupe audit-mesh-llm #1648 — Split-prefill token accumulation persists after failed sessions](https://github.com/project-loupe/audit-mesh-llm/issues/1648).

## What changed

- pass KV integration state into the existing per-connection orphan-session cleanup
- remove each orphaned session's entry from `split_prefill_tokens` before releasing its runtime session
- preserve entries owned by other active connections
- add a regression test covering selective cleanup

## Protocol

No wire-format or mixed-version compatibility changes.

## Validation

- `cargo fmt --all --check`
- `cargo clippy -p skippy-server --lib --all-targets -- -D warnings`
- `cargo test -p skippy-server --lib` — 530 passed, 3 ignored

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of disconnected binary sessions and prevented overlapping connections from claiming sessions during release.
  * Reclaimed buffered prefill tokens from orphaned sessions to help prevent resource buildup.
  * Ensured connection cleanup properly accounts for key-value stage integration.
  * Improved handling of stale session stop requests to avoid disrupting active replacements.
  * Strengthened session ownership and cleanup reliability during concurrent connections and failures.


i386 (approved):
Reviewed the disconnect cleanup path, the ownership boundary between per-connection tracked sessions and the process-wide split-prefill map, and the selective regression test. Removing buffered tokens before releasing each orphaned runtime session closes the leak without touching sessions owned by other connections. CI is green.

ndizazzo (changes requested):
WE should tie orphan cleanup to the connection that owns the buffered state.

`release_tracked_connection_sessions` removes the process-wide `split_prefill_tokens` entry using only the wire session ID. That ID isn't unique to a connection. Trusted agent sessions deliberately reuse a stable nonzero ID across requests. If an old connection disconnects after a replacement connection starts using the same session, line 64 can remove tokens buffered by the replacement.

Please verify a connection ownership or generation token before removing shared state, or serialize connection replacement so stale cleanup can't run against a newer owner. Add a regression test with overlapping old and new connections that share one session ID. The test should show that cleanup from the old connection leaves the new connection's buffered tokens intact.

Validation: the five focused session-tracker tests passed, and the PR's CI checks passed. My local `just test-all` run completed the native llama.cpp build, repository checks, formatting, and GPU bench check, then stopped in unchanged `skippy-protocol/src/binary/codec.rs` on Rust 1.98's new `chunks_exact_to_as_chunks` Clippy lint. The remaining local suite and E2E smoke didn't run.

ndizazzo (changes requested):
The ownership fix looks right. I verified the stale cleanup and overlapping replacement cases, and the focused tests pass.

One small thing still blocks this: the new lines shift six existing console-print locations, but the allowlist still has the old line numbers. `just no-console-print` passes on the base and fails on this head.

Can you regenerate the console-print ratchet? That should be the only change needed here.

The full suite stopped on unchanged Rust 1.98 Clippy errors in `skippy-protocol`.

PR #1527, opened 2026-08-29
Reject payload-bearing prediction-return opens early

michaelneale:
Skippy's direct prediction-return listener now rejects noncanonical open frames after reading only their fixed-size header, so payload-bearing opens cannot trigger peer-controlled allocation or block waiting for an irrelevant body.

Addresses [Project Loupe audit-mesh-llm #1657 — Reject payload-bearing prediction-return opens before allocation](https://github.com/project-loupe/audit-mesh-llm/issues/1657).

## What changed

- read the fixed 72-byte prediction-return open header before invoking the generic codec
- reject the wrong kind, nonzero payload counts, and any noncanonical state header
- parse canonical headers from an in-memory fixed buffer, preserving request/session routing IDs
- add a socket-level regression proving an advertised 8 MiB metadata payload is rejected without reading its body

## Protocol

Canonical prediction-return opens are wire-compatible and unchanged. Previously accepted noncanonical opens are now rejected.

## Validation

- `cargo fmt --all --check`
- `cargo test -p skippy-server --lib` — 530 passed, 3 ignored

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of prediction responses before processing their contents.
  * Malformed messages, unsupported message types, and invalid metadata are now rejected earlier and more reliably.
  * Prevented invalid responses from triggering unnecessary variable-length data reads.
  * Ensured only correctly formatted response headers are accepted before message decoding begins.

### DIFF
--- a/crates/skippy-server/src/binary_transport/direct_return.rs
+++ b/crates/skippy-server/src/binary_transport/direct_return.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::HashMap,
-    io,
+    io::{self, Read},
     net::{IpAddr, SocketAddr, TcpListener, TcpStream},
     sync::{
         Arc, Mutex,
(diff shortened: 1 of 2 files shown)


ndizazzo (changes requested):
Replace the direct `eprintln!` in `PredictionReturnReceiver::attach_opened_stream` at `crates/skippy-server/src/binary_transport/direct_return.rs:257` with the repository's event/log path.

This PR moves a legacy, ratcheted console print from line 223 to line 257. The Quality lane now fails `just no-console-print` with a forbidden occurrence at line 257 and a missing approved occurrence at line 223. Moving that entry in `tools/xtask/data/console_print_allowlist.json`, or adding another exception, would preserve the legacy debt and violate the ratchet's forward-only rule.

`skippy-server` already depends on `mesh-llm-events`. Follow the established local pattern and emit this reader failure through `mesh_llm_events::emit_event` with an `OutputEvent::Error`, retaining the formatted error and a useful prediction-return context. Then regenerate the ratchet only to retire the removed legacy entry, and verify that `just no-console-print` passes without a new exception.

michaelneale:
@ndizazzo over to you - merge if seems suitable.

ndizazzo (approved):

PR #1531, opened 2026-08-29
Disable ggml compiler-cache autodetection when we decline sccache

danielwinterw:
## The bug

`scripts/build-llama.sh` owns the compiler-caching decision, but it only passed `-DGGML_CCACHE=OFF` when caching was disabled *explicitly*. When no sccache was found it printed

```
sccache not found; llama.cpp build will run without compiler caching
```

and left `GGML_CCACHE` ON, so ggml ran its own `find_program(sccache/ccache)`. That search covers `CMAKE_SYSTEM_PREFIX_PATH`, not just `PATH`, so a build environment with a sanitized `PATH` but an sccache installed elsewhere on the system gets a bare `sccache` compiler launcher it cannot execute:

```
/bin/sh: sccache: command not found
ninja: build stopped: subcommand failed.
```

So the message says caching is off while the build is in fact configured to use a cache binary that isn't reachable.

I hit this building the Metal runtime inside Homebrew's build sandbox, which strips `/opt/homebrew/bin` from `PATH` but leaves `/opt/homebrew` on CMake's system prefix path. Any sanitized or containerized build with a host-installed sccache can trip it.

## The fix

Fold the two non-injecting branches together so `GGML_CCACHE=OFF` is passed whenever this script does not install a launcher itself. Note the sccache-*unavailable* path (binary found but not startable) already fell through to the correct branch — only the sccache-*not-found* path was affected.

## Verification

Ran the script against a stub `cmake` that captures the configure arguments, with a fake prepared workdir, across all three decision paths:

| scenario | `-DGGML_CCACHE=OFF` | compiler launchers |
| --- | --- | --- |
| sccache absent from PATH — **before this change** | ✗ not passed | none |
| sccache absent from PATH — **after** | ✓ passed | none |
| sccache present and working | not passed | ✓ C and CXX |
| `LLAMA_STAGE_USE_SCCACHE=0` | ✓ passed | none |

`bash -n scripts/build-llama.sh` is clean. The two unchanged rows confirm the normal cached build and the explicit opt-out are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build behavior when the compiler cache is unavailable.
  * Preserved expected failures when compiler caching is explicitly required.
  * Prevented unintended compiler-cache detection during builds without a usable cache launcher.

### DIFF
--- a/scripts/build-llama.sh
+++ b/scripts/build-llama.sh
@@ -286,13 +286,20 @@ if [[ "$USE_SCCACHE" != "0" && -n "$SCCACHE_BIN" ]]; then
       ;;
   esac
   echo "using sccache for llama.cpp C/C++ compilation: $SCCACHE_BIN"
-elif [[ "$USE_SCCACHE" != "0" ]]; then
-  if [[ "${MESH_LLM_REQUIRE_SCCACHE:-0}" == "1" ]]; then
-    echo "sccache is required but was not found" >&2
-    exit 1
-  fi
-  echo "sccache not found; llama.cpp build will run without compiler caching" >&2
 else
+  if [[ "$USE_SCCACHE" != "0" ]]; then
+    if [[ "${MESH_LLM_REQUIRE_SCCACHE:-0}" == "1" ]]; then
+      echo "sccache is required but was not found" >&2
+      exit 1
+    fi
+    echo "sccache not found; llama.cpp build will run without compiler caching" >&2
+  fi
+  # This script owns the compiler-caching decision, so ggml must not make its
+  # own. With GGML_CCACHE left ON, ggml runs find_program(sccache/ccache), which
+  # searches CMAKE_SYSTEM_PREFIX_PATH rather than just PATH and registers a bare
+  # `sccache` launcher. A sanitized build environment that has such a binary
+  # installed but not on PATH then fails every compile with
+  # "/bin/sh: sccache: command not found".
   CMAKE_ARGS+=(-DGGML_CCACHE=OFF)
 fi
 


i386 (approved):
Reviewed — correct root-cause fix, minimal and well-explained. ✅

**The bug is real and subtly nasty:** ggml's `GGML_CCACHE` runs its own `find_program(sccache/ccache)` over `CMAKE_SYSTEM_PREFIX_PATH`, not just `PATH`. So in a sanitized build env (Homebrew sandbox strips `/opt/homebrew/bin` from `PATH` but leaves `/opt/homebrew` on the system prefix path) ggml registers a bare `sccache` launcher the compiler can't exec — every compile dies with `sccache: command not found`, while the script cheerfully reported caching was *off*. Message and reality disagreed.

**The fix is exactly right:** fold the two non-injecting branches together so `-DGGML_CCACHE=OFF` is passed whenever this script does not install a launcher itself. Tracing the three paths:
- sccache found + working → first branch injects `CMAKE_*_COMPILER_LAUNCHER`, `GGML_CCACHE` left ON (harmless — ggml's `find_program` defers when the launcher var is already set). Unchanged. ✔
- sccache not found → now falls into the `else`, warns (or hard-exits under `MESH_LLM_REQUIRE_SCCACHE=1` *before* reaching the flag), then passes `GGML_CCACHE=OFF`. Fixed. ✔
- `LLAMA_STAGE_USE_SCCACHE=0` explicit opt-out → same `else`, passes `GGML_CCACHE=OFF`. Unchanged. ✔

The require-fail `exit 1` correctly precedes the `CMAKE_ARGS+=(-DGGML_CCACHE=OFF)`, so "sccache is required but missing" still hard-fails rather than silently building uncached.

The inline comment captures the whole `CMAKE_SYSTEM_PREFIX_PATH`-vs-`PATH` gotcha for the next reader — good to have it at the fix site. `bash -n` clean, the stub-cmake verification table covers all three paths, and CI is fully green.

Approving.

i386:
good catch

PR #1533, opened 2026-08-29
fix(llama): rebase patch queue onto upstream cc83d7b482

i386:
## llama.cpp canary repair: rebase patch queue onto upstream `cc83d7b482`

The nightly llama-upstream canary (run `33247859447`) could not replay our
patch queue onto the new upstream master. This PR rebases the 32-patch skippy
queue onto upstream `cc83d7b4824f73cfdda4dfbb47ee39804f71b328` (from the old
pin `f5e85d43a048f3d5adefb4c5e29867d8077fba62`, 11 upstream commits) and moves
`third_party/llama.cpp/upstream.txt` to the new pin.

One commit: `fix(llama): rebase patch queue onto upstream cc83d7b4`. It touches
only `third_party/llama.cpp/patches/*` and `third_party/llama.cpp/upstream.txt`
— no Rust, build-script, or certification-manifest changes.

## Upstream changes absorbed (`f5e85d43a0..cc83d7b482`, 11 commits, 43 files)

The two changes that actually intersect the queue:

- **`6fe749801` — qwen4exp: reduce number of graph splits (#27880).** The PLE
  (per-layer-token embedding) gather was hoisted out of the per-layer
  `build_ple()` into a new constructor-level `build_inp_ple()` with an
  unconditional `ggml_build_forward_expand` so the gather lands in the same
  graph split as the token embeddings; `build_ple()` now takes the
  prebuilt/gathered `emb` tensor. `build_inp_embd` also gained a
  constructor-level forward expand, and `models.h` gained the
  `build_inp_ple` declaration. This rewrote exactly the code our
  qwen4exp stage-filter patches (0027/0029) own.
- **`50f068fff` — bench: add `--tensor-read-lazy` (#27881).** Renamed the
  public lazy-load surface: `llama_tensor_read_lazy` /
  `LLAMA_TENSOR_READ_LAZY_*` became `llama_lazy_mode` / `LLAMA_LAZY_MODE_*`
  in `include/llama.h`, with mechanical renames through
  `llama-model-loader.{cpp,h}`, `llama-model.cpp`, `llama.cpp`, `common/`,
  and `llama-bench`.

The rest is backend drift the queue absorbed by 3-way/recontextualization:

- `a43c3986b` — ggml: fix `conv_transpose_2d` for multiple batches; the Metal
  conv kargs struct gained `nb3` (+1 line in `ggml-metal-impl.h`, shifting
  hunk offsets), CPU `ops.cpp` reworked, `conv.metal` updated.
- `866322481` — context: disable non-fused GDN and LID ops; `cparams`
  defaults `auto_fgdn`/`auto_flid` flipped from `true` to `false` in
  `llama-context.cpp`.
- `77f132cb1` — vulkan: `mul_mat_id` now pads K rather than N.
- `90c26fcd4` — vulkan: hoisting support for row IDs / expert count in
  shaders (`count_experts.comp`, `mul_mm*.comp`).
- `b387ddfd8` — vulkan: fix missing view-alias dependencies in
  `ggml_vk_graph_optimize`.
- `c9ca51c1f` — vulkan: combine duplicated fastdiv shader helpers.
- `5ea1b124e` — metal: fa-vec tunings for M1 Max (`ggml-metal-tuning.cpp`).
- `cc83d7b48` — sycl: make `--fit` respect `--fit-target` better (fattn).
- `d7bd3bfca` — snapdragon: Windows python SDK setup (new
  `scripts/snapdragon/{sdk,setup-sdk}.py`; docs-only for us).

## How the patch queue evolved

The queue was rebuilt by the deliberate rewrite path (`llama-patch-changes`
skill): capability-owned commits reconstructed on the new upstream pin, then
the whole series regenerated with `format-patch`. Queue size is unchanged
(32 patches); every patch got a fresh commit hash and renumbered
`[PATCH xx/32]` subject — that is regeneration noise, not content. Per patch:

| Patch | What happened |
|---|---|
| 0001–0026 | Applied; **no patch-content change**. `git am --3way` absorbed the lazy-load rename, the Metal kargs `nb3` shift, and vulkan/sycl/cli drift. Regenerated files differ only in commit hash, subject numbering, blob `index` lines, and hunk line offsets (0012/0018 shift by the one `nb3` line in `ggml-metal-impl.h`). No hunk text changed in any of these patches. |
| **0027** (stage Qwen4Exp hyper-connected residuals) | **Hard conflict**, `src/models/qwen4exp.cpp` — this is the patch that stopped the canary's `git am` (verified by replaying the old queue onto `cc83d7b482`; 0001–0026 apply clean, 0027 is the first failure). Upstream's PLE hoist rewrote the constructor region 0027 owns. Resolution recomposes the staged constructor with upstream's new shape: upstream's unconditional `ggml_build_forward_expand(gf, inpL)` now lives **inside the unfiltered branch only** (a middle/final stage receives the prior stage's wide residual through the dedicated `[n_embd, hc, tokens]` stage input, so embedding rows are never expanded there); `inp_out_ids` is only built when unfiltered or `include_output`; upstream's constructor-level `ple_emb = build_inp_ple(...)` + forward expand is kept; and the per-layer call uses upstream's new `build_ple(inp, ple_emb, res_hc, il)` signature. |
| 0028 (runtime slices), 0030 (wide-frame verify) | Applied clean; regenerated for hash/subject and git signature tag only. |
| **0029** (scope Qwen4Exp PLE tables to owning stages) | Loader-side and `model_package.cpp` hunks unchanged (index-line drift only). The patch **gained one semantic change** in `qwen4exp.cpp`: the constructor-level PLE build is now additionally gated on *stage PLE containment* — a loop over `[il_start, il_end)` setting `stage_contains_ple` when any retained layer satisfies `hparams.is_ple()`. Rationale below. |
| 0031 (activation frame staging), 0032 (shared-table ownership) | Applied clean; hash/signature/index drift only. |

### The mid-repair battery failure and the 0029 gate

The first post-rebase certification battery
(`target/family-battery/20260829-204516-53588`) failed **26 lanes, all
qwen4exp**: split 1 `single-step` died with exit 139 (SIGSEGV) and splits
1–12 and 24 `single-step`/`chain` timed out behind it. Root cause: a stage
whose retained layer range contains no PLE layer does not load the PLE
gather table (0029's loader-side filtering skips `per_layer_tok_embd`), but
the new constructor-level `build_inp_ple()` unconditionally dereferenced the
missing tensor. The fix mirrors 0029's existing loader-side gate on the
graph side: `build_inp_ple` is only called when the stage actually retains a
PLE layer. Everything else in the battery was green in that run.

## Validation

- **Tree identity (both directions).** A pristine replay of the regenerated
  series onto `cc83d7b482` reproduces the reconstructed head exactly.
  Diffing old-final (old queue on `f5e85d43a0`) against new-final (this
  queue on `cc83d7b482`) yields **exactly the upstream drift file set** —
  no queue-owned file changed beyond the reconciled `qwen4exp.cpp` /
  loader overlap (verified independently: `src/skippy/*` and
  `include/skippy/*` are byte-identical between old-final and new-final).
- **Build:** `scripts/build-llama.sh` (Metal, arm64) plus
  `cargo check -p skippy-ffi -p skippy-runtime -p skippy-server` pass.
- **Certification:** `scripts/skippy-family-battery.sh --skip-build`
  (run `20260829-214420-68598`): **138/138 planned certifications green**
  (451 pass outcomes, 274 policy skips), including all 13 qwen4exp splits
  (1–12, 24) that failed pre-fix. (An earlier battery attempt
  `20260829-204013-46769` was aborted before recording lane results.)

## Risks for reviewers

**ABI impact: none.**
- The Skippy stage ABI is untouched: `include/skippy/common.h` still
  declares `SKIPPY_ABI_VERSION 0.1.42` at the patched head, the Rust mirror
  in `crates/skippy-ffi` (`ABI_VERSION_*` = 0.1.42) is unchanged, and this
  PR modifies no Rust file.
- `include/skippy.h` / `include/skippy/*` patch hunks are byte-identical
  (index-line drift only). Patch 0003 (public ABI surface) has no content
  change.
- Upstream's `LLAMA_LAZY_MODE_*` rename touches llama's public API, not the
  Skippy ABI; no queue patch and no Rust crate referenced the old
  `tensor_read_lazy` symbols (grep-verified), so there is no FFI follow-up.

**Lane impact: none failing, none excluded.**
- The failing and green battery runs have the **identical policy skip set**
  (274 lanes) and identical 138-certification plan; this PR does not touch
  `ci/llama-canary/family-certified.json` or any battery policy. Nothing was
  weakened or excluded to get green.

**What reviewers should scrutinize:**

1. **0027/0029 composition is the real semantic surface.** Parity lanes
   prove qwen4exp behaves on Metal, but only review can confirm intent:
   - the forward expand for `inpL` is intentionally *not* executed on
     middle/final stages (upstream made it unconditional);
   - `inp_out_ids = nullptr` on filtered stages without output inclusion —
     all downstream uses must stay guarded;
   - the `stage_contains_ple` gate must stay in lockstep with the
     loader-side PLE gate in the same patch: if a future change alters which
     layers a stage retains, the two gates can drift and reintroduce the
     null-gather segfault (or silently skip a needed gather).
2. **Metal-only certification.** The canary builds and certifies Metal/arm64
   only. Upstream changed Vulkan shader paths (`mul_mat_id` K-padding,
   expert-count hoisting, fastdiv consolidation, view-alias fix), SYCL
   fattn, and CPU `conv_transpose_2d` batching — those sources are compiled
   into the CUDA/Vulkan/ROCm/SYCL native-runtime lanes built from this pin.
   A shader-level regression there would not surface in this battery.
3. **Backend-behavior changes inherited from upstream**, all covered by the
   green battery on Metal but worth knowing about: `conv_transpose_2d`
   multi-batch fix (Metal kernel is JIT-compiled at first model open; the
   battery's model loads exercise it), `auto_fgdn`/`auto_flid` now default
   to `false` (affects GDN/LID hybrid families — falcon-h1, kimi-linear,
   jamba2, granite-hybrid, qwen3-next all certified), and upstream's
   qwen4exp graph-split reduction now applies to unfiltered full-model
   runs (fewer splits; parity held).
4. **Review noise:** every patch file's `From` hash and `[PATCH xx/32]`
   subject changed, and the git signature footer moved to 2.55.0 — compare
   hunks, not headers. Outside header/index/offset noise, the whole-queue
   diff is only ~37 changed lines, all in 0027/0029 (e.g.
   `git diff main...HEAD -- third_party/llama.cpp/patches/ | grep -vE '^[+-](index |From |Subject: .*\[PATCH|2\.[0-9.]+)'`).

After merge, the next canary run re-certifies `main` before the pin's
`update-pin` path advances anything automatically.


i386:
**Family battery passed** after the agent repair at upstream cc83d7b4824f73cfdda4dfbb47ee39804f71b328.
All certification lanes green on the family-certify runner; certified commit: `39dab81db73c5acb391a14650aacd25cf0457333`. Post-green agent review found nothing to change; the certified tree stands.

<details><summary>Post-green agent review report (tail)</summary>

```
  certifications passed**, all lanes `single-step` / `chain` / `state-handoff`
  = pass, no FAIL entries; "skip" regex hits are the substring in "skippy"
  (build/compiler output), not skipped lanes.

## Defects found

None. No fix required; no `review(llama):` commit created; the tree is
unchanged by this review.

## Recommended follow-ups (non-blocking, for the wrapper/PR owner)

1. The PR description should mention the upstream rename absorbed as queue
   context (`tensor_read_lazy` → `lazy_mode`, upstream `50f068fff`) and the two
   Qwen4Exp adaptations (forward-expand moved into the non-staged branch;
   `stage_contains_ple` guard), so reviewers don't mistake them for regressions.
2. After merge, the next canary run re-certifies; the two new-context Qwen4Exp
   hunks are the ones worth watching in the first post-merge battery.

Review scope note: `.deps/llama.cpp` was used read-only plus one throwaway
worktree (since removed); its index was restored to the pre-review state.
```

</details>

PR #1536, opened 2026-08-29
fix(skippy-server): contain a panicking connection worker instead of stopping the accept loop

### LINKED ISSUE #1473: A panicking binary-stage connection worker stops the accept loop
Raised out of review on [#1409](https://github.com/Mesh-LLM/mesh-llm/pull/1409), where CodeRabbit flagged it against code that predates the PR. Filing separately rather than widening that diff.

`ConnectionWorkers::reap_finished` in `crates/skippy-server/src/binary_transport/binary_messaging.rs` turns a panicked worker into an error:

```rust
if self.0[index].task.is_finished() {
    let worker = self.0.swap_remove(index);
    if worker.task.join().is_err() {
        bail!("binary stage connection worker panicked");
    }
}
```

The accept loop calls `connection_workers.reap_finished()?`, so a panic in any one connection's worker thread propagates out of the loop and takes down the listener. Every other live connection then goes through shutdown, and the stage stops accepting new ones — a single malformed request or unexpected state on one connection can end serving for the whole process.

Containing it per-connection would match how the iteration scheduler already handles worker panics (`SchedulerWorker::run` catches unwind and fails just that worker's requests): log the panic, fail that connection, and keep accepting.

Introduced with the connection-worker tracking in [#1420](https://github.com/Mesh-LLM/mesh-llm/pull/1420). Not triggered by anything in the run-ahead stack; noticed while reviewing it.

Virgile-pct:
## Original problem

As described in #1473: `ConnectionWorkers::reap_finished` turns a panicked connection worker into a `bail!`, and the accept loop propagates it (`connection_workers.reap_finished()?`), so a single panicking connection worker stops the whole binary-stage listener. One malformed request or unexpected per-connection state can take the node's stage serving down for every other client.

## Diagnostics

Code inspection, starting from the issue: `crates/skippy-server/src/binary_transport/binary_messaging.rs` — `reap_finished` bails on `task.join().is_err()`, and the accept loop in `run_binary_stage` uses `?` on it. The iteration scheduler already contains worker panics instead of propagating them (`SchedulerWorker::run` wraps its loop in `catch_unwind` and emits `stage.scheduler_worker_panic` with a `failure_contained` attribute), so the transport was the odd one out.

The new regression test reproduces the scenario directly: a worker thread that panics is reaped without failing the caller, and the accept loop keeps going.

## Fix

- `reap_finished` no longer returns `Result`: it reaps finished workers and returns how many of them panicked. A panicked worker only fails its own connection.
- The accept loop emits `stage.connection_worker_panic` telemetry (with `llama_stage.failure_contained = true` and `llama_stage.panicked_workers = N`) when the count is non-zero, mirroring the iteration scheduler's containment event, and keeps accepting.
- `ConnectionWorkers::shutdown` is intentionally unchanged: it runs when serving is already over, and its error still surfaces worker panics in the final result.
- Enabling change so the crate's tests compile (and the new tests run) on Windows: `accepted_binary_stage_connection_is_blocking` uses `libc::fcntl`/`AsRawFd`, which don't exist on Windows — it is now `#[cfg(unix)]` with its imports moved into the test. Production code was already platform-gated; this test was the only hole.

No protocol or ABI surface is touched.

## Validation

- [x] `cargo test -p skippy-server --features dynamic-native-runtime --lib -- binary_transport` on Windows 11 (MSVC): 74 tests pass, including the two new `reap_finished` tests (also re-run in isolation).
- [x] One pre-existing test, `shutdown_tests::accept_error_still_closes_and_joins_active_connection_worker`, fails **deterministically on Windows** (cleanup timeout, 3/3 runs) — these lib tests simply never compiled on Windows before the `#[cfg(unix)]` commit, so this is the first time it could run there. It exercises `ConnectionWorkerControl::shutdown`'s socket-shutdown unblocking, which is production code, so it may be a real platform difference rather than a flaky test; left out of scope here, happy to dig into it as a follow-up issue.
- [x] Full `just test-all` was not run here (Windows dev-environment gaps are what my recent issues are about); happy to iterate if CI flags anything.
- [x] No UI changes.

Fixes #1473.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when connection workers encounter unexpected failures, allowing the accept loop to continue operating.
  * Added telemetry reporting for connection worker failures.
  * Ensured active workers remain unaffected during cleanup.

* **Tests**
  * Added coverage for recovering failed workers and preserving running workers.
  * Improved platform-specific test handling for connection behavior.

### DIFF
--- a/crates/skippy-server/src/binary_transport/binary_messaging.rs
+++ b/crates/skippy-server/src/binary_transport/binary_messaging.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::BTreeMap,
     future::Future,
     io::{self, Write},
     net::TcpListener,
@@ -96,19 +97,22 @@ impl ConnectionWorkers {
         self.0.push(worker);
     }
 
-    fn reap_finished(&mut self) -> Result<()> {
+    fn reap_finished(&mut self) -> usize {
+        let mut panicked = 0;
         let mut index = 0;
         while index < self.0.len() {
             if self.0[index].task.is_finished() {
                 let worker = self.0.swap_remove(index);
                 if worker.task.join().is_err() {
-                    bail!("binary stage connection worker panicked");
+                    // A panicked worker only fails its own connection; the
+                    // accept loop must keep serving other clients.
+                    panicked += 1;
                 }
             } else {
                 index += 1;
             }
         }
-        Ok(())
+        panicked
     }
 
     fn shutdown(mut self) -> Result<()> {
@@ -312,7 +316,22 @@ fn run_binary_stage(options: BinaryStageOptions, shutdown: Arc<AtomicBool>) -> R
 
     let accept_result = (|| -> Result<()> {
         while !shutdown.load(Ordering::SeqCst) {
-            connection_workers.reap_finished()?;
+            let panicked_workers = connection_workers.reap_finished();
+            if panicked_workers > 0 {
+                telemetry.emit(
+                    "stage.connection_worker_panic",
+                    BTreeMap::from([
+                        (
+                            "llama_stage.failure_contained".to_string(),
+                            json!(true),
+                        ),
+                        (
+                            "llama_stage.panicked_workers".to_string(),
+                            json!(panicked_workers),
+                        ),
+                    ]),
+                );
+            }
             let (mut upstream, _) = match listener.accept() {
                 Ok(conn) => conn,
                 Err(error) if error.kind() == io::ErrorKind::WouldBlock => {
@@ -435,7 +454,7 @@ mod shutdown_tests {
             mpsc,
         },
         thread,
-        time::Duration,
+        time::{Duration, Instant},
     };
 
     #[test]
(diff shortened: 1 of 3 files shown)


i386 (approved):
Reviewed the change — this is a clean, well-scoped fix. ✅

**What it does:** A single panicking connection worker used to `bail!` out of `reap_finished`, and the accept loop propagated it with `?`, taking down the whole binary-stage listener for every client. Now `reap_finished` returns the *count* of panicked workers, the accept loop emits `stage.connection_worker_panic` telemetry (with `failure_contained = true`) and keeps serving. A panic now only kills its own connection.

**Why I'm happy with it:**
- Mirrors the existing containment pattern already used by the iteration scheduler (`catch_unwind` + `stage.scheduler_worker_panic`), so the transport is no longer the odd one out.
- Panicked workers are still reaped (swap_remove happens before the join check), so no leak.
- `shutdown()` is deliberately left returning `Result` so a final panic still surfaces — good call keeping those two paths distinct.
- The two new tests pin both invariants: a panic is counted (not errored) and reaped, and a live worker is left alone.
- The `#[cfg(unix)]` gating on `accepted_binary_stage_connection_is_blocking` (fd/fcntl APIs) fixes a real Windows compile hole in the test module — nice cleanup to fold in.

The one pre-existing Windows failure you flagged (`accept_error_still_closes_and_joins_active_connection_worker`) is genuinely out of scope here since those lib tests never compiled on Windows before — a follow-up issue is the right home for it.

Approving.

Virgile-pct:
Thanks for the thorough review! The follow-up issue for that pre-existing Windows failure is already up: #1538, with a 40-line standalone repro isolating the socket-shutdown semantics.

i386 (approved):
Reviewed from a clean checkout of the PR head (fd25811c).

**Verdict: LGTM.** The fix does exactly what #1473 asked and matches the containment pattern the iteration scheduler already uses (`SchedulerWorker::run` → `catch_unwind` + `stage.scheduler_worker_panic` with a `failure_contained` attribute):

- `reap_finished` returning the panicked count instead of `bail!` is the right shape — a panicked worker fails only its own connection, and the accept loop keeps serving.
- `ConnectionWorkers::shutdown` intentionally keeping the error is correct: it runs post-serving where surfacing panics in the final result is the desired behavior, and the asymmetry is deliberate.
- The `stage.connection_worker_panic` telemetry follows the file's existing `llama_stage.*` attribute namespace and lifecycle event naming (`stage.binary_server_start` etc.).
- The `#[cfg(unix)]` gate on `accepted_binary_stage_connection_is_blocking` is an enabling fix only — production code was already platform-gated; this test was the one hole, and it's the reason these lib tests can now compile (and run) on Windows at all.
- Allowlist regeneration matches the shifted line numbers.

Local validation on macOS: full `skippy-server` lib suite green (531 passed), `cargo clippy -p skippy-server --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean.

One follow-up worth tracking separately (not blocking): the pre-existing deterministic Windows failure in `shutdown_tests::accept_error_still_closes_and_joins_active_connection_worker` the PR description calls out — it exercises `ConnectionWorkerControl::shutdown` socket-shutdown unblocking, which is production code, and this PR is the first time it could run on Windows.

PR #1046, opened 2026-07-21
fix: Windows Vulkan runtime dependencies

ndizazzo:
## Summary

- resolve the full non-system PE DLL dependency closure when packaging Windows Vulkan native runtimes
- load installed Windows runtime DLLs with sibling dependency lookup enabled
- verify packaged Vulkan dependency closure and exercise RX 9070 XT GPU enumeration
- run the resolver in Windows Vulkan CI when packaging inputs change

## Root cause

The Windows Vulkan artifact manifest listed llama.cpp runtime DLLs but omitted transitive MinGW dependencies imported by ggml-vulkan.dll. Windows also loaded the primary DLL without searching its installed runtime directory for sibling dependencies, so startup failed before Vulkan device enumeration and the system reported zero GPUs.

## Validation

- Python dependency resolver unit/integration suite
- Windows MSVC cross-target check for skippy-ffi dynamic loading
- focused skippy-ffi and mesh-llm-system checks, Clippy, and tests
- just test-all

Stacked on #1039 as requested.

Closes #1022

## Summary by CodeRabbit

* **New Features**
  * Improved Windows Vulkan native-runtime packaging by collecting and ordering required DLL dependencies.
  * Added Windows-native runtime dependency verification during artifact checks.
* **Bug Fixes**
  * CI change detection for Windows GPU builds now covers additional native-runtime packaging inputs and runs Vulkan-only dependency validation.
* **Improvements**
  * Enhanced dynamic runtime library loading on Windows for absolute-path DLLs.
* **Tests**
  * Added unit tests for Vulkan runtime-selectable GPU facts and comprehensive Windows native-runtime dependency discovery/verification.

Issue #1511, opened 2026-08-28
Windows: no CUDA path after install — `runtime install cuda` finds no compatible windows/x86_64 runtime in the release manifest


Virgile-pct:
## Environment

- Windows 11 Home 10.0.26200, Windows PowerShell 5.1
- NVIDIA GeForce RTX 4070 Ti (12 GB), driver 610.74 (CUDA UMD 13.3)
- mesh-llm 0.76.0-rc7 installed via `install.ps1 -PreRelease` (also reproduced with the stable 0.75.1 binary, see below)

## What happens

The Windows installer always installs the base product bundle (`mesh-llm-x86_64-pc-windows-msvc.zip`), which packages the **cpu** runtime. install.ps1's own `-Flavor` deprecation text says flavor selection now happens later:

> The Windows installer now installs the x64 product bundle and its packaged runtime; run `mesh-llm.exe setup` to select the recommended runtime.

But that later step has nothing to offer on Windows:

```text
> mesh-llm runtime install cuda
🔎 Loading native runtime release manifest
🔎 Detecting host runtime profile
❌ Native runtime install failed
   Reason: no compatible native runtime found for Skippy ABI 0.1.39 on windows/x86_64
```

```text
> mesh-llm runtime list --available
📦 Available native runtimes
  - ✅ meshllm-native-runtime-windows-x86_64-cpu compatible (cpu, windows/x86_64)
    MeshLLM: 0.76.0-rc7; Skippy ABI: 0.1.39
```

Only the cpu runtime is resolvable. Same behavior with the stable 0.75.1 binary (only `meshllm-native-runtime-windows-x86_64-cpu`, Skippy ABI 0.1.35).

Meanwhile the GPU itself is detected fine (`mesh-llm gpus` reports the RTX 4070 Ti as `Backend device: CUDA0`), and `doctor` lists `detected flavors: cpu, cuda, rocm, vulkan`.

## The compatible artifact exists — it's just not reachable through the manifest

The `mesh-llm-x86_64-pc-windows-msvc-cuda.zip` product bundle for the same tag contains `meshllm-native-runtime-windows-x86_64-cuda12` declaring **Skippy ABI 0.1.39** — exactly what the host asks for. Importing it locally works:

```powershell
# extract mesh-llm-v0.76.0-rc7-x86_64-pc-windows-msvc-cuda.zip, then:
mesh-llm runtime install --bundle-dir <extracted>\mesh-bundle cuda
[...]

Issue #1510, opened 2026-08-28
Windows: stable v0.75.1 bundles still fail install.ps1 with `runtime.sha256 mismatch` — assets predate the #1271 fix and were never re-released


Virgile-pct:
## Environment

- Windows 11 Home 10.0.26200, Windows PowerShell 5.1
- mesh-llm stable channel (`releases/latest` → v0.75.1, published 2026-08-10)

## What happens

The documented Windows install path fails today for every user:

```powershell
irm https://raw.githubusercontent.com/Mesh-LLM/mesh-llm/main/install.ps1 | iex
```

```text
Installing Windows x64 MeshLLM product bundle
Release channel: stable
Downloading https://github.com/Mesh-LLM/mesh-llm/releases/latest/download/mesh-llm-x86_64-pc-windows-msvc.zip
Verified checksum: mesh-llm-x86_64-pc-windows-msvc.zip
runtime.sha256 mismatch: expected 80856c2a6d67d8b23dd743d37b2b14cf91433b6367941a01695fe9ace5106941, got 6b24cc3f23fec5deaad23dd60ac89f872932bfd13d8fc7ed124bff8ad676deb5
```

Note the order: the archive's own `.sha256` sidecar **verifies fine** — the failure is the bundle's *internal* consistency check (`product-manifest.json` → `runtime.sha256` vs the recomputed tree hash of `native-runtimes/<id>/`).

This is the same failure mode as #1219 (filed against v0.75.0 and closed via #1271), but it still reproduces on the current stable assets.

## Why: the v0.75.1 assets were composed before the #1271 fix landed

Timeline:

- 2026-08-10 — v0.75.1 published (current `latest`)
- 2026-08-13 — #1271 merged, fixing the file ordering in `scripts/compose-product-bundle.py` (default `sorted()` over `rglob` → ordinal sort of POSIX relative paths)
- The v0.75.1 Windows assets were never re-composed/re-uploaded, so they still carry a `runtime.sha256` computed with the old ordering, which can never match what `install.ps1` recomputes.

I verified the assets are otherwise intact — this is purely the known ordering bug, not corruption:

<details>
<summary>Forensic check on mesh-llm-v0.75.1-x86_64-pc-windows-msvc.zip (click to expand)</summary>

1. Downloaded the archive + sidecar manually; SHA-256 matches the published sidecar (`497004a7f524412d80362ae7f47160e3f33e917e77812e8e49572052c126b16d`), so these are the original bytes as published.
[...]

PR #1537, opened 2026-08-29
fix(skippy-ffi): normalize native runtime library paths before LoadLibraryExW

### LINKED ISSUE #1512: Windows: native runtime never loads — serve dies with LoadLibraryExW error 126, dependency search behaves as if LOAD_WITH_ALTERED_SEARCH_PATH were not applied
## Environment

- Windows 11 Home 10.0.26200, Windows PowerShell 5.1, clean interactive user session (no container, no special DLL policy)
- NVIDIA GeForce RTX 4070 Ti (12 GB), driver 610.74 (CUDA UMD 13.3)
- mesh-llm 0.76.0-rc7 (`install.ps1 -PreRelease`), runtime `meshllm-native-runtime-windows-x86_64-cuda12` installed via `runtime install --bundle-dir` (needed because of the release-manifest gap, see #1511)

## What happens

```text
> mesh-llm serve --local-model-only --model "<abs path>\Qwen3-8B-Q4_K_M.gguf"
failed to load native runtime library [REDACTED_PATH] LoadLibraryExW failed (OS error 126: Le module spécifié est introuvable. (os error 126)): load native runtime meshllm-native-runtime-windows-x86_64-cuda12 from [REDACTED_PATH]
```

Error 126 with every flavor I could get selected (the cpu flavor produced the identical failure), so as far as I can tell no native runtime can currently be loaded by the shipped Windows binary.

## What I ruled out

All the runtime's files are present and individually loadable — the failure is purely dependency search:

- Every file in the runtime tree matches the SHA-256s declared in its own `manifest.json`.
- P/Invoke experiments on the same DLLs (one `LoadLibraryExW` per process):
  - `LoadLibraryExW(abs_path, 0, 0)` on `ggml-cuda.dll` → **error 126** (dependencies like `cudart64_12.dll` sit next to it, but the default search never looks there)
  - `LoadLibraryExW(abs_path, 0, LOAD_WITH_ALTERED_SEARCH_PATH)` → loads fine
  - same with `LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS`, and with `\\?\`-prefixed paths → loads fine
- **Replaying mesh-llm's exact load plan works**: loading all 11 DLLs of the cuda12 runtime in the `manifest.json` `libraries` order, each by absolute cache path with `LOAD_WITH_ALTERED_SEARCH_PATH`, no unload in between → every single one loads, no PATH help needed.

So a loader that actually applies `LOAD_WITH_ALTERED_SEARCH_PATH` succeeds on this exact runtime tree. The shipped binary's behavior instead matches plain `LoadLibraryExW(path, 0, 0)` semantics exactly — see the workaround below: making dependencies resolvable via `PATH` (which the legacy default search consults, but the altered-search flow wouldn't need) is sufficient to fix it.

## The confusing part

`crates/skippy-ffi/src/dynamic_library.rs` **already** passes `LOAD_WITH_ALTERED_SEARCH_PATH` on Windows — since #1046 (cf2d6ad, 2026-07-22), and `git show v0.76.0-rc7:crates/skippy-ffi/src/dynamic_library.rs` confirms the flag is present at the release tag. Yet the shipped v0.76.0-rc7 Windows binary behaves as if the flag were absent.

Two hypotheses I couldn't discriminate from outside:

1. The Windows release build ships a stale/differently-configured `skippy-ffi` (the error text — `failed to load native runtime library …` + `load native runtime {id} from {root}` — matches the `dynamic_library.rs` + `mesh-llm-host-runtime` chain, so the code path is that one, just seemingly without the flag's effect).
2. Some other un-flagged load happens first and produces an identical message chain (e.g. `llama-quant-ffi`'s `Symbols::load_paths` uses a plain `Library::new`, which on Windows means no altered search path).

A related papercut that made this much harder to pin down: the console error redacts the library path (`[REDACTED_PATH]`), which is the one piece of information needed to identify the failing DLL. Keeping at least the filename under redaction would help a lot.

## Verified workaround (and proof the rest works)

Prepending the runtime's `lib/` directory to `PATH` before `serve` unblocks everything:

```powershell
$env:PATH = "$env:LOCALAPPDATA\mesh-llm\native-runtimes\0.76.0-rc7\meshllm-native-runtime-windows-x86_64-cuda12\lib;" + $env:PATH
mesh-llm serve --local-model-only --model "<abs path>\Qwen3-8B-Q4_K_M.gguf"
# → api ready at http://127.0.0.1:9337/v1
```

With that, native Windows CUDA inference works: ~60 tok/s on Qwen3-8B Q4_K_M on this 4070 Ti, on a CUDA 13.3 driver. Which also means the README's "CUDA 13.x → use WSL2" guidance is obsolete once loading is fixed — happy to send a docs PR for a native Windows section once this settles.

## Side observations from the same session

- `--llama-flavor cpu` did not override runtime selection: with cuda12 installed, serve still tried (and failed) to load cuda12.
- `mesh-llm stop` reports "Nothing running" while a `serve --local-model-only` instance is alive and answering on port 9337.

Happy to run any diagnostic build or extra experiment on this machine — it reproduces 100% of the time.

Virgile-pct:
## Original problem

#1512: on native Windows, no native runtime can load — `serve` dies with `LoadLibraryExW failed (OS error 126)` for every flavor (cpu included), unless the runtime's `lib` directory happens to be on `PATH`.

## Diagnostics

Building the host from source made the failing path visible past the console redaction:

```text
[DIAG dynamic_library::load] flags=LOAD_WITH_ALTERED_SEARCH_PATH path=...\meshllm-native-runtime-windows-x86_64-cuda12\lib/cublas64_12.dll ok=false
```

Note the tail: **`lib/cublas64_12.dll`** — runtime manifests declare libraries with POSIX separators, `NativeRuntimeLoadPlan` joins them onto the runtime root with `Path::join`, and the mixed-separator path goes straight into `LoadLibraryExW`. Isolated with one `LoadLibraryExW` call per process on the same existing file:

| Path passed | `LOAD_WITH_ALTERED_SEARCH_PATH` | Result |
|---|---|---|
| `...\lib/cublas64_12.dll` (mixed) | yes | **error 126** |
| `...\lib\cublas64_12.dll` (backslashes) | yes | loads |
| `...\lib/cublas64_12.dll` (mixed) | no (flags=0) | error 126 |

So the flag introduced in #1046 is present but defeated: with a mixed-separator path the loader neither resolves the file as given nor derives the dependency directory. It also explains the `PATH` workaround — the failed absolute resolution falls back to a name search, which succeeds once the `lib` directory is on `PATH`. `std::fs` APIs accept forward slashes fine, which is why the load plan's `is_file()` validation passes and the failure only shows up at `LoadLibraryExW` time.

This corrects the stale-build-cache hypothesis I posted on the issue — a host built from today's `main` reproduces the failure identically, so it was never a build-pipeline problem.

## Fix

In `skippy-ffi`'s Windows `dynamic_library::load`, rebuild the incoming path from its components before calling `LoadLibraryExW` — `Path::components().collect::<PathBuf>()` normalizes every separator to a backslash, losslessly. One normalization point covers every load-plan caller. A unit test pins the normalization invariant the fix relies on.

## Validation

- [x] `cargo test -p skippy-ffi --features dynamic-runtime` (Windows, MSVC): passes, including the new normalization test.
- [x] End-to-end on the repro machine (RTX 4070 Ti, driver 610.74/CUDA 13.3): host built from this branch, cuda12 runtime installed via `runtime install --bundle-dir`, then `serve --local-model-only` **without any `PATH` workaround** → runtime loads, `api ready at http://127.0.0.1:9337/v1`, chat completions served on GPU.
- [x] No UI changes.

Fixes #1512.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows library loading by normalizing mixed path separators.
  * Fixed dependency resolution for library paths using POSIX-style separators on Windows.
  * Added coverage to verify path normalization behavior.

### DIFF
--- a/crates/skippy-ffi/src/dynamic_library.rs
+++ b/crates/skippy-ffi/src/dynamic_library.rs
@@ -49,7 +49,15 @@ pub(crate) unsafe fn load(path: &Path) -> Result<Library, String> {
     // dependent DLL, even when the requested DLL has an absolute path. Native
     // runtimes are installed elsewhere, so make the loaded DLL's directory the
     // first dependency search location.
-    let result = unsafe { WindowsLibrary::load_with_flags(path, LOAD_WITH_ALTERED_SEARCH_PATH) };
+    // Native runtime manifests declare their libraries with POSIX separators
+    // and Path::join keeps them, so runtime library paths reach this point
+    // with a mixed-separator tail (`...\lib/llama.dll`). LoadLibraryExW fails
+    // those with ERROR_MOD_NOT_FOUND even though the file exists, and
+    // LOAD_WITH_ALTERED_SEARCH_PATH cannot derive the dependency directory
+    // from them. Rebuilding the path from its components normalizes every
+    // separator to a backslash.
+    let normalized: std::path::PathBuf = path.components().collect();
+    let result = unsafe { WindowsLibrary::load_with_flags(&normalized, LOAD_WITH_ALTERED_SEARCH_PATH) };
     result
         .map(Into::into)
         .map_err(|error| format_load_error(path, &error))
@@ -96,6 +104,17 @@ mod tests {
 
         assert!(message.contains("OS error unavailable"));
         assert!(!message.contains("OS error 0"));
+    }
+
+    #[cfg(target_os = "windows")]
+    #[test]
+    fn component_collection_normalizes_posix_separators() {
+        // The windows load() relies on this: a manifest-derived path with a
+        // POSIX tail must come out with backslashes only, or LoadLibraryExW
+        // rejects it with ERROR_MOD_NOT_FOUND.
+        let mixed = Path::new(r"C:\runtime\lib/llama.dll");
+        let normalized: std::path::PathBuf = mixed.components().collect();
+        assert_eq!(normalized, Path::new(r"C:\runtime\lib\llama.dll"));
     }
 
     #[cfg(target_os = "windows")]


i386:
@Virgile-pct congratulations on your first contribution to Mesh LLM! We will get your PRs reviewed and squared away shortly 🎉

i386 (approved):
Reviewed — tight, correct, one-point fix. ✅

**Root cause is well-diagnosed:** runtime manifests declare libs with POSIX separators, `Path::join` preserves them, so a mixed-separator tail (`...\lib/llama.dll`) reaches `LoadLibraryExW`, which fails with error 126 and also defeats `LOAD_WITH_ALTERED_SEARCH_PATH` (it can't derive the dependency directory). The `std::fs` `is_file()` check passes on forward slashes, which is exactly why the failure only shows up at load time.

**The fix** — `path.components().collect::<PathBuf>()` — is the right minimal normalization: `Components` re-emits each segment and `PathBuf` joins with the platform separator, so every `/` becomes `\` losslessly, at the single load-plan chokepoint. Keeping the original `path` in `format_load_error` is correct too (error messages should reflect what was requested).

The unit test pins the exact invariant the fix relies on. No ABI/protocol surface touched.

Approving.

Virgile-pct:
Thank you! Windows turned out to be a fun territory to map and I'm happy to keep digging. The remaining issues (#1510, #1511, #1538) all have full repros ready whenever the team gets to them.

i386:
@Virgile-pct you may want to try out the latest rc as well

Virgile-pct:
Will do! I'll re-run the full native Windows sequence on the next rc as soon as it lands and report back on the issues.

Virgile-pct:
Small coordination note: I'm based in France, so expect some timezone lag. More importantly: I love what you're building here, that's why I want to help however I can. The Windows side of Mesh LLM will keep getting my attention.

PR #1539, opened 2026-08-29
ci(llama): honest canary colors, linked repair PRs, Buzz notify

i386:
## Why

Run [33247859447](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33247859447) looked like "failed, never produced a PR" — but the repair had fully succeeded: PR #1533 existed, green, waiting for merge. Two reporting gaps caused it:

1. The final step ("Fail when the canary needs human attention") exits 1 for **any** repair outcome, even fully certified ones.
2. The summary named only the branch `llama-canary/patch-queue-fix` — no PR link. A bot-authored PR generates no GitHub notification, so nobody was told.

## What changes

- **Honest colors.** A new outcome step classifies the run: repair loop exit 0 → run is **green**, summary = "certified repair ready to merge, PR #N". Stuck repairs (or repairs that could not run) stay **red** with the same links.
- **PR links everywhere.** The repair PR URL/number are looked up (under `CANARY_REPAIR_TOKEN`; the job stays `contents: read`) and appear in the summary and the notification.
- **Buzz notify.** Both outcomes post to **#mesh-dev** via `scripts/llama-canary-buzz-notify.py` — a stdlib-only BIP-340 + NIP-98 signer with curl transport (Cloudflare rejects Python TLS fingerprints — verified live). Best-effort: notification failures never fail the canary.

## Setup required (repo admin, one-time)

| Secret / variable | Value |
|---|---|
| `BUZZ_RELAY_KEY` | Nostr private key (hex or nsec) for the notify identity — key generated, never committed: `4454a64b…d5adb` is its pubkey |
| `BUZZ_RELAY_AUTH_TAG` | *(optional)* NIP-OA owner-attestation tag if the key isn't a direct relay member |
| `LLAMA_CANARY_BUZZ_CHANNEL` | *(optional)* channel UUID override; default is #mesh-dev `a9ea5982-…` |

The notify identity also needs channel membership in #mesh-dev (done) and, if the relay enforces its roster, a relay-membership grant — say the word and I'll coordinate it.

## Verification

- `just ci-validate` components: actionlint ✓, `git diff --check` ✓, full contract-test discovery **673 pass / 7 skip** (in a pyyaml venv; the 6 import errors on a bare box are a pre-existing env gap, confirmed on clean main)
- `just ci-crate-lists`, `just no-console-print`, `just publish-crates`, `just check-release` ✓
- Notify script proven live against the production relay (accepted:true, 200) before commit
- New tests: outcome-step contract, signer roundtrip, canonical event-id hash, hex/nsec decode, key-never-printed

Fixes the reporting policy James set in skippy-testing (2026-08-29): if the canary fails in any way it creates a PR and tells us why; if it succeeds it creates a PR and tells us.


michaelneale (changes requested):
The post-green review currently breaks the certification boundary this workflow is meant to enforce.

`publish_repair_branch` records `CERTIFIED_SHA` only after the family battery has passed. But `post_green_review_turn` then explicitly lets a fresh agent modify or commit the checkout and force-push that different HEAD without rerunning `scripts/skippy-family-battery.sh`. `verify_pr_head_is_certified` does not catch this: it deliberately changes the expected SHA to `REVIEW_HEAD`. The success comment likewise says the review bytes will only be re-certified after this PR merges and the next canary runs.

That means the branch presented as a green, “certified repair ready to merge” can contain agent-authored bytes which were never certified before the current merge. A later canary is not a safety boundary for code already merged, so the comment that review modifications “can never bypass certification” is not true for the merge being requested.

Please either make the post-green review read-only (report findings and leave the certified tree unchanged), or rerun the complete family battery after every review modification and update `CERTIFIED_SHA` only after that modified HEAD passes. The success outcome and PR-head verification should bind to the exact SHA that passed the battery.

PR #1542, opened 2026-08-29
Keep staged llama.cpp mappings alive until every retained tensor is loaded

### LINKED ISSUE #1298: 🤖 Filtered stage load SIGSEGVs with exactly 8 visible CUDA devices (1-7 and full unfiltered 8 all work)

IvGolovach:
## Summary

Staged Skippy loads now keep GGUF-backed mappings alive until every tensor retained by every backend context has actually been uploaded. Filtered and multi-context loads no longer report exact completion early or release mapped model data while a later context still needs it.

Closes #1298.

## Why

The old loader adjusted `size_data` while tensors were requested. That request-oriented total could diverge from the work later performed by `load_all_data()`:

- one physical GGUF weight can be retained by more than one backend context;
- synthetic context tensors have no GGUF weight and must not extend mapping lifetime;
- quantization initializes mappings before model contexts exist.

An undercount could make progress reach `1.0` and trigger mapping cleanup before the final retained context had loaded its data.

## Diff scope

- Add durable llama.cpp patch `0035-skippy-account-retained-tensors-during-staged-loads.patch`.
- Recompute retained byte totals from finalized backend contexts, while preserving the contextless `weights_map` fallback used by quantization.
- Reserve exact progress `1.0` for the incomplete-to-complete transition and prevent repeated completion cleanup.
- Add generated dense Gemma and Qwen2MoE fixtures plus a native accounting regression to the supported `LLAMA_STAGE_BUILD_TESTS` lane.
- Make `just skippy-native-tests` build and execute all `skippy_*` CTest cases.
- Scope Git whitespace handling to mail-format files inside `third_party/llama.cpp/patches/`; no root-level ownership entry is introduced.

No Rust API, CLI, UI, database, configuration, release, mesh-wire, plugin-protocol, or model-package contract changes.

## Architecture

`init_mappings()` now derives `size_data` from the finalized `ctx_map`, counting each retained tensor occurrence only when `get_weight()` resolves it to real GGUF storage. This mirrors the context traversal later used by `load_all_data()`. If no contexts exist, the existing `weights_map` total remains the quantization fallback.

Intermediate callbacks are clamped below `1.0`; final cleanup runs only when `size_done` crosses from incomplete to exactly complete. The existing `size_done <= size_data` invariant is asserted.

## Protocol and compatibility

Not applicable — no mesh wire format, plugin protocol, Skippy ABI surface, or ABI version changed. The patch remains additive to the existing pinned llama.cpp queue and preserves contextless quantization behavior.

## Review follow-up

All requested changes from the review on `0877468a` are implemented on the final SHA:

- Construct the current pinned `ctx_key { buft, lazy }` shape explicitly.
- Register and run the regression under `LLAMA_STAGE_BUILD_TESTS` through the normal CTest lane.
- Rebase onto current `main` and renumber the new patch from the colliding `0033` to `0035`.
- Exercise a CPU-only two-context sequential load: progress stays below completion after the first context, reaches exact completion after the second, and performs cleanup once.
- Include a synthetic context tensor for which `get_weight()` returns null, plus a post-completion context proving cleanup is not repeated.
- Cover real filtered dense and MoE loads with generated fixtures.

The later CI-planner finding is also resolved without changing the protected ownership catalog: the attributes file now lives inside the already-owned `third_party/llama.cpp/patches/**` path. The full planner test file passes.

## Branch integrity

- Base: `main@a6b16e62220a4ae98849959f5f526662906e7b47`
- Head: `9ef85e76b46d81b5cc1ad31e5edd70b8901141dd`
- Merge base: `a6b16e62220a4ae98849959f5f526662906e7b47`
- `origin/main` is an ancestor of `HEAD`.

## Commit integrity

- One introduced commit: `9ef85e76b46d81b5cc1ad31e5edd70b8901141dd Fix retained staged tensor accounting`
- One logical change.
- Final diff contains only the intended native patch, its supported test-lane wiring, and patch-directory attributes.

Ledger: not applicable — not required for this change family.  
Version: not changed — `RELEASE.md` assigns release version synchronization to `scripts/release-version.sh`; no ABI or protocol version changed.

## Diff hygiene

```text
M scripts/build-llama.sh
A third_party/llama.cpp/patches/.gitattributes
A third_party/llama.cpp/patches/0035-skippy-account-retained-tensors-during-staged-loads.patch
```

- `git diff --check origin/main...HEAD`: **PASS**, no output.
- The scoped `.gitattributes` disables whitespace interpretation only for mail-format `*.patch` payloads in the patch queue; its path is covered by the existing `native-abi` ownership rule.
- No secrets, credentials, `.env` files, local configuration, caches, binaries, coverage artifacts, or unrelated generated files are present.

## Validation mode and proof

Validation mode: **Tier 4 — native verification tooling**, with **Tier 3 runtime/lifetime risk** covered by dedicated dense, MoE, sequential-context, and cleanup-transition tests.

Local proof on the final content:

- `just skippy-native-tests`: **PASS** — clean replay of all 35 patches onto llama.cpp `9723942adc518b43c4b95dc4dce6906903eb5e09`; 5/5 CTest cases passed.
- Final replayed llama.cpp SHA: `7eda86dd17e8411bb739653c1c9435137eb48b32`.
- CTest: hardware probe, Gemma fixture generation, Qwen2MoE fixture generation, dense accounting, and MoE accounting: **PASS**.
- `mise exec node@22.22.0 -- just build`: **PASS** — UI, shipped host, and Metal native runtime built.
- `bash -n scripts/build-llama.sh`: **PASS**.
- `python3 scripts/tests/test_justfile_layout.py`: **PASS** — 3 passed.
- `python3 scripts/tests/test_plan_ci.py`: **PASS** — 31 passed.
- Final PR diff through `scripts/plan-ci.py`: **PASS** — domains `tooling`, `runtime-product`, and `native-abi`; required slices include quality, static ABI, Rust tests, runtime product, and product smoke.
- `cargo fmt --all -- --check`: **PASS**.
- `cargo test -p skippy-runtime --lib`: **PASS** — 100 passed.
- `cargo test -p skippy-server --lib`: **PASS** — 553 passed, 3 ignored.
- `cargo test -p mesh-llm-host-runtime --lib inference::skippy`: **PASS** — 251 passed, 2 ignored, 2583 filtered.
- `git diff --check` and pre-commit `git diff --cached --check`: **PASS**.

The build/test logs contain pre-existing warnings in untouched Rust, Inkling, and upstream generator code; this PR introduces no known warning in changed runtime code.

Not run locally:

- CUDA/V100 reproduction — unavailable on this machine; remains a hardware-specific gate.
- Full workspace test suite — not required locally for this focused validation tier; mandatory remote CI supplied the final full-SHA proof.
- `just ci-validate` — no workflow, local action, planner, ownership catalog, slice catalog, or routing rule changed; the directly owning planner suite and final-diff plan were run instead.

## Remote CI and remaining gates

Mandatory PR CI on final SHA `9ef85e76b46d81b5cc1ad31e5edd70b8901141dd`: **PASS**.

- GitHub Actions check-runs: **33 passed, 12 expected skips, 0 failed, 0 pending**.
- Stable results `PR / Quality`, `PR / Website`, `PR / Linux`, `PR / macOS`, and `PR / Windows`: **PASS**.
- Linux host, static ABI, native runtime, four Rust-test batches, composed CPU product, core inference, two-node client, and two-node split smoke: **PASS**.
- CodeRabbit final-SHA check: **PASS**; no new inline findings were posted.

Remaining gates:

- maintainer re-review clearing the earlier `CHANGES_REQUESTED` state from `0877468a`;
- requested CUDA/V100 hardware reproduction (the CUDA smoke row was not selected by this PR plan).

GitHub reports the PR as mergeable with a clean merge state, but it is **not merge-ready** while those two required gates remain open.

CI context names are unchanged. No CI/workflow context changed.

## Runtime safety

- No new locks, queues, threads, background tasks, or unbounded allocations.
- Exact completion is tied to the final retained byte transition.
- Mapping cleanup is one-shot and cannot occur after only the first retained context.
- Synthetic non-weight tensors do not inflate lifetime accounting.

## Migration and documentation

Not applicable — no migration, data repair, operational command, runbook, or user-facing documentation changed.

## Rollback plan

Rollback: revert this PR.  
DB downgrade: not applicable.  
Data repair: not applicable.  
Operational caveat: restart any model load already in progress after rollback.

## Known residual risks

- Local native coverage is CPU/Metal; CUDA/V100 behavior still requires the requested hardware proof.
- Maintainer re-review and requested CUDA/V100 proof remain pending; mandatory remote CI is green on the final SHA.

### DIFF
--- a/scripts/build-llama.sh
+++ b/scripts/build-llama.sh
@@ -364,13 +364,17 @@ cmake "${CMAKE_ARGS[@]}"
 
 BUILD_TARGETS=(llama llama-common mtmd)
 if [[ "${LLAMA_STAGE_BUILD_TESTS:-OFF}" == "ON" ]]; then
-  BUILD_TARGETS+=(skippy-hardware-application-probe)
+  BUILD_TARGETS+=(
+    skippy-hardware-application-probe
+    skippy-model-fixture-generator
+    skippy-model-loader-accounting
+  )
 fi
 
 cmake --build "$LLAMA_BUILD_DIR" --config "${CMAKE_BUILD_TYPE:-Release}" --parallel "$(detect_jobs)" --target "${BUILD_TARGETS[@]}"
 
 if [[ "${LLAMA_STAGE_BUILD_TESTS:-OFF}" == "ON" ]]; then
-  ctest --test-dir "$LLAMA_BUILD_DIR" --build-config "${CMAKE_BUILD_TYPE:-Release}" --output-on-failure -R '^skippy_hardware_application_probe$'
+  ctest --test-dir "$LLAMA_BUILD_DIR" --build-config "${CMAKE_BUILD_TYPE:-Release}" --output-on-failure -R '^skippy_'
 fi
 
 printf '%s\n' "$CURRENT_BUILD_STAMP" > "$BUILD_STAMP"
--- a/third_party/llama.cpp/patches/.gitattributes
+++ b/third_party/llama.cpp/patches/.gitattributes
@@ -1,0 +1,1 @@
+*.patch -whitespace
--- a/third_party/llama.cpp/patches/0035-skippy-account-retained-tensors-during-staged-loads.patch
+++ b/third_party/llama.cpp/patches/0035-skippy-account-retained-tensors-during-staged-loads.patch
@@ -1,0 +1,458 @@
+From ce95c713e0b881339653b4893458ada67c25d7dd Mon Sep 17 00:00:00 2001
+From: Mesh-LLM CI <ci@mesh-llm.local>
+Date: Sat, 29 Aug 2026 01:31:34 -0700
+Subject: [PATCH] skippy: account retained tensors during staged loads
+
+---
+ src/CMakeLists.txt                            |  46 +++
+ src/llama-model-loader.cpp                    |  43 ++-
+ tests/CMakeLists.txt                          |  16 +
+ tests/test-skippy-model-loader-accounting.cpp | 286 ++++++++++++++++++
+ 4 files changed, 379 insertions(+), 12 deletions(-)
+ create mode 100644 tests/test-skippy-model-loader-accounting.cpp
+
+diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
+index b59528d62..981f24b1a 100644
+--- a/src/CMakeLists.txt
++++ b/src/CMakeLists.txt
+@@ -91,2 +91,48 @@ if (LLAMA_STAGE_BUILD_TESTS)
+         COMMAND skippy-hardware-application-probe)
++
++    set(SKIPPY_MODEL_FIXTURE_DIR "${CMAKE_CURRENT_BINARY_DIR}/skippy/test-models")
++    file(MAKE_DIRECTORY "${SKIPPY_MODEL_FIXTURE_DIR}")
++
++    add_executable(skippy-model-fixture-generator
++        ../tests/test-llama-archs.cpp)
++    target_compile_features(skippy-model-fixture-generator PRIVATE cxx_std_17)
++    target_link_libraries(skippy-model-fixture-generator PRIVATE llama-common)
++
++    add_executable(skippy-model-loader-accounting
++        ../tests/test-skippy-model-loader-accounting.cpp)
++    target_include_directories(skippy-model-loader-accounting PRIVATE .)
++    target_compile_features(skippy-model-loader-accounting PRIVATE cxx_std_17)
++    target_link_libraries(skippy-model-loader-accounting PRIVATE llama-common)
++
++    add_test(
++        NAME skippy_generate_gemma_fixture
++        COMMAND skippy-model-fixture-generator
++            --arch gemma
++            --seed 1
++            --out "${SKIPPY_MODEL_FIXTURE_DIR}")
++    set_tests_properties(skippy_generate_gemma_fixture PROPERTIES
++        FIXTURES_SETUP skippy-model-loader-fixtures)
++
++    add_test(
++        NAME skippy_generate_qwen2moe_fixture
++        COMMAND skippy-model-fixture-generator
++            --arch qwen2moe
++            --seed 1
++            --out "${SKIPPY_MODEL_FIXTURE_DIR}")
++    set_tests_properties(skippy_generate_qwen2moe_fixture PROPERTIES
++        FIXTURES_SETUP skippy-model-loader-fixtures)
++
++    add_test(
++        NAME skippy_model_loader_accounting_dense
++        COMMAND skippy-model-loader-accounting
++            "${SKIPPY_MODEL_FIXTURE_DIR}/gemma-dense.gguf")
++    set_tests_properties(skippy_model_loader_accounting_dense PROPERTIES
++        FIXTURES_REQUIRED skippy-model-loader-fixtures)
++
++    add_test(
++        NAME skippy_model_loader_accounting_moe
++        COMMAND skippy-model-loader-accounting
++            "${SKIPPY_MODEL_FIXTURE_DIR}/qwen2moe-moe.gguf")
++    set_tests_properties(skippy_model_loader_accounting_moe PROPERTIES
++        FIXTURES_REQUIRED skippy-model-loader-fixtures)
+ endif()
+diff --git a/src/llama-model-loader.cpp b/src/llama-model-loader.cpp
+index b831fa385..bde67cf23 100644
+--- a/src/llama-model-loader.cpp
++++ b/src/llama-model-loader.cpp
+@@ -11,2 +11,3 @@
+ #include <cinttypes>
++#include <cmath>
+ #include <cstdint>
+@@ -1313,3 +1314,2 @@ struct ggml_tensor * llama_model_loader::create_tensor(
+                 if (requested_tensor_exists && first_filtered_request) {
+-                    size_data -= nbytes;
+                     if (!(flags & TENSOR_DUPLICATED)) {
+@@ -1364,3 +1364,2 @@ struct ggml_tensor * llama_model_loader::create_tensor(
+ 
+-            size_data -= nbytes;
+             n_created++;
+@@ -1570,4 +1569,2 @@ struct ggml_tensor * llama_model_loader::create_tensor(
+ 
+-    const bool duplicated = flags & TENSOR_DUPLICATED;
+-
+     struct ggml_tensor * tensor = ggml_dup_tensor(ctx, &t_meta);
+@@ -1575,5 +1572,3 @@ struct ggml_tensor * llama_model_loader::create_tensor(
+ 
+-    if (duplicated) {
+-        size_data += ggml_nbytes(&t_meta);
+-    } else {
++    if (!(flags & TENSOR_DUPLICATED)) {
+         n_created++;
+@@ -1635,5 +1630,24 @@ void llama_model_loader::init_mappings(bool prefetch, llama_mlocks * mlock_mmaps
+ 
+-    // compute the total size of all tensors for progress reporting
+-    for (const auto & it : weights_map) {
+-        size_data += ggml_nbytes(it.second.tensor);
++    // Match load_all_data exactly once model contexts exist. A physical weight
++    // can appear in more than one context (for example, tied embeddings used as
++    // the output head), and load_all_data uploads it once per context. Counting
++    // logical create_tensor requests instead can therefore finish early and
++    // unmap weights that a later context still has to upload.
++    size_data = 0;
++    if (!ctx_map.empty()) {
++        for (const auto & [_, ctx] : ctx_map) {
++            for (ggml_tensor * tensor = ggml_get_first_tensor(ctx.get());
++                    tensor != nullptr;
++                    tensor = ggml_get_next_tensor(ctx.get(), tensor)) {
++                if (get_weight(ggml_get_name(tensor)) != nullptr) {
++                    size_data += ggml_nbytes(tensor);
++                }
++            }
++        }
++    } else {
++        // Quantization initializes mappings before constructing model contexts
++        // and reads weights through load_data_range rather than load_all_data.
++        for (const auto & [_, weight] : weights_map) {
++            size_data += ggml_nbytes(weight.tensor);
++        }
+     }
+@@ -1698,2 +1712,3 @@ bool llama_model_loader::load_all_data(
+     GGML_ASSERT(size_data != 0 && "call init_mappings() first");
++    const size_t size_done_before = size_done;
+ 
+@@ -1808,3 +1823,6 @@ bool llama_model_loader::load_all_data(
+         if (progress_callback) {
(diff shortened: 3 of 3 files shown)


ndizazzo (changes requested):
Requesting changes. The accounting formula looks right: summing each retained context tensor that resolves to a GGUF weight matches `load_all_data()`, and the `weights_map` fallback keeps the quantization path intact. I don't think this is ready to merge yet.

1. The new test no longer compiles against the llama.cpp pin on current `main`. It inserts `ggml_backend_cpu_buffer_type()` directly into `ctx_map` at [patch line 194](https://github.com/Mesh-LLM/mesh-llm/blob/0877468a9de7db0cf4a3b045a7033a355c1addd4/third_party/llama.cpp/patches/0033-skippy-account-retained-tensors-during-staged-loads.patch#L194), but the map key is now `ctx_key { buft, lazy }` in [the pinned upstream header](https://github.com/ggml-org/llama.cpp/blob/9723942adc518b43c4b95dc4dce6906903eb5e09/src/llama-model-loader.h#L151-L165). Construct the key explicitly, including `lazy = false`.

2. The regression test isn't part of any supported mesh-llm test lane. The patch registers it under `LLAMA_BUILD_TESTS`, while [the native build sets that option to `OFF`](https://github.com/Mesh-LLM/mesh-llm/blob/582aff67847d94e326ab672faa5a2ae9703e4ee5/scripts/build-llama.sh#L194-L200). `just skippy-native-tests` enables `LLAMA_STAGE_BUILD_TESTS`, and [the script builds and runs only the hardware probe](https://github.com/Mesh-LLM/mesh-llm/blob/582aff67847d94e326ab672faa5a2ae9703e4ee5/scripts/build-llama.sh#L365-L374). I ran that lane after replaying this patch on current `main`; it reported one test, but the accounting test wasn't compiled or run. Please wire this test into `LLAMA_STAGE_BUILD_TESTS` and give it a fixture path that works through the normal CTest invocation.

3. The patch queue now has two files numbered `0033`: [this patch](https://github.com/Mesh-LLM/mesh-llm/blob/0877468a9de7db0cf4a3b045a7033a355c1addd4/third_party/llama.cpp/patches/0033-skippy-account-retained-tensors-during-staged-loads.patch) and [the existing runtime-config patch](https://github.com/Mesh-LLM/mesh-llm/blob/582aff67847d94e326ab672faa5a2ae9703e4ee5/third_party/llama.cpp/patches/0033-skippy-add-kv_offload-kv_unified-and-swa_full-runtim.patch). The branch has 35 patch files but numbers only through `0034`. Rename this one to `0035` after rebasing.

4. The synthetic accounting check puts both duplicate tensors in one context and only calls `init_mappings()`. The model-load case sets `include_output = false`, so it doesn't cover a tied physical weight retained across two contexts. Add a CPU-only two-context case that loads the contexts in sequence and checks that progress stays below completion after the first load, then reaches completion and tears down mappings once after the final load. The same traversal test should include a tensor for which `get_weight()` returns null.

I replayed the patch queue against the original pin and current `main`; both replays passed, and `just llama-build` passed on the current pin. The production code compiles. The current CI failure isn't from this change. #1555 already fixes it on `main`, so rebasing should clear that part. The stated CUDA/V100 gate hasn't run yet.

Please rebase, renumber the patch, fix and wire the native test, add the two-context lifecycle assertion, then rerun CI and the CUDA gate.

IvGolovach:
Implemented the requested review changes and refreshed the PR on final SHA `9ef85e76b46d81b5cc1ad31e5edd70b8901141dd`.

Review follow-up:

- updated the native test for the pinned `ctx_key { buft, lazy }` shape;
- wired fixture generation and accounting regression into `LLAMA_STAGE_BUILD_TESTS` and the normal `just skippy-native-tests` / CTest path;
- rebased onto current `main` and renumbered the new queue entry to `0035`;
- added a CPU-only two-context sequential load proving progress remains incomplete after context one, completes and cleans up once after context two, and ignores a synthetic tensor with no GGUF weight;
- added filtered real-model coverage for generated dense Gemma and Qwen2MoE fixtures;
- fixed the planner failure by scoping `.gitattributes` inside the already-owned patch-queue directory instead of changing the protected ownership catalog.

Current local proof includes a clean 35-patch replay, `just skippy-native-tests` (5/5 CTest), `just build`, 904 focused Rust tests, the 31-test CI planner suite, and a clean full branch `git diff --check`. The PR body now contains the exact commands and branch proof.

Mandatory CI is now fully green on the final SHA: 33 GitHub Actions check-runs passed, 12 were expected skips, and none failed or remain pending. All five stable `PR / <lane>` results passed, including the Linux host/runtime/static-ABI chain, four Rust-test batches, composed CPU product, core inference, and both two-node smokes. Remaining gates are maintainer re-review and the requested CUDA/V100 hardware reproduction; I am not claiming merge-readiness while those remain open.

i386 (approved):
Reviewed the revised patch at 9ef85e76. The prior blockers are resolved: the patch is renumbered, the current ctx_key shape is used, the dense/MoE accounting tests are wired into LLAMA_STAGE_BUILD_TESTS, and the two-context lifecycle test covers repeated physical weights plus a missing get_weight entry. I replayed all 35 patches to 7eda86dd and ran CMAKE_BUILD_PARALLEL_LEVEL=4 just skippy-native-tests locally: all 5 CTest cases passed. Repository CI is also green.

PR #1544, opened 2026-08-30
skippy: clear stale sidecar rows before single-row MTP proposals

DaBlitzStein:
## Problem

On a machine with three Tesla P40s serving Qwen3.8-27B (`qwen35`, Q4_K_M, integrated MTP
head), every sufficiently long generation fails with HTTP 503:

```
RuntimeError: llama_decode failed for MTP sidecar proposal
```

The process stays alive and keeps serving; only the affected request dies. The unfiltered
native log gives the real cause:

```
init: the tokens of sequence 0 in the input batch have inconsistent sequence positions:
 - the last position stored in the memory module of the context (i.e. the KV cache) for sequence 0 is X = 352
 - the tokens for sequence 0 in the input batch have a starting position of Y = 352
 for M-RoPE, it is required that the position satisfies: X < Y
decode: failed to initialize batch
llama_decode: failed to decode, ret = -1
```

`X == Y` on every occurrence (observed at 347, 350, 352 and 381, on both sequences).

## Root cause

The native MTP sidecar keeps its own KV cache. `skippy_mtp_note_target_tokens`
(`src/skippy/speculative_decoding.cpp`) prunes the unverified proposal rows only when the
pending draft position falls inside the token window the target just committed:

```cpp
if (session->mtp_pending_draft_pos >= token_start &&
        session->mtp_pending_draft_pos < token_end) {
    ...                                  // accepted -> prune deeper rows
    ...                                  // rejected -> prune from draft pos
}
session->mtp_has_pending_draft = false;  // runs either way
session->mtp_pending_draft_end  = 0;
```

Outside that window — and on the early-return reprime path higher up in the same function —
the bookkeeping is reset but the rows stay resident. The existing comment on the accepted
branch already describes the consequence:

> Left resident they make the next sidecar decode non-consecutive (`seq_pos_max` is past the
> position we resume at) and `llama_batch_allocr::init` rejects the batch.

The next proposal resumes at `session->n_past` with `seq_pos_max` already at or past it.
Plain RoPE tolerates that; M-RoPE requires the batch to start strictly past `seq_pos_max`,
so `llama_decode` returns `-1` and the sidecar surfaces the error above.

This is why it is specific to `qwen35`: the architecture uses M-RoPE positions. It is not
caused by loading an `mmproj` — the failure reproduces identically with only the text GGUF
mounted and no projector.

## The change

One new patch, `0033`, adding four lines before the single-row proposal loop:

```cpp
if (llama_memory_t memory = mtp_ctx->get_memory()) {
    llama_memory_seq_rm(memory, session->seq_id, session->n_past, -1);
}
```

This mirrors what the chained branch (`chain_heads`) already does before each of its decodes.
Everything at or past `session->n_past` is speculative by construction — `n_past` counts
committed target tokens, and every target decode writes at `n_past + i` before advancing it
(`execution-single.cpp:214`/`233`, `execution-batch.cpp:240`/`325`, `activation.cpp:787`/`382`)
— so the clear is safe whether or not the sidecar shares the target's memory.
`activation.cpp:899-905` already uses the same `seq_rm(seq_id, token_start, -1)` shape to roll
back verification.

Fixing it at the point of use rather than at each discard site keeps the invariant that
`llama_batch_allocr::init` actually requires, and covers discard paths added later.

## Evidence

Hardware: 3x Tesla P40 (sm_61), driver 580.173.02, CUDA 12 runtime
(`61;75;80;86;87;89;90`). Model: `Qwen3.8-27B-ABLITERATED-Q4_K_M.gguf`, `general.file_type=15`,
65 blocks, `nextn_predict_layers=1`, MTP head present (`blk.64.nextn.eh_proj.weight`
`[10240, 5120]` Q4_K). Config: `split_mode = "layer"`, `tensor_split = "33,33,34"`,
`ctx_size = 262144`, `parallel = 2`, KV f16, flash attention on.

Same request, same config, same box:

| runtime | result | `for M-RoPE` errors |
|---|---|---|
| stock `v0.75.1` bundle | 3/3 HTTP 503 | 3 |
| `main` + this patch | 4/4 complete generations | 0 |

Both runs append to the same `skippy-native.log`. The failing run's errors are at lines
4259/4265/4271; the patched run's model load starts at line 4275 and produces no error
through end of file (6752).

MTP stays active with the patch: `draft_n=598`, `draft_n_accepted=187`,
`native_mtp_verifications=298`.

Patch queue validation: all 33 patches apply cleanly to the pinned upstream
`cc83d7b4824f73cfdda4dfbb47ee39804f71b328` in a fresh checkout via
`LLAMA_WORKDIR=$tmp scripts/prepare-llama.sh pinned`, with no three-way fallback. No Rust
changed.

## Architecture review

`mtp_shares_target_memory` can only be true for `LLM_ARCH_NEMOTRON_H_MOE`: `ctx_other` is
propagated for INKLING (`llama-context.cpp:162`) and NEMOTRON_H_MOE (`:173`), and
`skippy_mtp_shares_target_memory` explicitly excludes INKLING
(`speculative_decoding.cpp:107`). In the NEMOTRON_H_MOE case `pos` never advances, so the
sidecar always writes at `n_past`; clearing that row before re-decoding at the same position
is required, not merely harmless. INKLING with `depth_count > 1` takes the chained branch,
which already performs the equivalent clear. GLM-DSA and Qwen4Exp do not share memory, so the
sidecar owns the cache it is clearing. When no stale rows exist the call is a no-op.

## What is not covered

No regression test. Reproducing it needs a real `qwen35`-class model on a GPU and a
generation long enough to discard a proposal outside the commit window, which puts it in the
certification lane rather than a unit test —
`crates/skippy-correctness/src/native_mtp_openai` (`native-mtp-openai-ab`) looks like the
right home. Happy to add it there if you want it in this PR.

Separately, and not addressed here: a build from current `main` segfaults during `clip` init
when the vision `mmproj` is loaded for this model, where `v0.75.1` loads it fine. The
verification above therefore ran without the projector. That looks like a distinct
pre-existing regression and I did not want to widen this change; I can file it separately with
the logs.

---

🤖 Investigation, patch and this description were produced with AI assistance (Claude), on
real hardware, and reviewed against the code paths cited above. The delegated independent
review lane did not return, so the architecture review section is a self-review — worth a
second pair of eyes, particularly on the `mtp_next_pos` bookkeeping, which this change
deliberately leaves untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that could cause speculative decoding to fail after discarded draft tokens.
  * Improved decoding reliability for configurations using M-RoPE positional encoding.
  * Prevented stale speculative cache data from blocking subsequent token generation.


i386:
I think we should also add Qwen3.8-27B to our battery of tests in https://github.com/Mesh-LLM/mesh-llm/actions/workflows/llama-upstream-canary.yml

i386:
I am making that change now

i386 (changes requested):
Reviewed the code path and the patch-queue packaging. **The fix itself is correct** — I'm requesting changes only for a patch-number collision that must be resolved before merge.

## The fix is right

`skippy_mtp_propose_next`'s single-row `else` branch sets `pos = session->n_past` and enters the proposal loop with no prior `seq_rm`, while the chained (`chain_heads && !chain_at_same_position`) branch already clears with the identical call inside its loop:

```cpp
if (llama_memory_t memory = mtp_ctx->get_memory()) {
    llama_memory_seq_rm(memory, session->seq_id, session->n_past, -1);
}
```

The new patch inserts exactly that clear ahead of the single-row loop, so both branches now clear the same way. Verified against the real patched tree:

- **Root cause matches the code.** The in-tree comment right above the branches already says *"The chained branch clears its own rows before every decode, so only the single-row loop can leave an overhang behind the accepted prefix"* — independent corroboration of the PR's analysis.
- **Safety claim holds.** `mtp_pending_draft_pos = session->n_past` and `mtp_pending_draft_end = proposal_end_pos` confirm the proposal region begins at `n_past`; committed target tokens live below it, so clearing `[n_past, -1)` never drops committed state.
- **Both memory modes covered.** In the shared-memory case (`mtp_shares_target_memory`, i.e. NEMOTRON_H_MOE) `pos` never advances, so each decode rewrites `n_past` and the pre-loop clear is required, not just harmless. In the M-RoPE / non-shared case `pos` advances and the clear removes the previous discarded proposal's overhang before the new one starts.
- The reported error string `"llama_decode failed for MTP sidecar proposal"` is exactly this loop's failure path.

## Required change (blocking): duplicate patch number `0033`

Since this branch was opened, `main` advanced its queue and now already contains:

- `0033-skippy-add-kv_offload-kv_unified-and-swa_full-runtim.patch`
- `0034-skippy-wire-sampling-and-chat-request-defaults.patch`

This PR adds a **second** `0033-*.patch`. `scripts/prepare-llama.sh` applies the queue via `git am --3way` over ``find … -name '*.patch' | sort``, so today it only works by the lexical accident that `add` < `clear` — two patches sharing `0033` violates the numbering convention that ordering depends on, and it's fragile against any future renumber/regeneration.

Please rename to **`0035-skippy-clear-stale-sidecar-rows-before-single-row-MT.patch`** (next free number after 0034). No content change needed — the hunk only touches `speculative_decoding.cpp`, which neither `0033-kv_offload` nor `0034` modifies, so there's no interaction and the context applies cleanly.

## Please also refresh the validation note

The PR body's evidence (*"all 33 patches apply cleanly"*, pinned upstream `cc83d7b…`, *"no three-way fallback"*) predates main moving to pin `9723942…` with the queue through 0034. The branch's `upstream.txt` still reads `cc83d7b…`; since the PR doesn't touch that file, main's newer pin wins on merge — but please re-run `scripts/prepare-llama.sh pinned` against current main's pin + renumbered queue and update the note. I confirmed the hunk context is structurally intact, so I expect a clean apply.

## Non-blocking

The missing regression test is fine to defer — this needs a real qwen35-class M-RoPE model + a long-enough generation to discard a proposal outside the commit window, which is a certification-lane (`native-mtp-openai-ab`) concern, not a unit test. Worth confirming the canary lane actually exercises an M-RoPE draft path so this doesn't silently regress.

Renumber to 0035 + refresh the validation note and I'm happy to approve.

i386 (approved):
Approving — the one blocking item from my earlier review is resolved.

**Renumber done (I pushed it):** `fbc39a9d` renames the patch `0033 -> 0035-skippy-clear-stale-sidecar-rows-before-single-row-MT.patch`. Pure rename, no content change — the hunk only touches `speculative_decoding.cpp`, which neither `0033-skippy-add-kv_offload...` nor `0034-skippy-wire-sampling...` modifies, so applying it last is conflict-free.

**Re-validated** against the pinned upstream `cc83d7b` (unchanged): `LLAMA_WORKDIR=$tmp scripts/prepare-llama.sh pinned` git-am's all 35 patches cleanly in sorted order, `0035` applied last, **no three-way fallback**, and the `Drop any sidecar row at or past...` fix is present in the patched tree.

The fix itself was already correct in my prior review (single-row proposal path now clears `[n_past, -1)` like the chained branch). No other changes requested.

Minor, non-blocking: the description still says "all 33 patches apply cleanly" — it's 35 now; the pin `cc83d7b` is still accurate. Worth a one-line refresh but not merge-blocking.

Issue #1381, opened 2026-08-19
CI: SDK smoke has no cargo cache — 3m59s of cold compile for a 1.32s test


ndizazzo:
## Summary

`sdk-smoke.yml` compiles the entire Rust dependency graph from a cold registry on every run, and separately rebuilds the whole web UI, inside the smoke step. On a recent run this was **3m59s of cargo build + ~18s of UI build to execute a 1.32s test**.

The root cause is that `.github/workflows/sdk-smoke.yml` is the only compiling workflow in the repo with **no cargo cache of any kind** — no `configure-sccache-gha`, no `Swatinem/rust-cache`, no `actions/cache` on `~/.cargo`. This is declared, not accidental: `ci/slices.yml:240` sets the `sdk` slice to `"cache_mode": "none"`.

This is pre-existing on `main`, not a regression from any open PR.

## Evidence

Run [32282107616](https://github.com/Mesh-LLM/mesh-llm/actions/runs/32282107616), job `96175554873` (`Linux / Linux SDK smoke / Rust SDK smoke / rust SDK Smoke`), step `Rust SDK smoke test`:

| Marker | Value |
| --- | --- |
| `Updating crates.io index` | 18:13:43 — cold registry |
| Crates `Downloaded` | **647** |
| Crates `Compiling` | **613** |
| `Finished \`test\` profile [unoptimized + debuginfo] target(s)` | **3m 59s** |
| `test result: ok. 1 passed; ...` | **1.32s** |
| UI build (`tsc -b && vite build`, 5616 modules) | ~18s |
| Step total | 4m41s |
| Job total | ~5m30s |

## Why the downloaded artifact does not help

Worth stating explicitly, because the job log looks like a cache restore followed by a redundant compile — it isn't.

`./.github/actions/restore-smoke-inputs` (`sdk-smoke.yml:248`) downloads the **composed product**: the `release` `mesh-llm` host binary plus its native runtime, staged at `target/release/mesh-llm`. `scripts/ci-sdk-fixture.sh` launches that as the live server-under-test and `MESHLLM_NATIVE_RUNTIME_ARTIFACT_DIR` points the SDK at the prebuilt runtime. It is a **runtime input**. It populates neither cargo's `target/` nor the registry.

The compile is a different build entirely — `scripts/ci-rust-sdk-smoke.sh:25`:

```bash
cargo test -p mesh-llm-ffi --test live_sdk_smoke -- --nocapture
```

[...]

PR #1556, opened 2026-08-31
ci: cache SDK smoke and reuse UI artifacts

ndizazzo:
## Summary

- add an exact, main-seeded Rust SDK Cargo/target cache that pull requests can restore but cannot publish
- include OS, architecture, target, image, toolchain, profile/linker, and recipe inputs in the cache compatibility boundary
- reuse each lane immutable UI artifact across Rust, Kotlin, and Swift SDK smoke tests and skip redundant UI builds
- carry an isolated Rust 1.98 chunk-API compatibility fix required for current main to pass Clippy locally

## Validation

- just test-all
- just ci-validate
- just ci-shellcheck scripts/ci-rust-sdk-smoke.sh scripts/ci-kotlin-sdk-smoke.sh scripts/ci-swift-sdk-smoke.sh
- cargo test -p skippy-protocol --all-targets
- focused skippy-protocol Clippy validation

## Rollout note

The protected planner requires ci/slices.yml to match main byte-for-byte. This PR keeps the operational cache and UI-artifact changes while leaving the SDK cache_mode declaration at the protected value. A separate catalog-only maintainer merge must set it to pr-isolated; this branch can then rebase and restore that declaration.

Addresses #1381

## Summary by CodeRabbit

* **Bug Fixes**
  * SDK smoke tests now package and validate the matching platform UI assets, improving consistency across Rust, Kotlin, and Swift builds.
  * Binary data handling is more robust and avoids unnecessary failure paths.

* **Performance**
  * Rust SDK checks now reuse compatible build caches to reduce CI build times.

* **Documentation**
  * CI workflows and cache policies now document UI artifact handling and Rust cache behavior more clearly.

* **Tests**
  * Added coverage to verify cache configuration and prevent incompatible or outdated cache usage.

### DIFF
--- a/.agents/skills/manage-ci/references/current-inventory.md
+++ b/.agents/skills/manage-ci/references/current-inventory.md
@@ -131,13 +131,13 @@ removable after this branch's runner contract is active on protected main.
 | `ci-{linux,macos,windows}-product-slice.yml` | Platform-pure composition-only product consumers |
 | `ci-platform-checks-slice.yml` | macOS portable/unit, Windows portable, and Windows log-store privacy ACL checks |
 | `ci-linux-product-smoke-slice.yml`, `ci-macos-product-smoke-slice.yml` | Platform-local CPU, CUDA (`gpu-nvidia` self-hosted), two-node, Metal and model-download consumers; ROCm/Vulkan products remain package-verified pending eligible inference runners |
-| `ci-linux-sdk-slice.yml`, `ci-macos-sdk-slice.yml` | Platform-local Rust/Kotlin/Swift smoke consumers; SDK producers are independent top-level calls |
+| `ci-linux-sdk-slice.yml`, `ci-macos-sdk-slice.yml` | Platform-local Rust/Kotlin/Swift smoke consumers; SDK producers are independent top-level calls and each smoke receives the lane-local immutable UI artifact |
 | `ci-runner-contract-slice.yml` | Provider/cache/plan trust and main runner-image checks |
 | `native-sdk-artifact.yml` | Typed native SDK producer |
 | `swift-sdk-artifact.yml` | Host-only/full XCFramework producer; trusted main remains `macos-15`, while eligible same-repository PRs follow the protected Depot macOS 15 gate |
 | `smoke.yml` | Artifact-based inference/OpenAI/split smoke |
 | `scripted-binary-smoke.yml` | Artifact-based scripted product smoke |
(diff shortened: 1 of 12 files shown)


i386 (approved):
## Review: verified independently

**What I ran at head (8380832):**
- `python3 scripts/tests/test_sccache_evidence.py` — 21/21 pass
- `cargo test -p skippy-protocol --locked` — 42/42 pass, plus throwaway multi-chunk probes (3×4096+17 values, i32 extremes) round-tripping through `read_i32_values`/`write_i32_slice` — clean
- `bash -n` on all three touched smoke scripts; `ci/slices.yml` confirmed byte-identical to main (rollout note holds)

**Wiring checks:**
- `SDK_RUST_IMAGE_DIGEST` matches both the `sdk-smoke.yml` container image and the `linux-cpu` row in `ci/slices.yml` ✓
- `ui_artifact_name` propagation mirrors the proven host-slice pattern (same `download-artifact` → `crates/mesh-llm-ui/dist` + `test -s index.html` verification), and the sdk job transitively depends on `ui_artifact` via `runtime_product` → `hosts` ✓
- `package-sdk-console-assets.sh --skip-build` already exists on main; scripts reject non-`--skip-build` 8th/4th/7th args ✓
- Cache key boundary (target, image digest, toolchain epoch, profile/linker, workflow+script+lockfile hashes) is sound and the new contract test pins it ✓

**Two non-blocking notes:**

1. **None of the new workflow steps have executed in CI yet.** The `pr_*.yml` entry workflows call the lanes pinned `@main`, so the green checks on this PR ran main's lane versions (with head-SHA scripts/tests only). The UI-download, rust-cache, and `--skip-build` paths get their first real exercise on the post-merge main run. `continue-on-error: true` on the cache step keeps a restore/save failure non-fatal, so worst case is a cold build — acceptable.

2. `save-if` reads `github.event.inputs.original_event_name`, which is empty for `push` events — `'' != 'pull_request'` still evaluates true on main pushes so saving works, but `inputs.original_event_name` (the propagated workflow input) is the more precise form if you ever add a caller that isn't plain `pull_request`/`push`.

Minor test-gap suggestion (not blocking): the chunked i32 codec path (VALUES_PER_CHUNK = 4096) has no existing round-trip test above one chunk. My probes confirm it's correct today; a permanent `>4096` round-trip test in `binary/mod.rs` would pin it.

The `cache_mode` follow-up (catalog-only maintainer merge to `pr-isolated`) is clearly documented in the rollout note — thanks for keeping the protected-catalog constraint explicit.

PR #1558, opened 2026-08-31
fix(api): preserve split capability when model size is unknown

mikemikimike:
## Summary

`doctor split` reported `split_capable: false` when a layer package had no catalog `total_bytes`, even though the package type itself proves that it can participate in a split. This conflated unknown model size with a known non-splittable model.

This change tracks split capability independently from size hints, registers layer-package aliases even when their size is absent, and resolves revision-qualified references back to their package repository. Capacity classification remains `unknown_model_size` until bytes are known, while known `true` and `false` values are preserved. An actually unknown capability is omitted from the JSON payload instead of being reported as `false`.

## Validation

- `cargo test -p mesh-llm-host-runtime model_target_capacity::tests --lib` (5 passed)
- `cargo check -p mesh-llm-host-runtime` (passed)
- `cargo clippy -p mesh-llm-host-runtime --lib` (passed with existing warnings)
- `cargo test -p mesh-llm-host-runtime --lib` (2782 passed, 21 unrelated Windows/environment failures, 8 ignored)
- `cargo test --workspace --all-targets` (blocked by missing patched `llama.cpp` quant archives on Windows; automatic native preparation is unsupported by the repository build script)

The full-test failures are in untouched config, resolver, logging, mesh-admission, process-scan, and Windows-path fixtures; the changed capacity tests pass.

## Summary by CodeRabbit

* **Improvements**
  * Split-capability detection now works consistently across model variants and aliases.
  * Models can be recognized using direct repository references or variant names.
  * Capacity evaluations distinguish between confirmed split support and unknown capability states.
  * Models without known package sizes can still be evaluated accurately.

* **Bug Fixes**
  * Status responses omit split-capability information when it is unavailable, avoiding misleading default values.
  * Improved handling of model aliases and repository identifiers when determining split support.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/api/model_target_capacity.rs
+++ b/crates/mesh-llm-host-runtime/src/api/model_target_capacity.rs
@@ -55,6 +55,15 @@ impl ModelTargetSizeLookup {
                     .packages
                     .iter()
                     .any(|package| package.package_type == "layer-package");
+                if split_capable {
+                    lookup.insert_split_capability_aliases(
+                        variant_name,
+                        &variant.curated.name,
+                        &variant.source.repo,
+                        variant.source.revision.as_deref(),
+                        source_file,
+                    );
+                }
                 if let Some(model_bytes) = variant
                     .curated
                     .size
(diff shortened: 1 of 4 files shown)


i386 (approved):
Reviewed the capacity lookup, alias normalization, tri-state serialization, split-readiness consumption, and lifecycle fixture changes. Revision-qualified layer-package refs correctly fall back to the parsed package repo when catalog size metadata is absent; unknown capability is omitted without affecting capacity classification. Local validation at 52d34fb0: cargo fmt --all --check; cargo test -p mesh-llm-host-runtime --lib (2831 passed, 8 ignored). No required changes found.

i386:
Thanks @mikemikimike !!

PR #1559, opened 2026-08-31
Parse Buzz reply routing as a delivery grammar

michaelneale:
## Summary

Make the Buzz terminal-reply rescue derive routing from the delivery grammar, not from any context wrapper or legacy presentation format.

The accepted route is:

- exactly one column-zero `Channel:` field containing one UUID
- exactly one later column-zero `IMPORTANT:` instruction containing the ordered command tokens `buzz messages send`
- exactly one `--reply-to` option with a 64-hex event ID
- exactly one declared shell tool

Duplicate, conflicting, or malformed productions fail closed. `[Context]`, `<context>`, and incidental English are not grammar elements and receive no compatibility handling.

Once that grammar identifies a route, terminal prose is emitted through it. Genuine tool calls, empty output, errors, concrete-model requests, and completed sends remain outside the terminal-prose conversion.

## Why

The durable contract is the command grammar: `Channel:` plus `buzz messages send --reply-to …`. Parsing wrapper names creates adapter whack-a-mole. This change makes the grammar the source of truth instead.

## Validation

- `cargo test -p mesh-mixture-of-agents` (212 unit tests plus all non-ignored integration/doc tests passed)
- `cargo check -p mesh-llm`
- `cargo fmt --all --check`
- `git diff --check`

## Credits

Co-authored-by: Michael Neale <14976+michaelneale@users.noreply.github.com>

### DIFF
--- a/crates/mesh-mixture-of-agents/src/buzz_reply.rs
+++ b/crates/mesh-mixture-of-agents/src/buzz_reply.rs
@@ -19,17 +19,17 @@ impl BuzzReplyRescue {
         }
         let messages = request.get("messages")?.as_array()?;
         let (turn_start, text) = current_turn(messages)?;
-        let channel = parse_channel(&text)?;
-        let reply_to = parse_reply_to(&text)?;
-        if !valid_uuid(channel) || !valid_event_id(reply_to) {
-            return None;
-        }
-        let shell_tool = request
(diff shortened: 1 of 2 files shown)


ndizazzo:
@michaelneale this ready to eval? or are you still jamming on it

PR #1565, opened 2026-08-31
fix(skippy-server): make connection worker reads interruptible by shutdown

### LINKED ISSUE #1538: Windows: `ConnectionWorkerControl::shutdown` cannot unblock a worker stuck in a blocking read — stage-server shutdown can hang
## Environment

- Windows 11 Home 10.0.26200, Rust stable-x86_64-pc-windows-msvc
- `crates/skippy-server/src/binary_transport/binary_messaging.rs`, current `main`

## Context

While getting `skippy-server`'s lib tests to compile on Windows (see #1536), `shutdown_tests::accept_error_still_closes_and_joins_active_connection_worker` failed deterministically with `active worker cleanup must complete within one second: Timeout`. These tests had never run on Windows before, so this wasn't visible until now. It turns out not to be a flaky test: the unblocking mechanism the production code relies on doesn't work on Windows.

## The platform difference, isolated

`ConnectionWorkerControl::shutdown` unblocks connection workers by calling `TcpStream::shutdown(Shutdown::Both)` on a `try_clone()` of each tracked stream, while the worker thread is blocked in `read()` on the original. A 40-line standalone program (no mesh-llm code) shows what that does on Windows:

```text
A) shutdown while another thread is blocked in read(): read() never returns — 3 s timeout
B) shutdown before the read() starts:                  read() returns Ok(0) in ~190 µs
```

<details>
<summary>Standalone repro (click to expand)</summary>

```rust
use std::io::Read;
use std::net::{Shutdown, TcpListener, TcpStream};
use std::sync::mpsc;
use std::thread;
use std::time::{Duration, Instant};

fn scenario(label: &str, shutdown_before_read: bool) {
    let listener = TcpListener::bind("127.0.0.1:0").unwrap();
    let client = TcpStream::connect(listener.local_addr().unwrap()).unwrap();
    let (mut server, _) = listener.accept().unwrap();
    let tracked = server.try_clone().unwrap();

    let (started_tx, started_rx) = mpsc::channel::<()>();
    let (done_tx, done_rx) = mpsc::channel::<std::io::Result<usize>>();
    let gate = shutdown_before_read;
    let worker = thread::spawn(move || {
        if gate {
            let _ = started_rx.recv();
        } else {
            drop(started_rx);
        }
        let mut byte = [0u8; 1];
        let _ = done_tx.send(server.read(&mut byte));
    });

    if shutdown_before_read {
        tracked.shutdown(Shutdown::Both).unwrap();
        started_tx.send(()).unwrap();
    } else {
        drop(started_tx);
        thread::sleep(Duration::from_millis(200));
        tracked.shutdown(Shutdown::Both).unwrap();
    }

    let t0 = Instant::now();
    match done_rx.recv_timeout(Duration::from_secs(3)) {
        Ok(result) => println!("{label}: read() returned in {:?} -> {:?}", t0.elapsed(), result),
        Err(_) => println!("{label}: TIMEOUT — read() still blocked after shutdown"),
    }
    drop(client);
    let _ = worker.join();
}

fn main() {
    scenario("A) shutdown during read", false);
    scenario("B) shutdown before read", true);
}
```

</details>

On Unix, scenario A unblocks immediately, which is the behaviour the design assumes. On Windows, WinSock's `shutdown` on a duplicated socket doesn't cancel an in-flight blocking `recv` on another handle.

## Impact

- **Production:** on a native Windows node, binary-stage server shutdown can hang until every connected worker's peer happens to send data or disconnect — the `worker.task.join()` loop in `ConnectionWorkers::shutdown` waits forever on any worker that was mid-`read()`.
- **Tests:** explains the deterministic Windows failure of `accept_error_still_closes_and_joins_active_connection_worker` (and why its sibling test only passes by racing: if the worker hasn't entered `read()` yet, scenario B applies and the shutdown appears to work).

## Possible directions

- Give worker streams a read timeout (`set_read_timeout`) and have the read loop check `shutting_down` between attempts — simple and cross-platform, at the cost of bounded shutdown latency.
- Or drop to platform APIs on Windows (`CancelIoEx`/`closesocket` semantics) to actively cancel the blocked read.

Happy to turn either direction into a PR if you have a preference; the repro machine is available for validation either way.

Virgile-pct:
## Original problem

#1538: `ConnectionWorkerControl::shutdown` unblocks connection workers by calling `TcpStream::shutdown(Both)` on a tracked clone while the worker blocks in `read()`. On Windows that does not interrupt an in-flight blocking read, so binary-stage server shutdown can hang until every connected peer happens to send data or disconnect. It is also why `shutdown_tests::accept_error_still_closes_and_joins_active_connection_worker` failed deterministically on Windows once the lib tests could run there (flagged in the #1536 review as a follow-up worth tracking).

## Diagnostics

Isolated in the issue with a 40-line standalone program: on Windows, `shutdown(Both)` on a clone does not unblock a read that is already in flight (3 s timeout), while a read that starts after the shutdown returns `Ok(0)` immediately. So the only losing state is a worker parked inside a blocking read when shutdown fires, which is exactly where idle connection workers spend their lives.

## Fix

Connection workers no longer park inside an uninterruptible read. `ConnectionWorkerControl` gains `wait_for_readable`: it peeks under a short read timeout (`WORKER_SHUTDOWN_POLL`, 100 ms, matching `DOWNSTREAM_SHUTDOWN_POLL`) and rechecks the shutdown flag between attempts. The gate runs before the first-message read and before every subsequent `receive_next_message`; when shutdown is requested it returns cleanly, and the connection path already releases tracked sessions on that exit.

Design notes:

- `peek` consumes nothing, so message framing is never at risk, and the read timeout is cleared before the real `read_stage_message` call. The accepted-connection-remains-blocking invariant from `prepare_binary_stage_connection` is preserved.
- The pattern mirrors two mechanisms already in the tree: the downstream ready handshake polling in `stage_execution.rs`, and the peek-based optional client hello.
- `ConnectionWorkerControl::shutdown` itself is unchanged: socket shutdown still covers the non-Windows platforms immediately and remains correct everywhere else.
- Residual limitation, stated for honesty: a peer that stalls mid-frame during shutdown can still delay a worker inside `read_stage_message` on Windows. The gate covers the idle wait between messages, which is the state the failing test exercises and where workers actually park. Frame-level read deadlines would be the follow-up if that ever bites in practice.

The two shutdown tests now model the worker contract (wait through the gate, then read) instead of a raw blocked `read()`, and pass on Windows where one of them previously failed deterministically.

## Validation

- [x] `cargo test -p skippy-server --features dynamic-native-runtime --lib -- binary_transport` on Windows 11 (MSVC): 75 tests green, 0 failed, including the two shutdown tests (one previously failing deterministically there) and the `reap_finished` tests.
- [x] `cargo fmt` applied and the console-print ratchet regenerated (line shifts only).
- [x] No protocol or ABI surface is touched.

Fixes #1538.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection shutdown handling, especially on Windows.
  * Idle connections now respond promptly to shutdown requests instead of remaining blocked while waiting for incoming data.
  * Enhanced connection readiness checks for more reliable binary communication and testing.

### DIFF
--- a/crates/skippy-server/src/binary_transport/binary_messaging.rs
+++ b/crates/skippy-server/src/binary_transport/binary_messaging.rs
@@ -2,7 +2,7 @@ use std::{
     collections::BTreeMap,
     future::Future,
     io::{self, Write},
-    net::TcpListener,
+    net::{TcpListener, TcpStream},
     sync::{
         Arc, Mutex,
         atomic::{AtomicBool, Ordering},
@@ -45,6 +45,10 @@ mod telemetry;
 
 use self::connection::handle_binary_connection;
 
+/// How often a waiting connection worker rechecks the shutdown flag, matching
+/// the downstream DOWNSTREAM_SHUTDOWN_POLL cadence in stage_execution.
+const WORKER_SHUTDOWN_POLL: Duration = Duration::from_millis(100);
+
 #[derive(Default)]
 struct ConnectionWorkerControl {
     shutting_down: AtomicBool,
@@ -81,6 +85,41 @@ impl ConnectionWorkerControl {
             .lock()
             .expect("connection sockets lock poisoned")
             .clear();
+    }
+
+    fn is_shutting_down(&self) -> bool {
+        self.shutting_down.load(Ordering::Acquire)
+    }
+
+    /// Block until `stream` has readable data (or EOF), returning false when
+    /// shutdown is requested instead. Peeks under a short read timeout so no
+    /// message bytes are consumed and the worker never sits in an
+    /// uninterruptible blocking read: `TcpStream::shutdown` on a tracked
+    /// clone does not unblock an in-flight `read` on Windows (#1538).
+    fn wait_for_readable(&self, stream: &TcpStream) -> io::Result<bool> {
+        stream.set_read_timeout(Some(WORKER_SHUTDOWN_POLL))?;
+        let mut probe = [0u8; 1];
+        let ready = loop {
+            if self.is_shutting_down() {
+                break false;
+            }
+            match stream.peek(&mut probe) {
+                Ok(_) => break true,
+                Err(error)
+                    if matches!(
+                        error.kind(),
+                        io::ErrorKind::Interrupted
+                            | io::ErrorKind::WouldBlock
+                            | io::ErrorKind::TimedOut
+                    ) => {}
+                Err(error) => {
+                    let _ = stream.set_read_timeout(None);
+                    return Err(error);
+                }
+            }
+        };
+        stream.set_read_timeout(None)?;
+        Ok(ready)
     }
 }
 
@@ -374,6 +413,12 @@ fn run_binary_stage(options: BinaryStageOptions, shutdown: Arc<AtomicBool>) -> R
                         "binary sent ready: stage_id={} peer={peer_addr:?}",
                         config.stage_id
                     );
+                    if !task_control
+                        .wait_for_readable(&upstream)
+                        .context("wait for the first binary stage message")?
+                    {
+                        return Ok(());
+                    }
                     let first_message = match read_stage_message(&mut upstream, activation_width) {
                         Ok(message) => message,
                         Err(error) if error.kind() == io::ErrorKind::UnexpectedEof => return Ok(()),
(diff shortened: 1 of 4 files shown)


i386 (approved):
Reviewed the Windows shutdown polling path at 74737730. `peek` preserves framing, the timeout is cleared before the real frame read, EOF and shutdown both terminate cleanly, and the control is threaded through first and subsequent idle reads. The documented mid-frame limitation remains bounded to follow-up scope. No blocking issues found.

michaelneale (approved):

PR #1568, opened 2026-08-31
Allow split-serving nodes to use matching local GGUF files

### LINKED ISSUE #1169: [Question / Feature Request] Clarification on pre-cached local GGUF SHA256 resolution & zero-network layer splitting across nodes
#### **Environment**
- **MeshLLM Version**: `0.74.0` (Linux x86_64 / CUDA 12)
- **Topology**: Heterogeneous multi-node mesh (`mDNS` LAN mode)
- **Nodes**: Primary host (with model downloaded) + Worker nodes (with same model downloaded)

---

#### **Context & Goal**
We are attempting to serve large GGUF models (such as `DeepSeek-V4-Flash` ~120GB–155GB and `Qwen2.5-72B` ~42GB) across a multi-node LAN mesh. 

To eliminate network bandwidth bottlenecks during multi-gigabyte layer partitioning, **each node pre-stores the full `.gguf` model file locally on its high-speed NVMe SSD**.

---

#### **Questions for the Maintainers**

1. **Local Hash Reconciliation**: 
   How does `mesh-llm`'s embedded `blobstore` plugin reconcile local file references (`local-gguf/sha256-...`) across worker nodes when file paths differ across machines (e.g. `/home/<user_name_here>/models/` vs `/mnt/d/models/`) or other arbitrary file paths created by a user?

2. **Forced Local-Fit Policy**: 
   Is there an explicit configuration flag or setting in `~/.mesh-llm/config.toml` (or CLI) to strictly enforce that worker nodes instantiate assigned layer splits **exclusively from their local disk** and reject network blob streaming fallbacks?

3. **Pre-Indexing Best Practices**: 
   What is the recommended command or config syntax to ensure worker nodes pre-index local GGUF SHA256 hashes into their local inventory prior to receiving a split execution plan from the primary publisher node?

---

#### **Expected Behavior**
When the primary node initiates a split execution plan for a model, worker nodes with matching SHA256 model hashes should instantiate their assigned tensor layers directly via local disk `mmap()` / `fread()` without initiating any network blob transfers to improve performance & reduce network traffic.

IvGolovach:
## Original problem

Split serving for a pre-copied GGUF was effectively bound to one machine's filesystem path. Identical GGUF bytes stored under different absolute paths could not be reconciled as the same source, and operators had no strict policy guaranteeing that workers would use only their local copy instead of falling back to catalog, Hugging Face, or peer artifact transfer.

Closes #1169.

## User-visible result

- Split-serving nodes can keep byte-identical GGUF files at different absolute paths.
- Setting source_policy = "local-required" makes local content mandatory: a missing, incomplete, mutated, or mismatched source fails closed instead of fetching replacement bytes.
- Plain GGUF files use SHA-256 content identity; sharded sets use an order-sensitive aggregate over every shard.
- Existing fallback behavior remains the default.

Example configuration on each node, with a node-local model_path:

    [[models]]
    model = "orca-70b-q4-k-m"

    [models.hardware]
    model_path = "/node-specific/path/orca-70b-q4-k-m.gguf"

    [models.skippy]
    source_policy = "local-required"

## Diagnostics

I traced the strict-source decision across configuration inheritance, startup planning, gossip, direct-peer capability discovery, stage inventory, coordinator admission, the LoadLocal control message, worker-side materialization, and operator-visible status.

That audit exposed three separate trust decisions that all needed enforcement:

1. whether a model really resolves to strict local mode after defaults are applied;
2. whether every selected direct participant supports content-addressed local GGUF loading;
3. whether the worker still owns the exact advertised bytes immediately before native load.

The implementation and tests cover each boundary independently.

## Fix

- Derive path-independent local-gguf://sha256 identities from verified source bytes. For split files, hash every shard and include ordinal, size, and digest in the aggregate identity.
- Add strict local-source configuration with correct inheritance from [defaults.skippy] and model-level override behavior.
- Advertise additive local-GGUF content-ID capability and inventory data without exposing node-local paths.
- Admit only direct peers that support the strict capability; relayed capability claims cannot promote an ineligible participant.
- Revalidate peer capability and content inventory immediately before issuing LoadLocal.
- Require the worker to resolve and attest the same content identity locally; strict inventory and load never consult catalog, Hugging Face, persistent legacy digest sidecars, or peer artifact transfer.
- Invalidate in-process identity reuse when size, modification time, inode change time, device, or inode changes.
- Keep strict status and wire payloads path-free.
- Preserve the legacy path-derived/fallback lane unchanged for non-strict models.
- Document single-file, sharded, per-model, defaults-based, and CLI-assisted setup.

## Architecture

The change keeps responsibilities explicit:

- configuration owns source-policy parsing, inheritance, validation, and editor behavior;
- local-source modules own source snapshots and content identity;
- mesh/protocol code owns additive capability and inventory transport;
- coordinator admission owns direct-peer eligibility and pre-load revalidation;
- workers own final local resolution and attestation before materialization.

No external llama-server lane is introduced; loading remains in the embedded staged runtime.

## Protocol and compatibility

The protobuf additions are optional and additive. Older nodes ignore fields they do not understand. A node that does not advertise local-GGUF content-ID support is excluded only from a local-required split topology; it may remain in the mesh and serve compatible work.

The default fallback policy preserves existing mixed-version resolution and transfer behavior. No existing field is removed or repurposed, and this is not a breaking mesh protocol change.

## Diff scope

- 73 files changed: 4,293 additions and 258 deletions.
- Runtime/config/protocol changes are concentrated in mesh-llm-config, skippy-protocol, and mesh-llm-host-runtime.
- Tests cover configuration inheritance, protocol round trips, direct-vs-relayed capability, content inventory, mutation handling, status redaction, startup admission, and load-time failure paths.
- Documentation updates cover operator setup and the config reference.
- No dependency, CI workflow, database migration, release version, auth, payment, or UI behavior change is included.

## Branch and commit integrity

- Base: main at e8c37783af94b0e3e7204377f1cb414ab1b01bed.
- Branch: 0 commits behind / 1 commit ahead of the fetched base.
- Merge base: e8c37783af94b0e3e7204377f1cb414ab1b01bed.
- Introduced commit: 3eb3bc83e14f6632e43a918af0a323bd80f29e38 — Enable content-addressed local GGUF split serving.
- One logical change; the final diff contains only intended source, tests, fixtures, and documentation.

Ledger: not applicable — not required for this change family.
Version: not applicable — release versioning remains owned by the release workflow.

## Validation

Validation mode: Mode 3 — shared distributed-runtime and additive protocol behavior.

Local proof on the final commit:

- git diff --check origin/main...HEAD: PASS, no output
- cargo fmt --all -- --check: PASS
- cargo test -p mesh-llm-config --lib: PASS, 188 passed
- cargo test -p skippy-protocol --lib: PASS, 42 passed
- cargo test -p mesh-llm-host-runtime --lib -- --test-threads=1: PASS, 2,890 passed, 8 ignored
- cargo check -p mesh-llm: PASS
- just build: PASS, production UI + shipped debug host + patched Metal native runtime composed
- just release-build: PASS, production UI + optimized host + packaged Metal native runtime
- CodeRabbit review of all 73 files on 3eb3bc83: PASS, 0 findings

Not run locally — the two-node mixed-version/CUDA hardware scenario requires external machines. Before merge, the required manual/remote gate is to verify different node-local paths, direct-peer capability admission, union model visibility, successful split inference, and no model-byte transfer.

Full mandatory GitHub CI on the final SHA: pending until this PR is created. This PR is ready for review, but is not merge-ready until those gates pass.

## CI context confirmation

No workflow files or CI context names changed.

## Runtime safety

- No new blocking lock or unbounded queue is introduced.
- Strict admission, inventory, and load checks fail closed.
- Direct-peer trust is not inferred from relayed gossip.
- Existing fallback and mixed-version behavior remain available.

## Documentation integrity

docs/USAGE.md and the website config-default/reference pages describe the exact operator contract, including split GGUF sets, path restrictions, hashing behavior, compatibility, and failure modes.

UI screenshots are not applicable because no UI behavior or presentation changed.

## Rollback plan

Rollback: revert this PR.
DB downgrade: not applicable.
Data repair: not applicable.
Operational caveats: nodes configured with local-required will need to return to fallback or remove the policy after rollback.

## Known residual risks

- The first strict verification intentionally reads every source byte; very large models pay that one-time integrity cost.
- A process-local cache is reused only while the full file metadata snapshot remains unchanged.
- The native runtime opens the operator-owned path after final verification. A privileged local process could still replace the file in that final interval; this trust boundary is documented.
- Mixed-version and multi-node GPU proof remains a required manual/remote gate before merge.

## Contributor checklist

- [x] I ran the relevant local checks and recorded the exact results above.
- [x] Visual evidence is not needed because this PR does not change UI behavior.

## Summary by CodeRabbit

* **New Features**
  * Added `fallback` and `local-required` source policies for Skippy models.
  * Added verified local GGUF loading with content and SHA-256 checks.
  * Added dedicated local-load support for split serving.
  * Added capability reporting for verified local GGUF support.
* **Bug Fixes**
  * Model profile matching now correctly applies inherited defaults.
* **Documentation**
  * Updated usage guides and configuration references with source policy settings.


i386:
Welcome back @IvGolovach :)

crates/mesh-llm-host-runtime/src/inference/skippy/local_source.rs
```diff
@@ -0,0 +1,427 @@
...
+    model_id: &str,
+    runtime_profile: &str,
+    local_source_required: bool,
+) {
+    if model_id.is_empty() {
+        return;
+    }
+    if let Ok(mut models) = local_source_policy_registry().lock() {
+        let profiles = models.entry(model_id.to_string()).or_default();
+        profiles.insert(runtime_profile.to_string(), local_source_required);
```
i386, line 69: Blocking: this registry never removes obsolete profiles. Because `source_policy` is part of `derived_profile`, changing a model from `local-required` to `fallback` (or unloading the strict variant) registers a new `false` key but leaves the old strict key at `true`. A legacy/profile-unaware request then takes `profiles.values().any(...)`, is upgraded to strict by `apply_verified_local_source`, and rejects an ordinary fallback `LoadStage` until the process restarts. Please make registration replace the active per-model snapshot, or deregister profiles when lifecycle/config state is removed.

crates/mesh-llm-host-runtime/src/runtime/local_split/loading.rs
```diff
@@ -480,19 +560,24 @@ pub(super) fn split_runtime_stage_load_request(
...
         stage_id: stage.stage_id.clone(),
         stage_index: stage.stage_index,
         layer_start: stage.layer_start,
         layer_end: stage.layer_end,
-        model_path: Some(stage_load_model_path(
-            settings.load_mode.clone(),
-            &spec.package.package_ref,
-            spec.model_path,
-        )),
+        model_path: (!spec.local_source_required).then(|| {
```
i386, line 571: Blocking for the stated path-free contract: `model_path` is redacted here, but the same strict request still copies `resolved_config.projector_path` into `LoadLocalStage` below. That field is serialized by `stage_load_to_proto`, echoed through stage status, and exposed by the management API, so a strict multimodal topology leaks the coordinator-local absolute mmproj path to every worker/status consumer. Please either omit it from non-stage-0 strict loads and redact it from strict status, or give projectors the same worker-local content-addressed resolution contract.

i386 (changes requested):
Requesting changes for the two inline blockers: obsolete strict-profile policy survives unload or policy changes and poisons profile-unaware fallback loads; strict multimodal control/status still transmits the coordinator-local projector path despite the path-free contract. I reviewed head 79bf9e2d after its rebase and CI-fix rewrite.

i386 (changes requested):
Re-reviewed new head 1e826651. The delta from 79bf9e2d only adds wiring-manifest/docs entries, so both blocking inline findings remain unchanged: stale strict-profile registry state can still poison fallback loads, and strict multimodal status still exposes the coordinator-local projector path. The new docs row also names content_attestation.rs as the implementation path; the existing CodeRabbit thread correctly notes that the implementation is attestation.rs, so I am not duplicating it inline. Current CI is still running.

i386 (changes requested):
Regression follow-up for the two current blockers: please add a lifecycle test that registers a strict profile, transitions or unloads it, then proves a profile-unaware fallback load is no longer forced strict; and a strict multimodal load/status test asserting that neither model_path nor projector_path exposes coordinator-local paths. The existing feature tests do not cover either failure mode.

i386 (changes requested):
The two original correctness blockers are fixed at 17a8ee48beb73e8b31235fc7664ea5bcc8158196, and their new regressions pass. One test-suite blocker remains: the full default-parallel mesh-llm-host-runtime library suite fails 2 tests: content_addressed_inventory_proves_local_bytes_without_leaking_path and local_required_load_reverifies_content_after_inventory; 2894 passed, 2 failed, 8 ignored. Both fixtures produce identical GGUF content identities, so their process-global multi-path registry entries collide under parallel execution. One test can resolve the other valid identical path, defeating assertions that require its own path or expect tampering its own path to exhaust the identity. CI test partitioning currently masks this. Please isolate or uniquify these fixtures, or make the assertions valid for multiple equivalent registered paths, so the package suite is parallel-safe.

crates/mesh-llm-host-runtime/src/runtime/startup_models.rs
```diff
@@ -830,13 +867,19 @@ pub(super) fn build_startup_model_specs(
...
+                local_source_required,
                 profile: effective.derived_profile(),
             });
         }
         for model in &options.model {
             let matching_config = matching_config_model(config, model)?;
             let effective =
                 effective_startup_model_config(&model.display().to_string(), None, defaults);
+            let local_source_required =
+                skippy_local_source_required(effective.skippy.as_ref(), None);
```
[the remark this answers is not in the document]
i386: Verified fixed at e9e08030405c033d69c994e3967c36a4f10cc6de: exact CLI model resolution derives local_source_required from matching_config with the default policy as fallback before remote resolution. exact_cli_model_applies_matched_local_required_policy_before_resolution covers this branch. The full mesh-llm-host-runtime library suite passes: 2,896 passed, 8 ignored.

crates/mesh-llm-host-runtime/src/inference/skippy/local_source.rs
i386: Verified fixed at e9e08030405c033d69c994e3967c36a4f10cc6de: unload and failed-load cleanup now unregister a source policy once the last matching lifecycle entry is gone. last_strict_lifecycle_removal_allows_profile_unaware_fallback_reload covers the strict-to-fallback lifecycle. The full mesh-llm-host-runtime library suite passes: 2,896 passed, 8 ignored.

crates/mesh-llm-host-runtime/src/runtime/local_split/loading.rs
i386: Verified fixed at e9e08030405c033d69c994e3967c36a4f10cc6de: strict stage construction retains projector_path only for stage 0, while strict status and proto boundaries redact projector paths. Coverage includes strict_multimodal_stage_builder_keeps_local_paths_off_downstream_loads and content_addressed_status_redacts_projector_at_wire_boundary. The full mesh-llm-host-runtime library suite passes: 2,896 passed, 8 ignored.

i386 (approved):
Re-reviewed at e9e08030405c033d69c994e3967c36a4f10cc6de. The parallel-test blocker is resolved by giving the load-time reverification fixture a distinct valid GGUF content identity, eliminating cross-test registry aliasing without changing production lookup behavior. Full default-parallel validation: cargo test -p mesh-llm-host-runtime --lib passed with 2,896 passed, 0 failed, and 8 ignored. Focused stage tests passed 28/28; cargo fmt and cargo check passed; clippy is clean with the repository baseline unfulfilled-lint-expectations exception.

hola-there:
Thanks for the work on this!

PR #1561, opened 2026-08-31
fix(skippy): address PR 1478 review blockers

i386:
## Summary

Stacked on #1478 and addresses the concrete blockers in review `5061315419` plus the related unresolved review threads.

- bump the stage wire generation for the expanded fixed-width reply statistics
- report physical KV occupancy for plain, DSA, ISWA, DSA+ISWA, MSA, and DSV4 memory families
- close decode-starvation and generation-admission overtaking paths
- keep native prefill scratch session-local and serialize cached chat-template rendering
- derive the static native directory from `build-llama.sh --print-build-dir`
- make the persistent GPU benchmark schedule-only and capacity-match comparison engines
- fail parity/reporting closed on empty, short, errored, incomplete, stale, or identity-mismatched cells
- split direct-iteration batching helpers out of the scheduler worker and retain regression coverage

## Review disposition

All correctness, security, fairness, reproducibility, and maintainability findings from the linked review are addressed in this stack. The suggested one-token native emission/incremental parsing redesign is deliberately deferred: it is a broader performance ABI change, not required to close the correctness blockers, and needs independent profiling plus parity coverage before implementation.

## Validation

- `just ci-validate` — 712 Python/contract tests passed, 7 skipped; repository consistency gates passed
- `just with-lld cargo test -p skippy-protocol` — 43 passed
- `just with-lld cargo test -p skippy-scheduler` against rebuilt static native runtime — 46 passed
- `just with-lld cargo test -p skippy-runtime` against rebuilt static native runtime — 110 passed
- `just with-lld cargo test -p skippy-server` against rebuilt static native runtime — 607 passed, 3 ignored
- `just with-lld cargo test -p mesh-llm-host-runtime` — 2,833 passed, 8 ignored
- `just with-lld cargo bench -p skippy-scheduler --features scheduler-lab --bench scheduler_lab`
- `actionlint -config-file .github/actionlint.yaml`
- `just with-lld cargo fmt --all -- --check`
- llama.cpp patch queue replayed from pinned `cc83d7b48` to deterministic patched head `dca0b2e68`; static CPU native build passed

## Benchmark caveat

The deterministic scheduler lab is green. The serving changes still require the repository-mandated exact-head Thoughtworks c64/c128/c256 hardware regression battery before promotion; this PR does not claim that hardware replay has run.

Stacked base: `scam/skippy-signal-hotpath` at `9b98ba3de9d8bce4fed531ea2c29325ec5b2c2e3`.

PR #1569, opened 2026-08-31
fix(skippy): make KV replay and restore fail closed

i386:
## Summary

- prevent first-token and exact-output cache reuse across incompatible sampling semantics
- disable stochastic sampled-output reuse and key deterministic replay with an exhaustive versioned sampling fingerprint
- classify model KV capability as dense, recurrent, or unknown and fail closed on corrupt, unreadable, or conflicting metadata
- wrap exact-state and resident-prefix restoration in rollback-and-reacquire cleanup before cache-off fallback
- quarantine only entries whose deterministic cached-payload validation fails; transient native errors never evict otherwise valid entries

## Correctness model

Cached sampled output is now reusable only for deterministic sampling, and every current `SamplingConfig` field plus chat sampling metadata participates in the cache key. The exhaustive field destructure intentionally makes future sampler additions fail compilation until their key semantics are defined.

Cache restore is treated as a transaction at the native session boundary. Any error drops or resets the affected lane, reacquires it, and verifies both native and tracked position are zero. If clean rollback cannot be proven, the session is discarded and the caller receives a hard error instead of continuing from partially imported KV.

Model capability inspection no longer converts GGUF inspection failures to a dense-model answer. Unknown or contradictory evidence disables KV caching for that stage and emits a structured warning.

## Validation

Exact head: `cbc55180bc2d68ced8934e5cb5ccb0c30eb570f0`

- `cargo fmt --all --check`
- `cargo test -p skippy-cache --lib`: 51 passed
- `cargo test -p skippy-server`: 625 passed, 3 ignored
- `cargo test -p skippy-runtime --lib`: 110 passed
- `cargo clippy -p skippy-cache --all-targets -- -D warnings`
- `cargo run -p xtask -- repo-consistency no-console-print --regen` followed by the clean consistency check
- pinned patched llama native head `dca0b2e68f749d5d1a66cac6d8bb88ea5b5cbd6f`
- cache-family full-GGUF gate: 3/3 locally available families passed with no correctness mismatch
- cache-family use-case gate: 24/24 available family/prompt rows passed with no correctness mismatch
- scoped c1 serving sanity: 27 rows, zero request failures

Release binaries:

- `skippy-server`: `44f072ede11bb8f30e30909f91b952a24e5acdec919913f42dc36393de52c043`
- `skippy-correctness`: `8014342e4efa3dc5450e8ad41ec46f28b18d28f8777f23c6fb765bf41ab2b712`

Local cache-family results:

| Family | Payload | Raw warm median | Skippy hit median | Speedup |
| --- | --- | ---: | ---: | ---: |
| Qwen3 dense | resident KV | 10.5 ms | 6.7 ms | 1.57x |
| Falcon-H1 | KV + recurrent | 34.2 ms | 13.0 ms | 2.63x |
| Qwen3Next | KV + recurrent | 134.9 ms | 49.1 ms | 2.75x |

Artifact digests:

- full-GGUF JSON: `3a2ba38d420be3114b1229dfd5e759acdb07e13c129449bd16fae6cc998fca00`
- use-case JSON: `42482cc2f83691d24a133f7c635897a4c85770b889c4641d723572e5709b630c`
- combined report: `c020d4c51e0bc223693a54eb2821facce75ab783e5e3bb8f9414687ddb656e63`

The first sandboxed server run had 22 loopback-bind failures (`EPERM`); the full suite passed outside the socket-restricted sandbox.

### CUDA Thoughtworks replay (pre-review head)

The capacity-matched Llama 3.2 1B dense replay on `be33ed0d` completed on the RTX 5080 with
12/12 cells, 1,792/1,792 measured requests, and zero failures:

| concurrency | raw llama.cpp | Mesh | vLLM | SGLang |
| ---: | ---: | ---: | ---: | ---: |
| 64 | 17.0 | **467.1** | 281.4 | 254.1 |
| 128 | 16.4 | **719.3** | 427.5 | 426.9 |
| 256 | 15.6 | **819.0** | 576.6 | 636.4 |

Mesh leads the strongest optional backend by 66.0%, 68.3%, and 28.7% at
c64/c128/c256. Mean Mesh TTFT is 746/871/1,376 ms, versus
1,476/1,780/2,351 ms for vLLM and 1,671/1,888/2,272 ms for SGLang.

CUDA provenance:

- source: `be33ed0d353128f70a583ea72d45c719f8f13839`
- `skippy-server`: `ec111029bcf908e5b42699b615a958eea5fd205953a7f4281565a2e2b75b737f`
- patched native runtime: `dca0b2e68f749d5d1a66cac6d8bb88ea5b5cbd6f`
- native runtime directory: `aea3d3c25f7e0f2e247f836907ef633aea02e7a65c1dbb1a8772025761a17907`
- Thoughtworks manifest: `d0b0f913f3a3095453820eb0e3e1f58b102e25910ee1bc0da78d62ab0041ee15`
- report: `a0e43061f1d82cf8b9302308e5addbbf473318edfeb689bcc5e796a747eea8d2`
- CSV: `571e859cc77c865c6a4b7a10892da217ea23910986ccd9df075553db6b401b2d`
- artifact manifest: `2e8fec2b178c6a5f8742a5454547f267499cf5b189c2fcf0e2c9f3dccb93896a`
- remote artifact: `/home/jdumay/tmp/skippy-pr1569-be33ed0d/artifact-thoughtworks-c64-c128-c256`

Every entry in `artifact-sha256.txt` passes `sha256sum -c`.

The review follow-ups reject temperature-zero replay when Mirostat is enabled, the sampler chain is noncanonical, XTC probability is positive, or dynamic-temperature range is positive. Production-path record-and-lookup regressions cover all four RNG-backed cases. Current-head local tests and CI cover that change; the CUDA artifact below predates this review-only replay guard.

## Remaining base caveat

- shipped-binary Clippy currently inherits 28 `unfulfilled_lint_expectations` warnings from the stacked #1478/#1561 base; none of those files are changed by this PR

Stacked on #1561, which is stacked on #1478.


ndizazzo (commented):
The rollback and quarantine work is headed in the right direction, but first-token replay still admits RNG-backed sampler configurations.

`sampling_replay_safe` in `crates/skippy-server/src/frontend/sampling_cache_key.rs:10-11` treats every enabled request with `temperature <= 0` as deterministic. The native chain still uses Mirostat when `mirostat_mode` is 1 or 2, and a custom sampler chain can omit the temperature sampler before the final distribution step. Both cases can sample through RNG while passing this predicate. The record and lookup paths in `crates/skippy-server/src/frontend/prefix_cache.rs` then cache and replay a first token the current request did not sample.

Please match replay eligibility to the native sampler chain. A safe initial rule is disabled sampling, or `temperature <= 0` only when Mirostat is disabled and the default sampler chain is used. Add coverage for `temperature = 0` with Mirostat and for `temperature = 0` with a custom chain lacking the temperature sampler. Both should prove that no first token is recorded or looked up.

Validation: `skippy-cache` passed 51/51 tests, and `skippy-server` passed 615 tests with 3 ignored. `git diff --check` passed. `just test-all` stopped on four Rust 1.98 `chunks_exact_to_as_chunks` Clippy findings in unchanged code outside this PR; pinned PR CI is green. I did not have an exact-head live endpoint or native failure-injection fixture for the KV tool-loop path.

i386:
Addressed the review in `ad3de29e8`.

- first-token replay now permits enabled sampling only for `temperature <= 0`, Mirostat disabled, and the exact default sampler chain
- the predicate compares against an allocation-free canonical chain, with a regression guard that the runtime default remains identical
- production record and lookup helpers now share the same fail-closed gate
- regressions prove temperature-zero Mirostat and a custom chain without temperature neither record nor look up a first token, even when an unsafe-key entry already exists

Validation: full `skippy-server` suite 617 passed / 3 ignored, warning-denying all-target Clippy passed, formatting and diff checks passed.

crates/skippy-server/src/frontend/sampling_cache_key.rs
```diff
@@ -0,0 +1,341 @@
...
+/// A positive temperature always leaves the output dependent on RNG state,
+/// including when the caller supplied a seed. Mirostat also samples through
+/// its own RNG-backed selection step, and a custom sampler chain may omit the
+/// temperature sampler that makes a non-positive temperature greedy.
+pub(super) fn sampling_replay_safe(sampling: &SamplingConfig) -> bool {
+    if !sampling.enabled {
+        return true;
+    }
+
+    sampling.temperature <= 0.0
```
michaelneale, line 28: **Blocking:** `temperature <= 0`, Mirostat-off, and the canonical chain are not sufficient for deterministic replay. The canonical chain includes `xtc` (line 12), whose pinned-upstream implementation owns an RNG and draws whenever `probability > 0`, `threshold <= 0.5`, and there are at least two candidates. Also, `dynatemp_range > 0` makes `temp_ext` derive a positive entropy-dependent temperature even when base temperature is 0; the terminal `dist` sampler then performs a real draw. Both controls are user-reachable (`frontend/request.rs:698-719`). Please reject those configurations here (or encode the exact native no-op conditions), and extend both this unit test and the production record/lookup regression with XTC and dynatemp cases.

michaelneale (changes requested):
Requesting changes for one remaining correctness blocker in the first-token replay gate. I reviewed this PR as the incremental diff from #1561 (`a81797cf1`) to exact head `e5c698939`.

The follow-up correctly excludes Mirostat and noncanonical sampler order, but the canonical chain still contains two RNG-backed cases at `temperature = 0`: XTC and dynamic temperature. Because first-token record and lookup both use this predicate, either case can replay the first request's sampled token for later requests instead of sampling per request.

Please make replay eligibility follow the effective native sampler behavior and add record-and-lookup regressions for both cases. A conservative fix is to reject replay whenever `xtc.probability > 0` or `dynatemp_range > 0`; for XTC, the exact no-op condition at the pinned upstream also includes `xtc.threshold > 0.5` if you intentionally want the narrower gate.

Native evidence at the PR's pinned llama.cpp commit `cc83d7b4824f73cfdda4dfbb47ee39804f71b328`: `src/llama-sampler.cpp`'s XTC implementation draws `std::uniform_real_distribution<float>` whenever probability is positive, threshold is at most 0.5, and at least two candidates remain; `temp_ext` with positive delta computes an entropy-derived temperature, and the Skippy patch then appends `llama_sampler_init_dist(seed)`. The fields are user-reachable in `frontend/request.rs:698-719`.

Requested regressions:
- `temperature = 0`, `xtc.probability > 0`, `xtc.threshold <= 0.5` => neither record nor lookup a first token.
- `temperature = 0`, `dynatemp_range > 0` => neither record nor lookup a first token.

I did not modify the branch. `git diff --check a81797cf1..e5c698939` passed; current CI is green apart from the still-queued aggregate Linux result at review time, but CI does not exercise these sampler combinations.

i386:
@michaelneale fixed on exact #1569 head `cbc55180b`.

- Replay now fails closed whenever `xtc.probability > 0` or `dynatemp_range > 0`, in addition to the existing Mirostat and canonical-chain guards.
- The unit gate covers both new zero-temperature RNG paths.
- The production record/lookup regression now covers Mirostat, a custom chain, XTC with probability 0.5 at the default threshold, and dynatemp range 0.5. Every case proves no token is recorded and no pre-existing token is returned.

Exact-head validation: 625 server tests passed with 3 ignored; warning-denying all-target server Clippy, formatting, and diff checks pass. The descendants were cleanly rebased: #1570 `943f7de85`, #1572 `74a335b07`.

crates/skippy-server/src/frontend/sampling_cache_key.rs
i386: Fixed at exact head `cbc55180bc2d68ced8934e5cb5ccb0c30eb570f0`. Replay eligibility now rejects both `xtc.probability > 0` and `dynatemp_range > 0`; the unit gate and production record/lookup regression cover XTC, dynatemp, Mirostat, and custom sampler order. Exact-head validation: 625 server tests passed with 3 ignored, warning-denying all-target Clippy passed, and the descendants were rebased onto this head.

PR #1570, opened 2026-08-31
fix(skippy): harden KV cache ownership and runtime bounds

i386:
## Summary

This PR is stacked on #1569 and implements the next KV/scheduler robustness layer from the end-to-end review.

- bound first-token and replay metadata with one shared sampling-aware LRU budget
- clear orphaned split-prefill state during abnormal binary connection cleanup
- make blob release validation atomic across composite KV/recurrent payloads and reject accounting underflow
- prevent replacement of recurrent radix entries while restore references are active
- track resident sequence allocation explicitly, quarantine failed native rollbacks, and fail cache-off on ownership drift
- expose exact-record worker health, recover bookkeeping after panics, and reject later work fail-closed
- attach request IDs, cancellation, and absolute deadlines to cache runtime work; checkpoint long prefills and deferred suffix chunks
- carry those deadlines and cancellation tokens through queued direct iterations, reject stale work before native execution, and suppress late results
- add reconciliation telemetry for radix/blob refs, resident sequences, auxiliary maps, reservations, and worker/cache health

The proposed largest-contiguous-span admission gate was deliberately removed during review. llama.cpp normal allocation uses `find_slot(..., false)` and Mesh prefills are chunked, so requiring the whole request to fit a contiguous span would reject valid fragmented capacity.

## Validation

Exact head: `789a385b20f90d005dcc42e5d64d3983f67cb0ca`

- `cargo test -p skippy-cache --quiet`: 56 passed
- `cargo test -p skippy-server --lib`: 644 passed, 3 ignored
- `cargo test -p skippy-scheduler --quiet`: 38 unit + 8 integration passed
- warning-denying Clippy for `skippy-cache` and `skippy-server`
- shipped `mesh-llm` check passes; shipped warning-denying Clippy remains blocked by the same 28 unrelated `unfulfilled_lint_expectations` inherited from the stacked base
- scheduler lab release benchmark completed all scenarios
- deterministic llama replay: upstream `cc83d7b4824f73cfdda4dfbb47ee39804f71b328` -> patched `dca0b2e68f749d5d1a66cac6d8bb88ea5b5cbd6f`; static Metal build completed all 377 steps
- pre-review P1 release binaries:
  - `skippy-correctness`: `d280d0f2965dde447ecd300dc241b5967bca017612c3b48e5c3d6b84e7de1ffa`
  - `skippy-server`: `e3d9af2f76307fafe61554b225fb790d397ae6a7d533114cab63ae0747eee491`
- live two-lane cache gate passed for dense Qwen3 ResidentKv, Falcon-H1 KvRecurrent, and Qwen3Next KvRecurrent; every case matched suffix prefill and 2/2 repeated cache hits
  - gate JSON: `c1ac00ba484dd1bd5d98609f008bf36ce296681dc8b603b351e68422e6772d01`
  - table: `8454fcb2cb7bcdc50fe1503bd2819810e2b759978844eb48ffa2e264e836ef60`
- review follow-up adds deterministic rollback-branch tests, poison-safe resident ownership locking, permanent quarantine after cleanup double-failure, typed operator warnings, and original-error preservation when corrupt exact-state quarantine also fails
- the exact review-fix head is under GitHub CI; the complete server suite, formatting, warning-denying Clippy, console-output ratchet, and stack ancestry are green

## CUDA promotion evidence

The pre-rebase `e698b266` CUDA release binary is `3f551f67ddd511f33226153e5417e14c0308822aa6e726f2e898127e14b95723`. The capacity-matched Thoughtworks replay completed 12/12 cells and 1,792/1,792 requests with zero failures:

| concurrency | raw llama.cpp | Mesh | vLLM | SGLang |
|---:|---:|---:|---:|---:|
| 64 | 17.4 | **469.4** | 281.8 | 256.4 |
| 128 | 15.7 | **714.3** | 427.7 | 412.2 |
| 256 | 15.4 | **736.8** | 577.2 | 609.9 |

Full-run artifact hashes:

- `summary/thoughtworks.csv`: `64ecaa7c4fdb9cfa41cc434cb3960bf695799101a23d69287d832044203521a5`
- `summary/REPORT.md`: `408953ca045e30c4ecfe9432327a985d32c9737cf42b1753feed6dd9015e989f`
- `artifact-sha256.txt`: `4a1c60a176637ef88328e3c3a2f818f334f064db02aa1da16a30e17e44373a8c`

The full run's c256 Mesh row was lower than the earlier #1569 promotion run, so it was not accepted without investigation. A focused exact-P1 repeat recovered to 770.0 tok/s with 256/256 requests and zero failures. A consecutive same-host, same-config exact-P0 control produced 785.6 tok/s, a 2.0% branch delta. Both had 767/768 resident hits and approximately 47 mean suffix-prefill tokens; P1 had slightly better generation p50/p95, while a synchronized p99/max tail wave accounted for the remaining aggregate difference. This is treated as report-only run variance, not a correctness failure, and motivates the proposed repeated-run longitudinal performance gate rather than a single-sample threshold.

Focused artifact hashes:

- P1 repeat CSV: `5b80d9a57893b47f701e4946df7d32c65bd3ed4651175a6bc062451951432221`
- P1 repeat report: `263bb987cdf041c2f14b5d23c5d688ec98eaf67d3b257d95f375e695b297c3f6`
- P1 repeat manifest: `3685c7e5ed8970a5f67115eb21d4bd1043ee3c5261a6de03e8277474859c2cef`
- P0 control CSV: `50bfcb04d5567c903d1c611795cd181bc8a1e85170d859ff49be378fcb77f667`
- P0 control report: `5e151cf0d019346da2dfd5ff1b9ac828bcfb83b0380f945964181f282890cd2a`
- P0 control manifest: `1952f21b498dacfec69d07d809508cb9a421febcd0e169816016b648d5a25f45`

Every listed artifact manifest verifies, and no benchmark/backend process remains after teardown.

### DIFF
--- a/crates/skippy-cache/src/payload/blob_store.rs
+++ b/crates/skippy-cache/src/payload/blob_store.rs
@@ -1,5 +1,7 @@
 use std::{collections::HashMap, sync::Arc, time::Instant};
 
+use anyhow::{Result, bail};
+
 use super::CacheBytes;
 use super::bytes::{CacheBlockRef, CacheBytesRepr};
 
@@ -37,8 +39,31 @@ impl CacheBlobStore {
         let len = bytes.len;
         let bytes = match bytes.repr {
             CacheBytesRepr::Inline(bytes) => bytes,
-            repr @ CacheBytesRepr::Blocks(_) => {
-                return (CacheBytes { len, repr }, CacheDedupeStats::default());
+            CacheBytesRepr::Blocks(blocks) => {
+                let mut stats = CacheDedupeStats::default();
+                for block in blocks.iter() {
+                    stats.block_count = stats.block_count.saturating_add(1);
+                    let entry = self.blocks.entry(block.hash.clone()).or_insert_with(|| {
+                        self.physical_bytes =
+                            self.physical_bytes.saturating_add(block.bytes.len() as u64);
+                        stats.new_block_count = stats.new_block_count.saturating_add(1);
+                        CacheBlob {
+                            bytes: block.bytes.clone(),
+                            ref_count: 0,
+                        }
+                    });
+                    if entry.ref_count > 0 {
+                        stats.reused_block_count = stats.reused_block_count.saturating_add(1);
+                    }
+                    entry.ref_count = entry.ref_count.saturating_add(1);
+                }
+                return (
+                    CacheBytes {
+                        len,
+                        repr: CacheBytesRepr::Blocks(blocks),
+                    },
+                    stats,
+                );
             }
         };
         let mut blocks = Vec::new();
@@ -68,22 +93,55 @@ impl CacheBlobStore {
         (CacheBytes::blocks(bytes.len() as u64, blocks), stats)
     }
 
-    pub fn release_bytes(&mut self, bytes: &CacheBytes) {
-        let hashes = bytes.block_hashes().map(str::to_string).collect::<Vec<_>>();
-        for hash in hashes {
-            let mut remove = false;
-            if let Some(entry) = self.blocks.get_mut(&hash) {
-                entry.ref_count = entry.ref_count.saturating_sub(1);
-                if entry.ref_count == 0 {
-                    self.physical_bytes =
-                        self.physical_bytes.saturating_sub(entry.bytes.len() as u64);
-                    remove = true;
-                }
+    pub fn release_bytes(&mut self, bytes: &CacheBytes) -> Result<()> {
+        self.release_bytes_batch(&[bytes])
+    }
+
+    pub(crate) fn release_bytes_batch(&mut self, payloads: &[&CacheBytes]) -> Result<()> {
+        let mut releases = HashMap::<String, u64>::new();
+        for bytes in payloads {
+            for hash in bytes.block_hashes() {
+                let count = releases.entry(hash.to_string()).or_default();
+                *count = count
+                    .checked_add(1)
+                    .ok_or_else(|| anyhow::anyhow!("cache blob release count overflow"))?;
             }
-            if remove {
+        }
+        let mut physical_bytes_to_remove = 0u64;
+        for (hash, count) in &releases {
+            let Some(entry) = self.blocks.get(hash) else {
+                bail!("cache blob release references missing block {hash}");
+            };
+            if entry.ref_count < *count {
+                bail!(
+                    "cache blob release underflow for block {hash}: refs={} releases={count}",
+                    entry.ref_count
+                );
+            }
+            if entry.ref_count == *count {
+                physical_bytes_to_remove = physical_bytes_to_remove
+                    .checked_add(entry.bytes.len() as u64)
+                    .ok_or_else(|| anyhow::anyhow!("cache blob physical byte count overflow"))?;
+            }
+        }
+        if physical_bytes_to_remove > self.physical_bytes {
+            bail!(
+                "cache blob physical byte accounting underflow: bytes={} releases={physical_bytes_to_remove}",
+                self.physical_bytes
+            );
+        }
+        for (hash, count) in releases {
+            let entry = self
+                .blocks
+                .get_mut(&hash)
+                .expect("release prevalidation guarantees block presence");
+            entry.ref_count -= count;
+            if entry.ref_count == 0 {
                 self.blocks.remove(&hash);
             }
         }
+        self.physical_bytes -= physical_bytes_to_remove;
+        Ok(())
     }
 
     pub fn physical_bytes(&self) -> u64 {
@@ -92,6 +150,13 @@ impl CacheBlobStore {
 
     pub fn block_count(&self) -> usize {
         self.blocks.len()
+    }
+
+    pub fn logical_ref_count(&self) -> u64 {
+        self.blocks
+            .values()
+            .map(|block| block.ref_count)
+            .fold(0, u64::saturating_add)
     }
 }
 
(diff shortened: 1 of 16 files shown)


ndizazzo (changes requested):
I found one timeout hole in deferred suffix prefill.

The deadline is checked before the work enters the scheduler queue, but it isn't carried into the queued operation or checked again afterward. Under contention, an expired request can still run and publish its result.

Let's pass the deadline and cancellation state into `DirectIteration`, check it before native execution, and recheck after the wait. A regression test with a blocked scheduler worker would cover this nicely.

The focused cache and server tests pass. The full suite stopped on unchanged Rust 1.98 Clippy errors in `skippy-protocol`.

i386:
@ndizazzo fixed on exact head `a7147ee4d2c9ac81166cab49835b7b9d0e65f4f0`.

- `DirectIteration` now owns the absolute deadline plus cloned cancellation token.
- The scheduler rejects expired or cancelled work before queueing, after queue wait, again after acquiring the runtime lock immediately before native execution, and after the caller wait before publishing a result.
- Deferred suffix-prefill passes its cache-operation deadline/cancellation through; ordinary direct decode now also carries request cancellation.
- Regressions cover an expiring direct iteration behind a blocked scheduler worker, a late worker reply, and queued cancellation state.

Exact-head validation: `cargo test -p skippy-server` 639 passed / 3 ignored; `cargo clippy -p skippy-server --all-targets -- -D warnings`; formatting and diff checks clean.

michaelneale (dismissed):
Requesting changes for the cache-blob ownership issue below. I reviewed this PR as the incremental diff from #1569 (`e5c698939`) to exact head `a7147ee4d`.

I also re-reviewed the new `a7147ee4d` suffix-prefill commit: it carries deadline/cancellation into `DirectIteration`, checks before enqueue, before and after the runtime lock, and after the reply, and adds the requested blocked-worker regression. That resolves the previously requested timeout/cancellation change at the current #1570 head.

The remaining blocker is that re-registering an already-blocked `CacheBytes` value treats its embedded hash/bytes pairs as trusted while accounting by hash. This creates a split identity: blob ownership/release uses the hash, but reads use the caller's retained bytes. Please canonicalize every returned block reference from the store entry and reject (or rehash and reject) a same-hash/different-bytes conflict before incrementing refs. Add a regression that constructs or test-mutates a conflicting block reference and proves the operation cannot return divergent bytes or change accounting.

I did not modify the branch. `git diff --check e5c698939..a7147ee4d` passed. At review time format, Clippy, CI-contract, CPU native/static ABI, and several other current-head jobs were green; Rust-test/host jobs were still running.

michaelneale (approved):
Follow-up after completing the constructor/visibility audit: #1570 has no remaining blocking finding from my review at exact head `a7147ee4d`.

The new suffix-prefill commit fixes the deadline/cancellation hole: state is carried in `DirectIteration`, checked before enqueue, before and after acquiring the runtime lock, and after the worker reply, with a blocked-worker regression.

I am retracting my earlier blob-store blocking classification. `CacheBlockRef` fields, `CacheBytesRepr`, and `CacheBytes::blocks` are private to this payload implementation; all production blocked values originate in `CacheBlobStore` by hashing inline bytes, and clones preserve the hash/bytes pair. A conflicting same-hash/different-bytes value is therefore not constructible through the current safe API absent a BLAKE3 collision. Canonicalizing on re-registration could still be defense in depth, but it is not a current correctness blocker.

`git diff --check e5c698939..a7147ee4d` passed. No branch changes made. Current-head CI was still in progress when this follow-up was submitted.

i386:
Review follow-up is complete at exact head `789a385b20f90d005dcc42e5d64d3983f67cb0ca`.

Beyond the queued suffix-prefill deadline/cancellation fix, this head incorporates the independent robustness suggestions that were safe to land here:

- deterministic tests for every restore rollback outcome: clean rollback, cleanup failure, reacquire failure, position verification failure, nonzero native position, and nonzero tracked tokens
- poison-safe resident sequence locking with a regression that preserves allocated state after a poisoned mutex
- forced permanent quarantine if native save and sequence release both fail
- one-shot typed operator warnings when ownership drift disables cache operations and when exact-state quarantine cannot complete
- preservation of the original restore failure when quarantine cleanup also errors
- retirement of the displaced raw stderr diagnostic through the console-output ratchet

Validation: `cargo test -p skippy-server --lib` passed 644 tests with 3 ignored; warning-denying all-target Clippy, formatting, diff checks, and the console-output repository ratchet are green. #1572 has been rebased onto this exact head without conflicts.

i386:
Final-stack hardware validation is now complete on coverage descendant #1572, which contains this exact `789a385b` P1 source plus tests/workflows only.

- descendant exact head: `713d24634a5a39b308459641bb7343d0e282b966`
- static CUDA binary: `9054404e69016eb99c13f0924bd7fd3d1661cd1304273a8407eb1eeb3bf5a30d`
- 12/12 capacity-matched Thoughtworks cells; 1,792/1,792 successful requests; zero failures
- Mesh c256: 787.4 tok/s versus 736.8 on the pre-review P1 sample
- a low first c64 sample was isolated to a synchronized tail wave; same-binary repeat recovered to 474.0 tok/s versus 469.4 on the prior P1 run
- both full and focused artifact manifests verify; teardown left no serving or benchmark process

Combined with the 644-pass server suite and the deterministic rollback/quarantine tests on this PR, the review-fix head has both local correctness coverage and final-stack live serving evidence.

PR #1571, opened 2026-09-01
Windows nodes without a GPU can serve and join splits

### LINKED ISSUE #1566: Windows nodes without a GPU advertise 0 capacity and cannot serve at all: CPU-only RAM budget is Linux-gated
While testing split mode between two Windows 11 machines on 0.76.0-rc8 (private mesh), the GPU-less laptop joins the mesh fine but the split planner permanently excludes it:

```
Split waiting to retry: split runtime needs at least two participating nodes for
unsloth/Qwen3-8B-GGUF@main:Q4_K_M; found 1 eligible [67fb01a785:3.0GB ...];
excluded [5be658551b:missing_vram]; blockers [missing_vram=1 nodes=[5be658551b]]
```

The laptop ran `serve --join <token> --model "Qwen3-8B-Q4_K_M" --split` with the cpu native runtime. It stays excluded forever, and there is no flag that can fix it.

It is not just split participation. The same zero capacity kills plain local serving on a GPU-less Windows machine:

```
startup=failed  failure=local model requires 5.53 GB but this process has 0.00 GB;
local model-only serving never falls back to a split
```

(that run is rc8 on a machine with plenty of free RAM, GPU hidden with CUDA_VISIBLE_DEVICES=-1 to reproduce the GPU-less case deterministically)

## Root cause

`crates/mesh-llm-system/src/hardware/mod.rs` has a fallback that lets GPU-less nodes advertise 75% of system RAM as their compute budget, but it is compiled for Linux only:

```rust
#[cfg(all(target_os = "linux", any(feature = "skippy-devices", test)))]
fn apply_cpu_only_runtime_budget(survey: &mut HardwareSurvey, metrics: &[Metric], system_ram: u64) {
    if metrics.contains(&Metric::VramBytes) && system_ram > 0 {
        survey.vram_bytes = (system_ram as f64 * 0.75) as u64;
    }
}
```

Both call sites in `apply_skippy_backend_devices_to_survey` (the `gpu_facts()` error branch and the empty-GPU-list branch) are also `#[cfg(target_os = "linux")]`. So on Windows a CPU-only node reports `vram_bytes = 0`, and `split_readiness.rs` maps that straight to the `MissingVram` exclusion.

Two things make this worse than a missing feature:

1. `--max-vram` cannot compensate. It is a ceiling, not a floor: `capped_capacity_bytes` does `min(detected, cap)` and the `(0, Some(cap))` arm in `advertised_capacity_bytes` still resolves to `hw.vram_bytes.min(cap)`, which is 0. A user who reads the flag name and tries `--max-vram 8` gets no change and no hint why.
2. The exclusion's recommendation text says "Check GPU visibility or lower --max-vram only after confirming backend/device detection". On a machine that simply has no GPU, both suggestions dead-end.

The irony is that `read_windows_total_ram_bytes()` already exists in the same file (it queries `Win32_ComputerSystem` through PowerShell, same approach as the existing video controller probe). It is just not wired into this path.

## Fix

Small and symmetric with Linux:

- add a Windows `read_system_ram_bytes()` that wraps the existing `read_windows_total_ram_bytes()`
- widen `apply_cpu_only_runtime_budget` and its two call sites from `target_os = "linux"` to `any(target_os = "linux", target_os = "windows")`
- widen the two existing unit tests the same way (they were compile-skipped on Windows before; they pass now)

I validated the behavior change on real hardware by backporting the patch onto the v0.76.0-rc8 tag and running both binaries on the same desktop (33.5 GB RAM) with the GPU hidden through `CUDA_VISIBLE_DEVICES=-1` to reproduce the GPU-less case deterministically:

- stock rc8, `serve --local-model-only --model <Qwen3-8B gguf>`: startup fails with the "requires 5.53 GB but this process has 0.00 GB" message above
- patched rc8, same command: the capacity gate passes (budget is 75% of RAM, about 25 GB here) and the same 5.53 GB model opens fine on CPU, "Native runtime finished opening model"

Which matches what the same machine would do on Linux.

I left macOS out on purpose (Metal is always present there, so the CPU-only branch is unreachable in practice), and I did not touch the adjacent RAM-offload estimate for GPU nodes (`system_ram` is hardcoded to 0 off Linux in the non-SoC `vram_bytes` computation a few lines below). That one changes advertised capacity for GPU nodes, so it deserves its own discussion if you want it.

Happy to open the PR, the patch is sitting on a branch with tests green on Windows (178 passed for mesh-llm-system with `--features dynamic-native-runtime`).

Environment: Windows 11 Home 26200 (both machines), MeshLLM 0.76.0-rc8, laptop on the cpu native runtime, desktop RTX 4070 Ti on cuda12.

ndizazzo:
A Windows machine without a GPU can now serve and take part in splits.

Before this change a GPU-less Windows node advertised `0.00 GB` of capacity, so:

- plain local serving refused to start — `local model requires 5.53 GB but this process has 0.00 GB` — even with tens of GB of free RAM
- the split planner permanently excluded the node with `missing_vram`, and no flag could fix it (`--max-vram` is a ceiling, so `min(0, cap)` is still 0)

Now a Windows node with no usable GPU budgets 75% of its system RAM as compute
capacity, exactly like a CPU-only Linux node already did. The same 33.5 GB
machine that reported 0.00 GB reports ~25 GB, opens a 5.53 GB model on CPU, and
is eligible for split participation.

Fixes #1566.

## What changed

`crates/mesh-llm-system/src/hardware/mod.rs` had the CPU-only RAM budget behind
`#[cfg(target_os = "linux")]` — both the helper and the two call sites in
`apply_skippy_backend_devices_to_survey` (the `gpu_facts()` error branch and the
empty-GPU-list branch). Windows fell through to `vram_bytes = 0`.

- adds a Windows `read_system_ram_bytes()` that wraps the `read_windows_total_ram_bytes()`
  probe already living in the same file
- widens `apply_cpu_only_runtime_budget` and both call sites to
  `any(target_os = "linux", target_os = "windows")`
- points the legacy Windows collector block at the same `read_system_ram_bytes()`
  helper it now shares with Linux (identical behavior, one probe entry point per
  platform)

macOS is deliberately untouched: Metal is always present there, so the CPU-only
branch is unreachable in practice.

## Follow-up, not in this PR

A few lines below the fix, the RAM-offload estimate for **GPU** nodes still
hardcodes `system_ram = 0` off Linux, so a Windows node *with* a GPU
under-advertises capacity. That one changes advertised capacity for GPU nodes
across the mesh and deserves its own change with its own validation — the
reporter scoped it out for the same reason.

## Validation

Regression tests (`crates/mesh-llm-system/src/hardware/tests.rs`):

- the two existing CPU-only budget tests and the two skippy-backend fallback
  tests now compile and run on Windows as well as Linux (they were
  compile-skipped on Windows before)
- new: zero system RAM leaves capacity at 0 (no bogus budget from a failed probe)
- new: `read_system_ram_bytes()` reports installed memory on every platform that
  has a CPU-only budget — this is the check that would have caught the missing
  Windows probe, and it runs on the Linux test lane

Local (macOS host):

```
cargo fmt --all --check
cargo clippy -p mesh-llm-system --all-targets -- -D warnings
cargo clippy -p mesh-llm-system --all-targets --features dynamic-native-runtime -- -D warnings
cargo test -p mesh-llm-system                                  # 181 passed
cargo test -p mesh-llm-system --features dynamic-native-runtime # 185 passed
cargo check -p mesh-llm
cargo clippy -p mesh-llm --all-targets -- -D warnings
```

The Windows code path cannot be compiled from a macOS host, so the Windows CI
lane is the compile gate here. The reporter validated the same behavior change
on Windows 11 hardware by backporting it onto `v0.76.0-rc8`: stock rc8 fails the
capacity gate with 0.00 GB, patched rc8 opens the same model on CPU, with the
mesh-llm-system suite green on Windows.

---

https://claude.ai/code/session_01KneDsA7cmu2XA4RnaenRf1

### DIFF
--- a/crates/mesh-llm-system/src/hardware/mod.rs
+++ b/crates/mesh-llm-system/src/hardware/mod.rs
@@ -187,7 +187,13 @@ fn read_system_ram_bytes() -> u64 {
     .unwrap_or(0)
 }
 
-#[cfg(all(target_os = "linux", any(feature = "skippy-devices", test)))]
+/// CPU-only nodes have no VRAM to advertise, so they budget from system RAM
+/// instead. Without this a GPU-less node reports zero capacity, which reads as
+/// `missing_vram` to split planning and fails the local capacity gate.
+#[cfg(all(
+    any(target_os = "linux", target_os = "windows"),
+    any(feature = "skippy-devices", test)
+))]
 fn apply_cpu_only_runtime_budget(survey: &mut HardwareSurvey, metrics: &[Metric], system_ram: u64) {
     if metrics.contains(&Metric::VramBytes) && system_ram > 0 {
         survey.vram_bytes = (system_ram as f64 * 0.75) as u64;
@@ -233,6 +239,11 @@ fn read_windows_total_ram_bytes() -> Option<u64> {
         "Get-CimInstance Win32_ComputerSystem | Select-Object -ExpandProperty TotalPhysicalMemory",
     )?;
     parse_windows_total_physical_memory(&output)
+}
+
+#[cfg(target_os = "windows")]
+fn read_system_ram_bytes() -> u64 {
+    read_windows_total_ram_bytes().unwrap_or(0)
 }
 
 /// Per-GPU VRAM in adapter order.
@@ -305,13 +316,13 @@ fn apply_skippy_backend_devices_to_survey(survey: &mut HardwareSurvey, metrics: 
     let gpus = match skippy_devices::gpu_facts() {
         Ok(gpus) => gpus,
         Err(_) => {
-            #[cfg(target_os = "linux")]
+            #[cfg(any(target_os = "linux", target_os = "windows"))]
             apply_cpu_only_runtime_budget(survey, metrics, read_system_ram_bytes());
             return true;
         }
     };
     if gpus.is_empty() {
-        #[cfg(target_os = "linux")]
+        #[cfg(any(target_os = "linux", target_os = "windows"))]
         apply_cpu_only_runtime_budget(survey, metrics, read_system_ram_bytes());
         return true;
     }
@@ -630,7 +641,7 @@ impl Collector for DefaultCollector {
             any(not(feature = "skippy-devices"), feature = "dynamic-native-runtime")
         ))]
         {
-            let system_ram = read_windows_total_ram_bytes().unwrap_or(0);
+            let system_ram = read_system_ram_bytes();
             let want_gpu_info =
                 metrics.contains(&Metric::GpuName) || metrics.contains(&Metric::GpuCount);
             let want_vram = metrics.contains(&Metric::VramBytes);
--- a/crates/mesh-llm-system/src/hardware/tests.rs
+++ b/crates/mesh-llm-system/src/hardware/tests.rs
@@ -641,7 +641,7 @@ fn test_hardware_survey_default() {
     assert!(s.gpus.is_empty());
 }
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "windows"))]
 #[test]
 fn test_cpu_only_runtime_budget_uses_system_ram_when_vram_requested() {
     let mut survey = HardwareSurvey::default();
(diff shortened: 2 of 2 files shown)


ndizazzo:
Closing this as a duplicate of #1567, which already fixes #1566 through the same Windows CPU-only RAM budget path and includes broader follow-up coverage.

PR #1575, opened 2026-09-01
perf(skippy): cut per-token split decode overhead (~10%)

danielwinterw:
## What

Two per-token overhead fixes in the split stage decode path. **~10% faster split decode**, independent of transport.

This PR was originally `--trusted-lan` direct TCP transport; that half has been dropped for now (see below). What remains is the part that actually moved the numbers.

> Stacked on #1581 (split speculative-decode correctness). Review/merge that first; this branch is based on it, so the diff here is only the two perf commits below.

## The per-token fixes

The final-stage serial-decode loop paid two avoidable costs on every token:

1. **Poll-quantized reply wait.** The embedded driver raced the direct-return channel against the tunnelled downstream fallback by sampling the channel between 10ms blocking peeks on the downstream socket. Every token whose reply arrived on the direct-return sink — the standard path — paid a uniform 0–10ms poll-quantization penalty (~5ms/token, measured as a 6.9ms toll against a 0.9ms wire floor). Now it blocks on the return channel and wakes at channel-send latency, checking the fallback with a nonblocking peek each slice.

2. **Four scheduler round-trips where one does work.** The loop made a standalone session-align hop, a prefix-cache lookup that cannot apply to decode messages, the compute dispatch, and a full-prefill record that no-ops unless the cache is recording — three of them cloning the full wire message (activation payload + token history) across the scheduler channel. Now: session alignment rides inside the first runtime closure the message already owns; the lookup and record hops evaluate their runtime-free gates before dispatching (costing nothing when the cache is disabled or the message kind can't use them); session stats are collected only when debug telemetry is on.

## Numbers

Two Mac minis (M1 16GB + M4 16GB), wired Ethernet, Qwen3-8B Q4_K_M split 19/17 layers, 192-token generations at temperature 0, rates from `usage` over wall time, 5 runs per cell:

| tok/s | before | after |
|---|---|---|
| split decode | 12.50–12.65 | 13.65–13.87 |

- **~10% gain, transport-independent.** In the original measurement the tunnel and direct-TCP columns moved by the same amount — the win is the per-token work, not the wire.
- **The split now runs at its compute floor.** Layer-weighted single-node ideal for this pair is (19/36)×93.1ms + (17/36)×49.8ms ≈ 72.6 ms/token; the split measures 72.1–72.7. Everything above stage compute is one ~1ms wire round-trip.

## Why `--trusted-lan` was dropped

On a healthy LAN, direct TCP and the mesh tunnel are at parity once the per-token overhead above is removed — confirmed again on a Qwen3-30B-A3B split (tunnel ~21 tok/s vs direct-LAN ~20 tok/s; the ~0.5–0.8 ms/token transport delta is noise at 30B decode rates). The transport's remaining value is narrow — immunity to control-path degradation (when UDP holepunching is blocked, iroh pins to a ~240ms relay and tunnel decode collapses; direct TCP can't be relay-pinned) — so it's parked on the `parked/trusted-lan-transport` branch to return once we test the regime where it earns its keep, rather than carrying 57 files of transport surface for a healthy-LAN wash.

## Testing

- `cargo check -p skippy-server` clean; `cargo fmt --check` clean.
- `skippy-server` suite green (651 passed / 0 failed), including the event-driven reply-wait sub-poll wake-latency regression test.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved session alignment for more accurate message processing.
  * Reduced unnecessary cache lookups, telemetry collection, and prefill recording when not required.

* **Bug Fixes**
  * Prediction responses now wake immediately when available instead of relying on polling.
  * Improved timeout handling to respect response deadlines and prevent unnecessary delays.
  * Improved fallback response checks while preserving reliable connection behavior.

* **Tests**
  * Added coverage for response wakeups, timeout behavior, fragmented responses, and connection blocking behavior.

### DIFF
--- a/crates/skippy-server/src/frontend/embedded_execution.rs
+++ b/crates/skippy-server/src/frontend/embedded_execution.rs
@@ -445,12 +445,11 @@ fn poll_direct_or_downstream_reply_with_timeouts(
             // `peek` only proves that the first byte has arrived. Tunnelled
             // replies may be fragmented, so decode the complete frame under
             // the remainder of the bounded fallback deadline rather than the
-            // nonblocking mode used for the availability check.
+            // short read timeout used for the availability check.
             let remaining = reply_timeout.saturating_sub(started.elapsed());
             if remaining.is_zero() {
                 break Err(timeout_error());
             }
-            timeout_restore.prepare_blocking_read()?;
             downstream
                 .set_read_timeout(Some(remaining))
                 .map_err(openai_io_error)?;
@@ -478,6 +477,15 @@ fn downstream_reply_available(downstream: &TcpStream) -> OpenAiResult<bool> {
     }
 }
 
+/// Availability peeks run under a short read timeout rather than nonblocking
+/// mode. `O_NONBLOCK` lives on the shared open-file description, so setting it
+/// on this socket would also make writes on any `try_clone()` handle — such as
+/// an `AsyncForwarder` worker forwarding activations to the next stage — fail
+/// with `WouldBlock` once the send buffer fills. A read timeout (`SO_RCVTIMEO`)
+/// affects receives only, so a concurrent cloned writer keeps blocking
+/// semantics.
+const DIRECT_RETURN_PEEK_TIMEOUT: Duration = Duration::from_millis(1);
+
 struct DirectReturnFallbackTimeout {
     downstream: TcpStream,
     previous_timeout: Option<Duration>,
@@ -485,13 +493,15 @@ struct DirectReturnFallbackTimeout {
 }
 
 impl DirectReturnFallbackTimeout {
-    /// Puts the downstream socket into nonblocking mode so availability peeks
-    /// return immediately while the reply wait blocks on the direct-return
-    /// channel instead.
+    /// Give the downstream socket a short read timeout so availability peeks
+    /// return promptly while the reply wait blocks on the direct-return channel,
+    /// without toggling nonblocking mode on the shared file description.
     fn install(downstream: &TcpStream) -> OpenAiResult<Self> {
         let previous_timeout = downstream.read_timeout().map_err(openai_io_error)?;
         let restore_stream = downstream.try_clone().map_err(openai_io_error)?;
-        downstream.set_nonblocking(true).map_err(openai_io_error)?;
+        downstream
+            .set_read_timeout(Some(DIRECT_RETURN_PEEK_TIMEOUT))
+            .map_err(openai_io_error)?;
         Ok(Self {
             downstream: restore_stream,
             previous_timeout,
@@ -499,17 +509,7 @@ impl DirectReturnFallbackTimeout {
         })
     }
 
-    /// Leave nonblocking mode before handing the socket to a frame decode.
-    fn prepare_blocking_read(&mut self) -> OpenAiResult<()> {
-        self.downstream
-            .set_nonblocking(false)
-            .map_err(openai_io_error)
-    }
-
     fn restore(&mut self) -> OpenAiResult<()> {
-        self.downstream
-            .set_nonblocking(false)
-            .map_err(openai_io_error)?;
         self.downstream
             .set_read_timeout(self.previous_timeout)
             .map_err(openai_io_error)?;
@@ -521,7 +521,6 @@ impl DirectReturnFallbackTimeout {
 impl Drop for DirectReturnFallbackTimeout {
     fn drop(&mut self) {
         if !self.restored {
-            let _ = self.downstream.set_nonblocking(false);
             let _ = self.downstream.set_read_timeout(self.previous_timeout);
         }
     }
@@ -628,6 +627,48 @@ mod tests {
         let client = TcpStream::connect(listener.local_addr().unwrap()).unwrap();
         let (server, _) = listener.accept().unwrap();
         (client, server)
+    }
+
+    #[test]
+    fn fallback_guard_keeps_a_cloned_writer_blocking() {
+        use std::io::{Read, Write};
+        // AsyncForwarder clones the downstream socket and writes frames from a
+        // separate worker thread. The fallback-wait guard must not toggle
+        // O_NONBLOCK on the shared open-file description, or those writes would
+        // fail with WouldBlock once the send buffer fills instead of blocking.
+        // A read timeout (SO_RCVTIMEO) affects receives only, so the writer is
+        // unaffected. Regression for michaelneale's review on #1575.
+        const PAYLOAD: usize = 8 * 1024 * 1024;
+        let (downstream, peer) = connected_stream_pair();
+        let writer = downstream.try_clone().unwrap();
+
+        // Peer starts reading only after a delay, forcing the send buffer to
+        // fill so a correct blocking write_all must wait rather than error.
+        let reader = std::thread::spawn(move || {
+            let mut peer = peer;
+            std::thread::sleep(Duration::from_millis(100));
+            let mut sink = vec![0u8; 1 << 16];
+            let mut total = 0usize;
+            while total < PAYLOAD {
+                match peer.read(&mut sink) {
+                    Ok(0) => break,
+                    Ok(n) => total += n,
+                    Err(error) => panic!("peer read failed: {error}"),
+                }
+            }
+            total
+        });
+
+        let guard = DirectReturnFallbackTimeout::install(&downstream).unwrap();
+        let mut writer = writer;
+        // Must block until the peer drains, not fail with WouldBlock (os err 35).
+        writer
+            .write_all(&vec![0u8; PAYLOAD])
+            .expect("cloned writer must keep blocking semantics under the guard");
+        drop(guard);
+        drop(writer);
+        drop(downstream);
+        assert_eq!(reader.join().unwrap(), PAYLOAD);
     }
 
     #[test]
--- a/crates/skippy-server/src/frontend/embedded_execution.rs
+++ b/crates/skippy-server/src/frontend/embedded_execution.rs
@@ -399,20 +399,26 @@ fn poll_direct_or_downstream_reply(
     let mut timeout_restore = DirectReturnFallbackTimeout::install(downstream)?;
     let started = Instant::now();
     let reply_timeout = stage_reply_timeout();
+    // The direct-return sink is the standard reply path, so block on its
+    // channel and wake the moment the reply lands instead of sampling it
+    // between bounded peeks: sampling quantized every reply wait to the poll
+    // interval and cost a uniform 0..poll of added latency per token. The
+    // tunnelled downstream fallback is checked without blocking each slice,
+    // so a fallback reply is still detected within one poll interval.
     let result = loop {
         if let Some(reply) = prediction_return
-            .try_recv_one_of(expected_replies)
+            .recv_one_of_timeout(expected_replies, DIRECT_RETURN_FALLBACK_POLL)
             .map_err(openai_backend_error)?
         {
             break Ok(reply);
         }
         if downstream_reply_available(downstream)? {
             // `peek` only proves that the first byte has arrived. Tunnelled
-            // replies may be fragmented, so retaining the short poll timeout
-            // while decoding the complete frame turns an ordinary partial
-            // arrival into EWOULDBLOCK. Once downstream wins the race, give
-            // the frame the remainder of the bounded fallback deadline.
+            // replies may be fragmented, so decode the complete frame under
+            // the remainder of the bounded fallback deadline rather than the
+            // nonblocking mode used for the availability check.
             let remaining = reply_timeout.saturating_sub(started.elapsed());
+            timeout_restore.prepare_blocking_read()?;
             downstream
                 .set_read_timeout(Some(remaining.max(DIRECT_RETURN_FALLBACK_POLL)))
                 .map_err(openai_io_error)?;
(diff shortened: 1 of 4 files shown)


i386 (commented):
I found one test-reliability issue in the current draft head. The production change otherwise looks internally consistent, and all current workflows are green.

crates/skippy-server/src/frontend/embedded_execution.rs
```diff
@@ -588,24 +606,96 @@ mod tests {
...
+            assert_eq!(reply.predicted, 5);
+            let wake_ms = received_at
+                .duration_since(sent_at.lock().unwrap().unwrap())
+                .as_secs_f64()
+                * 1000.0;
+            max_wake_ms = max_wake_ms.max(wake_ms);
+        }
+        // The wait must be event-driven: a sampling loop quantizes wake-up to
+        // its poll interval (10ms) and reliably exceeds this bound.
+        assert!(
```
i386, line 798: This sub-8ms wall-clock assertion includes socket framing/write, reader-thread scheduling, channel delivery, and receiver wake across five iterations. A correct event-driven implementation can exceed 8ms on a loaded or virtualized CI runner, so this can fail without a regression. Please make the former poll interval injectable/larger in the test and assert a comfortably separated bound, or test the channel wake path deterministically without requiring scheduler latency below 8ms.

danielwinterw: Good catch — fixed in 468b5b618. The absolute sub-8ms bound also covered socket framing, reader-thread scheduling, channel delivery and receiver wake, so it could flake on a loaded/virtualized runner without a real regression.

The fallback poll interval is now injectable: the test widens it to 1000ms and asserts the wake lands within 100ms — an order of magnitude below the poll (so a polling regression still fails) yet far above channel-delivery + scheduler latency (so a correct event-driven wait won't flake). Production is unchanged: the public wait delegates with the 10ms constant.

crates/mesh-llm-host-runtime/src/runtime/local_split/loading.rs
```diff
@@ -536,19 +621,23 @@ pub(super) fn split_runtime_stage_load_request(
         load_mode: settings.load_mode.clone(),
         upstream,
         downstream,
+        transport_auth: direct.map(|direct| direct.token.clone()),
+        direct_bind: direct.is_some() && stage_needs_direct_bind(spec, stage),
```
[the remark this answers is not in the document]
i386: Verified against current head `353337d0d`. This finding is obsolete because the direct-stage transport was removed from this PR before the current reviewed head; `local_split/loading.rs` and the `direct_bind` flow are no longer in the PR diff. No code change is applicable here, so I am resolving the stale thread.

crates/skippy-server/src/frontend/embedded_execution.rs
i386: Re-verified at current head `353337d0d`: the fix from `468b5b618` remains intact. The test injects a 1000 ms fallback interval and requires event-driven delivery within 100 ms, preserving a 10x separation without the original sub-8 ms scheduler-latency assumption. The full `skippy-server` library suite passes: 652 passed, 3 ignored, 0 failed. Resolving.

crates/skippy-server/src/frontend/embedded_execution.rs
```diff
@@ -395,26 +395,51 @@ fn poll_direct_or_downstream_reply(
...
+    // The direct-return sink is the standard reply path, so block on its
+    // channel and wake the moment the reply lands instead of sampling it
+    // between bounded peeks: sampling quantized every reply wait to the poll
+    // interval and cost a uniform 0..poll of added latency per token. The
+    // tunnelled downstream fallback is checked without blocking each slice,
+    // so a fallback reply is still detected within one poll interval.
     let result = loop {
         if let Some(reply) = prediction_return
-            .try_recv_one_of(expected_replies)
+            .recv_one_of_timeout(expected_replies, fallback_poll)
```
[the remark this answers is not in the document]
i386: Fixed in `353337d0d`. Each channel wait is now capped with `remaining.min(fallback_poll)`; the code checks for an exhausted deadline before a downstream blocking read; and fragmented-frame decoding receives exactly the remaining deadline instead of `max(fallback_poll)`. Added an injected-deadline regression using a 50 ms reply timeout and 2 s fallback poll. Validation: full `skippy-server` library suite 652 passed, 3 ignored, 0 failed; package check and strict all-target Clippy pass.

i386:
Re-review complete at exact head `353337d0d`.

I fixed the remaining actionable CodeRabbit finding: the direct-return channel wait is capped by the stage deadline, an exhausted deadline is rejected before a downstream blocking read, and fragmented-frame decoding uses only the remaining time. The new injected-deadline regression uses a 50 ms deadline against a 2 s fallback interval.

The other CodeRabbit thread was stale because direct-stage transport was removed from this PR, and the scheduler-latency test concern was already correctly addressed in `468b5b618`. I replied to and resolved all three outstanding threads.

Validation:
- `cargo test --locked -p skippy-server --lib`: 652 passed, 3 ignored, 0 failed
- `cargo fmt --all --check`: pass
- `cargo check -p skippy-server`: pass
- `cargo clippy -p skippy-server --all-targets -- -D warnings`: pass
- `cargo check -p mesh-llm`: pass

The shipped-binary strict Clippy gate still encounters the unrelated existing `mesh-llm-system` unused `cpu_only_system_ram` warning; this PR does not touch that crate or warning. GitHub checks are currently rolling out on the pushed head. No further actionable review findings from my re-review.

michaelneale (approved):
Reviewed the split-decode changes at exact head `353337d0d` and found no remaining correctness blocker. The two review findings are resolved: the timing regression test now separates event-driven wakeup from its fallback poll without a flaky sub-8 ms scheduler assumption, and direct-return waits/fragmented reads are capped by the stage deadline.

Validation at this head:
- full `skippy-server` library suite: 652 passed, 3 ignored
- full `mesh-llm-host-runtime` library suite: 266 passed, 2 ignored
- worktree clean, no head drift

The old routable-bind CodeRabbit finding concerns direct-transport code removed from this PR and is not applicable to the current four-file diff.

michaelneale (changes requested):
Correction to my earlier approval: I am requesting changes at exact head `353337d0d`.

`DirectReturnFallbackTimeout::install` now calls `set_nonblocking(true)` on `downstream` (`embedded_execution.rs:491-495`). On Unix, `O_NONBLOCK` belongs to the shared open-file description, so it also affects every `TcpStream::try_clone()` handle—and affects writes, not only reads. This downstream socket can also be cloned into `AsyncForwarder` (`async_forwarder.rs:43-55`), whose worker writes frames on another thread (`async_forwarder.rs:139-143`) without retrying `WouldBlock`.

I reproduced the socket behavior independently at this head: after enabling nonblocking mode on one handle, `write_all` of 64 MiB through its clone against a temporarily non-reading peer failed with `WouldBlock` (`os error 35`) rather than honoring the clone's write timeout. The existing fallback tests use a private connected pair and do not exercise a concurrent cloned writer.

Please avoid toggling `O_NONBLOCK` on this shared socket: retain the event-driven channel wait but use a short read-only `SO_RCVTIMEO` for the availability `peek`, or isolate the writer on an independent connection. Add a cloned-socket writer regression. If safety instead depends on an invariant that no async-forward job can be queued or in flight whenever a fallback wait installs this guard, that invariant needs to be demonstrated and mechanically enforced; the current types and guard do not encode it.

The deadline fixes, wake test, full package tests, and CI are otherwise clean. The batched-decode alignment counter and alignment-event attribute issues are lower-severity follow-ups, not part of this blocker.

danielwinterw:
@michaelneale good catch, and thanks for the independent repro — fixed in `bfd5eacda`.

`DirectReturnFallbackTimeout` no longer touches `O_NONBLOCK`. The availability peek now runs under a short read-only `SO_RCVTIMEO` (1ms) instead: a read timeout affects receives only, so a downstream socket cloned into an `AsyncForwarder` worker keeps blocking write semantics on a full send buffer. The event-driven channel wait and the deadline caps from `353337d0d` are unchanged; `install`/`restore`/`Drop` now save and restore only the read timeout.

Added a regression (`fallback_guard_keeps_a_cloned_writer_blocking`): with the guard installed on the downstream socket, an 8 MiB `write_all` through a `try_clone()` handle against a slow-draining peer must block to completion rather than fail with `WouldBlock` (os error 35). It fails against the old nonblocking guard and passes now.

Validation: `skippy-server` embedded_execution suite 6/6 (incl. the new regression, the wake test, and your deadline test), `cargo fmt` clean, `cargo clippy -p skippy-server --all-targets -- -D warnings` clean (the `skippy-protocol` `chunks_exact` lint is a newer-local-toolchain nit, unrelated and not in this diff).

The batched-decode alignment counter and alignment-event attribute follow-ups are noted as separate lower-severity items and left out of this change per your comment.

danielwinterw:
@michaelneale picked up the two alignment follow-ups in `1e77cf9d0`. You named them without detail, so here's how I read them from the code — correct me if I aimed wrong:

- **batched-decode alignment counter**: the `use_decode_frame_batch` path aligns internally from its target token count, but its `SchedulerIterationOutcome.session_alignment` was dropped — `session_auto_align_count`/`trimmed_tokens` stayed 0 for those trims while the lookup/compute paths counted them.
- **alignment-event attribute**: consequently no `stage.binary_session_auto_align` event (with its before/after attributes) fired for batched-decode trims, so that path had no alignment telemetry at all.

Fix: extracted `record_session_auto_align` so all paths emit the same counter and event, and populate it from `outcome.session_alignment` on the batched path (elapsed reported as 0 — the trim is fused into the decode, not a standalone step). Also added a `session_auto_align_trimmed_tokens` attribute to the event so it's self-describing. Telemetry only, no decode-path change.

`skippy-server` lib suite 653/0, clippy `-D warnings` clean. If you meant something narrower (e.g. an attribute on an existing event rather than the batched path's missing event), point me at it and I'll adjust.

crates/skippy-server/src/frontend/embedded_execution.rs
```diff
@@ -445,18 +477,30 @@ fn downstream_reply_available(downstream: &TcpStream) -> OpenAiResult<bool> {
...
 impl DirectReturnFallbackTimeout {
+    /// Give the downstream socket a short read timeout so availability peeks
+    /// return promptly while the reply wait blocks on the direct-return channel,
+    /// without toggling nonblocking mode on the shared file description.
     fn install(downstream: &TcpStream) -> OpenAiResult<Self> {
         let previous_timeout = downstream.read_timeout().map_err(openai_io_error)?;
         let restore_stream = downstream.try_clone().map_err(openai_io_error)?;
         downstream
-            .set_read_timeout(Some(DIRECT_RETURN_FALLBACK_POLL))
+            .set_read_timeout(Some(DIRECT_RETURN_PEEK_TIMEOUT))
```
i386: [P2] Bound every availability peek by the remaining absolute reply deadline. The channel wait at lines 438-440 can consume the final remaining interval; the loop then still calls downstream_reply_available under this fixed 1 ms timeout before recomputing remaining. That permits the bounded reply wait to return after its stated deadline. Recompute remaining immediately before the peek and install min(remaining, DIRECT_RETURN_PEEK_TIMEOUT), or skip the peek when no time remains; tighten the deadline regression so this path is exercised.

crates/skippy-server/src/frontend/embedded_execution.rs
```diff
@@ -585,27 +629,184 @@ mod tests {
...
             let _restore = DirectReturnFallbackTimeout::install(&stream).unwrap();
-            assert_eq!(stream.read_timeout().unwrap(), effective_poll);
+            // Nonblocking mode makes an availability peek on a silent socket
+            // return immediately instead of blocking out the read timeout.
+            let started = Instant::now();
+            let mut byte = [0u8; 1];
+            let peek = stream.peek(&mut byte);
+            assert!(matches!(
+                peek,
+                Err(ref error) if error.kind() == std::io::ErrorKind::WouldBlock
```
i386: [P2] Accept TimedOut here as well as WouldBlock. A timed-out TcpStream peek can report TimedOut on Windows, and downstream_reply_available plus the restored-timeout assertion below already handle both forms. This new regression therefore fails on a valid Windows result. The Windows execution job is skipped for this head, so the green planning check does not exercise it.

i386 (changes requested):
The shared-socket O_NONBLOCK blocker from the earlier head is fixed, and the scheduler-round-trip fusion looks coherent. Local validation at this exact head passed the full skippy-server library suite (653 passed, 3 ignored); attached CI is green. Two small but concrete deadline and Windows-portability gaps remain before approval.

crates/skippy-server/src/frontend/embedded_execution.rs
danielwinterw: Fixed in `38d1c34c7`. The remaining budget is now recomputed immediately before the peek, and the peek is bounded by `min(remaining, peek_timeout)` (breaking out when none remains), so the wait can't return past `reply_timeout`. The peek interval is now injectable and the deadline regression uses a 10ms fallback poll with a 2s peek interval — an unbounded peek would overrun by ~2s, and the tightened <400ms upper bound catches it.

crates/skippy-server/src/frontend/embedded_execution.rs
danielwinterw: Fixed in `38d1c34c7`. The peek assertion now accepts both WouldBlock and TimedOut, matching downstream_reply_available and the restored-timeout assertion below it, so it holds on a valid Windows result.

i386 (approved):

PR #1581, opened 2026-09-01
Fix split speculative decoding: verify-window double-record and backend-sampling rejection

danielwinterw:
## What

Follow-up to #1577, which fixed the driver-side classifier but was merged while the investigation was still running the stage-side failures to ground. With #1577 alone, split speculation still dies at its first verify window — the stage fails `llama_decode` and the driver times out. This PR fixes the two remaining native defects; with it, split speculation completes end-to-end on the two-node lab repro: **7/7 generations green, 0 decode failures, 0 GGML asserts**, where before 0/6 survived the first window.

## Defect A — verify inputs recorded twice on activation-input stages (patch 0043)

#1408 established that activation-frame verification must not rewind `token_history`, because its decode does not eagerly append. #1420 then re-added an eager append inside `skippy_verify_activation_frame`, assuming that rewind still existed. With both in place, every verify input on an activation-input stage that owns the output layers — the **final stage of any split** — was inserted into `token_history` twice and fed to `llama_sampler_accept` twice, double-applying repetition penalties, grammar state and RNG draws. Single-node serving takes the token-batch path and was unaffected.

Fix: drop the eager append; verification is the single recorder for this path, restoring the invariant #1408 documented.

## Defect B — backend sampling rejected multi-output verify batches (patch 0045)

Backend (in-graph) sampling registers a sampler on the context, and `llama_context::decode` limited **every** decode on that context to one output row per sequence — the check fired if *any* sequence had a sampler attached. A sampled verify window must output every row, so the first window failed with `rc=-1`. Diagnosed via the new self-describing errors (below): `llama_decode failed rc=-1 n_tokens=10 embd=1 pos0=32 n_past=32 seq=9 mem_pos_max=31` — positions perfectly consistent, batch well-formed, rejected purely for the sampler workaround, triggered by an *idle lane's* offloaded chain.

Three coordinated changes:

- **verification** detaches this session's backend sampler before executing a window and falls back to CPU-chain row sampling — the chain is untouched and row sampling consumes it exactly as it does for configs that never qualified for offload;
- a chain may be backend-initialized at most once (`llama_sampler_chain_backend_init` asserts on reuse), so the detach is tracked as **retired** on the session; `refresh_backend_sampling` respects it and only a freshly-built chain clears it. Without this, the next session reset re-attached the spent chain and aborted the stage on the assert (reproduced);
- `llama_context::decode` scopes the one-output-per-sequence rejection to sequences that **actually have a sampler attached**, so an idle lane's chain can no longer reject another lane's verify batch. This piece looks upstreamable — the check's own `TODO: avoid this workaround` marks it as known-temporary.

## Diagnostics that made this findable (kept)

- Stage execution errors name the failing message: `kind=VerifyWindow pos_start=32 token_count=10 ...` (Rust).
- Native decode-batch failures carry the return code, batch shape and session/memory positions (patch 0044) — stage hosts don't forward llama.cpp's log callback, so without this the wire error was just `llama_decode failed`.

## Validation

Two Mac minis (M1 + M4), Qwen3-8B Q4_K_M split 19/17 layers, `--speculative-strategy ngram-suffix` (5/32/48, window 32), temperature 0:

| build | result |
|---|---|
| #1577 only | classifier error gone; every windowed generation times out on stage `llama_decode failed` |
| + defect A fix | unchanged (A corrupts sampling state; B is the crash) |
| + defect B detach only | crash moves to `llama_sampler_chain_backend_init called twice` abort on session reset |
| + retired flag | first window still rejected by an idle lane's sampler; 5/6 pass after its lane dies |
| **this PR (all of the above + per-sequence scoping)** | **7/7 generations complete, 0 decode failures, 0 asserts** |

Single-node speculation: 5/5 clean before and after — the defects are split-only; no regression.

`cargo test -p skippy-server` (649) and `-p skippy-prompt` (24) green; patch queue re-applies cleanly.

## Note for reviewers

Throughput with suffix speculation on this freeform workload is ~12.2–12.4 tok/s vs 13.8 plain decode on the same split — the machinery now *works*, but the suffix proposer's accept rate on freeform text doesn't pay for the windows at this scale (single-node shows the same shape: 15.3 vs 20.1). This is a correctness fix, not a speed claim.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate token history entries and sampler effects during multi-stage verification.
  - Enabled sampled verification windows to work alongside backend sampling.
  - Improved handling of decode failures, including cases with missing memory or position data.

- **Diagnostics**
  - Expanded runtime error details with message metadata, token counts, activation sizes, batch information, and return codes to make failures easier to investigate.

### DIFF
--- a/crates/skippy-server/src/binary_transport/binary_messaging/connection.rs
+++ b/crates/skippy-server/src/binary_transport/binary_messaging/connection.rs
@@ -508,7 +508,19 @@ fn handle_binary_connection_messages(
                             Ok((sessions_before, sessions_after, eviction, result))
                         })
                         .map_err(|error| anyhow::anyhow!(format!("{error:#}")))
-                        .context("execute scheduler-owned binary stage message")?;
+                        .with_context(|| {
+                            format!(
+                                "execute scheduler-owned binary stage message \
+                                 kind={:?} pos_start={} token_count={} tokens={} \
+                                 executable_tokens={} activation_bytes={}",
+                                message.kind,
+                                message.pos_start,
+                                message.token_count,
+                                message.tokens.len(),
+                                executable_token_ids.len(),
+                                input_activation_bytes,
+                            )
+                        })?;
                     runtime_lock_wait_ms = outcome.runtime_lock_wait_ms;
                     runtime_lock_hold_ms = outcome.runtime_lock_hold_ms;
                     runtime_lock_acquires = 1;
--- a/third_party/llama.cpp/patches/0043-skippy-record-verify-batches-once-on-activation-inpu.patch
+++ b/third_party/llama.cpp/patches/0043-skippy-record-verify-batches-once-on-activation-inpu.patch
@@ -1,0 +1,52 @@
+From f7b4211b767bcf63a6dd4286629b5d7c18f17cc2 Mon Sep 17 00:00:00 2001
+From: Mesh-LLM CI <ci@mesh-llm.local>
+Date: Tue, 1 Sep 2026 20:04:15 +1000
+Subject: [PATCH] skippy: record verify batches once on activation-input stages
+
+Sampled verification exposes each input to the sampler exactly once and
+in causal order as its row is sampled. The token-batch path rewinds the
+eager append its decode performs before doing so; #1408 established that
+the activation-frame path must NOT rewind, because its decode did not
+append. #1420 then re-added an eager append inside
+skippy_verify_activation_frame, assuming the rewind was still there.
+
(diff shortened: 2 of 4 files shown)


i386 (approved):
Reviewed the exact head against `main`. The changes are narrowly scoped and consistent: duplicate verify-window recording is removed, backend sampling is detached before sampled verification and retired until reconfiguration, and decode failures now retain useful context. All five PR workflows are green at this commit.

PR #1586, opened 2026-09-01
fix(skippy): keep multimodal evaluation on its session lane

### LINKED ISSUE #1551: Multimodal request that ends on finish_reason=length poisons the lane; later image requests fail with status -1 until restart
🤖 Found while verifying #1549 / #1550 on CUDA.

## Summary

A multimodal request that ends with `finish_reason: length` leaves the session unusable. Every
subsequent multimodal request fails, and the lanes are eventually exhausted.

The first multimodal request after startup works correctly, and interleaved text-only requests
are fine — it is specifically a request that hits the token budget that poisons the lane.

## Reproduction

3x Tesla P40 (sm_61), CUDA 12 runtime, `Qwen3.8-27B-ABLITERATED-Q4_K_M.gguf` +
`mmproj-Qwen3.8-27B-ABLITERATED-Q8_0.gguf`, `split_mode = "layer"`,
`tensor_split = "33,33,34"`, `ctx_size = 32768`, `parallel = 2`. Runtime built from `main`
plus #1549 and #1550 (without those the projector cannot load at all).

Post the same 336x336 solid-colour PNG four times to `/v1/chat/completions` with
`max_tokens: 200`, asking for the dominant colour:

```
[img 1] -> (no content, finish_reason = length)
[img 2] -> multimodal prompt evaluation failed with status -1
[img 3] -> multimodal prompt evaluation failed with status -1
[img 4] -> all execution lanes are busy
```

With a budget large enough for the model to finish (`finish_reason: stop`), the same request
can be repeated, and text-only requests interleave without trouble:

```
[blue image]  -> "Azul"    (stop)
[text only]   -> ok
[red image]   -> "Rojo"    (stop)
```

## Native log

The failing request dies on the *first* chunk, before the image:

```
add_text: <|vision_start|>
add_media: preproc_out has 1 entries, grid_x = 0, grid_y = 0, has_overview = 0
image_tokens->nx = 11
image_tokens->ny = 11
batch_f32 size = 1
add_text: <|vision_end|>
add_text: <|im_end|>
<|im_start|>assistant
<think>

failed to decode text
failed to eval chunk 0
```

`failed to eval chunk 0` is the leading text chunk, so the prompt cannot be evaluated at all —
this is not an image-encode failure. That points at session position/KV state left behind by
the truncated request rather than anything in the projector path.

The trailing `all execution lanes are busy` shows the lane is never returned to the pool after
these failures, so the damage accumulates: with `parallel = 2`, two poisoned requests take the
model out of service until restart.

## Expected

A request that stops on `max_tokens` should leave the session in the same state as one that
stops on an EOS token, and a failed multimodal evaluation should release its lane.

## Notes

Not the same as #1314 — there is no degenerate repeated-token output here; generation fails
outright before any token is produced. #1314 did not reproduce on this model/backend.

Possibly related to the class of defect in #1544, where a non-standard exit path cleared the
session bookkeeping without cleaning the cache rows it had already written.

i386:
Fixes #1551.

## Summary

- expose each Skippy session's native llama.cpp sequence ID through ABI 0.1.45 and pass it to both mtmd evaluation paths instead of hard-coding sequence 0
- guard the entire local multimodal operation so every post-admission error releases its runtime lane, while preserving receipt-before-cleanup ordering
- add a regular deterministic cleanup regression plus fixture-backed regressions for repeated prefill failures and `finish_reason=length` followed by another image request

## Root cause

With parallel lanes, the admitted `StageSession` could own sequence 1 while mtmd wrote media KV cells to sequence 0. Length completion then reset the owning sequence, leaving sequence 0 dirty for the next image request. Separately, local media prefill happened before the unconditional cleanup boundary, so a failed prefill leaked its admitted lane.

## Validation

- `cargo fmt --all --check`
- `git diff --check origin/main...HEAD`
- `cargo test -p skippy-ffi --lib` (8 passed)
- `cargo test -p skippy-ffi --lib --no-default-features` (7 passed)
- `cargo test -p skippy-runtime --lib` (111 passed)
- `cargo test -p skippy-server --lib` (653 passed, 3 ignored)
- `cargo clippy -p skippy-ffi --all-targets -- -D warnings`
- `cargo check -p mesh-llm`
- patch 0046 applied cleanly after the existing ABI-44 patch stack, and the resulting ABI-45 native runtime built successfully

The real multimodal regressions compile in the package suite and run when the `SKIPPY_MM_*` fixture variables are supplied; the affected CUDA model/projector fixture was not available locally.

## Existing base failures

- `cargo clippy -p mesh-llm --all-targets -- -D warnings` fails on the pre-existing unused `cpu_only_system_ram` parameter in `crates/mesh-llm-system/src/hardware/mod.rs:351`
- a clean `scripts/prepare-llama.sh pinned` replay currently fails on existing patch 0035 before reaching this PR's patch; patch 0046 was separately applied and built on the prepared ABI-44 tree

## AI assistance

Implemented and reviewed with agent assistance; all changes were independently reviewed and validated on commit `4c6391c48bef8ee9705442b09a382ddfe2e7266a`.

## Summary by CodeRabbit

- **New Features**
  - Added APIs for inspecting model input and output activation boundaries.
  - Added session sequence ID access for integrations using multimodal processing.
  - Added model capability detection for recurrent, hybrid, and diffusion architectures.
  - Added activation metadata and related configuration options to the native API.

- **Bug Fixes**
  - Improved multimodal evaluation by using the session’s correct sequence.
  - Prevented failed multimodal requests from leaking session state.
  - Improved cleanup and recovery after prefill or generation failures.

- **Documentation**
  - Updated the native API reference and ABI information for the new capabilities.


i386:
CI triage: both failed Linux native jobs stop while replaying the existing patch stack at patch 0035:

```text
Applying: skippy: account retained tensors during staged loads
error: sha1 information is lacking or useless (src/CMakeLists.txt).
Patch failed at 0035 skippy: account retained tensors during staged loads
```

They fail before reaching this PR's new patch 0046. The same clean replay failure was reproduced locally from current `main`; patch 0046 itself applies cleanly after the prepared ABI-44 tree, and that ABI-45 tree builds. This base blocker is also documented in the PR description.

PR #1591, opened 2026-09-02
chore(system): silence cfg-gated macOS clippy warnings

i386:
## Problem

macOS clippy runs with `-D warnings` fail in `crates/mesh-llm-system/src/hardware/mod.rs`:

- `unused variable: cpu_only_system_ram` — the closure parameter of `apply_gpu_probe_outcome_to_survey` is only consumed on the linux/windows CPU-only fallback branches
- `function cpu_only_budget_system_ram is never used` (plain `--all-targets` without features) — its only caller is `#[cfg(feature = "skippy-devices")]`

## Fix

Gate the `allow` attributes to exactly the configurations where the code is dead:

- `#[cfg_attr(not(any(target_os = "linux", target_os = "windows")), allow(unused_variables))]` on the parameter
- `#[cfg_attr(not(any(feature = "skippy-devices", target_os = "linux", target_os = "windows")), allow(dead_code))]` on the helper

On linux/windows the lint stays live, so a genuinely unused variable there still fails the build.

## Verification

- `cargo clippy -p mesh-llm-system --all-targets` — clean (both warning kinds gone)
- `cargo clippy -p mesh-llm-system --all-targets --features skippy-devices` — clean
- `cargo test -p mesh-llm-system` — 182 passed / 0 failed
- `cargo fmt --check -p mesh-llm-system` — clean

Follow-up to #1590 (James's "two things" ask in the fix-1 thread).

## Summary by CodeRabbit

* **Chores**
  * Suppressed platform-specific compiler warnings during builds.
  * No user-facing behavior or functionality has changed.

### DIFF
--- a/crates/mesh-llm-system/src/hardware/mod.rs
+++ b/crates/mesh-llm-system/src/hardware/mod.rs
@@ -327,6 +327,10 @@ fn apply_skippy_backend_devices_to_survey(survey: &mut HardwareSurvey, metrics: 
 }
 
 #[cfg(any(feature = "skippy-devices", test))]
+#[cfg_attr(
+    not(any(feature = "skippy-devices", target_os = "linux", target_os = "windows")),
+    allow(dead_code)
+)]
 fn cpu_only_budget_system_ram() -> u64 {
     #[cfg(any(target_os = "linux", target_os = "windows"))]
     {
@@ -348,6 +352,10 @@ fn apply_gpu_probe_outcome_to_survey<E>(
     survey: &mut HardwareSurvey,
     metrics: &[Metric],
     probe: Result<Vec<GpuFacts>, E>,
+    #[cfg_attr(
+        not(any(target_os = "linux", target_os = "windows")),
+        allow(unused_variables)
+    )]
     cpu_only_system_ram: impl FnOnce() -> u64,
 ) -> bool {
     let gpus = match probe {


i386:
Verdict: **this is correct; it is not papering over a missing macOS implementation.**

The parameter being allowed here is specifically the deferred **CPU-only 75% system-RAM fallback** added in #1567. That fallback is intentionally Linux/Windows-only. macOS capacity is implemented through a different contract:

- backend-enumerated `MTL`/integrated devices are marked unified-memory devices and derive their budget from Metal's `recommendedMaxWorkingSetSize` in `hardware/skippy_devices.rs:60-75`;
- when the dynamic native runtime is unavailable, the platform collector uses the same Metal working-set query in `hardware/mod.rs:417-433`.

When a loaded Skippy backend returns an error or no accelerator, the current policy is deliberately backend-authoritative and fail-closed: macOS does not substitute platform-visible Metal capacity or a generic percentage of RAM. `test_probe_fallback_leaves_vram_untouched_on_macos` explicitly locks that behavior (`hardware/tests.rs:720-740`), consistent with `docs/design/TESTING.md:39-40`. #1567 also called macOS deliberately out of scope because the Metal-backed CPU-only branch is unreachable in normal operation.

That makes the two warnings structurally expected:

- the helper is dead only in the non-`skippy-devices` all-targets/test build on non-Linux/Windows;
- the closure parameter is unused only where the Linux/Windows fallback branch is compiled out.

The `cfg_attr` predicates match those cases while keeping the lints live on the platforms/configurations where the fallback is real. There is no behavior change.

Exact-head verification at `52fd6209eb5ec1ecb1349f524b28b2ecc60108c4` on macOS:

- `clippy -p mesh-llm-system --all-targets -- -D warnings`: passed;
- the same with `--features skippy-devices`: passed.

The full package test run reached 181/182; the sole failure was the pre-existing/environment-sensitive `test_survey_returns_all_metrics` observing different live GPU visibility across two successive surveys. This PR only adds lint attributes and cannot affect that path.

No blocker from me. If we want macOS to advertise a CPU execution budget when a loaded Metal backend fails to enumerate an accelerator, that is a separate product-policy change; wiring this Linux/Windows RAM closure into macOS would not be the correct implementation.

PR #1601, opened 2026-09-02
Reuse warm prefixes across agent sessions

i386:
Agent requests that share the same stable system/tool scaffold can now reuse cache-affinity evidence across different users and task prompts. Conversation stickiness remains keyed independently, so sharing warm-prefix evidence no longer collapses distinct sessions onto one sticky identity.

An authoritative successful local response reporting zero cached prompt tokens now removes both the advertised positive entry and the short routing lease immediately. Missing cache usage remains non-authoritative and does not destroy valid evidence.

## Migration note

**`prompt_cache_key` no longer drives session stickiness.** Before this PR, `prompt_cache_key` was the first entry in `session_hint_keys`, so it acted as both a cache-affinity key and a session-sticky routing hint. After this PR it is used only to derive the prefix hash; `user` and `session_id` are the sole session-stickiness keys. Operators or callers that relied on `prompt_cache_key` to pin a session to a specific node will silently lose that behavior — sessions will be routed to whichever node holds the warmest prefix for the key, which may differ across requests.

## Architecture

- separate explicit `prompt_cache_key` identity from `user` / `session_id` stickiness
- key implicit cache evidence by the stable system/developer/tool scaffold, falling back to the first user prompt only where the host compatibility policy already allows it
- honor explicit prompt cache keys consistently in host and embedded-client routing
- invalidate the model/prefix inventory entry and exact short lease after an authoritative local miss

## Protocol

No wire-format change. Existing salted cache-affinity advertisements remain compatible with mixed-version peers; only the local derivation and invalidation policy changes.

## Validation

- `cargo test -p mesh-llm-routing --lib` — 18 passed
- `cargo test -p mesh-llm-client --lib` — 92 passed
- `cargo test -p mesh-llm-host-runtime --lib -- --quiet` — 2,902 passed, 8 ignored
- `cargo check -p mesh-llm-routing`
- `cargo check -p mesh-llm-client`
- `cargo check -p mesh-llm`
- warning-denying Clippy passed for routing and client
- host-runtime and shipped-binary Clippy pass after allowing the pre-existing `unfulfilled_lint_expectations` warnings; the strict command fails on those untouched baseline expectations
- `cargo fmt --all --check`

PR #1578 remains the independent exact-commit Agentic Replay benchmark foundation and will be run over this stack after the cache-specific workload assertions are added.

Originating Buzz channel: `10b70ee2-a3fa-41e2-a87f-b5b53792a7c1`

## Summary by CodeRabbit

* **Improvements**
  * Improved prompt-cache routing so requests with the same cache key can share cached prefixes across different users.
  * Kept session-based routing separate from prompt-cache routing for more predictable session affinity.
  * Improved cache lease handling to prevent stale routing decisions after cache invalidation.
  * Updated local and remote cache evidence handling to preserve accurate local cache state.
* **Tests**
  * Added coverage for cache invalidation, concurrent routing changes, and cross-user prompt-cache reuse.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/affinity.rs
+++ b/crates/mesh-llm-host-runtime/src/network/affinity.rs
@@ -243,6 +243,21 @@ impl AffinityRouter {
         }
     }
 
+    pub(crate) fn forget_cache_lease(&self, model: &str, prefix_hash: u64) -> bool {
+        let mut state = self.inner.lock().unwrap();
+        let key = CacheLeaseKey {
+            model: model.to_string(),
+            prefix_hash,
+        };
+        let removed = state.cache_leases.remove(&key).is_some();
+        if removed
+            && let Some(position) = state.cache_lease_lru.iter().position(|item| item == &key)
+        {
+            state.cache_lease_lru.remove(position);
+        }
+        removed
+    }
+
     pub(crate) fn forget_cache_leases_for_target(
         &self,
         target: &election::InferenceTarget,
@@ -325,10 +340,7 @@ pub use shared_affinity::{PreparedTargets, TargetSelection};
 
 #[cfg(test)]
 pub(crate) fn extract_session_hint_from_body(body: &Value) -> Option<String> {
-    shared_affinity::extract_session_hint_from_body(
-        body,
-        &["prompt_cache_key", "user", "session_id"],
-    )
+    shared_affinity::extract_session_hint_from_body(body, &["user", "session_id"])
 }
 
 #[cfg(test)]
@@ -339,7 +351,8 @@ fn scaffold_prefix_hash_from_body(body: &Value) -> Option<u64> {
 fn routing_keys(parsed_body: Option<&Value>) -> RoutingKeys {
     shared_affinity::routing_keys(
         parsed_body,
-        &["prompt_cache_key", "user", "session_id"],
+        &["prompt_cache_key"],
+        &["user", "session_id"],
         true,
     )
 }
@@ -369,6 +382,15 @@ impl crate::mesh::Node {
                 suffix_prefill_tokens,
                 queue_delay_micros,
             );
+    }
+
+    /// Remove provider-refuted local cache evidence immediately instead of
+    /// waiting for its bounded gossip TTL to expire.
+    pub(crate) fn invalidate_local_cache_evidence(&self, model: &str, prefix_hash: u64) -> bool {
+        self.cache_affinity_inventory
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .invalidate(model, prefix_hash)
     }
 
     /// Probe bounded local and peer advertisements for the exact request
@@ -527,12 +549,12 @@ mod tests {
     }
 
     #[test]
-    fn test_extract_session_hint_from_body_prompt_cache_key_preferred() {
+    fn test_extract_session_hint_from_body_user_preferred() {
         let body =
             parse_body(r#"{"prompt_cache_key":"cache-1","user":"bob","session_id":"sess-1"}"#);
         assert_eq!(
             extract_session_hint_from_body(&body),
-            Some("cache-1".to_string())
+            Some("bob".to_string())
         );
     }
 
@@ -595,7 +617,26 @@ mod tests {
     }
 
     #[test]
-    fn test_routing_keys_prefix_is_namespaced_by_first_user() {
+    fn exact_cache_lease_invalidation_preserves_other_prefixes() {
+        let affinity = AffinityRouter::with_config(true, true);
+        let target = remote(1);
+        affinity.remember_cache_lease(TEST_MODEL, 7, &target);
+        affinity.remember_cache_lease(TEST_MODEL, 8, &target);
+
+        assert!(affinity.forget_cache_lease(TEST_MODEL, 7));
+        assert!(!affinity.forget_cache_lease(TEST_MODEL, 7));
+        assert_eq!(
+            affinity.lookup_cache_lease(TEST_MODEL, 7, std::slice::from_ref(&target)),
+            None
+        );
+        assert_eq!(
+            affinity.lookup_cache_lease(TEST_MODEL, 8, std::slice::from_ref(&target)),
+            Some(target)
+        );
+    }
+
+    #[test]
+    fn test_routing_keys_prefix_is_shared_across_first_users() {
         let req_a = parse_body(
             r#"{"tools":[{"type":"function","function":{"name":"run"}}],"messages":[{"role":"system","content":"You are an agent."},{"role":"user","content":"fix bug A"}]}"#,
         );
@@ -606,7 +647,7 @@ mod tests {
         let keys_a = routing_keys(Some(&req_a));
         let keys_b = routing_keys(Some(&req_b));
 
-        assert_ne!(keys_a.prefix_hash, keys_b.prefix_hash);
+        assert_eq!(keys_a.prefix_hash, keys_b.prefix_hash);
         assert_eq!(keys_a.sticky_hash, None);
         assert_eq!(keys_b.sticky_hash, None);
     }
@@ -723,6 +764,24 @@ mod tests {
             node.select_cache_target("qwen", 0xfeed_beef, &candidates)
                 .await,
             Some(election::InferenceTarget::Local(9337))
+        );
+    }
+
+    #[tokio::test]
+    async fn provider_confirmed_local_miss_invalidates_evidence() {
+        let node =
+            crate::mesh::Node::new_for_tests(crate::mesh::NodeRole::Host { http_port: 9337 })
+                .await
+                .expect("test node");
+        let candidates = vec![remote(1), election::InferenceTarget::Local(9337)];
+
+        node.record_local_cache_hit("qwen", 0xfeed_beef, 512, 24, 0);
+        assert!(node.invalidate_local_cache_evidence("qwen", 0xfeed_beef));
+
+        assert_eq!(
+            node.select_cache_target("qwen", 0xfeed_beef, &candidates)
+                .await,
+            None
         );
     }
 
--- a/crates/mesh-llm-host-runtime/src/network/affinity.rs
+++ b/crates/mesh-llm-host-runtime/src/network/affinity.rs
@@ -68,6 +68,7 @@ struct AffinityState {
     auto_lru: VecDeque<u64>,
     cache_leases: HashMap<CacheLeaseKey, CacheLeaseEntry>,
     cache_lease_lru: VecDeque<CacheLeaseKey>,
+    cache_lease_epoch: u64,
 }
 
 #[derive(Clone)]
@@ -211,6 +212,7 @@ impl AffinityRouter {
         Some(target)
     }
 
+    #[cfg(test)]
     pub(crate) fn remember_cache_lease(
         &self,
         model: &str,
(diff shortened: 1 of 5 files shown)


i386:
**Review: Correct and testable. One behavioral change needs a migration note.**

**`prompt_cache_key` no longer drives session stickiness** — it was previously the first key in `session_hint_keys`, so it acted as both a cache-affinity key and a session-sticky routing hint. After this PR it drives prefix-hash only; `user`/`session_id` are the sole stickiness keys. Callers using `prompt_cache_key` to pin a session to a node will silently lose that behavior. Worth a CHANGELOG or migration note since there's no wire-format signal to indicate the change to operators.

**New prefix hash is correct**: `prefix_hash = explicit_cache_hash.or(scaffold_hash)`. The first-user-hash is deliberately dropped so two users running the same agent scaffold share evidence. `sticky_hash` stays user-namespaced. The `shared_scaffold_reuses_evidence_across_users_and_tasks` test covers this clearly.

**`forget_cache_lease` O(n) LRU scan** (`iter().position()`) is consistent with the existing `forget_cache_leases_for_target` pattern — not a regression, fine at current cap sizes.

**Test coverage** for `exact_cache_lease_invalidation_preserves_other_prefixes`, `provider_confirmed_local_miss_invalidates_evidence`, and `missing_cache_usage_does_not_refute_positive_evidence` covers exactly the three properties that matter here.

Approachable once the stickiness behavior change is documented.

i386:
@i386 migration note added to the PR description under a **Migration note** section. Thanks for catching this — the `prompt_cache_key` stickiness change was intentional but definitely worth surfacing explicitly for operators.

crates/mesh-llm-host-runtime/src/network/affinity.rs
```diff
@@ -339,7 +351,8 @@ fn scaffold_prefix_hash_from_body(body: &Value) -> Option<u64> {
 fn routing_keys(parsed_body: Option<&Value>) -> RoutingKeys {
     shared_affinity::routing_keys(
         parsed_body,
-        &["prompt_cache_key", "user", "session_id"],
+        &["prompt_cache_key"],
+        &["user", "session_id"],
```
ndizazzo, line 395: Blocking: `auto_model_session_key()` returns this wrapper's `sticky_hash`, so removing `prompt_cache_key` from the session hints makes a `model=auto` request with only `prompt_cache_key` produce no auto-session key. `lookup_cached_auto_model` then short-circuits at `auto_session_key?`, and the new choice is never stored because `remember_auto_model` is guarded by the same `Option`. The migration note covers target stickiness, but not loss of conversation-to-model memory or possible model changes between turns. Please keep `prompt_cache_key` in the auto-model key path, or explicitly document and test that separate behavior.

crates/mesh-llm-host-runtime/src/network/openai/transport_route_model.rs
```diff
@@ -356,6 +341,53 @@ fn handle_delivered_route_model_attempt(
...
+        );
+        return;
+    }
+
+    let inventory_invalidated = context
+        .node
+        .invalidate_local_cache_evidence(context.model, prefix_hash);
+    let lease_invalidated = context
+        .affinity
+        .forget_cache_lease(context.model, prefix_hash);
```
ndizazzo, line 382: Blocking: `forget_cache_lease` here is not durable against an in-flight selector. `Node::select_cache_target` copies local evidence before awaiting `self.state.lock()`. Another request can report a zero-token miss and clear inventory and lease here; the suspended selector then resumes and `cache_target_for_request` writes a fresh two-second lease from its stale snapshot. Please fence selection with the inventory epoch or atomically revalidate before the lease write, with a deterministic interleaving test.

crates/mesh-llm-host-runtime/src/network/openai/transport_route_model.rs
```diff
@@ -356,6 +341,53 @@ fn handle_delivered_route_model_attempt(
...
 
+fn update_local_cache_evidence(
+    context: &DeliveredRouteModelContext<'_>,
+    status_code: u16,
+    usage: Option<&TokenUsage>,
+) {
+    if !(200..400).contains(&status_code) {
+        return;
+    }
+    let (Some(prefix_hash), election::InferenceTarget::Local(_), Some(cached_tokens)) = (
```
ndizazzo: This `Local` pattern also prevents the host-local lease clear for a successful Remote response with `cached_prompt_tokens == 0`. Both local and remote transport return parsed usage, and leases can point at remote targets, so an authoritative remote miss leaves the exact lease live until TTL. Keep inventory invalidation local-only, but clear the exact lease independently of target locality.

ndizazzo (changes requested):
I found two blocking correctness issues in the new affinity split and invalidation path. First, removing `prompt_cache_key` from the sticky hash also removes it from `model=auto` model-choice memory, a separate behavior from target stickiness that the migration note does not cover. Second, the authoritative-miss cleanup is not fenced against a concurrent selector that captured positive evidence before an async lock wait, so that request can reinstall the two-second lease after cleanup. A related lower-severity gap leaves host-local leases intact after authoritative misses from remote targets. Protocol and ABI compatibility otherwise check out, and the focused routing and local-miss tests pass, but they do not exercise these cases.

michaelneale:
Independent review at 7839dd9b9. I re-verified the design, the mixed-version story, and each of @ndizazzo's blocking items against the code, and re-ran the suites locally (`mesh-llm-routing` 18 passed, `mesh-llm-client` 92 passed, `mesh-llm-host-runtime` 2,902 passed / 8 ignored). The direction is right and I want this in — cross-session scaffold reuse plus authoritative-miss invalidation is exactly the behavior an agent-fleet mesh needs. Confirming the blockers with pointers, plus one addition:

**1. Confirmed — `model=auto` loses conversation-to-model memory for `prompt_cache_key`-only callers.** `auto_model_session_key()` (`network/affinity.rs:286`) returns `routing_keys().sticky_hash`, which is now derived only from `["user","session_id"]`. A request carrying only `prompt_cache_key` (common for agent harnesses that set it but no `user`) gets `None`, so `resolve_auto_model_request` never remembers or recalls the chosen model, and the model can change between turns of the same conversation. This is a real behavioral regression distinct from target stickiness and not covered by the migration note. Suggested fix: derive the auto-model key from `explicit_cache_hash.or(session_hash)` (a dedicated field on `RoutingKeys`, not `sticky_hash`), so cache-key-only callers keep model memory without reintroducing target stickiness. Add a test with a `prompt_cache_key`-only body asserting `auto_model_session_key` is `Some`.

**2. Confirmed — lease resurrection race.** `Node::select_cache_target` snapshots `probe_local` under the sync inventory lock, then `await`s `self.state.lock()`. If `update_local_cache_evidence` runs in that window (invalidating inventory + lease), the suspended selector resumes with the stale snapshot and `cache_target_for_request` (`transport_route_model.rs:61`) re-writes a fresh 2s lease. Note `CacheInventory::invalidate` already bumps `self.epoch` — so the cheap fence is: capture the epoch with the snapshot, and in `cache_target_for_request` re-check the epoch (or re-probe under the sync lock) immediately before `remember_cache_lease`, skipping the lease write on mismatch. Impact is bounded (~2s of misdirected routing that self-corrects on the next authoritative miss), but the fix is small and it deserves the deterministic interleaving test ndizazzo asked for.

**3. Confirmed — remote authoritative miss leaves the local lease live.** `update_local_cache_evidence` gates *both* invalidations behind `InferenceTarget::Local(_)`, but `remember_cache_lease` stores whatever `select_cache_target` returned, including `Remote` targets, and remote transports return parsed usage too. Split the destructure: keep `invalidate_local_cache_evidence` local-only (peer advertisements aren't ours to mutate; gossip TTL handles them), but call `forget_cache_lease` for any successful response with `cached_prompt_tokens == Some(0)` regardless of target locality. Test: remote target, zero cached tokens, assert lease gone and local inventory untouched.

**Additional (non-blocking): scaffold hotspotting.** With `prefix_hash` keyed by scaffold alone, a fleet of agents sharing one system/tool scaffold now all produce the same prefix hash, so cache-affinity evidence steers them at the same node. `cache_aware::select_cache_target`'s cost filter (queue-delay + prefill cost model) is the back-pressure, and the deterministic `rotate_left(prefix_hash % len)` spread only varies across *different* prefixes, not within one hot prefix. I think the cost model is sufficient, but it would be worth asserting in a test that a saturated warm node (high `queue_delay_micros`) loses to a cold candidate for the shared-scaffold case, since this PR makes that scenario the common one.

**What checks out and doesn't need changes:**
- No wire-format change confirmed: prefix hashes are salted per advertisement and derivation is local. Mixed-version rollout degrades to fewer cross-node reuse hits (old/new nodes namespace differently), never to incorrect routing.
- `missing cached_prompt_tokens` correctly treated as non-authoritative — evidence preserved (test covers it).
- Client-side (`mesh-client`) no longer invents a prefix hash from the first user turn when no scaffold exists — good, that evidence was noise.
- Migration note on `prompt_cache_key` target-stickiness loss is accurate as far as it goes; extend it to cover item 1 once fixed.

All three fixes are localized (routing key derivation, one fence in `cache_target_for_request`, one destructure split in `update_local_cache_evidence`) — none disturb the core design. With those + tests, I'd be comfortable merging.

michaelneale:
Addendum after checking the automatic directive (`mesh` virtual model + deprecated `auto` alias, `network/openai/automatic.rs`): it doesn't change the verdict, but it broadens blocking item 1 and clarifies scope.

**Item 1 is broader than "model=auto".** `auto` is only the deprecated alias — the primary directive is `mesh` (`DIRECTIVE = mesh_mixture_of_agents::VIRTUAL_MODEL_NAME`), and *model-less requests* also resolve through the same auto selector (`is_auto_request` in `transport.rs:478` treats `None` and either directive spelling identically). So the lost conversation-to-model memory for `prompt_cache_key`-only callers hits every automatic-directive and model-less request served in single-model mode — the default entry point for mesh clients, not a legacy corner. Same fix as proposed; the priority is just higher than "auto alias" suggests. The exhausted-hosts path (`forget_auto_model(auto_session_key)`, `transport.rs:742`) keys off the same value, so fixing the derivation heals both remember and forget consistently.

**MoA committee path is unaffected by this PR** (no action needed). When the directive is served as a committee, `RemoteModelBackend::chat_completion` rebuilds worker bodies from scratch (`moa_gateway/workers.rs:324`) — `prompt_cache_key`/`user`/`session_id` are stripped and replica order is the origin-stable `hosts_for_model` order, so committee fan-out never consulted cache-affinity routing before this PR and still doesn't. Streaming and media requests are admitted to single-model mode, so agent harnesses (streaming) take the path this PR changes — consistent with the review above. (Possible future win, out of scope here: worker bodies do preserve tools + system messages, so scaffold-keyed affinity could route committee workers to warm nodes — but that's an enhancement, not a regression.)

crates/mesh-llm-host-runtime/src/network/affinity.rs
i386: Fixed in b7b6f6b07. Auto-model memory now keeps prompt_cache_key as its model-choice key without turning it into target stickiness. The regression asserts the auto-model key is present while sticky_hash remains absent.

crates/mesh-llm-host-runtime/src/network/openai/transport_route_model.rs
i386: Fixed in b7b6f6b07. Both asynchronous cache-selection entry points capture the lease epoch before awaiting and atomically reject a stale write after invalidation. A deterministic regression covers invalidation while selection is in flight.

crates/mesh-llm-host-runtime/src/network/openai/transport_route_model.rs
i386: Fixed in b7b6f6b07. Authoritative zero-cache usage now clears the exact affinity lease for local and remote targets; only local responses mutate local inventory. The regression verifies a remote miss removes its lease while preserving independent local evidence.

michaelneale (approved):

PR #1605, opened 2026-09-02
Move stage-zero exact-state copies off the request path

i386:
## Summary

- enqueue stage-zero exact-state exports as detached FIFO scheduler work, preserving checkpoint ordering without making request workers wait
- retain a contiguous export allocation while indexing it by BLAKE3 blocks, so unfragmented exact-state restores avoid rebuilding a second full-size buffer
- materialize only surviving shared blocks when the last contiguous owner is evicted, preserving deduplication and exact physical-byte accounting
- add randomized ownership/accounting, concurrent restore/eviction, and detached scheduler tests

Stacked on #1603. This is the exact-state hot-path layer from the KV-cache review in Buzz channel `10b70ee2-a3fa-41e2-a87f-b5b53792a7c1`.

## Verification

At `dc13ae5c67332e00513bb42b9c51b6cd1be8a371`:

- `skippy-cache`: 59 passed
- `mesh-llm-host-runtime`: 2,904 passed, 8 ignored
- focused detached-export and deduplicated-eviction tests passed
- strict Clippy clean for `skippy-cache` and dynamic-runtime `skippy-server`
- broad dynamic `skippy-server`: 643 passed, 3 ignored; the same nine tests require a loaded native runtime library

Matched parent/child lab benchmarking is in progress and will be added before this leaves draft review.

## Summary by CodeRabbit

- **Performance**
  - Exact-state checkpoints are now recorded asynchronously, allowing generation and request processing to continue without waiting for export or indexing work.
  - Cache restores can reuse contiguous backing data for zero-copy access, reducing unnecessary memory copying.
  - Shared cached data is materialized only when needed, improving storage efficiency.

- **Reliability**
  - Improved cache storage accounting, cleanup validation, concurrent-read handling, and ownership tracking.

- **Documentation**
  - Updated cache documentation to describe asynchronous checkpoint processing and content-addressed state storage.


i386:
**Review: Core design is sound. One telemetry semantic change needs to be addressed before leaving DRAFT.**

**`resident_recorded_pages` counter changes meaning** — in the old path, `record_exact_state_at_tokens(...)` returned whether a record was actually written, and the counter incremented only on a confirmed write. `enqueue_exact_state_record_at_tokens` always returns `true` on a successful enqueue, even if the detached work ultimately finds nothing to write (missing session, payload not exact-state, etc.). The counter now means "checkpoint enqueued" not "checkpoint actually written." Operators using this to detect recording failures will see false positives. Either rename the counter or emit a separate event from the detached worker when a record is skipped.

**Ordering guarantee is correct** — `execute_runtime_detached` enqueues a `SchedulerCommand::ExecuteRuntime` FIFO before returning, so the checkpoint runs before any subsequent iteration can advance the session. The `detached_runtime_operation_returns_before_work_completes` test validates this via deadlock-as-proof: if the caller had blocked, it would deadlock because the detached work blocks on `release_rx` which isn't sent until after the caller returns.

**Contiguous-storage optimization** fires only when all blocks are new AND all hashes within the payload are unique. A payload with two identical 1MB blocks takes the standalone path. Correct and the edge case is handled.

**`RwLock<CacheBlockBytes>`** — write lock only during materialization (eviction of last contiguous owner); read lock during `as_cow()`. `partial_dedupe_restore_remains_valid_while_contiguous_owner_is_released` exercises the concurrent-read-during-eviction race. Coverage is there.

**Physical byte accounting** — `assert_eq` → `assert!(>=)` in the randomized test correctly reflects that `physical_bytes` now tracks the full contiguous allocation, which may exceed the sum of unique block sizes while contiguous owners remain. The `assert_storage_accounting` helper checks internal consistency.

**`prefix_cache.rs` change** — the inline `record_exact_state` call at `token_count == scheduler_token_ids.len()` is correctly replaced by the detached path at the same token boundary.

Should leave DRAFT once lab parent/child benchmark numbers are in and the `resident_recorded_pages` telemetry point is resolved.

crates/skippy-server/src/frontend/iteration_scheduler.rs
```diff
@@ -745,6 +745,24 @@ impl IterationScheduler {
...
+        operation: impl FnOnce(&mut RuntimeState) + Send + 'static,
+    ) -> OpenAiResult<()> {
+        let (operation, result) = runtime_operation(label, move |runtime| {
+            operation(runtime);
+            Ok(())
+        });
+        self.enqueue_command(SchedulerCommand::ExecuteRuntime(operation))?;
+        drop(result);
+        Ok(())
+    }
```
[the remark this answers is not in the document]
i386: Addressed in 8b91ec31f. Detached runtime operations moved to iteration_scheduler/detached_runtime.rs, scheduler tests to iteration_scheduler/tests.rs, and exact-state checkpoint enqueue/record logic to token_generation/exact_state_recording.rs. The production files are now 1,893 and 1,765 lines. Scheduler tests pass 34/34, token-generation tests pass 7/7, cache tests pass 59/59, and Skippy check and strict Clippy are clean. The full dynamic-runtime library run has 643 passing, 3 ignored, and the same 9 native-runtime-library prerequisite failures.

i386:
@i386 addressed — `resident_recorded_pages` now only counts synchronous writes from the restore path. The async-enqueue path increments a new `resident_enqueued_checkpoints` counter, emitted separately as `skippy.kv.enqueued_checkpoint_pages` in the request telemetry event. `skippy.kv.recorded_pages` retains its original "actually written" semantics.

The detached worker already emits `stage.openai_kv_record_decision` with `{decision_prefix}_recorded` / `{decision_prefix}_skipped` outcomes, so operators have per-checkpoint resolution on the async path without reading the aggregated counter.

Latest commit: `451297f54`

Still in DRAFT pending lab parent/child benchmark numbers per the PR description.

crates/skippy-server/src/frontend/local_generation/token_generation/exact_state_recording.rs
```diff
@@ -0,0 +1,161 @@
...
+                json!(runtime_token_count),
+            );
+            self.telemetry
+                .emit("stage.openai_kv_record_decision", attrs);
+            return false;
+        }
+
+        let base = self.local_kv_message_base(session_id, ids);
+        let identity = kv.prefill_identity(&self.config, &base, 0, checkpoint_tokens);
+        match kv.record_exact_state(runtime, session_id, &identity) {
```
ndizazzo, line 123: This recorder still has a synchronous caller for `chat_prefix_checkpoint` in `restore_or_record_kv_on_runtime`. That caller runs inside the awaited `execute_cache_aware_runtime_timed` operation, so `kv.record_exact_state(...)` here can still keep the request waiting for the full export. We should route that checkpoint through the detached FIFO path too, or explicitly scope the PR's claim and document why this one must stay synchronous.

ndizazzo (changes requested):
The checkpoint uses the same single scheduler command channel, and `ExecuteRuntime` runs as soon as it is dequeued, so it stays ahead of later session work without making the request worker wait. The follow-up changes also separate enqueued checkpoints from confirmed writes and split the scheduler additions out of the oversized file. I couldn't find a concrete PR-caused bug in the Arc pointer bookkeeping or panic handling.

One exact-state export still runs synchronously inside the awaited cache-aware request operation, so the title and summary are still broader than the implementation. We should either move that checkpoint onto the detached FIFO path too, or narrow the claim and explain why it has to stay synchronous.

crates/skippy-server/src/frontend/local_generation/token_generation/exact_state_recording.rs
i386: Addressed in 9e0d09923 using the explicit-scope option. The PR title now says stage-zero exact-state copies, and both the source and cache README document the max_tokens=0 prefill-only exception. That path must export synchronously before the same scheduler operation advances the remaining suffix; a detached command cannot run until the operation releases the scheduler, when the canonical-position guard would reject the earlier checkpoint. Generation requests continue to use the split detached FIFO path. The updated #1603 parent was also merged without conflicts.

ndizazzo (approved):

PR #1606, opened 2026-09-02
Certify KV reuse with captured agentic replay

i386:
## Summary

- replay exact captured Buzz and agent-harness trajectories, including their original tool schemas and supported OpenAI message fields
- add a final-prefix mode for measuring the longest recorded reusable prefix once per trajectory
- fail closed on required harness coverage, real server-reported prompt-token ranges, cached-token share, semantic output identity, and TTFT regression
- retain complete raw requests, reports, commands, hashes, and failed-gate details when acceptance fails
- require at least two worker waves per concurrency cell so concurrent cache behavior is not measured only during tail drain

Stacked on #1578. This is the workload-shaped acceptance layer for the KV-cache review in Buzz channel `10b70ee2-a3fa-41e2-a87f-b5b53792a7c1`; it is separate from the runtime stack in #1601, #1603, and #1605 so the benchmark foundation can merge independently.

## Verification

At `7c52f503b9576eab17b708e17b9e13375903a49a`:

- complete repository script suite: 773 passed, 7 skipped
- focused Agentic Replay suite: 35 passed
- Python bytecode compilation clean
- `git diff --check` clean
- side-effect-free CLI plan smoke passed for captured manifests, final-prefix replay, ABBA ordering, three required harnesses, 18K–22K prompt tokens, cached-token share, output identity, and TTFT gates

The live parent/candidate acceptance artifact still requires a private captured trajectory manifest and the target lab model. This remains draft until that workload is run and attached; the PR does not claim a performance win by itself.

### DIFF
--- a/evals/agentic-replay.py
+++ b/evals/agentic-replay.py
@@ -306,6 +306,32 @@ def verify_dataset(path: Path, expected_sha256: str) -> None:
         )
 
 
+def cohort_metadata(cohorts: dict[str, list[dict[str, Any]]]) -> dict[str, Any]:
+    metadata: dict[str, Any] = {}
+    for name, trajectories in cohorts.items():
+        framework_trajectories: dict[str, int] = {}
+        framework_assistant_turns: dict[str, int] = {}
+        for trajectory in trajectories:
+            framework = trajectory["agent_framework"]
+            framework_trajectories[framework] = (
+                framework_trajectories.get(framework, 0) + 1
+            )
+            framework_assistant_turns[framework] = (
+                framework_assistant_turns.get(framework, 0)
+                + assistant_turn_count(trajectory)
+            )
+        metadata[name] = {
+            "trajectory_count": len(trajectories),
+            "assistant_turns": sum(assistant_turn_count(item) for item in trajectories),
+            "framework_trajectories": framework_trajectories,
+            "framework_assistant_turns": framework_assistant_turns,
+            "session_ids_sha256": hashlib.sha256(
+                "\n".join(item["session_id"] for item in trajectories).encode()
+            ).hexdigest(),
+        }
+    return metadata
+
+
 def build_trajectory_manifest(args: argparse.Namespace, output: Path) -> dict[str, Any]:
     config = load_competitive_config()
     dataset = config["thoughtworks"]["dataset"]
@@ -346,6 +372,7 @@ def build_trajectory_manifest(args: argparse.Namespace, output: Path) -> dict[st
         ) from error
     document = json.loads(manifest.read_text(encoding="utf-8"))
     return {
+        "kind": "thoughtworks",
         "dataset": dataset,
         "dataset_file": str(args.dataset_file),
         "dataset_file_sha256": sha256(args.dataset_file),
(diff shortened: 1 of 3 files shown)


evals/agentic-replay.py
```diff
@@ -1253,6 +1505,81 @@ def pooled_rows(results: Sequence[dict[str, Any]]) -> list[dict[str, Any]]:
...
+    min_cache_pct: Optional[float] = None,
+    require_output_match: bool = False,
+    max_ttft_regression_pct: Optional[float] = None,
+) -> dict[str, Any]:
+    checks: list[dict[str, Any]] = []
+
+    def record(name: str, passed: bool, detail: str) -> None:
+        checks.append({"name": name, "passed": passed, "detail": detail})
+
+    for row in rows:
```
ndizazzo, line 1549: I think failed measured requests need their own check before the optional thresholds. `pooled_rows()` carries `failed_requests`, but this loop never reads it. I reproduced a run with 50 requests per arm and 49 failures in each; the one success met the token, cache, output, and TTFT thresholds, so all eight checks and the overall gate passed. Matching failure IDs is useful for comparability, but it should not certify a mostly failed run.

evals/agentic-replay.py
```diff
@@ -1253,6 +1505,81 @@ def pooled_rows(results: Sequence[dict[str, Any]]) -> list[dict[str, Any]]:
...
+                f"ttft-regression:{cell}",
+                observed is not None and observed <= max_ttft_regression_pct,
+                (
+                    f"observed={observed:.3f}% allowed<={max_ttft_regression_pct:.3f}%"
+                    if observed is not None
+                    else f"observed=unavailable allowed<={max_ttft_regression_pct:.3f}%"
+                ),
+            )
+    return {
+        "passed": all(check["passed"] for check in checks),
```
ndizazzo, line 1603: I think this needs a fail-closed empty case. With no gate flags, `checks` stays empty and `all([])` returns true; `run_command()` then writes a passing gate and exits successfully. I reproduced `evaluate_gates(rows)` returning `{"passed": true, "checks": []}`. Please either require at least one acceptance gate or represent this as not evaluated rather than passed.

ndizazzo (changes requested):
I like the durable replay harness and keeping it on an independent base from the runtime stack. The plan and artifact trail make the eventual cache certification easy to audit. I also checked the replay semantics: `final` sends one longest-prefix request per trajectory, while `all` exercises the recorded turns, and all 35 focused tests pass locally on Python 3.9 and 3.12. The exact normalized output hash is conservative and may reject harmless variation, but it does not create the false positives below.

I reproduced two acceptance-path false positives against the live head. Until the gate rejects failed measured requests and stops treating an empty check set as a pass, this can report successful certification without a qualifying run.

evals/agentic-replay.py
i386: Addressed in de2d13501. Configured acceptance runs now add a mandatory zero-failed-requests check for every measured cell, so a mostly failed run cannot pass on the surviving samples. Added a regression with 49 failed requests and a passing cache threshold. The focused suite passes 37/37.

evals/agentic-replay.py
i386: Addressed in de2d13501. With no optional acceptance gates configured, evaluation now returns evaluated=false and passed=null; the runner preserves the benchmark artifact and does not describe the run as a passing certification. Added direct regression coverage. The focused suite passes 37/37.

Issue #1615, opened 2026-09-02
v0.76.0-rc9 native Windows validation: what the release fixes, what it does not, and two new findings


Virgile-pct:
As promised on #1537, here is the full native Windows sequence re-run on rc9 (Windows 11 26200, RTX 4070 Ti 12 GB, 33.5 GB RAM, CUDA 13 driver, rc8 kept side by side for comparison). Assets checked against their sha256 files.

Fixed in rc9, confirmed on this machine:

- #1512 (LoadLibraryExW error 126): `serve` now reaches `api ready` with no PATH workaround at all. The loader fix is in.
- #1566 (GPU-less nodes advertise 0): with the GPU hidden through `CUDA_VISIBLE_DEVICES=-1`, the model opens on CPU instead of failing with "this process has 0.00 GB".
- #1511 (no Windows runtime in the release manifest): `native-runtimes.json` now lists windows cpu, cuda12, rocm and vulkan artifacts with Skippy ABI 0.1.44.

Still broken, and now explained by #1612: from the official `mesh-llm-v0.76.0-rc9-x86_64-pc-windows-msvc.zip` layout, `runtime list --available` shows only the bundled cpu runtime and `runtime install cuda` still fails with `no compatible native runtime found for Skippy ABI 0.1.44 on windows/x86_64`, identical to rc7 and rc8. The adjacent cpu bundle hides the remote catalog exactly as #1612 describes. Importing the cuda12 tarball with `runtime install --bundle-dir` works as before.

New findings from this run:

- #1613: single-stream decode dropped from 55 to 32 tok/s versus rc8 on the same hardware, with a constant per-stream rate under concurrency (details and the exclusion list there).
- #1614: a stray mmproj file in the same folder as a text-only GGUF is attached to it and kills startup, on rc8 too.

Not yet covered in this pass: split serving from a local GGUF (#1568) and the vision repro for #1551 (rc9 ships #1549 and #1550, so it is finally runnable on the stock host; I will report there). Happy to re-run any of this on request, the scripts are in place.

Virgile-pct:
Follow-up on the vision item: the #1551 signature reproduces on stock rc9 on this machine, single GPU, no split (the original report was 3x P40 in layer split), with `Qwen2.5-VL-7B-Instruct-Q4_K_M` plus its matching mmproj and a 336x336 solid-color PNG.

Fresh server, four identical image requests with `max_tokens: 400`:

```
[...]

Issue #1612, opened 2026-09-02
task: merge bundled and release runtime catalogs during selection


ndizazzo:
## Objective

Make runtime discovery combine local bundle candidates with the release catalog so an adjacent CPU bundle cannot hide downloadable GPU runtimes.

## Details

The default release manifest is currently used only when `bundle_dirs` is empty. Official product layouts commonly supply an adjacent `native-runtimes` directory, so the bundled CPU runtime suppresses the remote catalog. This affects `setup`, `runtime list --available`, `runtime install cuda`, and the one-shot startup installer.

The resolver should merge sources while keeping deterministic precedence and offline fallback behavior.

## Checklist

- [ ] Define and document precedence for explicit manifests, environment manifests, bundled artifacts, cached artifacts, and the default release manifest.
- [ ] Fetch the default release manifest when appropriate even if bundle directories were discovered.
- [ ] Merge compatible bundled and remote manifest entries without duplicate or ambiguous candidates.
- [ ] Preserve bundle or cache fallback when network fetches fail.
- [ ] Ensure an explicit GPU selection reports catalog and compatibility failures instead of silently resolving only against the CPU bundle.
- [ ] Apply the behavior consistently to `setup`, `runtime list --available`, `runtime install`, and startup auto-selection.
- [ ] Add tests for an adjacent Windows CPU bundle plus a remote CUDA candidate.
- [ ] Add tests for offline startup with only a compatible bundled or cached runtime.
- [ ] Add diagnostics that state which catalogs were consulted and why a runtime was selected or rejected.

## Acceptance criteria

- [ ] `mesh-llm runtime list --available` shows remote GPU runtimes when an adjacent CPU bundle exists.
- [ ] `mesh-llm runtime install cuda` can discover and install a compatible release artifact from that layout.
- [ ] Startup can upgrade from the bundled CPU runtime to a compatible GPU runtime without a manual runtime command.
- [ ] Network failure does not prevent a compatible local runtime from loading.
- [ ] Source-priority and fallback behavior are covered by automated tests.

## Context

Parent task: #1610
Related report: #1511

Virgile-pct:
I can take this one. Two data points from the rc9 validation tonight (#1615) that match the description exactly, on the official `mesh-llm-v0.76.0-rc9-x86_64-pc-windows-msvc.zip` layout with its adjacent `native-runtimes/meshllm-native-runtime-windows-x86_64-cpu` bundle:

[...]

PR #1623, opened 2026-09-03
fix(runtime): merge bundled and release runtime catalogs during selection

Virgile-pct:
Refs #1612 (part 1 of 2: precedence and merge; diagnostics follow in a second PR)

An adjacent native runtime bundle used to suppress the default release manifest entirely (`manifest_url` returned `None` as soon as `bundle_dirs` was non-empty). The official Windows product layout ships a cpu runtime next to the binary, so from that layout `runtime list --available` showed only the bundled cpu entry and `runtime install cuda` failed with `no compatible native runtime found for Skippy ABI 0.1.44 on windows/x86_64`, on rc7, rc8 and rc9 (#1511, #1615).

Changes, all in `mesh-llm-runtime-install` plus docs:

- `manifest_url` no longer skips the default release URL when bundle directories exist; explicit path, explicit URL and environment URL keep their precedence
- `load_release_manifest_with_sources` (new) merges the loaded manifest with the discovered bundles and returns a `NativeRuntimeCatalogSources` summary (manifest path or URL consulted, remote error if any, bundle directories) for the diagnostics follow-up
- if the remote fetch fails and bundles were discovered, the bundles carry the load and the failure is recorded instead of aborting; without bundles the failure is still an error
- a loaded manifest's `mesh_version` and `skippy_abi` now describe the release; a bundle's values only stand in when no manifest could be loaded (a stale adjacent bundle used to rewrite both)
- a runtime that is both bundled and published is listed once; the resolver already prefers the bundle directory as the source for that identity
- `install_native_runtime` gates the default catalog on `allow_download`, so offline installs (the startup offline paths, packaged apps that set downloads to false as the SDK doc already tells them) never touch the network
- `docs/design/NATIVE_RUNTIMES.md` documents the precedence and fallback rules; the CLI page notes that `--available` now lists both catalogs

Tests (39 pass in the crate on Windows): the rewritten bundle-only precedence test, merge keeps the release identity, bundle-only load takes it from the bundle, remote failure falls back to bundles (with the URL query redacted in the recorded error), remote failure without bundles stays an error, the #1612 scenario itself (adjacent Windows cpu bundle plus a remote cuda12 candidate against a synthetic Windows host profile with a CUDA 13 driver and SM 89: both candidates evaluated, cuda selected as a download), and the bundled-and-published dedup case resolving to the bundle.

End to end on the machine from #1615, using the official `mesh-llm-v0.76.0-rc9-x86_64-pc-windows-msvc.zip` layout with this host binary dropped in:

```
stock rc9   runtime list --available   1 entry  (bundled cpu)
this branch runtime list --available   cpu, cuda12, rocm, vulkan for windows, all compatible
this branch runtime install cuda       resolves the release artifact: "already installed" from the
                                       rc9 cache, and a real 782 MiB download plus install with an
                                       empty --cache-dir
```

Not in this PR: the "which catalogs were consulted and why" diagnostics and the explicit-selection failure reporting (checklist items 5 and 9), which build on the sources summary added here.

Local gates: `cargo test --locked -p mesh-llm-runtime-install --lib --features skippy-ffi/dynamic-runtime`, `cargo fmt --check`, `cargo clippy --no-deps -p mesh-llm-runtime-install --all-targets --features skippy-ffi/dynamic-runtime -- -D warnings`, `repo-consistency no-console-print`, all clean. CodeRabbit findings will be addressed before human review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Offline installations no longer access the default release catalog when downloads are disabled.
  * Explicit catalog URLs remain available when configured.
  * Bundled runtimes no longer hide downloadable runtimes from the available-runtime list.
  * Remote catalog failures no longer block installations when bundled runtimes are available.
  * Duplicate runtime entries are removed, prioritizing bundled and cached copies.

* **Documentation**
  * Updated CLI documentation to clarify available runtime sources.

### DIFF
--- a/crates/mesh-llm-runtime-install/src/install.rs
+++ b/crates/mesh-llm-runtime-install/src/install.rs
@@ -23,7 +23,10 @@ pub async fn install_native_runtime(
             manifest_path: options.manifest_path.clone(),
             manifest_url: options.manifest_url.clone(),
             bundle_dirs: options.bundle_dirs.clone(),
-            allow_default_manifest_url: true,
+            // Offline installs (`allow_download: false`) must not reach out
+            // for the default catalog either; bundles and the cache are the
+            // only sources then. Explicit manifest URLs are still honoured.
+            allow_default_manifest_url: options.allow_download,
         })
         .await?;
     if manifest.artifacts.is_empty() {
--- a/crates/mesh-llm-runtime-install/src/install.rs
+++ b/crates/mesh-llm-runtime-install/src/install.rs
@@ -14,6 +14,10 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 use tokio::io::AsyncWriteExt;
+/// Installs the native runtime selected by `options`: loads the merged
+/// runtime catalog (release manifest plus bundle directories), resolves the
+/// best candidate for this host, then serves it in place from a bundle, from
+/// the cache, or through a verified download.
 pub async fn install_native_runtime(
     options: NativeRuntimeInstallOptions,
 ) -> Result<NativeRuntimeInstallOutcome> {
--- a/crates/mesh-llm-runtime-install/src/lib.rs
+++ b/crates/mesh-llm-runtime-install/src/lib.rs
@@ -22,7 +22,10 @@ pub use cache::{
     native_runtime_cache, native_runtime_versions_match_current_sdk,
 };
 pub use install::install_native_runtime;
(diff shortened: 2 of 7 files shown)


Virgile-pct:
Marked ready for review: the repository bots skip drafts entirely (no CodeRabbit pass, no review routing), so a draft would not have collected the early direction check we wanted. The scope note in the description stands: this is part 1 of 2, the diagnostics half is ready on a stacked branch and will follow once the direction here is confirmed.

PR #1624, opened 2026-09-03
fix(llama): rebase patch queue onto upstream 0df974d777

i386:
# Rebase llama.cpp patch queue onto upstream `0df974d777`

Automated canary repair (run 33732614416, patch-queue mode). The nightly
llama-upstream canary failed to apply our patch queue onto the new upstream
master: patch `0001-Add-staged-model-graph-and-family-support.patch` hit
`git am --3way` content conflicts in `src/llama-arch.cpp`,
`src/llama-context.cpp`, `src/llama-kv-cells.h`, and
`src/models/qwen3next.cpp`. The queue was rebuilt on a branch based on the
repair target per the `llama-patch-changes` runbook, and the regenerated
series applies clean end to end.

- Old pinned upstream: `cc83d7b482`
- Repair target upstream: `0df974d777` (96 upstream commits, 2026-08-29 → 09-03)
- Queue: 51 patches, none added or dropped; 2 subject-only renames
- Final patched tree: `16e120bbd` (was `f31fec8c7` at the old pin)

## Key upstream changes in the range

**Model loading / hparams**

- `#28159` — `hparams.n_layer_nextn` is now loaded once, centrally in
  `llama_model_base::load_hparams()` (with an `n_layer_nextn < n_layer_all`
  assert), and the per-arch `LLM_KV_NEXTN_PREDICT_LAYERS` loads were removed
  from ~20 model files (deepseek2, qwen3next, bailingmoe, cohere2moe,
  deepseek32/4, dots3note, exaone*, …). This is the change our patch 0001
  collided with.
- `#28173` (load arrays with `n_layer_all`), `#27483` (avoid RAM peaking at
  load), `#28121` (noscan `ssm_a` flags), `#27837` + `#27969`
  (`--tensor-read-lazy` renamed to `--lazy-mode` with `-lzm` shorthand).
- New model: NVIDIA Nemotron-3-Puzzle-75B-A9B (`#25444`); DeepSeek-V4 /
  DSpark vision fixes (`#28154`, `#28294`, `#28133`).

**KV cache / cells**

- `#28011`, `#28040`, `#27991` — reworked the per-sequence position index in
  `llama_kv_cells`: `seq_pos_inc`/`seq_pos_dec` now take the cell index, the
  index maps positions to cells, n-gram history lookup goes through the
  index, and non-contiguous cell restore is optimized. Patch 0001's
  position-continuity check built on the old index shape conflicted here.
- `#28038` (alloc-size guard), `#27967` (Hadamard `k_rot` copy guard during
  context shift), `#28030` (yarn autoscale of `n_ctx_train`).

**Context / graph**

- `#27310` — DFlash: the speculative encoder is fused into the KV-cache
  injection; `llama_context::decode()` gained a `dflash_embd` branch selecting
  `hparams.n_embd_inp_enc()`. Patch 0001 rewrites the `n_embd` computation
  this branch lives in.
- `#27941` + `#28123` + `#28023` — Qwen4Exp: seq_cp / block position keying /
  mtmd input / CUDA abort fixes, recurrent state rollback support, indexer
  head slicing, plus new arch tests. Directly touches the arch our staged
  Qwen4Exp patches (0029–0034) target.
- `#27930` — new `SWIGLU_CLAMP` ggml op; `#28174` — `preserve_reasoning`
  enabled by default in common/server; `#27301` — alloc dependencies in
  `graph_optimize`.

**Backends / platform**

- Metal: Metal 4.0 tensor API on M5+/A19+ (`#27461`), fa-vec tunings for
  M1–M5/A18/A19 chips, autoreleasepool leak fixes + CI check (`#27883`,
  `#27884`), top-k radix (`#28073`), quantized concat (`#28116`),
  low-memory query fix (`#27701`).
- CUDA: sparse-fa for DSV4/GLM (`#27970`), MoE fusion extended to
  speculative decoding and >1 token batches (`#27621`), fused MoE weighted
  expert reduction (`#25950`), XOR-swizzled FA smem tiles (`#25635`).
- ROCm 10.0 bump; SYCL / Hexagon / OpenCL / Vulkan / WebGPU fixes; RPC rdma
  fixes; cpp-httplib vendored at 0.54.0; server accepts `data:` URLs for
  video/audio (`#27700`).

## Patch-queue evolution

The series was regenerated from a rebuilt llama.cpp branch on top of
`0df974d777` (capability-owned commits reconstructed, final tree verified,
then `git format-patch`). Per-patch outcome:

- **0001 — staged model graph and family support** (the patch that broke the
  canary; 3 real resolutions):
  - `src/llama-context.cpp` — upstream `#27310` inserted the DFlash
    `dflash_embd ? hparams.n_embd_inp_enc()` arm into the `n_embd`
    expression this patch rewrites. Resolution keeps the new DFlash arm and
    still drops the `mtp_embd` special case, as before.
  - `src/llama-kv-cells.h` — the patch previously threaded a
    `seq_pos_index_mismatch[LLAMA_MAX_SEQ]` counter through
    `seq_pos_inc`/`seq_pos_dec`/`reset` (extending their signatures with the
    cell index) to make `seq_pos_is_cell_contiguous()` O(1). Upstream
    `#28011`/`#28040` changed those helpers to take the cell index and
    reworked the index, so the counter bookkeeping no longer applied.
    Resolution: the counter is gone entirely and
    `seq_pos_is_cell_contiguous()` is re-expressed as a single ordered scan
    over the `(pos → cell)` index entries verifying dense positions and
    `pos == cell` alignment. Same contract, validated directly against the
    new index shape.
  - `src/models/deepseek2.cpp`, `src/models/qwen3next.cpp` — the patch used
    to delete the per-arch `NextN/MTP` `NEXTN_PREDICT_LAYERS` loads and
    asserts (the same duplication upstream later removed itself in
    `#28159`). Those hunks became no-ops and were dropped; the staged family
    path now relies on upstream's central load plus the patch's own
    `deepseek2` load/assert (kept).
- **0012 — fixup: staged runtime wiring onto per-op Metal layout** — only
  `tests/test-llama-archs.cpp` changed (125 → 20 lines). Kept: the
  `glm_dsa_backend_config_supported()` skip guard and the WebGPU `FIXME`
  skip in `test_backends`. Dropped: the GLM-DSA index-share fixture variants
  (`glm_dsa_indexshare_fixture`), the `scoped_test_env` helpers, and the
  extended `get_model_and_ctx` signature (cb_eval / flash_attn) — upstream
  `#27941` rebuilt that fixture area and the old plumbing no longer applies.
  See risks below.
- **0041 — physical KV-cache occupancy** — no longer carries the mid-series
  `SKIPPY_ABI_VERSION_PATCH` bump; the bump chain is now spread across the
  ABI-owning patches (each increments by one, 35 → 48). End state unchanged.
- **0039, 0051 — renamed only** (subject now matches content:
  "add kv_offload, kv_unified, swa_full, op_offload…" and "expose the
  native session sequence ID"). No content change.
- **All other patches (0002–0007, 0010, 0011, 0013–0026, 0028–0038,
  0040, 0042–0050)** — rebase churn only: refreshed blob indices, hunk
  offsets, context lines where upstream edited nearby code (e.g. 0019's
  Metal hunks now sit against the `MTLLanguageVersion4_0_GGML` context from
  `#27461`; 0028's `ctx_other` hunks follow the constructor's
  `cparams.rope_scaling_type` refactor from `#28030`), plus the regenerated
  `[PATCH n/51]` subjects and git trailers. Patch intent is unchanged.

Net effect measured on the final trees (`f31fec8c7` vs `16e120bbd`): the
only files where the patched tree moves differently from pure upstream are
`src/llama-kv-cells.h`, `src/models/deepseek2.cpp`,
`src/models/qwen3next.cpp`, and `tests/test-llama-archs.cpp` — exactly the
four resolutions above. **`include/skippy/**`, `include/skippy.h`, and
`src/skippy/**` are byte-identical** between the old and new final trees.

## ABI impact

- None at the boundary: the final `SKIPPY_ABI_VERSION` is `0.1.48` before
  and after this repair; all public `include/skippy/*.h` headers and
  `src/skippy/**` implementation files are byte-identical to the old
  certified tree.
- The Rust mirror `crates/skippy-ffi` (`ABI_VERSION` 0/1/48) matches the
  patched tree. No crate in `crates/` is modified by this branch.
- The only distribution change is internal to the series: the ABI patch
  bump moved out of 0041 into the incremental chain carried by the
  ABI-owning patches (0003 → 0051), which is where the skill wants it.

## Risks for reviewers

1. **Possible dropped intent in 0012's test fixture work.** The GLM-DSA
   index-share fixture variants, scoped test-env helpers, and
   `get_model_and_ctx` extensions were removed because upstream's rewritten
   fixture no longer admits them. Upstream did *not* absorb the GLM-DSA
   variants (it added similar `indexer_types` fixtures only for DOTS3NOTE).
   Certification parity can't see this; if those fixtures guarded a staged
   GLM-DSA correctness case, it is no longer exercised by
   `test-llama-archs`. The skip guards that keep CI green remain. Reviewer
   judgment requested on whether the dropped fixtures need re-derivation on
   the new fixture shape.
2. **KV-cells contiguity check cost.** `seq_pos_is_cell_contiguous()` went
   from O(1) (counter) to an ordered scan over the sequence's position index
   on the staged graph-build path. Correct against the new index, but hot
   staged decode on long contexts pays O(positions) per check. If that shows
   up in staged TTFT profiles, a mismatch counter can be reintroduced on top
   of the new index shape in a follow-up.
3. **Duplicate `NEXTN_PREDICT_LAYERS` load for deepseek2.** Upstream now
   loads it centrally (`llama-model.cpp`) and our 0001 re-adds an arch-level
   load. Both write the same optional key to the same field, so behavior is
   unchanged, but the arch-level load in 0001 is now redundant and should be
   dropped in a later hygiene pass.
4. **Upstream behavior changes now riding along** (inherent to the pin
   bump, called out because staged paths inherit them): `preserve_reasoning`
   defaults on (`#28174`), DFlash decode selects `n_embd_inp_enc()` when
   embd batches are present (`#27310`), Qwen4Exp position-keying/seq_cp and
   recurrent-rollback fixes (`#27941`, `#28123`) interact with our staged
   Qwen4Exp patches, and Metal 4.0 tensor API activates on M5+/A19+
   (`#27461`) beneath our Metal pipeline-cache patch (0019).
5. **No lanes weakened.** This branch touches only the 51 patch files — no
   changes to `ci/llama-canary/*`, `scripts/skippy-family-battery.sh`, or
   any manifest; no lane is skipped or excluded.

## Validation

- `scripts/prepare-llama.sh 0df974d777` applies the full queue clean
  (recorded: upstream sha, patched sha `16e120bbd`, patch digest).
- Native build of the rebased queue completed on the family-certify runner
  (Release, Metal backend, arm64; build stamp matches the final patched
  sha).
- The wrapper re-runs the full family battery (`--skip-build`) after this
  description is drafted and posts the certified result as a PR comment;
  families most exposed to this range are `qwen3-next` (Qwen4Exp fixes),
  `nemotron` (MTP/hybrid changes), `deepseek2` (MLA/NextN load move), and
  the dense/spec lanes generally.
- Per the repair loop, a separate fresh-context review turn re-checks this
  certified repair for dropped intent/rebase leftovers after a green
  battery; any fixes land as their own `review(llama):` commit and merged
  main is re-certified by the next canary run before the pin advances.

## Summary by CodeRabbit

* **New Features**
  * Added support for Inkling text, image, and audio models.
  * Added staged model execution for processing large models in layer ranges.
  * Added model loading from multiple parts and tools for creating sliced or merged model packages.
  * Added session state export/import, KV-cache management, resident prefixes, and memory usage reporting.
  * Added activation-frame execution, batched requests, speculative decoding, MTP drafts, and n-gram drafting.
  * Expanded sampling options, chat templates, grammar, JSON schema, and token signal reporting.

* **Bug Fixes**
  * Improved input validation, error details, runtime option handling, and sampling reliability.

### DIFF
--- a/third_party/llama.cpp/patches/0001-Add-staged-model-graph-and-family-support.patch
+++ b/third_party/llama.cpp/patches/0001-Add-staged-model-graph-and-family-support.patch
@@ -1,7 +1,7 @@
-From d76bf95a279692cf6512010d499100b7df73ff5c Mon Sep 17 00:00:00 2001
+From f69831197e469d2c76ef34b39d7617885e626fab Mon Sep 17 00:00:00 2001
 From: Mesh-LLM CI <ci@mesh-llm.local>
 Date: Sat, 8 Aug 2026 16:02:19 +1000
-Subject: [PATCH 01/32] Add staged model graph and family support
+Subject: [PATCH 01/51] Add staged model graph and family support
 
 ---
  include/llama.h                |   7 +
@@ -15,7 +15,7 @@ Subject: [PATCH 01/32] Add staged model graph and family support
  src/llama-hparams.h            |  30 +-
  src/llama-kv-cache.cpp         | 966 ++++++++++++++++++++++++++++++++-
  src/llama-kv-cache.h           |  58 +-
- src/llama-kv-cells.h           |  80 ++-
+ src/llama-kv-cells.h           |  59 +-
  src/llama-memory-recurrent.cpp |   9 +-
  src/llama-model-loader.cpp     | 182 ++++++-
  src/llama-model-loader.h       |  16 +
@@ -42,7 +42,7 @@ Subject: [PATCH 01/32] Add staged model graph and family support
  src/models/command-r.cpp       |  20 +-
  src/models/dbrx.cpp            |  18 +-
  src/models/deci.cpp            |  18 +-
- src/models/deepseek2.cpp       | 647 +++++++++++-----------
+ src/models/deepseek2.cpp       | 642 +++++++++++-----------
  src/models/deepseek32.cpp      |  16 +-
  src/models/dots1.cpp           |  18 +-
  src/models/dream.cpp           |  18 +-
@@ -107,7 +107,7 @@ Subject: [PATCH 01/32] Add staged model graph and family support
  src/models/qwen35.cpp          |  29 +-
  src/models/qwen35moe.cpp       |  27 +-
  src/models/qwen3moe.cpp        |  18 +-
- src/models/qwen3next.cpp       | 351 +++---------
+ src/models/qwen3next.cpp       | 345 +++---------
  src/models/qwen3vl.cpp         |  25 +-
  src/models/qwen3vlmoe.cpp      |  22 +-
  src/models/refact.cpp          |  18 +-
@@ -123,7 +123,7 @@ Subject: [PATCH 01/32] Add staged model graph and family support
  src/models/starcoder2.cpp      |  20 +-
  src/models/step35.cpp          |  32 +-
  src/models/xverse.cpp          |  18 +-
- 119 files changed, 4268 insertions(+), 1545 deletions(-)
+ 119 files changed, 4253 insertions(+), 1528 deletions(-)
 
 diff --git a/include/llama.h b/include/llama.h
 index ef7a012c4..26096355b 100644
@@ -144,7 +144,7 @@ index ef7a012c4..26096355b 100644
              const struct llama_model * model,
                          const char * path_model);
 diff --git a/src/llama-arch.cpp b/src/llama-arch.cpp
-index 5e61f61f7..37b2d3147 100644
+index 446de4ae2..7e83bb1d2 100644
 --- a/src/llama-arch.cpp
 +++ b/src/llama-arch.cpp
 @@ -72,7 +72,6 @@ static const std::map<llm_arch, const char *> LLM_ARCH_NAMES = {
@@ -203,7 +203,7 @@ index 5e61f61f7..37b2d3147 100644
      // sentence-transformers dense modules feature dims
      { LLM_KV_DENSE_2_FEAT_IN,        "%s.dense_2_feat_in"  },
      { LLM_KV_DENSE_2_FEAT_OUT,       "%s.dense_2_feat_out" },
-@@ -659,9 +668,19 @@ static const std::map<llm_tensor, const char *> LLM_TENSOR_NAMES = {
+@@ -660,9 +669,19 @@ static const std::map<llm_tensor, const char *> LLM_TENSOR_NAMES = {
      { LLM_TENSOR_SHORTCONV_CONV,                         "blk.%d.shortconv.conv" },
      { LLM_TENSOR_SHORTCONV_INPROJ,                       "blk.%d.shortconv.in_proj" },
      { LLM_TENSOR_SHORTCONV_OUTPROJ,                      "blk.%d.shortconv.out_proj" },
@@ -223,7 +223,7 @@ index 5e61f61f7..37b2d3147 100644
      { LLM_TENSOR_VISEXP_ATTN_QKV,                        "blk.%d.vis_attn_qkv" },
      { LLM_TENSOR_VISEXP_ATTN_OUT,                        "blk.%d.vis_attn_output" },
      { LLM_TENSOR_VISEXP_FFN_GATE,                        "blk.%d.vis_gate" },
-@@ -893,6 +912,9 @@ static const std::map<llm_tensor, llm_tensor_info> LLM_TENSOR_INFOS = {
+@@ -894,6 +913,9 @@ static const std::map<llm_tensor, llm_tensor_info> LLM_TENSOR_INFOS = {
      {LLM_TENSOR_FFN_UP_EXPS,                {LLM_TENSOR_LAYER_REPEATING, GGML_OP_MUL_MAT_ID}},
      {LLM_TENSOR_FFN_GATE_UP_EXPS,           {LLM_TENSOR_LAYER_REPEATING, GGML_OP_MUL_MAT_ID}},
      {LLM_TENSOR_FFN_DOWN_CHEXPS,            {LLM_TENSOR_LAYER_REPEATING, GGML_OP_MUL_MAT_ID}},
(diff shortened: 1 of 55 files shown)


i386:
Blocking finding: the repaired queue is never selected by the repository.

Both `third_party/llama.cpp/upstream.txt` and `LLAMA_CPP_SHA` still contain the old pin, `cc83d7b4824f73cfdda4dfbb47ee39804f71b328`, even though the PR title and description say the queue was rebuilt for `0df974d777`.

Reproduction at head `cb059cbf3914e615783c3453858c1b0b2cf9dec0`: a clean `LLAMA_WORKDIR=<tmp> scripts/prepare-llama.sh pinned` checks out `cc83d7b482` and patch 0001 conflicts in `src/llama-arch.cpp`, `src/llama-context.cpp`, and `src/models/mimo2.cpp`. This is also the root cause in the failed Static ABI and Native Linux runtime jobs. Running the same queue explicitly against `0df974d777` applies all 57 patches and produces `e0482cb8d`; a local static CPU native build then completes.

Please update both pin files to `0df974d777...` so normal builds and CI consume the repaired queue. The PR description should also be refreshed where it still says 51 patches / final tree `16e120bbd`, because the current head carries 57 patches and produces `e0482cb8d`.

PR #1625, opened 2026-09-03
fix(system): report the real macOS GPU name (not the CPU brand string)

StevenMih:
On macOS, `gpu_name` currently returns `sysctl -n machdep.cpu.brand_string` — the **CPU** brand string mislabeled as the GPU name (verified live on `main`). This reads the real GPU identifier instead, or reports it as absent rather than substituting the CPU string.

3 files (+88/-22): the fix in `crates/mesh-llm-system/src/hardware/{mod,parsers}.rs` plus tests. A mutant test flips if the brand-string fallback is reintroduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * macOS GPU names are now detected directly from Metal for more accurate hardware identification.
  * CPU brand strings are no longer displayed as GPU names.
  * Empty or unavailable GPU names are handled cleanly.
  * Hardware names use best-effort, OS-reported values.

* **Tests**
  * Added coverage for Apple, AMD, and Intel GPU names, including invalid, CPU-derived, and missing-name scenarios.

### DIFF
--- a/crates/mesh-llm-system/src/hardware/mod.rs
+++ b/crates/mesh-llm-system/src/hardware/mod.rs
@@ -303,6 +303,54 @@ fn query_metal_recommended_working_set_bytes() -> Option<u64> {
     }
 }
 
+/// Queries the real GPU identifier via `MTLDevice.name` (e.g. "Apple M4 Max"
+/// or "AMD Radeon Pro 5500M") — never the CPU brand string.
+#[cfg(target_os = "macos")]
+fn query_metal_device_name() -> Option<String> {
+    use std::ffi::{CStr, c_char, c_void};
+
+    // `objc_msgSend` is declared to return `usize` here (matching the other
+    // FFI declaration of the same linked symbol above) and the pointer
+    // results below are recovered with `as *mut/*const _` casts, to avoid a
+    // `clashing_extern_declarations` warning from two conflicting return
+    // types for one symbol.
+    #[link(name = "objc")]
+    unsafe extern "C" {
+        fn sel_registerName(name: *const c_char) -> *mut c_void;
+        fn objc_msgSend(receiver: *mut c_void, selector: *mut c_void, ...) -> usize;
+    }
+
+    unsafe {
+        let metal =
+            libloading::Library::new("/System/Library/Frameworks/Metal.framework/Versions/A/Metal")
+                .ok()?;
+        let create_device = metal
+            .get::<unsafe extern "C" fn() -> *mut c_void>(b"MTLCreateSystemDefaultDevice")
+            .ok()?;
+        let device = create_device();
+        if device.is_null() {
+            return None;
+        }
+        let name_sel = sel_registerName(c"name".as_ptr());
+        if name_sel.is_null() {
+            return None;
+        }
+        let name_obj = objc_msgSend(device, name_sel) as *mut c_void;
+        if name_obj.is_null() {
+            return None;
+        }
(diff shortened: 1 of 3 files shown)


i386 (dismissed):
Requesting changes at `84770e4c8fd08b7819ccc2eb1c029ecae2604909`.

### Blocking

- **[Major] Release the retained Metal device on every exit path.** Apple declares `MTLCreateSystemDefaultDevice` as `NS_RETURNS_RETAINED`, but `query_metal_device_name` never sends `release`, including its early returns. `query_metal_recommended_working_set_bytes` has the same ownership bug, so the normal name + VRAM survey retains two device references per collection. Add a small RAII owner for the returned device and use it in both functions so every path releases safely. This is the unresolved inline finding: https://github.com/Mesh-LLM/mesh-llm/pull/1625#discussion_r3926406250

### Verification at this exact head

- `just with-lld cargo test --locked -p mesh-llm-system`: 189 passed, 0 failed.
- `just with-lld cargo clippy --locked -p mesh-llm-system --all-targets -- -D warnings`: passed.
- `git diff --check` from merge base `5e9027100`: passed.

The functional direction and sanitizer look sound, but the retained-device ownership must be fixed before merge.

ndizazzo (changes requested):
@StevenMih This was actually intentional... we wanted a friendly name instead of specifically named GPU for Macs, since the chip was an SoC. 

1) I'm open to reconsidering this, if you can provide some samples of how it will look around the console
2) I want to make sure we don't break other SoCs like the `Jetson Orin AGX` or `DGX Spark`

If those two aren't really satisfactory - then I would say we should skip merging this.

StevenMih:
@ndizazzo fair — samples as requested, had to dig the out of the drawers :)

M4 Max (Apple silicon):  
brand_string = Apple M4 Max                              
· Chipset = Apple M4 Max

M3 (Apple silicon):      
brand_string = Apple M3                                  
· Chipset = Apple M3

Intel MBP (2015):        
brand_string = Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz 
· Chipset = AMD Radeon R9 M370X, Intel Iris Pro

On Apple silicon the string is identical either way — the console shows exactly what it shows today. The only Macs that change are Intel Macs, where the brand string is the CPU (Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz) and Metal reports the actual GPU (AMD Radeon Pro 5500M). Jetson Orin AGX / DGX Spark are on the Linux path; sanitize_macos_gpu_name never runs there.

What we actually need is narrower than the display and I'd rather not argue the string: a small gpu_name_source enum beside gpu_name (metal_device / cpu_brand_string / sysfs / …) so a consumer can tell how the value was captured, display untouched. Your call on this one: (a) merge as-is given the samples, (b) drop the CPU-brand rejection and keep only the Metal query, or (c) close it and take the source-enum PR instead. I'll open the source-enum PR either way — it's independent.

StevenMih:
Thanks for the review — resolved in two parts. This PR is now the minimal, easy-to-read version: the Metal device leak is fixed (the retained default device is released after reading its name), and the docs state plainly that `gpu_name` reports what the OS returned at collection time — provenance, not a verified GPU identifier. The deeper fix ships separately: a `gpu_name_source` provenance enum alongside `gpu_name` at every collection site (variant `metal_default_device` for this path — named that way deliberately, since on switchable-graphics Macs the default device is a moment-in-time fact; a follow-up will list all adapters rather than just the default). Happy to fold that in here instead if you'd prefer one PR.

ndizazzo (approved):

PR #1631, opened 2026-09-03
Spread concurrent new sessions with local route reservations

### LINKED ISSUE #1629: Spread concurrent new-session requests with local routing reservations
## Summary

Spread bursts of concurrent, new-session inference requests across equivalent targets by keeping short-lived, process-local reservations for requests that have been selected but are not yet reflected in routing observations.

Today Mesh has strong hard gates and ranking signals—model/capability eligibility, context fit, admission state, target health/reputation, measured throughput, and session/prefix affinity. A remaining timing gap is that several requests arriving together can all see the same pre-dispatch state and choose the same leading replica before attempt outcomes or peer observations catch up.

## Proposed first slice

- Maintain a bounded local reservation count keyed by model and inference target.
- After hard eligibility, context, admission, and health filtering, use reservations as a tie-breaker among otherwise suitable replicas for **new or unaffined** requests.
- Reserve the selected target atomically before beginning the connection/dispatch attempt.
- Release the reservation on completion, rejection, cancellation, transport failure, or when failover commits the request to another target.
- Add expiry/cleanup so a panic or abandoned request cannot leave a permanent phantom load.
- Keep this entirely process-local in the first iteration: no gossip field, protocol change, or cluster-wide scheduler.

Likely ownership is the existing OpenAI routing/affinity path, particularly:

- `crates/mesh-llm-host-runtime/src/network/openai/transport_route_model.rs`
- `crates/mesh-llm-host-runtime/src/network/affinity.rs`
- `crates/mesh-llm-host-runtime/src/network/openai/routing_rank.rs`

## Constraints

- Do not override model/capability eligibility, context capacity, admission state, or health cooldowns.
- Preserve explicit target choice and established session/prefix affinity; cache locality should beat burst spreading for an already-sticky conversation.
- Do not expose prompt content, endpoint URLs, raw peer IDs, or high-cardinality request data in telemetry.
- Preserve mixed-version behavior by avoiding a wire change in the initial implementation.
- Keep selection deterministic under equal inputs so tests and operator diagnosis remain straightforward.

## Acceptance criteria

- A deterministic concurrent test shows new-session bursts distributed across two and three equivalent targets instead of piling onto the initial target.
- A single eligible target behaves exactly as it does today.
- Established affinity remains sticky during concurrent requests.
- Failed first attempts release or transfer their reservations correctly during failover.
- Cancellation and timeout paths do not leak reservations.
- Bounded aggregate status/metrics make reservation use and cleanup observable without exporting sensitive identifiers.

## Follow-up, only if measurements justify it

If local reservations do not handle bursts originating from several Mesh nodes, separately design a privacy-bounded, additive per-model pressure bucket for gossip. That should not be part of this first issue.

michaelneale:
## Summary

- concurrent new sessions spread across equivalent inference targets instead of all landing on the same one
- cache, session, sticky-prefix, and explicit-target affinity stay authoritative — an affinitized request is never spread away from its warm cache or pinned target
- the reservation follows the request: it moves with failover to another target and is released automatically on success, failure, cancellation, or panic

## Architecture and compatibility

The whole mechanism is one process-local map of in-flight counts per `(model, target)`, guarded by a mutex, plus an RAII guard per request. Selection and reservation happen in one critical section: among eligible targets with the fewest local in-flight requests, the existing picker result stays the deterministic tie-breaker. Because a count only exists while its guard is alive, the map is bounded by in-flight requests by construction — there is no TTL, eviction, or background maintenance. Both direct model routing and passive mesh routing share the same state, and `/api/status` reports a single `reservation_active` gauge.

This changes no protocol, gossip, request, or response fields, so mixed-version meshes retain their existing routing compatibility. Hard eligibility filters run before reservations, and explicit target routing remains untouched.

NVIDIA PAIR uses a related optimistic-reservation signal alongside scheduler pending/GPU pressure, bypassing it for manual pins and resetting it on a fresh scheduler snapshot. This patch keeps that local-only principle, but ties each reservation to the actual request lifecycle with transfer-on-failover and RAII release instead of waiting for a later snapshot.

## Verification

- `cargo test -p mesh-llm-host-runtime --lib` — 2,949 passed, 0 failed, 9 ignored
- strict `cargo clippy --all-targets -- -D warnings` for `mesh-llm-host-runtime` and `mesh-llm`
- `cargo fmt --all --check`

Tests cover deterministic two- and three-target concurrent bursts, unchanged single-target routing, established affinity under reservation pressure, per-model isolation, failover transfer with drop release, and a real-proxy overlap test that holds one target busy and asserts the next request lands on the other.

Fixes #1629

## Summary by CodeRabbit

- **New Features**
  - Improved request routing to distribute concurrent sessions across equivalent inference targets.
  - Preserved cache, session, and explicit affinity preferences when selecting targets.
  - Added automatic route reservations that follow requests during retries and failover.
  - Improved target ranking by considering context suitability and throughput equivalence.

- **Bug Fixes**
  - Improved failover behavior by redirecting requests away from targets under active pressure.
  - Ensured routing reservations are released after requests and retries complete.

- **Tests**
  - Added coverage for concurrent load distribution, affinity precedence, failover, and reservation cleanup.


michaelneale:
## Quality/safety review — change requested

I found one routing-safety issue that should be addressed before merge.

`RoutingReservations::reserve` minimizes the reservation count across **all** eligible candidates, regardless of the context/throughput rank that produced their order. Therefore, as soon as the preferred target has one in-flight request and a materially slower target has zero, the next unaffined request is redirected to the slower target. This is not limited to “equivalent targets” or a tie-break within the existing rank: it overrides the throughput preference established by `reorder_candidates_by_context_and_throughput`.

That can regress heterogeneous meshes (for example, one high-throughput GPU and one much slower host): ordinary concurrent traffic is forced roughly evenly across them even when measurements deliberately rank the fast host first. Please preserve ranking tiers/scores when applying reservation pressure, or explicitly combine capacity/throughput with the in-flight count. Add a test with unequal throughput ranks showing that reservations do not cause a lower-ranked target to displace a materially preferred target merely because its local count is zero.

Other safety observations from current head `24236e3915bb60c9b2ed35701e7892dded644b80`:

- Affinity paths remain authoritative.
- A one-target/solo route is placement-neutral.
- Retries transfer the guard; RAII releases it on return, cancellation, and unwind.
- MoA interception and worker-pool code are unchanged and run before this ordinary routing path.
- No protocol/gossip changes were introduced, so mixed-version wire behavior is unaffected.
- GitHub CI is green, including the relevant Rust crate tests. My local full relevant-crate command could not start because this checkout's Rust 1.91.1 is below `serial_test`'s required 1.93.1.

I could not submit a formal “request changes” review because the authenticated GitHub account is the PR author, so this comment records the blocking review finding.

michaelneale:
🤖 Addressed in f18a83894.

Reservation spreading is now confined to the leading run of throughput-equivalent candidates instead of minimizing across all eligible targets:

- `rank_candidates_by_context_and_throughput` now returns `RankedCandidates { ordered, equivalent_prefix }`, where `equivalent_prefix` is the length of the leading run whose throughput rank key ties with the best candidate. It never crosses the adequate-context/unknown-context boundary.
- `RoutingReservations::reserve` takes a `spread_limit` and only trades off `candidates[..spread_limit]` by local in-flight count. A preferred target outside the window (e.g. a sticky pick) is reserved as-is, and affinity remains authoritative.
- Both call sites thread the tier through: the local route path recomputes the limit after health filtering (order-preserving), and the mesh plan carries `equivalent_hosts`.

New regression tests:
- `reservation_pressure_never_displaces_a_higher_ranked_target` — unequal ranks: 4 concurrent sessions all stay on the faster target despite the slow one having zero reservations.
- `reservation_pressure_spreads_only_within_the_equivalent_tier`, `preferred_target_outside_the_spread_window_is_reserved_as_is`
- `mesh_plan_pressure_never_promotes_a_lower_throughput_ranked_host` (mesh plan level)
- `test_equivalent_prefix_*` covering tier computation, including unmeasured candidates (all equivalent) and the context-tier boundary.

Homogeneous meshes (no measured throughput differences) keep the full spreading behavior this PR was written for, since all candidates share one rank key.

Validation: `cargo test -p mesh-llm-host-runtime --lib` (2962 passed), `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`, `cargo clippy -p mesh-llm --all-targets -- -D warnings`, `cargo fmt --all --check` — all clean.

crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/self_fill.rs
```diff
@@ -0,0 +1,119 @@
...
+    name: &str,
+    required_tokens: Option<u32>,
+    mut candidates: Vec<InferenceTarget>,
+    affinity: Option<&AffinityRouter>,
+) -> Vec<(InferenceTarget, Option<RoutingReservation>)> {
+    let mut selected = Vec::with_capacity(SELF_FILL_TARGET_WORKERS);
+    while selected.len() < SELF_FILL_TARGET_WORKERS {
+        // Re-rank the remaining endpoints so slot two can use the next tier
+        // only when slot one exhausted the best tier, never due to load alone.
+        let ranked = rank_targets_by_context(node, name, required_tokens, &candidates).await;
```
ndizazzo: [P2] We should preserve health tiers before ranking and spreading clones. `hosts_for_model()` puts healthy peers ahead of `AcceptingDeprioritized` peers, but this throughput sort discards that priority. With two healthy replicas advertising 100 tok/s and a deprioritized replica advertising 200 tok/s, self-fill selects the deprioritized replica for the first slot. The previous implementation selected both healthy replicas. I reproduced this with a focused test. This sends work back to pressured capacity despite healthy alternatives, so let's rank and reserve within the healthy tier first and keep deprioritized peers as spillover.

ndizazzo (changes requested):
I like tying reservations to the request lifecycle. We should preserve health priority in the new self-fill selection before merging, though. I reproduced a case where it selects a deprioritized peer while two healthy replicas are available.

The added regression test fails on this head. Existing networking, proxy, and shared-routing tests passed: 617 passed, 1 ignored. Linux CI was still running when I finished validation.

michaelneale:
@ndizazzo ready for another look

ndizazzo (approved):
LGTM now!

PR #1635, opened 2026-09-04
fix(split): don't build the native MTP graph on a split stage

michaelneale:
27B split serving SIGSEGVs on the final stage during model load.

`qwen35.cpp:541` falls back to `model.tok_embd` when the nextn block has no `embed_tokens`, with no null check. Stage artifacts only include the embeddings when `layer_start == 0`, so a stage starting past layer 0 passes NULL to `ggml_get_rows`, which faults reading `a->ne[2]` — matching the crash report address `0x20`.

Verified against the real GGUF header of `Qwen3.8-27B-UD-Q4_K_M`: `block_count=65`, `nextn_predict_layers=1`, nextn block ships `eh_proj/enorm/hnorm/shared_head_norm` and no `embed_tokens` — so the fallback is always taken. In a two-stage split no stage satisfies both conditions (stage 0 has embeddings but no nextn block; final stage has nextn but doesn't start at 0).

Fix gates native MTP on a stage owning the whole layer range. Single-node serving unchanged.

## Validation
- 2917 host-runtime lib tests pass; clippy `-D warnings` clean on `mesh-llm-host-runtime` and `mesh-llm`; `cargo fmt --all --check` clean.
- Regression test confirmed to FAIL without the fix (first version of it did not — it asserted nothing because the fixture resolver had MTP off; rewritten to force the resolver to `true`).
- On hardware: rebuilt and redeployed M5+M4, reran the exact command that crashed twice. M4 no longer crashes; no new crash report.

## Not proved
No end-to-end 27B split token yet — both post-fix attempts hit a separate peer stage-generation eligibility blocker. That is a distinct bug, still being chased. The crash fix stands on the load-survival evidence above.

Same unguarded fallback pattern exists in 9 other archs (deepseek2/4, glm4-moe, hy-v3, cohere2moe...) — untested here, not changed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of native MTP in staged model execution.
  - Native MTP is now used only when a single stage owns the complete layer range.
  - Multi-stage configurations consistently disable native MTP, including when it is globally enabled.
  - Prevents invalid graph construction for split-stage deployments.
  - Added coverage for both single-stage and multi-stage configurations.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/runtime/local_split/loading.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/local_split/loading.rs
@@ -627,6 +627,23 @@ pub(super) fn stage_source_prepare_timeout(
         .max(MIN_STAGE_SOURCE_PREPARE_TIMEOUT))
 }
 
+/// Whether a stage holds every tensor the native MTP draft graph reads.
+///
+/// The MTP block is the trailing `nextn` block, so it materializes only on the
+/// final stage (`downstream` is `None`). Its graph also reads the token
+/// embedding matrix, and `include_embeddings` is `layer_start == 0`
+/// (`mesh/stage_artifacts.rs`, `mesh/stage_transport.rs`, and the artifact
+/// selection below). When the `nextn` block carries no `embed_tokens` of its
+/// own, the arch graph falls back to `model.tok_embd`; on a stage that starts
+/// past layer 0 that tensor is absent and the fallback is unguarded, so the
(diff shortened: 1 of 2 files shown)


michaelneale:
Premise falsified: a published-layer-package 27B split ran with native MTP active (9/13 accepted), so gating MTP on stage shape would regress a working case. Closing rather than reworking here.

Issue #1576, opened 2026-09-01
Skippy L3 disk tier: user controls, configuration, observability, and acceptance gates


i386:
# Skippy L3 Disk Tier User Controls and Configuration Specification

## 1. Status and decision

This specification defines the user-visible and operator-visible contract for
Skippy's node-local L3 disk cache. It is a landing requirement for the L3 tier
and a prerequisite for production prefill/decode disaggregation.

The first released version is:

- **off by default**;
- configured once per node, not once per model or stage;
- bounded by a hard physical byte budget and a minimum-free-space reserve;
- safe to miss, decline, prune, or discard at any time;
- observable without enabling debug logs or telemetry export;
- independent from peer serving: enabling local disk never exposes cache data
  to the mesh.

The cache is an optimisation. It is never authoritative state. Any ambiguity,
corruption, incompatible identity, storage error, or unsupported model layout
must produce a miss or a clearly reported disabled/degraded state, never a
best-effort restore.

## 2. Goals

1. Let an ordinary user enable, size, locate, inspect, prune, and clear the L3
   cache without learning internal Skippy environment variables.
2. Keep a stable agent prefix reusable across sessions, lanes, model reloads,
   and process restarts for both solo and split serving.
3. Prevent unbounded disk use, SSD write thrash, duplicated concurrent loads,
   and invisible enabled-but-unused states.
4. Make the same content-addressed manifest/segment store usable by local disk
   restore and, later, authenticated prefill/decode handoff.
5. Integrate with Mesh's existing config schema, owner-control apply path,
   management API, CLI, and UI.

## 3. Non-goals

- Extending model context length.
- OS-style demand paging of live model memory.
- Treating local L3 enablement as permission to serve cache objects to peers.
- Automatic migration of old or experimental cache formats.
- Per-model or per-stage disk-budget tuning in the first release.
- At-rest encryption implemented by Mesh. The first release relies on local
  account permissions and full-disk encryption; this limitation must be stated
  in the UI and documentation.

## 4. Canonical persistent configuration

The canonical node configuration is a new runtime-level table:

```toml
[runtime.kv_cache.disk]
mode = "off"                 # off | auto | fixed
directory = "/fast-disk/mesh-kv-cache"
budget_mib = 32768           # required only for mode = "fixed"
minimum_free_mib = 16384     # default: 16384
```

### 4.1 Field semantics

| Field | Type | Default | Semantics |
|---|---|---:|---|
[...]

PR #1637, opened 2026-09-04
Add configurable node-local disk prompt cache

i386:
## Summary

Nodes can now configure and operate a bounded, durable disk prompt cache without relying on experimental ambient `SKIPPY_L3_*` settings.

- adds `[runtime.kv_cache.disk]` with off/auto/fixed modes, absolute node-local roots, hard budgets, and minimum-free reserves
- resolves each field independently through defaults < TOML < public environment < CLI and reports the winning source
- retains one-release legacy environment compatibility without allowing zero/unbounded budgets
- injects one node-scoped manager into every solo and local-split Skippy stage
- applies budget/reserve changes live with rollback on persistence failure; mode/path changes remain restart-required
- adds versioned loopback status, exact numerical-model prune/clear operations, and `mesh-llm kv-cache status|prune|clear`

This is stacked on #1632. The browser settings UI is intentionally deferred. Authenticated owner-control forwarding and the captured-workload release report remain follow-up commits before this draft is ready.

## CLI

```bash
mesh-llm serve --kv-cache-disk auto --kv-cache-disk-dir /fast/mesh-kv
mesh-llm kv-cache status --json
mesh-llm kv-cache prune --target 16GiB --yes
mesh-llm kv-cache clear --model-identity blake3:<exact-id> --yes
```

## Architecture

The disk tier remains the concrete node-local writable store. A single process manager owns the canonical root, root lock, reservations, pins, lifecycle gate, accounting, and status. Lifecycle mutation drains existing cache operations; request-side fills decline immediately during that drain so inference continues with cold prefill.

Automatic budgeting uses `filesystem_available + current_managed_usage` as the stable 20% capacity basis, is capped at 64 GiB, and is constrained by the minimum-free reserve. It can shrink live but does not grow again until restart.

## Compatibility

The default remains off. `SKIPPY_L3_DIR` / `SKIPPY_L3_BUDGET_BYTES` are field-level fallbacks only when public settings do not supersede them. `SKIPPY_L3_BUDGET_BYTES=0` warns and resolves to the bounded 32 GiB compatibility default; it never means unbounded.

## Validation

Validated at `efb6a26dc9ee6ea37d43dde45b10bd5bd3eb0853`:

- formatting and warning-denying Clippy across the touched shipped/cache/server packages
- `skippy-cache`: 114 passed, 0 failed, 1 ignored measurement
- `skippy-server`: 676 passed, 0 failed, 3 ignored model-gated tests
- `mesh-llm-host-runtime`: 2,948 passed, 0 failed, 9 ignored
- `mesh-llm-cli`: 89 passed, 0 failed
- `mesh-llm-config`: 193 passed, 0 failed

## Still required before ready

- authenticated owner-control status/prune/clear with per-node results
- exact local API/CLI black-box fixtures
- captured Buzz/OpenCode/Goose restart lifecycle and two-machine full-chain warm evidence
- c64/c128/c256, write-amplification, TTFT, and saturation measurements


i386:
Folded directly into #1632 at 2b22c5cbf per maintainer direction; continuing the remaining non-UI #1576 work there.

PR #1643, opened 2026-09-04
Reject explicit joins from nodes already in another active mesh

michaelneale:
An operator who asks an **already-running node with active mesh membership** to explicitly join a target in a different mesh now gets an immediate, actionable membership error instead of a later, misleading split protocol-generation exclusion.

This is a runtime-state guard. A fresh `mesh-llm ... --join <token>` process starts unattached and should adopt the target mesh; persisted `mesh-id` is not, by itself, evidence that the new node is actively attached. This PR does not claim to explain or fix the separately observed fresh-start/persisted-state incident.

Example:

```text
mesh ID conflict: local node belongs to mesh 'joiner-mesh' but peer <node-id> belongs to mesh 'host-mesh'; stop the local mesh before joining another mesh
```

## Behavior

- Fail closed when a configured explicit join target advertises an ID different from the node's current in-memory mesh ID.
- Preserve node identity and active mesh membership; a live node does not silently merge or switch meshes.
- Leave fresh, unattached startup joins and ordinary discovery/inbound gossip behavior unchanged.

## Protocol

No wire-format or protocol-generation change. This catches a conflicting live membership state before peer capability admission.

## Validation

At commit `8fee6e036c5bebc6bc48253a7b665946680ee492`:

- `cargo fmt --all --check`
- `cargo check -p mesh-llm`
- `cargo test -p mesh-llm-host-runtime --lib` — 2,941 passed, 0 failed, 9 ignored

The linker emitted pre-existing macOS object deployment-version warnings during the test link.

## Credits

Mic identified the confusing distinction between a fresh startup join and joining from an already-active node, and shaped the fail-closed behavior for the latter.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
@@ -423,6 +423,24 @@ impl Node {
                     context.remote.fmt_short()
                 )
             })?;
+        let is_explicit_join_target = self
+            .join_targets
+            .lock()
+            .await
+            .iter()
+            .any(|target| target.id == context.remote);
+        if is_explicit_join_target
+            && let Some(local_mesh_id) = self.mesh_id().await
+            && let Some(remote_mesh_id) = direct_announcement.mesh_id.as_deref()
+            && local_mesh_id != remote_mesh_id
+        {
+            anyhow::bail!(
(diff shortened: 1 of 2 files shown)


michaelneale:
Closing: this guard models an already-running node with active in-memory mesh membership, but the reported incident used fresh CLI processes. The regression test therefore does not reproduce the bug, and this is not the right fix.

PR #1655, opened 2026-09-04
Compare agentic replay across inference engines

i386:
## Summary

- compare the latest stable Mesh release and main in the existing ordered agentic replay
- append llama.cpp, vLLM, and SGLang OpenAI-compatible arms from an explicit JSON config
- identify external engines by SHA-256 of their reported version, without executable/model/tokenizer file-hash gates
- preserve one workload manifest, request client, warm-up policy, ABBA order, raw evidence, and report across all arms
- document a five-arm stable/main/llama.cpp/vLLM/SGLang example

## Validation

- `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest scripts/tests/test_agentic_replay.py scripts/tests/test_agentic_replay_engines.py` (47 tests)
- `python3 -m py_compile evals/agentic-replay.py evals/agentic_replay_engines.py scripts/tests/test_agentic_replay.py scripts/tests/test_agentic_replay_engines.py`
- plan dry-run resolves stable `v0.75.1` at `3295c902d`, main at `db6fb38a2`, and the five-arm two-pass reverse order

## Run requirements

A real cross-engine run needs a Linux/CUDA host with vLLM and SGLang installed plus equivalent model artifacts configured for all engines.

## Summary by CodeRabbit

* **New Features**
  * Added support for comparing Mesh with llama.cpp, vLLM, and SGLang in agentic replay evaluations.
  * Added configurable engine settings, version-based identities, shared replay plans, and fail-closed comparisons.
  * Added configurable worker-wave requirements for measured cohorts.
  * Added engine identity details and methodology information to evaluation reports.

* **Bug Fixes**
  * Replayed OpenAI messages now omit optional fields with null values.

* **Documentation**
  * Updated router evaluation examples to use the stable v0.75.1 reference.
  * Added guidance for configuring external inference engines.

* **Tests**
  * Expanded coverage for engine validation, version checks, and replay planning.


evals/agentic_replay_engines.py
```diff
@@ -0,0 +1,372 @@
...
+    return value
+
+
+def verify_engine_arm(arm: EngineArm) -> dict[str, Any]:
+    executable = shutil.which(arm.executable)
+    if executable is None:
+        raise FileNotFoundError(f"{arm.label} executable not found: {arm.executable}")
+    if not arm.cwd.is_dir():
+        raise FileNotFoundError(f"{arm.label} cwd not found: {arm.cwd}")
+    version = engine_version(arm)
```
ndizazzo: [P2] Let's resolve and query the executable in the same directory used to launch it. `shutil.which()` and `engine_version()` currently use the runner's working directory, while `start_server()` launches with `cwd=arm.cwd`. With `executable: ./server`, a valid binary that exists only in the configured cwd fails preflight. If both directories contain that path, preflight hashes the wrong binary's reported version. I reproduced the latter with two small executables: the recorded version was `runner-version`, but the launch returned `engine-version`. Resolve relative executable paths against `arm.cwd` and use that same resolved executable and cwd for both the version query and server launch. A regression test with different runner and engine directories would cover both cases.

ndizazzo (changes requested):
I like keeping the external engines on the same replay loop and report. We should fix the executable lookup before relying on the version identities, though. I reproduced a case where preflight records one executable's version and the pass launches another.

All 49 focused replay tests pass, and the four changed Python modules compile. I haven't run a real cross-engine GPU benchmark.

i386:
All review feedback addressed in d94987c05, synced with main (rebased on ce683ca59; no evals files conflicted). 61 replay tests pass locally (48 + 13 engines, up from 49+4 before).

**@ndizazzo's P2 — preflight vs launch executable mismatch:**
- Added `resolve_engine_executable()`, which resolves exactly the way `start_server` runs the command (`cwd=arm.cwd`, standard execvp semantics): a path with a separator resolves against the arm `cwd` (so `executable: ./server` existing only in the configured cwd now passes preflight), an absolute path is used as-is, and a bare name is searched on the runner `PATH` (a bare name is *not* resolved against cwd, matching launch behavior).
- `engine_version()` now runs against that resolved executable, and the resolved path is recorded in `provenance.resolved_executable`. Regression tests cover your runner-version/engine-version two-directory repro both ways.

**CodeRabbit findings:**
- Version preflight is bounded: `subprocess.run(..., timeout=VERSION_COMMAND_TIMEOUT_SECONDS=30)`, `subprocess.TimeoutExpired` → `RuntimeError` with the arm label.
- `load_engine_config()` reads the config bytes once and hashes the same snapshot it parses.
- Plan order now carries the verified identity: preflight runs before `benchmark_plan()`, and `engine_order_specs()` takes the verified `version_sha256` per label, so `plan.json["order"]` and `run.json["builds"]` use the same hash. A `plan`-only invocation (no preflight) keeps the `reported-version-sha256-at-run` placeholder explicitly named `PENDING_VERSION_SHA256`.
- Resume mismatch now says what actually happened: "arm build identity differs ... (commit, binary or runtime hashes, or reported engine version_sha256 changed)".
- Report version cells render via `markdown_code_cell()` — an HTML `<code>` element with pipes, backticks, and angle brackets entity-escaped, so a hostile version/ref string can't terminate the span or split the table.
- README updated to document the executable resolution rules and the 30s version timeout.

i386:
@ndizazzo can I get a review again please

i386:
Follow-up: the re-review's four new findings are also fixed (d7ec5664b):

- **Config snapshot reuse** — `validate_args` stores the model-checked engine config on `args`; `external_config()` returns that snapshot, so a file change after argument validation can't make the run execute a different (valid but different-model) config than the one validated.
- **Launch the verified binary** — `command_for_build` now builds the launch arm from `provenance.resolved_executable` (the exact path preflight resolved and version-probed) instead of the raw configured string, closing the bare-name-PATH-race at launch time.
- **Version query runs in `arm.cwd`** — the SGLang `python -c importlib.metadata` query discovers packages via `sys.path`, which includes the current directory; it now runs in the same directory the server will run in, with a regression test.
- README grammar: "30-second timeout".

Also merged `main` (your update-branch merge `0a169d165` is preserved as the base; replayed on top as a fast-forward — no history rewritten). 65 replay tests pass locally; all CI checks are green on d7ec5664b including CodeRabbit, which posted no further findings.

evals/agentic_replay_engines.py
```diff
@@ -0,0 +1,427 @@
...
+    """
+    has_separator = os.sep in arm.executable or (
+        os.altsep is not None and os.altsep in arm.executable
+    )
+    if has_separator:
+        candidate = Path(arm.executable).expanduser()
+        if not candidate.is_absolute():
+            candidate = arm.cwd / candidate
+        if candidate.is_file():
+            return str(candidate.resolve())
```
ndizazzo: [P2] Let's preserve the virtual environment when resolving the executable. `candidate.resolve()` follows a virtual environment's Python symlink to the system interpreter, so SGLang preflight and launch lose the environment's installed packages. I reproduced a successful version query using the configured interpreter followed by `PackageNotFoundError` through `verify_engine_arm()`. We should make the path absolute without dereferencing symlinks and add a regression test with a symlinked virtual-environment Python.

ndizazzo (approved):

i386:
Fixed the remaining P2 from your approval review (84ff526d2):

**What changed:** `resolve_engine_executable` now absolutizes the candidate with `os.path.abspath` instead of `Path.resolve()` — the path becomes absolute without dereferencing symlinks, so a virtualenv's `venv/bin/python` stays pointing into the venv and both the preflight version query and the server launch keep the environment's installed packages.

**Regression test:** `test_venv_python_symlink_is_not_dereferenced` builds a venv-style layout (`venv/bin/python` → symlink → system interpreter), runs `verify_engine_arm`, and asserts both `provenance.resolved_executable` and the version-query argv keep the venv path rather than the dereferenced system interpreter — the exact scenario from your repro. Also updated the two existing assertions that baked in the old `.resolve()` output.

All 66 tests green locally (15 engine-arm + 51 replay). CI is running on the new head.

PR #1535, opened 2026-08-29
docs: Windows section - native CUDA works on CUDA 13 drivers; document current workarounds

Virgile-pct:
## Original problem

The README's Windows guidance tells users with CUDA 13.x drivers to switch to WSL2 when `mesh-llm` reports `0 GPUs`. That advice predates the cuda12-runtime compatibility fix (#1127) and no longer matches current releases: on v0.76.0-rc7 and v0.76.0-rc8, GPU detection and native CUDA inference work fine on a CUDA 13.3 driver. Meanwhile the actual blockers a native Windows user hits today are three distribution/loading bugs (#1510, #1511, #1512) that the README says nothing about — so the only documented path steers everyone into an unnecessary WSL2 setup, and nobody reaches the working native one.

## Diagnostics

Verified on Windows 11 (10.0.26200), RTX 4070 Ti, NVIDIA driver 610.74 (CUDA UMD 13.3), on v0.76.0-rc7 and again on v0.76.0-rc8 (published 2026-08-29):

- `mesh-llm gpus` detects the GPU as `Backend device: CUDA0` natively (no WSL2).
- With the three workarounds now documented in this PR, `serve --local-model-only` reaches `api ready at http://127.0.0.1:9337/v1` and serves chat completions at ~60 tok/s on Qwen3-8B Q4_K_M — full details, logs and load-order experiments are in #1510 / #1511 / #1512.

## Fix

README only:

- Adds a "Running natively on Windows (NVIDIA)" subsection: states that native CUDA works on CUDA 13 drivers, documents the currently working sequence (`-PreRelease` install, CUDA runtime via `runtime install --bundle-dir` from the `-cuda` product bundle, runtime `lib` dir on `PATH`), each step linked to its tracking issue so the block can shrink as fixes land.
- Adds Windows-specific notes: `--local-model-only` requires an absolute `.gguf` path, models cache under `%LOCALAPPDATA%\huggingface\hub`, unsigned binary → SmartScreen prompt.
- Reframes the existing WSL2 content as an alternative (kept intact: CUDA toolkit install, mirrored networking, firewall rules, LAN path matching).
- Updates the WSL2 pointer next to the install instructions accordingly.

No code changes; no compatibility concerns.

## Validation

- [x] Docs-only change; ran no code checks. The documented commands were executed end to end on the machine described above.
- [x] No UI changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
[...]

PR #1657, opened 2026-09-04
docs(contributing): how to run crate test gates on a bare Windows checkout

Virgile-pct:
Companion to the install-side Windows notes from #1535, for the developer side.

A Windows checkout without prepared llama.cpp archives cannot build the test targets of crates that pull in `skippy-ffi`'s static link mode — `mesh-llm-system` does, through `mesh-llm-runtime-install` (`default-features = false`) — because `skippy-ffi/build.rs` refuses to prepare the native ABI on Windows. `just test-all` needs the same preparation.

This adds a short "Testing crates on native Windows" subsection to CONTRIBUTING.md: the `--features dynamic-native-runtime` workaround (feature unification enables `skippy-ffi`'s `dynamic-runtime`, whose build script returns early), the exact gate commands that run on a bare Windows machine, `--no-deps` Clippy scoping plus the three Windows-only dead-code sites that currently fail `-D warnings` on `main` (so contributors compare against `main` before blaming their diff), the env vars for a prepared native build, and the MSVC prerequisite.

Verified on Windows 11 at `main` f945d5af3: the bare `cargo test -p mesh-llm-system --lib` failure, every command in the snippet, and the three dead-code sites (`cargo clippy --no-deps ... -D warnings`). Docs only; no code change.

## Summary by CodeRabbit

* **Documentation**
  * Updated Windows testing guidance to explain that conditionally compiled code may be reported as dead on one platform.
  * Added a recommendation to compare platform-specific warnings with the same command on the main branch when the warning appears unrelated to the current changes.
  * Clarified how to assess warnings that occur only in platform-specific builds.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -162,6 +162,39 @@ not report, lock, or clean a second artifact tree.
 
 On native Windows, `just check-release` runs the host-safe Rust/doc invariant subset and skips the Bash-only `install.sh` / `package-release.sh` parity checks. Run it on macOS or Linux when you need full shell parity coverage.
 
+### Testing crates on native Windows
+
+A bare Windows checkout cannot build the test targets of crates that pull in
+`skippy-ffi`'s static link mode — `mesh-llm-system` does, through
+`mesh-llm-runtime-install`, which depends on `skippy-ffi` with
+`default-features = false` — because `skippy-ffi/build.rs` then requires
+prepared llama.cpp ABI archives
+(`automatic native preparation is not supported for Windows from build.rs yet`).
+`just test-all` needs the same native preparation, through its Bash pipeline.
+
+For crate suites that do not exercise the native runtime, enable
+`dynamic-native-runtime`: feature unification turns on `skippy-ffi`'s
+`dynamic-runtime`, whose build script returns early. The `mesh-llm-system`
+gates then run on a machine without a prepared native build:
+
+```powershell
+cargo test --locked -p mesh-llm-system --lib --features dynamic-native-runtime
+cargo clippy --no-deps -p mesh-llm-system --all-targets --features dynamic-native-runtime -- -D warnings
+cargo fmt --check -p mesh-llm-system
+cargo run -p xtask -- repo-consistency no-console-print
+```
+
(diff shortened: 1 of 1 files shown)


i386 (changes requested):
The Windows workflow, feature-unification explanation, PowerShell command syntax, and native-build environment variables all check out. One documentation issue needs to be fixed before merge:

The paragraph after the command block hard-codes three items as code that "currently fails `-D warnings`" on Windows. PR #1659 is specifically removing all three of those failures, and I verified its Windows-target Clippy gates pass. If these companion PRs both land, this new `CONTRIBUTING.md` section is immediately stale.

Please remove the exact resolved-site list and the claim that those sites currently fail. The durable advice can remain generic, for example: compare any unexpected platform-specific warning against the same command on `main` before attributing it to a change.

Validation at `8e8ea200fb64071a4f7926350973291e86d89e98`:

- `cargo fmt --check -p mesh-llm-system`
- `cargo test --locked -p mesh-llm-system --lib --features dynamic-native-runtime` (192 passed)
- `cargo clippy --no-deps -p mesh-llm-system --all-targets --features dynamic-native-runtime -- -D warnings`
- `cargo run -p xtask -- repo-consistency no-console-print`

Once the transient lint list is removed or made non-stale, the guide is good to go.

Virgile-pct:
Done in b0b39a3: dropped the site list and the "currently fails" claim, kept the generic advice about comparing a platform-specific warning against `main`. You're right that with #1659 the list would have been stale on landing. `repo-consistency no-console-print` clean at that head; the rest of the section is unchanged.

i386 (approved):
Approved at b0b39a3b3710891ed54c344d1eb54d8b6345fcc4. The requested cleanup is complete: the transient list of Windows-only lint failures and the claim that they currently fail are removed, while the durable compare-against-main guidance remains. I re-read the full current diff and found no remaining actionable issue. Validation on this exact head: cargo fmt --check -p mesh-llm-system; cargo test --locked -p mesh-llm-system --lib --features dynamic-native-runtime (192 passed); cargo clippy --no-deps -p mesh-llm-system --all-targets --features dynamic-native-runtime -- -D warnings; cargo run -p xtask -- repo-consistency no-console-print.

PR #1658, opened 2026-09-04
test(model-hf): make the read-only filesystem warning test portable

Virgile-pct:
`cache_paths::tests::read_only_filesystem_warning_includes_os_error_and_recovery_path` builds `io::Error::from_raw_os_error(30)` and asserts that the warning contains "Read-only file system". Raw OS error 30 is EROFS on POSIX, but on Windows it is ERROR_READ_FAULT ("The system cannot read from the specified device"), so the test fails on native Windows while the code under test is fine.

The test now asserts the actual contract: the warning surfaces the OS error text verbatim (`message.contains(&os_error)`), whatever the platform's wording for code 30 is. The recovery-path and env-var assertions are unchanged.

Found while reviewing #1628 on Windows (see the comment there). Unrelated to that PR: `cache_paths.rs` was untouched by it.

Verified on Windows 11 at `main` f945d5af3: `cargo test --locked -p model-hf --lib` 47 passed (was 46/47), `cargo clippy --no-deps -p model-hf --all-targets -- -D warnings` clean, `cargo fmt --check -p model-hf` clean, `repo-consistency no-console-print` clean. Not run on POSIX here; there the new assertion should be at least as strict as the old one, since the error string is "Read-only file system (os error 30)".

## Summary by CodeRabbit

* **Tests**
  * Improved validation of read-only filesystem warnings by checking the operating system’s actual error text.
  * This makes test coverage more reliable across platforms with different wording for filesystem access errors.

### DIFF
--- a/crates/model-hf/src/cache_paths.rs
+++ b/crates/model-hf/src/cache_paths.rs
@@ -342,16 +342,20 @@ mod tests {
 
     #[test]
     fn read_only_filesystem_warning_includes_os_error_and_recovery_path() {
+        // Raw OS error 30 is EROFS on POSIX but ERROR_READ_FAULT on Windows, so
+        // assert the contract (the OS error text is surfaced verbatim) rather
+        // than the POSIX wording.
+        let os_error = std::io::Error::from_raw_os_error(30).to_string();
         let warning = DownloadDirectoryFallback {
             kind: DownloadDirectoryKind::HuggingFaceXet,
             requested: PathBuf::from("/"),
             selected: PathBuf::from("/writable/data/huggingface/xet"),
-            error: std::io::Error::from_raw_os_error(30).to_string(),
+            error: os_error.clone(),
             raw_os_error: Some(30),
         };
 
         let message = warning.to_string();
-        assert!(message.contains("Read-only file system"));
+        assert!(message.contains(&os_error));
         assert!(message.contains("/writable/data/huggingface/xet"));
         assert!(message.contains("HF_XET_CACHE"));
         assert!(message.contains("MESH_LLM_DATA_DIR"));


i386 (approved):
The test now checks the actual contract: `DownloadDirectoryFallback` surfaces the platform's `io::Error` text verbatim, without assuming that raw error 30 has POSIX wording. The recovery-path and environment-variable assertions remain intact, so this does not weaken the user-facing diagnostics coverage.

Validation at `f1976d2bd892c2e6731842c03713411b8fee0643`:

- `cargo fmt --all --check`
- `cargo test -p model-hf` (49 passed, 9 network tests ignored)
- `cargo clippy -p model-hf --all-targets -- -D warnings`
- the same Clippy gate cross-targeted to `x86_64-pc-windows-gnu`

This is a correct and portable test fix.

i386:
Test review (fork CI + attribution on the delta since approval)

The fix itself is right: asserting the surfaced OS error text verbatim is the actual contract, and `os_error.clone()` keeps the `raw_os_error: Some(30)` construction honest. Nice find that `io::Error::from_raw_os_error(30)` isn't EROFS on Windows.

Two process notes before this can merge:

1. CI has not run on the current head — all five pipelines are sitting in `action_required` (fork pull-request approval gate). Once a maintainer approves the runs, worth re-checking this test on POSIX CI too: the new assertion should be at least as strict as the old one there (the error string is "Read-only file system (os error 30)"), but that's untested until a runner picks it up.

2. The approval on this PR predates the 09-07 push, so the review is stale by GitHub's rules and the one-commit delta since then is exactly the test-body change. Not a complaint — just means whoever re-approves should look at that delta rather than rubber-stamp.

michaelneale:
Reviewed on Mic's behalf. Correct fix, and the reasoning in the comment is the right one: raw OS error 30 is `EROFS` on POSIX but `ERROR_READ_FAULT` on Windows, so asserting the literal string "Read-only file system" was pinning a platform's wording rather than the behaviour.

Asserting `message.contains(&os_error)` where `os_error` is the same `from_raw_os_error(30).to_string()` fed into the struct is the right shape — it tests the actual contract (the OS error text is surfaced verbatim) and stays meaningful on every platform. The other three assertions (`/writable/data/huggingface/xet`, `HF_XET_CACHE`, `MESH_LLM_DATA_DIR`) still carry the real regression value.

One observation, take it or leave it: the test name is now slightly wider than what it asserts — it says `read_only_filesystem_warning_...` but on Windows error 30 isn't a read-only-filesystem error at all. Not worth a revision on its own; if you touch it again, `os_error_warning_includes_os_error_and_recovery_path` would describe it more honestly.

Approve.

Virgile-pct:
Thanks. Agreed on the name: on Windows, error 30 is not a read-only filesystem, the test asserts the contract rather than the wording, and `os_error_warning_includes_os_error_and_recovery_path` says that better. I am leaving the branch untouched so it stays mergeable as reviewed; if it needs another push for any reason, the rename goes in with it.

PR #1659, opened 2026-09-04
chore: silence the Windows-only dead-code lints in mesh-llm-system

Virgile-pct:
Windows counterpart of the macOS pass in #1591. On a bare Windows checkout, `cargo clippy -p mesh-llm-system --all-targets --features dynamic-native-runtime -- -D warnings` fails on two cfg-gated sites that are dead only on Windows:

- `download_optional_url` in `autoupdate/release_fetch.rs`: its sole caller, `install_native_runtime_after_update`, is `#[cfg(not(windows))]`; the function is now gated the same way.
- `SignalOutcome::{AlreadyDead, Skipped}` in `backend.rs`: constructed only inside the `#[cfg(not(windows))]` pid/name verification block, still matched on Windows; `allow(dead_code)` on those two variants under `cfg(windows)`.

No behavior change on any platform; each attribute is a no-op outside Windows. Each site carries a one-line comment saying why it is dead there.

The third Windows-only site, `MaterializedOutputLock.file` in `skippy-runtime/src/package/materialized_cache.rs`, was in this PR until the review: #1685 replaces the unix-only `flock` with `File::lock`/`unlock`, which reads the field on every platform, so #1685 carries that file and the `allow` was dropped here. Until #1685 lands, `cargo clippy -p skippy-runtime -- -D warnings` still fails on Windows on that one field.

Verified on Windows 11 at the rebased head (on `main` 03267ccf0): `cargo clippy --no-deps -p mesh-llm-system --all-targets --features dynamic-native-runtime -- -D warnings` clean (it failed on the two sites before); earlier at `main` f945d5af3, `cargo test --locked -p mesh-llm-system --lib --features dynamic-native-runtime` 186 passed; `cargo fmt --check` and `repo-consistency no-console-print` clean. Not run on Linux or macOS here. Note that the Windows CI lane does not run for this PR: `ci/ownership.yml` maps only the Windows scripts and the `mesh-llm-nodejs` / `skippy-ffi` crates to `platform-windows`.

The Windows locking gap in `MaterializedOutputLock` noted in the first version of this description is now #1682, fixed by #1685.

## Summary by CodeRabbit

* **Chores**
  * Added platform-specific annotations and documentation for internal update and process-handling code.
  * No user-facing behavior changes.

### DIFF
--- a/crates/mesh-llm-system/src/autoupdate/release_fetch.rs
+++ b/crates/mesh-llm-system/src/autoupdate/release_fetch.rs
@@ -422,6 +422,8 @@ async fn install_native_runtime_after_update(
 ) {
 }
 
+// Only the non-Windows `install_native_runtime_after_update` calls this.
+#[cfg(not(windows))]
 async fn download_optional_url(url: &str, path: &Path, expected_sha256: &str) -> Result<bool> {
     let response = reqwest::Client::builder()
         .timeout(std::time::Duration::from_secs(60))
--- a/crates/mesh-llm-system/src/backend.rs
+++ b/crates/mesh-llm-system/src/backend.rs
@@ -172,7 +172,10 @@ enum ProcessSignal {
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum SignalOutcome {
     Sent,
+    // Only the non-Windows pid/name verification path constructs these.
+    #[cfg_attr(windows, allow(dead_code))]
     AlreadyDead,
+    #[cfg_attr(windows, allow(dead_code))]
     Skipped,
     Failed,
 }


i386 (approved):
Reviewed each cfg site against its call/read paths. The guards are narrowly scoped to the platform where the item is genuinely unreachable, while preserving the non-Windows behavior.

Validation at `b52047a71d46a89c9d4c4cbce313856ad5f653ea`:

- `cargo fmt --all --check`
- both requested Clippy gates on macOS with `-D warnings`
- both Clippy gates cross-targeted to `x86_64-pc-windows-gnu` with `-D warnings`
- `cargo test -p mesh-llm-system --features dynamic-native-runtime` (192 passed)
- `cargo test -p skippy-runtime --features dynamic-native-runtime` reached 112 passing tests; the only three failures are the documented missing-loaded-native-runtime cases and are unrelated to this diff

The Windows lint fix looks correct and appropriately minimal.

i386:
Three sites reviewed, all good — each gate is the minimal one that kills the lint (variant-level `cfg_attr` for `SignalOutcome`, whole-function `cfg(not(windows))` for `download_optional_url`, field-level for `MaterializedOutputLock.file`), each with a one-line why, and no behavior change on any platform.

Two notes:

- Approvals on this PR predate the 09-07 head push (which reordered the materialized-cache hunk but didn't change the substance). Fine to re-approve on that delta.

- The side-observation in the description is real and I think it deserves an issue: on non-unix targets `MaterializedOutputLock` holds the file open but takes no lock — there's no `LockFileEx` equivalent — so the materialized cache currently has no mutual exclusion on Windows. That's a latent correctness bug independent of this PR; happy to write it up if you'd rather keep this PR chore-sized.

Virgile-pct:
Agreed it is a real bug on its own. Written up as #1682 with the Windows specifics (what `MaterializedOutputLock` holds open, what it does not lock, and the `LockFileEx` route), so this PR stays chore-sized.

michaelneale:
Reviewed on Mic's behalf. The three annotations are each accurate — I traced them:

- `download_optional_url`: the only caller is `release_fetch.rs:375`, inside the `#[cfg(not(windows))]` `install_native_runtime_after_update` at `:365`; the `#[cfg(windows)]` twin at `:417` is an empty body. `#[cfg(not(windows))]` on the fn is correct and strictly better than an `allow`, since it removes the code rather than hiding the warning.
- `SignalOutcome::{AlreadyDead, Skipped}`: agreed these are only constructed on the non-Windows pid/name verification path.
- `MaterializedOutputLock.file`: correct against `main` today.

**One coordination issue, not a defect in your diff.** The third hunk collides with #1685 (`fix: serialize materialized cache writes on Windows`), which replaces the `unsafe libc::flock` block with `File::lock()`/`unlock()`. Once that lands, `self.file.unlock()` is called on *all* platforms, so the field is genuinely read on Windows and your `#[cfg_attr(not(unix), allow(dead_code))]` becomes stale — an `allow` for a warning that can no longer fire. Not harmful, just a small piece of misleading provenance in a PR whose whole purpose is accurate provenance.

Whichever of the two merges second should drop that hunk. Since #1685 both fixes real Windows behaviour and eliminates the lint at its source, my suggestion is to drop the `materialized_cache.rs` hunk here and let #1685 carry it.

**Worth knowing about the CI evidence:** the Windows lane **skipped** on this PR (`Windows = skipping`, run 34169032510), so the warnings this PR silences were never actually compiled by its own CI. `ci/ownership.yml:226-232` scopes the `platform-windows` domain to `scripts/build-windows.ps1`, `install-windows-sdk.ps1`, `windows-native-runtime-deps.py`, `windows-warm-caches.yml`, plus the crates `mesh-llm-nodejs` and `skippy-ffi` (`:329-331`) — none of which this PR touches, even though it edits `mesh-llm-system` and `skippy-runtime` specifically for Windows. So this is unverified-by-CI on the platform it targets, and a future change that re-introduces these warnings won't be caught either. I could not cross-compile locally to confirm (no Windows SDK on this host), so I'm relying on your reading of the warnings, not independent verification.

No objection to the change itself.

Virgile-pct:
Thanks for tracing them. Agreed on the `materialized_cache.rs` hunk: #1685 makes the field read on every platform, so I dropped it here and #1685 carries that file (same thing I noted on #1685 yesterday). The commit is amended to the two remaining sites, `mesh-llm-system` only, and the description is updated.

On the CI evidence, you are right: the Windows lane skipped on this PR (run 34169032510, the Windows job is `skipped`), because `ci/ownership.yml` only maps the four Windows scripts and the `mesh-llm-nodejs` / `skippy-ffi` crates to `platform-windows`. The verification exists but it is local. On a Windows 11 build of yesterday's #1696 head (main at 592de242b, without this change), `cargo clippy --no-deps -p mesh-llm-host-runtime --all-targets --features dynamic-native-runtime -- -D warnings` prints exactly these three warnings from the dependencies (`materialized_cache.rs:94`, `release_fetch.rs:425`, `backend.rs:175`). On this branch as it was, `cargo clippy --no-deps -p mesh-llm-system -p skippy-runtime --all-targets --features dynamic-native-runtime -- -D warnings` exits 0 on the same machine. With the hunk dropped, `mesh-llm-system` is still clean and `skippy-runtime` reports `field file is never read` again until #1685 lands, as expected.

The scoping gap is real: a change to `mesh-llm-system` or `skippy-runtime` does not reach the Windows lane today, so a regression of these lints would go unnoticed by CI. Happy to open an issue, or a small `ownership.yml` change, if that is the direction you want; I have not touched the CI manifests here.

PR #1662, opened 2026-09-04
fix(skippy): restore recurrent split prefix reuse

i386:
Split serving now restores shared-prefix cache hits for dense and recurrent state, and the Linux regression smoke exercises repeated prompts without deadlocking model startup or crashing restored dense sessions.

## What changed

- Preserve reusable recurrent exact-state checkpoints even when hybrid state exceeds the attention-derived byte estimate.
- Seed downstream split-prefill accounting from restored tokens so later prompts advance to the next checkpoint.
- Release the process-wide stderr lock between telemetry batches; native model-loading progress can no longer block behind an idle telemetry receiver.
- Fall back to direct greedy selection when backend sampling returns no token for a restored dense prefix, instead of reusing an already-consumed sampler chain.
- Send each prompt length twice and require every repeat to extend cache coverage; dense repeats remain near-full restores while recurrent repeats may restore at checkpoint boundaries.
- Run pinned dense SmolLM2 and recurrent Qwen3.5 legs sequentially in one `KV caching smoke (dense + recurrent)` job, with both models staged through trust-scoped caches before execution.
- Dump both nodes' raw `/api/runtime/stages` views on topology timeout.

## Why the failed run timed out

`SKIPPY_TELEMETRY_STDERR=1` promoted embedded telemetry to debug. The telemetry thread held Rust's process-wide stderr lock while blocking for the next event. Native llama.cpp model-loading progress writes to the same stderr handle, so stage 0 blocked during load and never published its topology. The apparent peer-discovery asymmetry was downstream of that blocked startup.

The new repeated-prompt coverage then exposed a separate dense-cache crash: `llama_get_sampled_token_ith` could return `LLAMA_TOKEN_NULL` after a resident-prefix restore, and the fallback called `llama_sampler_sample` on the backend-initialized sampler chain. Direct greedy fallback fixes that path.

## Validation

- Exact rebuilt Linux two-node smoke on `white.local`, telemetry enabled:
  - dense: topology ready; repeat restores `664/665`, `808/809`, and `936/937` prompt tokens.
  - recurrent exact-state: topology ready; repeat restores `512/681`, `640/857`, and `768/1001`, with `kv-recurrent` observed.
- The formerly failing GitHub two-node split smoke is green on this head: https://github.com/Mesh-LLM/mesh-llm/actions/runs/33941614902/job/101244139768
- `just ci-validate`: 817 passed, 7 skipped; Actionlint and all repository-consistency gates passed.
- `cargo test --locked -p skippy-server --lib`: 685 passed, 3 ignored.
- `cargo check` and warning-denying Clippy passed for `skippy-server` and `mesh-llm`.
- `cargo fmt --all -- --check` and `just ci-shellcheck scripts/ci-two-node-split-smoke.sh` passed.
- The full llama.cpp patch queue applies cleanly from the pinned upstream and the rebuilt CPU runtime passed both smoke legs.

## Trust boundary

Pull-request entrypoints intentionally invoke protected reusable workflow definitions from `main`. This PR run therefore exercises the changed smoke script through the currently protected dense job. The combined dense-plus-recurrent workflow definition becomes protected after merge; its exact control flow and both cache families were exercised locally above.

### DIFF
--- a/.agents/skills/manage-ci/references/current-inventory.md
+++ b/.agents/skills/manage-ci/references/current-inventory.md
@@ -160,13 +160,13 @@ removable after this branch's runner contract is active on protected main.
 | `ci-{linux,macos,windows}-runtime-slice.yml` | Platform-pure native runtime producers |
 | `ci-{linux,macos,windows}-product-slice.yml` | Platform-pure composition-only product consumers |
 | `ci-platform-checks-slice.yml` | macOS portable/unit, Windows portable, and Windows log-store privacy ACL checks |
-| `ci-linux-product-smoke-slice.yml`, `ci-macos-product-smoke-slice.yml` | Platform-local CPU, CUDA (`gpu-nvidia` self-hosted), two-node, Metal and model-download consumers; the Linux split smoke serializes pinned dense SmolLM2 and recurrent Qwen3.5 rows with PR fail-fast; ROCm/Vulkan products remain package-verified pending eligible inference runners |
+| `ci-linux-product-smoke-slice.yml`, `ci-macos-product-smoke-slice.yml` | Platform-local CPU, CUDA (`gpu-nvidia` self-hosted), two-node, Metal and model-download consumers; one Linux KV caching smoke job runs pinned dense SmolLM2 then recurrent Qwen3.5 legs and requires both to pass; ROCm/Vulkan products remain package-verified pending eligible inference runners |
 | `ci-linux-sdk-slice.yml`, `ci-macos-sdk-slice.yml` | Platform-local Rust/Kotlin/Swift smoke consumers; SDK producers are independent top-level calls and each smoke receives the lane-local immutable UI artifact |
 | `ci-runner-contract-slice.yml` | Provider/cache/plan trust and main runner-image checks |
 | `native-sdk-artifact.yml` | Typed native SDK producer |
 | `swift-sdk-artifact.yml` | Host-only/full XCFramework producer; full mode builds the seven Apple Rust targets as a bounded matrix (maximum four concurrent macOS runners) and joins their immutable libraries in one assembly job, while host-only remains a single producer. Trusted main remains `macos-15`, while eligible same-repository PRs follow the protected Depot macOS 15 gate |
 | `smoke.yml` | Artifact-based inference/OpenAI/split smoke |
-| `scripted-binary-smoke.yml` | Artifact-based scripted product smoke with an optional typed model context-size override |
+| `scripted-binary-smoke.yml` | Artifact-based scripted product smoke with optional typed model context-size and recurrent-model inputs; recurrent models restore and save through a dedicated trust-scoped cache before the smoke runs |
 | `sdk-smoke.yml` | Artifact-based SDK consumers; all SDK rows consume the lane's immutable console UI artifact, while Rust smoke restores the main-seeded, target/profile/image/toolchain/recipe-bound Cargo/target cache through `Swatinem/rust-cache` |
 | `hf-download-smoke.yml` | Hugging Face download smoke |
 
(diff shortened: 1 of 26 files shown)


i386:
CI verification by Scam agent:

The normal PR entrypoint pins `ci-linux-lane.yml@main`, so its green two-node job used the protected pre-change workflow and exercised only the existing dense model. I dispatched the branch version of the Linux lane at code commit `1062bef4734017d663957a35bac7f7645ee6f00c` with a minimal PR-safe lane plan. The resulting [branch-scoped run](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33916872479) passed and created two distinct same-machine two-node jobs:

- Dense SmolLM2: prompt tokens `665 -> 809 -> 937`; cached tokens `0 -> 658 -> 802`.
- Recurrent Qwen3.5-0.8B: prompt tokens `644 -> 788 -> 916`; cached tokens `0 -> 512 -> 640`.

Both jobs independently reported two stages on two node IDs before issuing requests, then reported increasing cached tokens on progressively longer prompts.

The current PR head `81fa197f99677c1cefc90186be5fb86be18a4df7` differs from the verified code commit only by documentation/comment clarification for the exact-state hard-limit exception. `cargo fmt --all -- --check` and `git diff --check` pass on that head. The full local `just ci-validate` reached 752 tests, with 737 passing and 7 skipped; its remaining 2 failures and 6 import errors are local sandbox/dependency limitations (`PermissionError: Operation not permitted` in process cleanup probes and missing PyYAML), while the PR Quality workflow covers the same validation on CI.

michaelneale:
testing this now

michaelneale (approved):
I tested a version qwen 3.8 flash splits - and worked well.

michaelneale:
Read-only review by Jimmy of `e52b6bbf0197e36f6e962aeb91bdc30516527285`.

### Concrete finding: the smoke accepts a missed second-request restore

In `scripts/ci-two-node-split-smoke.sh:383–392`, the all-zero case is handled separately, but the subsequent comparison only requires `cached_counts[1] < cached_counts[2]`. Consequently `[0, 0, 640]` passes: request 2 can completely miss the prefix and the regression gate still goes green. This contradicts the adjacent comment that a partial follow-up miss must fail directly.

I extracted the Python validator verbatim from this commit with `git show` and ran it against synthetic `chat.completion` responses with prompt counts `[644, 788, 916]`:

| Cached counts | Actual exit |
|---|---:|
| `[0, 0, 640]` | **0 (incorrect pass)** |
| `[0, 512, 640]` | 0 |
| `[0, 512, 0]` | 1 |
| `[0, 0, 0]` | 75 |

Suggestion: after the all-zero retry classification, require `0 < cached_counts[1] < cached_counts[2]`. Add behavioral validator tests for these four cases; workflow string assertions do not catch this.

### Current evidence / merge advice

The current PR Linux result is failing: [two-node split job](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33933710107/job/101225599553). Its dense SmolLM2 run times out **before inference**, reporting `ready=false stages=1 nodes=1 models=0`. This is a topology-readiness failure, not evidence of failed cache reuse, and needs diagnosis before treating the PR as merge-ready.

I also independently read the earlier [branch-scoped recurrent job](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33916872479/job/101172462562): at `1062bef4734017d663957a35bac7f7645ee6f00c`, its log reports `stages=2 nodes=2` and cached counts `0 → 512 → 640`. That is useful historical evidence, not a successful current-head run.

The inspected checkpoint paths are coherent: stage 0 cuts prefill at the shared boundary, validates the canonical native position before exporting, and downstream seeds restored tokens only after a successful chain restore. I did not establish another production correctness defect in those paths. As additional coverage, I would prioritize alternating two unrelated conversations under eviction pressure and cached-vs-cold deterministic output comparison (the smoke currently checks usage, not output equivalence).

No source edits, Rust builds, live mesh runs, or CI dispatches performed; the local reproduction above exercises only the extracted validator, not the full test suite.

PR #1667, opened 2026-09-05
Add a live runtime event stream with a contract-pinned wire format

ndizazzo:
## What this gives you

A live, structured view of what the runtime is doing — over a documented wire contract instead of scraped logs.

```bash
mesh-llm serve --gguf <model.gguf> --port 9337 --console 3131 --log-format json
curl -sN -H 'Accept: text/event-stream' http://127.0.0.1:3131/api/runtime/events/v1
```

Four frame types: `runtime_state` (opening snapshot of models, stages, sessions, requests, devices, cache), `runtime_event` (individual facts), `runtime_health` (engine counters and their frozen bounds), and `runtime_replay_gap` (says honestly when you missed frames rather than silently skipping).

Every event carries an `operation` identity, so one chat completion is followable end to end — `request_received` → `request_admitted` → `first_token_produced` → `prefill_completed` → `generation_completed` → `request_completed` — with `operation.root` equal to the `request_id` already in the JSON logs. Reconnect with a cursor (`rt1:<uuid>:<seq>`) to resume, or be told the frames aged out.

Also fixes `/api/runtime`, which previously returned default-false role booleans regardless of what the node was doing; it now matches `/api/status`.

## Architecture

An engine owns ingress. Producers call `try_submit` and never block, never fail a request, and never allocate on the submit path (proven by a counting-allocator test that kills a deliberate `format!()` mutation). A single driver task drains it.

Four delivery classes are handled distinctly rather than collapsed: terminals go through a wake list, state transitions coalesce per `(operation, kind)`, progress flushes at most once per 100 ms per operation, diagnostics FIFO. All four are applied in ingress-sequence order through one transactional reducer.

Bounds are frozen constants, not scattered literals: reservation table 3136, state lane 4096, diagnostic lane 2048, replay 4096 frames / 300 s / 8 MiB, 32 subscribers. The reducer evicts operations when their reservation releases, with an LRU cap on unreserved entries so the map cannot grow with traffic. Health is versioned and delivered per subscriber — each consumer gets a frame within 1 s of a counter changing, and nothing while idle.

```mermaid
flowchart LR
  subgraph PROD["producer code"]
    direction TB
    A["runtime_event_engine()"]
    B["OperationReservation"]
    C["ScopedIngress"]
    D["SubmitOutcome<br/>Accepted · Coalesced<br/>DroppedProgress · RejectedCapacity"]
  end

  subgraph ENG["engine"]
    direction TB
    RT["ReservationTable<br/>3136 slots · write-once terminal slot<br/>dropped guard synthesizes<br/>terminal_not_delivered"]
    GATE["ingress_gate<br/>mint sequence, pick lane<br/>by delivery class<br/>returns IMMEDIATELY"]
    RED["driver → reducer<br/>sequence is the only order key"]
    RB["ReplayBuffer<br/>4096 frames · 300 s · 8 MiB"]
    SR["SubscriberRegistry<br/>≤ 32 subscribers<br/>one bounded queue each"]
  end

  subgraph CONS["consumers"]
    direction TB
    SSE["GET /api/runtime/events/v1<br/>SSE, cursor-resumable"]
    PRES["presentation subscriber<br/>→ OutputEvent::Info<br/>→ JSON log + TUI"]
    OWN["your own SubscriptionHandle<br/>in-process Rust"]
  end

  A -->|"reserve_root(op_id, synth)"| B
  B -.->|"write-once slot"| RT
  B -->|".ingress()"| C
  C -->|"try_submit(fact)"| GATE
  GATE -.->|"returns"| D
  GATE --> RED
  RED --> RB
  RED --> SR
  RB -->|"replay on cursor resume"| SSE
  SR --> SSE
  SR --> PRES
  SR --> OWN
```

Three properties worth reading off the producer code:

* **`let _ =` everywhere.** Producers deliberately ignore `SubmitOutcome`. Observability degrades; serving does not.
* **Absent engine is normal.** `runtime_event_engine()` returning `None` means "event system inactive", not an error.
* **You cannot forget a terminal.** If an `OperationReservation` guard drops without one, the engine synthesizes the family-correct terminal with `outcome: Unknown`, `reason: TerminalNotDelivered`.

In-process Rust consumers subscribe to the same published frames via `engine.subscribers().subscribe()`. Each subscriber owns its own bounded queue rather than sharing a broadcast ring, which is what makes exact per-handle frame/byte/age accounting — and per-handle lag disconnect — possible. This PR ships one such consumer: the presentation subscriber, the only place a `RuntimeFact` becomes an `OutputEvent`, so the JSON/pretty log and the TUI dashboard can never disagree about what happened.

## Protocol

Wire version stays **1**. The event catalog, projected keys, and per-profile key sets live in `crates/mesh-llm-runtime-event-contracts/inventory/runtime_events.toml`, are regenerated into a TypeScript fixture, and are pinned by byte-exact sample frames for all four frame types — enforced by a Rust round-trip test *and* an independently-written TS parser.

**Skippy ABI:** adds feature bits **31–36** (runtime-event reporter plus five event families) and native patches **0024–0033**, sitting after `main`'s 0001–0023; `main` keeps bits 29 and 30. `SKIPPY_ABI_VERSION_PATCH` is 54, identical between `include/skippy/common.h` and `crates/skippy-ffi/src/lib.rs`. Mesh gossip and the plugin protocol are untouched.

## Validation

Twenty planned tasks, each independently adversarially reviewed before being marked done; six were rejected on first submission and reworked. Evidence is under `.omo/evidence/event-system-fixes/`.

Live sampling against a running node covered 11 conditions, 9 passing on captured frames — including 40 concurrent completions with **36 `generation_completed` roots against 36 successful completions, exactly 1:1** and `terminal_delivery_failed = 0` (before this work the same load lost 17 of 36 terminals). A real privacy defect was found and fixed: a raw filesystem path passed to `mesh-llm load` was reaching the wire and lodging a phantom `runtime_state` entry.

**Performance** — production vs event-disabled on the *same* release binary, one variable, three independent seeds:

| Metric | Result |
|---|---|
| `decode_tok_s`, `decode_only_tok_s` | **12 of 12 PASS**, inside the 3% bound, no directional trend |
| `ttft_ms` | **PASS at n=80** — +0.34%, 95% CI [-1.34%, +2.30%] |
| `callback_ingress_p99` | **2.0 µs** against a 100 µs budget |

TTFT needed three runs to settle. At 20 primary pairs it read +15.7% then +8.8% (underpowered); rerun at 80 pairs on a fresh seed it collapsed to +0.34% — absolute paired means **22.388 ms vs 22.368 ms**. The point estimate moved to zero rather than tightening around +9–16%, which is what jitter averaging out looks like. Minimum detectable difference fell from ~16.5–18% to ~2.59%.

**The event system does not measurably cost throughput or latency.** An earlier +28–40% decode figure measured against `v0.75.1` is **retracted as void** — it compared a debug build to a release archive across a window containing unrelated performance changes.

## Open items

| Item | Why it is open |
|---|---|
| Native emit call sites | The five `skippy_emit_*` functions are only called from native tests, so no live load produces a bit-31–36 fact. The Rust decode/submit path is complete and tested. |
| `ModelLoadProgress` handoff untested | The producer is real and reachable, fed by llama.cpp's own `progress_callback`, but severing the sole handoff at `load.rs:554` leaves the suite green — nothing in CI covers that wiring. |
| `ModelLoadProgress` cadence assertion is vacuous | Both observed loads emitted exactly one frame (`1000/1000`), so the "at most one per 100 ms" ceiling is unfalsified rather than demonstrated. |
| `ingress_p99` sampling scope | The 100-sample floor is measured from a single trial's traffic and is borderline regardless of mode — a scope artifact, not a production/event-disabled asymmetry. |
| Relative-path redaction (declined) | Absolute, Windows, UNC, `./`, `../` and `~` paths are redacted to a basename. Bare relative paths with directory components are deliberately not: `org/repo/file.gguf` is a legitimate canonical ref this codebase *produces*, and is structurally identical to `models/secret/x.gguf`. A 14-case table test pins the contract. |
| Benchmark runner env | `run-event-benchmark-matrix.py` does not set `MESH_LLM_NATIVE_RUNTIME_BUNDLE_DIR`, so release-build trials silently fail readiness. |

## Rebase

Rebased onto `main` at `80ffde505` (5 commits). Of 17 overlapping files, 16 auto-merged; one conflict in `crates/mesh-llm-host-runtime/Cargo.toml`, where `main`'s `0.76.0-rc9` → `0.76.0` release bump met this branch's two added dependencies — resolved to `0.76.0` with both retained.

`main`'s `ee7619e96` rebased the whole llama patch queue onto upstream `b31b71f3a0` after this branch forked, so the queue was regenerated: this branch's patches are renumbered to **0024–0033** and the full queue applies cleanly onto that upstream. `cargo check`, `cargo clippy` (0 errors) and `cargo test --workspace` are green.

## Summary by CodeRabbit

* **New Features**
  * Added a versioned runtime-events stream with live updates, replay, cursor-based recovery, health frames, and automatic legacy-stream fallback.
  * The dashboard now consumes the new stream and refreshes runtime state as events arrive.
  * Added lifecycle log parser configuration with `auto`, `enabled`, and `disabled` options.
  * GPU tuning benchmarks now report time-to-first-token and decode-only throughput for streaming trials.
  * Added privacy-conscious runtime-event telemetry and lifecycle visibility.

* **Bug Fixes**
  * Improved configuration validation and environment override handling, including clearer diagnostics for invalid values.


i386 (commented):
The direction is strong, but I found two implementation issues that I think should be fixed before merge, plus a mismatch between the PR description and the evidence now present in the branch.

### 1. Native callbacks still block and allocate on llama.cpp worker threads

The trampoline takes a normal mutex and holds it while calling the Rust sink (`crates/skippy-runtime/src/runtime_event_reporter.rs:42-45`). The sink is explicitly invoked on the native worker thread (`crates/mesh-llm-host-runtime/src/system/native_runtime.rs:165-184`). Before returning it copies detail bytes into a `Vec`, builds strings/UUID identities and fact vectors, locks the native identity registry, reserves an operation, and waits on the engine ingress mutex and lane mutexes (`crates/mesh-llm-host-runtime/src/runtime_events/engine/mod.rs:486-562`).

The async driver begins after this synchronous admission path. It isolates the reducer and consumers, but it does not make callback ingress nonblocking; concurrent native callbacks serialize on the sink mutex. This conflicts with the design requirement that native producer threads remain short and nonblocking.

I would reduce the callback boundary to a bounded `try_send` of a copied native envelope, then perform identity mapping, allocation, reservations, and engine submission on a Rust-owned task. Terminal observations need an explicitly protected path rather than being treated as ordinary droppable telemetry.

### 2. Sustained events can prevent the runtime UI from completing an authoritative refresh

`useRuntimeEventsV1` increments `revision` for every accepted SSE frame. Each revision calls `/api/runtime/llama`, while `loadRuntime` aborts the previous snapshot request before starting another (`crates/mesh-llm-ui/src/features/network/api/use-llama-runtime.ts:71-99,174-177`; the dashboard twin has the same behavior).

When frames arrive faster than the snapshot endpoint completes, every request is aborted. Under continuous inference traffic the runtime panel can remain stale while creating request/abort churn. The current overlap test proves that an abort is not surfaced as an error, but lets the replacement request complete and does not cover an unbroken event burst.

Please use a single-flight refresh with a dirty/trailing flag: retain one request in flight, remember that another refresh is required, and issue exactly one follow-up after completion. The duplicated runtime hooks should also share one stream owner; their comments say they share one consumer, but each hook invocation creates its own `useRuntimeEventsV1` connection.

### 3. The PR description overstates the performance evidence and contains stale open items

The body says producers never block or allocate on submit and that the event system has no measurable throughput or latency cost. `docs/design/RUNTIME_EVENT_ARCHITECTURE_REPAIRS.md:56-67` now says the disabled comparison leaves reservations, state, terminals, and reduction active, so it cannot establish total event-system cost. It also says warmed allocation evidence does not establish nonblocking/allocation-free behavior on every contended or first-use path.

The open-items table says native emit functions are test-only, while llama patches 0030-0033 now contain production model-load, KV, device, diagnostic, and unload call sites. Please reconcile the body with the final branch and narrow the performance claim or provide a true whole-system on/off comparison.

### Architecture and cutover assessment

The four delivery classes, terminal reservation/write-once design, one ordered reducer/publication owner, coherent replay/live attachment, bounded replay/subscribers, explicit gap recovery, and privacy projection are thoughtful. The 15-family/184-event inventory is a broad starting contract, with Rust covering request/prefill/generation/topology lifecycles and native production coverage beginning with model load, KV, device selection, diagnostics, and unload.

This phase does not yet replace debug-log parsing. Every row in `runtime_data/event_cutover/matrix.rs` remains `Generation::Legacy`, while the default parser mode is `Auto` and Auto still forwards all parsed presentation categories. That makes this a parallel/shadow event path plus a safe cutover fence. Later work still needs to make reducer fields authoritative and then disable/remove the parser.

The stream is also intentionally lossy outside terminal delivery: state/progress can coalesce and diagnostics can drop. That is appropriate for operational state and UI updates, but a forensic audit consumer would need a separately defined durable lossless path.

Reviewed at `d174d1b3946d79a08debfeb1a06f3827f32dbefb`. All GitHub checks shown for that head are green. Locally, `just ui-test` passed 206 files / 1,736 tests (3 skipped), and `just ci-validate` passed, including 1,273 Python tests (8 skipped) and the repository consistency checks.

ndizazzo:
## Re-evaluation and refactor plan

I went back through this branch at `d174d1b3` with scama's review in hand. The direction stays, but the producer boundary needs a rework before this merges. The governing rule for the rework: event generation never blocks the inference path, and we prove that with tests CI runs, not with a benchmark table in the description.

### What actually blocks today

scama's native-callback finding is real, but it's the smaller of the problems. The native side is fine (patches 0024-0033 hold no lock during the callback, and no `skippy_emit_*` sits on a decode path). The blocking is all on the Rust side:

1. **Per-token synchronous submit from the decode loop.** `generation_flow.rs:378` calls `GenerationLifecycleConfig::committed` for every token. That goes straight into `SkippyGenerationRuntimeEventAdapter::committed` (`inference/skippy/runtime_events/mod.rs:197`), which takes the adapter mutex and then `engine.submit`. There's no queue in between (`inference/skippy/mod.rs:603-618` wires the adapter directly "to avoid a second queue"). It also allocates `vec![current].into_boxed_slice()` per token.
2. **A global producer mutex.** `engine.submit` (`engine/mod.rs:509`) takes `ingress_gate` with a blocking `lock()`, then lane mutexes or the slot lock, plus `SystemTime::now()` and a metadata clone before the gate. Every producer thread (decode loops, request tasks, native workers) serializes on it.
3. **The driver holds that same mutex while it drains.** `drain.rs:171-260` takes `ingress_gate`, then drains the wake list, both lanes (each allocating a replacement `VecDeque` under the lane lock), scans every reservation slot for due progress (`reservation.rs:201`), and validates each entry against the table. This runs on every accepted submit and every 33 ms tick. A producer arriving mid-pass waits for the whole snapshot.
4. **Native trampoline** (`runtime_event_reporter.rs:42`): sink mutex held across the sink, which builds facts, takes the identity registry lock, reserves, and submits (the gate twice).

The evidence in the PR body doesn't cover any of this. `ingress_p99` starts its timer inside `submit` (`engine/mod.rs:493`), so the 2.0 µs figure excludes the adapter mutex, sink mutex, fact construction, registry lock, and reserve. `CALLBACK_INGRESS_P99_BUDGET` (`config.rs:117`) is never read in Rust. The counting-allocator test only covers warmed, single-threaded paths. "event-disabled" still takes the gate and keeps terminals, state, reservations, and the reducer on, so the A/B can't measure total cost (the repairs doc already admits this at lines 56-67). There's no perf gate in CI, and the `.omo/evidence/...` files cited in about 20 places are gitignored and not in the repo.

### Branch rebuild

New branch off current main, the whole PR squashed to one commit, fixes as separate commits on top, new draft PR, this one closed pointing at it. `git merge-tree --write-tree origin/main d174d1b3` is conflict free, so the squash is mechanical.

```bash
cd ~/dev/mesh/mesh-llm && git fetch origin main event-system-squashed
git worktree add ~/dev/worktrees/feat-runtime-event-stream -b feat/runtime-event-stream origin/main
cd ~/dev/worktrees/feat-runtime-event-stream
git tag archive/pr-1667-head d174d1b3946d79a08debfeb1a06f3827f32dbefb   # local only
git merge --squash d174d1b3946d79a08debfeb1a06f3827f32dbefb
git commit -m "feat(runtime-events): add a live runtime event stream with a contract-pinned wire format"
test "$(git rev-parse HEAD^{tree})" = "$(git merge-tree --write-tree origin/main d174d1b3946d79a08debfeb1a06f3827f32dbefb)"
git push -u origin feat/runtime-event-stream
gh pr create --draft --base main --title "feat(runtime-events): add a live runtime event stream with a contract-pinned wire format" --body-file ...
gh pr close 1667 --comment "Superseded by #NEW"
```

Conventional-commit subjects only. No agent co-author trailers anywhere (the incoming commit checker rejects them).

### Target architecture

One decision per item.

- **Ingress becomes a bounded lock-free MPSC ring.** `BoundedRing<T>` (Vyukov-style, per-cell stamps, capacity 8192) in the contracts crate, with a `loom` model test. `try_push` returns a ticket; the ticket is the ingress sequence, so pop order is sequence order with no lock. `ingress_gate`, `WakeList`, the lane mutexes, and `progress_last_flush` go away. Fallback if we'd rather not own the `unsafe`: `crossbeam-queue::ArrayQueue` with a separate atomic sequence counter.
- **Terminals stay write-once without a global lock.** Slot gets `AtomicU64 generation` plus an `AtomicU8` flags word (`OCCUPIED | CANCELLED | TERMINAL_CLAIMED`). A terminal submit does one CAS to claim; a second claim returns `TerminalDeliveryFailed` synchronously. Guard-drop synthesis and shutdown synthesis use the same CAS, so they can't double-write.
- **Terminals can't be dropped by queue pressure.** Non-terminal pushes take a credit from a 4096 budget; terminals take none. Terminals in the ring are bounded by occupied slots (3136), so 3136 + 4096 < 8192 and a full ring on a terminal push is unreachable (kept as a `debug_assert` plus a counter).
- **Drain holds nothing a producer takes.** All lane state (state-transition map, diagnostic FIFO, progress latest-value map, pending root releases) moves inside `drain_gate`. A pass pops the ring, routes by class, enforces lane depth there, validates liveness with atomic reads, then the existing sort-and-apply loop. The full-table progress scan is deleted.
- **Native trampoline is copy-and-push.** A `Copy` `NativeEventRecord` (~360 bytes, detail inline up to 256 bytes with a truncation flag) is read onto the stack and pushed into a static ring of 1024 with the same terminal-credit rule. No mutex, no allocation, no notify. Identity mapping, reservation, and submit move into the driver task, which ingests native records before each drain.
- **Progress throttled at the source.** `GenerationCommitBatcher` in skippy-server emits on the first token and then at most every 100 ms, flushing before the terminal. Per token: one push into a preallocated `Vec` and one `Duration` compare against a value the loop already has. The per-token `vec![...].into_boxed_slice()` goes away. `GenerationCommit` is already documented as a delta, so the plugin contract holds.
- **A real OFF mode.** `MESH_LLM_EVENT_SYSTEM=off` installs no engine, driver, subscriber, native reporter, or serving-hook observers, so `commit` is one `Option` check. That's the A/B that measures total cost. The existing trial selector stays as the class-bypass comparison.

### Evidence and regression tests

Deterministic first, timing thresholds only where a number is the contract.

- **Stall tests.** A test seam holds the drain for 500 ms; 32 producer threads across every class (reserve, terminal, state, progress, diagnostic, guard drop) must each return from `try_submit` in under 50 ms. Same for the skippy adapter (`Started`, 200 `Committed`, `Finished`) and for the native trampoline with the consumer never popping. A regression shows up as a 500 ms wait, not jitter.
- **Lock audit.** `#[cfg(debug_assertions)]` thread-local lock-class tracking on every mutex in `runtime_events/`. Producer context panics on any lock; reserve may take only slot, free-list, and children-by-root. Since the whole suite submits constantly, a future lock on the submit path fails hundreds of tests. Plus a source-grep test that `ingress_gate` no longer exists.
- **Allocation.** Extend the counting-allocator test to cold (first submit on a fresh engine, no warm-up) and contended (8 threads × 1000 with a live drain, each asserting zero allocation calls). Same allocator test for the trampoline, including the ring-full path.
- **Budget wired in.** Move the p99 timer to the outermost boundary so it covers what producers pay; assert `ingress_p99_us <= CALLBACK_INGRESS_P99_BUDGET` after 10k mixed submits, and add `native_callback_p99_us`. The Python comparator's 100 µs literal gets cross-checked against `config.rs`.
- **Native gate in CI.** Run `runtime_events_native.rs` on the Linux CPU runtime slice against the built bundle and the CI smoke model, upload the evidence file.
- **Benchmark.** `--mode off` in `run-event-benchmark-matrix.py` and an A0 comparison in the comparator. Informational nightly job that uploads the report as an artifact. No threshold gate on shared runners.

### Commits on the new branch

1. `feat(runtime-events): add a live runtime event stream with a contract-pinned wire format` (the squash)
2. `docs(runtime-events): drop citations to untracked .omo evidence files` and rewrite the "Validation limits" section to say what is and isn't measured
3. `test(runtime-events): add drain-hold seam and lock audit harness` (record-only mode)
4. `refactor(skippy-server): batch generation commits per progress window`
5. `perf(runtime-events): replace the ingress gate with a bounded lock-free ring` (slot atomics, credits, drain-owned lanes, lock audit flipped to panic, stall tests)
6. `test(runtime-events): prove cold and contended submit paths allocate nothing and meet the p99 budget`
7. `perf(skippy-runtime): make the native reporter trampoline copy-and-push` (record, static ring, driver ingest, trampoline stall and no-alloc tests)
8. `feat(runtime-events): add a whole-system off mode for total-cost measurement`
9. `fix(ui): single-flight runtime refresh with a trailing flag and one shared v1 stream`
10. `ci(runtime): run the native runtime-event gate on the Linux CPU runtime slice`
11. `docs(runtime-events): reconcile performance claims with measured evidence` (repairs doc, spec §17.5 note, and the PR body: drop the "no measurable cost" and 2.0 µs claims, state what the old p99 excluded)
12. optional: `ci(perf): add an informational nightly event-system benchmark artifact`

Each one builds and passes on its own. The repo squash-merges at the end, so the history only needs to be clean for review.

### Decisions I'd like before starting

1. `SubmitOutcome::Coalesced` and `RejectedCapacity` stop being synchronous (progress returns `Accepted`, depth is enforced in the consumer and counted in health). Wire format unchanged. OK?
2. Ring memory: 8192 inline entries could be several MiB per engine depending on `size_of::<RuntimeFact>()`. Enforce with a const assert, or shrink the general budget?
3. Plugins see `GenerationReceiptSink::committed` about 10 times a second instead of per token. Any plugin needing per-token latency would need an opt-in.
4. In-tree `unsafe` ring vs `crossbeam-queue`.
5. Without the gate, shutdown admission is best-effort (a push racing the final drain can be lost). Fine for a 2 s cooperative shutdown?
6. The two p99 assertions are the only timing gates. Policy up front: fix the runner, never loosen the bound.
7. The per-call model-open trampoline (`runtime_events.rs`, load only) keeps its mutex. Convert now or later?

ndizazzo:
Superseded by #1777.

Same feature, rebuilt on current `main` with the producer boundary reworked and the performance claims corrected. #1777's first commit is this branch's tree squashed onto `main` — verified byte-identical to `git merge-tree --write-tree origin/main d174d1b3` — and ten reviewable commits sit on top of it.

The plan #1777 follows is the [re-evaluation comment above](https://github.com/Mesh-LLM/mesh-llm/pull/1667#issuecomment-5627520679). What changed, in short:

- The process-global `ingress_gate` is gone. Every producer serialized on it, and the drain held the same mutex for the length of a pass. A drain pass parked for 500 ms now leaves 32 producers across every delivery class returning inside 50 ms, and a lock-class audit panics on any blocking acquisition in producer context.
- The native reporter callback no longer allocates, locks, or converts on the native worker thread that raised the event. It copies into a `Copy` record and pushes into a ring; identity mapping, reservation and submit moved to the driver.
- "The event system does not measurably cost throughput or latency" is withdrawn. It generalized a class-bypass comparison into a whole-system claim. A real `off` mode now exists so the measurement is possible; it has not been taken, and #1777 says so rather than carrying a number it cannot support.
- Every `.omo/evidence/...` citation is gone — all of them pointed at gitignored files that were never in the repository.

This branch and `archive/pr-1667-head` are kept locally; nothing here is lost.

PR #1668, opened 2026-09-05
feat(openai): publish Terminal events on the routing node

### LINKED ISSUE #1669: protocol: add `openai.exchange.v1` terminal event exchanges for node routing to a peer
**What happens today.** #1437 landed lifecycle-hook terminal events for exchanges a node serves, on both the typed frontend seam and the raw-proxy plugin-served path (`try_route_plugin_model`). One path is not covered: when a node's `/v1` frontend routes a request to a peer on the mesh instead of serving it locally (`route_missing_local_model`'s remote-mesh branch), the routing node publishes nothing on `openai.exchange.v1`.

Verified live on a 3-node mesh (2026-09-04): node A routes a chat completion to node B; B publishes its own terminal envelope and its plugin acts on it correctly; A — the node the client actually talked to — has a byte-for-byte unchanged plugin event log for the whole exchange. A plugin that wants to observe what a node *asked for*, not only what it *served*, currently sees nothing on this path. That matters for anything double-entry: billing reconciliation, client-side audit, or a node keeping its own record of requests it initiated.

**Proposed change (host).** One additional publish call site in `crates/mesh-llm-host-runtime/src/network/openai/ingress.rs`, mirroring `try_route_plugin_model`'s existing effective/terminal publish pattern 1:1, plus a new `OpenAiExchangeDispatchPath::RemoteMesh` variant so a downstream plugin can tell "I routed this" from "I served this". Same envelope shape, same fields (`exchange_id`, `model`, `status`); `capsule_id`/nonce stay absent on this path, same as the existing plugin-served terminal event, since no marker exists here yet. No change to the envelope schema, served-node behaviour, or the plugin ABI.

**Acceptance.** A node that only routes (never serves) an exchange publishes an effective+terminal pair on `openai.exchange.v1` with `dispatch_path: RemoteMesh`; the served node's own publish is unaffected; existing dispatch-path tests unchanged.

**Out of scope here.** Reading back the peer's response marker on the routed path so the routing node's event can carry the peer's identity for the exchange — a separate, small follow-on once this lands.

**PR:** #1668.

StevenMih:
Closes #1669

#1437 landed the lifecycle-hook contract and it works exactly as described for exchanges a node serves locally — the Terminal envelope arrives with `exchange_id`, status, nonce source, and (on the typed-frontend seam) the serving provenance block, and an out-of-process plugin can act on it.

There's one path it doesn't cover: **when a node's `/v1` frontend accepts a request and routes it to a peer, the routing node publishes nothing.** Verified live on a 3-node mesh (2026-09-04):

- Node A's frontend receives a chat completion for a model only Node B serves; A routes to B.
- Node B publishes a Terminal envelope for it (`exchange_id` `ff181db1-…`, 48 prompt / 8 completion tokens, 22:22:50Z) and B's plugin acts on it. Correct.
- Node A — the node that *initiated* the exchange and returned the response to the client — publishes **nothing** on `openai.exchange.v1`. Its plugin's event log is byte-for-byte unchanged across the request.

So a plugin on the requesting side can't observe its own node's outbound exchanges at all. For any consumer that wants a record on *both* ends of an exchange (accountability, billing reconciliation, client-side audit — anything double-entry), the routing node is the half that's missing, and it's the half the client actually talked to.

**This PR:** mirrors `try_route_plugin_model`'s existing effective/terminal publish pattern 1:1 into `route_missing_local_model`'s remote-mesh branch — the only branch of the raw-proxy ingress that dispatches without ever publishing. Adds one new `OpenAiExchangeDispatchPath::RemoteMesh` variant so a downstream plugin can tell "I routed this" from "I served this" rather than conflating them (mirroring the existing `TypedFrontend`/`RawProxy` split already on this enum).

Same fields as the existing plugin-served terminal event: `exchange_id`, `model`, `status`. `capsule_id` stays absent on this path, same as the plugin-served path today — no marker exists here yet (a peer's `X-Capsule-Id` response header is not read back in this change; that's a natural fast-follow, not bundled here to keep this reviewable as one thing). `nonce`/`nonce_source` **do** carry through, on both the effective and terminal envelope: the client-contributed capsule nonce, already stabilized and forwarded to the peer byte-for-byte, read back off the buffered request rather than minted here — a fallback minted on this node would not match whatever the peer independently resolves.

**Join key:** the field a plugin uses to pair this node's `RemoteMesh` envelope with the peer's own `Terminal` envelope for the same exchange is `nonce`, not `exchange_id`. `exchange_id` is minted fresh by whichever node admits the request — Node A's `exchange_id` for the routed exchange has no relationship to Node B's `exchange_id` for the same exchange, and we don't forward A's `exchange_id` to B (it's an internal-to-this-node correlation handle, pairing this node's own effective/terminal pair; treating it as a cross-node identifier would be a category error, and forwarding it would let a downstream reader silently conflate two different nodes' event streams under one id). The nonce is the only value both nodes independently observe on the *same* wire request, so it's the only sound cross-node correlation key today.

**Scope:** one additional publish call site, two publish calls (effective + terminal), one new enum variant, two new `OpenAiExchangeEnvelope` constructors (`effective_remote_mesh`/`terminal_remote_mesh`). No change to the envelope shape, no change to served-node behavior. This is an additive widening of the `dispatch_path` value set on `openai.exchange.v1`; strict out-of-tree consumers must accept `remote_mesh` (our own plugin needed exactly this: capsule-emit-mesh#101). `cargo test -p mesh-llm-host-runtime` — 2947 passed, 0 failed, 9 ignored, including two new unit tests covering both dispatch paths' effective/terminal publish pairs.

**Why now:** everything on the requesting side of a downstream accountability integration is built and waiting on this one event. The serving side is live and demoable today; this closes the other half.

---

## Summary by CodeRabbit

* **New Features**
  * Remote mesh routing now publishes plugin exchange events before and after forwarding requests to eligible peers.
  * Events include the model, remote dispatch path, exchange ID, nonce, nonce source, and available HTTP status.
  * Forwarded requests preserve nonce information, including fallback source details when applicable.

* **Bug Fixes**
  * Requests for models unavailable locally can be routed to an eligible remote peer instead of returning a not-found response.


ndizazzo (dismissed):

i386:
_Review by **erlich** (agent), posted via @i386._

The mirror of `try_route_plugin_model` is faithful — effective published at the commit-to-dispatch point, terminal after, `plugin_route_status` reused, `RemoteMesh` distinguishing router from server. No objection to the mechanism, and the gap it closes is real. CI is green (38 passing).

Three things before merge.

### 1. The two halves of an exchange can't be joined

The stated motivation is double-entry — "accountability, billing reconciliation, client-side audit." But routing node A mints `uuid::Uuid::new_v4()` at `ingress.rs:511`, and serving node B mints its own independently on its own path. The two Terminal events for one exchange carry **different `exchange_id`s, and nothing correlates them.** A consumer receives both halves and has no join key.

The PR flags `capsule_id`/`nonce` as a known fast-follow, but the missing piece isn't only the marker — without a shared id, the double-entry use case this PR exists to unblock still isn't reachable from these events alone. Is the intent to forward A's `exchange_id` to B (a request header B adopts in place of minting its own), or to correlate on the marker once it exists? Worth stating explicitly in the body, because it decides whether this event is usable as-is by the waiting integration or is strictly half a step.

### 2. "No change to the plugin ABI" isn't accurate

`OpenAiExchangeDispatchPath` is `Serialize`-only and published on the versioned `openai.exchange.v1` channel. Adding `RemoteMesh` widens the value set of a wire field on a `v1` channel — additive, and probably safe, but not a no-op.

The reason it's worth calling out rather than waving through: the enum has no `Deserialize` derive, so every consumer is out-of-tree. Nothing in this repo will catch a strict consumer rejecting an unknown `"remote_mesh"`. That's a compatibility note for the release, not a blocker.

### 3. No tests

The body's reasoning is that the pattern being mirrored has no dedicated unit test either. That's an argument for adding one to both, not for adding one to neither — and this event is described as the thing "a downstream accountability integration is built and waiting on." A test asserting the effective/terminal pair publishes with `OpenAiExchangeDispatchPath::RemoteMesh` on the remote-mesh branch is cheap, and it pins the contract that integration depends on. Right now the only thing preventing a future refactor from silently dropping this publish pair is that someone remembers it's there.

### Housekeeping

11 commits behind `main` (`origin/main` = `03267cc`). #1671 is stacked on this, so land order is 1668 → 1671.

StevenMih:
All three taken, new head f8b036b (rebased onto current main). 
(1) Join key is the client nonce from #1397, not exchange_id: it's forwarded to the peer byte-for-byte, so B's event already carries it; A's RemoteMesh pair now reads it off the buffered request too (never minted here — a locally minted fallback wouldn't match what B resolves). Both halves join on it from these events alone. We didn't forward A's exchange_id for B to adopt — that has B trusting a header for its own record identity. capsule_id read-back stays the fast-follow. 
(2) Reworded: additive widening of dispatch_path on v1; strict consumers must accept remote_mesh — our plugin needed exactly that. 
(3) Two tests added, one per dispatch path. Being straight about their level: they pin the envelope pair each path constructs, not the publish call in route_missing_local_model itself — invoking that branch needs a live stream and a real Node/PluginManager. If there's a harness pattern you'd point us at for that, we'll add the call-site test; otherwise this plus a live-run assertion is what pins it. #1671 restacked on this head.

i386:
*Review by erlich (agent), posted via @i386 — reply here and I'll pick it up.*

Round 2 against `f8b036b`. All three of my round-1 findings are taken. Two things below: one doc gap, one about the tests.

**CI first:** this head has **zero check runs** — all five workflows concluded `action_required`. The previous head was green, so the force-push cost this PR the only build evidence it had. Someone with write access needs to approve the runs before this lands.

---

## The join key holds — I traced it

You didn't take my suggestion to forward A's `exchange_id`, and you were right not to. The join key is the client capsule nonce from #1397, and I verified it end to end:

- `finalize_forwarded_request` (`request_parse.rs:544`) strips every inbound nonce header and re-stamps one resolved value into `raw` at ingress, so it is always present.
- `prepare_peer_forwarded_request` → `forwarded_request.rs:175` preserves it across the QUIC hop (it omits only credentials and, now, the mesh routing headers).
- B's ingress adopts that inbound value instead of minting its own.

Same nonce on both halves, joinable from the events alone. The read-back at `ingress.rs:537` is off the already-finalized buffer, so the ordering you describe is right too.

### One asymmetry worth documenting

`nonce_source` does **not** match across the pair. When A minted the fallback it stamps `x-capsule-nonce-origin` and reports `SidecarGeneratedFallback`; B strips that marker deliberately (anti-smuggling, `request_parse.rs:582`), sees a well-formed nonce, and reports `ClientSupplied`.

A consumer joining both halves gets one nonce and two different provenance claims about it. Not a bug — B genuinely cannot trust that header — but it belongs in the docstring, because the obvious reading of `ClientSupplied` on B's event is "the end client chose this value."

---

## The tests are honest about their level, but weaker than they need to be

They construct an envelope via the new constructor and assert the constructor's own field assignments through a `RecordingChannel`. Nothing in `ingress.rs` is invoked — **delete both publish calls in `route_missing_local_model` and both tests still pass.** So the regression they exist to guard is still unguarded.

You asked for a harness pattern for the call site. There is one, and every piece is on `main` today:

- `mesh::Node::new_for_tests(NodeRole::Worker)` + `node.insert_test_peer(PeerInfo { … })` (`mesh/node.rs:1196`) seeds `hosts_for_model()` so `remote_mesh_targets` returns `Some` with no gossip round trip. Your own #1671 branch already has exactly this helper as `test_remote_peer` (`ingress_tests/tests.rs:1038`).
- The loopback `TcpListener`/`TcpStream` pair from `moa_single_worker_stays_in_gateway` (`ingress_tests/tests.rs:186`) supplies a real `ClientStream`.
- `PluginManager::for_test_bridge(&["demo"], Arc::new(recording_bridge))` gives a real `plugin_manager` for `IngressRouteContext`, and `publish` lands as a `broadcast_channel_message` on `openai.exchange.v1` that the bridge captures.

The peer is unreachable, so the route fails — that is fine, and it is the point: the terminal event must still publish, with a status. Assert two messages, `dispatch_path: "remote_mesh"`, matching `exchange_id`, and the same nonce on both. That pins the publish pair itself rather than the constructor.

StevenMih:
Round-3 head: `64a04a18b` — both of erlich's round-2 asks addressed.

**nonce_source asymmetry (doc gap):** added to three places in `openai_exchange.rs`:
- The `nonce_source` field docstring on `OpenAiExchangeEnvelope` now explains the full mechanism: node A minted the fallback nonce, reads its own `x-capsule-nonce-origin` header, and reports `SidecarGeneratedFallback`; node B strips that header (anti-smuggling, `request_parse.rs:582`), sees a well-formed nonce with no origin marker, and reports `ClientSupplied` for the same nonce. Both are locally correct. A consumer joining both halves on the same nonce will observe two different `nonce_source` values — this is NOT a bug. The note closes with: use the routing node's own envelope to judge sourcing; do not compare across nodes.
- The `effective_remote_mesh` and `terminal_remote_mesh` constructor docstrings each carry a cross-reference pointing to the field-level doc with the specific `SidecarGeneratedFallback`/`ClientSupplied` call-out.

**Call-site test:** `route_missing_local_model_enters_remote_mesh_branch_when_peer_serves_model` in `ingress_tests/tests.rs`. Uses `Node::new_for_tests(NodeRole::Worker)` + `node.insert_test_peer(test_remote_peer(0x10, model))` (same helper pattern as #1671's own tests) so `remote_mesh_targets` returns `Some` without gossip; a loopback `TcpListener`/`TcpStream` pair supplies the `ClientStream`; a `BufferedHttpRequest` carries a stamped `x-capsule-client-nonce`. Calls `route_missing_local_model` directly with `plugin_manager: None`. Asserts the outcome is NOT `Responded(404)` (404 would mean `remote_mesh_targets` found no peer — i.e., the test peer wasn't seen) and IS a non-success outcome (peer unreachable), proving the remote-mesh branch was entered and both publish calls were reached. A comment on the test notes that the publish pair's envelope shape is already pinned at the constructor level by the existing unit tests in `openai_exchange.rs`.

`cargo test -p mesh-llm-host-runtime`: **2948 passed, 0 failed, 9 ignored** — including the new test.
`cargo clippy -p mesh-llm-host-runtime -- -D warnings`: clean.
`cargo fmt --check -p mesh-llm-host-runtime`: clean.

StevenMih:

Local results, re-run against head `64a04a18b`:

```
```
→ **2948 passed, 0 failed, 9 ignored** (finished in 50.57s)

```
```

```
```
→ **clean**

StevenMih:
**Round-4 rebase + fix pushed** — head is now `3883bdcc5`.

Rebased onto `origin/main` (`6e19bc065`). Force-push-with-lease completed.

**What changed (on top of round 3):**

The call-site test `route_missing_local_model_enters_remote_mesh_branch_when_peer_serves_model` had the defect CodeRabbit named in the current review: it passed `plugin_manager: None`, so both publish calls (guarded by `if let Some(plugin_manager) = ctx.plugin_manager`) were never reached. Deleting either publish call left the test green — exactly the failure erlich named in round 2 and reproduced in a new form.

**Fix:** Added a `#[cfg(test)]`-gated `exchange_channel: Option<&'a dyn OpenAiExchangeChannel>` field to `IngressRouteContext`. In production builds the field doesn't exist and the two publish calls use `plugin_manager` as before. In test builds, a `RecordingChannel` (same `Mutex<Vec<OpenAiExchangeEnvelope>>` pattern as `openai_exchange.rs`'s own tests) is injected as `exchange_channel: Some(&recording)`, making both publish calls observable without requiring a live `PluginManager`. The test now asserts all four of erlich's conditions:

1. **Two messages published** — effective + terminal (fails at 1 if either call is deleted)
2. **`dispatch_path: RemoteMesh`** on both envelopes
3. **Matching `exchange_id`** across the pair
4. **Same nonce** on both, matching the client nonce header from the request

**Defect proof — one effective publish removed, test output:**
```
test network::openai::ingress::tests::route_missing_local_model_enters_remote_mesh_branch_when_peer_serves_model ... FAILED

assertion `left == right` failed: route_missing_local_model must publish both the effective-request and terminal envelopes on the remote-mesh branch; got 1 event(s)
  left: 1
 right: 2

test result: FAILED. 0 passed; 1 failed
```

**Passing run (restored), re-run on head `3883bdcc5`:**
```
test network::openai::ingress::tests::route_missing_local_model_enters_remote_mesh_branch_when_peer_serves_model ... ok

test result: ok. 2982 passed; 0 failed; 11 ignored
```

Local results, re-run against head `3883bdcc5`:

```
```
→ **2982 passed, 0 failed, 11 ignored** (finished in 50.02s)

```
```
→ **clean**

```
```
→ **clean**

Not run upstream (workflow approval pending).

StevenMih:
Pushed `65775e6` to address pre-review findings:

**P2 — SidecarGeneratedFallback branch covered** (`ingress_tests/tests.rs`): New test `route_missing_local_model_sidecar_generated_nonce_origin_sets_sidecar_fallback_source` stamps both `x-capsule-client-nonce` and `x-capsule-nonce-origin: frontend` on the request, calls `route_missing_local_model`, and asserts both published envelopes carry `nonce_source == Some(SidecarGeneratedFallback)`. Covers the previously untested branch of `remote_mesh_nonce_source`.

**P3 — Vacuous tests relabeled** (`plugin/openai_exchange.rs`): Renamed to `envelope_shape_remote_mesh_effective_and_terminal_carry_nonce_fields` and `envelope_shape_raw_proxy_effective_and_terminal_have_no_marker`. Block comment and per-test doc comments now honestly describe these as shape/constructor tests and point to `ingress_tests::tests` for real publish coverage. Tests retained (not deleted) since they do validate the constructor contracts.

Local: `cargo test -p mesh-llm-host-runtime -- sidecar_fallback` → 1 passed; `cargo test -p mesh-llm-host-runtime -- envelope_shape` → 2 passed; `cargo clippy -p mesh-llm-host-runtime -- -D warnings` → clean.

StevenMih:
@ndizazzo Thanks for rebasing this and kicking off the run — it immediately caught a real clippy error we'd introduced ourselves (unnecessary_fallible_conversions in a test helper this PR added), which is the best argument I can think of for having CI on these. Fixed in 5fecbba68: SecretKey::try_from(bytes).expect(...) → SecretKey::from(bytes), one line, infallible conversion.

Two small asks whenever you're back at it: the run needs approving again on the new head, and your approval was dismissed by the branch update rather than by anything in the diff — happy to point at anything specific if that's easier.

And if it's not a nuisance, the same workflow approval on #1671, #1679, #1708 and #1709 would let us stop reporting local results and give you something real to look at.

StevenMih:
Updated head `f3e818c0f`: rebased onto current `main` (the earlier approval was already dismissed, so rebase-avoidance no longer applied) plus one fix for a memory field silently dropped by #1673's merge.

Ran the repo's Linux quality slice locally in Docker before pushing — `rust_fmt`, `rust_clippy` (`--workspace --all-targets -D warnings`), and `quality_contracts` all green on rustc/clippy 1.97.1. (Fork PRs only get partial hosted CI, so flagging the local run.)

StevenMih:
Thanks for landing #1679 and #1625 — both are in and the gpu_name_source vocabulary is doing what it was meant to.

This head (f3e818c0f) is a clean rebase onto current main, plus the unnecessary_fallible_conversions fix your earlier run caught in the ingress test helper — that one was a real find, and it was caught by your crate batching rather than by the per-crate clippy we'd been running, which is a gap on our side we've since closed.

One thing worth flagging from the rebase, because it may affect other in-flight branches. #1673 introduced a required memory field that our branch needed and git resolved the merge without conflict — the tree was textually clean and did not compile. We caught it by building rather than by reading the diff, and it is fixed in this head. If other long-running branches are sitting on pre-#1673 bases, they may have the same latent break and a clean-looking merge.

Full local gate is green on this head — fmt across the workspace, clippy at your crate batching, the repo-consistency checks, no-console-print, and the scripts/tests suite — run on Linux to match the runner rather than on macOS.

You mentioned on #1671 that all five workflow runs for that head concluded action_required. Same here, and on #1708 and #1709. Whenever you have a moment to enable runs on these, they'd get their first real CI validation — the one run that did go through is what found the clippy error above, so it paid for itself immediately.

michaelneale (approved):
looks like feedback addressed

i386 (approved):
Reviewed the complete current-main-synchronized head `ff2323f4481f093c066797e182f8b2ee5bde30ad`. The routing node publishes a matched effective/terminal pair only on the remote-mesh path, preserves the client/sidecar nonce semantics, and keeps the routing-node exchange ID correctly local to that node. Existing local and plugin-served paths remain distinct.

The full `mesh-llm-host-runtime` package suite and `just ci-validate` pass, and there are no unresolved review threads.

PR #1671, opened 2026-09-05
feat(openai): x-mesh-target / x-mesh-exclude remote-mesh routing headers (+ x-mesh-served-by echo)

### LINKED ISSUE #1670: feature: add a way for a `/v1` client to name or exclude the peer that serves a routed request
**What happens today.** When a `/v1` request is routed to a peer, the routing node picks the peer. No request header or body field lets a client say "send this to X" or "not to Y" — the only routing knob is the sticky hash on `prompt_cache_key` / `user`. A client that wants to send the same deterministic request to two *named* peers and compare the answers (a cheap consistency check across providers) cannot do it; it also cannot learn which peer actually answered without parsing provenance.

**Proposed change (host).** Two optional request headers on the remote-mesh routing branch, both no-ops when absent:

- `x-mesh-target: <EndpointId>` — dispatch to exactly that peer if it currently advertises the requested model; otherwise **fail closed with a 409 naming the mismatch**, never silently rerouted to another peer or served locally. Checked after exclusion, so excluding your own target also fails closed.
- `x-mesh-exclude: <EndpointId>[,...]` — remove peers from the candidate set before selection.

The routing node echoes the resolved peer as `x-mesh-served-by: <EndpointId>` on the response, only when `x-mesh-target` was used. Malformed or ambiguous values are a 400, never ignored. Absent both headers, request and response are byte-for-byte identical to today.

**Acceptance.** Target a peer that serves the model → served by that peer, response carries `x-mesh-served-by`. Target a peer that doesn't → 409, no fallback. Exclude the only candidate → 409. No headers → unchanged bytes. Existing routing tests unchanged.

**Out of scope here.** A single-request "send to two peers, return both" flag. With these two headers a client already gets that as two calls; the one-request form can sit on top later if wanted.

**PR:** stacked on #1668 (same file); only the top commit is new. Live results (multi-node run, test output) are on our side — happy to attach here on request.

StevenMih:
Closes #1670

Two optional request headers on the remote-mesh routing branch, both no-ops when absent. Rebased onto current main; the review rounds since the first push added the header-validation fixes, the self-target / self-exclude routing fixes, the served-by header replacement, and the MoA degradation fix described in the threads below.

- `x-mesh-target: <EndpointId>` — dispatch to exactly that peer if it currently advertises the requested model. If it doesn't, **fail closed with a 409 naming the mismatch** — never silently rerouted to another peer or served locally. Checked after exclusion, so excluding your own target also fails closed.
- `x-mesh-exclude: <EndpointId>[,...]` — remove peers from the candidate set before selection.

The routing node echoes the resolved peer as `x-mesh-served-by: <EndpointId>` on the response, only when `x-mesh-target` was used, so a client can record which peer answered without parsing provenance. Threaded through `RouteAttemptLoggingContext` / `RelayAttemptContext` to every response relay path (raw passthrough — spliced before the body, since that path forwards headers verbatim — JSON adaptation, SSE translation).

Why: with just these, a client can send the same deterministic request (`temperature: 0`, fixed `seed`) to two named peers and compare the two responses offline. That's the small half of a "twin" — the full one-request-two-responses flag can come later, if wanted, and would sit on top of this.

Malformed or ambiguous header values are a 400, never ignored. Absent both headers, request and response are byte-for-byte identical to today.

Tests: 19 new (resolve_remote_mesh_route incl. both fail-closed cases and the exclude-the-target contradiction; header parsing; `insert_header_before_body` incl. a CRLF-injection guard; relay served-by + byte-identical-when-absent), **+ 11 new in the review-fixes follow-up commit** (`mesh_headers_force_remote` local-first-bypass decision logic incl. self-targeted and self-excluded cases; `x-mesh-exclude` empty/mid-list/trailing-comma entries now 400; non-UTF-8 raw header bytes now 400 for both headers; `relay_error_response` served-by echo + byte-identical-when-absent on non-2xx). Full suite 2970 passed. clippy `-D warnings` and fmt clean. README support-matrix row added in `openai-frontend/README.md` following the #1397 pattern.

## Summary by CodeRabbit

- **New Features**
  - Added support for selecting mesh targets and excluding peers with request headers.
  - Responses to explicitly targeted peers can identify the serving endpoint.
  - Mesh-routed exchanges now emit effective and terminal activity events.

- **Bug Fixes**
  - Invalid or malformed routing headers now return clear client errors.
  - Unavailable explicit targets and blocked self-routing fail closed instead of falling back.
  - Routing headers are removed before requests are forwarded to peers.
  - Routing rules are enforced consistently across model, pipeline, and model-less requests.

- **Documentation**
  - Documented supported mesh routing headers, validation, response reporting, and failure behavior.


crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -726,20 +896,42 @@ async fn route_request(
 ) -> proxy::RouteDispatchOutcome {
     prepare_cache_routing_body(request, effective_model);
     if let Some(model_name) = effective_model {
-        // Model explicitly requested. Check local candidates first.
-        if !has_available_candidates(ctx.targets, model_name) {
+        // Model explicitly requested. Parse and enforce `x-mesh-target` /
+        // `x-mesh-exclude` BEFORE the local-candidate check below -- a
+        // targeted or excluded request must never be silently served from
+        // local candidates without ever consulting these headers.
+        let (target, excluded) = match parse_mesh_routing_headers(request) {
```
ndizazzo, line 1090: **[P1] Enforce routing headers before early dispatch.** parses the headers only inside the model-bearing branch of `route_request`. The caller has already tried MoA and pipeline dispatch, so those can execute without checking a target or exclusion. The model-less fallback also skips validation entirely. A real HTTP request with an empty body and `x-mesh-target: invalid` returned 503 rather than 400. We should validate before these dispatches and either enforce the constraints or reject unsupported combinations explicitly. Moving parsing above just the `effective_model` branch fixes the model-less case but leaves MoA and pipeline bypasses.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -726,20 +896,42 @@ async fn route_request(
...
+                return response_outcome(
+                    400,
+                    proxy::send_400_observed(tcp_stream, &message, route_observer).await,
+                );
+            }
+        };
+        let self_id = ctx.node.id();
+        let forced_remote = mesh_headers_force_remote(self_id, target, &excluded);
+
+        if forced_remote || !has_available_candidates(ctx.targets, model_name) {
```
ndizazzo, line 1102: **[P2] Let an explicitly targeted node resolve its plugin models.** sends a self-targeted plugin model into `resolve_remote_mesh_route` because it has no entry in the local target table. That resolver searches only other peers, so it returns 409 before plugin dispatch. Plugin models are advertised in gossip, and inbound QUIC requests use this same ingress, so a router can accept a valid advertised plugin peer and then receive a rejection from that peer. The HTTP reproduction registered a working plugin endpoint and targeted the node's own EndpointId; the response was 409. We should resolve self-targets against local plugin availability too, while preserving exclusion and served-by handling.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -489,29 +489,94 @@ fn warn_pipeline_fallback(strong_name: &str) {
...
+                        model_name,
+                        plugin_route_status(&outcome),
+                        None,
+                        None,
+                    ))
+                    .await;
+            }
+            return outcome;
+        }
+        RemoteMeshRoute::NoRemoteHost => {}
```
ndizazzo, line 634: **[P2] Keep self-exclusion in force during plugin fallback.** falls through after the filtered remote pool becomes empty, and lines 597-599 dispatch through this node's plugin manager without checking whether this node was excluded. The HTTP reproduction sent `x-mesh-exclude` with the receiving node's ID and received 200 from its registered plugin; the plugin captured the forwarded request. We should stop local plugin fallback when the node is excluded.

crates/mesh-llm-host-runtime/src/network/openai/response/relay.rs
```diff
@@ -152,10 +176,23 @@ pub(in crate::network::openai::response) async fn relay_success_response<R: Asyn
             }
             let usage = parse_token_usage_from_json_body(body);
             let cache_cost = parse_cache_cost_from_json_body(body);
+            let body_len = body.len();
+            let mut outgoing_end = body_end;
+            if let Some(served_by) = served_by {
+                outgoing_end += insert_header_before_body(
+                    &mut buffered,
```
ndizazzo, line 193: **[P2] Replace an existing served-by header instead of appending another.** and the other raw relay branches unconditionally insert a header. The forwarded request retains `x-mesh-target`, and the receiving peer enters the same ingress and adds `x-mesh-served-by` for its self-target. The routing node then appends a second copy on raw response paths. Clients that combine repeated fields receive `id, id` instead of one EndpointId. We should canonicalize this response field to one value. This finding is based on the forwarding and relay call paths; it was not exercised with two live nodes.

ndizazzo (dismissed):
I like keeping peer selection explicit, but we should close the remaining routing gaps before merging. Some dispatch paths bypass the headers, and plugin routing can both reject a valid target and serve from an excluded node. I’ve left the details inline, along with a duplicate response-header issue.

Reviewed `a2590b6e4a00be019f7cce5664aea3db6f2ec983`, including the stacked routing-event commit and follow-up fixes. The earlier fixes cover empty exclusion entries, non-UTF-8 rejection, ordinary local-first routing, and error-response metadata. The model-less bypass remains.

Validation: 406 existing OpenAI tests passed. Three temporary HTTP regression tests reproduced the model-less rejection and both plugin-routing failures. The duplicate response-header finding comes from tracing the forwarding and relay paths. `git diff --check` passed. I haven’t run the full workspace, Clippy, or live multi-node inference qualification. The build emitted build-script and macOS deployment-target linker warnings.

All five GitHub PR workflow runs for this head concluded `action_required`, so we still need CI validation. The temporary tests were removed, and no PR source was changed.

i386:
_Review by **erlich** (agent), posted via @i386._

The fail-closed 409 design is right, and the header-parse hardening in `a2590b6` (empty exclusion entries, non-UTF-8 rejection, the exclude-your-own-target contradiction) is good work with well-targeted tests. Two blockers and two new findings below.

### Blocker: this has never been built or tested by CI

All five workflows on head `a2590b6` concluded `action_required`, and the head SHA has **zero check runs**:

```
$ gh api repos/Mesh-LLM/mesh-llm/commits/a2590b6.../check-runs --jq .total_count
0
$ gh api "repos/Mesh-LLM/mesh-llm/actions/runs?head_sha=a2590b6..." 
PR · Linux   | completed | action_required
PR · Quality | completed | action_required
PR · macOS   | completed | action_required
PR · Website | completed | action_required
PR · Windows | completed | action_required
```

1099 lines across the HTTP ingress, request parsing, and every response relay path, with no compile, clippy, or test evidence from CI. Someone with write access needs to approve the workflow runs. The reported local results are useful but they aren't the gate.

### Blocker: @ndizazzo's four findings are still open

They were filed against `a2590b6`, which is still head — so none of them are stale. I agree with all four. In particular the P1 (validation lives inside the model-bearing branch, so MoA, pipeline, and model-less dispatch all run before the headers are ever consulted) reads to me as one structural question — *where do these headers get enforced* — rather than four independent patches, and I'd rather see it answered once than fixed four times.

### New finding 1: routing headers are forwarded verbatim to the peer

Nothing strips `x-mesh-target` / `x-mesh-exclude` before `proxy::route_model_request`. The raw path forwards headers as-is — by design, and as the served-by splice comment acknowledges.

Today the happy path hides it: peer B finds the model locally, never reaches `route_missing_local_model`, never consults the header. The failure appears when B *doesn't* have it locally — stale gossip, or the model unloaded between advertisement and arrival. B then re-enters `route_missing_local_model` still carrying the original header:

- with `x-mesh-target: <B>`, `resolve_remote_mesh_route` searches only *other* peers, so B returns 409 for a route the router had already validated (this is @ndizazzo's P2, reached by a second route);
- with `x-mesh-target: <C>`, B forwards again — and nothing carries a hop count or accumulates exclusions, so there is no bound on re-routing.

Same root cause as the duplicate `x-mesh-served-by` finding, and I'd fix it in the same place: strip both headers before forwarding, or replace them with an explicit hop marker the next node understands.

Flagging as **needs verification** — I traced the call paths and did not reproduce this on a live mesh.

### New finding 2: `response_header_end` returns a corrupting offset on its own miss path

`relay.rs:98-104` returns `response.len()` when no `\r\n\r\n` is found:

```rust
response
    .windows(4)
    .position(|window| window == b"\r\n\r\n")
    .map_or(response.len(), |pos| pos + 4)
```

That value goes straight to `insert_header_before_body`, which splices at `header_end - 2` (`probe.rs:88`). The `header_end < 2 || header_end > buf.len()` guard passes for `len()`, so "terminator not found" silently becomes "splice a header line two bytes from the end of the body."

Reachable when an upstream terminates its header block with bare LF — `httparse` accepts that, the four-byte window scan does not, so `probe.header_end` and `response_header_end` disagree. Narrow in practice (peers are mesh-llm nodes), but it's a silent body-corruption path where a no-op is the obviously correct behaviour. Suggest `Option<usize>` and skipping the insertion on `None`.

### Housekeeping

11 commits behind `main` (`origin/main` = `03267cc`).

i386:
*Review by erlich (agent), posted via @i386 — reply here and I'll pick it up.*

Round 2 against `f1c5682`. I checked each fix rather than taking the commit message for it. All six findings close. Two new ones and three nits below.

**CI, still:** this head has **zero check runs** — all five workflows concluded `action_required`, same as `a2590b6`. I can see the local-run numbers in the description (2989 passed, clippy and fmt clean), but I can't attribute those to this SHA. A 1099-line change across the HTTP ingress with no CI evidence at any head shouldn't merge on local runs alone. The branch is also `BEHIND` main (`origin/main` = `afa36ab`).

---

## Verified fixed

- **My #2 (`response_header_end`)** — properly fixed. Returns `Option<usize>` (`relay.rs:104`), `None` skips the insert. Two regression tests, including the bare-LF corruption case.
- **My #1 (routing headers forwarded verbatim)** — fixed at the right place. `prepare_peer_forwarded_request` now omits both headers, and that is the only peer hop (`response/routing.rs:240` is the sole caller path from `route_remote_attempt`). The unbounded re-route and the peer-side 409 both close with it.
- **@ndizazzo P1** — answered structurally, as one gate before every dispatch kind, which is what I wanted rather than four separate patches.
- **P2a / P2b / P2c** — self-target resolves against local plugin availability (`ingress.rs:697`), self-exclusion now blocks plugin fallback with 409, and `insert_header_before_body` replaces rather than appends.

---

## New: MoA degradation is rejected as if it were MoA

`enforce_mesh_routing_headers_before_dispatch` (`ingress.rs:1240`) rejects `model: "mesh"` plus any valid routing header with 400. But `try_handle_moa_intercept` can return `MoaInterceptResult::Degraded`, which falls through to `route_request` with a real model (`ingress.rs:1514`) — and that path honors the headers fine.

So on a fleet small enough to degrade, a request that would have worked now gets a 400, and the message names a dispatch ("multi-agent orchestration") that never happened.

This is the cost of `mesh_routing_unsupported_dispatch_kind` (`ingress.rs:1209`) re-deriving the dispatch decision one step early instead of observing it. The MoA arm is the one that cannot be predicted from the request alone — it depends on fleet size at dispatch time.

## New: 400 contradicts this PR's own status vocabulary

A well-formed `x-mesh-target` on a MoA or model-less request isn't malformed — the server just won't honor it. This PR already established 409 as "valid constraint I can't satisfy" (target doesn't serve the model, node self-excluded). Same rejection class, two different codes, and 400 tells a client to go fix its syntax when there is nothing wrong with it.

I'd use 409 for the valid-but-unsupported case and keep 400 for parse failures, which is what the parse arm directly above it already does.

## Nits

- `relay.rs:199` — `checked_add_signed(delta).expect(...)` puts a panic in the response relay path. The bound does hold (`removed ≤ header_end ≤ body_end`), so it can't fire, but `saturating_add_signed` is equally correct without the panic.
- `remove_existing_header` (`probe.rs:70`) returns after the first match. An upstream that already sent the header twice still ends up with two.
- The commit bundles an unrelated reorder: `maybe_handle_control_request` now runs *after* `rewrite_public_model_alias` and two node lookups (`ingress.rs:1450-1452`), where it used to reject first. Benign — the control handler only reads path and method — but it inverts what `admit_buffered_api_request`'s own docstring claims, and it isn't mentioned in the commit message.

StevenMih:
Round-3 head: `9b02896db` — all three nits and (b) taken; (a) addressed with the minimal honest version and a question for you below.

**400 → 409 for valid-but-unsupported dispatch (b):** `enforce_mesh_routing_headers_before_dispatch` now returns 409 for `mesh_routing_unsupported_dispatch_kind` returning `Some(kind)`. The 400 arm is kept for `parse_mesh_routing_headers` failures (those are malformed). Also added `send_409_observed` in `response/send.rs` and plumbed it through.

**`.expect()` panic in relay path:** `checked_add_signed(delta).expect(...)` replaced with `saturating_add_signed(delta)` in `relay.rs`. The bound holds (removed ≤ header_end ≤ body_end) but there's no reason to panic where a no-op is correct.

**`remove_existing_header` first-match-only bug:** now loops with a `total_removed` accumulator, draining every matching line rather than returning after the first. `insert_header_before_body` already uses the return value as total bytes removed, so `header_end - removed` is still correct with the new semantics.

**Unrelated reorder dropped:** `maybe_handle_control_request` is back in `handle_buffered_api_request` before `admit_buffered_api_request`, matching the structure before this PR. `admit_buffered_api_request` is back to wrapping only `check_activity_admission`.

**MoA degradation / 400 naming a dispatch that never happened (a):** kept refusing, switched to 409, and rewrote the message to name only what this node observed, not what may or may not have dispatched:

> `x-mesh-target/x-mesh-exclude are not honored when the requested model is "mesh" (multi-agent orchestration); if the fleet degraded to a specific model, retry with that model name and the routing header`

This is truthful: we're not claiming MoA ran, only that we can't honor the header for a request naming the `"mesh"` virtual model. The retry hint covers the degraded-fleet case without widening behavior.

---

**Open question for you:** when MoA degrades (`MoaInterceptResult::Degraded`), the effective dispatch is an ordinary model route — `route_request` would honor `x-mesh-target`/`x-mesh-exclude` against real candidates. Should a well-formed routing header on a `model: "mesh"` request be **honored** on the degraded path instead of being rejected here? That changes what `x-mesh-target` means (is it "target the named peer for this model, whatever model ends up dispatched" or "target the named peer only if you route directly, not through orchestration?"). I don't want to widen behavior unilaterally. Staying with the reject-and-hint approach until you weigh in.

---

`cargo test -p mesh-llm-host-runtime`: **2989 passed, 0 failed, 9 ignored**.
`cargo clippy -p mesh-llm-host-runtime -- -D warnings`: clean.
`cargo fmt --check -p mesh-llm-host-runtime`: clean.

StevenMih:

Local results, re-run against head `9b02896db`:

```
```
→ **2989 passed, 0 failed, 9 ignored** (finished in 51.95s)

```
```

```
```
→ **clean**

StevenMih:
@coderabbitai review

StevenMih:
**Round-4 rebase + fix pushed** — head is now `5d1be7b93`.

Rebased onto `origin/main` (`6e19bc065`). Force-push-with-lease completed.

**What changed (on top of round 3):**

**1. Transient plugin-resolution-as-409 → 503 (fix)**

In `route_self_targeted_model`, the previous code used `.ok().flatten().is_some()` on the result of `inference_endpoint_for_model`. A transient `Err` from that call was silently swallowed by `.ok()`, making the whole condition false, and execution fell through to the 409 "x-mesh-target '...' does not serve model '...'" path. A transient resolution failure is not a "valid-but-unsupported" conflict — it's an internal/transient error. Changed to an explicit `match`:

- `Err(error)` → warns + returns **503** with "plugin endpoint for model '...' unavailable (resolution error)"
- `Ok(Some(_))` → routes via plugin (guarded by `!excluded.contains(&self_id)`)
- `Ok(None)` → falls through to the existing 409 close (correct: the peer genuinely doesn't serve the model)

**2. Docstring coverage (75.89% → ≥80%)**

Added `///` docstrings to 10 functions touched by this PR's diff that were missing documentation:
`terminal_outcome_for_dispatch`, `model_access_succeeded`, `response_outcome`, `resolve_remote_mesh_route`, `parse_endpoint_id_hex`, `try_route_plugin_model`, `prepare_cache_routing_body`, `prepare_auto_route_decision`, `send_media_unsupported`, `callable_models_with_local_served`.

**3. MoA behavior question**

Not changing behavior — no dispatch change for the `model: "mesh"` path. CodeRabbit's most recent review reached the same conclusion (no actionable comment on the MoA dispatch behavior). Flagging under "## Needs decision" if you want to honor the header on the degraded path instead.

---

Local results, re-run against head `5d1be7b93`:

```
```
→ **3023 passed, 0 failed, 11 ignored** (finished in 52.03s)

```
```
→ **clean**

```
```
→ **clean**

Not run upstream (workflow approval pending).

StevenMih:
Pushed `04fb852` to address pre-review findings:

**P2 — README narrowed** (`crates/openai-frontend/README.md`): The `x-mesh-served-by` claim "only when `x-mesh-target` was used" was too broad. The `route_self_targeted_model → try_route_plugin_model` path calls `route_http_endpoint_request` which hardcodes `served_by: None` in `RouteAttemptLoggingContext`, so a plugin-served self-target does NOT echo the header. README now reads: "when `x-mesh-target` resolves to a remote peer or a locally-served (non-plugin) model; plugin-served self-targets do not yet echo this header."

**P3 `remote_mesh_test_ctx` unused**: Not applicable — has 6 active call sites in the test file.

**P3 trim on 409/400 body**: Not applicable — header values are already trimmed in the parse functions before format interpolation.

Local: `cargo test -p mesh-llm-host-runtime -p openai-frontend` → all passed; `cargo clippy -p mesh-llm-host-runtime -p openai-frontend -- -D warnings` → clean.

StevenMih:
@ndizazzo — your four are all addressed, and I should be clear that three of them landed in pushes after your review at a2590b6e, so you haven't seen them. Rather than ask you to re-derive that from the diff, each thread below says where the fix is and which test pins it. One new item from this round (the model: "mesh" case) is also fixed and is described in its own thread.

The README support-matrix row is in the same push, deliberately, so there's one head to look at.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
StevenMih: Fixed after your review. enforce_mesh_routing_headers_before_dispatch now runs ahead of the MoA/pipeline dispatch and enforces both malformed-header (400) and unsupported-dispatch-kind (409) up front; the model-less case you called out is flagged through mesh_routing_unsupported_dispatch_kind's "no model specified" arm. ingress.rs:1298-1370, pinned by mesh_routing_unsupported_dispatch_kind_flags_model_less_dispatch.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
StevenMih: Fixed after your review, in route_self_targeted_model — a self-targeted plugin model now tries the local plugin instead of being handed to resolve_remote_mesh_route and 409'd, and a transient endpoint-resolution failure returns 503 rather than failing closed. Pinned by route_self_targeted_model_attempts_a_registered_plugin_instead_of_failing_closed.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
StevenMih: Fixed after your review. x-mesh-exclude naming this node now blocks local-plugin fallback even once the filtered remote pool is empty. Pinned by route_missing_local_model_excluding_self_blocks_local_plugin_fallback.

crates/mesh-llm-host-runtime/src/network/openai/response/relay.rs
StevenMih: Fixed after your review. insert_header_before_body removes any existing header of that name before splicing, and drains every match rather than stopping at the first, so a response that already carried x-mesh-served-by can't come back with two values. response/probe.rs:98-133, pinned by insert_header_before_body_replaces_an_existing_value_instead_of_duplicating.

StevenMih:
> _Review by **erlich** (agent), posted via @i386._
> 
> The fail-closed 409 design is right, and the header-parse hardening in `a2590b6` (empty exclusion entries, non-UTF-8 rejection, the exclude-your-own-target contradiction) is good work with well-targeted tests. Two blockers and two new findings below.
> 
> ### Blocker: this has never been built or tested by CI
> All five workflows on head `a2590b6` concluded `action_required`, and the head SHA has **zero check runs**:
> 
> ```
> $ gh api repos/Mesh-LLM/mesh-llm/commits/a2590b6.../check-runs --jq .total_count
> 0
> $ gh api "repos/Mesh-LLM/mesh-llm/actions/runs?head_sha=a2590b6..." 
> PR · Linux   | completed | action_required
> PR · Quality | completed | action_required
> PR · macOS   | completed | action_required
> PR · Website | completed | action_required
> PR · Windows | completed | action_required
> ```
> 
> 1099 lines across the HTTP ingress, request parsing, and every response relay path, with no compile, clippy, or test evidence from CI. Someone with write access needs to approve the workflow runs. The reported local results are useful but they aren't the gate.
> 
> ### Blocker: @ndizazzo's four findings are still open
> They were filed against `a2590b6`, which is still head — so none of them are stale. I agree with all four. In particular the P1 (validation lives inside the model-bearing branch, so MoA, pipeline, and model-less dispatch all run before the headers are ever consulted) reads to me as one structural question — _where do these headers get enforced_ — rather than four independent patches, and I'd rather see it answered once than fixed four times.
> 
> ### New finding 1: routing headers are forwarded verbatim to the peer
> Nothing strips `x-mesh-target` / `x-mesh-exclude` before `proxy::route_model_request`. The raw path forwards headers as-is — by design, and as the served-by splice comment acknowledges.
> 
> Today the happy path hides it: peer B finds the model locally, never reaches `route_missing_local_model`, never consults the header. The failure appears when B _doesn't_ have it locally — stale gossip, or the model unloaded between advertisement and arrival. B then re-enters `route_missing_local_model` still carrying the original header:
> 
> * with `x-mesh-target: <B>`, `resolve_remote_mesh_route` searches only _other_ peers, so B returns 409 for a route the router had already validated (this is @ndizazzo's P2, reached by a second route);
> * with `x-mesh-target: <C>`, B forwards again — and nothing carries a hop count or accumulates exclusions, so there is no bound on re-routing.
> 
> Same root cause as the duplicate `x-mesh-served-by` finding, and I'd fix it in the same place: strip both headers before forwarding, or replace them with an explicit hop marker the next node understands.
> 
> Flagging as **needs verification** — I traced the call paths and did not reproduce this on a live mesh.
> 
> ### New finding 2: `response_header_end` returns a corrupting offset on its own miss path
> `relay.rs:98-104` returns `response.len()` when no `\r\n\r\n` is found:
> 
> ```rust
> response
>     .windows(4)
>     .position(|window| window == b"\r\n\r\n")
>     .map_or(response.len(), |pos| pos + 4)
> ```
> 
> That value goes straight to `insert_header_before_body`, which splices at `header_end - 2` (`probe.rs:88`). The `header_end < 2 || header_end > buf.len()` guard passes for `len()`, so "terminator not found" silently becomes "splice a header line two bytes from the end of the body."
> 
> Reachable when an upstream terminates its header block with bare LF — `httparse` accepts that, the four-byte window scan does not, so `probe.header_end` and `response_header_end` disagree. Narrow in practice (peers are mesh-llm nodes), but it's a silent body-corruption path where a no-op is the obviously correct behaviour. Suggest `Option<usize>` and skipping the insertion on `None`.
> 
> ### Housekeeping
> 11 commits behind `main` (`origin/main` = `03267cc`).

Good catch, and fixed. The pre-MoA check was returning 409 before try_handle_moa_intercept ran, so a model: "mesh" request that MoA was about to degrade to a concrete model never got the chance to have its headers honoured against that model.

Malformed-header validation (400) still runs unconditionally ahead of MoA. The unsupported-dispatch 409 moved into try_handle_moa itself and now fires only at the point a committee is actually about to be convened — a committee fans a turn across every admitted worker, which is genuinely incompatible with naming or excluding one peer. MoaDispatchResult::Passthrough, the degrade case, continues untouched, so ordinary routing re-parses and honours the headers against the rewritten model. Two tests cover both sides: reject-once-convened, and degrade-and-continue.

StevenMih:
> _Review by erlich (agent), posted via @i386 — reply here and I'll pick it up._
> 
> Round 2 against `f1c5682`. I checked each fix rather than taking the commit message for it. All six findings close. Two new ones and three nits below.
> 
> **CI, still:** this head has **zero check runs** — all five workflows concluded `action_required`, same as `a2590b6`. I can see the local-run numbers in the description (2989 passed, clippy and fmt clean), but I can't attribute those to this SHA. A 1099-line change across the HTTP ingress with no CI evidence at any head shouldn't merge on local runs alone. The branch is also `BEHIND` main (`origin/main` = `afa36ab`).
> 
> ## Verified fixed
> * **My [feat: Add discover meshes feature to console UI #2](https://github.com/Mesh-LLM/mesh-llm/pull/2) (`response_header_end`)** — properly fixed. Returns `Option<usize>` (`relay.rs:104`), `None` skips the insert. Two regression tests, including the bare-LF corruption case.
> * **My [macos menu app #1](https://github.com/Mesh-LLM/mesh-llm/pull/1) (routing headers forwarded verbatim)** — fixed at the right place. `prepare_peer_forwarded_request` now omits both headers, and that is the only peer hop (`response/routing.rs:240` is the sole caller path from `route_remote_attempt`). The unbounded re-route and the peer-side 409 both close with it.
> * **@ndizazzo P1** — answered structurally, as one gate before every dispatch kind, which is what I wanted rather than four separate patches.
> * **P2a / P2b / P2c** — self-target resolves against local plugin availability (`ingress.rs:697`), self-exclusion now blocks plugin fallback with 409, and `insert_header_before_body` replaces rather than appends.
> 
> ## New: MoA degradation is rejected as if it were MoA
> `enforce_mesh_routing_headers_before_dispatch` (`ingress.rs:1240`) rejects `model: "mesh"` plus any valid routing header with 400. But `try_handle_moa_intercept` can return `MoaInterceptResult::Degraded`, which falls through to `route_request` with a real model (`ingress.rs:1514`) — and that path honors the headers fine.
> 
> So on a fleet small enough to degrade, a request that would have worked now gets a 400, and the message names a dispatch ("multi-agent orchestration") that never happened.
> 
> This is the cost of `mesh_routing_unsupported_dispatch_kind` (`ingress.rs:1209`) re-deriving the dispatch decision one step early instead of observing it. The MoA arm is the one that cannot be predicted from the request alone — it depends on fleet size at dispatch time.
> 
> ## New: 400 contradicts this PR's own status vocabulary
> A well-formed `x-mesh-target` on a MoA or model-less request isn't malformed — the server just won't honor it. This PR already established 409 as "valid constraint I can't satisfy" (target doesn't serve the model, node self-excluded). Same rejection class, two different codes, and 400 tells a client to go fix its syntax when there is nothing wrong with it.
> 
> I'd use 409 for the valid-but-unsupported case and keep 400 for parse failures, which is what the parse arm directly above it already does.
> 
> ## Nits
> * `relay.rs:199` — `checked_add_signed(delta).expect(...)` puts a panic in the response relay path. The bound does hold (`removed ≤ header_end ≤ body_end`), so it can't fire, but `saturating_add_signed` is equally correct without the panic.
> * `remove_existing_header` (`probe.rs:70`) returns after the first match. An upstream that already sent the header twice still ends up with two.
> * The commit bundles an unrelated reorder: `maybe_handle_control_request` now runs _after_ `rewrite_public_model_alias` and two node lookups (`ingress.rs:1450-1452`), where it used to reject first. Benign — the control handler only reads path and method — but it inverts what `admit_buffered_api_request`'s own docstring claims, and it isn't mentioned in the commit message.

Good catch, and fixed. The pre-MoA check was returning 409 before try_handle_moa_intercept ran, so a model: "mesh" request that MoA was about to degrade to a concrete model never got the chance to have its headers honoured against that model.

Malformed-header validation (400) still runs unconditionally ahead of MoA. The unsupported-dispatch 409 moved into try_handle_moa itself and now fires only at the point a committee is actually about to be convened — a committee fans a turn across every admitted worker, which is genuinely incompatible with naming or excluding one peer. MoaDispatchResult::Passthrough, the degrade case, continues untouched, so ordinary routing re-parses and honours the headers against the rewritten model. Two tests cover both sides: reject-once-convened, and degrade-and-continue.

StevenMih:
Force-pushed updated head `b78c0f0c7`: reconciled the review-fix lineage and hand-resolved a real merge conflict against `main` — two independently-added `MoaRoutingContext` structs that git had silently merged into non-compiling duplicate code, now merged into one struct with call sites fixed and the routing regression tests passing.

Ran the Linux quality slice locally in Docker before pushing — `rust_fmt`, `rust_clippy` (`--workspace --all-targets -D warnings`), and `quality_contracts` all green on rustc/clippy 1.97.1.

StevenMih:
Thanks again for the detailed review, @ndizazzo. I've pushed `52919b841` (current head) — that's the reviewed `b78c0f0c7` plus two new tests that pin the two cases you and CodeRabbit flagged. I verified each item below by *running* the named test at this head, not by reading the diff.

## Per-item

**P1 — model-less bypass** (parsing moved above only the `effective_model` branch left MoA/pipeline bypasses open). **Fixed** (`15f72a26b`): `enforce_mesh_routing_headers_before_dispatch` parses headers unconditionally, before any dispatch-kind check (`ingress.rs:1339-1372`), so a malformed header returns 400 regardless of `effective_model`; the MoA/pipeline bypasses are closed the same way via `mesh_routing_unsupported_dispatch_kind`. New end-to-end test in `52919b841`: `enforce_mesh_routing_headers_before_dispatch_rejects_malformed_header_without_model` drives a real `TcpStream` request with a malformed `x-mesh-target` and no `model`, and asserts a 400 on the wire. Reverting the ordering fix makes it fail with exactly the symptom CodeRabbit predicted; restoring it passes clean.

**P2a — self-target rejected even when a local plugin serves it.** **Fixed** (`15f72a26b`): a target naming this node now routes through `route_self_targeted_model` (which checks local host- and plugin-served availability) instead of the peers-only `resolve_remote_mesh_route`; 409 only if neither serves it. Covered by `route_self_targeted_model_attempts_a_registered_plugin_instead_of_failing_closed` (re-run — passes).

**P2b — excluded node still served from its own plugin.** **Fixed** (`15f72a26b`): `route_missing_local_model` now checks `excluded.contains(&ctx.node.id())` and returns 409 before the local-plugin-dispatch branch. Covered by `route_missing_local_model_excluding_self_blocks_local_plugin_fallback` (passes).

**P2c — duplicate `x-mesh-served-by`**, plus CodeRabbit's follow-on at `probe.rs:86`. **Fixed** (`15f72a26b`): `insert_header_before_body` now calls `remove_existing_header` (case-insensitive, loops to remove *every* match) before splicing in the new value.

### Correction to my earlier inline reply on P2c

I want to flag this directly rather than bury it: in my earlier inline reply I cited `insert_header_before_body_replaces_an_existing_value_instead_of_duplicating` as the proof. That test only ever sets up **one** pre-existing header, so it can't actually distinguish "stops after the first match" from "removes all matches" — it passes unchanged even with the CodeRabbit bug reintroduced. The multi-duplicate case is now properly pinned by a new test in `52919b841`: `insert_header_before_body_removes_every_duplicate_not_just_the_first`, which sets up two pre-existing occurrences in mixed case (`x-mesh-served-by` / `X-Mesh-Served-By`) and asserts exactly one survives and that it's the fresh value. Reintroducing the "stop after the first removal" bug (a stray `break`) fails this new test with exactly the duplicate you described, while all four pre-existing `probe.rs` tests — including the one I'd cited — still pass, confirming they don't cover this case.

## Two CodeRabbit findings closed in the same pass

- **`ingress.rs:580` — peer-forward loop** (a forwarded request could re-enter the peer's own client-side routing). **Fixed** (`15f72a26b`): `prepare_peer_forwarded_request` strips `x-mesh-target`/`x-mesh-exclude` before the request goes to the peer over QUIC (`forwarded_request.rs`, `OMITTED_ON_PEER_FORWARD`). Covered by `peer_forwarding_strips_mesh_routing_headers` (passes).
- **`ingress.rs:1260-1270` — `model="mesh"` + routing headers 409'd before MoA could degrade.** **Fixed** (`b78c0f0c7`): the malformed-header 400 still runs unconditionally ahead of MoA, but the unsupported-dispatch 409 moved into `try_handle_moa` right before it would convene a committee, so a degrade-to-Passthrough now continues with the headers intact. Covered by `try_handle_moa_rejects_routing_headers_once_a_committee_is_convened` and `try_handle_moa_degrades_and_continues_when_no_committee_can_form` (both pass).

## CI

You'd noted that all five workflow runs on this PR conclude `action_required` without a maintainer approval — still the case at `52919b841` (only the CodeRabbit check reports; no Actions runs). Once you're able to approve the pending run (or point me at who can), that closes the one gap we can't close from the fork side. In the meantime `ci-local.sh` (docker/linux) runs green locally on this head: `quality_contracts` (6 checks) pass, `rust_fmt` pass, `rust_clippy` (workspace, `-D warnings`) pass; full `cargo test -p mesh-llm-host-runtime --lib` = **3089 passed, 0 failed, 11 ignored**.

<sub>Branch-history note: your review anchor `a2590b6e4` (and the earlier `d2b50b7f` / `f1c5682b` / `9b02896d`) aren't reachable from `52919b841` by ancestry — the branch was rebased/reconciled — but they're patch-identical to `b50b48fcb` / `15f72a26b` / `a078a3835` in the current chain (`git patch-id`), so nothing was lost, just renumbered.</sub>

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
i386: Verified at 52919b84123af8fe239ec19152bda8aa5a43cc2d. Routing-header parsing and unsupported-dispatch enforcement now precede ordinary MoA, pipeline, and model-less dispatch. The branch includes dedicated regression coverage for each path, while allowing the explicit MoA degradation passthrough handled below.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
i386: Verified at 52919b84123af8fe239ec19152bda8aa5a43cc2d. A self-target now resolves through route_self_targeted_model and attempts the registered local plugin. The targeted regression covers the previously failing 409 path.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
i386: Verified at 52919b84123af8fe239ec19152bda8aa5a43cc2d. The local plugin fallback checks self exclusion before dispatch, with a regression test that confirms the excluded node cannot serve the request.

crates/mesh-llm-host-runtime/src/network/openai/response/relay.rs
i386: Verified at 52919b84123af8fe239ec19152bda8aa5a43cc2d. insert_header_before_body removes every case-insensitive existing field before inserting one canonical value; the branch tests replacement, duplicate removal, and case-insensitive matching.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -888,6 +1197,110 @@ fn pipeline_route_model<'a>(
...
+            409,
+            proxy::send_409_observed(
+                tcp_stream,
+                "x-mesh-target/x-mesh-exclude are not honored when the requested model is \
+                 \"mesh\" (multi-agent orchestration); if the fleet degraded to a specific \
+                 model, retry with that model name and the routing header",
+                route_observer,
+            )
+            .await,
+        )),
```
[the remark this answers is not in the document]
i386: Verified at 52919b84123af8fe239ec19152bda8aa5a43cc2d. Commit b78c0f0c7065dc531c7c4ef8d3fccb942bb668d2 defers the unsupported-dispatch decision until the MoA outcome is known, and its tests cover degraded passthrough with both target and exclusion headers.

StevenMih:
@ndizazzo — following up on the changes-requested from the 7th. @i386 went through all four items thread-by-thread on Friday and marked each verified at 52919b84: the model-less bypass (now enforced before MoA/pipeline dispatch, with an end-to-end test), the self-target and self-exclusion plugin paths, and the duplicate x-mesh-served-by (a new test covers the multi-header case — my earlier reply cited one that didn't, which he caught). All ten threads are resolved and CI is green on this head. Whenever you have a moment to re-review — happy to point at anything that still looks open.

Same ask on #1668, where the rebase dismissed your earlier approval and nothing in the diff changed beyond the clippy fix your run caught.

michaelneale (dismissed):
looks good - feedback addressed, I like it. have thought about it too.

i386 (dismissed):
Reviewed the complete current-main-synchronized head `77a4f8618b5ac5318d6b71682c4f385f003a3a30`. Target/exclude parsing fails closed on malformed or contradictory input, routing directives are stripped before forwarding, self-target/self-exclude behavior is explicit, and the served-by header is added only on the targeted response paths while preserving untouched responses when absent. The MoA degradation path remains available before directive rejection.

The full `mesh-llm-host-runtime` package suite and `just ci-validate` pass, and there are no unresolved review threads.

i386 (dismissed):
Reviewed exact head dc3de36d92f7c35082d40acb4f8b1e010723e1d0 after synchronization with current main. I fixed the CI Clippy failure in the request-header parser and verified the full mesh-llm-host-runtime all-target Clippy gate with -D warnings, formatting, and diff checks. The feature behavior and earlier full host-runtime suite remain sound.

StevenMih:
Rebased onto current main (post-#1841), new head b9af9ab. Two commits: the feature, and one round-2 fixes commit covering @ndizazzo's four findings (headers parsed and enforced once at the top of route_request — MoA/pipeline/model-less get an explicit 400 rather than silently ignoring them; self-target resolves against local host or plugin availability; self-exclude blocks local plugin fallback; served-by replaces rather than appends), @i386's two (both headers stripped before forwarding to the peer, so a peer can never re-route or loop; response_header_end returns Option and skips the insert on a miss), and CodeRabbit's model-less case. The #1668 commit that used to sit underneath is gone — it's on main. @michaelneale your approval was at the old head; re-requesting you and @ndizazzo.

PR #1672, opened 2026-09-05
feat(skippy): complete graph-derived v2 model splitting

i386:
## Problem and resulting behavior

A split model could reach native graph construction with a tensor closure or state boundary that disagreed with the coordinator plan. Stage behavior was also spread across hand-maintained model-family patches, which made upstream llama.cpp rebases large and easy to get wrong.

This PR cuts package serving over to package schema v2 and graph-derived stage planning. Every realized stage is bound to the admitted package, graph/profile identities, exact resident tensor closure, state effects, boundary schema, backend, and configuration. Mismatches fail before topology publication.

## Delivered

- graph-observed, metadata-only stage planning across dense, MoE, hybrid, recurrent, MTP, and multimodal graphs
- source-complete package-v2 writing and verification, exact tensor/sidecar admission, and v2-only serving
- generation-8 stage-control descriptors and participant echo validation
- lossless activation encoding selection over the complete F32 frame, including sideband planes
- recurrent state ownership, rollback, KV/state handoff, MTP/shared-embedding, hyperconnection, sideband, and multimodal handling
- Darwin-safe binary transport timeout polling and deterministic two-node evidence reconciliation
- a structural Clang transformer that proves stage-control edits across llama.cpp builders without family-name or tensor-name eligibility rules

## Patch queue and generated-family maintenance

The maintained llama.cpp queue is now **11 capability-owned core patches**, down from 76. They are followed by **33 deterministic generated shards**: 32 mapped shard groups covering every certified family and one fail-closed shard for model sources that do not yet have a certified artifact. Generated shards always apply after the core queue. They are byte-compared in CI and are not hand-maintained.

The current GitHub compare reports **315 files, +64,586 / −34,187**. Those are accurate raw diff statistics after rewriting tracked patch artifacts: deleted historical patch bodies and added consolidated/generated patch bodies are both counted. They do not represent 63,000 new runtime source lines.

The transformer classifies all **147 builders** found across **152 model source files**: 128 already transformed, 18 supported auxiliary builders, and one supported whole-model builder. The final report contains **zero unsupported shapes and zero analyzer errors**.

For a llama.cpp pin update, CI diffs the actual old and new upstream revisions. Changes confined to mapped `src/models/*.cpp` files run the affected manifest families plus dense, MoE, recurrent, and multimodal sentinels. Shared upstream code, an unmapped model source, or an unavailable diff fails closed to the full battery. Nightly and manual-full runs remain full.

## Validation

At model-affecting head `f184496b3`:

- clean replay of 11 core patches followed by all 33 generated shards
- generated shard series is byte-identical across independent runs; a second transformer pass emits zero edits
- full patched llama.cpp build: 476 targets
- full configured llama.cpp CTest suite: 92/92 passed, including all 27 Skippy native tests and the upstream suite
- `just ci-validate`: 914 tests executed with 7 expected skips; one signal-timing cleanup test failed once and passed on isolated rerun; actionlint, shellcheck, and repository consistency checks otherwise passed
- immutable local cache inventory: all 35 certified families and their required model/projector artifacts passed
- Windows CRLF generated-series integration test and the complete four-test prepare suite passed
- prepared llama.cpp tree: `bd7bfd7c46f235c114aaf568be53335e4383d916`

At model-runtime head `d68fa9ca2`, review fixes preserve unbraced conditional-continue semantics, derive targeted family selection from the actual upstream source diff, and keep native MTP/NextN inside its single target model. The family battery no longer reopens an MTP model as its own draft; single-step and chain require the native draft sideband and verify its proposed next token against the target. Targeted Python, shellcheck, actionlint, formatting, and the full `skippy-correctness` package pass. A real two-family MTP preflight records one target and no draft for both affected families. Regenerating all 33 family shards produces no diff, and a clean replay still produces the exact tree `bd7bfd7c46f235c114aaf568be53335e4383d916`.

Current PR head `8747286ef` addresses all 16 open review findings and the GLM-4.5-Air Metal OOM. Package-v2 admission now resolves the manifest tensor closure to canonical native names and the loader enforces exact set equality. Planned and realized descriptors come from independent discovery and realization passes. The commit also fixes codec policy propagation, live status routing, sparse layer-byte fallback, timeout retry bounds, non-finite codec handling, typed admission state, quantization-before-write, deterministic package identity, CI smoke selection, shell portability, raw source-byte handling, transformer refusal rules, and real post-rewrite native verification.

Validation at `8747286ef`:

- clean replay: 11 capability patches plus 33 generated shards; prepared llama.cpp tree `55e71f34aeb4a83c6b4a0cdba28956e236aae73a`
- generated-series check: 152 model sources and 147 builders, zero unsupported shapes or analyzer errors, byte-identical regeneration, zero-edit second pass, post-rewrite build, and 27/27 Skippy native tests
- llama.cpp CPU and Metal builds with 27/27 Skippy native tests on each backend
- `just ci-validate`: 922 tests executed, 7 expected skips, all repository consistency, actionlint, and shellcheck gates passed
- `skippy-runtime`: 118/118; `mesh-llm-host-runtime --lib`: 2969/2969 with 11 ignored; `skippy-server`: 694/694 with 3 ignored
- GLM-4.5-Air three-stage runtime slice `0–15 / 15–31 / 31–47`: pass, baseline match, native MTP sideband present and verified; disabling mmap for filtered runtime slices removes the sparse-file Metal residency failure

The full configured 92-test llama.cpp suite passed at `6164c569a`; the current GitHub canary is rerunning that suite and the family selection against `8747286ef`. The table below records the model evidence already completed.

## Certified-family evidence matrix

All rows run load, single-step split parity, a three-stage chain, and state handoff (including the in-run KV/state continuation oracle). Native-MTP rows additionally require the in-model draft sideband and verify the proposed next token against the target without loading a second model. Rows with a sweep period exercise additional internal cuts. Multimodal rows with a projector run local and split mmproj smoke.

| Family | Class | Artifact | Split parity | 3-stage chain | KV/state handoff | Boundary sweep | Native MTP sideband | mmproj |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| qwen3-dense | dense | `Qwen3-0.6B-Q8_0.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| llama | dense | `Llama-3.2-1B-Instruct-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| glm47-flash | dense / MLA | `GLM-4.7-Flash-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| glm4 | dense | `glm-4-9b-chat-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| glm45-air | MoE + MTP | `GLM-4.5-Air-Q4_K_M-00001-of-00002.gguf` | Pass @ `6164c569a` | Pass @ `8747286ef` | Pass @ `6164c569a` | — | Pass @ `8747286ef` | — |
| deepseek2 | dense / MLA | `DeepSeek-Coder-V2-Lite-Instruct-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| gpt-oss | MoE | `gpt-oss-20b-MXFP4.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| qwen3-moe | MoE | `Qwen3-30B-A3B-Instruct-2507-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| qwen2-moe | MoE | `Qwen1.5-MoE-A2.7B-Chat.Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| falcon-h1 | hybrid | `Falcon-H1-1.5B-Instruct-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (12 cuts) @ `f184496b3` | — | — |
| jamba2 | hybrid | `ai21labs_AI21-Jamba2-3B-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (18 cuts) @ `f184496b3` | — | — |
| kimi-linear | hybrid MoE | `moonshotai_Kimi-Linear-48B-A3B-Instruct-IQ2_XS.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (12 cuts) | — | — |
| qwen3-next | hybrid MoE | `Qwen_Qwen3-Next-80B-A3B-Thinking-IQ2_XS.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (12 cuts) | — | — |
| qwen4exp | hybrid | `Qwen3.8-Flash-Next-UD-IQ1_S-00001-of-00003.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (12 cuts) @ `f184496b3` | — | — |
| nemotron | hybrid MoE + MTP | `NVIDIA-Nemotron-3.5-Lightning-30B-A3B-UD-Q4_K_M.gguf` | Pass @ `6164c569a` | Pass @ `6164c569a` | Pass @ `6164c569a` | Pass (6/6 internal cuts plus terminal cut) @ `6164c569a` | Pass @ `6164c569a` | — |
| mamba | recurrent | `mamba-130m-hf.Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (6/6 cuts) @ `f184496b3` | — | — |
| mamba2 | recurrent | `falcon-mamba-7b-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (6/6 cuts) @ `f184496b3` | — | — |
| rwkv6 | recurrent | `rwkv-6-finch-1b6-Q4_0.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (3/3 cuts) @ `f184496b3` | — | — |
| rwkv7 | recurrent | `RWKV7-Goose-World3-2.9B-HF-f16-q4_k.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (3/3 cuts) @ `f184496b3` | — | — |
| granite-hybrid | hybrid | `granite-4.0-h-350m-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (31/31 cuts) @ `f184496b3` | — | — |
| qwen35 | hybrid | `Qwen3.5-0.8B-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (12/12 cuts) @ `f184496b3` | — | — |
| granite-dense | dense | `ibm-granite_granite-3.2-2b-instruct-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| lfm2 | hybrid | `LFM2-350M-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (6/6 cuts) @ `f184496b3` | — | — |
| lfm2-vl | multimodal hybrid | `LFM2-VL-450M-Q8_0.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass (6/6 cuts) @ `f184496b3` | — | Pass: real image + projector, local and split @ `f184496b3` |
| laguna | dense | `Laguna-XS-2.1-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| qwen2-vl | multimodal | `Qwen2-VL-2B-Instruct-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | Pass: real image + projector, local and split @ `f184496b3` |
| qwen3-vl | multimodal | `Qwen3-VL-8B-Instruct-Q4_K_M.gguf` | Pass @ `6164c569a` | Pass @ `6164c569a` | Pass @ `6164c569a` | — | — | Pass: real image + projector, local and split @ `6164c569a` |
| gemma2 | dense | `2b_it_v2.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| phi3 | dense | `Phi-3-mini-4k-instruct-q4.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| phi4 | dense | `phi-4-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| falcon-dense | dense | `tiiuae-falcon-7b-instruct-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| internlm2 | dense | `internlm2-chat-1_8b.F16.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| mistral-small | dense | `Mistral-Small-3.2-24B-Instruct-2506-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| cohere2 | dense | `c4ai-command-r7b-12-2024-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |
| smollm3 | dense | `SmolLM3-Q4_K_M.gguf` | Pass @ `f184496b3` | Pass @ `f184496b3` | Pass @ `f184496b3` | — | — | — |

## Direct-GGUF contract and two-host proof at `e032d07d6`

Generation-8 split serving again accepts a direct local GGUF source. It derives the source-complete package-v2 planning envelope in memory, validates every GGUF shard and tensor, binds the admitted resident-tensor closure into native loading, and writes no package or layer shard. The peer status responder now reports host-embedded stages from its own runtime state while excluding cached peer snapshots.

The exact-head release binary (`d37731489ef9c7498afcd5011c2bdbb790768e2f989896f9a4df05e623477571`) ran Qwen3.8 Flash Next across micstudio and studio54 with no explicit context-window setting. Both LAN consoles independently reported stage 0 `0–47` and stage 1 `47–48` Ready at auto-selected context 262,144. A LAN chat request returned HTTP 200 with `split online`; both isolated data directories contained no generated package or layer artifact.

Exact-tree validation: `just ci-validate` passed 922 tests with 7 expected skips; `mesh-llm-host-runtime` passed 2,973 tests with 11 expected ignores plus 2/2 audit tests; all-target clippy with warnings denied, rustfmt, diff checks, and the release host build passed.

## Review state

All 16 original review findings are addressed in `8747286ef`; each review thread has a commit-specific response and is resolved. Local validation is green through `e032d07d6`, including the GLM-4.5-Air and direct-GGUF Qwen3.8 regressions. GitHub CI is running on the pushed head. The PR remains unmerged and awaits CI completion and reviewer approval.


crates/skippy-model-package/src/package_v2.rs
```diff
@@ -0,0 +1,227 @@
...
+            .as_secs(),
+    })
+}
+
+fn copy_artifact(source: &Path, output: &Path, resume: bool) -> Result<()> {
+    if resume && output.is_file() {
+        ensure!(
+            fs::canonicalize(source)? != fs::canonicalize(output)?,
+            "package artifact must not be the independent source file"
+        );
```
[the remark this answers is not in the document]
i386: Fixed in 673e133af. Resume now uses same_file::is_same_file, and a hard-link regression proves the manifest is not written when source and artifact share an inode. skippy-model-package passes 81/81.

crates/skippy-package-format/src/lib.rs
```diff
@@ -0,0 +1,2424 @@
+use std::collections::{BTreeMap, BTreeSet};
+use std::fmt;
+use std::path::{Component, Path};
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use sha2::{Digest, Sha256};
+
+pub mod materialization;
+pub mod stage_admission;
```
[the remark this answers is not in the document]
i386: Fixed in 673e133af. The inline tests moved to src/tests.rs; lib.rs is now 1,335 lines, below the repository limit, with public paths unchanged. skippy-package-format passes 42/42 plus Clippy.

scripts/build-llama.sh
```diff
@@ -363,17 +370,37 @@ fi
...
 
 cmake --build "$LLAMA_BUILD_DIR" --config "${CMAKE_BUILD_TYPE:-Release}" --parallel "$(detect_jobs)" --target "${BUILD_TARGETS[@]}"
 
-if [[ "${LLAMA_STAGE_BUILD_TESTS:-OFF}" == "ON" ]]; then
+if [[ "$LLAMA_STAGE_FULL_REPLAY" == "ON" ]]; then
+  REPLAY_FIXTURE_DIR="$LLAMA_BUILD_DIR/tests/test-models"
+  mkdir -p "$REPLAY_FIXTURE_DIR"
+  "$LLAMA_BUILD_DIR/bin/skippy-model-fixture-generator" --arch gemma --seed 1 --out "$REPLAY_FIXTURE_DIR"
+  "$LLAMA_BUILD_DIR/bin/skippy-model-fixture-generator" --arch qwen2moe --seed 1 --out "$REPLAY_FIXTURE_DIR"
+  ctest --test-dir "$LLAMA_BUILD_DIR" --build-config "${CMAKE_BUILD_TYPE:-Release}" --output-on-failure --timeout 900 --fixture-exclude-any '^generate-models$' -R '^(skippy_|test-skippy-)'
```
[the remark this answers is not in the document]
i386: Fixed in 673e133af. Full-replay mode bypasses the cached early return, while ordinary cached builds retain the shortcut. The regression executes both paths twice, and the canonical native replay passes 28/28.

scripts/plan-ci.py
```diff
@@ -510,9 +510,11 @@ def _select_rows(
         for domain in domains:
             smoke_ids.extend(slices.get("smoke_domain_rows", {}).get(domain, []))
         if "product-smoke" in selected and not smoke_ids:
-            smoke_ids.append("core")
+            smoke_ids.append("product-integration-cpu")
         if profile == "pr-draft" and smoke_ids:
-            smoke_ids = [row_id for row_id in smoke_ids if row_id == "core"] or [smoke_ids[0]]
+            smoke_ids = [
+                row_id for row_id in smoke_ids if row_id == "product-integration-cpu"
+            ] or [smoke_ids[0]]
```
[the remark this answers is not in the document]
i386: Resolved in 673e133af by removing the unreachable branch. Draft profiles are intentionally empty under the current planner policy; the existing planner tests preserve that behavior and pass 31/31.

third_party/llama.cpp/patches/0060-skippy-select-graph-derived-stage-slices.patch
```diff
@@ -0,0 +1,873 @@
...
++            if (finish == node_index.end()) {
++                error = "stage end boundary is not a graph node";
++                return false;
++            }
++            add_live_edges(finish->second, live_end);
++        }
++
++        const ggml_tensor * primary_output = layer_end == layer_count
++                ? ggml_graph_node(graph, -1)
++                : boundaries[2 * (layer_end - 1) + 1].tensor;
```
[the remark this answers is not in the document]
i386: Fixed in 673e133af through the cumulative final ABI patch. The planner threads the explicit graph result output (logits, pooled embedding, or embedding) into final-stage selection. A trailing unrelated ordered node regression proves it cannot replace the output dependency cone; missing output fails closed. Native replay passes 28/28.

third_party/llama.cpp/patches/0061-skippy-construct-isolated-executable-graph-slices.patch
```diff
@@ -0,0 +1,547 @@
...
++            }
++        }
++    }
++    for (int32_t index : node_indices) {
++        const ggml_tensor * tensor = source->nodes[index];
++        if (tensor->flags & GGML_TENSOR_FLAG_COMPUTE) {
++            ggml_build_forward_expand(graph, builder.result.tensors.at(tensor));
++        } else {
++            ggml_build_forward_order(graph, builder.result.tensors.at(tensor));
++        }
```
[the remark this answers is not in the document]
i386: Verified as a false positive at the pinned cc83d7b48 tree. The available headers define the ggml_can_fuse overload used here, and skippy-stage-slice-plan compiled and passed within the 28/28 full native replay at 673e133af. No code change was needed.

third_party/llama.cpp/patches/0063-test-skippy-validate-guarded-profile-stage-chains.patch
```diff
@@ -0,0 +1,318 @@
...
++                    }
++                    if (!error.empty()) {
+                         return false;
+                     }
+                 }
+@@ -488,7 +527,6 @@ bool skippy_build_guarded_stage_slice_plan(
+         }
+         if (profile->guard.n_tokens == 0 || profile->guard.n_sequences == 0 ||
+                 profile->guard.n_outputs == 0 ||
+-                profile->guard.n_tokens % profile->guard.n_sequences != 0 ||
```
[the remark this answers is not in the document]
i386: Fixed in 673e133af by restoring the divisibility guard and changing the uneven 5/2 case back to rejection, keeping the recurrent write-count contract consistent. Native replay passes 28/28.

crates/mesh-llm-host-runtime/src/runtime/local_package.rs
```diff
@@ -832,19 +831,78 @@ pub(super) fn split_participant_set_hash(participants: &[SplitParticipant]) -> S
...
         hasher.update(stage.layer_end.to_le_bytes());
         hasher.update(stage.parameter_bytes.to_le_bytes());
+        if let Some(admission) = admissions.get(&stage.stage_id) {
+            hasher.update([1]);
+            hasher.update(admission.version.to_le_bytes());
+            hash_field(&mut hasher, admission.package_id.as_bytes());
+            hash_field(&mut hasher, admission.plan_id.as_bytes());
+            hasher.update(admission.layer_start.to_le_bytes());
+            hasher.update(admission.layer_end.to_le_bytes());
+            hasher.update((admission.resident_tensor_ids.len() as u64).to_le_bytes());
```
ndizazzo, line 880: This hasher is the only non-test consumer of `resident_tensor_ids` in the whole tree. Grepping crates/ turns up this loop, the protocol test fixtures, and nothing else.

So the closure we spend the entire native partitioner deriving gets folded into `split_topology_hash` and then thrown away. The actual load still goes through `llama_model_loader_set_stage_filter` and `g_skippy_filter`, which re-derives retention from the coarse `LLM_TENSOR_LAYER_*` classes plus the four hardcoded arch exceptions.

We should bind the load to this list instead: pass `resident_tensor_ids` through `PackageManifest::materialization_tensors` (it already returns native_name, ggml_type, dimensions, artifact, data_offset, stored_length, alignment, which is exactly what the loader needs) and make retention a set-membership test. Once that's in place the whole `stage_filter_excludes` function and its arch ladder can be deleted, which I think was the point.

crates/mesh-llm-host-runtime/src/runtime/stage_admission.rs
```diff
@@ -0,0 +1,1750 @@
...
+        ranges,
+        profiles,
+        graph_configuration_id,
+        backend_id,
+        &sidecars_by_stage,
+    )?;
+    realized
+        .iter()
+        .map(|realized| {
+            let planned = PlannedStageAdmission::from(realized);
```
ndizazzo: `planned` is built by cloning `realized` right here, and this is the only non-test path into `admit_stage_plan`. The `From<&RealizedStagePlan>` impl copies package_id, plan_id, layer_start/end, resident_tensor_ids, sidecars, and every profile identity verbatim.

So every planned-vs-realized check downstream is comparing a value to a copy of itself. `PackageIdMismatch`, `PlanIdMismatch`, `LayerRangeMismatch`, `TensorClosureMismatch`, `SidecarMismatch`, `ProfileSetMismatch`, `ProfileIdentityMismatch`, and all of `admit_profiles` are unreachable. The only thing in here with any force left is the `resolve_stage_admission` tail.

Which means if the native realizer ever produces a stage over the wrong layer range, against a different package, with a dropped sidecar, or with a profile the backend actually rejected, we publish it into the topology anyway. The module doc promises the opposite. The unit tests build the two sides independently, so they pass while production compares X to X.

We should build `PlannedStageAdmission` from the topology plan and the manifest *before* `realize_native_stage_chain` runs, pass it in as an independent expectation, and delete the `From` impl so a self-comparison isn't expressible.

crates/mesh-llm-host-runtime/src/mesh/stage_proto.rs
```diff
@@ -68,6 +68,9 @@ pub(super) fn stage_runtime_status_from_snapshot(
         node_id,
         layer_start: status.layer_start,
         layer_end: status.layer_end,
+        admission: status.admission,
+        activation_codec: status.activation_codec,
+        activation_codec_policy: Default::default(),
```
ndizazzo: `activation_codec` on the line above is propagated correctly, but the policy right under it gets defaulted, which lands on `Fixed`. Same thing at 117, 804, 913, 950, and 981.

Every new split generation sets `AutoLosslessV1` (`local_split/coordinator.rs:74`), so every status we record or serve reports the wrong policy. That matters more now that peer `Status` is answered from this cache. And `local_split/attestation.rs:42` and `:76` compare this exact field for equality, so anything routed through a cached or derived status fails closed on a healthy topology.

While we're in here: stage 0's own `StageConfig` never picks up the negotiated policy either, so `forwarding.rs:119` takes the `Fixed` branch and stage 0 never compacts a frame. That's the only hop in the two-node split, so as far as I can tell the feature is inert end to end right now. `skippy-cache/src/identity.rs:78` also hashes the policy, so stage 0 and stages 1..N end up in different KV namespaces inside one topology.

Just `status.activation_codec_policy` in all six spots should do it.

crates/mesh-llm-host-runtime/src/mesh/stage_artifacts.rs
```diff
@@ -232,12 +232,17 @@ impl Node {
...
         {
-            self.record_stage_topology(stage_topology_from_load(self.endpoint.id(), load))
-                .await;
+            self.record_stage_load_topology(load).await;
         }
-        let response = self
-            .execute_stage_control_request_for_peer(remote, request)
-            .await?;
+        let response =
+            if let crate::inference::skippy::StageControlRequest::Status(filter) = &request {
```
ndizazzo: Answering a peer's `Status` from our own mesh cache instead of the live stage-control loop loses the thing the caller is asking for. `cached_stage_statuses` reads `stage_topologies.runtime_statuses()`, which only ever gets written by `record_stage_load_topology` (Starting) and `record_stage_control_response` (Ready/Status).

Two ways that bites. A stage whose native runtime dies or gets torn down without a control request keeps answering Ready forever, so `refresh_stage_runtime_statuses` never notices. And if we have no cached entry at all (topology withdrawn, mesh state restarted, stage started outside `handle_stage_control`) we return `Status(vec![])`, which `record_stage_status_refresh` reads as `MissingFromRuntime` and turns into a Failed status for a stage that's alive.

There's a related problem in the cache itself: `stage_transport.rs:733` blanks `admission` for every stage except the one being loaded, but `stage_status_from_load` still copies the non-empty `model_id`. That's exactly the combination `validate_status_stage_admission` rejects, and it rejects the whole `StageStatusList`, not just the bad row. So serving placeholders out of here can fail an entire status refresh.

I'd keep routing `Status` through the live loop.

crates/mesh-llm-host-runtime/src/inference/skippy/package.rs
```diff
@@ -40,6 +58,372 @@ pub struct SkippyPackageSourceFile {
...
+                source.sha256 == artifact.sha256,
+                "package-v2 artifact {:?} SHA-256 differs from manifest",
+                artifact.id
+            );
+            Ok(source)
+        })
+        .collect()
+}
+
+fn package_v2_layer_weight_bytes(manifest: &PackageManifestV2) -> Result<Vec<u64>> {
```
ndizazzo, line 278: This always returns all zeros. `layer_ordinal` is `None` on every tensor a v2 manifest can contain, because `source_inventory.rs:256` deliberately sets it that way (GGUF has no structural per-tensor layer field and the name-derived one was intentionally dropped). So the loop `continue`s on every entry and we hand back `vec![0; layer_count]`.

The trouble is that both consumers gate on length rather than content: `split_planning.rs:380` and `skippy-coordinator/src/topology.rs:569` both check `len() == layer_count`, so the all-zero vector is accepted as authoritative and the `model_weight_bytes / layer_count` fallback never kicks in. `topology.rs:600` then fits with `candidate_bytes_per_layer(0, kv_per_layer, ...)`, so on a 70B the weights contribute nothing to the VRAM decision, `fit_candidate` packs every layer onto the first node whose KV cache alone fits, the plan is accepted, and the stage load OOMs. It also zeroes `parameter_bytes` in the topology hash and collapses `assigned_bytes` in `local_split/loading.rs:630`, which shrinks the prepare timeouts to the floor.

The direct-GGUF sibling at line 896 gets this right, it returns `Ok(Vec::new())` when the layout is unknown precisely so the fallback engages. Let's do the same here until v2 actually carries layer ordinals.

crates/skippy-server/src/binary_transport/binary_messaging.rs
```diff
@@ -114,13 +141,20 @@ impl ConnectionWorkerControl {
                             | io::ErrorKind::WouldBlock
                             | io::ErrorKind::TimedOut
                     ) => {}
+                Err(error) if timeout_armed && error.raw_os_error() == Some(EINVAL) => {}
```
ndizazzo: This arm matches every `EINVAL` from `peek`, not just a timeout expiry. `timeout_armed` is true on every normal path (it's only false when `set_read_timeout` failed, and that returns early), and the body is empty with no sleep and no retry cap.

The arm above it is safe because `WouldBlock` and `TimedOut` only show up after the 100ms actually elapses. `EINVAL` has no such guarantee: an fd the kernel rejects after a partial teardown returns it immediately, and on Linux it has nothing to do with `SO_RCVTIMEO` at all. When that happens the worker spins on `peek()` at full speed forever, since the only way out is `is_shutting_down()` becoming true and nothing is tearing this connection down.

Either cap the retries, or check that the timeout window really did elapse before treating `EINVAL` as benign.

crates/skippy-protocol/src/binary/activation_codec.rs
```diff
@@ -0,0 +1,843 @@
...
+    codec: StageActivationCodec,
+    shape: ActivationShape,
+    values: &[f32],
+) -> io::Result<Vec<u8>> {
+    let elements = shape.elements()?;
+    shape.validate_decoded_limit()?;
+    if values.len() != elements {
+        return Err(invalid_data("activation value count does not match shape"));
+    }
+    validate_finite(values)?;
```
ndizazzo: This makes the RawF32 path reject NaN and infinities, and `decode_raw_f32` re-checks on the way back in. On main, `encode_f32_activation_payload_with_state_flags` was just `Ok(f32_payload.to_vec())` and the reader stored the bytes verbatim, so a non-finite value went through harmlessly.

Now the default `Fixed(RawF32V1)` path refuses the frame, which is a problem because non-finite values are a normal thing to see: an fp16 overflow upstream, a fully-masked attention lane, a degenerate softmax. `forwarded_stage_message_timed` returns the error, it escapes the connection worker, and the stage link dies mid-request. Under a rolling upgrade it's worse, since an old producer keeps emitting the frame and only the new consumer rejects it.

The round-trip arithmetic looks right to me for all four codecs, it's specifically the value-domain restriction on the path we advertise as byte-exact that I'd drop. If we want a finiteness check it probably belongs behind a debug assert or on the lossy codecs only.

crates/skippy-protocol/src/validation.rs
```diff
@@ -199,9 +243,294 @@ pub fn validate_stage_control_response(
...
+        ));
+    }
+    Ok(())
+}
+
+fn validate_status_stage_admission(
+    status: &proto::stage::StageStatus,
+) -> Result<(), StageFrameError> {
+    let Some(admission) = status.admission.as_ref() else {
+        if status.model_id.is_empty() {
```
ndizazzo: An empty `model_id` gets a status admitted with no descriptor at all: no package_id, no plan_id, no tensor closure, no profile identities, and the layer-range cross-check below gets skipped along with it. Same shape at line 387 for `StagePreparationStatus`.

That's the whole barrier keeping a generation-7 node out of a generation-8 topology, and it's an absent value plus a permissive default standing in for a type. A peer that just omits `model_id` walks through.

I'd rather model "idle, nothing loaded" as its own variant (a `oneof` of `Idle` vs `Admitted{descriptor}`) so an admitted-but-undescribed stage isn't constructible in the first place.

crates/skippy-model-package/src/package_v2.rs
```diff
@@ -0,0 +1,227 @@
...
+pub(crate) fn write_package(
+    model: String,
+    out_dir: PathBuf,
+    projectors: Vec<PathBuf>,
+    artifact_hook: ArtifactHook,
+    artifact_transform: ArtifactHook,
+    explicit: ExplicitSourceIdentity,
+    resume_existing_artifacts: bool,
+) -> Result<()> {
+    ensure!(
```
ndizazzo, line 30: This ensure fires on every `skippy-quantize` run. `main.rs:106` now points `Command::WritePackage` at this function, and `skippy-quantize/src/main.rs:872` unconditionally passes `--transform-artifact-command <hook>`. Clap still accepts the flag so it parses fine, then we bail and the quantize pipeline dies at the packaging step.

`git diff --stat origin/main...HEAD -- crates/skippy-quantize/` is empty, so the caller never got updated. Takes out both shell entry points too (`glm52-q2-routed-down-layer-package.sh:83` and `skippy-bf16-to-quant-layer-package.sh:12`).

The rule itself is reasonable, we just need to teach skippy-quantize to transform the source before packaging rather than passing a hook.

crates/skippy-model-package/src/package_v2.rs
```diff
@@ -0,0 +1,227 @@
...
+        sidecars: Vec::new(),
+        generation: None,
+        native_abi_version: format!(
+            "{}.{}.{}",
+            skippy_ffi::ABI_VERSION_MAJOR,
+            skippy_ffi::ABI_VERSION_MINOR,
+            skippy_ffi::ABI_VERSION_PATCH
+        ),
+        generator_version: env!("CARGO_PKG_VERSION").to_string(),
+        created_at_unix_secs: SystemTime::now()
```
ndizazzo, line 176: `created_at_unix_secs` is a plain serialized field on `PackageManifest`, and `computed_package_id()` only clears `package_id` before hashing. So the timestamp is inside the digest, and `package_id` is clock-dependent rather than content-addressed.

Two builds from byte-identical source GGUFs with the same generator and ABI version get different ids if they start in different seconds. That breaks reproducible-build verification, cross-machine dedup, and anything keyed on the id. And since this id is the root of the plan digest chain, the determinism story downstream inherits it.

Easiest fix is clearing it in `normalized` alongside `package_id`.

.github/actions/plan-ci/action.yml
```diff
@@ -309,7 +309,7 @@ runs:
                 hosts: [.matrices.hosts[] | select(.platform == "macos")],
                 runtime_products: [.matrices.runtime_products[] | select(.platform == "macos")],
                 platform_checks: [.matrices.platform_checks[] | select(.platform == "macos")],
-                smoke: [.matrices.smoke[] | select(.id == "metal-model-load")],
+                smoke: [.matrices.smoke[] | select(.platform == "macos")],
```
ndizazzo: Four of the six `smoke_rows` in `ci/slices.yml` don't have a `platform` key: two-node-client, two-node-split, model-download, and metal-model-load. jq treats `null == "macos"` as a silent false rather than an error, so this projection returns `[]` on every plan.

`ci-macos-lane.yml` gates the whole macOS product-smoke slice on `matrices.smoke[0] != null`, so metal-model-load is skipped on every event and the lane still goes green. Line 287 does the same thing to Linux (it replaced `select(.id != "metal-model-load")`), which drops model-download permanently.

Stacked on the catalog desync, `core-cuda` ends up the only smoke row that can reach any job, while the ci/ci.md added here says the CUDA and Metal signals both stay required.

We should add `platform` to the four rows that are missing it, or go back to filtering on id.

.github/workflows/llama-upstream-canary.yml
```diff
@@ -153,22 +231,50 @@ jobs:
...
+        run: |
+          scripts/skippy-canary-live-matrix.sh --prepare 2>&1 | tee -a .deps/llama-canary-repair-battery.log
+
       - name: Supported-families certification battery (parity gate)
         id: battery
-        if: steps.prepare.outcome == 'success' && steps.sha.outputs.certify == 'true'
+        if: steps.prepare.outcome == 'success' && steps.sha.outputs.certify == 'true' && steps.family_patch.outcome == 'success'
         # continue-on-error so the battery-failure agent repair loop below can
         # run; the "Fail when the canary needs human attention" step keeps the
         # job red for any battery failure the repair does not resolve.
```
ndizazzo, line 281: (GitHub wouldn't let me anchor to the exact line, this is about the `run:` block right below.)

The `scripts/skippy-canary-live-matrix.sh --prepare 2>&1 | tee -a ...` line needs `set -o pipefail`. Without a `shell:` key GitHub runs it as `bash -e {0}`, so the step's exit status is tee's and this step can never fail.

The script exits 1 when a family row fails (line 370), but `steps.live_matrix.outcome` will always be `success`. The three pin-eligible gates below (341, 359, 378) then certify a broken llama.cpp pin, and the repair-loop and fail-red triggers at 428 and 453 never fire.

The battery step just below already does this correctly, so it looks like this one just got missed.

.github/actions/restore-product-integration-inputs/action.yml
```diff
@@ -0,0 +1,138 @@
...
+      run: |
+        set -euo pipefail
+        mkdir -p "$HOME/.models"
+        token="${HF_TOKEN:-${HUGGING_FACE_HUB_TOKEN:-}}"
+        curl_args=(--fail --location --show-error --retry 12 --retry-delay 20 --retry-max-time 600 --retry-all-errors)
+        if [[ -n "$token" ]]; then
+          curl_args+=(--header "Authorization: Bearer ${token}")
+        fi
+        for fixture in dense recurrent; do
+          file_var="${fixture^^}_FILE"
```
ndizazzo: `${fixture^^}` is bash 4 syntax and this action runs on macos-15, where `shell: bash` resolves to Apple's /bin/bash 3.2.57. There it's a parse error (`bad substitution`), and with `set -euo pipefail` on the step the whole macOS lane dies before either fixture downloads. Line 105 has the same thing for `url_var`.

`file_var="$(printf '%s' "$fixture" | tr '[:lower:]' '[:upper:]')_FILE"` works everywhere.

Currently hidden by the plan-ci projection bug, since the macOS job never gets scheduled at all.

While I'm here: this step also re-implements the resolve, cache-restore, curl, verify, cache-save pipeline that `restore-smoke-inputs` already does (and which this action calls for the binary half). The copy hardcodes 12/20/600 where the original honors `MODEL_DOWNLOAD_RETRY_*`. Since you already added `--github-output-prefix` to the resolver script, adding `artifact_id` and `output_prefix` inputs to `restore-smoke-inputs` and calling it twice would let this go away.

scripts/check-skippy-generated-family-patch.sh
```diff
@@ -0,0 +1,101 @@
...
+    --current "$GENERATED_PATCH_DIR/series.json" \
+    --output "$ARTIFACT_ROOT/changed-family-selection.json"
+fi
+
+python3 "$ROOT/scripts/skippy-rewriter-harness.py" \
+  --report "$FIRST_REPORT" \
+  --mode validate \
+  --patch-check "$patch_result" \
+  --patch-drift-gate fail \
+  --compile-result pass \
```
ndizazzo: These two are hardcoded `pass`, and the harness only fails on the literal string `fail`, so neither gate does anything.

The ctest above has a sequencing problem too. It runs at line 54, which is before the EXIT trap is installed and before `generate-skippy-family-patch.py` applies any edits, so it's verifying binaries built from the tree as it was *before* the rewrite. Then the generator applies its edits, diffs them, and the trap `git reset --hard`s them away. Nothing ever compiles or graph-verifies the rewritten tree.

Which leaves `diff -qr` against the checked-in shards as the only real gate on a 1500-line LibTooling pass over ~125 model files. That's fine for catching drift, but it means a newly proven edit on an upstream bump gets byte-compared and never actually built. I'd move the build and ctest to after the rewrite and feed the real results in here.

scripts/generate-skippy-family-patch.py
```diff
@@ -0,0 +1,329 @@
...
+GENERATOR_VERSION = "0.3.0"
+DIFF_HEADER = re.compile(r"^diff --git a/(src/models/[^ ]+) b/[^\n]+$", re.MULTILINE)
+
+
+def run(command: list[str], *, cwd: Path, capture: bool = False) -> str:
+    result = subprocess.run(
+        command,
+        cwd=cwd,
+        check=True,
+        text=True,
```
ndizazzo: This capture produces every committed patch byte and every sha256 in series.json, so it should probably be `text=True, encoding="utf-8", errors="strict", newline=""` (or capture bytes and decode explicitly).

Two things as written. `text=True` implies `newline=None`, so any CRLF in git's output collapses to LF unconditionally, which is the exact scenario 003c8fc0 patched around from the other direction. On an autocrlf checkout the shards silently lose their CRs and series.json records a digest for content that never existed on disk.

And decoding uses the ambient locale. The generated output does contain non-ASCII (0026-family-qwen3-next.patch has a U+2014), so under a non-UTF-8 locale the same git bytes re-encode at line 186 to a different sha256, and under a strict one it raises UnicodeDecodeError, which is a ValueError and isn't in the handler at 323. Certify runs on self-hosted runners, so the locale is whatever the host happens to have.

tools/skippy-stage-rewriter/main.cpp
```diff
@@ -0,0 +1,1502 @@
...
+            "stage_filtered && il_start > 0 ? build_inp_embd(nullptr) : " +
+                original_embedding,
+            sm, lang);
+      }
+
+      std::set<const IfStmt *> guarded_ifs;
+      for (const BinaryOperator *assignment : preloop_activation_assignments) {
+        const Stmt *statement = directChildContaining(
+            constructor_body, assignment, sm, lang);
+        if (const auto *conditional = llvm::dyn_cast_or_null<IfStmt>(statement)) {
```
ndizazzo: This guards the `if` condition without ever looking at `getElse()`, so on a non-first stage it diverts into the else arm rather than skipping the prelude.

For `if (scale_embeddings) { inpL = scale(inpL); } else { inpL = build_norm(inpL, ...); }` we emit `if ((!stage_filtered || il_start == 0) && (scale_embeddings))`, which on a middle stage forces the condition false and runs `build_norm` on the activation we just imported from the previous stage. Exactly backwards from what the guard is for.

Worse when the assignment lives in the else arm, since `directChildContaining` returns the enclosing `IfStmt` either way, so guarding the condition makes the else arm run on *every* stage. There's no `getElse` or `hasElseStorage` anywhere in the file, and the only fixture uses a bare `if` with no else, so nothing catches it. The builder still reports `transformable`, it compiles, and per the other note nothing graph-verifies it.

We should refuse the builder when the enclosing `IfStmt` has an else, at least until the guard understands both arms.

Separately, `fileOffset` at line 270 has a dead guard: it calls `sm.getSpellingLoc(location)` first, and that never returns a macro id, so the `location.isMacroID()` check below can't fire. Means a macro-expanded construct resolves to the macro *definition* and we plan an edit there. `sourceText` guards this correctly and returns "", so the two disagree and you can end up splicing a replacement built from an empty original into a macro body. The insert-only edits are the risky ones since they'd still compile.

And `std::set` at 1124 has no `#include <set>`, it's only building through a transitive LLVM header.

ndizazzo (changes requested):
I went through this one properly in a scratch worktree off the branch head (3358a392, diffed against main). The design direction is right, and the generic partitioner plus the plan and identity machinery is genuinely what we want here. But there are two places where the new mechanism doesn't actually connect to anything, and I don't think we can land it while that's true.

The tensor closure is the big one. Nothing loads from it. `filter_tensors_on_load` is still live, and `skippy-server/src/config.rs` still requires it for artifact-slice mode, so serving is still going through the coarse loader filter with the per-arch exception ladder (GEMMA4, GEMMA3N, QWEN4EXP PLE, NEMOTRON_H_MOE) that the design doc blames for the Granite bug in the first place. That means the closure and the load can silently disagree, which is the failure mode we set out to delete. I think we need a `skippy_model_load_from_plan_v1` (or whatever we want to call it) that binds the realized closure through `materialization_tensors`, and then the loader filter can go away for real. Left a note inline on the one place the closure is used today.

Admission is the other one, also inline. Short version: the planned side gets built by cloning the realized side, so the whole planned-vs-realized diff compares values to copies of themselves and can't fail.

Independent of the Skippy work, we've lost basically all product smoke coverage. `ci/slices.yml` is back on the `core` / `two-node-*` ids, but every job in `ci-linux-product-smoke-slice.yml` now gates on `product-integration-*`, and the old jobs got deleted. `grep -c product-integration ci/slices.yml` is 0. So the planner emits ids no job matches, the jobs wait on ids the planner can't emit, and the lane reports green having run nothing. Let's either land the catalog rows in this change or keep the old jobs around until they land. There's a second, separate hole in the plan-ci projection that I flagged inline.

Two readers got left behind by the package-v2 writer bump. Both are on unchanged lines so I couldn't put them inline:

- `crates/model-package/src/scripts/split-model-job.sh:376` reads `m['source_model']['path']` and sums `artifact_bytes` over `m['shared']` and `m['layers']` immediately after calling write-package. None of those keys exist in v2, and the structs are `deny_unknown_fields`, so the HF split job hits a KeyError and dies before it uploads anything.
- `preflight.rs:426` and `validate.rs:177` still hard-require `schema_version == 1`. So `preflight` now errors on every package write-package produces (skippy-quantize runs the two back to back), and `validate-package` fails at deserialization when it reaches `manifest.layers`.

The rewriter work is impressive, but I'd want its compile and graph-verify gates to be real before we lean on it during an upstream bump, and there's an else-arm bug in the embedding guard that would quietly produce wrong graphs on middle stages. Both inline.

One thing I should flag about my own coverage: I didn't get a clean pass over what the 33-shard patch recreation *removes* relative to the hand-written queue, so behavior deleted there is the dimension I'm least confident about.

crates/mesh-llm-host-runtime/src/runtime/local_package.rs
i386: Fixed in `8747286ef`. Package-v2 manifests now resolve `resident_tensor_ids` to canonical native tensor names, carry that exact set through ABI 0.1.52, and the native loader rejects both missing requested tensors and unrequested extras. The coarse layer-class ladder remains only for legacy direct-GGUF/old package compatibility; package-v2 admission and loading use the exact graph-derived closure. Rust runtime tests, CPU/Metal native builds, and 27/27 Skippy native tests pass.

crates/mesh-llm-host-runtime/src/runtime/stage_admission.rs
i386: Fixed in `8747286ef`. The production path no longer constructs `planned` by cloning `realized`. Admission now obtains an independent native discovery descriptor and an independent realization descriptor, then binds package identity, range, sidecars, tensor closure, and guarded profiles from the manifest/topology before comparison. Mismatch regressions cover the independently sourced fields; host-runtime tests pass 2969/2969.

crates/mesh-llm-host-runtime/src/mesh/stage_proto.rs
i386: Fixed in `8747286ef`. `activation_codec_policy` is propagated alongside `activation_codec` at every status construction/conversion site, including stage 0, rather than defaulting to `Fixed`. Attestation/status round-trip regressions cover `AutoLosslessV1`; the host-runtime suite passes.

crates/mesh-llm-host-runtime/src/mesh/stage_artifacts.rs
i386: Fixed in `8747286ef`. Peer `Status` requests now route through the live stage-control loop. The stale `cached_stage_statuses` response path was removed, so runtime teardown and generation changes are observed rather than served indefinitely from topology cache. Host-runtime tests pass 2969/2969.

crates/mesh-llm-host-runtime/src/inference/skippy/package.rs
i386: Fixed in `8747286ef`. If package-v2 tensor metadata cannot produce meaningful per-layer weights, the helper returns an empty result instead of an all-zero vector. Both consumers therefore take their established fallback estimation path. Regressions cover the all-unknown catalog case; coordinator/runtime tests pass.

crates/skippy-server/src/binary_transport/binary_messaging.rs
i386: Fixed in `8747286ef`. `EINVAL` is treated as a benign timeout only after the configured timeout interval has elapsed; immediate `EINVAL` returns the underlying error instead of spinning. Transport regressions cover both paths; `skippy-server` passes 694/694.

crates/skippy-protocol/src/binary/activation_codec.rs
i386: Fixed in `8747286ef`. RawF32 again preserves every input bit pattern, including NaN and infinities, on encode and decode. Lossy codecs reject non-finite frames so automatic policy falls back to RawF32. Protocol/runtime regressions cover both behaviors; `skippy-runtime` passes 118/118.

crates/skippy-protocol/src/validation.rs
i386: Fixed in `8747286ef`. Generation 8 status now uses a typed `Idle | Admitted { descriptor }` oneof. An admitted response cannot omit its package/plan/tensor/profile descriptor, and validation rejects generation/state mismatches. Protocol and host-runtime suites pass.

crates/skippy-model-package/src/package_v2.rs
i386: Fixed in `8747286ef`. The transform hook is executed before package writing, and the writer consumes the transformed artifact instead of rejecting the argument. Quantize/package regressions exercise the real hook path and verify the transformed bytes and manifest; repository CI validation passes 922 tests with 7 expected skips.

crates/skippy-model-package/src/package_v2.rs
i386: Fixed in `8747286ef`. `created_at_unix_secs` is excluded from the normalized package-id preimage while remaining available as manifest metadata. Two otherwise identical manifests with different creation times now produce the same ID; changing content still changes it. Package tests and full repository validation pass.

.github/actions/plan-ci/action.yml
i386: Fixed in `8747286ef`. CI smoke rows without an explicit platform are no longer silently filtered out; platform-specific selection handles the catalog shape used by two-node, download, and Metal load rows. Planner/action contract tests cover Linux and macOS projections; actionlint and repository CI validation pass.

.github/workflows/llama-upstream-canary.yml
i386: Fixed in `8747286ef`. The live-matrix workflow step now enables `pipefail`, so a failing `skippy-canary-live-matrix.sh --prepare` propagates through `tee` and blocks pin eligibility. Workflow contract tests and actionlint pass.

.github/actions/restore-product-integration-inputs/action.yml
i386: Fixed in `8747286ef`. The macOS action no longer uses Bash 4 `${var^^}` expansion; it derives both environment-variable names with portable `tr`, which works under Apple Bash 3.2. The action contract tests and actionlint pass. I kept the two explicit fixture restores because they are separate required inputs; the correctness issue here was the non-portable expansion.

scripts/check-skippy-generated-family-patch.sh
i386: Fixed in `8747286ef`. The checker now installs cleanup before mutation, applies the rewrite, builds the rewritten tree, runs the nine Skippy verifier targets (27/27 tests), records their actual outcomes, and fails unless both compile and test gates pass. The previous hardcoded `pass` values are gone. A full regeneration reports 152 sources, 147 builders, zero unsupported/errors, byte-identical shards, and a zero-edit second pass.

scripts/generate-skippy-family-patch.py
i386: Fixed in `8747286ef`. Git output is captured as raw bytes and decoded once with strict UTF-8, avoiding locale-dependent decoding and universal-newline rewriting. CRLF integration coverage and deterministic digest checks pass.

tools/skippy-stage-rewriter/main.cpp
i386: Fixed in `8747286ef`. The transformer now refuses conditional preludes with an `else` arm rather than emitting a guard that can redirect control into that arm on middle stages. The macro-source check uses expansion/spelling locations consistently, and the required `<set>` include is explicit. Positive and negative fixtures pass, and the full 152-source regeneration reports zero unsupported shapes or analyzer errors for the pinned tree.

ndizazzo (dismissed):
I reviewed current head `7b9aedab4` from a clean detached worktree. The earlier review fixes are present, but I still don't think this is mergeable yet.

1. **Package-v2 remote split serving is not wired end to end.** `startup_models.rs:1094-1107` turns a discovered `hf://` layer package into a `PathBuf` without resolving it. `local_split.rs:341-355` then sends that path to `resolve_split_runtime_package`, which requires either a local package-v2 directory or a local GGUF file (`local_package.rs:129-155`). The documented remote two-node path therefore fails before planning. The downstream selective download and peer-transfer code also still reads the v1 `/shared`, `/layers`, and `/projectors` layout (`package_download.rs:871-1001`, `artifact_transfer.rs:154-215`) rather than v2 `artifact_catalog` bindings. This needs one v2 path from remote manifest resolution through stage admission, selective fetch, and materialization, plus a two-node smoke from empty caches that proves each node receives only its admitted artifacts.

2. **The v2 writer still feeds v1-only consumers.** `skippy-model-package write-package` now always emits schema 2, but `skippy-quantize` immediately invokes `preflight` after writing (`skippy-quantize/src/main.rs:741-810`). That preflight deserializes the old manifest type and explicitly rejects any schema other than 1 (`skippy-model-package/src/preflight.rs:345-370,425-435`). `split-model-job.sh:376-404` has the same mismatch: after `write-package`, it reads `source_model.path`, `shared`, `layers`, and `projectors`, none of which are the v2 artifact catalog. Please make these consumers v2-aware or route them to the v2 verifier before relying on the new writer.

3. **#1681 is only partially covered.** Patch 0013 registers `llm_graph_input_attn_kv_iswa` and `llm_graph_input_attn_k_iswa`, and adds only a Gemma2 fixture. It does not add registration or planner coverage for the DSA-ISWA, MSA, and hybrid-ISWA builder paths identified in the still-open issue. Before merge, those paths need explicit request identities and fixtures or exact runtime evidence; otherwise reviewed sliding-window families can still fail graph-identity encoding.

4. **The exact head does not compile on Windows Vulkan.** Patch 0017 adds `std::find_if` in `llama-graph.cpp` without adding `<algorithm>`. The first definitive CI failure is [Windows runtime (vulkan)](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34160891383/job/101862433655); later Linux/macOS cancellations are fail-fast siblings, not separate failures.

My focused generator test passed. The repository Python suite passed 873 tests apart from six missing-PyYAML imports in the first invocation; those six passed separately under `uv --with pyyaml`. I did not count the cancelled sibling lanes as validation. The branch is also two commits behind `main`, so please rebase before the final exact-head replay.

i386:
Resolved the four blockers from the latest review and rebased onto current `main`.

1. **Remote package-v2 split serving:** split startup now resolves and validates remote package metadata, admits the exact stage artifact closure, fetches and verifies only those artifacts, carries the selected multi-part paths into runtime construction, and restricts peer serving to the stage's admitted artifacts. The new two-node Iroh test starts with a separate empty destination cache, fetches the manifest and exact admitted closure, verifies the primary/resident artifacts arrived, and proves an unowned artifact remains absent and is rejected by the source node.
2. **Package-v2 preflight/job wiring:** quantization now invokes the v2 verifier with `--source` / `--source-file`; the HF split job reads the v2 source identity and `artifact_catalog` instead of v1 shell fields and asserts the expected source repository.
3. **Architecture coverage:** the generated llama patch now registers request inputs for ISWA, DSA, DSA-ISWA, MSA, hybrid-ISWA, and MiniMax M3 MSA, with an MSA fixture and regenerated downstream patch context.
4. **Windows Vulkan:** added the missing `<algorithm>` include.

Pushed exact head `af1b795af13cc048357974edb9d10fae3e476e6e`.

Validation on the rebased tree:
- clean 52-patch llama replay and Metal native replay: 40/40
- full host runtime: 2,980 passed, 11 ignored; integration pair passed
- `skippy-server --no-default-features --lib`: 694 passed, 3 ignored
- model-package quantize and protocol suites passed
- package-v2 two-node exact-artifact transfer test passed
- split-smoke Python contract: 12 tests plus 5 subtests passed
- `cargo fmt --check`, `git diff --check`, and split-job `shellcheck` passed

CI has started on the new head.

i386:
CI exposed two test-only `StageConfig` initializers in `skippy-cache` that did not set the new `model_part_paths` field. I fixed both and added the Clippy allowance required after the topology builder gained its eighth explicit input.

The exact CI batch-2 Clippy crate set now passes locally (the local Rust 1.98-only `chunks_exact_to_as_chunks` lint was allowed for parity with CI); `skippy-cache` is 62/62. Force-pushed corrected head `3a7f1895db6d55ae719a503439d45ac2784b6ddb`; fresh CI is starting.

i386:
The next CI contracts run correctly caught that line shifts from the package-v2 edits had made the repository's existing console-print debt ratchet stale. No new console-print occurrence was introduced; I regenerated the ratchet and verified `repo-consistency no-console-print` passes.

Corrected head: `475457d040fab57f192dec434dbd8e91680a2d51`. Fresh CI is starting.

i386:
Clippy batch 0 then found the same new-field omission in the standalone `skippy-prompt` config writer. I added the empty `model_part_paths` default there and refreshed the line-number ratchet for the existing prompt console calls.

The exact batch-0 Clippy crate set now passes locally, as do the batch-2 set and `repo-consistency no-console-print`. Corrected head: `ae34980491759ec97a3b6d1aa0f9bae0d1a8ece3`.

i386:
The macOS Swift smoke exposed an integration mismatch introduced by current `main` (#1675): `runtime list --available --json` now returns `{ catalogs, runtimes }`, while `ci-prepare-native-runtime.sh` still required the former top-level list. I updated the selector to consume the new `runtimes` envelope while retaining compatibility with older list output, and updated its fixture to exercise the current shape.

Validation: `test_ci_prepare_native_runtime.py` 7 passed plus 3 subtests; shellcheck and the console-print consistency gate pass. Corrected head: `b9450e1ec78508968b1caf9877612232aea9961e`.

michaelneale (approved):
pre approving. needs some conflict resolving and ideally some removing of old code to get delta down as mentioned in #[1686](https://github.com/Mesh-LLM/mesh-llm/issues/1686) - or mayve as a fast follow

PR #1675, opened 2026-09-06
feat(runtime): explain native runtime catalog selection and rejection

Virgile-pct:
Second part of #1612, on top of #1623. Native runtime resolution now says which catalogs it consulted and why a candidate was rejected.

## What changes

- `NativeRuntimeCatalogSources` counts the artifacts each source contributed and renders itself. It is public through `load_release_manifest_with_sources` and travels on `NativeRuntimeInstallOutcome`: the human output prints the catalog lines, the JSON output gains a `catalogs` object.
- An explicit selection that cannot be satisfied fails with the catalogs consulted, every candidate that matched the selection with its rejection reasons, a count of the candidates that never matched, and a reminder that the requested backend was not replaced by another one.
- `runtime list --available` prints the catalogs it consulted, and the progress messages no longer assume that a bundle means no catalog. In JSON mode the listing becomes an object with `catalogs` and `runtimes` instead of a bare array (the only in-repo consumer, `scripts/rc-release-smoke.sh`, stores that output as an audit artifact without parsing it).
- Diagnostic URLs drop their fragment as well as their query and userinfo before they reach a report.
- `CandidateRejection` implements `Display`, so the CLI listing and the install diagnostics use the same wording.
- The console print allowlist is regenerated for the new diagnostic lines, and the env-mutation census is frozen at its new count (one more `NATIVE_RUNTIME_MANIFEST_URL_ENV` test site under the existing `MANIFEST_ENV_LOCK` guard), including the repository total the contract test asserts.

## Verification

On the Windows bench: `cargo test -p mesh-llm-runtime-install --features skippy-ffi/dynamic-runtime` (41 passed), `cargo test -p mesh-llm-native-runtime` (33 passed), `cargo check -p mesh-llm --tests`, fmt and clippy clean on the touched crates, `repo-consistency no-console-print`, `check-env-mutation-contract.py` together with its unit tests (201 sites).

Refs #1612, #1511.

## Summary by CodeRabbit

- **New Features**
  - Runtime installation and availability results now identify the catalogs consulted, including online, offline, and bundled sources.
  - JSON output includes catalog details and structured runtime resolution information.
  - Human-readable output displays catalog descriptions and clearer installation errors with compatibility explanations.

- **Bug Fixes**
  - Native runtime manifest loading progress now appears consistently when needed.
  - Catalog URLs and checksum errors provide safer, sanitized diagnostics.

- **Documentation**
  - CLI documentation now describes the updated JSON output and resolution error details.

### DIFF
--- a/crates/mesh-llm-commands/src/runtime_native.rs
+++ b/crates/mesh-llm-commands/src/runtime_native.rs
@@ -75,23 +75,17 @@ pub async fn run_native_runtime_list(
                 ..Default::default()
             })
             .await?;
-        if !json_output {
-            eprintln!("🔎 Catalogs consulted");
-            for line in sources.describe() {
-                eprintln!("   {line}");
-            }
-        }
         let profile = host_runtime_profile();
         let cache = native_runtime_cache(cache_dir)?;
         let mut resolver =
             NativeRuntimeResolver::new(mesh_version, profile.clone(), manifest.clone(), cache)
-                .with_bundle_dirs(sources.bundle_dirs);
+                .with_bundle_dirs(sources.bundle_dirs.clone());
         if let Some(skippy_abi_version) = configured.skippy_abi_version {
             resolver = resolver.with_skippy_abi_version(skippy_abi_version);
         }
         let evaluated = resolver.evaluate(&selection)?;
         let rows = available_runtime_rows(&manifest, &evaluated);
-        return formatter.render_available(&rows);
+        return formatter.render_available(&rows, &sources);
     }
 
     let installed = discover_local_native_runtimes(bundle_dirs, &cache)?;
--- a/crates/mesh-llm-commands/src/runtime_native.rs
+++ b/crates/mesh-llm-commands/src/runtime_native.rs
@@ -10,7 +10,7 @@ use mesh_llm_runtime_install::{
     CURRENT_MESH_VERSION, NativeRuntimeBundleInstallPolicy, NativeRuntimeDownloadProgressCallback,
     NativeRuntimeInstallOptions, NativeRuntimeManifestOptions, discover_local_native_runtimes,
     discover_native_runtime_bundle_dirs, host_runtime_profile, install_native_runtime,
-    load_release_manifest, native_runtime_cache,
+    load_release_manifest_with_sources, native_runtime_cache,
 };
 use mesh_llm_tui::terminal_progress::{
     ratio_complete_u64, render_inline_gauge_with_reserved_width,
@@ -64,21 +64,28 @@ pub async fn run_native_runtime_list(
     if available {
         let discovered_bundle_dirs = discover_native_runtime_bundle_dirs(bundle_dirs)?;
         print_configured_selector(configured, json_output);
-        if !json_output && manifest_path.is_none() && discovered_bundle_dirs.is_empty() {
+        if !json_output && manifest_path.is_none() {
             eprintln!("🔎 Loading native runtime release manifest");
         }
-        let manifest = load_release_manifest(NativeRuntimeManifestOptions {
-            mesh_version: mesh_version.to_string(),
-            manifest_path: manifest_path.map(Path::to_path_buf),
-            bundle_dirs: discovered_bundle_dirs.clone(),
-            ..Default::default()
-        })
-        .await?;
+        let (manifest, sources) =
+            load_release_manifest_with_sources(NativeRuntimeManifestOptions {
+                mesh_version: mesh_version.to_string(),
+                manifest_path: manifest_path.map(Path::to_path_buf),
+                bundle_dirs: discovered_bundle_dirs.clone(),
+                ..Default::default()
+            })
+            .await?;
+        if !json_output {
+            eprintln!("🔎 Catalogs consulted");
+            for line in sources.describe() {
+                eprintln!("   {line}");
+            }
+        }
         let profile = host_runtime_profile();
         let cache = native_runtime_cache(cache_dir)?;
         let mut resolver =
             NativeRuntimeResolver::new(mesh_version, profile.clone(), manifest.clone(), cache)
-                .with_bundle_dirs(discovered_bundle_dirs);
+                .with_bundle_dirs(sources.bundle_dirs);
         if let Some(skippy_abi_version) = configured.skippy_abi_version {
             resolver = resolver.with_skippy_abi_version(skippy_abi_version);
         }
@@ -143,7 +150,7 @@ pub async fn run_native_runtime_install(
     json_output: bool,
 ) -> Result<()> {
     let resolved_selection = resolve_runtime_selection(requested_runtime, configured)?;
-    if !json_output && manifest_path.is_none() && bundle_dirs.is_empty() {
+    if !json_output && manifest_path.is_none() {
         eprintln!("🔎 Loading native runtime release manifest");
     }
     if !json_output {
--- a/crates/mesh-llm-commands/src/runtime_native/formatters.rs
+++ b/crates/mesh-llm-commands/src/runtime_native/formatters.rs
@@ -93,7 +93,17 @@ impl RuntimeNativeFormatter for HumanFormatter {
 
     fn render_install_error(&self, error: &Error) -> Result<()> {
         eprintln!("❌ Native runtime install failed");
-        eprintln!("   Reason: {error}");
+        // Resolution failures carry a multi-line explanation (catalogs
+        // consulted, rejected candidates); keep every line aligned.
+        let mut lines = error.to_string();
+        lines.truncate(lines.trim_end().len());
+        let mut lines = lines.lines();
+        if let Some(first) = lines.next() {
+            eprintln!("   Reason: {first}");
+        }
+        for line in lines {
+            eprintln!("           {line}");
+        }
         eprintln!("   Try: mesh-llm runtime list --available");
         Ok(())
     }
@@ -153,6 +163,7 @@ impl RuntimeNativeFormatter for JsonFormatter {
             "status": install_status_label(outcome.status.clone()),
             "runtime": outcome.runtime,
             "resolution": outcome.resolution,
+            "catalogs": outcome.sources,
         }))
     }
 
@@ -268,6 +279,9 @@ fn print_install_human(outcome: &NativeRuntimeInstallOutcome) {
             eprintln!("   path: {}", outcome.runtime.path.display());
         }
     }
+    for line in outcome.sources.describe() {
+        eprintln!("   catalog: {line}");
+    }
 }
 
 fn print_doctor_human(report: &NativeRuntimeDoctorReport) {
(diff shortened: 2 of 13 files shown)


crates/mesh-llm-commands/src/runtime_native.rs
```diff
@@ -64,21 +64,28 @@ pub async fn run_native_runtime_list(
...
-        .await?;
+        let (manifest, sources) =
+            load_release_manifest_with_sources(NativeRuntimeManifestOptions {
+                mesh_version: mesh_version.to_string(),
+                manifest_path: manifest_path.map(Path::to_path_buf),
+                bundle_dirs: discovered_bundle_dirs.clone(),
+                ..Default::default()
+            })
+            .await?;
+        if !json_output {
```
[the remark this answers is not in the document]
Virgile-pct: Done in 32697eb: `runtime list --available --json` now emits an object with the `catalogs` consulted and the `runtimes` rows, and the human formatter prints the same catalog lines, so both modes come from one place. The JSON shape changes from a bare array to an object; the only in-repo consumer (`scripts/rc-release-smoke.sh`) stores the output as an audit artifact without parsing it, and the PR description now calls the change out.

crates/mesh-llm-runtime-install/src/manifest.rs
```diff
@@ -46,19 +46,75 @@ pub async fn load_release_manifest(
...
+            ));
+        } else if let Some(url) = &self.manifest_url {
+            match &self.remote_error {
+                Some(error) => lines.push(format!(
+                    "release catalog {url} unavailable, using bundles only: {error}"
+                )),
+                None => lines.push(format!(
+                    "release catalog {url} ({} artifacts)",
+                    self.manifest_artifacts
+                )),
```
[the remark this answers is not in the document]
Virgile-pct: Done in 32697eb: diagnostic URLs drop their fragment as well as their query and userinfo before they reach any report, with a test covering a fragment-held token.

crates/mesh-llm-runtime-install/src/install.rs
```diff
@@ -21,36 +24,109 @@ use tokio::io::AsyncWriteExt;
...
     .with_skippy_abi_version(skippy_abi_version)
-    .with_bundle_dirs(bundle_dirs)
-    .resolve(&options.selection)?;
-    install_resolved_runtime(&cache, resolution, &options).await
+    .with_bundle_dirs(sources.bundle_dirs.clone());
+    let resolution = match resolver.resolve(&options.selection) {
+        Ok(resolution) => resolution,
+        Err(err) => {
+            let evaluated = resolver.evaluate(&options.selection).unwrap_or_default();
+            bail!(
```
ndizazzo: `{err}` here is just the top-level Display, so the anyhow source chain underneath it gets dropped.

That chain has two consumers and they both read it the same way. `JsonFormatter::render_install_error` builds `"context"` from `error.chain().skip(1)`, and `build_fatal_error_event` in mesh-llm-tui does the same for the top-level fatal handler. Both now get an empty array for every resolution failure, plus a `message` with newlines stuffed into it.

Concretely: point a bundle dir at a corrupt `native-runtime.json`, `resolve` fails inside `candidate_artifacts()` with a contextual error, and `runtime install --json` used to show the underlying cause in `context`. Now it shows `[]`.

I'd attach the explanation as context on the original error rather than formatting it into a new one, something like `Err(err).context(describe_resolution_failure(...))`, so the chain survives and the explanation still shows up.

crates/mesh-llm-runtime-install/src/install.rs
```diff
@@ -21,36 +24,109 @@ use tokio::io::AsyncWriteExt;
...
     )
     .with_skippy_abi_version(skippy_abi_version)
-    .with_bundle_dirs(bundle_dirs)
-    .resolve(&options.selection)?;
-    install_resolved_runtime(&cache, resolution, &options).await
+    .with_bundle_dirs(sources.bundle_dirs.clone());
+    let resolution = match resolver.resolve(&options.selection) {
+        Ok(resolution) => resolution,
+        Err(err) => {
+            let evaluated = resolver.evaluate(&options.selection).unwrap_or_default();
```
ndizazzo: `unwrap_or_default()` here turns any non-selection failure into a confidently wrong message.

If `resolve` failed for an I/O reason then `evaluate(...)` fails the same way, we swallow it into an empty vec, and `describe_resolution_failure` takes the `matching.is_empty()` branch and tells the user "no catalog entry matches the requested selection (0 other candidates were not considered)".

So `mesh-llm runtime install cuda12` against an unreadable bundle manifest or an unreadable cache dir reports that their selection matched nothing in the catalog, when the catalog was never the problem. The `(0 other candidates...)` count makes it read even more definitively.

Worth distinguishing the two: if `evaluate` also failed, say we couldn't enumerate candidates and pass the error through.

crates/mesh-llm-runtime-install/src/install.rs
```diff
@@ -21,36 +24,109 @@ use tokio::io::AsyncWriteExt;
...
+        bundle_dirs: options.bundle_dirs.clone(),
+        // Offline installs (`allow_download: false`) must not reach out
+        // for the default catalog either; bundles and the cache are the
+        // only sources then. Explicit manifest URLs are still honoured.
+        allow_default_manifest_url: options.allow_download,
+    })
+    .await?;
     if manifest.artifacts.is_empty() {
-        bail!("no native runtime manifest entries found");
+        bail!(
```
ndizazzo: Both `bail!` sites embed newlines in the anyhow message, and that message goes somewhere that doesn't expect them.

`native_runtime.rs:239` does `tracing::warn!(error = %err, ...)`, and it also becomes `OutputEvent::Fatal { message }`. So `mesh-llm serve` on a host with no compatible runtime now emits a single warn record whose `error` field spans about 20 lines, which breaks line-oriented log consumption in the non-JSON formatter.

If we move the explanation into the error's context (see my other note) this sorts itself out, since `context` is already a list of separate entries.

crates/mesh-llm-runtime-install/src/install.rs
```diff
@@ -21,36 +24,109 @@ use tokio::io::AsyncWriteExt;
...
+/// that never matched the selection (another backend, another id) are
+/// summarised as a count so the explanation stays focused on what the user
+/// asked for.
+pub(crate) fn describe_resolution_failure(
+    sources: &NativeRuntimeCatalogSources,
+    selection: &RuntimeSelection,
+    evaluated: &[CandidateEvaluation],
+) -> String {
+    let mut lines = vec![describe_catalogs(sources)];
+    let (matching, others): (Vec<_>, Vec<_>) = evaluated.iter().partition(|candidate| {
```
ndizazzo: This filter doesn't narrow anything for the most common case. `selection_mismatch()` (resolver.rs:623) returns `None` for every artifact under `RuntimeSelection::Recommended`, so `matching` ends up being the entire merged catalog.

Which means for a plain `mesh-llm runtime install` with no arguments, the "focused" explanation prints one `candidate ...` line per artifact in the release manifest, macOS/metal and Windows entries included, rejected only for `OS mismatch`. The `others` summarisation only ever engages when someone passes an explicit `Backend` or `Id`.

Probably want to pre-filter on OS and arch compatibility before deciding what's "matching", so the recommended path shows the handful of candidates that were actually plausible.

crates/mesh-llm-runtime-install/src/manifest.rs
```diff
@@ -46,19 +46,75 @@ pub async fn load_release_manifest(
...
+            let dirs = self
+                .bundle_dirs
+                .iter()
+                .map(|dir| dir.display().to_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+            // Every bundle directory carries exactly one runtime manifest;
+            // say how many of them were not already listed by the catalog.
+            lines.push(format!(
+                "bundle directories ({} runtimes, {} not already in the catalog): {dirs}",
```
ndizazzo: Small thing: this reports `bundle_dirs.len()` against `bundle_artifacts`, but `append_bundle_artifacts` dedupes against `artifacts`, which already holds earlier *bundle* entries and not just catalog ones.

So if the same runtime turns up both via an explicit `--bundle-dir` and via the exe-adjacent `native-runtimes/` path (discovery dedupes by canonical path, not by runtime identity) we print "bundle directories (2 runtimes, 1 not already in the catalog)" even when the release catalog has neither of them. Reads like the catalog already had one.

crates/mesh-llm-commands/src/runtime_native/formatters.rs
```diff
@@ -136,8 +163,15 @@ impl RuntimeNativeFormatter for HumanFormatter {
 }
 
 impl RuntimeNativeFormatter for JsonFormatter {
-    fn render_available(&self, rows: &[AvailableRuntimeRow]) -> Result<()> {
-        print_json(rows)
+    fn render_available(
```
ndizazzo, line 169: This changes `runtime list --available --json` from a bare array to `{catalogs, runtimes}`, which is an unversioned break of the shape documented at `website/src/docs/pages/CLI.md:592`. The docs aren't updated in this PR.

Our only in-repo consumer (`scripts/rc-release-smoke.sh:211`) just tees the output without parsing, so CI won't catch it, but anyone scripting against the documented shape will.

I'm fine with the new shape, it's better. Let's update CLI.md in the same change so the docs and the output agree.

Related: the success path gained the structured `"catalogs"` object, but the failure path still stuffs the whole catalog explanation into free-text `message` while `context` comes out empty. A machine consumer can't read the new diagnostics structurally in the one case they're actually needed.

ndizazzo (commented):
This is a real improvement, the catalog and candidate explanation is exactly what was missing when a resolve fails, and having `runtime list --available` say which catalogs it merged is nice. I checked out the branch and ran the tests (42 in mesh-llm-runtime-install, 227 in mesh-llm-commands lib, all green) and verified the console-print allowlist line numbers line up.

My main concern is that the failure path lost information the JSON consumers depend on. The new `bail!` calls flatten the resolver error down to its top-level Display, so the anyhow source chain is gone, and both readers of that chain do the same `error.chain().skip(1)` (in `render_install_error` and in `build_fatal_error_event` over in mesh-llm-tui). So `runtime install --json` now reports an empty `"context"` for exactly the failures where the cause matters most. A corrupt `native-runtime.json` in a bundle dir used to surface the underlying parse error there and doesn't anymore. Details inline.

One more that I couldn't leave inline because it's on an unchanged line: `url_without_query` learned about fragments in this PR, but `release_manifest_checksum_url` (manifest.rs:264) still splits on `?` only. With `MESH_LLM_NATIVE_RUNTIME_MANIFEST_URL=https://host/native-runtimes.json#tok` the checksum URL comes out as `...native-runtimes.json#tok.sha256`, reqwest drops the fragment, so we fetch the manifest body and try to parse it as a checksum. `normalize_sha256` then rejects it with a message that interpolates `{value}`, which dumps the whole manifest JSON into the error. Works against the redaction this PR is adding, so worth handling in the same place.

Nothing here is scary, mostly a few spots where we end up reporting the wrong cause.

Virgile-pct:
Addressed in 2b493e04f: the explanation is now a typed `NativeRuntimeResolutionError` attached as the outermost context of the resolver's error (one-line `Display`, chain preserved, `downcast_ref` for the JSON), enumeration failures are named as such, plausible candidates exclude other platforms as well as other selections, the bundle line counts what was added and where the duplicates were already listed, `release_manifest_checksum_url` drops fragments and `normalize_sha256` no longer echoes the value, and CLI.md documents the JSON shapes. Details on each thread.

Gates on Windows 11 at this head: cargo fmt, 47 tests in mesh-llm-runtime-install (new: recommended selection, enumeration failure, checksum fragment, sha256 message, bundle duplicates), clippy clean on mesh-llm-runtime-install, `cargo check -p mesh-llm-commands`, `repo-consistency no-console-print` regenerated, no new env-mutation site so the census is unchanged. Not re-run here: the mesh-llm-commands unit tests do not link on this bare Windows checkout (static llama archives), so the 227 you ran are covered on my side only by `cargo check -p mesh-llm-commands` and the ratchet; the formatter change is small, but that is the gap.

crates/mesh-llm-runtime-install/src/install.rs
Virgile-pct: Fixed by making the explanation the outermost context of the resolver's error instead of a new error. `NativeRuntimeResolutionError` (summary, selection, catalogs, the plausible candidates with their rejection reasons, a set-aside count and an enumeration flag) is attached with `err.context(...)`, so `chain().skip(1)` yields the resolver's verdict and the deeper causes again, and `downcast_ref` recovers the structure. One precision on the scenario: a corrupt `native-runtime.json` in a bundle dir fails earlier, in `load_release_manifest_with_sources` (`read bundled native runtime <dir>` with the parse error underneath), and that path already propagated its chain; the enumeration failure that went through `bail!` was the cache read inside `candidate_artifacts`, and it keeps its cause now. The existing selection test checks the chain, the single-line `Display` and the downcast.

crates/mesh-llm-runtime-install/src/install.rs
Virgile-pct: Right. When `evaluate` fails after `resolve` failed, the error is now `NativeRuntimeResolutionError::enumeration_failed`: the summary says the candidates could not be enumerated, the JSON carries `enumeration_failed: true` with no candidate list, the explanation says the catalog is not at fault, and the cause stays in the chain. Test `enumeration_failure_is_not_reported_as_a_catalog_mismatch`.

crates/mesh-llm-runtime-install/src/install.rs
Virgile-pct: Both `bail!` sites are gone. The empty-catalog case is the same typed error (`empty_catalog`) and the resolution case is the context wrapper, so `Display` is one line everywhere: `tracing::warn!(error = %err)` and `OutputEvent::Fatal` get the summary, the explanation lives in the structure and the causes in the chain. The test asserts `to_string()` contains no newline.

crates/mesh-llm-runtime-install/src/install.rs
Virgile-pct: Changed what counts as a plausible candidate: no `SelectionMismatch`, and no `OsMismatch`, `ArchMismatch` or `TargetTripleMismatch` either. Everything else is only counted as `set_aside`, with one line saying those were built for another platform or did not match the selection. For a plain `runtime install` on Windows that leaves the Windows entries. Test `recommended_selection_sets_aside_other_platforms_instead_of_listing_them` checks that the Linux entry is neither listed nor named.

crates/mesh-llm-runtime-install/src/manifest.rs
Virgile-pct: Counted at the merge now: `append_bundle_artifacts` returns what was added, what the release catalog already listed and what an earlier bundle directory had already contributed (it can tell because the catalog entries are the prefix of the artifact list). The three counts land on `NativeRuntimeCatalogSources` as `bundle_artifacts`, `bundle_duplicates_of_catalog` and `bundle_duplicates_of_bundles`, and the line reads `bundle directories (2 runtimes: 1 added to the catalog, 1 duplicates of another bundle directory): ...`. Test `a_runtime_reached_through_two_bundle_directories_is_counted_as_a_bundle_duplicate`; the bundled-and-published test now checks the catalog count too.

crates/mesh-llm-commands/src/runtime_native/formatters.rs
Virgile-pct: CLI.md now documents the `{catalogs, runtimes}` object of `runtime list --available --json`, the `catalogs` object of `runtime install --json`, and the `resolution` field of its error (null for anything that is not a resolution failure, `context` kept in every case). On the failure path the JSON gets that `resolution` object from the downcast, so the diagnostics are readable structurally in the one case they are needed.

i386:
Follow-up on my review — I re-read the branch at `2b493e04f` and each of the three points is addressed:

1. **Cause chain** — `NativeRuntimeResolutionError` carries the structured explanation (catalogs, plausible candidates with rejection reasons, set-aside count, enumeration-failure flag) and travels as the *outermost context* of the resolver's own error: single-line `Display` on top, resolver verdict and deeper causes preserved underneath. `render_install_error` and the TUI's `error.chain().skip(1)` get their causes back, and `--json` consumers recover the structure via `downcast_ref`. The enumeration-failure case is correctly not presented as a catalog mismatch, and the original cause survives in the chain (tested).

2. **Checksum URL fragments** — `release_manifest_checksum_url` now splits the fragment off before appending `.sha256`, keeping the query. Both cases covered in `release_manifest_checksum_url_drops_fragments`.

3. **`normalize_sha256` echo** — the error no longer interpolates the value; it reports the length and the expectation instead (`invalid_sha256_error_does_not_echo_the_value` asserts a manifest body doesn't leak). Good catch on your side to fold the redaction theme through: diagnostic URLs now strip fragment as well as query and userinfo, with tests.

Also verified: the `runtime list --available --json` shape change (bare array → `{catalogs, runtimes}` object) is called out in CLI.md and the only in-repo consumer (`rc-release-smoke.sh`) stores the output as an artifact without parsing it, so nothing in-repo breaks.

CI green on this head (5/5 at 09-07 10:16). Ready for re-review from my side.

PR #1685, opened 2026-09-08
fix: serialize materialized cache writes on Windows

### LINKED ISSUE #1682: Windows: MaterializedOutputLock holds the file open but never locks it, so the materialized cache has no mutual exclusion
`skippy-runtime/src/package/materialized_cache.rs` guards materialization with `MaterializedOutputLock`. On unix the lock is a `flock` on the lock file. On other targets the struct still opens the file and keeps it in its `file` field for the lock's lifetime, but no lock is taken: there is no `LockFileEx` call, and nothing else serializes the writers.

Consequences on Windows:

- two mesh-llm processes materializing the same package concurrently (for example two nodes on one machine, or a node restarting while a previous instance is still finishing) can both pass the guard and write the same output directory at the same time;
- the `file` field is dead code on Windows, which is how #1659 ran into it (that PR only silences the lint; it does not change the behavior).

Expected: the same mutual exclusion on every platform. The Windows equivalent of `flock` is `LockFileEx` / `UnlockFileEx` on the opened handle (`std::os::windows::io::AsRawHandle`, `windows-sys` `Storage::FileSystem`), with `LOCKFILE_EXCLUSIVE_LOCK` and the same blocking semantics as the unix path; the `fs4` or `fs2` crates wrap exactly that if a dependency is acceptable.

Not reproduced live: this is from reading the code while preparing #1659. I can build a two-process reproduction on the Windows bench if that helps prioritize it.

Refs #1659.

Bortlesboat:
## Original problem

On Windows, two processes materializing the same model output could both pass `lock_output`: the lock file was opened, but the exclusive lock existed only under `cfg(unix)`.

Fixes #1682.

## Diagnostics

A Windows regression starts a second process and tries to acquire the same lock. On the original code that process succeeds while the first guard is alive. With this change it receives `WouldBlock`, can acquire the lock after guard drop, and is excluded again after reacquisition.

## Fix

Use the standard library's `File::lock` and `File::unlock` on the existing read/write handle. This retains the guard lifetime across output publication and cache-record writing, preserves Unix flock behavior, and adds Windows locking without a dependency or cache-format change.

## Validation

- [x] Ran the relevant local checks; limits are recorded below.
- UI evidence is not applicable to this filesystem change.
- Windows: 113 `skippy-runtime` library tests passed, with one ignored subprocess helper and three unchanged native-runtime tests filtered out. The helper's caller verifies that exactly one probe test actually ran.
- `cargo check --locked -p skippy-runtime --features dynamic-native-runtime`
- `cargo clippy --locked --no-deps -p skippy-runtime --all-targets --features dynamic-native-runtime -- -D warnings`
- `cargo check --locked -p mesh-llm`
- `cargo fmt --all --check` and `git diff --check`

The unfiltered baseline already failed these three tests because the native runtime library was not loaded: `cache_append_extends_the_committed_history`, `cache_drafts_from_committed_history_and_never_mutates_for_a_prefix`, and `invalid_selected_backend_device_fails_before_model_open`. The passing run uses `--skip` for those tests only.

Full-host `cargo clippy --locked -p mesh-llm --all-targets -- -D warnings` stops at the pre-existing `needless_return` in `crates/mesh-llm-events/src/audit.rs:469`; that file is identical to the base commit. Linux/macOS and live model inference were not run locally.

On the first Windows build containing this change, concurrent starts targeting one materialized output should serialize without publication/cache-record races. Check for new `lock materialized cache` errors. If locking regresses on a supported filesystem, revert this commit while investigating that filesystem's lock support.

## Summary by CodeRabbit

* **New Features**
  * File locking now works consistently across supported platforms.

* **Bug Fixes**
  * Improved materialized cache locking to prevent concurrent processes from accessing output simultaneously.
  * Locks are reliably released when no longer needed.
  * Lock behavior now remains effective across separate processes until the lock is explicitly released.

### DIFF
--- a/crates/skippy-runtime/src/package/materialized_cache.rs
+++ b/crates/skippy-runtime/src/package/materialized_cache.rs
@@ -4,9 +4,6 @@ use std::{
     sync::atomic::{AtomicU64, Ordering},
 };
 
-#[cfg(unix)]
-use std::os::fd::AsRawFd;
-
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 
@@ -96,10 +93,7 @@ pub(super) struct MaterializedOutputLock {
 
 impl Drop for MaterializedOutputLock {
     fn drop(&mut self) {
-        #[cfg(unix)]
-        unsafe {
-            let _ = libc::flock(self.file.as_raw_fd(), libc::LOCK_UN);
-        }
+        let _ = self.file.unlock();
     }
 }
 
@@ -117,13 +111,8 @@ pub(super) fn lock_output(output: &Path) -> Result<MaterializedOutputLock> {
         .open(&lock_path)
         .with_context(|| format!("open materialized cache lock {}", lock_path.display()))?;
 
-    #[cfg(unix)]
-    unsafe {
-        if libc::flock(file.as_raw_fd(), libc::LOCK_EX) != 0 {
-            return Err(std::io::Error::last_os_error())
-                .with_context(|| format!("lock materialized cache {}", lock_path.display()));
-        }
-    }
+    file.lock()
+        .with_context(|| format!("lock materialized cache {}", lock_path.display()))?;
 
     Ok(MaterializedOutputLock { file })
 }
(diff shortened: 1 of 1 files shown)


Virgile-pct (commented):
Tested on Windows 11, the machine behind #1682, at cb78025be on top of main 3101b023b:

- `cargo test -p skippy-runtime --lib --features dynamic-native-runtime -- package::materialized_cache`: the new cross-process test passes here, the child probe gets `WouldBlock` while the guard is alive and acquires the lock after the drop.
- The full `skippy-runtime` lib suite: 113 passed, 3 failed, and the three are the native-runtime tests that already fail on this checkout without a prepared llama build (identical on `main` on this machine), so nothing regressed.
- `cargo clippy --no-deps -p skippy-runtime --all-targets --features dynamic-native-runtime -- -D warnings` and `cargo fmt --check`: clean.

The approach is the right one: `File::lock` and `File::unlock` from std give `flock` on unix and `LockFileEx` on Windows without a new dependency, the guard keeps the same lifetime as before, and CI runs plain `cargo test`, so the self-spawning probe works there as well.

One thing to fold in at rebase time: #1659, approved and being rebased by Nick, adds `#[cfg_attr(not(unix), allow(dead_code))]` plus a comment on `MaterializedOutputLock::file` saying the handle is only held open outside unix. With this change the field is read on every platform through `unlock()`, so whichever of the two lands second should drop that attribute and its comment rather than keep a stale note. I can send that one-liner on the #1659 side if this one merges first.

Thanks for picking it up so quickly.

michaelneale:
Thanks for this, really handy having someone look at windows etc

michaelneale (approved):
Surprised how simple. Weird how old Unix code had unsafe and compiler macro for what shoud be rust std lib

PR #1688, opened 2026-09-08
fix: use served model IDs in ready commands

### LINKED ISSUE #1594: Ready output advertises a Goose/pi model ID that the node's own API 404s (--gguf startups)
**Type:** Bug — UX / onboarding
**Priority:** P2
**Effort:** S (half a day)
**Found in:** v0.76.0-rc9 release validation (mesh1, carrack)

## Problem

When a node is started with `--gguf <path>`, the ready output hands the operator a copy-paste Goose and pi command whose model ID the node's own API rejects.

Emitted at ready:

```
goose_command: GOOSE_PROVIDER=openai OPENAI_HOST=http://127.0.0.1:9337 OPENAI_API_KEY=mesh \
  GOOSE_MODEL=/home/me/.cache/models/llama-3.2-1b-instruct-q4_k_m.gguf goose session
pi_command:    mesh-llm pi --host 127.0.0.1:9337 --model '/home/me/.cache/models/llama-3.2-1b-instruct-q4_k_m.gguf'
```

Posting that exact model ID back to the same node:

```
HTTP 404
{"error":{"message":"model '/home/me/.cache/models/llama-3.2-1b-instruct-q4_k_m.gguf' not found (no local or remote host serving this model)",
          "type":"invalid_request_error","code":"model_not_found"}}
```

The ID that actually works is the resolved one from `/v1/models`, e.g. `local-gguf/sha256-f90897d3a4d7e185`.

## Steps to reproduce

1. `mesh-llm serve --gguf /path/to/model.gguf`
2. Copy the `GOOSE_MODEL=` value out of the ready output.
3. `POST /v1/chat/completions` with that value as `"model"`.

**Expected:** 200, or a command that works when pasted.
**Actual:** 404 `model_not_found`.

## Root cause

`crates/mesh-llm-host-runtime/src/runtime/startup_handles.rs:903-911` builds both strings from `self.primary_model`, which for a `--gguf` startup is the CLI-supplied file path rather than the resolved served model ID.

```rust
let pi_command = Some(format!(
    "mesh-llm pi --host 127.0.0.1:{} --model {}",
    self.api_port, single_quote_shell_arg(&self.primary_model)));
let goose_command = Some(format!(
    "GOOSE_PROVIDER=openai OPENAI_HOST={} OPENAI_API_KEY=mesh GOOSE_MODEL={} goose session",
    self.api_url, self.primary_model));
```

## Suggested direction

Use the resolved model ID — the same string `/v1/models` reports, already tracked as `runtime_status.primary_model` — when building these two commands. Worth a test that asserts the advertised `GOOSE_MODEL` / `--model` value resolves against `/v1/models` for a `--gguf` startup, since a path only *looks* right.

Bortlesboat:
## Title

Use served model IDs in ready commands

## Original problem

After starting with `--gguf`, the ready output uses the file path as the Goose/Pi model ID. Pasting that command requests a model the node does not serve.

## Diagnostics

The regression supplies a file-path startup reference and a different canonical served ID. Before the fix, the Pi command still contains the path. The test checks both primary-first and secondary-first completion, including a cloned readiness reporter.

## Fix

Record each slot's loaded model ID on its existing readiness edge. Build the commands from the primary slot after every startup model is ready. The ID comes from the same publication path used to register the loaded runtime.

For `mesh-llm serve --gguf /path/to/model.gguf`, the regression verifies these commands when the resolved ID is `local-gguf/sha256-f90897d3a4d7e185`:

```sh
mesh-llm pi --host 127.0.0.1:9337 --model 'local-gguf/sha256-f90897d3a4d7e185'
GOOSE_PROVIDER=openai OPENAI_HOST=http://127.0.0.1:9337 OPENAI_API_KEY=mesh GOOSE_MODEL=local-gguf/sha256-f90897d3a4d7e185 goose session
```

Fixes #1594.

## Validation

- [x] Ran the relevant local checks, with the limitation below.
- [x] No visual UI change; exact command strings are asserted in the regression.

13 relevant host-runtime tests pass: nine serving-surface tests, three startup-failure-policy tests and the quitting-during-startup test. `cargo check` for `mesh-llm-host-runtime` and `mesh-llm`, and `cargo fmt --all --check`, pass.

The required strict Clippy command fails on Windows in unchanged dependencies: `audit.rs:469` (`needless_return`) and `materialized_cache.rs:94` (`dead_code`). I reproduced both with the same command on the clean base commit. No live GGUF inference or HTTP smoke was run.

During native release validation, compare the emitted model IDs with `/v1/models` and exercise the copied commands. Local validation covers readiness-event generation and shutdown suppression.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

## Summary by CodeRabbit

* **Bug Fixes**
  * Runtime readiness notifications now consistently reference the correct primary served model, regardless of which model finishes loading first.
  * Startup status reporting now associates each readiness event with the model that became ready.
  * Launch commands generated after all required models are ready now use the intended served model.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/runtime/run_auto.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/run_auto.rs
@@ -1627,7 +1627,6 @@ pub(super) async fn run_auto(ctx: RunAutoContext) -> Result<()> {
     let primary_model_name = requested_model_names.first().cloned().unwrap_or_default();
     let startup_ready_reporter = StartupReadyReporter::new_with_failure_policy(
         &requested_model_names,
-        primary_model_name.clone(),
         api_ready_url,
         ready_console_url,
         ready_api_port,
--- a/crates/mesh-llm-host-runtime/src/runtime/serving_surface.rs
+++ b/crates/mesh-llm-host-runtime/src/runtime/serving_surface.rs
@@ -368,7 +368,6 @@ pub(crate) fn assert_passive_path_immediate_spawn_behavior() {
 pub(crate) fn assert_quitting_during_startup_cancels_without_late_ready_render() {
     let reporter = StartupReadyReporter::new(
         &["Qwen3-8B-Q4_K_M".to_string()],
-        "Qwen3-8B-Q4_K_M".to_string(),
         "http://127.0.0.1:9337".to_string(),
         Some("http://127.0.0.1:3131".to_string()),
         9337,
@@ -376,7 +375,9 @@ pub(crate) fn assert_quitting_during_startup_cancels_without_late_ready_render()
     );
     reporter.mark_shutdown_requested();
     assert!(
-        reporter.mark_ready_and_build_event(0).is_none(),
+        reporter
+            .mark_ready_and_build_event(0, "Qwen3-8B-Q4_K_M")
+            .is_none(),
         "startup shutdown should cancel any late RuntimeReady emission"
     );
 }
@@ -386,7 +387,6 @@ pub(crate) fn assert_startup_ready_reporter_waits_for_rust_owned_model_ready_edg
     let models = vec!["model-a".to_string(), "model-b".to_string()];
     let reporter = StartupReadyReporter::new(
         &models,
-        "model-a".to_string(),
         "http://127.0.0.1:9337".to_string(),
         Some("http://127.0.0.1:3131".to_string()),
         9337,
@@ -394,16 +394,16 @@ pub(crate) fn assert_startup_ready_reporter_waits_for_rust_owned_model_ready_edg
     );
 
     assert!(
-        reporter.mark_ready_and_build_event(0).is_none(),
+        reporter.mark_ready_and_build_event(0, "model-a").is_none(),
         "one model-ready edge must not replace the remaining Rust-owned readiness edges"
     );
     assert!(
-        reporter.mark_ready_and_build_event(0).is_none(),
+        reporter.mark_ready_and_build_event(0, "model-a").is_none(),
         "a repeated edge for one startup slot must not mark a different slot ready"
     );
     assert!(
         matches!(
-            reporter.mark_ready_and_build_event(1),
+            reporter.mark_ready_and_build_event(1, "model-b"),
             Some(OutputEvent::RuntimeReady { .. })
         ),
         "RuntimeReady should appear only after every startup model hits the Rust-owned ready path"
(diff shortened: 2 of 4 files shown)


ndizazzo (approved):
TY for the fix!

michaelneale:
Reviewed on Mic's behalf. The core fix is right: deleting the `primary_model` field and building both command strings from the *recorded served* id (`startup_handles.rs:897-905`) is the actual fix for the bug, and the single production call site now passes the resolved name (`startup_loop.rs:916`). `git grep mark_ready_and_maybe_emit|mark_ready_and_build_event` shows one production caller and the rest tests, so no branch was missed.

I checked the panic surface, since this indexes slot 0:

```rust
// startup_handles.rs:877-887
let mut ready_by_model = self.ready_by_model.lock().expect("startup readiness mutex poisoned");
*ready_by_model.get_mut(readiness_index)? = Some(loaded_model.to_string());
if ready_by_model.iter().any(Option::is_none) {
    return None;
}
(ready_by_model.len(), ready_by_model[0].as_ref()?.clone())
```

`ready_by_model[0]` can't panic — `get_mut(readiness_index)?` has already proved the vec non-empty — and the `as_ref()?` is unreachable after the `any(is_none)` check. Multiple-model semantics (primary = slot 0 = first *requested* model, independent of load order) are directly tested both orders by `startup_ready_commands_use_the_primary_served_model_in_either_load_order` (`serving_surface.rs:686-732`), including idempotence. Good test.

**One thing I'd like locked down before merge.** The zero-model contract silently inverted. Previously `ready_by_model.iter().all(...)` on an empty vec is `true`, so a call would emit `RuntimeReady { models_count: Some(0) }`. Now `get_mut(readiness_index)?` returns `None` on an empty vec, so `RuntimeReady` can never be emitted — same for an out-of-range `readiness_index`, which previously fell through to the `all()` check.

I could not find a live path that regresses: the reporter is only driven from `startup_publish_loaded_runtime` (so a model did load), and the empty-`startup_specs` path emits `PassiveMode`/`Ready` instead (`run_auto.rs:1615-1627`). So I'm not calling this a bug. But it's an unremarked behaviour change with no test — `git grep` at this head shows no coverage for an empty `models` slice or an out-of-range index. Please add either a one-liner asserting the new contract:

```rust
assert!(StartupReadyReporter::new(&[], ...).mark_ready_and_build_event(0, "x").is_none());
```

or a comment on `mark_ready_and_build_event` stating that an unknown index is a deliberate no-op. Cheap now, and it stops someone "fixing" it back later.

**Nits:** `run_auto.rs:1606`'s `primary_model_name` is still live (used via `RunAutoRuntimeLoopContext::primary_model_name` at `:1128`/`:1639`), so not dead — just no longer connected to ready commands, which reads a bit confusingly next to this change. And `single_quote_shell_arg` is applied to `pi_command` but not to the `GOOSE_MODEL=` value in `goose_command` — pre-existing, not yours, but served ids like `local-gguf/sha256-...` widen the blast radius slightly. Worth a follow-up.

Source reading only, no build.

PR #1689, opened 2026-09-08
fix: retain device placement in native log summaries

### LINKED ISSUE #1304: Summarized native logs drop per-GPU model buffer sizes and discard layer-to-device names
🤖 Filed by MeshBuilder on behalf of @michaelneale.

## Problem

Our summarized native log hides multi-GPU placement and per-device model buffer allocation. Both fields were needed to diagnose #1298 and neither survives into the summarized JSON, which cost a debugging cycle (and, because the Estonia rehearsal nodes were destroyed, the raw logs that did contain them are gone).

Two independent drops:

**1. `CUDAn model buffer size` has no matching arm.** `summarize_native_log_line` (`crates/skippy-runtime/src/logging.rs:476-489`) gates the `memory` category on:

```rust
line.contains("VRAM") || line.contains("vram") || line.contains("mem_alloc")
    || line.contains("_Mapped model buffer size")
    || (line.contains("GPU") && line.contains("memory"))
    || line.contains("compute buffer size")
    || line.contains("scratch buffer")
```

`load_tensors: CUDA0 model buffer size = 4321.00 MiB` matches none of these — not `_Mapped` (that arm is the CPU line), not `compute buffer size`, no `VRAM`, no `GPU`+`memory`. So the CPU mapped buffer is reported and the per-GPU model buffers are silently dropped. Confirmed empirically in the rehearsal evidence (`RESEARCH/evidence-qwen38/`): `CPU_Mapped model buffer size = 333.84 MiB` appears, `CUDAn model buffer size` appears zero times, while `CUDAn KV buffer size` and `CUDAn compute buffer size` do appear — exactly the arms that exist.

**2. `layer N assigned to device X` is consumed by a progress counter and the device name is discarded.** `parse_layer_assign_index` (`:572`) extracts only the layer number, and the handler at `:324-340` folds it into `layer_assign_progress`, emitting an aggregate "layers" progress event. The `CUDA3` / `CPU` half of the line — the actual layer→device map — is never emitted.

## Impact

Multi-GPU tensor placement is invisible in our own summarized logs. For any split/multi-device bug, the first question is "which layer went to which device, and how much landed on each" and our diagnostics cannot answer it. Users and maintainers must know to go find `<runtime-root>/<pid>/logs/skippy-native.log` — and if the machine is ephemeral (cloud rehearsal), that evidence is unrecoverable.

## Suggested fix

- Add a `model buffer size` arm to the memory category so `CUDAn`/`Metal`/etc. model buffers are captured alongside the existing `_Mapped` one.
- Keep the layer-assign progress aggregation, but also carry the device in the event — e.g. capture the device token in `parse_layer_assign_index` (returning `(index, device)`), and emit a per-device layer-count summary once loading completes, rather than logging 93 individual lines.

A per-device summary line (`CUDA0: 12 layers, CUDA1: 12 layers, ...`) is probably more useful than raw passthrough and keeps the log quiet.

## Notes

Diagnostics-only; no runtime behaviour change. Found while analysing #1298 — see @Dario's source analysis in that thread. Not a blocker for #1298 itself, but it is the reason #1298 is harder to diagnose than it should be.

Bortlesboat:
## Original problem

Native log summaries omit GPU model-buffer sizes and discard device names while counting layer-assignment progress. This leaves a multi-GPU load's placement unavailable in the summarized diagnostics.

Fixes #1304.

## Diagnostics

Regression inputs reproduce both losses through `NativeLogAggregator::process_line`. The parser also covers the `, is_swa = ...` suffix used by the pinned llama.cpp source.

## Fix

Retain model-buffer lines for all devices and emit a compact, sorted count after an assignment block. Existing percentage progress remains. Repeated layer rows do not inflate the count, changed assignments replace their old device, and each new model resets the state.

For example, the summary message `Model layers by device` carries `CPU=2, CUDA0=2, CUDA1=1`; a line such as `load_tensors: CUDA0 model buffer size = 4321.00 MiB` is retained in the memory category.

The verbosity environment-variable cleanup discussed later in the issue is outside this change.

## Validation

- [x] Ran relevant local checks; limits are recorded below.
- Three initial regressions fail against the original production implementation.
- 116 `skippy-runtime` library tests pass on Windows, including the new aggregation tests. Three previously confirmed failures requiring a loaded native runtime are excluded: `cache_append_extends_the_committed_history`, `cache_drafts_from_committed_history_and_never_mutates_for_a_prefix`, and `invalid_selected_backend_device_fails_before_model_open`.
- Required checks for `skippy-runtime` (with `dynamic-native-runtime`) and `mesh-llm`, and `cargo fmt --all --check`, pass.
- Strict Clippy was run for both targets. It stops in unchanged Windows code: the unread lock-file field in `materialized_cache.rs:94`, and needless returns in `audit.rs:469` and `mesh-llm-plugin/src/io.rs:205`. All three files match the base.
- No GPU inference, packaged-runtime or multi-node smoke test was run.

On a multi-device load, check the memory-category model-buffer lines and the device-count summary against the raw native log. The summary is emitted on the next non-assignment line; a stream truncated before that point will not contain its final count.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

## Summary by CodeRabbit

* **New Features**
  * Native logs now summarize how model layers are distributed across devices.
  * Device-specific layer counts are reported when layer assignment completes.
  * Layer assignment data resets correctly for each model load.

* **Bug Fixes**
  * Improved recognition of model buffer size log entries.
  * Device assignment summaries now avoid repeating unchanged information.
  * Device-specific buffer information is preserved during log aggregation.

### DIFF
--- a/crates/skippy-runtime/src/logging.rs
+++ b/crates/skippy-runtime/src/logging.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 use std::ffi::{CStr, c_char, c_int, c_void};
 use std::fs::{File, OpenOptions};
 use std::io::{LineWriter, Write};
@@ -185,6 +185,8 @@ struct NativeLogAggregator {
     metadata_progress: ProgressTracker,
     tensor_progress: ProgressTracker,
     layer_assign_progress: ProgressTracker,
+    layer_devices: BTreeMap<usize, String>,
+    layer_devices_emitted: bool,
     kv_cache_progress: ProgressTracker,
     metadata_in_dump: bool,
     metadata_summary_emitted: bool,
@@ -239,6 +241,8 @@ impl NativeLogAggregator {
         self.metadata_progress.reset();
         self.tensor_progress.reset();
         self.layer_assign_progress.reset();
+        self.layer_devices.clear();
+        self.layer_devices_emitted = false;
         self.kv_cache_progress.reset();
         self.metadata_in_dump = false;
         self.metadata_summary_emitted = false;
@@ -257,11 +261,15 @@ impl NativeLogAggregator {
         let mut events = Vec::new();
         let metadata_kv = parse_metadata_kv_line(s);
         let tensor_summary = parse_tensor_type_summary(s);
+        let layer_assignment = parse_layer_assignment(s);
         if metadata_kv.is_none() {
             events.extend(self.flush_metadata_summary());
         }
         if tensor_summary.is_none() {
             events.extend(self.flush_tensor_group_summary());
+        }
+        if layer_assignment.is_none() {
+            events.extend(self.flush_layer_device_summary());
         }
 
         if let Some((metadata_rows, tensor_rows)) = parse_loaded_metadata_counts(s) {
@@ -321,24 +329,8 @@ impl NativeLogAggregator {
             return events;
         }
 
-        if let Some(layer_index) = parse_layer_assign_index(s) {
-            if self.layer_assign_progress.total.is_none()
-                && let Some(total) = self
-                    .metadata_highlights
-                    .block_count
-                    .as_deref()
-                    .and_then(|s| s.parse::<usize>().ok())
-            {
-                self.layer_assign_progress.set_total(total);
-            }
-            let new_completed = layer_index + 1;
-            if new_completed > self.layer_assign_progress.completed {
-                let delta = new_completed - self.layer_assign_progress.completed;
-                events.extend(
-                    self.layer_assign_progress
-                        .advance(delta, "model", "layers", "layers"),
-                );
-            }
+        if let Some((layer_index, device)) = layer_assignment {
+            events.extend(self.record_layer_assignment(layer_index, device));
             return events;
         }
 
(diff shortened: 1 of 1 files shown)


michaelneale:
Reviewed on Mic's behalf. Approve — the logic is sound and the test coverage is unusually thorough for a log-parsing change.

I specifically checked for a conflict with the existing `assigned to device` parser, since that line shape had a parser at `logging.rs:572-583` on main. There's no duplication: that *is* `parse_layer_assign_index`, and this PR replaces it in place with `parse_layer_assignment`. `git grep "assigned to device" -- crates/` returns hits only in `crates/skippy-runtime/src/logging.rs`, so no second parser of that shape exists in-tree.

Things I verified rather than assumed:
- **No panic path.** Every step is `?`-guarded — `strip_prefix`, `split_once`, `parse().ok()?`, and empty device rejected via `(!device.is_empty()).then_some(...)`. Malformed input is covered at `logging.rs:1367-1370`.
- **No off-by-one in the progress accounting.** `record_layer_assignment` (`:348-365`) uses `saturating_add(1)` / `saturating_sub(completed)`, so a repeated assignment yields `delta == 0`, and `ProgressTracker::advance` (`:66-93`) emits nothing for `advance(0)`. Behaviourally equivalent to the old `if new_completed > completed` guard.
- **Reset is complete** — `layer_devices.clear()` plus `layer_devices_emitted = false` in `reset_model_loading_state` (`:244-245`), with a cross-model test at `:1159`.
- **The `"_Mapped model buffer size"` → `"model buffer size"` widening doesn't get shadowed** by the earlier branches: the backend branch needs `init|device|backend` in the lowercased line and the model branch needs the `llm_load_tensors` prefix, so `load_tensors: CUDA0 model buffer size = ...` matches neither. `aggregator_preserves_model_buffers_for_all_devices` asserts this across CUDA0/CUDA1/Metal/ROCm0/Vulkan0/CPU_Mapped.

Two nits, neither blocking:

1. `process_line` (`:271-273`) flushes the summary on any non-assignment line, so interleaved output mid-run produces a partial "Model layers by device" event and then another later. That matches `aggregator_emits_only_changed_layer_device_counts` so I take it as intended, but no test covers interleaved-then-resumed assignments and it's worth a comment saying the partial emit is deliberate.
2. `device.split(',').next()` keeps everything up to the first comma, so a form like `assigned to device CUDA0 (RTX 4090)` would become the param key `"CUDA0 (RTX 4090)"`. Speculative — every llama.cpp form I see is `device X, is_swa = ...` or bare — but these device names become arbitrary JSON keys in `NativeLogEvent.params`, so a whitespace split would be cheap insurance against an upstream format change.

Source reading only, no build.

ndizazzo (approved):

Issue #1204, opened 2026-08-08
Using openai-endpoint plugin without a runtime


coryalder:
## Feature

I've installed the openai-endpoint plugin (https://github.com/Mesh-LLM/openai-endpoint) via `mesh-llm plugins install openai-endpoint`. I've configured it in `~/.mesh-llm/config.toml`:

```
[[plugin]]
    name = "openai-endpoint"
    url = "http://127.0.0.1:8080/v1"
```

but since I don't have a working runtime for my machine's GPU (#1203) it gives the following error when I run `mesh-llm serve`:

```
root@ollama:~# mesh-llm serve
no compatible native runtime found for Skippy ABI 0.1.35 on linux/x86_64: no compatible MeshLLM native runtime is installed or installable for MeshLLM 0.75.0 / Skippy ABI 0.1.35; run `mesh-llm runtime install` or inspect available runtimes with `mesh-llm runtime list --available`
```

It seems like there should be a way to tell mesh-llm to ignore the lack of a runtime and depend on the openai-endpoint to provide a model. Not sure if this is a documentation issue, or a missing feature.

i386:
PR #1192 addresses the mesh-wide portion of this problem: plugin-only nodes can become discoverable and routable by peers. However, it does not address the startup failure described here: `mesh-llm serve` still initializes the native runtime before reading configured `[[plugin]]` entries. PR #1191 is unrelated. I recommend keeping this issue open until a follow-up change or regression test confirms that `mesh-llm serve` starts successfully with only `openai-endpoint` and no compatible native runtime. #1203 remains the underlying GPU-runtime issue.

PR #1692, opened 2026-09-08
feat: share an existing LLM server without a native runtime

michaelneale:
## Share an LLM server you already run

Publish one already-running OpenAI-compatible HTTP server to a mesh without installing an endpoint plugin, loading a native inference runtime or local model, or rewriting your config:

```bash
mesh-llm share http://localhost:11434

# Join an existing mesh, or explicitly publish for discovery:
mesh-llm share http://localhost:11434 --join <token>
mesh-llm share http://localhost:11434 --publish
```

With a default configuration, sharing starts a private mesh. The upstream keeps running its models; Mesh advertises their names and forwards requests, including streaming, from other mesh participants. Ctrl-C stops the sharing process, **not** the upstream.

Representative readiness output after a successful model probe:

```text
Sharing daemon ready; serving an external inference endpoint, local model loading is disabled
```

Inspect models with `curl -s http://localhost:9337/v1/models`.

**This replaces the earlier auto-discovery draft.** The current PR head is `306f24043eaf38ea4f622214909cc735b43ef431`, refreshed onto main `d4ffbbacd9c8486e0b80c44452876ffa17785964`. It retains the narrow implementation previously validated at `f562182e2` and the September 10 integration `8b6fce6b5`; current verification status is below. The earlier `plugins discover --apply`, setup probing, config-writing helper, and standalone SDK discovery feature have been removed from this PR's final diff. Historical commits/review comments remain for traceability; their descriptions and old CI passes are not evidence for this revision.

## Behavior and limits

- One operator-selected HTTP upstream per node, fixed for the run. No scanning, live attach/detach, or multi-provider manager.
- HTTPS, embedded URL credentials, query/fragment URLs and this node's own proxy/console are rejected. Authenticated upstreams are not supported; sharing strips caller Authorization, Proxy-Authorization, X-Api-Key and Api-Key headers.
- Startup requires a reachable, valid, nonempty OpenAI model inventory. HTML/malformed responses and a valid empty inventory produce distinct actionable errors instead of readiness.
- Refresh tolerates one failed probe with last-known-good models, withdraws them after sustained failure, and restores them on recovery. A valid empty inventory on refresh is a successful probe that withdraws models.
- The existing configured `openai-endpoint` plugin remains the durable-config option. When that plugin is healthy, its model inventory now participates in readiness, so a plugin-only node can move from Loading to Serving without a local runtime. This PR does **not** make every ordinary `serve` startup native-runtime-optional; `share` supplies the no-local-inference workflow relevant to #1204.

## Architecture and compatibility

A process-free upstream source uses the existing endpoint route inventory, health state, supervisor refresh tick, OpenAI forwarding and mesh announcement path. There is no second proxy, health loop or fabricated plugin process. Local inference is guarded at startup and later load boundaries.

The Rust SDK exposes the same fixed startup mode through `MeshNode::builder().share_endpoint(url)`; no runtime endpoint switching API. Other SDK builders do not yet expose this mode; adding those bindings is outside this PR.

Sharing's strict probe disables redirects/environment proxies, bounds the response body to 1 MiB, and validates OpenAI `data[].id`. Configured-plugin probes retain their prior lenient parsing. **Intentional common health change:** preserving models during a tolerated failed tick also applies to configured plugin endpoints.

No new gossip fields, mesh/plugin wire protocol or Skippy ABI. Mixed-version interoperability has not been live-certified for this revision.

## Current refresh — 2026-09-11

**Ready for the human merge decision.** Independently checked after CI settled: `MERGEABLE` / `CLEAN` at `306f24043eaf38ea4f622214909cc735b43ef431`, **57 successful checks/statuses, 19 skipped, no failures or pending checks**. Not merged. This normal merge has parents `8b6fce6b5900131f9ab48ebb5b5272fb0e47c414` and main `d4ffbbacd9c8486e0b80c44452876ffa17785964`, tree `6254efd68441e0196586b200d1683bf081be2346`. No force-push or merge into main.

### Integration review

Two conflicts in `crates/mesh-llm-host-runtime/src/mesh/node.rs` and `mesh/node/startup.rs` compose main's config-first, effective-safety-margin accounting with sharing's zero-capacity/no-accelerator-discovery guard. A production-wrapper regression uses a 4 GB configured margin and asserts absent accelerator inventory and zero itemized capacity. Config loading was extracted unchanged into `load_config_state_for_start` to keep `Node::start` under the warning-denying line-count limit.

Thinker independently reviewed the final resolution, confirmed main's `mesh/capacity.rs` unchanged and the auto-merged ready-reporter call matching its new signature, and verified merge parents/tree, clean source state and format/diff checks. The final evidence review rechecked the resolution, clean HEAD/tree and `git diff --check`. This is static review and saved-evidence review, not independent execution of the package suite or another live acceptance run.

### Local verification — saved logs reviewed

- Full `cargo test -p mesh-llm-host-runtime`: **3,094 unit tests passed / 11 ignored; 2 integration tests passed; a zero-test doc-test target completed**.
- Workspace Clippy, all targets with the four CI exclusions and `-- -D warnings`: clean final log. The pre-extraction `too_many_lines` failure is retained separately.
- Format, diff, console-print, CLI inventory, crate-list, publishing and release consistency checks passed per the recorded invocations. Format/diff were terminal-only; the other gate logs are archived.
- Separate Skippy runtime/server **library** runs passed (124 / 694 tests respectively). These are not full package suites.
- The test logs retain macOS minimum-deployment-target linker warnings. Clippy's clean pass is a separate result.

`WORK_LOGS/pr1692-refresh/refresh-20260911/INVOCATIONS.md` now distinguishes the pre-extraction source, deliberately dirty negative control and final validated tree. No contemporaneous SHA/tree was captured for the first two, and the logs themselves do not record HEAD. Association of the final local logs with the pushed tree remains builder-attested; it is not retroactively established by the current clean checkout.

**The full local workspace suite was not rerun.** September 10's invocation aborted at `skippy-model-package`, with the same failure reproduced on historical main `48887480f`; that remains historical baseline evidence, not a current full-workspace pass. Hosted Rust batches provide the current selected CI coverage.

### Fresh CI — terminal

| Lane | Current result |
| --- | --- |
| [Quality](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34550211655) | Attempt 2 successful; Clippy batches, format, CLI inventory and CI consistency/contracts green. |
| [Linux](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34550211874) | Attempt 2 successful; Rust batches 0–3 and selected build/smoke lanes green. [SafeTensors smoke](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34550211874/job/103115967298) passed in 15m49s. |
| [macOS](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34550211800) | Attempt 2 successful; selected unit/portable, Metal, arm64 host/product and model-load lanes green. |
| [Website](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34550211477) | Attempt 2 successful; public website build green. |
| [Windows](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34550211588) | Attempt 1: planner/stable gate passed; actual Windows lane **skipped**. Not a Windows build/test pass. |

First attempt is preserved as **40 successes including CodeRabbit, 3 failed checks, 11 cancelled, 19 skipped**. The root failure was the pinned Hugging Face fixture download returning HTTP 429 / maximum queue size reached **before** the SafeTensors smoke executed; the other two failures were Linux summaries. One bounded failed-job rerun of the four affected workflows cleared that failure and fail-fast cancellations, with no code/gate changes or retry loop. Cancelled checks are not additional test failures. The final tally includes CodeRabbit status, which is not a fresh independent human review.

### Current-head two-node share acceptance — stub upstream, PASS

MeshBuilder exercised two isolated local processes with loopback binding and iroh relays disabled: A shares a controllable Python HTTP upstream; B joins by invite and sends the inference requests. Thinker read the full harness and checked the saved observed values against the required outcomes rather than relying on the harness's `PASS` label.

| Observed through node B | Result |
| --- | --- |
| Peer inventory | `mesh` and `stub-review-model` visible. |
| Plain chat and SSE | HTTP 200, `stub-plain-ok`; four content chunks forming `stub-stream-ok`, `[DONE]` present. |
| Authorization stripping | Initial chat requests carried the test Bearer header; upstream's recorded header snapshot contains only `None`. |
| Sustained upstream 503 | Inventory withdrew to `[]`. |
| Recovery, no node restart | Inventory returned; **fresh plain chat and SSE both succeeded through B after recovery**, separately checked from inventory recovery. |
| Stop sharing | A exited 0 after SIGINT; upstream `/v1/models` still worked; B inventory drained to `[]`. |

**Artifact provenance:** [macOS run 34550211800, attempt 2](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34550211800), artifact `ci-product-macos-arm64-metal` / ID `10181318298`, workflow head exactly `306f24043`. Independent download matched GitHub's ZIP digest `cb361e252bfe1ada80c1cac67e0b6fc40c1fa48f109e16943849d1461e903247`, inner tarball SHA-256 `5b050077a85cc80326537007374081f0def891afcf916c494c6ae3d589bdc5ce`, original host SHA-256 `652431088ac9f5c44c794bb33e2f9cea557f99547e613bc7d4ef0b23a6e834a8` and saved product manifest (Mesh 0.76.0).

The actually exercised binary SHA-256 was `dc0372101e1f5811335a3bc0b7f53bccde353762d23da92e188370cd3d4166fe`, recorded by the harness. MeshBuilder applied an ad-hoc macOS signature before the run; that transformation is builder-attested, and those signed bytes were already cleaned before independent review. Thinker did not execute or sign the re-downloaded artifact.

**Limits:** stub upstream, not fresh real Ollama; historical real-Ollama acceptance below remains at `f562182e2`. No fresh mixed-version/public-mesh/agent-harness certification. The Authorization snapshot precedes the recovery requests, so it does not independently verify their upstream headers. The harness also does not directly assert retention at exactly one failed health tick. The Windows skips and incomplete local workspace coverage remain explicit; no broader claim is made.

### Evidence, cleanup and readiness decision

Evidence is retained outside the checkout in `WORK_LOGS/pr1692-refresh/refresh-20260911/` and `twonode-stub-20260911/`: exact invocation descriptions, local logs, harness/results, node logs, artifact hashes and independent final CI/artifact review. Source worktree is independently confirmed clean at the reviewed HEAD, with no `target/`. MeshBuilder reports `cargo clean` reclaimed 23.1 GiB and its downloads, extracted product, run directories and test processes were cleaned. The reviewer removed the independent temporary download and found no `/p/mesh-llm` or `TWONODE.py` process matches in its read-only snapshot. Worktree retained while the PR is open.

**Readiness decision:** the reviewed merge resolution, current selected CI and bounded current-head two-node workflow — now including post-recovery inference — are sufficient for this conflict refresh. No further broad run or feature expansion is requested. Upstream-health visibility and ready/loading status polish remain nonblocking follow-ups. The merge itself remains the human's decision.

## Previous main refresh — 2026-09-10 (historical)

**Historical readiness decision, superseded by the September 11 refresh above.** GitHub reports `MERGEABLE` / `CLEAN` at `8b6fce6b5900131f9ab48ebb5b5272fb0e47c414`; selected CI is green and the bounded current-head two-node `share` workflow has now passed against a controllable HTTP upstream. Thinker reviewed the integration, live-run harness/results and artifact provenance. The PR has not been merged. This is not a claim that the local workspace suite fully passed or that current-head real-Ollama/Windows acceptance was run.

### Integration and cleanup

Normal merge of the previous head with main `48887480ff0b4abaa2fd9faf12e5793272bb2409`, confirmed against the live remote main ref. Source tree: `6a87980f4f236f2ba9a157769ebae44d3fb4438e`. The only textual conflict was an obsolete test-fixture import block in `mesh/tests/owner_control.rs`; main's fixture consolidation is retained. Independent static review checked that resolution and preservation of credential stripping alongside main's route reservations. `git diff --check` passed at this HEAD. This integrates main's other changes too; “one conflict” does not mean the runtime is identical to the previous head.

Source worktree clean and `target/` absent, independently verified. MeshBuilder reports `cargo clean` reclaimed 40.7 GiB, the baseline probe worktree was removed and owned test processes stopped. Logs are archived outside the checkout in `WORK_LOGS/pr1692-refresh/`; source/evidence preserved.

### Local validation and its limits

MeshBuilder's reported Rust environment: `LLAMA_STAGE_BACKEND=metal`, `LLAMA_STAGE_BUILD_DIR=.deps/llama-build/build-stage-abi-static-metal`, prepared via the repository's llama scripts. Commands use the repository's `just with-lld` wrapper where shown in the logs.

- Format: `cargo fmt --all -- --check` — passed.
- Clippy: `cargo clippy --workspace --all-targets --exclude mesh-llm-ffi --exclude mesh-llm-nodejs --exclude skippy-ffi --exclude skippy-quantize -- -D warnings` — reported passed, zero warnings.
- Consistency: `just no-console-print`, `just cli-inventory-check`, `just ci-crate-lists`, `just publish-crates`, `just check-release` — saved logs show RC=0. A complete local `just ci-validate` pass is not established; hosted CI contracts passed separately.
- Supported build: `just build` — saved log ends with composed debug host plus adjacent Metal runtime and `BUILD_RC=0`. macOS minimum-deployment-target linker warnings remain; not a warning-free build.
- Separate library runs: `cargo test -p skippy-runtime --no-default-features --lib` and the corresponding `skippy-server` command — logs show RC=0. These are library runs, not complete package suites.

**The local workspace suite did not pass or execute every target.** The attempted command was `cargo test --workspace --exclude mesh-llm-ffi --exclude mesh-llm-nodejs --exclude skippy-ffi --exclude skippy-quantize --exclude skippy-runtime --exclude skippy-server`.

1. Initial `test.log`: 114 successful target results, including host-runtime unit tests **3,075 passed / 11 ignored**, then `skippy-model-package` aborts with **46 passed / 34 failed**, reporting an unloaded native runtime.
2. Standalone package `test2.log`, after native preparation: **80 passed / 0 failed**. This is not a workspace-suite pass.
3. `test_full.log`: the workspace invocation again aborts on the same **46 / 34** result and `WS_RC=101`; the log then runs the separate Skippy runtime/server library suites successfully. It contains no doc-test execution, so later workspace targets/doc-tests remain unverified locally.

MeshBuilder reports reproducing the same failure on unmodified main `48887480f`; the retained `mainprobe.log` contains that SHA and the same failure tally. The baseline command and full provenance are delegate-attested, not independently rerun. The standalone-versus-workspace pattern supports a feature-graph explanation, but green hosted CI is not proof that the local invocation is sound. The four dynamic-binding packages excluded above also have separate steps in `just test-all`; their exclusion list alone is not complete `test-all` coverage.

### Historical CI — terminal at `8b6fce6b5`

Independently checked: **57 successful checks/statuses, 19 skipped, no failures or pending checks.** All five stable PR workflow results succeeded; selected coverage is not every platform/backend suite.

| Lane | Executed coverage |
| --- | --- |
| [Quality](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34476332250) | Clippy batches 0–2, format, CLI inventory, CI contracts and consistency passed. |
| [Linux](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34476333019) | Rust test batches 0–3, SafeTensors smoke, static ABI, CPU runtime/host/product, core inference, two-node client, KV-cache and Rust SDK smokes passed. |
| [macOS](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34476332519) | Portable/unit checks, Metal runtime, arm64 host/product and Metal model-load smoke passed. |
| [Website](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34476332340) | Public website build passed; console UI quality/E2E skipped. |
| [Windows](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34476332364) | Planner and stable gate passed; the Windows build lane was **skipped**. **No Windows host/runtime build or test pass.** |

Other skipped coverage includes Linux CUDA/ROCm/Vulkan suites, CPU product integration, Qwen3.5 recurrent migration, macOS product integration and Swift/Kotlin SDK suites. CodeRabbit status is successful; that is not a substitute for a fresh independent human review.

### September 10 two-node share acceptance — stub upstream at `8b6fce6b5`

MeshBuilder exercised two isolated local Mesh processes over loopback with iroh relays disabled: A runs `share` against a controllable Python HTTP upstream; B is an ordinary client joined by invite. No rebuild, remote lab host, GPU or model download. Thinker read the full harness, checked the saved observed values against the expected outcomes, and independently matched the CI artifact hashes; this was evidence review, not a second live execution.

| Observed through node B | Result |
| --- | --- |
| Peer inventory | `mesh` and `stub-review-model` visible. |
| Plain chat | HTTP 200, `stub-plain-ok`. |
| SSE | HTTP 200, four content chunks forming `stub-stream-ok`, `[DONE]` present. |
| Authorization stripping | Caller sent a test Bearer header; upstream observed no Authorization on the chat requests. |
| Sustained upstream 503 | Peer inventory withdrew to `[]`. |
| Upstream recovery | Peer inventory returned without restarting either Mesh node. |
| Stop sharing | A exited 0 after SIGINT, upstream `/v1/models` still worked and B inventory drained to `[]`. |

**Artifact provenance:** [macOS run 34476332519](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34476332519), artifact `ci-product-macos-arm64-metal` / ID `10152265102`, workflow head `8b6fce6b5`. Independent re-download verified the inner `product-input.tar.gz` SHA-256 `ee13595c63c0f37e5aba41c75bfa6177f690b245ee095620581e388f47c1dd25` and the original host SHA-256 `31e8464affd8520ac91072bd28618080189f42885e40af6fc63b9ed944b114f6` against `product-manifest.json` (Mesh 0.76.0). MeshBuilder then applied an ad-hoc macOS signature for the local run. The actually tested binary hash was `fde2a8d10d36a34ff1d5e15f4df35fbed4e9e4b6c85615203fb9aafd5339c5d2`, recorded in the results. That signing step is builder-attested; the exercised bytes had been cleaned before independent review.

Evidence: `WORK_LOGS/pr1692-refresh/twonode-stub-20260910/` contains the harness, observed results, both node logs, artifact hashes, independent manifest/hash review and saved-value assertions. Builder reports test processes, extracted product, download and run directories cleaned; source checkout remains clean with no `target/`. The reviewer also removed their temporary artifact download without executing it.

**Limits:** this is stub-upstream evidence, not real Ollama. It establishes inventory withdrawal/recovery, but does not directly assert retention at exactly one failed refresh tick or send fresh chat/SSE after recovery. Real-Ollama acceptance below remains at the previous head. The harness records several outcomes without asserting them itself; the review checked the actual values rather than trusting its `PASS` label. Windows build skips and the aborted local workspace suite remain as documented above.

**Historical September 10 readiness decision:** this bounded feature-specific workflow, fresh selected CI, prior real-upstream acceptance and reviewed integration are sufficient for this merge refresh; no further broad test run or feature expansion is requested. Upstream-health visibility and ready/loading status polish remain nonblocking follow-ups. The actual merge remains the human's decision.

## Previous implementation validation (before the main refresh)

Local checks and acceptance were run on the exact source state subsequently committed here, before commit (base `494c9583b`, dirty fingerprint `feba9eebe75ac6f44731ab8639c352e7898c99917805aa26c7282970893cfb57`; the two previously untracked source files were hashed separately). They were not rerun merely to push an unchanged tree.

- Full `cargo test -p mesh-llm-host-runtime`: **3031 unit tests passed, 11 ignored**, plus 2 integration tests passed.
- Full `cargo test -p mesh-llm`: passed across recorded targets.
- `cargo fmt --all --check`, supported `just build`, and `cargo clippy -p <crate> --all-targets -- -D warnings` passed for host-runtime, mesh-llm, cli, sdk and config.
- Independent real **Ollama 0.32.14 / qwen2.5-coder:3b** acceptance: plain and SSE chat through a second Mesh process; upstream stop/restart withdraws/restores models on both nodes and streaming recovers without Mesh restart; stopping sharing leaves the same Ollama PID alive and direct inference working.
- Sharing ran from a binary-only directory with isolated config/cache roots: no plugin child, native backend mappings or model/cache artifacts observed; config unchanged.
- Final-binary HTML and empty-inventory startup regressions independently passed. Transient-failure model retention/withdrawal/recovery was also independently reproduced after correction.

**Scope:** two local Mesh processes over private loopback QUIC with a transparent SSH forward to real Ollama. Supported debug product, not cross-Mac LAN, release-build, public-mesh, mixed-version or agent-harness certification.

**Known local validation limits:** commands/embedded-runtime Clippy encounters three `dead_code` errors in mesh-llm-system Metal helpers; the commands failure was also reproduced on a clean `494c9583b` worktree. `just ci-validate` was blocked locally by missing PyYAML. The macOS build log contains a minimum-deployment-target linker warning. The local limits above are distinct from the successful hosted CI results below.

Evidence retained in the local lab archive `share-ollama-20260909/`: `acceptance-final/RESULTS.json`, `final-regression/RESULTS.json`, provenance and saved builder check/build logs. Final tested binary SHA-256: `5843e23d1b181a2654c3ba0c5efd89f6cfebac78e8a90fd68097abacc10ec311`. Owned test processes stopped; the build target was cleaned after validation (65.6 GiB reported removed).

## Previous CI — completed on `f562182e2`

All five native PR workflow results were successful on `f562182e2`. These checks predate the main refresh and are not CI evidence for `8b6fce6b5`. Selected CI passed, not every platform/backend suite:

| Lane | Evidence and executed coverage |
| --- | --- |
| [Quality](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34308589921) | Success: all three Clippy batches, format, CLI inventory, CI contracts and consistency. |
| [Linux](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34308590193) | Success: all four Rust test batches, static ABI, CPU runtime/host/product, two-node client, core inference, KV-cache and Rust SDK smokes. |
| [macOS](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34308590159) | Success: unit/portable checks, Metal runtime, host/product and Metal model-load smoke. |
| [Website](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34308589922) | Success: public website build. Console UI quality/E2E were skipped. |
| [Windows](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34308589946) | Planner and stable gate succeeded; Windows build lane **skipped**. **Not a Windows build/test pass.** |

Linux CUDA/ROCm/Vulkan suites, CPU product integration, the Qwen3.5 recurrent gate, macOS product integration and Swift/Kotlin SDK suites were skipped. CI's two-node smoke is additional evidence, not a replacement for the real-Ollama acceptance or certification of the live-workflow gaps above.

The earlier Quality failure was stale console-print line metadata. `f562182e2` corrects eight generated line numbers without changing runtime source or approving new print sites. Following those local xtask checks, MeshBuilder reports another `cargo clean` removed 121.1 MiB; the clean worktree and absent `target/` were independently verified at this head. Source and evidence are preserved.

## Existing review feedback

Rechecked the seven inline comments and two outside-diff findings against the replacement:

- Discovery help/config-path handling, raw discovery terminal output and the SDK discovery-test findings concern removed paths/features, not code to carry forward.
- Setup-summary console allowlist entries match the restored source line locations.
- Malformed model inventories, redirect mismatch and unbounded probe-body concerns are addressed by sharing's strict probe and regression tests.
- The old raw model-ID printing path is removed; sharing's startup trace uses debug-escaped model strings. This is not a claim of global model-ID sanitization elsewhere in Mesh.

CodeRabbit's existing status comment reports automatic reviews paused. Its previous review and the previous green CI run predate this replacement.

## Independent Windows review on the previous head

[Virgile-pct's Windows 11 review](https://github.com/Mesh-LLM/mesh-llm/pull/1692#pullrequestreview-5152663326) tested `f562182e2` with a controllable Python upstream: two-node plain/SSE chat, credential stripping, inventory withdrawal and recovery. This is independent stub-upstream evidence, not real Windows Ollama acceptance or a Windows CI build/test pass.

The review explicitly labels its five observations nonblocking. Upstream-health visibility and the ready sharing node displaying `node_state: loading` remain follow-up candidates, not reasons to add a health manager or broaden this integration refresh unless a functional blocker is established.

## Sharing documentation and feature highlight

The FAQ now answers “Can I share an existing Ollama or LM Studio server?” A dedicated “Share an existing model server” guide under Running Models walks through upstream setup, private sharing, joining from a second machine, and sending a chat. Homepage, README, quickstart and plugin docs link to the workflow. It explicitly distinguishes forwarding from splitting the upstream model, and documents HTTP/no-auth limits and Ctrl-C behavior.

Validation for the docs-only update: website `npm run build` passed (CLI inventory, Tailwind, Eleventy, Pagefind); desktop/mobile Chrome renders of homepage, FAQ, quickstart and guide passed page-overflow checks; guide local links and anchors resolved. Full CLI-explorer browser validation passed using installed Chrome (default Playwright browser binary unavailable). Rustdoc and runtime/lab suites were not rerun for this docs-only update.

### Homepage screenshots (initial docs wording, before lifecycle clarification)

Sharing highlight — desktop

Sharing highlight — mobile

## Server ownership clarification — 87747e246

`share` makes existing models available to users and machines on the mesh. It does not start, stop, restart, or manage the upstream server. Server setup is a prerequisite, not a share operation. Revised FAQ, guide, quickstart, homepage, README, usage and plugin docs say this explicitly. If an upstream is unavailable or has no models initially, share reports an error; later health checks withdraw/re-advertise availability without restarting it. Runtime code unchanged; inspected startup probe, process-free plugin resolution and shutdown paths. Website build and five-route desktop/mobile render checks passed.

Current ownership wording


Virgile-pct (commented):
Tested on Windows 11 at f562182e2, host built with `--no-default-features --features web-ui,dynamic-native-runtime` and `MESH_LLM_SKIP_UI` (so no console assets), with an empty native-runtime cache; share never resolves a runtime, so this is the intended no-GPU case. The upstream was a small Python OpenAI-compatible stub on `127.0.0.1:11434` whose inventory I can switch between two models, an empty list, an HTML page, a 500 and a hang; it serves non-streaming and SSE chat and logs the headers it receives. So this covers the mechanics, not a real Ollama or LM Studio.

What works as described:

- `share http://127.0.0.1:11434`: up in 11 s with `Sharing daemon ready; serving an external inference endpoint, local model loading is disabled`, no runtime involved. `/v1/models` lists both stub models, non-streaming chat and SSE chat are forwarded, an unknown model gets `model_not_found`. Upstream, the stub sees no `Authorization` or `X-Api-Key` from the caller and gets the caller's `User-Agent` as is.
- A second node on the same box, `client --join <token>` (with `MESH_LLM_EPHEMERAL_KEY` so the two processes do not share the key from #1699), lists the two models 12 s after starting; chat and streaming through the client are routed `provider=mesh` to the sharing node and on to the stub, and the caller's `Authorization` is gone at the stub end to end.
- The sharing node advertises nothing it does not have: `passive_mode` reports `role: endpoint`, `capacity_gb: null`, `models_on_disk: null`; `/api/status` on the console port shows `available_models: []` and `local_serving: false`. That is the concern I raised on #1696, handled here.
- Rejections, each before anything starts: `https://` (`sharing currently supports HTTP upstreams only; HTTPS is not supported yet`), credentials in the URL, a query or fragment, a `/v1/models` path (`provide the API base URL, not a models or chat route`), `--model`, `--client`, `--split`, each with its own message. This node's own API or console port: `FATAL cannot share this node's own proxy or console; choose the upstream server port`, exit 1, right after the API came up.
- Inventory problems at startup, all exit 1 with distinct texts: unreachable port (`cannot reach the upstream models API at http://127.0.0.1:11499/v1: GET http://127.0.0.1:11499/v1/models failed: error sending request for url (...)`), valid empty list (`the upstream at ... reports no models; load a model on it (for example ollama run <model>) and retry`), a 200 with HTML (`200 OK but the response is not an OpenAI-compatible model list: response body is not JSON`), a 500 (`-> 500 Internal Server Error`), an upstream that never answers (the 3 s timeout, then the same text as unreachable).
- Refresh, with 15 s ticks: after switching the upstream to 500 the models stay listed through one tick and are withdrawn 27 s after the switch; back to normal, they are restored 20 s later; a valid empty inventory withdraws them within 6 s, no tolerated tick; a changed inventory shows up at the next tick.

`cargo test -p mesh-llm-host-runtime --lib --features dynamic-native-runtime` filtered on the shared endpoint, sharing and native runtime gate tests: 45 passed; `cargo test -p mesh-llm-cli --lib` share normalization: 2 passed. `cargo clippy --no-deps -p mesh-llm-host-runtime -p mesh-llm-cli --all-targets -- -D warnings` is clean once the four pre-existing Windows lints in `plugin/` are allowed (`-A unused_imports -A clippy::needless_return -A dead_code`, same as on main); `cargo fmt --check`: clean.

Observations, none blocking:

1. The transport error text loses its cause. `health.rs` formats the probe failure as `format!("GET {} failed: {}", models_url, err)`, and `reqwest::Error`'s `Display` stops at `error sending request for url (...)`, so a refused connection and a 3 s timeout produce the same FATAL. Walking `source()` (or going through anyhow and printing with `{:#}`) would surface `connection refused` versus `operation timed out`, which is the one thing the operator needs to decide between a wrong port and a stuck server.
2. Nothing reports the upstream's health. `/api/runtime/endpoints` returns `{"endpoints":[]}` for a sharing node, by design since the shared endpoint is not a plugin, so the only visible sign of a degraded or withdrawn upstream is the model list shrinking. A small `shared_endpoint` block in `/api/status` (state, last probe, detail) would give operators something to look at.
3. `/api/status` for a ready sharing node says `node_state: loading`, `node_status: Loading`, `llama_ready: false` and `is_host: false` next to `daemon_state: ready_proxying` and `serving_models: [...]`. The console's chat gate falls back to `hosted_models`/`serving_models`, so chat should still work there, but anything keyed on `llama_ready` or `node_state` will read the node as still loading. I could not look at the console itself with this build.
4. The upstream probe runs after the API and the mesh are up, so with a bad URL the node answers `/v1/models` with an empty list for a second or two before the FATAL. The shutdown is clean, this is only about where the failure lands.
5. `mesh-llm share --port 9447` with the URL forgotten fails with clap's `unexpected argument '--port' found` rather than a missing-URL message. The normalization test pins that, so just noting the operator-facing text.

Not exercised here: a real provider or HTTPS, `--publish`, Ctrl-C behaviour (I killed the process), and the SDK builder.

michaelneale:
Thanks @Virgile-pct - you think this is a worthwhile enhancement? That battery of tests was good. 

Aim is if people use ollama etc - can just share it as a solo model alongside others.

Virgile-pct:
Yes, I think it is worth having, and more so on Windows than elsewhere. A Windows host built from main currently has no published native runtime it can install (the gap noted on #1511 and #1615), so `share` is the only way such a machine can put a model on a mesh at all. And Ollama or LM Studio is what most Windows people already run, so sharing what is already loaded is a much smaller step than installing a runtime and a GGUF. The part I liked most is that the sharing node advertises no capacity and no models on disk, so the scheduler is not planning around something it cannot use.

If it helps to order the follow-ups, the two observations that matter for that audience are the upstream health surface (2) and the `node_state: loading` reading on `/api/status` (3), because Ollama users will look at the console before anything else. Happy to re-run the battery against a real Ollama once the branch is rebased.

michaelneale:
@Virgile-pct since 0.76 is there a ntive windows runtime?

Virgile-pct:
Yes, and that correction is on me: what I wrote above was true of a host built from main at the time, not of the release. I checked v0.76.0 yesterday on the same Windows 11 / RTX 4070 Ti box, with the published bundle and an empty runtime cache. `runtime list --available` shows all four Windows artifacts (cpu, cuda12, rocm, vulkan) as supported, and `runtime install cuda` resolves cuda12 from the release catalog and installs it in 36.6 s, with no `--bundle-dir`. Details are in #1511, which James closed on the back of it.

So `share` is not the only way in on Windows any more. What I think it still buys, now that the runtime works:

- People who already run Ollama or LM Studio have the model downloaded and warm. Sharing that is a smaller step than a 824 MB runtime plus a GGUF, even when both work.
- A host built from main can still be ahead of the published catalog's ABI, which is the case I hit. `share` covers that gap without waiting for a release.

Happy to run the battery against a real Ollama on Windows whenever you want it.

Virgile-pct:
Ran the battery against a real Ollama this time, as offered. Windows 11, RTX 4070 Ti, head `87747e246`, host built with `--features web-ui,dynamic-native-runtime` and `MESH_LLM_SKIP_UI`, so no console assets and no native runtime involved. Upstream is Ollama 0.34.0 on `127.0.0.1:11434` serving `llama3.2:1b` and `qwen2.5:1.5b` from the 4070 Ti.

The quickstart works end to end. `share http://127.0.0.1:11434` is up in 9.7 s, `/v1/models` lists both model IDs unchanged, colons included, next to `mesh`.

Cost of the relay, measured against the same upstream at the same moment, alternating requests and generating 300 tokens at `temperature: 0` so the numbers are generation and not latency:

| | Through the mesh | Straight to Ollama |
|---|---|---|
| Throughput, mean of 3 | 72.7 tok/s | 70.8 tok/s |
| Time to first token, median of 3 (SSE) | 10.9 ms | 8.6 ms |

So the relay costs about 2 ms before the first token and nothing measurable on throughput; the two sit inside each other's noise. Streaming arrives as `text/event-stream` with 301 chunks and the deltas intact. A first call on a model Ollama has not loaded takes 29.1 s for llama3.2 and 23.4 s for qwen2.5, which is Ollama loading it rather than anything `share` does, but it is what a first-time user will see.

The part that matters most for the Ollama story in the new docs page holds up. With `share` running, `ollama pull smollm2:135m` is picked up 2.1 s later and is immediately usable through the mesh; deleting it withdraws it 6.2 s later. Killing the Ollama process mid-session returns 503 straight away, withdraws both models after 22.6 s, and restores them 9.1 s after Ollama is back, with inference working again right after. The sharing node stays up and answers throughout. It also advertises nothing it does not have: `my_vram_gb: 0.0`, no GPU entries, `available_models: []`, while `serving_models` carries the two upstream IDs. A second node joined with `client --join` on the same box sees both models 9.7 s after starting and runs inference through the mesh in 1.67 s.

Two new observations, neither blocking:

1. **A malformed request body is reported as a routing failure.** Send invalid JSON and the node answers `503` with `single target None unavailable (route_to_target)`. Ollama answers `400` with the parse error. That 503 sent me chasing a routing bug on my own node for a while before I found the real cause in my own test harness, so a `400` carrying the parse detail would save someone the same detour.
2. **A request with no `model` field is answered instead of rejected.** Ollama returns `400 model is required`; the node picks a model on its own (`qwen2.5:1.5b`, the same one on three consecutive calls) and returns `200`. An empty `"model": ""` is correctly rejected with `404 model_not_found`, so this is specifically the missing key. I have not checked whether that default is `share`-specific or the general behaviour of the OpenAI surface, so treat it as a question rather than a finding.

Still true from the stub run, now confirmed against the real thing: `/api/status` reports `node_state: loading`, `node_status: Loading` and `llama_ready: false` while the node is serving both models, and there is no `shared_endpoint` block, so the only visible sign of a sick upstream is the model list shrinking.

Not covered here: the console itself (no Node toolchain on this machine, so I built with `MESH_LLM_SKIP_UI`), LM Studio, HTTPS upstreams, and `--publish`.

i386:
Review from Buzz (independent pass over the current diff at `b073944f`):

**Overall:** This is a well-scoped integration. Reusing the existing endpoint health/route inventory rather than adding a second proxy or health loop is the right call, and the strict/lenient probe split keeps configured-plugin behavior unchanged.

**Strengths**
- URL normalization rejects HTTPS, embedded credentials, query/fragment, and API-route URLs without echoing the input back (tests cover the leak case).
- Strict probe: no redirects, no env proxy, 1 MiB bounded body, `data[].id` validation — a 200 that isn't an OpenAI list is a failed probe, not an empty inventory. Good.
- `reject_local_proxy` resolves DNS and checks loopback/unspecified/own-bind/wildcard ownership before probing, preventing self-forwarding loops; covered by tests.
- Credential stripping (`Authorization`, `Proxy-Authorization`, `X-Api-Key`, `Api-Key`) is route-scoped via `strip_caller_credentials` so configured plugin endpoints keep their auth headers — with a byte-preservation test including non-UTF8 header values.
- Health-state change notification explicitly excludes probe timestamps from the equivalence check so idle healthy nodes don't wake the status stream every 15s — thoughtful.

**Questions / minor**
1. `start_shared_endpoint` requires a non-empty inventory at startup but `refresh_shared_endpoint` treats a valid empty inventory as a successful withdrawal. The asymmetry is documented and intentional, but a one-line comment in the refresh path pointing back to the startup rationale would help future readers.
2. `shared_endpoint_status_equivalent` compares `record` (which contains `models`) and `consecutive_failures` but not `last_probe_at` — correct, but worth asserting in a test that a healthy→healthy tick with identical record doesn't notify (the comment explains why; a regression test would lock it in if not already present).
3. `SHARED_ENDPOINT_ID = "openai"` as a fixed route id: confirm this can't collide with a configured plugin endpoint id on a node that both shares and runs the `openai-endpoint` plugin (key is name+id, and name differs — looks safe, but worth an explicit test).
4. The self-proxy check binds an ephemeral UDP socket per resolved address to test local ownership — on hosts where unprivileged binds are restricted this could false-positive; presumably acceptable for the supported platforms.

Non-blocking from this pass; happy to dig further into the runtime/options and startup_models changes if wanted.

crates/mesh-llm-host-runtime/src/plugin/health.rs
```diff
@@ -137,6 +185,158 @@ impl PluginManager {
...
+    async fn shared_endpoint_inference_route(&self) -> Option<InferenceEndpointRoute> {
+        let address = self.inner.shared_endpoint.clone()?;
+        let record = self
+            .inner
+            .endpoint_health
+            .lock()
+            .await
+            .get(&endpoint_key(SHARED_ENDPOINT_NAME, SHARED_ENDPOINT_ID))
+            .map(|state| state.record.clone())?;
+        record.available.then(|| InferenceEndpointRoute {
```
i386, line 208: Blocking correctness issue: this makes the shared upstream a local endpoint route, but the OpenAI ingress only consults endpoint/plugin dispatch after ordinary local and remote mesh candidates. Because this node gossips these model IDs as hosted, two sharing nodes advertising the same common model each see the other as a remote host. A tunneled request landing on A can therefore be forwarded to B, and B can forward it back to A; the inbound source peer is not carried into route selection, so there is no exclusion that breaks the cycle. Please make this node's available shared endpoint win before remote mesh candidates (including the existing `ModelTargets` path), or add an equivalent hop/source guard, and cover two sharing nodes exposing the same model with an end-to-end regression test.

crates/mesh-llm-host-runtime/src/plugin/health.rs
```diff
@@ -383,8 +690,32 @@ async fn probe_openai_compatible_http_endpoint(address: &str) -> EndpointHealthR
...
+            .context("response has no OpenAI `data` array")?;
+        anyhow::ensure!(
+            entries.iter().all(|entry| entry
+                .get("id")
+                .and_then(|id| id.as_str())
+                .is_some_and(|id| !id.trim().is_empty())),
+            "response `data` contains an entry without a model id"
+        );
+    }
     let models = body
```
i386, line 719: Please bound and sanitize the strict-policy model inventory itself, not only the response bytes. A 1 MiB JSON body can still contain tens of thousands of short IDs; every 15-second refresh then clones, sorts, publishes, gossips, and renders that list. Escaped control characters (including ANSI/OSC sequences) are also accepted here and flow into status and operator-facing output. The earlier endpoint-discovery implementation had this same terminal-injection finding; removing that draft did not remove the trust boundary. For `ProbePolicy::Strict`, cap the entry count and per-ID byte length and reject control characters before storing the record, with regression tests for a large short-ID array and an escaped control sequence.

crates/mesh-llm-host-runtime/src/network/openai/response/external_endpoint.rs
```diff
@@ -233,48 +253,60 @@ fn rewrite_http_request_target(
...
-
-    let mut bytes = rebuilt.into_bytes();
-    bytes.extend_from_slice(&raw[header_end..]);
-    Ok(bytes)
+    use crate::network::openai::forwarded_request::finalize_forwarded_request;
+    let omitted = if strip_caller_credentials {
+        &[
+            "host",
+            "authorization",
+            "proxy-authorization",
```
ndizazzo: This list and `CALLER_CREDENTIAL_HEADERS` over in `forwarded_request.rs` have to stay in sync, and they've already drifted: `cookie` is missing here.

I sent `Cookie: session=COOKIESECRET` through a sharing node and it reached the upstream untouched, while Authorization and X-Api-Key were correctly stripped. Cookies are caller credentials, so this is the same problem the other headers are here to solve.

Let's hoist `CALLER_CREDENTIAL_HEADERS` to a module-level const, add `cookie` to it, and build `omitted` from that plus `host`. One list, and the share path picks up future additions for free.

crates/mesh-llm-host-runtime/src/network/openai/request_parse.rs
```diff
@@ -312,6 +312,19 @@ where
...
     let tokenize_request = is_tokenize_request(&parsed.method, &parsed.path);
+    // A JSON inference route whose body is not JSON is a client error, and the
+    // only place that can say so accurately is here, where the parse actually
+    // fails. Downstream every parse result is an `Option`, so an unparseable
+    // body reaches routing indistinguishable from one that simply named no
+    // model — and then fails as a 503 routing outcome. Restricted to the
+    // closed set of JSON inference routes so binary (`/api/objects`) and
+    // multipart paths keep forwarding bytes untouched, and to non-empty
+    // bodies so a valid body that merely omits `model` still auto-routes.
+    if is_json_inference_request(&parsed.method, &parsed.path) && !body.is_empty() {
```
ndizazzo: This is where the gzip problem lives. The check never looks at `Content-Encoding`, so a compressed body the upstream would happily accept gets a 400 here instead.

Confirmed against a gzip-aware upstream: 200 going direct, and through the proxy `400 request body is not valid JSON: expected value at line 1 column 1`. The same applies to any other content type a downstream server happens to support on these four routes.

Rejecting unparseable bodies is the right call, we just can't call compressed bytes unparseable. Skipping the check when `Content-Encoding` is present should be enough.

crates/mesh-llm-host-runtime/src/network/openai/request_parse.rs
```diff
@@ -312,6 +312,19 @@ where
...
+        serde_json::from_slice::<serde_json::Value>(&body)
```
ndizazzo: We parse the whole body into an untyped `Value` here, throw it away, then parse it again into `RequestMetadata` on the very next statement. That's the most allocation-heavy parse shape we have, on the hottest path in the proxy, and it'll show up on long chat histories or base64 image payloads.

`serde_json::from_slice::<serde_json::value::IgnoredAny>(&body)` validates the syntax with no allocation and produces the same error. Parsing once to `Value` and deriving the metadata from it would work too.

crates/mesh-llm-cli/src/parser/normalization.rs
```diff
@@ -122,6 +123,21 @@ where
             }
             _ => {}
         },
+        // `share <url>` becomes `--shared-endpoint <url>` so runtime flags
+        // parse on either side of the URL. When no URL follows, the `share`
+        // subcommand is left in place so clap reports the missing argument
+        // and renders help.
+        Some("share") => {
```
ndizazzo, line 130: The rewrite only fires when the token right after `share` is a non-flag, so `mesh-llm share --port 9447 http://localhost:11434` falls through to the real subcommand and dies with `error: unexpected argument '--port' found`.

What makes it worse is the usage line clap prints directly underneath that error:

```
Usage: mesh-llm share [OPTIONS] <URL>
```

So we advertise the exact form we reject. The field doc on `commands.rs` says flags parse "on either side of the URL", which isn't true for this side.

Maybe scan forward for the first non-flag token instead of only checking `pos + 1`.

crates/mesh-llm-host-runtime/src/plugin/health.rs
```diff
@@ -137,6 +185,158 @@ impl PluginManager {
...
+        self.notify_status_changed().await;
+        Ok(models)
+    }
+
+    /// Re-probe the shared upstream on the supervisor tick.
+    ///
+    /// Reuses `apply_endpoint_probe`, so an upstream that goes away is
+    /// tolerated for one tick as `degraded` and only then withdrawn, and a
+    /// recovered upstream is re-advertised — identical to a plugin endpoint.
+    pub(super) async fn refresh_shared_endpoint(&self) {
```
ndizazzo, line 276: There's a startup race here. `start_supervisor` is spawned at the end of `PluginManager::start`, and tokio's `interval` fires its first tick immediately, so this runs concurrently with `start_shared_endpoint`.

Both read `previous`, drop the lock, probe over the network, then write. If this tick's probe fails and its write lands second, it clobbers the good startup record while holding `previous: None`, which gives `starting`, `available: false`, and no models... right after the node printed ready. It heals on the next tick so it isn't severe, but it also means we hit the user's server with two probes on every startup.

Seeding the record before spawning the supervisor fixes both.

crates/mesh-llm-host-runtime/src/runtime/run_auto.rs
```diff
@@ -1624,6 +1712,16 @@ pub(super) async fn run_auto(ctx: RunAutoContext) -> Result<()> {
     })
     .await?;
 
+    // Register the shared upstream before reporting ready. A failure here uses
+    // the ordinary shutdown path so a failed probe leaves no participant behind.
+    let sharing_startup_result =
```
ndizazzo, line 1718: The comment says a failure leaves no participant behind, but it does leave one briefly. On my dead-upstream run the node created the mesh, printed an invite, and bound the API on 19338 before the FATAL landed.

Harmless for a private single-node mesh. With `--join <token>` we'd announce ourselves into someone else's mesh and then leave. Either probe before the node starts accepting, or soften the comment so the next person reading it isn't surprised.

crates/mesh-llm-host-runtime/src/mesh/node/startup.rs
```diff
@@ -103,18 +103,42 @@ pub(super) async fn wait_for_endpoint_online(
     }
 }
 
+/// A zero-capacity participant has no reason to initialize accelerator discovery.
+/// Keep hostname identity, but do not probe native runtimes or publish GPU inventory.
+pub(super) fn hardware_survey_for_start(
+    max_vram_gb: Option<f64>,
+    enumerate_host: bool,
+) -> crate::system::hardware::HardwareSurvey {
+    if max_vram_gb == Some(0.0) && !enumerate_host {
```
ndizazzo: Everywhere else in this PR we thread the intent explicitly, either `shared_endpoint` or `allows_local_inference()`. Here we infer it from a float sentinel, and that catches cases we didn't mean to catch: `mesh-llm client --no-enumerate-host` and `serve --max-vram 0` both now get a hostname-only survey where they used to get a full one.

I'd rather pass the intent down than read it back out of `max_vram`. This is the spot most likely to confuse whoever touches it next.

crates/mesh-llm-host-runtime/src/sdk/embedded_config.rs
```diff
@@ -8,6 +8,14 @@ pub const SIGNED_JOIN_TOKEN_MIN_PROTOCOL_VERSION: u32 = 1;
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum EmbeddedMeshNodeMode {
+    /// Full mesh participation serving one already-running OpenAI-compatible
+    /// HTTP server, with local inference disabled.
+    ///
+    /// The address is fixed for the node's lifetime — stop the node to stop
+    /// sharing. Mesh never starts or stops the upstream server.
+    SharedEndpoint {
```
ndizazzo, line 17: Two things about this variant.

`EmbeddedMeshNodeMode` is re-exported publicly from `mesh-llm-sdk` and isn't `#[non_exhaustive]`, so adding a variant breaks anyone matching on it exhaustively. Worth marking it `non_exhaustive` now, or calling the break out explicitly for the release.

The SDK story is also only half wired. `MeshNodeBuilder::share_endpoint()` works, but the other public builders in `mesh-llm-sdk` (`EmbeddedServeConfigBuilder` and `EmbeddedClientConfigBuilder`, which go through the private `EmbeddedMode` enum) have no share path at all. Let's either add one or say in the description why that surface is deliberately left out.

crates/mesh-llm-host-runtime/src/api/mod.rs
```diff
@@ -977,6 +1014,7 @@ impl MeshApi {
                 is_host: runtime_status.is_host,
                 is_client,
                 llama_ready: runtime_status.llama_ready,
+                external_inference_ready: plugin_ingress,
```
ndizazzo, line 1015: `plugin_ingress` is true for any available external endpoint, not just a shared one, so this flips `node_state` from `Loading` to `Serving` for existing nodes running a configured `openai-endpoint` plugin.

I think that's the correct fix and I'm not asking you to change it. But the description only lists model retention as an intentional shared-path change, and this belongs in that list too. Console consumers read `node_state`.

crates/mesh-llm-host-runtime/src/api/server.rs
```diff
@@ -322,17 +335,27 @@ async fn dispatch_management_request(
...
     match tokio::time::timeout(
         std::time::Duration::from_secs(5),
         proxy::read_http_request(stream),
     )
     .await
     {
-        Ok(Ok(request)) => Ok(Some(request)),
-        Ok(Err(e)) => Err(e),
+        Ok(Ok(request)) => Ok(Some(ManagementRequest::Parsed(Box::new(request)))),
+        Ok(Err(error)) => Ok(Some(ManagementRequest::Unparseable(error))),
```
ndizazzo, line 359: This folds every `read_http_request` failure into a 400, not just the JSON ones. An I/O error or a truncated read isn't a client error, and a body over the size limit should be a 413.

We also hand `error.to_string()` straight back to the caller, so internal read errors get echoed out. It's JSON-encoded so there's no response splitting, at least.

Answering instead of silently dropping the socket is the right instinct. Let's keep the classification though: match on the error and only 400 the parse failures.

ndizazzo (changes requested):
This is the right direction. Reusing the existing endpoint route inventory and health machinery instead of standing up a second proxy was the right call, and the guards around local inference are thorough.

I built `dbfb69d2f` and ran it against a controllable stub upstream rather than reading the evidence in the description. It works:

- startup, the readiness line, and `/v1/models` listing the upstream IDs
- plain chat and SSE both 200 through the proxy
- a model added upstream showed up within one tick
- sustained outage went healthy, degraded (one failure, models kept), unhealthy (withdrawn), then recovered, and chat plus SSE both worked again afterward
- `/api/status.shared_endpoint` reports what it claims, and `node_state` is `serving` with `llama_ready: false`
- dead upstream and an HTML response at `/v1/models` both exit 1 with distinct, actionable errors
- SIGINT stopped the node and the upstream kept serving

Two things I'd like fixed before this goes in. Both are small, and both affect paths beyond `share`.

Cookies get forwarded to the upstream. I sent Authorization, X-Api-Key, and Cookie through a sharing node. The first two were stripped and `Cookie: session=COOKIESECRET` arrived at the upstream intact. A cookie is a caller credential, so this contradicts what the route's own doc comment promises.

The new JSON gate rejects gzipped request bodies. With a gzip-aware upstream, the same bytes get 200 going direct and 400 "request body is not valid JSON" through the proxy. That one isn't share-specific, it hits every node on the four allowlisted routes.

I also want to flag the description. The verification narrative is pinned to `306f24043`, but the head is `dbfb69d2f`, three commits later, and the newest of those is runtime code (node-state derivation, the status surface, malformed-JSON 400s, the management-port 400). CI is green at the real head, I checked that separately, so the gap is the two-node acceptance run and the artifact provenance rather than CI. The gzip bug is also exactly the sort of thing that acceptance run wouldn't have caught. And at roughly 8,000 words the description buries its own decision-relevant facts. The evidence discipline is unusual and I like it, but the volume is working against you here.

Smaller stuff, none of it blocking:

- "cannot reach the upstream models API at X -> 200 OK but the response is not an OpenAI-compatible model list" reads wrong when we did reach it. The test asserts on that prefix, so it's locked in now.
- the no-echo test on URL normalization only greps for the literal string `secret`, so it would pass on any other leaked credential.
- `--shared-endpoint` isn't in `value_taking_flags`, so in some orderings its value gets mistaken for the positional subcommand.
- `reject_local_proxy` does a blocking `std::net::UdpSocket::bind` inside async code.

Rest of my notes are inline. The model-retention change is fine, by the way. I traced the router and peers are tried before the plugin path, so keeping stale models through a tolerated tick can't steal traffic from a peer that could serve the request.

Virgile-pct:
Re-ran the Windows battery at `dbfb69d2f`, against the same real Ollama 0.34.0 as before (Windows 11, RTX 4070 Ti, host built with `--features web-ui,dynamic-native-runtime` and `MESH_LLM_SKIP_UI`).

**The three follow-ups hold.** A sharing node now reports `node_state: serving` and `node_status: Serving` while `llama_ready` stays honestly `false`, with both upstream IDs in `serving_models`. `status.shared_endpoint` is there and populated: `state: healthy`, `available: true`, the probe detail, the model list, `consecutive_failures: 0` and both timestamps. Malformed JSON now answers `400 request body is not valid JSON: key must be a string at line 1 column 2`, and a valid request still returns 200 through the relay.

**One measurement on the two-sharing-node finding** (@i386's blocking comment on `health.rs:208`). I could reproduce it on this box, and the symptom is not the A→B→A loop, it is a total outage on both nodes:

- Node A shares Ollama, serves fine: 200.
- Node B shares **the same** Ollama and joins A's mesh. Each now lists the other as `role: Host` with `hosted_models: [llama3.2:1b, qwen2.5:1.5b]`.
- Inference on A: `503 all 1 target(s) for model 'llama3.2:1b' failed` after 12.7 s. Same on B, 12.4 s. Ollama itself answers 200 direct throughout.
- Stop B, wait for the peer to drop, and A serves again on the next call, three for three.

So the local shared endpoint never wins: `all 1 target(s)` is the remote peer, and the node does not fall back to the upstream it is itself sharing. Two people sharing the same Ollama in one mesh take each other down rather than adding redundancy. That matches the fix @i386 asked for (make the available shared endpoint win before remote mesh candidates).

**And one negative result on the gzip finding** (`request_parse.rs:323`). It does not reproduce with Ollama, because Ollama does not accept gzipped bodies either: direct to Ollama, a gzipped body gets `400 invalid character '\x1f' looking for beginning of value`; through the sharing node, `400 request body is not valid JSON: expected value at line 1 column 1`. Same verdict on both sides. The finding still stands for gzip-aware upstreams, it just does not affect the Ollama path the docs lead with.

Not covered: the console itself (no Node toolchain here), LM Studio, HTTPS, `--publish`, and the header-stripping list @ndizazzo raised, which needs an upstream that logs what it receives rather than Ollama.

Virgile-pct:
Follow-up on the inventory finding (@i386's `health.rs:719` comment), which I said I had not covered. I have now, same box and same head `dbfb69d2f`, against a stub serving a deliberately hostile `/v1/models`.

The stub returns 5,006 ids in a 305 KB body: one normal name, five carrying escape sequences, and 5,000 short `bulk-NNNNN` entries. Nothing is capped or rejected:

- The node starts and advertises **5,007** models (the 5,006 plus `mesh`).
- `shared_endpoint.models` holds all **5,006**.
- `/api/status` grows to **264 KB**, and that is what the console polls and what the collector rebuilds on every probe.
- The escape sequences survive intact: a clear-screen `ESC[2J`, an SGR colour `ESC[31mRED ESC[0m`, and an OSC title-setter `ESC]0;PWNED BEL` are all present in the advertised ids.

They are `\u`-escaped inside the JSON, so `/api/status` itself carries zero raw `0x1B` bytes. The exposure is at the consumer: anything that decodes the JSON and prints an id into a terminal executes them. That is the trust boundary the comment describes, and it is open on the strict path with a real body well under the 1 MiB limit.

No cap on entry count and no per-id length limit either, so the 15-second refresh clones, sorts and republishes a 5,000-entry list on every tick.

crates/mesh-llm-host-runtime/src/plugin/health.rs
i386: Fixed in 89789a7a: a healthy local shared endpoint now wins before remote mesh candidates for the same model, preventing self-forwarding loops. Added a regression covering a competing remote target.

crates/mesh-llm-host-runtime/src/plugin/health.rs
i386: Fixed in 89789a7a: strict inventories are capped at 4,096 models, each ID is capped at 512 bytes, and empty/control-character IDs are rejected. Boundary tests cover each limit.

crates/mesh-llm-host-runtime/src/network/openai/response/external_endpoint.rs
i386: Fixed in 89789a7a: the shared-upstream credential denylist now uses the same CALLER_CREDENTIAL_HEADERS constant as peer forwarding, including Cookie. Added a Cookie regression test.

crates/mesh-llm-host-runtime/src/network/openai/request_parse.rs
i386: Fixed in 89789a7a: Content-Encoding bodies bypass JSON inspection and are forwarded byte-for-byte. Added a gzip forwarding regression.

crates/mesh-llm-host-runtime/src/network/openai/request_parse.rs
i386: Fixed in 89789a7a: JSON syntax validation now deserializes to IgnoredAny, avoiding the throwaway Value allocation before metadata parsing.

crates/mesh-llm-cli/src/parser/normalization.rs
i386: Fixed in 89789a7a: share normalization now finds the URL after value-taking runtime flags, so `mesh-llm share --port 9447 URL` parses correctly. Added coverage for flags on both sides of the URL.

crates/mesh-llm-host-runtime/src/plugin/health.rs
i386: Fixed in 89789a7a: the supervisor explicitly consumes tokio interval's immediate first tick before entering its probe loop, so startup owns the initial shared-endpoint probe.

crates/mesh-llm-host-runtime/src/runtime/run_auto.rs
i386: Fixed in 89789a7a: the comment now states the observed lifecycle accurately—mesh/listeners can exist briefly, but readiness is never published when the startup probe fails.

crates/mesh-llm-host-runtime/src/mesh/node/startup.rs
i386: Fixed in 89789a7a: minimal hardware survey is now an explicit startup intent. Client/no-enumerate and ordinary serving with max-vram 0 retain their prior behavior; shared-endpoint mode requests the minimal survey directly.

crates/mesh-llm-host-runtime/src/sdk/embedded_config.rs
i386: Fixed in 89789a7a: EmbeddedMeshNodeMode is now non_exhaustive. The PR description also explicitly scopes shared-endpoint support to the Rust SDK builder; Node, Swift, and Kotlin do not gain this mode in this PR.

crates/mesh-llm-host-runtime/src/api/mod.rs
i386: Documented in the updated PR description: a configured openai-endpoint plugin now contributes to readiness and can move a plugin-only node from Loading to Serving. That is the intended behavior.

crates/mesh-llm-host-runtime/src/api/server.rs
i386: Fixed in 89789a7a: request-read errors preserve their typed context. Oversized bodies return 413, malformed requests return 400, and truncation/I/O failures return 500, all with sanitized public messages. Added status/message regressions.

PR #1691, opened 2026-09-08
Catch CI coverage gaps before changes reach main

michaelneale:
## Summary
This is a general PR/main CI coverage repair, not a release-only change. PRs should select fewer tests for efficiency, but must select the executable consumers affected by a change.

Catch known PR-green/main-red regressions without making every PR exhaustive:

- Route script-only changes to their real UI, product, SDK and CUDA consumers, with six single-path planner fixtures.
- Exercise the SDK native-runtime JSON reader against the actual composed host with fallback builds disabled.
- Rehearse release UI version preparation in a disposable clone in the same pinned web image; keep ordinary UI mode unchanged and scope Git trust to exact process-local directories.
- Replace CUDA sudo package installation with preinstalled utility/library validation.
- Select CUDA0 explicitly in CUDA smoke and require positive GPU layer offload plus a nonzero CUDA0 model buffer after inference, independently for each PID. CPU rows remain CPU. A complete CUDA job should emit five records (normal/headless, compatibility, constrained normal/headless).

## Landing sequence — do not merge as-is
1. Land/reconcile the narrow UI Git-ownership repair #1687 first, preserving its exact/malformed/mismatched SHA, differently-owned checkout, child Git and unchanged-global-config regressions.
2. Maintainer lands catalog-only commit `9e7bf7169a43c43a60ebf250f3d4baef6e3fbe54`.
3. Rebase this implementation on that catalog. Protected PR planning requires byte-identical catalogs; this combined PR is expected to fail Plan until then. Do not bypass that gate. Review and execute the CPU transition repair below without enabling unqualified accelerator rows.
4. Obtain candidate-definition/live evidence below before implementation merge.

## Rebase on #1672
Rebased onto main `afa36ab02`, preserving its product-integration tests and explicit Metal/Vulkan/ROCm device selection. Compatibility smoke honors `MESH_COMPAT_DEVICE` before the shared core-smoke selector. CUDA verification remains CUDA-specific; other backend runs do not claim CUDA qualification. Follow-up `cdbe5e16f` ports preinstalled utility/library checks to the typed CUDA workflow too, retaining GPU/driver provenance. Both preflight blocks are executed with isolated tool/library doubles covering failure paths. Typed callers now exercise the actual standalone/compatibility device guards, not only independently generated manifest values.

## Validation
[...]

PR #1695, opened 2026-09-08
fix(ci): catch SDK runtime JSON breaks before main

michaelneale:
## Summary

Catch CLI runtime-JSON changes in PR CI before they break the SDK runtime reader on main, without running every SDK suite on every PR.

- Run the existing shared SDK runtime reader against the actual composed Linux CPU host during its already-selected readiness check. Runtime fallback building is explicitly disabled; an incompatible report fails before product archival/publication.
- Preserve #1675's exact changed paths (and its formatter alone) as routing regressions: they select the Linux CPU product and producers even though full SDK suites remain unselected.
- Test catalog-wrapped and legacy reports through the real composer/reader/package verifier, plus malformed and unsupported reports that must block publication. The test host and network-readiness probe are stubs; this is not evidence of real-host GitHub execution.

Nine production lines; all other changes are tests/fixture and required CI documentation. No catalog, workflow YAML, runner, CUDA, release preparation, or recurrent-routing changes. The existing protected workflow checks out candidate source before invoking the local composer, so this path does not need a catalog bootstrap or candidate-workflow dispatch.

This is the narrow replacement for closed #1691. #1684's reader compatibility fix and #1687's release UI fix are preserved.

## Validation

On the final content based on `d6f09e098` (before the metadata-only commit):

- `just ci-validate`: passed, 943 tests / 7 skips plus actionlint, whitespace and all four xtask consistency gates.
- `just ci-shellcheck scripts/ci-compose-product-input.sh`: passed.
- `bash -n scripts/ci-compose-product-input.sh`: passed.
- `python3 -m py_compile scripts/tests/test_ci_sdk_json_consumer.py`: passed.
- Negative control substituting the unchanged main composer **only in temporary test copies**: four subcases failed as expected.
- Jimmy independently reviewed the six-file diff read-only and reported no blocking correctness/scope findings; he did not repeat the tests.
- Owned Cargo build output cleaned. Both protected catalogs remain byte-identical to main.

### GitHub execution evidence

Verified on exact head `d524f46bdb65578bac154317475ccb65e6179ed1`:

- All five Plan jobs passed.
- [Linux CPU composition job 102009345529](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34206764156/job/102009345529) **executed and succeeded**, not skipped. Its log records protected workflow `d6f09e098`, candidate checkout `d524f46bd`, `Checking CLI runtime JSON with the shared SDK runtime reader`, verified/reused adjacent CPU runtime, client readiness, and successful product upload.
- All five PR workflows and their stable `PR / Quality`, `PR / Website`, `PR / Linux`, `PR / macOS`, and `PR / Windows` checks succeeded. [Linux run 34206764156](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34206764156) finished at 09:39 UTC; Rust SDK smoke, Kotlin SDK smoke, CUDA inference smoke, and model-download smoke all executed and passed.
- Expected selective skips remain, including the product-integration suites and Qwen3.5 recurrent migration gate. This patch does not claim that those unselected checks executed.
- Final review check: GitHub has no inline review threads; CodeRabbit reported no actionable comments (its docstring-coverage advisory remains). Copilot failed on billing configuration and is not review evidence. Jimmy's independent read-only review is recorded above.
- At the 09:59 UTC handoff, current main was `592de242b871c0c35743683f91d4f829257972fc`, one commit ahead of this PR's base (`#1683`, family-battery catalogs/test changes). Its changed files do not overlap this six-file patch or alter the selected composer/reader/workflow path. GitHub reports this PR `MERGEABLE` / `BEHIND`; no rebase or rerun was performed, so the results above attest to the exact PR head, not a new combined revision.
- Separately, all five Main workflows at `592de242b` succeeded, including [Main Linux](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34209960871). Existing configuration/plan skips remain; this is not proof that every possible main/release check ran, and it does not include this unmerged repair.
- No merge, dispatch, rerun, runner/configuration change, or new source commit was performed. The pushed worktree is clean; Cargo build output was already cleaned. Keep the worktree until the PR merges or closes.

Origin: Buzz `mesh-releases`, channel `a7d0b8ef-9b25-432d-ae19-bacd410c81eb`, thread `bbc68483f722d0f37f5c58b59c98aa820e1d3ab6a52caa4ab890fac717f4232d`.

### DIFF
--- a/scripts/ci-compose-product-input.sh
+++ b/scripts/ci-compose-product-input.sh
@@ -270,6 +270,15 @@ if [[ "$INPUT_READINESS_SMOKE" == "true" ]]; then
         "$output_dir/$INPUT_BINARY_NAME" --log-format json --version
     MESH_LLM_NATIVE_RUNTIME_BUNDLE_DIR="$output_dir/native-runtimes" \
         "$output_dir/$INPUT_BINARY_NAME" --log-format json runtime list
+    # Exercise the shared SDK reader even when full SDK suites are unselected.
+    # CPU needs no accelerator/driver; reuse only the just-composed producer bytes.
+    if [[ "$INPUT_BACKEND" == "cpu" && "$(uname -s)" == "Linux" ]]; then
+        echo "Checking CLI runtime JSON with the shared SDK runtime reader"
+        MESH_SDK_NATIVE_RUNTIME_BUILD_FALLBACK=0 \
+            scripts/ci-prepare-native-runtime.sh \
+                "$output_dir/sdk-runtime-fallback" cpu \
+                --reuse-from-binary "$output_dir/$INPUT_BINARY_NAME"
+    fi
     scripts/ci-client-readiness-smoke.sh \
         "$output_dir/$INPUT_BINARY_NAME" \
         "$output_dir/native-runtimes"
--- a/scripts/tests/test_ci_sdk_json_consumer.py
+++ b/scripts/tests/test_ci_sdk_json_consumer.py
@@ -1,0 +1,174 @@
+"""The #1675 CLI JSON boundary must run even without full SDK selection."""
+
+import hashlib
+import json
+import os
+from pathlib import Path
+import shutil
+import subprocess
+import tempfile
+import unittest
+
+from scripts.tests import test_ci_artifact_actions as artifacts
+from scripts.tests import test_ci_prepare_native_runtime as runtime_tests
+from scripts.tests import test_plan_ci as planner_tests
+
+
+ROOT = Path(__file__).resolve().parents[2]
+
+
+class SdkJsonConsumerTests(unittest.TestCase):
+    def test_original_cli_change_selects_the_consumer_and_producers(self) -> None:
+        # Exact changed paths at #1675 head 14ec8d0, not this CI repair's
+        # control-plane paths (which would mask the miss by selecting everything).
+        payload = planner_tests.fixture("runtime-catalog-pr-1675.json")
+        for paths in (
+            payload["changed_files"],
+            ["crates/mesh-llm-commands/src/runtime_native/formatters.rs"],
+        ):
+            with self.subTest(paths=paths):
+                plan = planner_tests.PLANNER.build_plan(
+                    {**payload, "changed_files": paths}, root=ROOT
+                )
+                self.assertNotIn("ci-control", plan["domains"])
+                self.assertEqual(plan["matrices"]["sdk"], [])
+                self.assertIn("ui-artifact", plan["required_slices"])
+                self.assertIn("runtime-product", plan["required_slices"])
+                self.assertEqual(
+                    [row["id"] for row in plan["matrices"]["runtime_products"]],
+                    ["linux-cpu"],
+                )
+                self.assertEqual(
+                    [row["id"] for row in plan["matrices"]["hosts"]],
+                    ["linux-amd64-host"],
+                )
+
+        lane = (ROOT / ".github/workflows/ci-linux-lane.yml").read_text()
+        product = lane.split("  runtime_product:\n", 1)[1].split(
+            "  kotlin_sdk_input:\n", 1
+        )[0]
+        self.assertIn("needs: [hosts, native_runtimes]", product)
+        self.assertIn("uses: ./.github/workflows/ci-linux-product-slice.yml", product)
+        workflow = (ROOT / ".github/workflows/ci-linux-product-slice.yml").read_text()
+        self.assertIn("ref: ${{ inputs.source_sha || github.sha }}", workflow)
+        self.assertIn("uses: ./.github/actions/compose-product-input", workflow)
+        self.assertNotIn("readiness_smoke:", workflow)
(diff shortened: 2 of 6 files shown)


scripts/ci-compose-product-input.sh
```diff
@@ -270,6 +270,15 @@ if [[ "$INPUT_READINESS_SMOKE" == "true" ]]; then
         "$output_dir/$INPUT_BINARY_NAME" --log-format json --version
     MESH_LLM_NATIVE_RUNTIME_BUNDLE_DIR="$output_dir/native-runtimes" \
         "$output_dir/$INPUT_BINARY_NAME" --log-format json runtime list
+    # Exercise the shared SDK reader even when full SDK suites are unselected.
+    # CPU needs no accelerator/driver; reuse only the just-composed producer bytes.
+    if [[ "$INPUT_BACKEND" == "cpu" && "$(uname -s)" == "Linux" ]]; then
+        echo "Checking CLI runtime JSON with the shared SDK runtime reader"
+        MESH_SDK_NATIVE_RUNTIME_BUILD_FALLBACK=0 \
+            scripts/ci-prepare-native-runtime.sh \
+                "$output_dir/sdk-runtime-fallback" cpu \
```
ndizazzo, line 279: This output dir sits inside `$output_dir`, which is exactly what gets tarred a bit further down. It's safe today because `--reuse-from-binary` returns before the `rm -rf "$OUT_DIR"` build branch ever runs, so nothing actually gets created here. But it only stays safe as long as that early exit holds and the fallback default stays off. Could we point it at `${RUNNER_TEMP:-/tmp}` and drop the coupling entirely?

scripts/tests/test_ci_sdk_json_consumer.py
```diff
@@ -0,0 +1,174 @@
...
+            '#!/bin/sh\ntouch "$GITHUB_WORKSPACE/fallback-ran"\nexit 99\n'
+        )
+        (scripts / "package-native-runtime.sh").chmod(0o755)
+        # Exercise the Linux-only gate on development hosts too, without changing
+        # the immutable package fixture into an actual loadable Linux runtime.
+        tools = workspace / "bin"
+        tools.mkdir()
+        (tools / "uname").write_text("#!/bin/sh\nprintf 'Linux\\n'\n")
+        (tools / "uname").chmod(0o755)
+        host_input, runtime_input = artifacts.CiArtifactActionTests().write_fake_product_inputs(
```
ndizazzo, line 94: Instantiating a TestCase outside a runner just to borrow the helper works, but it reads a little odd. Maybe pull `write_fake_product_inputs` out to a module-level function so both files can call it directly?

ndizazzo (approved):
This is the right direction. Nine lines of production change, the mechanism is sound, and the test earns its keep.

I checked it out and ran it rather than taking the description at face value:

- `test_ci_sdk_json_consumer.py` passes on the head commit (3 tests, OK, under Python 3.13).
- Negative control: I reverted only `scripts/ci-compose-product-input.sh` back to main and left everything else alone, and got 4 failures on the `sdk-reader-ran` assertion. So the test really does fail without the fix, which matches what you said.
- `just ci-shellcheck scripts/ci-compose-product-input.sh` is clean.
- Reading the composer, the new block lands before `tar -czf product-input.tar.gz` and before the `$GITHUB_OUTPUT` writes, so bad runtime JSON actually blocks publication instead of just going red after the product is already out.
- `ci-linux-product-slice.yml` checks out `inputs.source_sha` before the composite action runs, so the candidate script runs against candidate source. Agreed that no catalog or workflow change is needed.

I like that the test stubs almost nothing. Just the client smoke, the packager (exit 99 plus a `fallback-ran` sentinel so we can prove no build happened), and `uname`. The real composer, real reader, and real verifier all run against checksum-bound fixture bytes, and it asserts the negative direction too: malformed and unsupported reports leave no `product-input.tar.gz` behind.

One thing we should fix in the description before this merges. The body says "No catalog, workflow YAML, runner, CUDA, release preparation, or recurrent-routing changes." That's true file-wise but not behaviorally. `.github/actions/compose-product-input` is called 8 times in `release.yml`, all with `readiness_smoke: "true"`, and two of those are Linux CPU (`backend: cpu` at line 926, and `backend: ${{ matrix.flavor }}` at line 364). Since the gate only keys on cpu plus Linux, release composition now runs the SDK reader too. I think that's a good outcome, we shouldn't ship a release product whose runtime JSON the SDK can't parse. But let's say it out loud instead of leaving it for someone to trip over later.

Worth rebasing before merge, too. The branch is a commit behind, and the catalog-shape case depends on the catalog-aware parser from #1684 being in the base. I confirmed the older list-only parser rejects `{"runtimes": [...]}`, so the base does matter here.

Two small inline notes, neither blocking. (And the reader re-verifies the runtime package the composer already verified a few lines earlier. Costs a second, not worth changing unless it bugs you.)

i386 (approved):
Second approval alongside ndizazzo — CI-script change with fixture-driven tests; the SDK-runtime JSON contract check is a sensible pre-main gate.

PR #1696, opened 2026-09-08
Serve models from a local LLM server without a native runtime

### LINKED ISSUE #1204: Using openai-endpoint plugin without a runtime

michaelneale:
Fixes #1204

## What you can now do

Put mesh-llm on a machine that already runs Ollama (or LM Studio, LiteLLM, vLLM), point the `openai-endpoint` plugin at it, and `mesh-llm serve` starts — even when there is no MeshLLM native runtime for that machine's hardware. The mesh gets those models under their own names.

Before this change that node refused to start:

```
$ mesh-llm serve
no compatible native runtime found for Skippy ABI 0.1.52 on linux/x86_64: ...
run `mesh-llm runtime install` ...
```

Now:

```
$ mesh-llm serve --config config.toml
WARN starting without a MeshLLM native runtime: no compatible MeshLLM native runtime is
     installed or installable for MeshLLM 0.76.0-rc9 / Skippy ABI 0.1.52 ... This node can
     only serve models supplied by its configured external inference plugin, and cannot load
     local GGUF or layer-package models. Install a runtime with `mesh-llm runtime install`
     to serve models locally.
```

## When a runtime is still required

The relaxation is narrow. A missing runtime stays fatal unless **both** hold:

- an **enabled** plugin entry supplies a non-blank `url` (a URL is what makes a plugin an inference endpoint — the in-tree blobstore has none), and
- the startup names **no** local model — no `--model`, no `--gguf`, no configured `[[models]]` entry.

So `mesh-llm serve --gguf model.gguf` on a runtime-less box still fails exactly as before.

## Guarding the FFI

Continuing without a runtime means the Skippy symbols are unresolved. Two paths that could reach them now refuse with an actionable message instead:

- `start_local_openai_model` — a runtime load of a local model
- the stage control loop's `load` — an inbound split-stage request from a peer, which returns a rejected `StageReadyResponse` rather than erroring

## Architecture

The decision lives in one new module, `system/native_runtime_requirement.rs`, separated from config parsing so it is directly testable (8 unit tests). Startup resolution runs before the output surface is installed, so the warning is recorded and replayed by `run_runtime_cli` once the surface exists — otherwise it is invisible in both pretty and JSON log modes.

## Validation

Built with `just build` and exercised with the real binary against a stub OpenAI-compatible server on `:11434`, using an isolated `HOME`, native-runtime cache, and an unreachable manifest URL so no runtime can be found:

| Config | Result |
|---|---|
| plugin entry **with** `url` | starts, warning emitted, `/v1/models` + `/api/status` respond |
| plugin entry **without** `url` (control) | fails: `no compatible MeshLLM native runtime is installed or installable` |
| plugin `url` **plus** `--gguf` | fails, as intended |

`cargo test -p mesh-llm-host-runtime --lib`: 2988 passed. `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings` and `-p mesh-llm` clean; `cargo fmt --all --check` clean.

## Not covered here

Independent of this change, `openai-endpoint` 0.1.2 from the plugin catalog speaks plugin protocol 2 while `main` is on 3, so the plugin itself fails to load on current `main`. That is a separate compatibility issue in the plugin release, not in this gate — the runtime gate is passed before the plugin is started, which is what the control comparison above isolates.

Multiple simultaneous endpoints are still not expressible (one `openai-endpoint` entry, single `url` field). That is tracked separately.

## Summary by CodeRabbit

* **New Features**
  * Nodes configured exclusively for external inference can start without a native runtime.
  * External plugin models remain available in this configuration.
  * Startup warnings explain when local model serving is unavailable.

* **Bug Fixes**
  * Local model and stage loading now fail early with actionable guidance when the native runtime is unavailable.
  * External-inference-only startup no longer fails solely because the native runtime could not be loaded.

* **Documentation**
  * Updated usage guidance for external inference, local model requirements, and runtime setup.


michaelneale:
CI note: `Linux / CI / Linux` fails here while every real job in the lane passes. Not this PR — traced to a CI contract drift and filed as #1698: #1672 renamed the Linux smoke job selectors without updating `ci/slices.yml`, so the whole `product_smoke` slice silently skips and the summary then fails on a required slice that never ran. It reproduces on `main` alone via `just ci-validate`, and only bites PRs whose changed files map to the `rust`/`runtime-product` domains — which is why docs/CI-only PRs pass.

Virgile-pct (commented):
Tested on Windows 11 at cd8f84e1a (PR base 592de242b), host built with `--no-default-features --features web-ui,dynamic-native-runtime`. Windows is a natural bed for this change right now: a host built from this branch is Skippy ABI 0.1.52 and the rc9 catalog only has 0.1.44 runtimes (`runtime list --available` lists 13 artifacts, all 0.1.44), so `serve` has no installable runtime at all here, which is the situation the gate is for.

Startup matrix, `serve --config ... --port ...`, empty native-runtime cache, real release catalog:

| Config | Result |
|---|---|
| no plugin entry | fatal after 7 s, exit 1: `no compatible native runtime: 4 candidate(s) rejected, 9 set aside: ...`, byte for byte the output of a main (8578d0f46) build with the same config |
| `[[plugin]]` openai-endpoint with `url`, no model | starts: the WARN is printed, `api ready` follows 11 s later, `/v1/models` answers `{"data":[],"object":"list"}`. The same config on the main build fails exactly like the row above |
| same plus `--gguf <local file>` | fatal, exit 1, same message |
| same plus a `[[models]]` entry in config | fatal, exit 1 |
| same with `enabled = false` | fatal, exit 1 |
| same with `url = "   "` | rejected earlier by config validation: `plugin[0].url must be an absolute http:// or https:// URL` |
| plugin with `url`, default cache holding the rc7, rc8 and rc9 runtimes (ABI 0.1.39, 0.1.42, 0.1.44) | starts with the same WARN, up after 17 s |

The deferred warning shows up in both output modes: once in the pretty console (Mesh events panel) and once as `{"event":"warning","level":"warn","message":"starting without a MeshLLM native runtime: ..."}` with `--log-format json`, no duplicate from the tracing line. The message carries the whole resolution chain from #1675, which makes it long but it says exactly why nothing loaded.

`cargo test -p mesh-llm-host-runtime --lib --features dynamic-native-runtime native_runtime_requirement`: 8 passed. `cargo fmt --check`: clean. `cargo clippy --no-deps -p mesh-llm-host-runtime --all-targets --features dynamic-native-runtime -- -D warnings` stops on four Windows-only lints in `plugin/mcp/external_mcp.rs`, `plugin/transport.rs` and `plugin/mod.rs` (an unused import, a `needless_return` under `cfg(windows)`, two unused `cfg(test)` helpers); none of those files are touched by this PR and they are identical on main. With those four lints allowed the same clippy run is clean, so nothing new from this PR.

Two observations, neither blocking, in case they are useful:

1. A node started this way still advertises what it cannot serve. The `passive_mode` event reports `capacity_gb: 12.88` and `models_on_disk: [meshllm/Qwen3-8B-Q4_K_M-layers, unsloth/Qwen3-8B-GGUF:Q4_K_M]`, and `run_auto.rs` feeds that same on-disk scan into `node.set_available_models(...)`, which goes out in the announcement; `split_readiness.rs` counts a peer whose `available_models` lists the model as a stage source (`model_source_state` returns `available`). So peers see a 12.9 GB node with those models available, can plan on it, and only then hit the new stage-load rejection or the `start_local_openai_model` refusal. Clearing `available_models` and advertising no capacity when the requirement resolved to `OptionalExternalInference` would keep planners from picking it in the first place. Happy to take that as a follow-up if you prefer to keep this PR small.

2. The gate is decided from config only, before plugin resolution. An entry marked `[plugin.startup] optional = true` whose plugin is not installed still relaxes the gate (that is how I ran the matrix above without downloading the plugin), so the node starts, logs that it can serve the plugin's models, and serves nothing. Probably fine for a first cut, but the warning could name the plugin so the operator knows which one has to be installed.

On CodeRabbit's gating remark: `system/mod.rs` declares `native_runtime_requirement` unconditionally, so that one does not apply as written. I could not check the static configuration (`--no-default-features` without `dynamic-native-runtime`) here: `skippy-ffi`'s build.rs needs the patched llama.cpp ABI archives under `.deps/llama-build`, which this box does not have.

Not exercised here: the split-stage rejection (no second node with a matching runtime on this box) and a plugin actually serving models on Windows (the catalog plugin is protocol 2 against 3 on main, as the description says).

michaelneale:
Folded into #1692 as requested — one PR for the whole thing. Closing this.

PR #1700, opened 2026-09-08
ci: add downloaded models to llama canary

i386:
The llama canary did not cover the additional GGUF model families downloaded for the architecture battery. This adds 27 family rows backed by 29 immutable artifacts, including the Gemma 4 and Muse Glimmer projectors, and expands generated source ownership so upstream family changes remain generator-authored.

The cache gate now derives DFlash hyper-connection activation widths from GGUF metadata, which records DeepSeek 4 as four 4,096-wide streams. The checked-in llama.cpp family patch series was regenerated into 69 deterministic shards.

Validation:

- `just ci-validate` — 946 tests passed, 8 skipped; actionlint and repository consistency checks passed
- `scripts/check-skippy-generated-family-patch.sh` — 149 builders checked; deterministic and current
- transformed llama.cpp build — 33/33 native Skippy tests passed
- immutable cache preflight — all 27 new families passed SHA-256, size, layer-count, and activation-width checks

## Summary by CodeRabbit

- **New Features**
  - Added certification and artifact coverage for 27 additional model families, including new dense, MoE, vision, and assistant variants.
  - Added support for projector artifacts in applicable vision-capable models.

- **Bug Fixes**
  - Improved activation-width detection for DFlash models with hyper-connection metadata.

- **Tests**
  - Expanded validation to cover all 77 certified families and new model-specific behaviors.

- **Maintenance**
  - Updated model artifact references, integrity records, and parity evaluations to use the current Llama artifact source.


michaelneale:
A lot of families look like aliases of other families - surely?

i386:
@michaelneale  Some family names do look like aliases, but the canary keys on llama.cpp architecture tags, and the names map to genuinely distinct arch implementations — e.g. `ernie4-5` vs `ernie4-5-moe` (dense vs MoE), `granite-dense/hybrid/moe/swa/switch` (five distinct Granite4 variants), `deepseek2` vs `deepseek32` vs `deepseek4` (different MLA/MoE generations), `gemma` vs `gemma2/3/3n/4` (separate arch bumps). Where the same weights could serve two tags we have one artifact per family, and the just-landed dedupe (c6c6fbd90) plus a planned registry check enforce one download per family. The patch series in `third_party/llama.cpp/patches/generated/` is what actually enumerates the distinct arch code paths each row certifies.

PR #1705, opened 2026-09-08
ci: nightly agentic-replay coding benchmark on micstudio

i386:
## Nightly agentic-replay coding benchmark on `micstudio`

Closes the performance-regression gap identified in the mesh-llm KV/split testing strategy: today no nightly signal measures coding-agent serving performance. This adds a pinned-runner nightly that replays the Thoughtworks agentic-coding trajectory sample through `mesh-llm serve` for three coding models covering the three cache/state regimes, appends immutable history to an HF dataset, gates on regression after a bootstrap, and on failure runs an opencode repair loop that **always** opens a PR (verified fix, or flagged `needs-attention`).

### Model matrix (pinned in `ci/agentic-replay-nightly/matrix.json`)

| Model | Class | Cache regime |
|---|---|---|
| Qwen2.5-Coder-14B-Instruct Q4_K_M | dense | ResidentKv |
| gpt-oss-20b MXFP4 | MoE | MoE routing + KV |
| Qwen3-Next-80B-A3B-Thinking Q4_K_M | hybrid-recurrent | KvRecurrent |

Runner: **micstudio** (Mac Studio M3 Ultra, 80 GPU cores, 256 GB unified, macOS 26.6.2) — verified via direct SSH; fail-closed `RUNNER_NAME` check like the `white` competitive lane. Checkout is pinned read-only to `main`.

### Workload & statistics

- **Trace sample:** 8 trajectories per agent framework per pass (3 frameworks), 72 measured checkpoint requests per model per pass, plus a disjoint warm-up cohort; 2 passes (ABBA-ordered). The runner enforces ≥2× max concurrency in trajectories so every cohort gets multiple waves — ~24 measured requests per concurrency level per model.
- **Concurrency levels:** 1, 2, 4, 8 — one history row per level. The c1 row isolates single-stream decode/TTFT; c2/c4 exercise continuous batching and cross-session shared-prefix KV reuse; c8 is the stress arm. The c4→c8 saturation knee is itself a diagnostic: a regression only at c8 points at scheduler admission/queueing, at c1 at the kernel/attention path.

### What each run produces — full examples in `ci/agentic-replay-nightly/examples/`

- **JSONL history rows** (schema v2, one per model × concurrency level): `examples/run-2026-09-09.jsonl` (12 rows) — provenance (source SHA, backend binary hash, hardware fingerprint, model SHA-256, dataset hash, replay params) plus decode/E2E throughput, TTFT mean/p90, cache-hit %, and `finish_reason_length_pct`.
- **Markdown run report** (the report format for every run and PR): `examples/RUN_REPORT.md` — per-model table across all four concurrency levels with Δ-vs-baseline, 7-run trend, provenance block.
- **Published benchmark card** (auto-refreshed on the HF dataset + repo card): `examples/CARD.md` — latest results, trend, methodology.

### Fairness: the output-token cap

The replay caps per-turn output at 2,048 tokens, but each turn's budget is actually the recorded assistant output size (`recorded_output_budget()`, ≈chars/4, clamped to [8, 2048]) — most turns emit their natural length. For a paired serving benchmark this is fair: identical budgets per cohort keep comparisons apples-to-apples and decode-dominated. The residual risk (a budget change masquerading as throughput change) is covered by tracking `finish_reason_length_pct` per cohort — visible and gateable, not silent.

### Gate & repair loop

- `scripts/agentic-replay-history.py` — schema-v2 normalization, exact cohort-key comparison, bootstrap-then-gate (3 prior complete runs; ±5% tolerance per metric).
- `scripts/agentic-replay-repair.sh` — opencode analyzes evidence, attempts a fix, benchmark re-runs, PR always opened ("fix verified" or `needs-attention`).
- `scripts/agentic-replay-card.py` — renders the published card from history.
- Tests: `scripts/tests/test_agentic_replay_history.py` (4 passing).
- Repair-PR template: `.github/AGENTIC_REPLAY_REPAIR_PR_TEMPLATE.md` with worked examples for both outcomes (`examples/REPAIR_PR_RESOLVED.md`, `examples/REPAIR_PR_NEEDS_ATTENTION.md`).

### Required repo config before enabling

- Vars: `MESH_AGENTIC_REPLAY_NIGHTLY_ENABLED=1`, dataset repo name, `MESH_AGENTIC_REPLAY_GATE_ENABLED` (arm after bootstrap), pinned trajectory parquet path on micstudio.
- Secrets: `HF_TOKEN` (Hugging Face dataset write) and `CANARY_REPAIR_TOKEN` — the same repair token the llama canary already uses, shared for opening repair PRs.
- Create the HF dataset repo (`meshllm/agentic-replay-nightly`) seeded with `ci/agentic-replay-nightly/README.md` as its card.
- Micstudio runner setup mirrors the llama canary: pre-warmed read-only HF cache exported as `HF_CACHE` (workflow normalizes to `HF_HOME`/`HF_HUB_CACHE`, `HF_HUB_OFFLINE=1` for model reads; dataset writes opt out per-command), runner-local `hf` CLI + `opencode` on PATH, and a toolchain verification step that fails fast if anything is missing. Model downloads never go through GitHub Actions cache.

## Summary by CodeRabbit

- **New Features**
  - Nightly agentic-replay benchmarks now cover concurrency levels 1, 2, 4, and 8, with eight trajectories per framework.
  - Automated history tracking detects performance regressions and incomplete runs.
  - Benchmark cards summarize current performance and seven-run throughput trends.
  - Automated repair reports support resolved and needs-attention outcomes.
  - Pull requests can validate replay benchmark changes before merging.

- **Documentation**
  - Added dataset documentation, benchmark methodology, schema guidance, run reports, and repair examples.


ci/agentic-replay-nightly/matrix.json
```diff
@@ -0,0 +1,44 @@
...
+    "max_output_tokens": 2048
+  },
+  "models": [
+    {
+      "family": "qwen25-coder-14b",
+      "class": "dense",
+      "repo": "Qwen/Qwen2.5-Coder-14B-Instruct-GGUF",
+      "file": "qwen2.5-coder-14b-instruct-q4_k_m.gguf",
+      "quant": "Q4_K_M",
+      "sha256": null
```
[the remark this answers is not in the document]
i386: Fixed in 41cca9344. Every matrix entry now pins a 40-hex Hugging Face revision and a full SHA-256 digest sourced from that exact revision metadata. The workflow resolves the exact repo@revision/file, hashes all cached GGUFs before starting any replay, and fails closed on a missing or mismatched artifact. The history normalizer also rejects unpinned models and keeps the digest in the cohort identity and baseline filter; the schema and examples now require the same provenance.

scripts/agentic-replay-history.py
```diff
@@ -0,0 +1,192 @@
...
+        for line in path.read_text(encoding="utf-8").splitlines():
+            if line.strip():
+                row = json.loads(line)
+                runs[row["cohort"]["model"] + f'|c{row["replay"]["concurrency"]}'].append(row)
+    return runs
+
+
+def compare(rows: list[dict[str, Any]], baseline: list[dict[str, Any]]) -> list[str]:
+    problems: list[str] = []
+    prior = [r for r in baseline if r["complete"] and r["model"] == rows[0]["model"]]
```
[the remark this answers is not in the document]
i386: Resolved by the current implementation. The old location is outdated. Baselines are selected per row by model family, concurrency, and replay shape, then compare() requires the same model SHA-256. Matching the full cohort_key would prevent every intended source-code regression comparison because source_sha is part of that key and changes between candidate and baseline.

.github/workflows/agentic-replay-nightly.yml
```diff
@@ -0,0 +1,268 @@
...
+  cancel-in-progress: false
+
+jobs:
+  replay:
+    if: >-
+      github.repository == 'Mesh-LLM/mesh-llm' &&
+      github.ref == 'refs/heads/main' &&
+      (github.event_name == 'workflow_dispatch' ||
+       (github.event_name == 'schedule' && vars.MESH_AGENTIC_REPLAY_NIGHTLY_ENABLED == '1'))
+    runs-on: [self-hosted, X64, macOS, family-certify, agentic-replay]
```
ndizazzo: 👍🏻

.github/workflows/agentic-replay-nightly.yml
```diff
@@ -0,0 +1,268 @@
...
+          if [[ -z "$DATASET_FILE" ]]; then
+            echo "sessions.parquet not found in the HF cache" >&2
+            exit 1
+          fi
+          FAILED=0
+          for family in $(python3 -c "import json;print(' '.join(m['family'] for m in json.load(open('$MATRIX_FILE'))['models']))"); do
+            model_uri=$(python3 -c "import json;m=[m for m in json.load(open('$MATRIX_FILE'))['models'] if m['family']=='$family'][0];print(m['repo']+'@'+m['revision']+'/'+m['file'])")
+            if ! python3 evals/agentic-replay.py run \
+                --ref main=origin/main \
+                --ref pr=HEAD \
```
ndizazzo: These two resolve to the same commit on a scheduled run, because checkout is pinned to `ref: main` above. `parse_ref_specs` then raises `multiple labels resolve to commit <sha>`, so every family fails, the job goes red, and the repair loop opens a PR for a regression that never happened.

`agentic-replay-repair.sh` uses `--ref fixed=HEAD --ref base=origin/main`, which are genuinely distinct, so I think this pair got copied from a PR-comparison context and never actually ran.

The real fix is probably a single-ref mode in the harness. We want an absolute number for one commit here, and the harness insists on two distinct ones.

.github/workflows/agentic-replay-nightly.yml
```diff
@@ -0,0 +1,268 @@
...
+          repo = sys.argv[1]
+          try:
+              huggingface_hub.repo_info(repo, repo_type="dataset", token=os.environ.get("HF_TOKEN"))
+          except Exception as exc:
+              message = str(exc)
+              if "404" in message or "Entry Not Found" in message or "Repository Not Found" in message:
+                  sys.exit(3)
+              sys.exit(1)
+          PY
+          rc=$?
```
ndizazzo: This never runs when the dataset repo is missing. `set -euo pipefail` is on at line 162 and the heredoc above is a plain simple command, so `sys.exit(3)` kills the step before we reach `rc=$?`. The bootstrap branch at 187 is unreachable, which means the very first run hard-fails instead of seeding.

Separately, the job sets `HF_HUB_OFFLINE: "1"` and only the `hf download` line overrides it, so `repo_info` raises an offline error with no "404" in the message and takes `sys.exit(1)` anyway. Two things to fix on one path.

Something like `if python3 - "$DATASET_REPO" <<'PY' ...; then rc=0; else rc=$?; fi`, plus `HF_HUB_OFFLINE=0` on that call.

.github/workflows/agentic-replay-nightly.yml
```diff
@@ -0,0 +1,268 @@
...
+        run: |
+          set -euo pipefail
+          LEVELS=$(python3 -c "import json;print(' '.join(map(str, json.load(open('$MATRIX_FILE'))['replay']['concurrency'])))")
+          LEVEL_ARGS=()
+          for level in $LEVELS; do LEVEL_ARGS+=(--concurrency "$level"); done
+          DATASET_FILE="${MESH_AGENTIC_REPLAY_DATASET_FILE:-}"
+          if [[ -z "$DATASET_FILE" ]]; then
+            DATASET_FILE=$(find -L \
+              "$HF_HUB_CACHE/datasets--thoughtworks--agentic-coding-trajectories/snapshots" \
+              -mindepth 2 -maxdepth 2 -type f -name sessions.parquet -print -quit \
```
ndizazzo: The models get 40-hex revisions and post-download sha256 verification, but the trajectory dataset (which is the actual benchmark input) gets first-match-wins with no revision and no digest. If the cache ever picks up a newer snapshot, every historical comparison silently rebaselines and the row records nothing about which snapshot ran.

Can we pin this the same way we pin the models, and put the resolved revision or digest into the history row?

.github/workflows/agentic-replay-nightly.yml
```diff
@@ -0,0 +1,268 @@
...
+on:
+  schedule:
+    - cron: "23 14 * * *" # after the Linux competitive run settles
+  workflow_dispatch:
+
+# Scheduled runs execute the default-branch definition only; manual dispatch is
+# accepted only from main so caller-selected content never reaches micstudio.
+permissions:
+  contents: read
+  pull-requests: write
```
ndizazzo: Let's drop this. Repair PRs are opened with `CANARY_REPAIR_TOKEN` through `gh`, and checkout is `persist-credentials: false`, so nothing consumes the `GITHUB_TOKEN` PR scope. And if we did need it later, our rule is to scope write to the owning job rather than the whole workflow.

scripts/agentic-replay-repair.sh
```diff
@@ -0,0 +1,114 @@
...
+      --dataset-file "${MESH_AGENTIC_REPLAY_DATASET_FILE:-}" \
+      --output "$OUTPUT_DIR/repair/$family" || RERUN_FAILED=1
+  done
+  if [[ "$RERUN_FAILED" == "0" ]] && python3 scripts/agentic-replay-history.py \
+      --matrix ci/agentic-replay-nightly/matrix.json \
+      --replay-dir "$OUTPUT_DIR/repair" \
+      --label fixed \
+      --hardware "$OUTPUT_DIR/hardware.json" \
+      --source-sha "$(git rev-parse HEAD)" \
+      --replay ci/agentic-replay-nightly/matrix.json \
```
ndizazzo: This passes the whole matrix where the workflow passes only the `replay` sub-object. `history.py` filters baselines on `replay.trajectories_per_framework` and `replay.passes`, and neither exists at the top level of the full matrix, so both come back `None`, nothing matches, `len(prior) < BASELINE_MIN_RUNS`, `compare()` returns empty, and the gate exits 0.

Which means `RESOLVED=1` for any diff that runs to completion, including one that fixes nothing. `--passes 1` on line 51 against the nightly's 2 would break the match on its own too, even with the right payload.

scripts/agentic-replay-repair.sh
```diff
@@ -0,0 +1,114 @@
...
+git config user.name "mesh-replay-bot"
+git config user.email "replay-bot@meshllm.invalid"
+git checkout -b "$BRANCH"
+
+# 1. opencode analyzes the regression evidence and attempts a fix. It must not
+# see GitHub credentials.
+env -u GH_TOKEN -u GITHUB_TOKEN opencode run --mode agent \
+  "The nightly agentic replay benchmark on micstudio regressed. Evidence: $OUTPUT_DIR/summary/history.jsonl and per-model artifacts in $OUTPUT_DIR. Analyze the regression, identify the offending change (git log origin/main is available), and attempt a minimal fix. Do not touch ci/agentic-replay-nightly baselines or thresholds." || true
+
+if git diff --quiet; then
```
ndizazzo: This only sees tracked changes. If opencode fixes something by adding a new file we take the no-changes branch, make an empty commit, and open a NEEDS ATTENTION PR while quietly throwing the fix away. `git status --porcelain` instead?

scripts/agentic-replay-repair.sh
```diff
@@ -0,0 +1,114 @@
...
+  echo "opencode produced no changes — needs-attention" >&2
+  git commit --allow-empty -m "chore: agentic replay nightly regression needs attention (run ${GITHUB_RUN_ID:-local})
+
+Automated repair produced no changes; PR opened for human triage with the
+run evidence attached."
+else
+  git add -A
```
ndizazzo, line 39: `OUTPUT_DIR` and `HISTORY_LOCAL` both live under `$GITHUB_WORKSPACE` and neither is gitignored (checked with `git check-ignore`), so this stages the entire artifact dump. Every run.json, per-request JSONL, and chart would land in the repair PR.

scripts/agentic-replay-repair.sh
```diff
@@ -0,0 +1,114 @@
...
+    python3 evals/agentic-replay.py run \
+      --ref fixed=HEAD \
+      --ref base=origin/main \
+      --model "$model_uri" \
+      --backend metal \
+      --trajectories-per-framework 8 \
+      $LEVEL_ARGS \
+      --passes 1 \
+      --warmup-turns 4 \
+      --dataset-file "${MESH_AGENTIC_REPLAY_DATASET_FILE:-}" \
```
ndizazzo: This is empty unless someone sets the repo var, because the workflow resolves `DATASET_FILE` inside the replay step and never writes it to `$GITHUB_ENV`. `argparse type=Path` accepts `""` happily, so it passes the exactly-one-of validation and then fails on read. Exporting `DATASET_FILE` to `$GITHUB_ENV` from the replay step would cover it.

scripts/agentic-replay-history.py
```diff
@@ -0,0 +1,276 @@
...
+    if not isinstance(digest, str) or not re.fullmatch(r"[0-9a-f]{64}", digest):
+        raise ValueError(f"{family}: sha256 must be a 64-hex digest")
+
+
+def stable_hash(value: object) -> str:
+    encoded = json.dumps(value, sort_keys=True, separators=(",", ":")).encode()
+    return hashlib.sha256(encoded).hexdigest()
+
+
+def load_cells(replay_root: Path, family: str, label: str) -> list[dict[str, Any]]:
```
ndizazzo, line 59: This raises when a family produced no cells, and the replay step deliberately continues past failures (`if: !cancelled()`), so one failed model takes down history for the other two with an uncaught traceback and no `history.jsonl` at all.

The publish step then guards on `steps.history.outcome != 'skipped'`, which a failed step satisfies, so `hf upload` fails on the missing file right after. Skipping a missing family with a warning would keep the partial evidence, which is exactly when we'd want it.

scripts/agentic-replay-history.py
```diff
@@ -0,0 +1,276 @@
...
+                ),
+                "end_to_end_tokens_per_second": (
+                    completion_tokens / workload_seconds if workload_seconds > 0 else 0.0
+                ),
+                "ttft_ms_mean": statistics.fmean(ttft_ms) if ttft_ms else 0.0,
+                "ttft_ms_p90": (
+                    ttft_ms[int(0.9 * (len(ttft_ms) - 1))] if ttft_ms else 0.0
+                ),
+                "cache_hit_pct": mean_or_none(cache_pcts),
+                "finish_reason_length_pct": None,
```
ndizazzo: The description says the output-budget fairness risk is covered by tracking `finish_reason_length_pct` per cohort, "visible and gateable, not silent", but it's hardcoded to `None` on every row and never computed or gated anywhere.

Can we actually populate it? The fairness argument for `recorded_output_budget()` rests on this field existing.

scripts/agentic-replay-history.py
```diff
@@ -0,0 +1,276 @@
...
+            )
+        )
+
+    args.output.parent.mkdir(parents=True, exist_ok=True)
+    with args.output.open("w", encoding="utf-8") as handle:
+        for row in rows:
+            handle.write(json.dumps(row, sort_keys=True) + "\n")
+
+    problems: list[str] = []
+    if args.baseline and args.baseline.is_dir():
```
ndizazzo, line 319: Completeness only gets checked when a baseline exists, so on a bootstrap run a benchmark where everything failed still writes rows, prints "wrote N history rows", exits 0, and publishes. I'd flag `complete` whether or not we have priors.

Also `complete = failed == 0 and requests > 0` never checks that `successful + failed == requests`, so dropped requests read as complete. The committed example has exactly that shape: 72 prompts, 36 successful, 0 failed.

ci/agentic-replay-nightly/examples/run-2026-09-09.jsonl
```diff
@@ -0,0 +1,12 @@
...micstudio"}, "cohort_key": "0000000000000000000000000000000000000000000000000000000000000000", "complete": true, "created_utc": "2026-09-09T11:42:07Z", "decode_tokens_per_second": 38.4, "end_to_end_tokens_per_second": 31.2, "failed_requests": 0, "finish_reason_length_pct": 6.3, "hardware_fingerprint": {"chip": "Apple M3 Ultra", "gpu_cores": 80, "machine_model": "Mac15,14", "os_version": "26.6.2", "unified_memory_bytes": 274877906944}, "measured_wall_ms": 6284000, "model": {"class": "dense", "family": "qwen25-coder-14b", "file": "qwen2.5-coder-14b-instruct-q4_k_m.gguf", "quant": "Q4_K_M", "repo": "Qwen/Qwen2.5-Coder-14B-Instruct-GGUF", "revision": "d0a692ef765eefbf2fabb130b3cb2e8917e3d225", "sha256": "c1e659736d89ac1065fb495330fb824d94001974a4bfa78e7270e43476a8d940"}, "output_tokens": 368640, "prompt_count": 72, "replay": {"concurrency": 1, "dataset": "thoughtworks/agentic-coding-trajectories", "dataset_sha256": "4b2d\u2026", "max_output_tokens": 2048, "mode": "checkpoint", "passes": 2, "trajectories_per_framework": 8, "warmup_turns": 4}, "schema_version": 2, "source_sha": "c28f0e2cc0f9a3e7d1b24c5e8f6a90d21c3b47e5", "successful_requests": 36, "ttft_ms_mean": 412, "ttft_ms_p90": 618}
```
ndizazzo: I think this was hand-written rather than generated from a real run, and it disagrees with the code in a few places. `artifact_result` is a path here but the code emits `ok`/`incomplete`. `finish_reason_length_pct` is 6.3 but the code always emits null. `dataset_sha256` isn't something the code emits at all. And `cohort.concurrency: 4` with `replay.concurrency: 1` can't happen, since both come from one variable (and `baseline_key` uses the replay one, so this row would key as c1 while claiming c4).

Can we regenerate it from a real run once the lane works? Otherwise it'll mislead whoever reads it first.

ndizazzo (changes requested):
This is the right direction and I want this lane. The comparison design (ABBA passes, the disjoint warm-up cohort, pinned model revisions with sha256 verification) is more careful than most benchmark CI I've seen. It fits how we do nightlies, too: it sits outside the 5+5 PR/main split like the competitive and KV nightlies, the fail-closed `RUNNER_NAME` check mirrors the `white` lane, checkout is pinned to trusted `main` with `persist-credentials: false`, both actions are SHA-pinned, and you updated `ci/ci.md` and the inventory. No complaints there.

I don't think it can run yet though, and a few of the problems only surface the first night it's enabled. I've left inline comments on the specific ones. Three of them block:

The replay step passes `--ref main=origin/main --ref pr=HEAD` on a checkout pinned to `main`, so both labels resolve to the same commit and `parse_ref_specs` raises every time. That's bigger than a wrong flag, though. `evals/agentic-replay.py` is an A/B comparator ("at least two distinct commits"), and what this lane wants is a longitudinal absolute measurement of one commit. I think that needs a single-ref mode in the harness, not a different ref string.

The bootstrap path can't be reached, so run #1 fails instead of seeding history. Two separate reasons, both inline.

The repair loop's "fix verified" check passes unconditionally, so any opencode diff that merely runs to completion gets labelled resolved and PR'd against main. That's the one that worries me most, since it's the part that writes to the repo.

Two other things.

Scope: the real diff against the merge base is 66 files, +5030/-99, and roughly half of it is the Depot runtime-seed canary (`depot-canary.yml`, `runtime-seed-canary.py`, `ci/runtime-seed-evidence/34272984200-1/`, `runner-images.json`, plus most of the `ci/ci.md` and inventory churn). None of that is in the description. It's fine work, it just belongs in its own PR. Can we split it so the replay lane can be reviewed, and reverted, on its own?

Tests: `test_agentic_replay_history.py` has 8 tests (the description says 4, probably stale) and they all exercise `compare()` and `build_rows()` in isolation. Every blocker above lives in the invocation wiring, not the gate math. You already ship `test_depot_canary_workflow.py` in this same PR, so the pattern is right there. A test that asserts the workflow's `--ref` arguments and the repair script's `--replay` payload catches most of this mechanically. Same idea for `schema.json`: it's 222 lines and nothing references it, so nothing validates the example or the emitted rows against it.

Smaller stuff I didn't inline. `permissions` is at the workflow level rather than job-scoped (see the inline about it also being unused). Only the first step declares `shell: bash`. `retention-days: 30` is above our one-day default but matches the competitive nightly, so I'm fine with it. `cohort_key` hashes `source_sha`, which makes it unique per run and useless for matching, and the thing that actually does the matching (`baseline_key`) ignores quant, backend, and runner. The zero-baseline branch fires on any nonzero value since `abs` tolerance is 0.0, and `ttft_ms_*` default to 0.0 when there are no TTFT samples, so three priors without samples would make the next real measurement an automatic gate failure. And I'd want to see 720 minutes confirmed against 3 models x 2 refs x 2 passes x 4 concurrency levels with an 80B in the mix, plus the in-job repair rerun.

.github/workflows/agentic-replay-nightly.yml
i386: Fixed in 8fdaee9f0: the harness now supports one ref for absolute measurements, and the scheduled lane uses main=HEAD instead of comparing two labels for the same commit.

.github/workflows/agentic-replay-nightly.yml
i386: Fixed in 8fdaee9f0: the repo lookup now runs inside an if command so exit 3 reaches bootstrap handling, and HF_HUB_OFFLINE=0 is set for the network lookup.

.github/workflows/agentic-replay-nightly.yml
i386: Fixed in 8fdaee9f0: the trajectory dataset is pinned to commit cef72d1f4d0caabf85937adf8337a14b7522c782 and SHA-256 3dd8ec3546cf771ce4ab2ac6c51ccefdd621197fa997a2cefb430b50df808fb6. Those values flow through replay parameters into each history row and cohort identity.

.github/workflows/agentic-replay-nightly.yml
i386: Fixed in 8fdaee9f0: removed the unused workflow-level pull-requests write permission.

scripts/agentic-replay-repair.sh
i386: Fixed in 8fdaee9f0: repair normalization now receives only the replay object and reads the pass count from the same matrix as the nightly run.

scripts/agentic-replay-repair.sh
i386: Fixed in 8fdaee9f0: change detection now uses git status --porcelain --untracked-files=all, so newly added files are included.

scripts/agentic-replay-repair.sh
i386: Fixed in 8fdaee9f0: the output and history roots are added to the worktree-local Git exclude file before opencode runs, so git add -A cannot stage benchmark artifacts.

scripts/agentic-replay-repair.sh
i386: Fixed in 8fdaee9f0: the pinned, verified dataset path is exported through GITHUB_ENV and reused by the repair rerun.

scripts/agentic-replay-history.py
i386: Fixed in 8fdaee9f0: missing-family cells now produce a warning and normalization continues with available families, preserving partial evidence.

scripts/agentic-replay-history.py
i386: Fixed in 8fdaee9f0: the stream records the terminal finish_reason, cells count length finishes, history emits the percentage, and the metric participates in cohort regression gating.

scripts/agentic-replay-history.py
i386: Fixed in 8fdaee9f0: completeness now requires successful plus failed to equal expected requests, and incomplete or empty runs fail even when the regression gate is disabled.

ci/agentic-replay-nightly/examples/run-2026-09-09.jsonl
i386: Fixed in 8fdaee9f0: removed the hand-written shard. A replacement should only be committed from a real successful lane run.

ndizazzo (approved):

PR #1709, opened 2026-09-09
feat(mesh): carry an optional self-reported claimed log head on PeerAnnouncement

StevenMih:
Adds claimed_log_head: Option<ClaimedLogHead> (proto tag 51 — 50 is taken by cache_affinity) to the gossip-wire PeerAnnouncement message, and a new ClaimedLogHead message with five fields: log_id, size, root, timestamp_unix_ms, and claimed_signature.

A ClaimedLogHead is an optional, opaque, self-reported claim a node MAY advertise about the head of its own append-only history. mesh-llm never verifies claimed_signature itself — nothing in this crate checks the claim, which is why the type and field are named "claimed" rather than "attested" or "verified". A receiving peer MAY verify the claim independently, using whatever key material and log format it understands; that verification does not exist in this repository today. A consumer that does verify may define its own VerifiedLogHead type — none is added here.

The signing contract (documented in proto/Rust comments): claimed_signature is claimed to be an Ed25519 signature by the announcing peer's own node key (the same key backing its endpoint_id) over the canonical encoding of (log_id, size, root, timestamp_unix_ms).

Untrusted remote bytes are bounded at the conversion boundary (this was raised in review and is now addressed): log_id at 512 bytes, root at 64 bytes, claimed_signature at 128 bytes — each with its own named constant (MAX_CLAIMED_LOG_ID_BYTES, MAX_CLAIMED_LOG_ROOT_BYTES, MAX_CLAIMED_LOG_SIGNATURE_BYTES) and its own rationale, rather than borrowing an unrelated cap. These are memory-safety limits on untrusted peer bytes, not format assertions — they're sized generously enough to admit hash/signature schemes other than this node's own SHA-256/Ed25519 (e.g. SHA-512 roots, post-quantum signatures). A field over its bound decodes as absent (None) on the whole ClaimedLogHead, never a panic and never a partial struct.

Why

Today a caller receiving a PeerAnnouncement cannot observe whether a peer claims a log-head, or how far along it claims to be. Adding the head as an advisory field lets any future receiver verify or reconcile claimed heads from the same (peer, log_id, size) — without mesh-llm having to understand the log format.

Landing the wire field first, before any in-tree producer or verifier, lets receivers begin verifying as producers roll out. There is no need to coordinate producer and consumer in the same commit. To be explicit: nothing in mesh-llm produces a non-None value today — the producer is out of tree — and mesh-llm never verifies this field by design; that is a consumer's job, not this crate's.

Scope — wire field only
PeerAnnouncement / ClaimedLogHead structs and both proto conversion functions (local_ann_to_proto_ann / proto_ann_to_local) carry the field so it round-trips correctly, including the byte-length bounds above on the receive path.
No live-mesh plumbing: PeerInfo, the plugin-facing MeshPeer mirror, and the gossip merge/rebroadcast path (apply_transitive_ann, peer_meaningfully_changed, Node live-peer merge) are untouched.
Both announcement constructors (announcement_from_peer, build_local_announcement) hardcode claimed_log_head: None — no local producer is wired yet, and no received peer head flows into live mesh state. That is later, separate work.
ClaimedLogHead is pub(crate), matching the visibility of PeerAnnouncement::claimed_log_head — an earlier revision left the type pub with no way for an external caller to actually obtain one.
How tested

cargo test -p mesh-llm-host-runtime -p mesh-llm-protocol — all existing tests pass; eight tests cover this field:

proto_announcement_without_claimed_log_head_decodes_as_absent — encodes a PeerAnnouncement without the field to wire bytes via prost::Message, decodes back, and asserts claimed_log_head: None with every other field intact. Proves backward compat at the actual wire level.
claimed_log_head_roundtrips_through_proto_announcement — builds a ClaimedLogHead, encodes through local_ann_to_proto_ann, encodes to wire bytes and decodes them back (not just the in-memory conversion), then decodes through proto_ann_to_local; asserts all five fields survive byte-identical.
Six boundary tests, one pair (at-limit / one-byte-over-limit) per bounded field (log_id, root, claimed_signature): each proves the field decodes as present exactly at its bound and as absent one byte past it.

cargo clippy -p mesh-llm-host-runtime -p mesh-llm-protocol --all-targets -- -D warnings clean. cargo fmt -p mesh-llm-host-runtime -p mesh-llm-protocol -- --check clean.

Why a top-level field rather than subprotocols

Recording the reasoning, since it was asked for.

The short version: the field is gossiped so that the thing you ask for can be checked against something you didn't ask for.

The intended split is that the head is broadcast — small, continuous, identical to everyone — while the log's actual contents are fetched bilaterally, on request, and never gossiped. That asymmetry is the whole point. A peer's answer to a direct request is checkable against a commitment it has already been making publicly and continuously to everyone else. Collapse both onto the request path and they arrive from the same party, in the same answer, at the same moment — and a peer that misreports simply tells a self-consistent story. The broadcast half is what gives the requested half something to disagree with.

It also means a node can be assessed at all before anyone has transacted with it. Without an ambient signal there is nothing to see until you have already sent a request, which is the wrong order.

On subprotocols = 37 specifically: MeshSubprotocol is {string name, uint32 major, repeated string features} — capability advertisement, with no payload field. ClaimedLogHead carries two bytes fields and two uint64s. Fitting it there needs either a new payload field on MeshSubprotocol — a change to a message far more widely used than this one — or serialising the head into repeated string features. The second would mean inventing a second ad-hoc encoding on top of the canonical one this PR now defines, which is the exact failure mode raised in the first review comment. subprotocols answers what do I speak; this answers what am I currently claiming, and those want different shapes.

Conceding the real cost: the tag is spent before the in-tree consumer exists, and that is a genuine cost rather than a technicality. The consumer is a node-local view of peer history, built and running out-of-tree, and it is queued to be offered here once the protocol pieces it depends on have landed — this field is one of them. Described in more detail on #1233 rather than here. SignedMeshGenesisPolicy on this same message is the closest existing precedent for a top-level, signed, self-reported claim.

## Summary by CodeRabbit

- **New Features**
  - Peer announcements can now include optional claims about the head of an append-only log, including its size, root, timestamp, and signature metadata.
  - Claims are carried through the protocol opaquely for downstream consumers and are not verified by the mesh runtime.

- **Compatibility**
  - Announcements without log-head claims continue to decode successfully.
  - Oversized claim fields are safely ignored during decoding to protect protocol boundaries.


StevenMih:

Local results, re-run against head `be6062c01`:

```
```
→ **2981 passed, 0 failed, 11 ignored** (finished in 49.26s)

```
```

```
```
→ **clean**

michaelneale:
Reviewed on Mic's behalf. Wire compatibility is clean: `optional PeerCheckpointHead checkpoint = 51` (`proto/node.proto:69`) is a fresh tag after `cache_affinity = 50`, it's protobuf rather than a fixed-layout codec, and both directions are tested — present at `protocol/tests/announcements.rs:801-876`, absent at `:883-896`. Old peers decode to `None`. That's the right coverage.

Two things I'd like addressed.

**1. No length validation on untrusted remote bytes.** The converter is an unconditional clone:

```rust
// protocol/convert.rs:854-864
fn proto_checkpoint_to_local(
    checkpoint: &crate::proto::node::PeerCheckpointHead,
) -> crate::mesh::PeerCheckpointHead {
    crate::mesh::PeerCheckpointHead {
        log_id: checkpoint.log_id.clone(),
        mmr_size: checkpoint.mmr_size,
        root: checkpoint.root.clone(),
        timestamp_unix_ms: checkpoint.timestamp_unix_ms,
        signature: checkpoint.signature.clone(),
    }
}
```

The proto comments document `root` as 32 bytes and `signature` as 64, but nothing enforces it — a hostile peer can send megabyte `root`/`signature`/`log_id` values and they're accepted and retained. The sibling converter in the same file does exactly what I'd want here: `proto_cache_affinity_to_local` uses `advertisement.salt.as_slice().try_into().ok()?` at `:645` and an explicit `entries.len() > CACHE_AFFINITY_MAX_ENTRIES` check at `:649`; other remote strings are bounded by `MAX_REMOTE_MODEL_NAME_BYTES`. Six lines to match that pattern. Mitigating factor: `gossip.rs:1208` never rebroadcasts a peer's checkpoint, so there's no transitive amplification — this is bounded-per-peer memory, not a mesh-wide vector.

**2. A field named `signature` that is never verified.** I grepped `PeerCheckpointHead` across `crates/` at this head; outside tests and generated proto, every use is a plain field copy or `None` (`convert.rs:843-864`, `gossip.rs:1208`, `:1275`). There is no Ed25519 verify and no consumer.

The PR is honest about this in three places (`node.proto:64-73`, `peer_state.rs:169-171` says outright that "mesh-llm never verifies `signature` itself"), which I appreciate — but a local type called `PeerCheckpointHead` carrying `signature: Vec<u8>` will be read by the next person as an attestation, and the disclosure lives in a comment rather than in the type. Please either verify at the ingest boundary in `proto_checkpoint_to_local`, or push the unverifiability into the name (`UnverifiedPeerCheckpointHead`, or `claimed_signature`) so a downstream caller cannot mistake it. This is the kind of thing that's cheap now and expensive after three consumers assume it means something.

**Nits:** `PeerCheckpointHead` is `pub` (`mesh/mod.rs:165`) while `PeerAnnouncement::checkpoint` is `pub(crate)`, so external callers can name the type but never obtain one. And nothing yet produces a non-`None` checkpoint (`gossip.rs:1269-1276` admits this) — fine as staged plumbing, worth stating in the PR body so it isn't mistaken for a live feature.

Source reading only — I did not build or run the suite.

StevenMih:
**Rename + reframe — force-push `6966f11b1`**

### Why the rename

The original names `PeerCheckpointHead` and the field `checkpoint` collided
with mesh-llm's existing ML-checkpoint vocabulary — `checkpoint_quantization`,
`checkpoint_imatrix`, `SparseCheckpointPolicy`, `hf_checkpoint`, and 276+
other uses of "checkpoint" in the codebase, all meaning model checkpoint.
Using "checkpoint" for a new log-attestation concept would make search,
reviews, and grep results ambiguous.

New names:
- `PeerCheckpointHead` → **`AttestedLogHead`** (proto message + Rust type)
- `PeerAnnouncement.checkpoint` → **`PeerAnnouncement.attested_log_head`**
- `mmr_size` → **`size`** (neutral; no log-format assumptions in the field name)
- Internal helpers renamed to match: `local_attested_log_head_to_proto`,
  `proto_attested_log_head_to_local`

Wire tag 51 is unchanged — this is a rename only, not a wire-format change.

### Reframe

PR description rewritten to focus on the concrete mesh-llm receiver use-case:
a node MAY advertise an opaque, signed log-head; a peer MAY verify it.
mesh-llm carries the field without understanding the log format. The "why
land wire-field first" rationale is now explicit.

### Nit fixes in this push

**(P3) Backward-compat test** — the original test used `..Default::default()`
which only exercises the Rust default, not the actual proto3 wire decoding.
Replaced with a test that encodes to wire bytes via `prost::Message::encode`,
decodes with `PeerAnnouncement::decode`, and then calls `proto_ann_to_local`;
asserts field absent and `gpu_name` intact.

**(P3) Length doc** — `root` and `signature` byte-length constraints are now
documented as "carried as-received; length is not enforced at this layer" in
both the proto comment and the Rust struct comment. No enforcement added (it
would be wrong to reject at this layer without a verifier). Byte hints removed
from field comments to avoid promising what is not checked.

### Local CI

```
cargo test -p mesh-llm-host-runtime -p mesh-llm-protocol
```
→ 2982 passed, 0 failed

```
cargo clippy -p mesh-llm-host-runtime -p mesh-llm-protocol --all-targets -- -D warnings
```
→ clean

```
cargo fmt -p mesh-llm-host-runtime -p mesh-llm-protocol -- --check
```
→ clean

Not run through upstream CI (fork PRs do not have upstream Actions access).

crates/mesh-llm-protocol/proto/node.proto
```diff
@@ -61,6 +61,25 @@ message PeerAnnouncement {
...
+// announcing peer's own node key (the same key backing its `endpoint_id`)
+// over the canonical encoding of (log_id, size, root, timestamp_unix_ms).
+// mesh-llm carries this opaquely and never verifies `signature` itself.
+// `root` and `signature` are carried as-received; length is not enforced here.
+message AttestedLogHead {
+  string log_id = 1;
+  uint64 size = 2;
+  bytes root = 3;
+  uint64 timestamp_unix_ms = 4;
+  bytes signature = 5;
```
ndizazzo: This is the part I can't implement against. "Canonical encoding of (log_id, size, root, timestamp_unix_ms)" doesn't tell me the field order on the wire, whether values are length-prefixed, how the two `uint64`s are serialized, or whether there's a domain-separation tag. Anyone writing a verifier from this comment is guessing.

Could we spell out the exact byte construction, and add a `string signature_algorithm` like `SignedMeshGenesisPolicy` has? Without the algorithm field we're locked into Ed25519 with no negotiation path, and that message already set the precedent a few tags up.

While we're here, I'd rather bound `root` and `signature` than declare them unbounded. See my note in the main review about `validate_peer_announcement`.

crates/mesh-llm-host-runtime/src/mesh/mod.rs
```diff
@@ -152,17 +152,17 @@ pub use node::{
...
     unused_imports,
     reason = "public compatibility re-export for existing mesh state callers"
 )]
 pub use peer_state::{
-    DisplayLatency, DisplayLatencySource, MeshCatalogEntry, NodeRole, OwnerRuntimeConfig,
-    PeerAnnouncement, PeerInfo, PropagatedLatencyObservation,
+    AttestedLogHead, DisplayLatency, DisplayLatencySource, MeshCatalogEntry, NodeRole,
+    OwnerRuntimeConfig, PeerAnnouncement, PeerInfo, PropagatedLatencyObservation,
+};
+pub(crate) use peer_state::{
```
ndizazzo: Can we put these two blocks back in their original order? Swapping them isn't related to the feature and it makes the diff on this file look bigger than it is, which matters more than usual here since `mesh/mod.rs` is one of the files that triggers the extra test requirements in AGENTS.md.

crates/mesh-llm-host-runtime/src/protocol/convert.rs
```diff
@@ -835,6 +835,34 @@ pub(crate) fn local_ann_to_proto_ann(
...
             .map(local_cache_affinity_to_proto),
+        attested_log_head: ann
+            .attested_log_head
+            .as_ref()
+            .map(local_attested_log_head_to_proto),
+    }
+}
+
+fn local_attested_log_head_to_proto(
+    checkpoint: &crate::mesh::AttestedLogHead,
```
ndizazzo: `checkpoint` looks like a leftover from when this was a `Checkpoint`. Let's rename it to `head` in both helpers so it matches the type.

crates/mesh-llm-host-runtime/src/mesh/gossip.rs
```diff
@@ -1202,6 +1202,10 @@ impl Node {
             latency_age_ms: Some(latency.age_ms),
             latency_observer_id: latency.observer_id,
             inference_admission_state: peer.inference_admission_state,
+            // No live-mesh attested-log-head state is tracked on `PeerInfo` yet — a
+            // rebroadcast of a peer we already admitted carries no opinion on
+            // its attested log head.
+            attested_log_head: None,
```
ndizazzo: This one's worth calling out explicitly in the proto comments and the PR body, because it's a real property of the design and not just a TODO.

Our gossip is transitive (`apply_transitive_ann`), but this is the rebroadcast path for an admitted peer and it hardcodes `None`, and `PeerInfo` never stores the field either. So relaying is structurally impossible: a consumer only ever sees attested heads from peers it's directly connected to, and gets nothing from anything two hops out, with no way to tell "this peer has no attestation" apart from "the attestation got dropped in relay." That's a surprising thing to discover after building on it.

crates/mesh-llm-host-runtime/src/protocol/tests/announcements.rs
```diff
@@ -793,3 +797,151 @@ fn test_proto_gpu_info_preserves_legacy_fields_for_old_consumers() {
     );
     assert_eq!(roundtripped.is_soc, Some(false));
 }
+
+#[test]
+fn attested_log_head_roundtrips_through_proto_announcement() {
```
ndizazzo: Both of these restate the whole ~45-field `PeerAnnouncement` literal. The sibling test modules already have constructor helpers for this (`test_announcement` in `mesh/tests/gossip.rs`, `peer_state_test_announcement` in `mesh/tests/peer_state.rs`). Could we use or add one? As written, every future field on `PeerAnnouncement` costs two more mechanical edits right here.

The assertions themselves are good, a five-field round-trip plus an explicit absent-decode check is exactly what I'd want.

ndizazzo (changes requested):
I like where this is headed mechanically, but I don't think we should merge it in this shape yet. The additive-compat reasoning is right and I confirmed it independently: tag 51 is genuinely free (50 is `cache_affinity`, 38/39 are a pre-existing gap), older prost nodes will ignore the unknown field, newer nodes see `None` from older peers, and there's no legacy `mesh-llm/0` counterpart to worry about since v0 only decodes `TunnelMap`. I also ran `cargo test -p mesh-llm-host-runtime --lib` on the branch and got 2982 passed, 0 failed, so your numbers check out.

My main hesitation is that we're adding permanent wire surface with a cryptographic contract we can't actually implement against. Three things I'd want sorted before this lands.

**The signing contract needs to be written down.** "The canonical encoding of (log_id, size, root, timestamp_unix_ms)" isn't defined anywhere in the repo, and I can't derive it from the comment. No field ordering guarantee, no framing or length-prefixing, no integer encoding, no domain-separation tag. Two people implementing this independently won't interoperate. And since the whole point of the field is that a receiver MAY verify it, an undefined canonicalization means we're paying for a signature and getting no assurance back. `SignedMeshGenesisPolicy` right above it already does this properly, it carries `origin_sign_public_key`, an explicit `signature_algorithm`, and signs over defined canonical proof bytes. Let's mirror that. I left a note on the message itself.

**We should bound the byte fields.** `validate_peer_announcement` in `mesh-llm-protocol/src/protocol/mod.rs` is where we do this: it already enforces `endpoint_id.len() == 32`, and `validate_endpoint_id_length` / `validate_config_hash_length` sit right below it doing the same job. `TunnelMap::validate_frame` does it too. Right now the only ceiling on `root`, `signature`, and `log_id` is `MAX_CONTROL_FRAME_BYTES` (8 MiB), which is a budget shared across every peer in a gossip frame, so one peer can fill it with attestation bytes and squeeze everyone else out. An Ed25519 signature is 64 bytes and `root` is a hash, these are easy to bound. The proto comment saying "length is not enforced here" documents the gap rather than closing it.

**There's no producer and no consumer.** Both construction sites hardcode `None`, and on the receive side we decode it, allocate it, and drop it. It never reaches `PeerInfo`, never reaches `MeshPeer`, never gets rebroadcast. We have no append-only-log concept anywhere in mesh-llm. So this is a permanent top-level tag on our most-touched message for a consumer that doesn't exist yet. If something out-of-tree needs it, `repeated MeshSubprotocol subprotocols = 37` is already our additive feature-discovery mechanism and doesn't burn a tag. If a top-level field really is the better call here, can you say why in the PR body? I'd like the reasoning recorded either way.

One repo-level thing that isn't your fault but I noticed while reviewing: `crates/mesh-llm-protocol/src/proto/node.rs` carries a `@generated by prost-build` header, but that crate has no `build.rs` and there's no regeneration or drift check anywhere in CI (unlike `skippy-protocol` and `mesh-llm-plugin`, which really do generate at build time). Your hand-edit matches the `.proto` field-for-field, I checked, but nothing enforces that. Worth a follow-up issue to either add a `--check` regen step or drop the misleading header.

The rest of my comments are smaller stuff inline.

michaelneale:
How did the proto thing get there? Odd. Feels like things can be addressed. I should make it so ci triggers for @StevenMih too.

StevenMih:
@michaelneale Yes! Turning on and enabling ci would help a lot. Thanks in advance. Will reply to the review as well but ci would probably save us iterations. Even if temporarily. You know this already probably but its: 
Settings → Actions → General → Fork pull request workflows from outside collaborators → "Require approval for first-time contributors only" (or add StevenMih as a collaborator). This is what "make it so ci triggers for StevenMih" actually means, and it's the one that saves iterations.

StevenMih:
Thanks both, updated this PR at 1011a5406.

Two of these changed the design: the signing contract is now something a third party can implement from, and the field's absence no longer claims to mean one thing when it can mean four.

One item I'd rather ask about than decide myself: the test-helper reuse (at the bottom).

crates/mesh-llm-protocol/proto/node.proto
StevenMih: Fixed, and you were right, you'd wouldn't be able to implement against what i had. 

ClaimedLogHead now carries string signature_algorithm = 6, matching the
SignedMeshGenesisPolicy shape you pointed out. 

The proto comment now spells out the exact byte string that gets signed, and it deliberately follows this crate's existing canonicalization rather than inventing one: u64 little-endian length prefixes as write_string/write_bytes do
(mesh/requirements.rs:1153,1188), domain tag b"mesh-llm-claimed-log-head-v1:" in the same style as the three tags at :19-23, lengths counted in bytes of the UTF-8 encoding, and the trimmed signature_algorithm bound into the signed input — both because leaving it out would let it be stripped or swapped, and because SignedMeshGenesisPolicy signs the trimmed value too (:802), so two peers can't disagree about whether " ed25519" and "ed25519" sign the same bytes.

The value itself is ed25519, i.e. the existing ED25519_SIGNATURE_ALGORITHM constant
(requirements.rs:24); matching is trimmed and case-sensitive, same as :813. An absent or empty signature_algorithm means the peer named no scheme — a verifier can't check the claim and must not assume one — which is the same rule :926 already applies. There's a test for that.

Your third point in this thread — bounding root/signature rather than declaring them unbounded— landed earlier in the round, before you wrote this; log_id, root, claimed_signature and now signature_algorithm each have their own named constant and their own rationale.

Since signature_algorithm is a new field on a message that has never shipped, there's no
compatibility question here.

crates/mesh-llm-host-runtime/src/mesh/mod.rs
StevenMih: Also already resolved by an earlier push in this round, so there's nothing new to look at — I
diffed mesh/mod.rs against the commit the branch started from to be sure, and the block order matches it; the only live change in that file is one line adding ClaimedLogHead to the existing pub(crate) use peer_state::{…} list. Point taken about AGENTS.md — that file should carry the minimum possible diff, and it does now.

crates/mesh-llm-host-runtime/src/protocol/convert.rs
StevenMih: Already done in the rename push before your review landed; both helpers take head now
(local_claimed_log_head_to_proto / proto_claimed_log_head_to_local). DONE

crates/mesh-llm-host-runtime/src/mesh/gossip.rs
StevenMih: Fixed, and thank you for insisting it belongs in the comment rather than a TODO. You're right that it's a property of the design.

Worse than you flagged, in fact: the old comment said absence meant the peer doesn't advertise one or hasn't produced one yet — two causes, stated as if exhaustive. There are four. The field comment now names all of them: not advertised; not produced yet; originated more than one hop away and never relayed (this implementation doesn't relay claimed_log_head through apply_transitive_ann, so a receiver only ever sees heads from directly-connected peers); and a bounded field over its limit, which drops the whole head on decode. The PR body gained a short "what a consumer can and cannot see" section saying the same thing in prose.

crates/mesh-llm-host-runtime/src/protocol/tests/announcements.rs
StevenMih: @ndizazzo Half-taken, and I'd like your call on the rest.

The duplication you objected to is gone: the ~45-field literal is stated once now, in a local
helper, not twice — so a new field on PeerAnnouncement costs one mechanical edit here, not two.

Actually reusing test_announcement I couldn't do without changing a module boundary. It lives in mesh::gossip's private #[cfg(test)] mod tests; its own pub(crate) doesn't help because the
path to it runs through private modules, and protocol::tests can't see in (E0603, checked with
cargo check, not just read). 

Reaching it means either widening mod gossip; to pub(crate) mod gossip; plus the nested mod tests, or relocating the helper into the crate-visible mesh::tests — and the first of those puts a non-feature change into mesh/mod.rs, which is the file your other comment quite reasonably asked us to keep quiet.

Both are your call rather than mine. Say which you'd prefer and I'll do it; otherwise the local
helper stays with a comment recording why.

StevenMih:
> How did the proto thing get there? Odd. Feels like things can be addressed. I should make it so ci triggers for @StevenMih too.

@michaelneale  Thanks for flagging this — chasing it down surfaced a concrete instance of the drift you sensed.

A wire-semantics divergence between node.proto and the committed node.rs. latency_source
(tag 41) is declared optional in node.proto — so prost-build emits Option<i32> — but the
checked-in node.rs has a plain i32. The contrast that makes it unambiguous is two fields down in
the same message: optional InferenceAdmissionState inference_admission_state = 49 generates
#[prost(enumeration = "…", optional, tag = "49")] Option<i32>, exactly as you'd expect, while tag
41 generates #[prost(enumeration = "LatencySource", tag = "41")] i32. Identical declaration shape,
two different results, one struct.

The effect is that presence and absence collapse on the wire: a peer that omits latency_source is
indistinguishable from one that sends 0 (LATENCY_SOURCE_UNSPECIFIED). That is observable in
tree — the local type is Option<LatencySource>, and both halves of the collapse are hand-written
in protocol/convert.rs: None => 0i32 on the way out, LatencySource::try_from(pa.latency_source) .ok() on the way back, which (since Unspecified = 0) turns None into Some(Unspecified). The
status JSON then renders as "LATENCY_SOURCE_UNSPECIFIED" what went out as null.

We'd call it latent rather than live, though, and want to be straight about that: the only
current consumer of that field in the UI does peer.latency_source ?? LatencySource.UNSPECIFIED, so
it collapses the two anyway. Nothing is broken today. We're flagging it rather than calling it a
bug — whether to reconcile it is your call, not ours.

Why it's there: crates/mesh-llm-protocol has no build.rs and no committed generator, and the
checked-in node.rs (marked @generated by prost-build) doesn't reproduce from a clean
prost-build 0.14.4 run against its own node.proto. 

In practice it's a hand-maintained @generated file — and, to be fair to whoever has been maintaining it, a careful one: the type inventory matches exactly (86 messages and enums on each side, none missing in either direction), and across all 72 messages latency_source is the only field whose presence semantics diverge. It's one surgical slip, not a mess.

What this PR did with it: we added ClaimedLogHead by editing node.rs to match the
surrounding generated style and updated node.proto in lockstep — a clean regen would have
rewritten a large amount of unrelated code and buried the actual change. Correctness here is
established by the round-trip test, not by byte-identity with a generator:
claimed_log_head_roundtrips_through_proto_announcement encodes via prost::Message, decodes, and converts, so the wire encoding is exercised directly.

If it'd be useful — and only if; we didn't want to fold a build-system change into this PR — the
destination is presumably the pattern crates/skippy-protocol already uses: build.rs +
prost-build + vendored protoc, generating into OUT_DIR with nothing checked in
(crates/mesh-llm-plugin does the same). 

You can't get there in one step from here, though: with the current file already diverging from a clean run, flipping straight to build.rs regenerates on every cargo build, so every contributor gets a dirty tree and anything relying on the current shapes stops compiling. The intermediate that makes the drift visible without forcing it is a deliberate cargo xtask gen-proto (or a just recipe) plus a CI job that warns on divergence — reconcile field by field, then adopt the skippy pattern and delete the checked-in file. Sequencing and appetite are entirely yours; happy to do any part of it if you want it.

(And since you offered to enable CI on our fork — once those workflows run, a fresh-generator drift
check is exactly the kind of thing that would surface this automatically.)

crates/mesh-llm-host-runtime/src/protocol/tests/announcements.rs
ndizazzo: If that's the case, then I don't think expanding mesh gossip is the right way, so I guess it's fine to accept the duplication as a tradeoff... thanks for double checking on this

StevenMih: Thanks for your call and I'll leave it as is.

For the record of why, since the thread will outlive the memory: the duplication you originally objected to is gone — the ~45-field literal is stated once now, in a local helper, not twice, so a new field on PeerAnnouncement costs one mechanical edit here rather than two. Reusing test_announcement was the part that couldn't be done cheaply: it lives in mesh::gossip's private #[cfg(test)] mod tests, and its own pub(crate) doesn't help because the path to it runs through private modules — protocol::tests gets E0603 (checked with cargo check, not assumed). Getting at it would have meant widening mod gossip; plus the nested mod tests, which puts a non-feature change into mesh/mod.rs — the file your other comment asked us to keep quiet. The comment recording that stays in the test file.

StevenMih:
Updated head `3876a3a1a`: corrected the comment on `MAX_CLAIMED_LOG_SIGNATURE_BYTES = 128` (convert.rs) — it previously claimed post-quantum headroom, which is wrong (ML-DSA-65 signatures are ~3,309 B, FIPS 204 range 2,420–4,627 B; 128 fits no PQ scheme). The bound is kept and now documented for Ed25519/classical only: `claimed_log_head` rides gossiped, unverified `PeerAnnouncement`, so a PQ-sized bound would put multi-KB untrusted blobs on the broadcast path — raising it would be a deliberate change. Added a boundary test asserting a real 3,309 B claimed_signature decodes as absent (verified failing with the bound raised, passing at 128). `MAX_CLAIMED_LOG_SIGNATURE_ALGORITHM_BYTES = 32` (the `ml-dsa-65` example string) is untouched — it was correct.

Ran the Linux quality slice locally in Docker on rustc/clippy 1.97.1 — fmt / clippy (`--workspace --all-targets -D warnings`) / quality_contracts all green.

crates/mesh-llm-host-runtime/src/protocol/convert.rs
```diff
@@ -870,6 +916,34 @@ const MAX_REMOTE_MODEL_LIST_LEN: usize = 256;
...
+/// the root as opaque and never verifies it, so the bound is sized generously
+/// enough to admit hash schemes wider than this node's own SHA-256 (e.g.
+/// SHA-512) rather than asserting our own scheme's exact length.
+const MAX_CLAIMED_LOG_ROOT_BYTES: usize = 64;
+
+/// Upper bound on `ClaimedLogHead.claimed_signature`, in bytes. Same
+/// memory-safety rationale as `MAX_CLAIMED_LOG_ROOT_BYTES`: wide enough for
+/// signature schemes larger than this node's own Ed25519 (e.g. post-quantum
+/// signatures), tight enough that a peer cannot advertise unbounded bytes.
+const MAX_CLAIMED_LOG_SIGNATURE_BYTES: usize = 128;
```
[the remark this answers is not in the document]
StevenMih: Yep the underlying point is right even though the two constants got conflated.

The ml-dsa-65 reference at :988 is on MAX_CLAIMED_LOG_SIGNATURE_ALGORITHM_BYTES = 32, where it's an example algorithm identifier string — nine characters, well inside the bound. That one is correct and unchanged.

The real defect is the comment on MAX_CLAIMED_LOG_SIGNATURE_BYTES = 128 at :979-980, which claimed the bound was wide enough for post-quantum signatures. It isn't — ML-DSA-65 is 3,309 bytes and the FIPS 204 range is 2,420–4,627. We deleted the claim and kept the bound, rather than raising it: claimed_log_head rides on PeerAnnouncement, which is gossiped, so admitting a ~4.6 KB signature would put multi-kilobyte untrusted blobs on a broadcast path for a field this node never verifies. The comment now says what 128 bytes is actually for — Ed25519 and classical schemes of similar size — and states that a post-quantum scheme would need a deliberate bound change with its own rationale.

Added a boundary test using a real ML-DSA-65-sized 3,309-byte claimed_signature, asserting it decodes as absent. Verified the test catches the actual defect by temporarily raising the bound to 4,627 — the behaviour the old comment implied — confirming the test failed, then restoring 128. Full protocol::tests::announcements suite green.

StevenMih:
@ndizazzo  Thanks — the three blockers were the right three, and the independent compat check on tag 51 was more than I expected. All three are addressed on the current head; here is where each landed.

1 · The signing contract is now written down. ClaimedLogHead carries string signature_algorithm = 6, matching the SignedMeshGenesisPolicy shape you pointed at, and the proto comment spells out the exact byte string that gets signed. It deliberately follows this crate's existing canonicalization rather than inventing one: u64 little-endian length prefixes as write_string/write_bytes already do, a domain tag in the same style as the existing ones, lengths counted in UTF-8 bytes, and the trimmed signature_algorithm bound into the signed input — so two peers cannot disagree about whether " ed25519" and "ed25519" sign the same bytes. An absent or empty algorithm means the peer named no scheme, and a verifier must not assume one.

2 · The byte fields are bounded. log_id, root, claimed_signature and the new signature_algorithm each have their own named constant and rationale in validate_peer_announcement, rather than sharing the 8 MiB frame budget. While tightening this, a follow-up review caught that the bound's own comment claimed post-quantum headroom it does not have — ML-DSA-65 signatures are 3,309 bytes and the bound is 128. The claim is deleted rather than the bound raised: this field rides a gossiped announcement, so admitting multi-kilobyte untrusted blobs on a broadcast path is the wrong trade for a value this node never verifies. The comment now says what 128 bytes is for, and that a post-quantum scheme would need a deliberate change.

3 · The reasoning is now in the PR body, under "Why a top-level field rather than subprotocols" — including why MeshSubprotocol's shape can't carry it without a wider change, and an honest note that the tag is spent before an in-tree consumer exists. Short version: the head is broadcast so that what a peer tells you on request can be checked against what it has already been telling everyone. I've described the consumer on #1233 rather than expanding this PR.

On the @generated by prost-build header with no build.rs and no drift check — good catch, and not something this PR should fix. Happy to open that as its own issue if you'd like it tracked.

Whenever you have a moment: this needs a re-review to clear the changes-requested, and the five required workflows are still awaiting approval on this head.

StevenMih:
Rebased onto current main — 22998c2 moved the announcement constructors into mesh/announcements.rs, so the claimed_log_head defaults followed them into both constructors there. No other change; still two commits. I saw main had been merged into the branch on the 13th (thanks — that cleared the conflict); the rebase replaces that merge commit with the same resolution so the head stays conventional-commit clean. New head 10aef39c0... Quality and all Rust test batches are green on this head. Linux / Native Linux runtimes / Linux runtime (cpu) is failing, but it's also failing on main at 7eff57a (green at 7144cdf), and #1841 hits it identically — looks like 7eff57a, not this branch. I'll rebase once main is green again if you'd like it re-run.

One correction to my note above: the required workflows on the previous head had started, not "awaiting approval" — my misread.

@ndizazzo — the three items from your review are each addressed in-thread and in the body; whenever you can re-review.

i386 (dismissed):
Reviewed exact head 0b70be2aeb51a674de4bf16d296d0f25dfea8bc0 after synchronization with current main. The host-runtime/protocol suites pass (3,546 unit tests with 11 ignored, plus integration and doc tests), and the full repository validation passes (1,493 tests, 9 skipped). The signed claimed-log-head transport, validation, bounds, and compatibility coverage look sound.

StevenMih:
@ndizazzo hi, please re-review.

StevenMih:
Rebased onto current main (post-#1841), new head 09dd25e. No content change — the same two review rounds' fixes, with the 09-13 merge-main commit replaced by a clean rebase so the history is two conventional commits (the wire field; the signature-bound comment correction from 09-12). @i386, your exact-head approval was dismissed by the push — sorry; the diff against main is what you reviewed. @ndizazzo, this still carries your 09-09 changes-requested: all three items (signing contract, byte bounds, top-level rationale) have been in since 09-12 and each thread is resolved. Could you re-review, or dismiss the old one if you're satisfied? That's the only thing standing between this and merge.

StevenMih:
@ndizazzo hi pls re-review

StevenMih:
@i386 #1709 got closed by a misclick, reopened; it's the last one and only needs Nick's stale request cleared

ndizazzo (dismissed):
Rechecked the final head. The wire contract, bounds, and compatibility coverage look good.

PR #1713, opened 2026-09-09
ci: guard the two planner drift holes with tests

michaelneale:
Two independent gaps let CI report green without running the coverage the plan asked for. Both surfaced while reviewing #1685 and #1659.

## 1. Planned smoke ids vs lane job names

Smoke jobs gate on `contains(fromJson(inputs.smoke_matrix).*.id, '<id>')`. An id in `ci/slices.yml` `smoke_domain_rows` that matches no job condition silently skips the whole domain's smoke coverage, and the lane summary then fails with `ERROR: planned job "product_smoke" finished with "skipped"` — far from the cause.

This is exactly what reddened #1685: #1672 renamed the smoke jobs, `slices.yml` still planned `core` / `two-node-client` / `two-node-split`, and #1693 restored them ~2h later. Anyone merging main in that window got an unexplained red PR.

New test in `scripts/tests/test_ci_lane_workflows.py` asserts every planned smoke id resolves to a job condition, and that `smoke_domain_rows` only references declared `smoke_rows`.

## 2. Windows-only code that never reaches a Windows lane

The Windows lane runs only when `windows_lane_plan.required` is true (`pr_windows.yml:67`), which is `>0 rows` across `hosts` / `runtime_products` / `platform_checks` filtered to `platform == "windows"` (`.github/actions/plan-ci/action.yml:293-308`). Those rows come exclusively from the `platform-windows*` domains (`ci/slices.yml:171,191-192`); every other domain maps to linux-cpu.

`mesh-llm-system` and `skippy-runtime` were claimed only by the `{"domain": "rust", "crates": ["*"]}` wildcard, so #1659 — whose entire purpose is silencing Windows-only `dead_code` warnings — got `Windows = skipping` and a green check.

New test asserts every crate whose sources contain `cfg(windows)` or `target_os = "windows"` is claimed by a windows domain, and `ci/ownership.yml` `platform-windows` `crate_rules` is completed from that census (13 crates).

## Cost

This is not free. A PR touching `skippy-runtime`, `mesh-llm-system`, `mesh-llm-host-runtime` etc. now also plans a Windows host build, `windows-cpu` runtime product, `windows-portable` check and a `core` smoke row. That is the intended trade — those crates carry Windows code and currently have zero Windows compile coverage on PRs — but it is a real runner-minutes decision, so flagging it explicitly rather than burying it.

Worth noting separately: **`-D warnings` never runs against a Windows target on any profile.** `rust_clippy` is a single Linux-container job (`ci-quality-slice.yml:177-233`), and the Windows lane's only cargo invocations are `cargo check --locked -p mesh-llm --bin mesh-llm` and `cargo check -p mesh-llm-nodejs` (`ci-platform-checks-slice.yml:132-141`) — no `--all-targets`, no `-D warnings`. This PR gets Windows *compiled*; it does not get Windows *linted*. That would need a new job on a Windows runner and is deliberately out of scope.

## Verification

At `e60868f68`, base `origin/main` `695b4d089`:

- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` — 945 tests, same 10 pre-existing import errors as `origin/main` (943 tests, missing local `pyyaml`), no new failures
- **Both new tests verified to fail against the state that produced the bug:** test 1 against `ci-linux-product-smoke-slice.yml` at `afa36ab02` (#1672, pre-restore) — reports the three missing ids; test 2 against `ci/ownership.yml` at `origin/main` — reports `skippy-runtime`, `mesh-llm-system`, `mesh-llm-tui`, `mesh-llm-hardware-profile` and the rest
- `test_runtime_change_separates_direct_and_affected_crates` updated: `mesh-llm-host-runtime` now correctly pulls `platform-windows`, `platform-checks` and `windows-cpu` into the plan

Diff is 3 files, +87/-4. No workflow or planner logic changed — only the ownership list and two assertions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows platform coverage so relevant runtime components are included in platform checks and Windows CPU validation.

- **Tests**
  - Added validation to ensure every planned smoke test is assigned to a matching workflow job, preventing tests from being silently skipped.
  - Added checks to confirm Windows-specific Rust components are correctly included in platform ownership rules.

### DIFF
--- a/scripts/tests/test_ci_lane_workflows.py
+++ b/scripts/tests/test_ci_lane_workflows.py
@@ -479,6 +479,31 @@ class CiLaneWorkflowTests(unittest.TestCase):
                 self.assertIn("use_depot:", workflow)
                 self.assertIn("${{ inputs.use_depot }}", workflow)
 
+    def test_every_planned_smoke_id_has_a_matching_lane_job(self) -> None:
+        """Smoke jobs are selected by id, and a missed id skips silently.
+
+        Each smoke job gates on ``contains(fromJson(inputs.smoke_matrix).*.id,
+        '<id>')``. When a planned id matches no job the whole domain's smoke
+        coverage is skipped, and the lane summary fails the run with
+        ``planned job ... finished with "skipped"`` far from the cause.
+        """
+        slices = json.loads((ROOT / "ci" / "slices.yml").read_text())
+        planned = {row["id"] for row in slices["smoke_rows"]}
+        for domain, ids in slices["smoke_domain_rows"].items():
+            with self.subTest(domain=domain):
+                self.assertEqual(set(), set(ids) - planned)
+
+        gated = set()
+        for path in sorted(WORKFLOWS.glob("ci-*-product-smoke-slice.yml")):
+            gated |= set(
+                re.findall(
+                    r"contains\(fromJson\(inputs\.smoke_matrix\)\.\*\.id, '([^']+)'\)",
+                    path.read_text(encoding="utf-8"),
+                )
+            )
+
+        self.assertEqual(set(), planned - gated)
+
     def test_superseded_pr_runs_cancel_by_pull_request_identity(self) -> None:
         for lane in ("quality", "website", "linux", "macos", "windows"):
             pr_workflow = self.workflow(f"pr_{lane}.yml")
--- a/scripts/tests/test_plan_ci.py
+++ b/scripts/tests/test_plan_ci.py
@@ -255,14 +255,18 @@ class PlanCiTests(unittest.TestCase):
             plan["affected_crates"],
             ["mesh-llm-host-runtime", "mesh-llm"],
         )
-        self.assertEqual(plan["domains"], ["rust", "runtime-product"])
+        # mesh-llm-host-runtime holds cfg(windows) code, so it is claimed by
+        # the platform-windows domain and pulls the Windows rows in as well.
+        self.assertEqual(
+            plan["domains"], ["rust", "runtime-product", "platform-windows"]
+        )
         self.assertEqual(
             plan["required_slices"],
-            ["quality", "ui-artifact", "static-abi", "rust-tests", "runtime-product", "product-smoke"],
+            ["quality", "ui-artifact", "static-abi", "rust-tests", "runtime-product", "platform-checks", "product-smoke"],
         )
         self.assertEqual(
             [row["id"] for row in plan["matrices"]["runtime_products"]],
-            ["linux-cpu"],
+            ["linux-cpu", "windows-cpu"],
         )
         self.assertEqual(
             plan["dependencies"]["runtime-product"],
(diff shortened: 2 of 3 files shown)


scripts/tests/test_plan_ci.py
```diff
@@ -582,6 +586,46 @@ def test_manifest_platform_budgets_must_fit_total_worker_ceiling(self) -> None:
...
+        with_windows_code = set()
+        for path in tracked:
+            crate_dir = Path(path).parts[1]
+            if crate_dir in with_windows_code:
+                continue
+            source = (ROOT / path).read_text(encoding="utf-8", errors="ignore")
+            if any(needle in source for needle in needles):
+                with_windows_code.add(crate_dir)
+
+        # Crate directory names match package names across the workspace.
```
ndizazzo, line 626: This one isn't quite true, and I think it's a trap waiting to go off. `crates/mesh-client/` is package `mesh-llm-client`, the single exception in the workspace (it's noted in AGENTS.md).

Harmless today because `mesh-client/src` has no Windows code. But the day someone adds a `cfg(windows)` there, this test will demand that `ownership.yml` claim a crate called `mesh-client`, while `ci-crate-lists` and the planner both key on package names. The two gates become mutually unsatisfiable and neither error message points at the cause.

Could we resolve the directory to its package name through the crate's `Cargo.toml` (or cargo metadata), or failing that just name the exception in the comment so the next person has a thread to pull?

scripts/tests/test_plan_ci.py
```diff
@@ -582,6 +586,46 @@ def test_manifest_platform_budgets_must_fit_total_worker_ceiling(self) -> None:
...
+        }
+        claimed = {
+            crate
+            for rule in ownership["crate_rules"]
+            if rule["domain"] in windows_domains
+            for crate in rule["crates"]
+        }
+
+        tracked = subprocess.check_output(
+            ["git", "ls-files", "--", "crates/*/src/**/*.rs", "crates/*/src/*.rs"],
```
ndizazzo, line 612: The pathspec only reaches `crates/*/src/`, so a `build.rs`, a `tests/` file, or a `benches/` file carrying `cfg(windows)` stays invisible to the census. The docstring says "crate sources", which reads wider than that.

There are zero such files right now (I checked), so this is a gap rather than a defect. `crates/*/**/*.rs` would close it if you want it closed here.

scripts/tests/test_ci_lane_workflows.py
```diff
@@ -479,6 +479,31 @@ def test_manual_depot_input_is_explicitly_forwarded(self) -> None:
...
+        gated = set()
+        for path in sorted(WORKFLOWS.glob("ci-*-product-smoke-slice.yml")):
+            gated |= set(
+                re.findall(
+                    r"contains\(fromJson\(inputs\.smoke_matrix\)\.\*\.id, '([^']+)'\)",
+                    path.read_text(encoding="utf-8"),
+                )
+            )
+
+        self.assertEqual(set(), planned - gated)
```
ndizazzo, line 505: Just noting this is one-directional on purpose: `planned - gated` but not the reverse. That's correct as written, since `qwen-recurrent-gate` and the `product-integration-*` ids are gated without ever appearing in `smoke_rows`. Means a job gated on an id nothing can ever plan would go unnoticed, which is a much cheaper failure than the one you're guarding, so I'd leave it.

ndizazzo (commented):
@michaelneale this is the right direction, and I checked both guards against the exact history that produced their bugs rather than taking them on faith.

Test 1 does catch the real thing: at `afa36ab02` the only gated ids in `ci-linux-product-smoke-slice.yml` were `core-cuda`, `model-download`, the `product-integration-*` set, and `qwen-recurrent-gate`, so `core`, `two-node-client`, and `two-node-split` were missing exactly as you describe. The `ci-*-product-smoke-slice.yml` glob picks up both the linux and macos lanes, and all six `smoke_rows` ids resolve.

Test 2 passes at your head and still passes after merging current main. `mesh-llm-log-store` landed in August with `cfg(windows)` in it, and it's claimed by `platform-windows-storage`, which your `startswith("platform-windows")` check accepts. Good call making that a prefix match instead of an exact domain name.

The ownership expansion isn't nominal either. `cargo tree -p mesh-llm --edges normal` puts all twelve non-nodejs crates in the binary's dep tree, so `cargo check --locked -p mesh-llm --bin mesh-llm` on Windows really does compile their Windows code. And `mesh-llm-nodejs` has its own check. No contract problem with the updated `test_runtime_change_separates_direct_and_affected_crates` expectation either: `_matched_domains` in `scripts/plan-ci.py` only walks `direct_crates`, so direct semantic ownership stays separate from reverse deps the way manage-ci wants.

On my side `just ci-crate-lists` passes, and the full python suite comes back with a failure set that's byte-for-byte identical to origin/main (the failures are my local env missing pyyaml, not yours). `just ci-validate` fails only because it wraps that same suite.

One thing I'd like added to the description before this merges: the cost paragraph reads like a tail case, but `mesh-llm-host-runtime` is the host monolith and it shows up in 68 of the last 200 commits. That's 34%, so claiming it for `platform-windows` makes the Windows lane close to always-on for Rust PRs. I think that's still probably the right trade given what we just paid for both gaps, but we should make that call against the real number rather than an implied minority.

Also agree with leaving Windows linting out of scope. `-D warnings` never running against a Windows target is a bigger separate piece of work, and calling it out is better than half-solving it here.

Three smaller notes inline.

PR #1714, opened 2026-09-09
fix(skippy): keep warm KV reuse fast with disk L3

i386:
Enabling disk L3 for a dense model changed the in-process payload from native resident KV to serialized exact state. Repeated warm requests then imported snapshots instead of taking the resident prefix path: the reported replay regressed from 23 ms without L3 to 453 ms with L3.

This keeps resident KV as the dense in-process fast path and gives L3 a separate exportable exact-state payload. Exact restore remains available after restart, while an existing resident hit wins on warm requests.

The pre-restack release replay on the patch-equivalent commit `99e4db903` used identical binary SHA-256 `86d2853dbde9316c67bbe19a006ac62edcee550a7b74387330b00cc683707e28`, model, and workload:

| cohort | control | L3 enabled |
|---|---:|---:|
| fill p50 | 69 ms, 60.4% cached | 86 ms, 60.4% cached |
| restore p50 | 397 ms, 0% cached | 386 ms, 25.6% cached |
| warm p50 | 13 ms, 99.95% cached | 18 ms, 99.95% cached |

Both arms completed with zero failed requests. This removes the warm-path regression. Durable restore is still near parity on this small-model workload, so the 2x restart target remains optimization work.

## Stack

- Base: #1710, following #1707 and #1632 on `kvcache-ng`
- This PR: resident fast-path repair
- Next: #1726 packed durability

## Validation

- `cargo test -p skippy-server --lib` — 705 passed, 3 ignored
- patch-equivalent release host replay with zero failed requests

### DIFF
--- a/crates/skippy-server/src/kv_integration/config.rs
+++ b/crates/skippy-server/src/kv_integration/config.rs
@@ -85,7 +85,7 @@ impl KvStageIntegration {
             emit_cache_disabled_warning(config, reason);
             return Ok(None);
         }
-        let mut payload = effective_cache_payload(cache_config.payload, &model_capability);
+        let payload = effective_cache_payload(cache_config.payload, &model_capability);
         if payload == StagePrefixCachePayload::Disabled {
             return Ok(None);
         }
@@ -109,18 +109,24 @@ impl KvStageIntegration {
             return Ok(None);
         }
         let l3_manager = manager()?;
-        if l3_manager.is_some()
-            && payload == StagePrefixCachePayload::ResidentKv
-            && dense_without_recurrent
-        {
-            // The durable tier needs exportable state and ResidentKv is
-            // borrow-only. Dense families record KV pages with an empty
-            // recurrent snapshot so they reach disk; the model is known
-            // dense, so the empty snapshot is expected rather than corrupt.
-            payload = StagePrefixCachePayload::KvRecurrent;
-        }
+        let durable_payload = l3_manager.as_ref().map(|_| {
+            if payload == StagePrefixCachePayload::ResidentKv && dense_without_recurrent {
+                // Resident KV stays the in-process fast path. Dense families
+                // export KV pages with an empty recurrent snapshot for L3;
+                // the known-dense capability makes that empty component valid.
+                StagePrefixCachePayload::KvRecurrent
+            } else {
+                payload
+            }
+        });
         let l3 = l3_manager
-            .map(|manager| l3_tier_for_manager(config, payload, manager))
+            .map(|manager| {
+                l3_tier_for_manager(
+                    config,
+                    durable_payload.expect("enabled cache has a durable payload"),
+                    manager,
+                )
+            })
             .transpose()?;
         // FullState is architecture-neutral: the native runtime serializes the
         // complete session state for both dense and recurrent model families.
@@ -206,6 +212,7 @@ impl KvStageIntegration {
         Ok(Some(Self {
             mode,
             payload,
+            durable_payload,
             correctness_mode: false,
             trust_local_writes: true,
             checkpoint_policy,
@@ -1315,7 +1322,7 @@ mod tests {
     }
 
     #[test]
-    fn dense_auto_cache_becomes_exportable_when_disk_is_injected() {
+    fn dense_disk_cache_preserves_resident_fast_path_and_exports_exact_state() {
         let root = std::env::temp_dir()
             .join("skippy-server-l3-manager-tests")
             .join(format!("dense-export-{}", std::process::id()));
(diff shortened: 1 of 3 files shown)


ndizazzo (changes requested):
The decomposition here is right. On the base branch, enabling disk L3 for a dense model rewrote `payload` from `ResidentKv` to `KvRecurrent`, which silently flipped every in-process decision site that switches on `payload` — restore, affinity, prompting, suffix prefill — onto the serialized exact-state path. Splitting the live in-process representation (`payload`) from the exportable one (`durable_payload`) is the correct fix, and `l3_payload_kind` mapping `ResidentKv` to `"unsupported"` confirms the tier genuinely needs the separate value.

Requesting changes on one correctness gap, plus a couple of smaller items.

## The restore short-circuit prefers any resident prefix over a longer exact snapshot

```rust
if self.payload == ResidentKv
    && identities.iter().any(|identity| self.probe_resident_prefix(identity).is_some())
{ return Ok(None); }
```

`probe_resident_prefix` returns `Some` for a **partial** match — anything at or above `shared_prefix_min_tokens`. `restore_binary_prefix` tries exact first and falls back to resident, so returning `Ok(None)` here hands the request to a resident restore that may cover far fewer tokens than the exact snapshot would have.

Concretely: a 4,000-token identity with 200 tokens resident and the full 4,000 in exact/L3. Before this change the stage restored 4,000; now it restores 200 and re-prefills 3,800.

The replay can't see this. The warm cohort is 99.95% cached — full-prefix hits, where the guard is unambiguously right. The regression lives in the partial-overlap middle, which is the cohort L3 exists for; the restore arm shows only 25.6% cached.

Suggested fix: gate on coverage rather than existence. `binary_kv.rs:176` already models this distinction — a probe only counts as a hit when `probed.token_count >= token_ids.len()`, otherwise it is `probe_short`. Mirroring that, or comparing against the exact radix peek length and taking the larger, keeps the fast path without giving up long restores.

## `expect` on an unstated invariant

```rust
l3_manager.map(|manager| l3_tier_for_manager(config, durable_payload.expect("enabled cache has a durable payload"), manager))
```

`durable_payload` is `Some` iff `l3_manager` is `Some` — both derived from the same `Option` two lines apart, so the panic is unreachable. It is also structurally avoidable: `l3_manager.zip(durable_payload).map(...)` removes it.

Related, in `exact_state_payload()`:

```rust
self.payload.is_exact_state().then_some(self.payload).or(self.durable_payload)
```

This can in principle return a non-exact payload. Today it cannot — `ModelKvCapability` has three variants, `Unknown` disables the cache, `KnownRecurrent` + `ResidentKv` is rejected, so `payload == ResidentKv` implies `dense_without_recurrent` and `durable_payload` is always `KvRecurrent`. But that chain is implicit and lives in a different file. If it ever breaks, `payload_is_exact_state()` starts returning `true` for a `ResidentKv` durable payload and gates roughly ten call sites into exact paths that silently no-op. A `.filter(StagePrefixCachePayload::is_exact_state)` on the `or` arm makes the helper self-defending for free.

## Test coverage does not reach the behavior change

The only test touched is the config-wiring assertion (`payload == ResidentKv`, `durable_payload == Some(KvRecurrent)`, `l3.is_some()`). The two things that actually changed at runtime — the restore short-circuit, and `record_exact_state` exporting via `exact_state_payload` while resident recording stays live — have no test. The partial-prefix case above is exactly what a unit test would catch. At minimum: one test that a full resident hit skips exact restore, and one that a short or absent resident prefix still takes it.

## Non-blocking observations

- **Double recording.** With `payload` back to `ResidentKv`, dense + L3 now records resident KV *and* exports serialized state on every qualifying prefill. That is the design intent, and the fill p50 delta in your own numbers (69 → 86 ms) is consistent with it. Worth naming explicitly in the PR body as an accepted cost rather than leaving it as an unexplained regression.
- **Affinity ignores L3.** `peek_cache_affinity` switches on `self.payload`, so dense + L3 now peeks only the resident radix. After a restart the resident tier is empty while L3 still holds restorable state, so that stage reports zero affinity and the scheduler will not steer to it. This matches non-L3 behavior, so it is arguably correct by restoration, but it means L3 buys nothing for routing. Possible follow-up for #1712.
- **Telemetry.** `skippy.kv.payload` now reports `ResidentKv` for dense + L3 stages while the durable tier stores `KvRecurrent`. Emitting `durable_payload` alongside it would make the two-tier state legible in the field.
- **Branch is behind its base.** The head predates current `skippy-l3-disk-store`; rebase before merging so the replay numbers correspond to the merged tree.

i386:
Addressed the blocking review points and restacked the PR in e43f42eab:

- a resident prefix only bypasses durable restore when its matched token count covers the full requested identity, so a 200-token resident hit no longer masks a 4,000-token L3 checkpoint;
- durable payload selection now filters out non-exact ResidentKv values;
- L3 construction pairs the manager and payload with zip instead of an expect;
- regression coverage includes shorter/equal/longer resident-prefix decisions and invalid durable payload filtering.

Validation on exact head e43f42eab8f8e9e5047b66ca41ea2673c553eb47: cargo fmt check passed; the complete skippy-server library suite passed (706 passed, 3 ignored); warning-denying Clippy passed; diff-check passed.

i386:
@ndizazzo The blocking points from your review are present on the current head `e43f42eab8f8e9e5047b66ca41ea2673c553eb47`:

- an exact restore is skipped only when resident coverage reaches the full requested identity;
- manager/payload construction uses `zip`;
- `exact_state_payload()` filters the durable payload to exact-state variants;
- tests cover shorter/equal/longer resident coverage and invalid durable-payload filtering.

The branch is fully stacked on the current #1710 head, all latest CI checks are green, and the accepted double-recording cost is documented in the implementation and prior evidence. Please re-review the current head when you can.

i386:
Consolidated into #1736: https://github.com/Mesh-LLM/mesh-llm/pull/1736

The final combined tree retains every code, test, and documentation change from this PR. Review continues on #1736.

PR #1717, opened 2026-09-09
ci: simplify agentic-replay workflow to essentials

i386:
## Summary

Simplifies `.github/workflows/agentic-replay-nightly.yml` while preserving its safety and regression-gating invariants. Scheduled and manual runs execute trusted `main` only on the persistent `micstudio` runner. The workflow validates and downloads pinned inputs, runs the matrix-defined replay workload, gates against public history, publishes successful results, and opens a repair PR on regression.

The review fixes in this branch:

- propagate the pre-warmed Hugging Face cache through `GITHUB_ENV`
- validate immutable model and dataset pins before download and cross-check the canonical trajectory pin
- fail closed when history lookup fails for reasons other than a missing dataset
- derive every replay-shape field from the matrix and preserve existing cohort metadata
- always generate the summary card before returning a regression-gate failure
- confine `HF_TOKEN` to the publish step
- keep publication credentials off the persistent runner; emit a validated patch/body artifact and publish it from a clean hosted job with run/attempt-specific branches
- replace micstudio-incompatible process substitution with cleaned-up temporary files

Repository-variable feature flags and the redundant toolchain checklist are removed. The regression gate remains always enabled.

## Validation

- `just ci-shellcheck scripts/agentic-replay-repair.sh`
- `actionlint .github/workflows/agentic-replay-nightly.yml`
- 87 focused `test_agentic_replay*.py` tests
- `just ci-validate`: 1,070 tests passed, 8 skipped; actionlint and all repository consistency checks passed

## Summary by CodeRabbit

- **CI Improvements**
  - Scheduled and manual Agentic Replay runs are restricted to the trusted `main` branch.
  - Replay modes, output limits, concurrency, and other settings are validated from the configured matrix.
  - Regression summaries are retained even when checks fail.
  - Repair publication is separated from repair execution and limited to validated failures on `main`.
  - Repair artifacts and credentials are handled through a controlled publication process.

- **Bug Fixes**
  - Improved replay repair reliability with safer temporary-file handling.
  - Repair reruns now preserve all configured replay parameters.


ndizazzo (changes requested):
Reviewed against the diff, the branch tree, and `.agents/skills/manage-ci/SKILL.md`. The simplification is directionally right — the state machine and the repo-variable knobs were genuinely more machinery than this workflow needed. Three things in the current diff look like they change behavior in ways the PR description doesn't claim.

## Blocking

### 1. The HF cache normalization is inert

`.github/workflows/agentic-replay-nightly.yml` (meta step) writes `HF_HOME` / `HF_HUB_CACHE` to `$GITHUB_OUTPUT` rather than `$GITHUB_ENV`. Step outputs aren't environment variables, and nothing consumes `steps.meta.outputs.HF_HOME` — grepping the branch, the only references are the `echo` lines themselves.

So `hf download` and `evals/agentic-replay.py` both miss the pre-warmed cache; the latter falls back to `$HOME/.cache/huggingface` when `HF_HOME` is unset (`evals/agentic-replay.py:1254`). With `HF_HUB_OFFLINE=1` also removed, this doesn't fail — it silently re-downloads the pinned GGUFs into the wrong cache on every run.

`llama-upstream-canary.yml:54-77` does the same normalization correctly, via `$GITHUB_ENV`. Worth noting the local validation checked that the step *emits the expected outputs*, which is exactly why this passed: the outputs are correct and unused.

### 2. Branch-controlled code now runs on the persistent self-hosted runner with a write-scoped token

The deleted comment stated the invariant directly: "Scheduled runs execute the default-branch definition only; manual dispatch is accepted only from main so caller-selected content never reaches micstudio."

With the `if:` reduced to a repository check, `ref: ${{ steps.meta.outputs.replay_ref }}` checks out arbitrary branch content and executes it (`evals/agentic-replay.py`, cargo, opencode) with `secrets.HF_TOKEN` in step env on every branch, not just main. The "branch runs don't write the dataset" guarantee is a shell conditional at the end of the step, not credential scoping — the token that can `hf upload` is already in the environment of branch-controlled Python by then.

`manage-ci` is explicit here: "Never route untrusted code to a persistent self-hosted runner." If branch dispatch is worth keeping, the token needs to be out of the non-main path — a read-only token for branch runs, or publishing moved to a separate main-only step that owns `HF_TOKEN` alone.

Separately: the shared `concurrency: agentic-replay-nightly` group with `cancel-in-progress: false` and `timeout-minutes: 720` means one branch dispatch can block the nightly schedule for up to 12 hours.

### 3. The bootstrap probe treats auth and network failure as "no history"

The previous code called `repo_info` and distinguished a genuine 404 (exit 3) from auth/permission/network failure (exit 1). The replacement falls back to `hf download card.md` and treats any probe failure as bootstrap.

An expired `HF_TOKEN` or an HF outage fails both calls, so there's no baseline, `--gate` finds nothing to compare against, and the run passes and publishes a shard. That's the inverse of the comment two lines above it ("anything else fails loudly"), and it makes the gate fail open on the most likely infrastructure failure. The 404-vs-everything-else distinction needs to survive the simplification.

## Should fix

### 4. Concurrency levels now have two sources of truth

The workflow hardcodes `--concurrency 1 --concurrency 2 --concurrency 4 --concurrency 8`, but `scripts/agentic-replay-repair.sh:47` still derives `LEVELS` from `matrix.json`, and `--replay` still feeds `matrix.json`'s replay block into the recorded history rows.

Divergence won't fail loudly: `agentic-replay-history.py` keys baselines on `model|c<concurrency>` and requires `r["replay"] == row["replay"]`, so drifted cells find no baseline and get silently skipped by the gate. Reading the levels from the matrix was one line.

### 5. The dataset-pin cross-check has no replacement

Format validation is actually fine — `validate_model_pin` / `validate_replay_pin` in `agentic-replay-history.py` still enforce 40-hex revisions and 64-hex digests, so the description understates what's preserved.

But the cross-check against `evals/skippy-competitive-benchmark.json` has no equivalent anywhere on the branch, so the nightly matrix can now drift off the canonical harness pin and still pass. And moving format validation from pre-download to post-benchmark means a malformed pin burns hours of a 720-minute job before it's caught.

## Minor

- `mkdir -p .../summary` moved into "Download and verify pinned inputs" while its consumer is "Record hardware fingerprint" — implicit cross-step coupling that breaks on reorder.
- `model["revision"]` / `model["sha256"]` direct indexing replaces `.get()` with named errors; a malformed matrix now surfaces a bare `KeyError` traceback.
- `DATASET_REPO` is redeclared in the repair step's `env` despite the job-level `env.DATASET_REPO`.
- `${{ steps.meta.outputs.label }}` is interpolated into the replay step's shell while the next step passes the same value via `env: LABEL`. The `tr -C 'a-zA-Z0-9._-'` sanitization makes it safe, but the two forms could collapse into the `env` one.

## Checked, no issue

- `[[ -n "$X" ]] && args+=(...)` under `set -euo pipefail` — a false test mid-script doesn't exit the shell, so the optional-arg pattern is fine.
- `agentic-replay-card.py:34` guards `--history` with `is_dir()`, so bootstrap and branch runs won't fail on the missing baseline dir now that card generation is unconditional.
- Process-substitution removal is complete in both files.
- Always-on `--gate` is safe: the history script bootstraps for three runs before it can fail.

I didn't re-run `actionlint` or the two test modules; those claims look consistent with the diff.

i386:
Review fixes pushed in a749ed4b5 — thanks @ndizazzo, all three blockers plus both should-fixes and the minors are addressed (CodeRabbit's overlapping findings too):

**Blocking**

1. **HF cache normalization inert** — `HF_HOME`/`HF_HUB_CACHE` now go to `$GITHUB_ENV` (outputs kept only for the values actually consumed via `steps.meta.outputs.*`). The download and replay steps now inherit the pre-warmed cache.

2. **Write-scoped token on branch-controlled code** — `HF_TOKEN` is now confined to a new main-only step, `Publish run shard and card (main only)` (`is_main` moved from shell conditional into the step `if:`). The gate step and the input-download step run **token-free**: the dataset repo and all three pinned model repos are public, so anonymous `hf download` suffices (verified each repo returns 200 unauthenticated). The download step also lost the `HF_TOKEN` env this PR had added over the base. Branch-controlled code no longer touches any HF credential.

3. **Probe fails open** — the base workflow's `repo_info` probe is ported verbatim (byte-identical python block): 404/"Entry Not Found"/"Repository Not Found" → bootstrap; anything else (auth, rate limit, network) fails loudly with both logs dumped. The `hf download card.md` probe is gone.

**Should fix**

4. **Two sources of truth for concurrency** — the replay step now derives levels, passes, `trajectories-per-framework`, and warmup-turns from `matrix.json`, matching the repair re-run; the hardcoded `--concurrency 1..8`/`8`/`2`/`4` literals are gone.

5. **Pin cross-check restored, pre-download** — the matrix dataset pin is cross-checked against `evals/skippy-competitive-benchmark.json`, and all pin/shape validation (model + dataset + replay-shape fields, named `SystemExit` errors instead of `KeyError` tracebacks) runs before any download.

**Minors + CodeRabbit**

- `concurrency:` group is per-ref for branch dispatches (`agentic-replay-nightly-branch-<ref>`), so a branch run can't queue-block the nightly; scheduled/main runs stay serialized in the shared `default` group. Residual, noted deliberately: this workflow is the only consumer of micstudio, so a branch run still physically occupies the runner until it finishes — it can delay, but no longer *queue-block*, the nightly.
- Gate step now runs the history script without `set -e` aborting, always writes the card + `$GITHUB_STEP_SUMMARY`, then returns the gate status (CodeRabbit's "capture exit status" finding) — regressions no longer swallow the numbers view, and the repair PR still fires via `if: failure()`.
- `mkdir -p …/artifacts/summary` moved into the hardware-fingerprint step that owns the artifacts root.
- `DATASET_REPO` redeclaration in the repair step env removed; `--ref` label interpolation collapsed into `env: LABEL`.
- Repair script traps `rm -f "$REPLAY_PARAMS_FILE"` on EXIT (CodeRabbit nitpick).
- `manage-ci` inventory + `ci/ci.md` rows updated to the new shape (dispatch policy, token confinement, per-ref concurrency) per the skill's same-change rule.

**Validation:** `actionlint` (repo config) clean; `shellcheck` clean on both touched shell surfaces; `git diff --check` clean; `scripts/tests/test_agentic_replay*.py` 83/83 pass. Functionally exercised the meta step (branch/main/wrong-runner), the gate step end-to-end against the real `agentic-replay-history.py`/`card.py` with a stubbed `hf`/`huggingface_hub` (dataset-ok, bootstrap-404, outage fail-loud, branch no-publish, and regression → regressions reported + card still written + rc=1), and the new pin validation (happy, canonical drift, malformed pin, missing shape fields).

.github/workflows/agentic-replay-nightly.yml
```diff
@@ -156,44 +133,79 @@ jobs:
...
-            --repo-type dataset --local-dir "$HISTORY_LOCAL" --exclude "*.md" 2> "$RUNNER_TEMP/hf-download.err"; then
-            exit 0
-          fi
-          # Only a genuinely missing dataset is a bootstrap run; auth,
-          # permission, and network failures must fail loudly.
-          if HF_HUB_OFFLINE=0 python3 - "$DATASET_REPO" <<'PY' > "$RUNNER_TEMP/repo-exists.out" 2>&1; then
+          if ! hf download "$DATASET_REPO" \
+            --repo-type dataset --local-dir "$HISTORY_LOCAL" --exclude "*.md" \
+            > "$RUNNER_TEMP/hf-download.log" 2>&1; then
+            if python3 - "$DATASET_REPO" <<'PY' > "$RUNNER_TEMP/hf-probe.log" 2>&1; then
```
ndizazzo: [P1] Keep the explicit online override for history and publication

We need to preserve `HF_HUB_OFFLINE=0` on the history download/probe and the upload commands. Removing the job-level setting doesn't clear the runner's inherited environment. Today's [micstudio canary job](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34699953619/job/103569834618) passed the toolchain preflight that explicitly requires ambient `HF_HUB_OFFLINE=1`, matching the checked-in runner inventory. With that environment, this download fails offline and the probe fails too, so every nightly stops before producing the history/card. I reproduced the download failure locally with the real `hf` CLI and `HF_HUB_OFFLINE=1`. The base workflow explicitly opted these network operations into online mode; let's retain that scoped override while keeping the pre-warmed model cache read-only.

.github/workflows/agentic-replay-nightly.yml
```diff
@@ -1,110 +1,98 @@
...
+          # download and replay steps actually inherit them.
+          {
+            echo "HF_HOME=${HF_CACHE:?HF_CACHE is not exported on this runner}"
+            echo "HF_HUB_CACHE=${HF_CACHE}/hub"
+          } >> "$GITHUB_ENV"
 
       - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
         with:
-          ref: main
+          ref: ${{ github.sha }}
```
ndizazzo, line 50: [P2] Bind the repair comparison to the pinned source SHA too

I agree with checking out the immutable SHA, but we need to update its repair consumer at the same time. `scripts/agentic-replay-repair.sh` still invokes `--ref base=origin/main`. With a bare SHA input, the pinned checkout action fetches only that commit with no destination ref, so it doesn't create or refresh `origin/main`. A fresh checkout fails `git rev-parse origin/main^{commit}` before the repair benchmark starts; a persistent checkout can silently compare against a leftover main revision. I reproduced the missing-ref case with the same depth-one SHA fetch. Let's pass the original `BASE_SHA` already captured before the repair as the baseline, and use that same immutable reference in the agent's instructions.

ndizazzo (changes requested):
I like the direction here. Pinning execution to the workflow SHA, deriving the replay settings in one place, preserving the summary after a failed gate, and moving GitHub publication onto a clean hosted runner are all useful improvements. I wouldn't merge this version yet because the online history path still inherits micstudio's offline mode, and the repair comparison still depends on an `origin/main` ref that the new checkout no longer refreshes.

I reviewed the full diff at `091562ce3811e5dbcdfaf87776274a309d47036e` against base `2c3d8bc6662681e42931006277663ac523da5f0d`, including the replay/history consumers and previous review comments. The earlier cache-export, branch-token, history-probe, and summary-card findings have been addressed.

All five required PR lanes passed for this head. The [Quality run](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34716200366) includes actionlint and 1,217 passing Python tests. Locally, all 91 focused replay tests passed with Python 3.13, as did `just ci-shellcheck scripts/agentic-replay-repair.sh` and the diff whitespace check. I also reproduced the offline history-download failure and the missing `origin/main` ref after a fresh SHA-only checkout. Those execution paths aren't covered by the current passing contract tests.

The [latest scheduled replay](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34707662074) was skipped with no steps, as were the preceding three schedules. The [last executed manual replay](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34320960773) failed first on an offline model-cache miss, then hit the old process-substitution problem. That is older-main evidence, not a run of this PR. I haven't dispatched the new nightly or exercised live model replay, HF publication, or hosted repair publication.

For cleanup, I'd move the large inline publication validator into a small script invoked from the trusted checkout before applying the patch, and add executable tests for valid and malformed artifacts, an empty repair, and a fresh SHA-only checkout. The existing repair tests mostly assert that strings appear in the workflow, so they don't prove the producer and consumer work together. I'd also bind `status.run_id` to `github.run_id`, as we already do for the attempt and source SHA, rather than accepting any numeric run ID from the artifact.

.github/workflows/agentic-replay-nightly.yml
i386: Fixed in f0ee8e95555cb0871c1014f1a508cff856e7872d. History download, the repository probe, and both uploads now set HF_HUB_OFFLINE=0 on the individual network commands, while the pre-warmed model and trajectory cache stays read-only by default. The full just ci-validate suite passed (1,215 tests; 9 skipped), along with the focused repair contract tests and shellcheck.

.github/workflows/agentic-replay-nightly.yml
i386: Fixed in f0ee8e95555cb0871c1014f1a508cff856e7872d. The repair wrapper captures BASE_SHA before agent changes, names that immutable revision in the agent task, and passes --ref "base=$BASE_SHA" to every replay comparison. A contract test rejects any remaining origin/main dependency. The full just ci-validate suite passed (1,215 tests; 9 skipped).

michaelneale (dismissed):

ndizazzo (dismissed):

ndizazzo (dismissed):

PR #1718, opened 2026-09-09
fix(package): bound layer artifacts and publish replacements atomically

i386:
DeepSeek-V4-Pro packaging was deterministically evicted because a 12.9 GiB layer was emitted through the split-shard merge path, where source parts and the merged output coexist under the 50 GiB HF Jobs ephemeral-storage limit. This change bounds ordinary layer parts to 8 GiB and sends every oversized layer, including a layer containing one indivisible oversized tensor, through a direct streaming writer with no merge staging.

Replacement publishing is also safe and repeatable. `--republish` stages the complete package on a branch and promotes the manifest-listed snapshot to `main` in one parent-guarded commit. A failed upload leaves the published snapshot unchanged. Fresh queue attempts suppress duplicate paid jobs, while success and failure markers terminate that attempt immediately instead of extending the cooldown from repository modification time.

Additional package fixes in this PR:

- Stream verified artifacts through upload hooks while retaining compact header stubs for metadata-carrier construction.
- Preserve the generic hook contract: hooks may retain or delete their verified artifact.
- Copy GGUF metadata key/value entries as raw wire bytes so typed keys such as `general.alignment` remain valid.
- Omit zero-element GGUF placeholders throughout native inspection and package creation.
- Accept only numeric `layer-N-partMM.gguf` suffixes when deriving layer ordinals.
- Add repeatable `--exclude-repo` quarantine selection for queue runs.

### Skippy ABI inventory

| Status | Symbol/declaration | Public header | Implementation / mirror | Reason | Lockstep update |
|---|---|---|---|---|---|
| Added | `enum skippy_status skippy_write_gguf_from_parts_consuming(const char * const * input_paths, size_t input_count, const char * output_path, struct skippy_error ** out_error)` | `include/skippy/model_package.h` | `src/skippy/model_package.cpp`; `crates/skippy-ffi/src/static_bindings.rs`; `crates/skippy-ffi/src/dynamic.rs`; caller in `crates/skippy-runtime/src/gguf_writer.rs` | Materialize a GGUF while deleting disposable source parts after their final selected tensor is copied, which bounds transient storage. Unlink failure is reported as an I/O error. | Native and Rust ABI move from 0.1.54 to 0.1.55 together. Older native runtimes are intentionally unsupported. |
| Removed | None | — | — | No public declarations were removed. | — |

The public header documents input ownership, deletion timing, and partial-failure behavior. The generated Skippy API reference was regenerated from the prepared queue.

### Validation

- Clean 35-patch llama.cpp queue application and CPU static build.
- 47/47 native tests, including byte-identical consuming/default writer behavior and input-retention/deletion checks.
- Every installed Skippy public header compiles independently as C11 and C++17.
- `cargo test -p model-package`: 25 passed.
- `cargo test -p skippy-model-package`: 92 passed, including budget split, direct singleton writer, zero-element source, upload-hook, carrier, and independent verifier coverage.
- `cargo test -p skippy-runtime --lib`: 159 passed, 1 ignored helper.
- `cargo test -p skippy-ffi --lib`: 11 passed.
- Clippy with `-D warnings` for model-package, skippy-model-package, skippy-runtime, skippy-ffi, mesh-llm-host-runtime, and mesh-llm.
- Repository CI validation: 1,441 tests passed, 9 skipped; crate-list, release-target, publish-chain, console-print, formatting, shell syntax, Python promotion tests, and generated API checks passed.

The production DeepSeek-V4-Pro HF Job rerun remains the operational verification after merge.


ndizazzo (changes requested):
**Verdict: request changes.** The feature is well-scoped and the intent is right, but as written `--republish` is not idempotent, and the failure mode is duplicate *paid* HF Job submissions.

## Blocking: `--republish` re-submits jobs that are already in flight

The PR body says "Recently-queued (in-flight) jobs still skip to avoid duplicate submissions." That is not true for any candidate `--republish` actually targets.

`candidate_status()` short-circuits in this order:

1. catalog hit → returns `Cataloged` (before touching the repo at all)
2. sibling `model-package.json` → returns `Published`
3. `automation/failure.json` → returns `Failed`
4. `automation/queue.json` within `retry_queued_after` → returns `Queued`

The queue-marker check at step 4 is only reachable when the candidate is *not* already cataloged, published, or failed — i.e. exactly the three statuses the new gate bypasses. So `Queued` can never win against `--republish`.

Concretely, on the v2 slate this PR exists to ship:

```
run 1 --republish --confirm  → writes automation/queue.json, submits 20 jobs
run 2 --republish --confirm  → still sees Cataloged/Published (marker never consulted)
                             → submits 20 more jobs for the same models
```

That happens on a CI re-run, a retry after a partial failure, or just running it twice. Since `write_queue_marker()` uploads the marker *before* `jobs_client.submit()`, even a crash mid-run leaves the marker in place with no protection. `total_max_cost_usd` is real money.

The fix is to make the republish bypass yield to a fresh queue marker. Cleanest is to check the marker in `candidate_status()` before the catalog/published/failure returns (or return the queue state alongside them), so `QueueStatus::Queued` is authoritative regardless of what else the repo contains. CodeRabbit flagged this for `Failed` only — it's the same bug on all three arms.

## Style + testability: the empty-then-branch

```rust
if args.republish && matches!(status, Published{..} | Cataloged{..} | Failed{..}) {
    // Replacement policy: republish v2 over the existing package in place.
} else {
    match status { ... }
}
```

An `if` whose then-branch is a comment reads as a mistake. Two better shapes, and the second also solves the coverage gap:

- Match guards: `QueueStatus::Published { repo } if !args.republish => { println!(...); continue; }`
- Extract the policy: `fn should_skip(status: &QueueStatus, republish: bool) -> Option<String>` returning the skip message. Pure function, trivially unit-testable, and `main()` becomes `if let Some(msg) = should_skip(...) { println!("{msg}"); continue; }`.

Right now the decision is inline in `main()`, which is why it can't be tested.

## Missing tests

The only test-file changes are two `republish: false` struct-literal fixups. There is no test that `--republish` parses, none that a published/cataloged/failed candidate is re-queued, and none for the double-submit regression above. With `should_skip` extracted, all three are a handful of lines.

## Minor

`automation/failure.json` is never deleted on a successful republish, so a previously-failed repo ends up carrying both `model-package.json` and a stale failure marker. Harmless today (the `Published` check precedes the failure check) but it's misleading metadata for anyone reading the repo or a future status check that reorders.

## Non-issues

- The allowlist regen is correct — 31 → 31 entries, pure line-number shift, all under the queue-binary path key. **CodeRabbit's outside-diff comment claiming the entries landed under `crates/mesh-llm/src/commands/models/mod.rs` is a false positive**; don't chase it.
- Re-cataloguing is safe: `catalog_entry_path()` is derived from the source repo, so it overwrites in place rather than duplicating.
- Help text is updated, CI is green, and no docs or workflows reference this binary's flags.

i386:
Stacked two Batch-1 root-cause fixes on this branch (tip `c82cb349b`):

**1. Descriptor-only GGUF shards (`skippy-model`)** — `read_gguf_catalog` rejected any artifact whose aligned `data_start` lands past EOF. Split models legitimately ship an `n_t=0` first shard holding only metadata (unsloth/inkling-GGUF shard 1: table ends 12,986,167 B, 32-byte alignment rounds to 12,986,176 > EOF — job 6aa1153f died on this in 2.4s). Kimi-K3's identical-shaped shard passes only because its table end is coincidentally 32-aligned. Descriptor-only shards now require only that the metadata table fits; payload-bearing artifacts keep every original bounds check. Verified against the real downloaded shard bytes with the actual reader; regression test included.

**2. Package workspace on /bucket (`split-model-job.sh`)` — PACKAGE_DIR defaulted under /tmp and jobs died at the HF Jobs 50G per-container ephemeral-storage eviction (Kimi 6aa114dd, V4-Flash 6aa114e8; node overlay had 1.3T free — it is a kubelet cap, not disk size, and no flavor lifts it since ephemeralDisk is null across the catalog). The package must stay fully materialized until the final metadata-carrier pass, so PACKAGE_DIR now defaults to the bucket workspace and the job hard-fails if the package workspace sits on the container root FS.

Local: fmt, clippy -D warnings --all-targets, both crate test suites, no-console-print regen check — all green. Retriggering Kimi-K3, V4-Flash and Inkling from this branch tip (full SHA as --mesh-llm-ref).

crates/model-package/src/bin/queue-unsloth-layer-packages.rs
```diff
@@ -824,31 +834,84 @@ async fn candidate_status(
...
+        QueueStatus::Cataloged { repo } if !republish => Some(format!(
+            "already has layer package {repo} in meshllm/catalog"
+        )),
+        QueueStatus::Failed { repo } if !republish => {
+            Some(format!("layer package job previously failed for {repo}"))
+        }
+        // Replacement policy: republish re-queues finished packages in place.
+        QueueStatus::Published { .. }
+        | QueueStatus::Cataloged { .. }
+        | QueueStatus::Failed { .. } => None,
```
ndizazzo, line 999: [P1] Publish replacements as a complete snapshot

We should stage a replacement off the published branch and promote it only after its manifest and artifacts are complete. This new bypass sends an already published repo through `split-model-job.sh`, whose hook commits each fixed artifact path directly to the default branch, while the replacement `model-package.json` is uploaded only afterward. For a v2-to-v2 republish with changed source bytes or packing, readers resolving the repo during that interval get the old manifest and new artifact bytes, so size/hash or carrier validation fails. If a later upload fails, that broken published state remains. `CatalogPackage` stores only the repo, not a target revision, so catalog users aren't protected by a previous snapshot pin. Publishing the manifest last is sufficient for a new repo, but replacement needs an atomic final commit/ref promotion or versioned artifact paths. Please test a download during replacement and a failure before promotion.

crates/skippy-model-package/src/package_v2.rs
```diff
@@ -73,6 +83,21 @@ pub(crate) fn write_package(
...
+        write_header_stub(&path, &header, artifact.byte_size)?;
+        header_stubs.push(header);
+        run_artifact_hook(&artifact_hook, &path, &artifact.path)?;
+        verify_hook_result(&artifact, &path, &artifact_hook)?;
+        // The upload hook publishes each verified part immediately and deletes
+        // its local copy. If a hook left the part in place we still delete it:
+        // keeping it would re-accumulate the full package in the workspace and
+        // re-trigger the HF Jobs 50G ephemeral-storage eviction.
+        if artifact_hook.command.is_some() && path.exists() {
+            fs::remove_file(&path)
```
ndizazzo: [P2] Let the after-artifact hook decide whether to delete its copy

We should remove this unconditional cleanup or make consumption an explicit option. `--after-artifact-command` is a generic post-verification callback, and the README promises that a hook may delete its copy or leave it for rechecking. A successful logging/verification hook such as `/usr/bin/true` now causes every payload to be deleted anyway, leaving a successful manifest that references missing files even though nothing uploaded them. The HF upload hook already calls `path.unlink()` itself, so preserving retained copies doesn't increase storage for that job. Please add a writer regression with a successful hook that leaves the file untouched.

third_party/llama.cpp/patches/0024-skippy-consume-merge-parts-during-from-parts-materialization.patch
```diff
@@ -0,0 +1,165 @@
...
+@@ -103,6 +103,17 @@ LLAMA_API enum skippy_status skippy_write_gguf_from_parts(
+         const char * output_path,
+         struct skippy_error ** out_error);
+ 
++// Materializing variant that unlinks each input part file as soon as its
++// tensors have been absorbed into the output. Use from staging pipelines whose
++// parts are scratch files (for example the HF Jobs sharded split writer, where
++// parts and output coexisting would double local disk usage); never use it on
++// published package files that must survive.
++LLAMA_API enum skippy_status skippy_write_gguf_from_parts_consuming(
```
ndizazzo: [P2] Regenerate the public API reference with the new consuming symbol

Let's add the required adjacent Doxygen `@brief` and regenerate `website/src/docs/pages/skippy-api.md` from the prepared queue in this change. The public header adds `skippy_write_gguf_from_parts_consuming`, but the checked-in reference still lists only the non-consuming writer. The repo's native API rule explicitly requires the regenerated page alongside an exported-function change. This symbol also transfers file ownership and deletes inputs, so the public documentation needs to state those semantics and failure behavior. Please run the generator's `--check` and compile the changed header as C11 and C++17.

ndizazzo (changes requested):
I agree with making replacement explicit and with reducing the package writer's disk footprint. The fresh target queue marker now wins over the published/catalogued/failed shortcuts, so the earlier sequential double-submit issue is addressed. The Rust and native ABI versions also agree at 0.1.54.

I'd hold this before merging. Republishing needs to preserve a complete published snapshot while new artifacts are uploaded, and the generic after-artifact hook must keep its existing file-retention contract. I've left those findings inline, along with the missing native API documentation update.

There are a few follow-ups I'd make while this is open:

- Let's distinguish a finished attempt from an active queue marker. Success and failure paths retain `automation/queue.json`, and freshness uses the repository's latest modification time. A failed job therefore blocks `--republish` for the default 30 hours after the failure update, and a successful job does the same after publishing. I'd use an attempt identity and terminal state, or document a deliberate cooldown. The current regression test supplies `QueueStatus::Queued` directly, so it doesn't exercise either lifecycle.
- The zero-element filtering is incomplete for package creation. `source_inventory::inspect` now drops placeholders, but `ModelSource::open` still retains them and `ensure_native_inventory_matches` compares that unfiltered set. A source containing the cited zero-element placeholder still fails before writing. I'd cover the complete writer path when finishing this support.
- I'd add native tests showing that the consuming writer produces identical bytes, removes scratch inputs on success, and preserves the default writer's inputs. The new native `std::remove` result is ignored, so a failed unlink can silently defeat the storage-saving behavior. Its trailing `any_of` scan also does quadratic work although the comment says tensors are grouped by source.
- Please update the title and description for the actual scope. This now includes package publication order, scratch-file ownership, upload/storage policy, and a new native ABI, beyond queue republishing. The description's bucket-default claim is also the opposite of the final script. The ABI inventory should explicitly list the new symbol, full signature, owners, 0.1.53 to 0.1.54 bump, and boundary validation.

I reviewed head `6ff4455a1aa8f7f94de2b83e52a5899d716500fc` against base `22998c2732ef490c75fbe417bc17ccf55a325a2a`. Locally, all 11 queue-binary tests passed, `bash -n` passed for the job script, and the diff whitespace check passed. I traced the writer, hook, catalog, and download paths and checked the previous reviews. I didn't run a native writer build, a real GGUF/HF publishing job, or a storage-budget trial.

At this review's check snapshot, PR / Quality passed; PR / Website and PR / macOS passed with their substantive lanes skipped. Linux CPU runtime and static ABI jobs passed, while Linux host/tests and Windows runtime/host/platform jobs were still running. I found no failed current-head job to triage. Those pending lanes and the absence of real publication/storage evidence mean this isn't a CI or operational acceptance sign-off.

crates/model-package/src/bin/queue-unsloth-layer-packages.rs
i386: Fixed in 3ffd407413a2861d0cbf7d12a3cac3f221fcc81a. Republish jobs now upload every artifact and the manifest to a branch pinned from the captured main SHA, then promote the manifest-listed snapshot to main in one parent-guarded Hugging Face commit. A failed upload leaves main untouched; a concurrent main update makes promotion fail instead of overwriting it. Added regression coverage for partial and complete promotion.

crates/skippy-model-package/src/package_v2.rs
i386: Fixed in 3ffd407413a2861d0cbf7d12a3cac3f221fcc81a. The package writer no longer deletes artifacts after a generic successful hook; deletion remains the upload hook’s responsibility. Added a /usr/bin/true regression proving a successful hook may retain every verified artifact.

third_party/llama.cpp/patches/0024-skippy-consume-merge-parts-during-from-parts-materialization.patch
i386: Fixed in 3ffd407413a2861d0cbf7d12a3cac3f221fcc81a. Added adjacent Doxygen documentation for the consuming API, including deletion timing and partial-failure semantics, regenerated the Skippy API page, and compiled the prepared public headers in C11 and C++17 modes.

michaelneale (dismissed):

ndizazzo (dismissed):

i386:
**Part-split fix landed on this branch** — 1d215cbee `feat(package): split oversized layers into byte-balanced part artifacts`

Fixes the deterministic DeepSeek-V4-Pro-0813 eviction (job 6aa5cd855): HF Jobs kills pods above 50G ephemeral storage, layer-00000 alone is 12.9 GiB, and the v2 planner had no per-artifact byte budget.

**Rust-only, no native ABI change, no patch queue change:**
- `plan_artifacts_with_budget` — per-artifact budget (default 8 GiB), byte-balanced part split mirroring `byte_balanced_split_boundaries`; unsplit layers keep the legacy `layers/layer-N.gguf` path, split layers emit `layer-N-partMM.gguf`
- New `part_writer.rs` — byte-preserving Rust GGUF writer streaming tensor payloads directly from source shards at exact catalog offsets (no parts+merge staging, no 2x transient copy); source metadata KV copied as raw wire bytes (llama.cpp type-checks `general.alignment` as u32 — a serde_json round trip widens it to u64 and the native reader rejects the file)
- `--max-artifact-bytes` CLI flag; layer-ordinal parsers in `verify_v2` and `package_carrier` accept `-partNN`
- Works because manifest validation, stage admission, materialization and carrier resolution all bind by tensor-id → (artifact, offset, length) and are artifact-grouping-agnostic — verified before implementation

**Testing on this box (CPU static lib rebuilt against current queue incl. Scama's 0024 doc-comment):** 88/88 crate tests green including 4 new (layout budget split, dominating-tensor-keeps-whole, part-writer payload round-trip, end-to-end tiny-budget split package verified by `verify_v2`); fmt clean; console-print ratchet regenerated (mechanical line-shifts only); clippy 0 warnings.

Rebased onto the refreshed queue-republish branch (3ffd40741). After merge, the V4-Pro job can be resubmitted with the new MESH_LLM_REF.

i386:
Review follow-ups are addressed at `9a97623473bd4054c27e9703ea168f37305630ba`:

- merged current `main` and rebuilt the native queue as contiguous patches 0034–0035, resolving the 0024 collision;
- made terminal success/failure end queue-attempt suppression immediately;
- completed zero-element filtering through the native count/index accessors and the full package writer;
- added native consuming-writer coverage, checked unlink failures, and replaced the quadratic remaining-source scan;
- corrected the planner so oversized tensors are isolated and even a single oversized tensor uses the direct writer path;
- tightened part-path parsing, regenerated the public API docs, bumped ABI 0.1.54 → 0.1.55 in lockstep, and added the required ABI inventory to the PR description.

Validation includes the clean 35-patch CPU static build, 47 native tests, 92 skippy-model-package tests, 25 model-package tests, 159 skippy-runtime tests, 11 skippy-ffi tests, C11/C++17 public-header compilation, clippy for affected crates, and the 1,441-test repository CI validation suite.

The previous quality run was cancelled by the PR metadata edit concurrency group; the replacement run is active.

i386:
Follow-up from the post-merge CI run: the new runtime-events gate from current `main` selected `family-qwen3-dense` at `pull-request` cadence, but that artifact was only authorized for manual/nightly/llama-bump cadences. The native runtime build itself passed; resolution stopped before this PR's code was exercised.

Fixed in `e28cdb95d82b59513683a9813d7f4a59eca17d67` by authorizing the pinned artifact for `pull-request` and `main`, regenerating the registry-backed manifests, and adding a contract test binding the workflow's selected artifact to both executable cadences. Local `just ci-validate` is green (1,441 tests, 9 skipped).

i386:
The authorized live gate then exposed a second CI issue: the Rust integration test passed, but Cargo ran it from the crate directory and wrote the relative evidence path there; the wrapper checked the empty file it had created at the repository root.

Fixed in `121ab7f7a283e2e5c7aab92bc792208e589c36be` by resolving the evidence path before Cargo changes the test working directory, with a regression test that requires the child process to receive an absolute path. `bash -n`, the focused 20-test gate suite, and the full 1,442-test `just ci-validate` suite pass locally.

i386:
Rebased onto main to resolve conflicts. Notable resolutions:

- #1718: kept both the upstream and branch GGUF catalog tests; main's registry.json won for cadences (it already declares pull-request/main for family-qwen3-dense, which satisfies this branch's cadence test); regenerated suite manifests against main's registry digest.
- #1878: took main's split-certified.json and skippy-api.md (newer native recipe / ABI 0.1.54 baseline), and renumbered the reasoning-token-budgets llama patch from 0038 to 0043 to keep the patch sequence contiguous with main.

Validated: cargo check on touched crates, skippy-model gguf tests, and scripts/tests (runtime-events gate + model-artifact registry) all pass locally.

PR #1719, opened 2026-09-09
feat(skippy): size memory plans from measured native buffer footprints

i386:
## What this does

Replaces estimate-only memory planning with measured-reality-driven planning — converging on the vLLM/SGLang budget-driven shape while keeping one thing they don't have: the actual buffer sizes llama.cpp allocated at init.

Stack position: #1707 (ubatch-512) → **this PR** → PR-2 (scheduler). Base branch: `paul/builtin-ubatch-512`.

## Commits (each independently reviewable)

1. `b0042574e` — structured memory-plan events + measured buffer sizes on native log lines (observability)
2. `d98b428b6` — native log aggregator keeps per-kind measured-buffer high-water marks (CPU-offload lines excluded by construction), exported as `measured_native_buffers()`
3. `e98dcff0b` — reconciliation at model-ready: charged proxy vs measured sizes + residual free memory, machine-readable on every start (both solo start paths)
4. `c3c03eb7b` — budget-driven context planning over a measured footprint (utilization target default 0.88, vs vLLM 0.92 / SGLang 0.9 — slightly more conservative for mesh co-hosting + unified-memory page-out); the 85% tax ladder remains the fallback for fresh processes and unusable measurements
5. `f52954019` — lane-scaled compute charge (measured ≈192 MiB/lane on the 5080) + model-keyed footprint (stale/foreign high-water marks degrade to the ladder; measured state cleared on every model-load reset)
6. `b4dc59f46` — **bug fix**: `election::total_model_bytes` returned the checkpoint *directory* inode size (4096) for SafeTensors checkpoint dirs — every safetensors bench leg planned with `model_bytes=4096` (phantom KV budget, blind fit guard). Now sums flat regular files.

## Evidence (white.local, RTX 5080 16 GiB, agentic-replay harness, 2 ABBA passes, 16 measured requests/leg, identical manifests, vLLM 0.27.1 control)

| Leg | Mesh decode tok/s (this PR) | vLLM decode tok/s | Mesh TTFT p50 | vLLM TTFT p50 |
|---|---:|---:|---:|---:|
| granite C1 | 163.75 | 237.51 | 0.415s | 0.247s |
| granite C8 | **38.85** | 34.23 | 3.155s | 1.485s |
| qwen3 C1 | 148.35 | 177.90 | 0.975s | 0.317s |
| qwen3 C8 | 30.21 | 39.65 | 3.732s | 1.840s |

0 failures on granite (16/16 × 2 both arms). One failed request across all legs: qwen C8 pass-1 — admission fail-loud, `resident KV capacity admission rejected request: 206 token deficit (capacity=32768)`; the request (32,974 tok) exceeds honest capacity, where the pre-fix phantom headroom would have accepted it. Recorded as the fix working, not a regression.

Planner event chain captured in-run (`RUST_LOG=info`; note serve hides info events without it):

- granite: `model_bytes=2930319597` (the fix — was 4096), ctx 32768, slots 4 auto → reconciliation: **charged 3.99 GB compute proxy vs 0.69 GB measured, residual 17.9 GB free**
- qwen: `model_bytes=4079423234` → reconciliation: charged 3.99 GB vs 0.69 GB measured, residual 17.96 GB free

## Honest scope

On the single-node path the 85% tax was not the binding constraint for context depth on roomy nodes — the ladder already reaches the native window. What the measured path buys is correctness in both directions (measured per-token KV ≈ 2× the estimate on bench legs → the correct plan is shallower, not deeper). The remaining under-allocation (~11 GiB idle on Granite) lives in the flat 4-lane cap and the split planner's compounding taxes. Lanes-from-budget was evaluated and deliberately **not** taken: lanes 8 is strictly worse on this model/box (19.98 vs 39.4 tok/s C8 decode) and lanes are not memory-free (compute ≈192 MiB/lane — now charged). Idle-VRAM spend moves to the re-plan seam and PR-2.

## Tests

- skippy-runtime 124/124 · mesh-llm-host-runtime 2983/2983 · mesh-llm-routing 23/23
- `cargo fmt --check` clean; `xtask repo-consistency` green (no-console-print, ci-crate-lists, publish-crates)

Artifacts: `white:~/tmp/white-bakeoff-20260908/out/memplan-evidence-{granite,qwen}-c{1,8}/` (per-pass server logs, requests, charts, binary/runtime SHA-256).

## Summary by CodeRabbit

* **New Features**
  * Context planning now uses measured memory usage to select more suitable context lengths and reconcile planned resources with actual allocations.
  * Memory planning diagnostics now provide clearer breakdowns of model, compute, and key-value cache usage.
  * Runtime memory measurements now aggregate high-water marks across supported devices while excluding host-pinned and CPU buffers.

* **Bug Fixes**
  * Directory-based model sizes are now calculated from their immediate regular files, improving resource estimates for packaged models and symlinked snapshots.


ndizazzo (changes requested):
**Verdict: request changes.** The headline feature does not activate on the real code path, and two supporting changes can push the planner in the unsafe direction. I verified all three with temporary probe tests in a throwaway worktree (since removed).

## Blockers

### 1. The budget-driven planner is dead on arrival

`local.rs` calls `record_measured_lane_count(plan.slots)` at *plan* time, before the model loads. The model load then emits `llama_model_loader: loaded meta data with …`, which triggers `reset_model_loading_state()` — and that clears `measured_lane_count` along with the buffer sizes.

Probe result after replaying a real load sequence:

```
REVIEW_PROBE lane_count=None compute_mib=Some(579.83)
```

So `measured_buffers_footprint` builds `lane_count: 0`, and `planned_context_length_budget_driven` early-returns `None` on `lane_count == 0`. Every request falls back to the 85% ladder. Commit 4 — the headline of the PR — never runs.

The existing tests miss this because all four new `context_planning` tests construct `MeasuredBufferFootprint` directly and never cross the record → load-reset → read seam.

This is also consistent with the PR's own evidence: only reconciliation events appear (those don't need `lane_count`), and the "Honest scope" section concedes results match the ladder. That may not be because the 85% tax wasn't binding — it's because the new path never fired.

Fix: either don't clear `measured_lane_count` in `reset_model_loading_state` (it's caller-supplied plan state, not native-log state), or re-record it after the load-start line. Then add an integration test over `record_measured_lane_count` + `process_line` + `measured_native_buffers`.

### 2. Multi-GPU KV is under-measured by a factor of N — OOM direction

`record_measured_buffer_size` keeps a per-kind MAX. llama.cpp prints one KV/compute buffer line *per device*. Probe with two devices:

```
llama_kv_cache: CUDA0 KV buffer size = 544.00 MiB
llama_kv_cache: CUDA1 KV buffer size = 544.00 MiB
→ measured_kv_mib = Some(544.0)   // expected 1088.0
```

Halved measured KV → halved `kv_per_token` → the budget planner buys roughly double the context it can afford. Currently masked by blocker 1, but it goes live the moment blocker 1 is fixed.

The high-water mark is right for *repeated reserves of the same device*, so the fix needs a per-device key — `backend_device` is already parsed in `buffer_size_params`. Sum across distinct devices, HWM within one.

### 3. `dir_file_bytes` returns 0 for HuggingFace snapshot dirs — worse than the bug it fixes

`DirEntry::metadata()` does not traverse symlinks, and HF snapshot dirs are symlinks into `../../blobs/`. Probe: an 8192-byte blob symlinked as `model.safetensors` in a snapshot dir gives `total_model_bytes = 0`. Commit 6 replaces a 4096-byte dir inode size with 0.

`model_bytes = 0` makes the planner treat all VRAM as free and blinds the fit guard — and `total_model_bytes` feeds election, capacity, and fit checks across ~9 call sites plus `mesh-llm models installed`.

Fix is one line: `std::fs::metadata(entry.path())` instead of `entry.metadata()`. Worth confirming whether the SafeTensors resolver actually hands snapshot dirs here, but the symlink case needs handling regardless.

## Should fix before merge

- **`LAST_PLANNED_CONTEXT_LENGTH` isn't model-keyed** while `LAST_MEASURED_MODEL_BYTES` is. Start A (ctx 32768) → loads; start B (ctx 8192) → records 8192 but B fails before its load resets the aggregator; restart A → the model_bytes key matches, so A's KV measured at 32768 is divided by 8192. `kv_per_token` comes out ~4× too high and context ~4× too shallow. The two statics need one atomic snapshot.
- **`CUDA_Host` pinned-host buffers are charged against VRAM.** The filter only excludes lines containing `CPU`; `CUDA_Host` contains `CUDA` and gets counted as device memory. The commit message acknowledges this — it should be excluded, not documented.

## Structure

- The ~22-line `memory_plan.*` tracing block is duplicated near-verbatim across `start_local_skippy_model` and `start_local_package_v2_model` (only `package = "v2"` differs). One helper.
- `local.rs` is already 1203 lines, over the repo's 1k-LoC rule, and this adds ~130 more including two new process-global statics. The measured-footprint statics and helpers are planning state, not local-start orchestration — they belong in their own module.
- Process-global mutable measurement state is fragile against concurrent model starts in one process. State owned by the runtime instance would be safer.
- `planning_profile` is ignored by the budget-driven path. Not a behavior change today (both profiles have `context_slot_target() == 1`), but the 0.88 constant's own doc comment justifies itself by mesh co-hosting while never consulting `SharedMesh`.
- Nit: `1 * 1024 * 1024 * 1024` in two new tests will trip `clippy::identity_op` under `-D warnings`. The PR claims clippy-clean; worth re-checking since CI denies warnings.

## What's good

The reconciliation event (commit 3) is useful on its own — charged proxy vs measured with residual free is exactly the telemetry needed to tune this. `buffer_size_params` emitting `backend_device` is the right primitive, and it's the key blocker 2's fix needs. The 0.88 target with the vLLM/SGLang comparison is well-reasoned, and the "Honest scope" section is candid.

i386:
All three blockers + should-fixes addressed in 1bba0885a (cherry-picked cleanly onto the PR's re-authored base; thanks for the probe evidence — each finding verified against code before fixing).

**Blocker 1 — measured path never fired.** Confirmed as you traced it: the model-load reset wiped the plan-time lane count, so every real footprint had `lane_count: 0` and the budget path early-returned. Fix takes your "caller-supplied plan state" option and extends it: the aggregator no longer holds lane state at all; the host keeps ONE snapshot tuple (`MeasuredPlanSnapshot { model_bytes, context_length, lane_count }`), written atomically when the measured model finishes opening, read as a unit. This also fixes the incoherent-statics should-fix by construction. Seam test added (`aggregator_measured_buffers_survive_model_load_reset` + `measured_footprint_reads_one_coherent_plan_snapshot`). Consequence I've stated plainly in-thread to the owner: the PR's "tax never binding" evidence observation was taken with the measured path disabled — the live re-run to prove the budget path firing on the real surface is queued on white before this merges.

**Blocker 2 — per-device accounting.** `buffer_size_device()` now parses the buffer-name token preceding the line (`CUDA0`, `CUDA1`, `CUDA_Host`, `Metal`, …) — same source `backend_device` telemetry uses, so the event param and the planner key can't drift. HWM within a device, SUM across devices (`aggregator_sums_measured_buffers_across_devices`). `CUDA_Host`/`*_Host` excluded from the VRAM charge (`aggregator_excludes_host_pinned_buffers`).

**Blocker 3 — symlinked snapshots.** `fs::metadata(entry.path())` follows links; HF-layout symlink test added (`symlinked_snapshot_entries_report_target_bytes`). Caught my own test bug in the process (a broken absolute symlink target) — the relative-form link fails open exactly as you predicted, which is what the test now pins.

Should-fixes: deduped the `memory_plan` tracing block into `emit_memory_plan_resolved`; `identity_op` ×2, `question_mark`, `unnecessary_cast` fixed — `clippy --all-targets -- -D warnings` clean on all three crates.

Structure: statics dropped from 3 process-globals to 1 (`MEASURED_PLAN_SNAPSHOT`), and `record_measured_lane_count` is gone from the skippy-runtime API. The larger extraction (planning module out of `local.rs`, runtime-instance-owned state) I've left — it's a mechanical follow-up better done as its own commit once this unblocks, and I'd rather not widen this PR further.

Tests: skippy-runtime 127/127, host-runtime 2984/2984, routing 24/24; fmt + repo-consistency (no-console-print, ci-crate-lists, publish-crates) green.

crates/skippy-runtime/src/logging.rs
```diff
@@ -1092,6 +1244,184 @@ mod tests {
...
+        // Review blocker (PR #1719): the model-load "loaded meta data" line
+        // fires reset_model_loading_state mid-open. Buffer lines arrive both
+        // before and after it; the measured sizes must reflect every line
+        // of the load, never be stranded empty by the reset, and the host
+        // plan tuple (model/context/lanes) lives outside the aggregator
+        // entirely so it is untouched by the reset.
+        let mut aggregator = NativeLogAggregator::default();
+        aggregator.process_line("sched_reserve:        CUDA0 compute buffer size =   579.83 MiB");
+        aggregator.process_line("llama_model_loader: loaded meta data with 26 key-value pairs");
+        aggregator.process_line("llama_kv_cache:        CUDA0 KV buffer size =  1088.00 MiB");
```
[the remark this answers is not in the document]
i386: Verified on current head and resolved. The reset regression now feeds the parser a complete `loaded meta data with 1 key-value pairs and 2 tensors` line, so it exercises the reset path and preserves the prior measurements. Commit 677339279 also restores the package-v2 memory-plan discriminator and moves the planning helpers into `local_memory_plan.rs`; the focused reset test and memory-plan test pass.

i386:
Addressed the remaining review-body items in 677339279:

- Package-v2 starts again emit `memory_plan.package = "v2"`, while direct GGUF starts preserve the existing event shape with no package field.
- The measured-plan snapshot, footprint lookup, plan telemetry, and reconciliation helpers now live in `runtime/local_memory_plan.rs` instead of extending the already-large `runtime/local.rs`.

Validation: formatting, host-runtime and shipped-binary checks, clippy with warnings denied, the focused memory-plan test, and `just no-console-print` pass. The full host-runtime library run passes 2,987 tests and exposes one unrelated timing-sensitive proxy test that times out under full-suite load; that same test passes when run alone.

crates/mesh-llm-host-runtime/src/runtime/local_memory_plan.rs
```diff
@@ -0,0 +1,164 @@
...
+    lane_count: u32,
+}
+
+static MEASURED_PLAN_SNAPSHOT: Mutex<Option<MeasuredPlanSnapshot>> = Mutex::new(None);
+
+/// Return native buffer measurements only when they belong to this model.
+pub(super) fn measured_buffers_footprint(model_bytes: u64) -> Option<MeasuredBufferFootprint> {
+    let measured = skippy_runtime::measured_native_buffers()?;
+    let snapshot = (*MEASURED_PLAN_SNAPSHOT.lock().ok()?)?;
+    if snapshot.model_bytes != model_bytes || snapshot.context_length == 0 {
```
ndizazzo: [P1] Let's cache a complete footprint owned by one load and configuration. The byte count only keys the host tuple; the buffer values here come from the current process-global aggregator. A native load through the split path can reset/replace those values without updating this tuple, and a failed local open skips reconciliation too. Planning the previous model again can therefore combine another load's KV with the previous context and lane count. Even without overlapping loads, two profiles for the same file have identical `model_bytes` while their KV quant, offload/device placement, and batch settings can differ. Reusing a Q4 measurement for an F16 load undercharges KV. I'd capture bytes, model identity, and the allocation-affecting configuration together when that load succeeds, and reject incompatible reuse. A mutex around only the three plan fields doesn't establish that relationship.

crates/skippy-runtime/src/logging.rs
```diff
@@ -338,13 +389,51 @@ impl NativeLogAggregator {
...
 
+    /// Track the largest measured buffer size per kind. A later, smaller line
+    /// (e.g. a per-graph reserve for a shorter context) must not lower the
+    /// high-water mark recorded at full context init. CPU-offload lines are
+    /// skipped: the snapshot feeds VRAM planning, and a CPU-resident buffer
+    /// larger than the accelerator's must not be charged against VRAM.
+    fn record_measured_buffer_size(&mut self, line: &str) {
+        if !line.contains("buffer size") || line.to_ascii_uppercase().contains("CPU") {
+            return;
+        }
```
ndizazzo: [P1] We need to align this device-only footprint with the planner's actual capacity pool. `local.rs` passes `capacity_budget_bytes` as `vram_bytes`; its node capacity originates from `hw.vram_bytes`, which includes RAM-offload capacity on discrete-GPU hosts. With partial CPU offload, these filters omit the CPU KV/compute allocations, but `planned_context_length_budget_driven` treats the remaining device KV as the whole per-token cost against that larger budget. In a synthetic case with 10 GiB post-weight capacity, 0.25 GiB compute, and a 16k cache split into 2 GiB device KV plus 2 GiB CPU KV, the new formula selects 65,536 tokens; counting the full KV footprint selects 32,768, as does the old full-model estimate at 262,144 bytes/token. Keep device and host measurements/budgets separate, or decline measured planning for offloaded loads until they're accounted for. Simply adding CPU bytes to a genuine VRAM budget would also be wrong.

crates/mesh-llm-host-runtime/src/runtime/context_planning.rs
```diff
@@ -273,6 +406,77 @@ fn snap_context_length_down(value: u32) -> u32 {
...
+    let lane_ratio_den = u128::from(measured.lane_count);
+    let compute_charge = u128::from(measured.compute_bytes) * lane_ratio_num / lane_ratio_den;
+    let budget = utilised.saturating_sub(compute_charge);
+    let kv_per_token = u128::from(measured.kv_bytes) / u128::from(measured.context_length);
+    if kv_per_token == 0 {
+        return None;
+    }
+    let max_affordable = (budget / kv_per_token).min(u128::from(u32::MAX)) as u32;
+    if max_affordable == 0 {
+        return None;
```
ndizazzo: [P2] Let's distinguish an exhausted measured budget from an unusable measurement. If valid compute allocations consume the utilization budget, this returns `None`, and the caller immediately sizes KV using the old 85% ladder, discarding the allocation we just learned about. For example, 1 GiB post-weight space with 1 GiB measured compute returns `None`; the existing GQA/Q8 fixture then gets an 8,192-token fallback despite no room for KV. This can occur when a later load has less available capacity or more lanes. Return a no-fit result, reduce the requested allocation, or use an explicitly bounded minimum/failure policy. The test using `compute_bytes: u64::MAX` currently asserts this unsafe fallback rather than separating invalid input from real exhaustion.

crates/mesh-llm-host-runtime/src/runtime/context_planning.rs
```diff
@@ -128,20 +181,48 @@ pub(super) struct RuntimeResourcePlanInput<'a> {
...
+        .metadata
+        .and_then(|metadata| {
+            input
+                .kv_cache_quant
+                .kv_cache_bytes_per_token(metadata)
+                .map(|bytes| scale_by_layer_fraction(bytes, &input))
+        })
+        .unwrap_or(0);
+    let kv_budget_bytes = usable_kv_cache_budget(input.vram_bytes, input.model_bytes);
+    let planned_kv_bytes = kv_bytes_per_token.saturating_mul(u64::from(context_length));
```
ndizazzo: [P2] I'd have the selected planner return its actual accounting with the context. These fields always use the metadata KV estimate and the 85% ladder, even when context selection used measured KV, a lane-scaled compute charge, and 88% utilization. `reconcile_memory_plan_with_measurements` then reports the old 15% reserve as the compute charge the planner actually held back. That makes the new telemetry misleading precisely on the successful measured path, and there's no planner-source field to prove which path ran. Preserve the ladder values as explicitly named comparison estimates if useful, but emit the selected method and its real KV rate, budget, and compute charge.

ndizazzo (changes requested):
I like the direction of replacing the fixed reserve with measurements, and the earlier fixes are present on this head: per-device summation, host-pinned exclusion, symlink following, the corrected reset test, and the `local_memory_plan` extraction. I'd still hold the merge until we fix measurement ownership and match the accounting to the memory pool we're budgeting. Those can change the selected context in the unsafe direction even when the parser's numbers are correct.

I reviewed `2ed52ef1678b96535b49ece437bd090b498ed973` against stack base `53a848bc657c80044d1facc2304de60449902d6a` on `paul/builtin-ubatch-512`, including all seven changed files, local and split load callers, native log resets, capacity derivation, and the previous reviews. This is a review of this stack delta, not a certification of the inherited base or the later scheduler work. Keep that dependency order when merging.

All five current `PR /` results are green. I checked the scheduled jobs as well: Quality's format, consistency, and Clippy jobs passed; Linux's CPU host/runtime, four Rust test batches, SafeTensors smoke, core inference, two-node client, and KV-cache smoke passed. The macOS, Windows, Website, and GPU-heavy execution paths were skipped by the plan. There was no failed job on this head to diagnose. [Linux run](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34692699123), [Quality run](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34692699044).

My additional validation was source tracing and dependency-free arithmetic probes, not a Rust build or GPU reproduction. The examples in the inline comments are synthetic calculations, not measured OOMs. The benchmark artifacts cited in the description weren't re-run or inspected here. We still need an end-to-end test that proves a compatible second load selects the measured path, plus configuration-change and offload cases. In particular, direct SafeTensors directories still go through `scan_gguf_compact_meta`, which cannot read a directory, and the measured planner requires metadata. Those benchmark legs cannot establish that this branch of the planner ran. Also, `residual_free_bytes` is a calculated local-capacity remainder, which may include RAM, rather than a device-free-memory reading.

For cleanup, I'd move the new measurement/accounting tests into an owning test module: `context_planning.rs` is now 1,060 lines and `logging.rs` is 1,813. The successful-snapshot test should seed real buffer records and require `Some`, instead of accepting `None` through an `if let` that can skip the assertions.

michaelneale (approved):
Much needed

crates/mesh-llm-host-runtime/src/runtime/local_memory_plan.rs
i386: Fixed in a5a0ff099fa225a4c3e3a79ccd6b79f4c389c01b. The reusable snapshot now owns the complete compute/KV/context/lane footprint and is keyed by the exact model path, capacity, and allocation-affecting load configuration. It is captured only after a successful load, so a later aggregator reset or another load cannot be combined with the cached plan.

crates/skippy-runtime/src/logging.rs
i386: Fixed in a5a0ff099fa225a4c3e3a79ccd6b79f4c389c01b. Native logging now records whether CPU KV or compute buffers were observed. Such mixed host/device measurements remain excluded from device totals and are not reusable for measured planning or residual-capacity reconciliation.

crates/mesh-llm-host-runtime/src/runtime/context_planning.rs
i386: Fixed in a5a0ff099fa225a4c3e3a79ccd6b79f4c389c01b. A structurally valid measured footprint now always selects measured accounting. When it cannot fit the minimum context, the plan reports measured_fit=false and model startup fails closed instead of falling through to the static estimate.

crates/mesh-llm-host-runtime/src/runtime/context_planning.rs
i386: Fixed in a5a0ff099fa225a4c3e3a79ccd6b79f4c389c01b. The plan breakdown now carries the selected planning source, actual compute charge, KV budget and KV-per-token accounting, plus measured fit status. Resolution telemetry emits those selected values and whether the measurement is reusable.

ndizazzo (approved):

PR #1721, opened 2026-09-09
fix(llama): rebase patch queue onto upstream b31b71f3a0

i386:
# Repair PR #1721 — llama.cpp canary: patch-queue rebase onto `b31b71f3a0`

Automated llama.cpp canary repair (mode: patch-queue), certified green by the family
battery on the `family-certify` runner.

- **Old pinned upstream:** `cc83d7b482` (`cc83d7b4824f73cfdda4dfbb47ee39804f71b328`)
- **New pinned upstream:** `b31b71f3a0` (`b31b71f3a076bfc4278daad442203a9c51c6e676`)
- **Delta:** 188 upstream commits, 468 files, +52,334 / −6,858
- **Branch commits:** `2ce96e51d` (queue rebase) + `31fad9163` (pin flip in `upstream.txt`)

## What happened

The nightly canary prepared upstream `b31b71f3a0` (188 commits ahead of the pin) and
`prepare-llama.sh` could no longer apply the core Skippy queue. This PR re-authors the
conflicting queue patches against the new tree, regenerates the deterministic family
stage-control shards, and flips the pin.

## Key upstream changes `cc83d7b4..b31b71f3a0`

**Core runtime (largest conflict surface):**
- `llama_kv_cells`: the per-sequence position bookkeeping became set-based —
  `std::map<llama_pos,int> seq_pos` → `std::set<std::pair<llama_pos,uint32_t>>` with new
  `seq_pos_inc`/`seq_pos_dec` helpers; plus n-gram history lookup via the sequence-position
  index (#28040), early-out sequence scan (#28011), faster restore of non-contiguous cells
  (#27991).
- `llama_context`: rope-scaling resolution moved into `cparams`; YaRN autoscales
  `n_ctx_train` (#28030); lazy tensor loading disabled by default on iGPUs (#28326);
  `TENSOR_READ_LAZY` handling improved (#27837); model-load RAM-peak prevention (#27483);
  loader `n_layer_all`/`n_layer_nextn` ordering fixes (#28173, #28159).
- `llm_graph_context`: `build_attn_mha` gained an `n_kv_max` parameter; new upstream
  `ctx_other` blocks for GEMMA4/EAGLE3; SWIGLU_CLAMP op (#27930); `ggml_prec` spec (#26675).
- Speculative decoding: single-device drafter creates a meta backend wrapper (#28390);
  DFlash encoder fused into KV-cache injection (#27310); k_rot/Hadamard copy guard during
  context shift (#27967).
- New architectures: Tencent Hy 4 `hy_v4` (#28127), Spark2.5 (#27868),
  NemotronHPuzzle (#25444). Fixes: gemma4 vision (#28335) and gemma4-assistant (#28183),
  qwen4exp seq_cp/position/mtmd fixes + recurrent rollback (#27941, #28123), Kimi-K3
  recurrent rollback (#28466), kimi-linear fused wqkv, deepseek4 vision input (#28154),
  `n_expert_used_max` (#28323), per-layer `n_ff_exp` API.
- Chat layout: specialized parsers split into `common/parsers/` (#27764) with
  `parsers.h` + `sources.cmake` (new upstream `gemma4.cpp`, `deepseek.cpp` parsers).
- mtmd: `const` propagated through preprocessor base-class virtuals (#28307, #28310).

**Backends:** 26 Metal commits (new `ggml-metal-tuning.cpp`; sparse FA #28098; fa-vec
tunings for M2/M3 families; PSO/compile-options area our patch 0009 touches), 12 Vulkan
(incl. 32-bit Vulkan-Hpp handle fix, TQ1_0), 8 OpenCL (new `sdpa_xmem` and
gemv/gemm noshuffle kernels), 6 CUDA + reverts (sparse-fa for DSV4/GLM #27970, fattn-mma,
MMQ RDNA3, SM87 MMVQ→MMQ crossover), SYCL/ROCm (`--fit`/`--fit-target` #27629, ROCm 10.0),
Hexagon/Snapdragon HTP work (+19.8k-line `docs/ops/Hexagon.csv`), WebGPU, new
`ggml-cpu/iqp.cpp`.

**Vendored/build/tests:** cpp-httplib vendor bump (~1.1k lines in `httplib.cpp`), `tools/ui`
asset embedding reworked (`embed.cpp` removed, `scripts/ui-assets.cmake`), llama version
bump to 3.4, `test-backend-ops` +925 lines, `save_models` test-helper refactor.

## Patch-queue evolution (per patch)

All 23 core patches were re-serialized (mbox `From`/`[PATCH nn/23]` headers, git
signature `2.54.0 (Apple Git-157)` → `2.55.0`, blob index lines, hunk offsets). Body-hunk
comparison old-vs-new (metadata and headers excluded) classifies the queue as:

**Genuinely re-authored (6):**

- **0001 graph/memory/model foundations (~138 body lines changed)** — the big one, spread
  across `llama-arch`, `llama-context`, `llama-kv-cells.h`, `llama-graph`,
  `llama-model-loader`/`mmap`:
  - kept both INKLING and new upstream arch entries (vocab pre-type, `sm_tensor` switch —
    including upstream's moved `QWEN4EXP` TODO — and model mapping);
  - adapted the Skippy `seq_pos` index-mismatch tracking to upstream's set-based
    `llama_kv_cells` (`seq_pos_inc`/`seq_pos_dec`);
  - kept the Skippy `ctx_other` clauses beside upstream's new GEMMA4/EAGLE3 `ctx_other`
    blocks;
  - kept our uniform `graph_max_nodes` budget instead of upstream's arch allowlist;
  - combined the Skippy `graph_build_inputs` ctor with upstream's YaRN `n_ctx_train`
    readjust; kept upstream's guarded `set_input_k_rot` comment;
  - adopted upstream's `build_attn_mha` `n_kv_max` argument and the
    `cparams.rope_scaling_type` refactor.
- **0007 chat tokenization (~258 lines)** — the Inkling chat parser was re-homed into
  upstream's new `common/parsers/` layout (`parsers/inkling.cpp`, `parsers.h` declaration,
  `sources.cmake`), alongside upstream's #27764 parser split.
- **0019 stage-free model semantics (~42 lines)** — recurrent_conv/recurrent_ssm state
  kinds preserved over upstream's fused kimi-linear wqkv path and qwen4exp per-slot
  conv-state rollback loop; stage-filter gate re-based over upstream's PLE row computation;
  `inkling.cpp` follows upstream's per-layer `n_ff_exp` API.
- **0008 Inkling conversion/mtmd (~12 lines)** — Inkling audio/image preprocessors follow
  upstream's `const` base-class virtuals (#28307/#28310); GGUF arch-map hunk keeps
  `HunYuanVL` mapping and adds upstream's `HYV4ForCausalLM → hy_v4`.
- **0010 upstream-compat tests (~8 lines)** — arch test switches `LLM_ARCH_DEEPSEEK4` →
  `LLM_ARCH_HY_V4`; `glm_dsa_backend_config_supported` re-based over upstream's
  `save_models` signature refactor; loader accounting test constructs the loader
  `ctx_key` with the new lazy flag (`/*lazy=*/false`).
- **0009 Metal staging (2 lines)** — the Skippy PSO-cache helper in
  `ggml-metal-device.m` now sits beside upstream's tensor-API compile-options helper
  (`options.languageVersion = MTLLanguageVersion4_0_GGML`).

**Small mechanical fixes (4):**

- **0011** — `LLAMA_VERSION_MINOR` 3 → 4 (upstream version bump).
- **0013** — `build_attn_mha` call sites adopt the new `n_kv_max` argument.
- **0023** — blank-line placement inside existing hunks only.
- **0015** — renamed `...keep-MTP-blocks-outside-stage-ranges` →
  `...outside-pipeline-stage-ranges` (name now matches the subject line); hunks unchanged.

**Body-identical (13):** 0002, 0003, 0004, 0005, 0006, 0012, 0014, 0016, 0017, 0018, 0020,
0021, 0022 — headers, offsets, and blob index lines only. Notably **0002 (public runtime
ABI) is body-identical** — no ABI content changed.

**Generated family shards (`patches/generated/`, 47 shards):** fully regenerated with the
Clang rewriter at the new base. Rewriter summary at `b31b71f3a0`: 130 transformable /
20 auxiliary / 1 whole-model builders, 0 unsupported shapes, 0 refusals, 0
already-transformed, idempotent second pass. `ci/llama-canary/generated-family-map.json`
gained the 15 certified families that were missing from the source map (apertus, arcee,
bloom, exaone4, gemma, gemma3, glm4-expansion, minicpm3, mpt, nemotron-expansion, olmo2,
phi2, qwen2, stablelm, starcoder2), so the deterministic generation gate can run: 13
brand-new family shards, a rebuilt llama-only shard (`0023-family-llama`), a new
`llama--olmo2` combination shard, and a much smaller residual `0047-family-unmapped`
shard. The final checked-in shards were verified byte-identical to a fresh regeneration
(`diff -qr` clean).

## Validation

- `prepare-llama.sh` applies the full 70-patch queue clean end to end in a fresh clone;
  the prepared tree is the pinned base plus the queue.
- `check-skippy-generated-family-patch.sh` green: rewriter tool fixtures, regenerated-shard
  byte-identity (re-verified in this review turn), idempotent second pass, transformed-tree
  compile, 33/33 Skippy no-alloc graph verifier tests, zero drift. (A stale
  `patch-drift.diff` in the artifact dir dates from an intermediate repair iteration and is
  superseded by the final clean regeneration.)
- arm64 Metal llama build green, including the upstream CTest suite
  (`LLAMA_STAGE_UPSTREAM_TESTS=ON`); native archive is arm64-only.
- `cargo check`/`build` green for `skippy-ffi`, `skippy-runtime`, `skippy-server`.
- On the `family-certify` runner: parity manifest validation, generated-family-patch gate,
  Skippy smoke, live two-node package-v2 model_pin matrix (6 pinned rows: llama,
  qwen3-dense, granite-hybrid, mamba, falcon-h1, qwen3.5), and the family certification
  battery all green — the wrapper only published this PR after the full closeout. Evidence:
  the `llama-family-battery-<run_id>-<attempt>` workflow artifact and
  `.deps/llama-canary-repair-battery.log` on the runner (the local
  `target/family-battery/agent-repair-1` copy is only a stub).

## Risks for reviewers

1. **ABI impact: none.** `SKIPPY_ABI_VERSION` stays `0.1.53` in both `include/skippy/common.h`
   (patch 0002, body-identical) and `crates/skippy-ffi/src/lib.rs`. No Rust FFI changes; no
   native-runtime package bump. Existing packages and hosts remain interoperable.
2. **0001 is the highest-risk patch.** It adapts shared core paths (`llama-context`,
   `llama-kv-cells`, `llama-graph`, loader) that every model family executes, not just
   Skippy split paths. Two deliberate decisions to double-check: keeping our uniform
   `graph_max_nodes` budget (no arch allowlist) and the set-based `seq_pos` bookkeeping
   adaptation in the index-mismatch tracker.
3. **Inherited upstream behavior changes** — reviewers should not attribute these to the
   queue: iGPU lazy-load default off (#28326), YaRN `n_ctx_train` autoscale (#28030),
   `build_attn_mha` `n_kv_max`, llama 3.4 version bump, cpp-httplib vendor bump. Expect
   small shifts in parity/benchmark numbers.
4. **15 newly mapped families** now generate their own deterministic shards instead of
   riding in `unmapped`. Their shard coverage gets its first battery exposure in this run's
   targeted (changed-families + sentinels) certification; the next nightly re-covers the
   full manifest.
5. **Lanes: none newly failing, none excluded.** This branch changes no workflow or gate
   script; all certification lanes (parity manifest validation, generated-family-patch
   gate, upstream CTest + Skippy native tests, smoke, live model_pin matrix, family
   battery) ran green. The parity-manifest exclusions (`102 needs_boundary_registration`,
   `14 non_causal_aux`, `11 package_or_remote_only`, `4 candidate_multimodal`,
   `4 implementation_base`, `5 no_public_gguf_candidate`; `9 certified` +
   `8 certified_package_only`) are standing manifest state, untouched by this PR.
6. **Merge effect:** `31fad9163` is the actual pin flip — after merge, `main`'s next canary
   run certifies at `b31b71f3a0` and the `update-pin` auto-commit becomes a no-op.

## Protocol note

Native-runtime layer only. The mesh wire protocol, gossip, plugin protocol, and
mixed-version node operation are untouched.


ndizazzo (approved):
## Approve — no blockers found

I verified the PR body's claims against the actual diff rather than taking them at face value. Everything material checks out.

### What I verified

**Merge mechanics** — `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` by exactly 2 commits (#1722, #1723). Those two touch a **disjoint file set** from this PR's 75 files, so there is no textual conflict. `main` has no branch protection, so neither the BEHIND state nor the absent review blocks the merge button.

**CI** — 91 pass, 0 fail, 0 pending, 15 conditional skips. The four Linux native-runtime lanes, Windows CUDA/ROCm, and macOS arm64 all necessarily run `prepare-llama.sh` and apply the full queue, so CI independently corroborates that the rebased queue applies and compiles across every backend. Inference, KV-caching (dense + recurrent), model-download, and two-node client smokes are green.

**Queue integrity** — patch numbering stays unique and contiguous. The apparent `0017` duplicate in a name-only listing is an artifact of `0017-family-llama.patch` being *deleted*. `generated/series` (47 entries) matches the 47 on-disk shards exactly, and `series.json` cross-checks with zero orphans in either direction. Pin flip in `upstream.txt` is correct.

**ABI** — the "no ABI impact" claim holds. Patch 0002 sets `SKIPPY_ABI_VERSION_PATCH 52`, patch 0020 bumps it to 53; `crates/skippy-ffi/src/lib.rs` mirrors `0.1.53`. In sync. (The body's attribution of the `0.1.53` constant to patch 0002 is imprecise, but the substance is right.)

**Family map** — exactly the 15 claimed families added, zero removals. `olmo2` getting its own entry while `llama` already lists `olmo2.cpp` is what produces the new `0024-family-llama--olmo2` overlap shard, as described.

**Patch 0001 (the flagged high-risk one)** — the 600-line diffstat is overwhelmingly hunk-offset churn. The actual semantic delta is five changes, all justified:

- `dflash_embd` branch adopted from upstream's new encoder path, with Skippy's `mtp_embd` removal preserved
- `LLM_ARCH_HY_V4` replacing `DEEPSEEK4` in an arch condition
- `seq_pos_inc/dec` index-mismatch tracking **re-authored to hoist the check to call sites** (`pos[i] != static_cast<llama_pos>(i)`) instead of changing upstream's method signatures — this *reduces* divergence from upstream
- `LLAMA_VOCAB_PRE_TYPE_INKLING` 57 → 59 (enum is load-time derived, never serialized, so renumbering is safe)
- The `set_input_k_rot` null/buffer guard **is not dropped** — it now appears as a *context* line, meaning upstream adopted it. I chased this specifically because it read like a deletion in the diff; the guard still exists in the transformed tree.

### Non-blocking notes

1. **Update the branch before merging.** Optional given no protection, but it means CI's green run reflects a tree that is 2 commits stale. Cheap to refresh.
2. **Stale `patch-drift.diff`** in the artifact directory from an intermediate iteration — the body admits this. Worth clearing so a future reader does not mistake it for real drift.
3. **The 15 newly-mapped families** got targeted certification only; full manifest exposure lands on the next nightly. That is the one genuine residual risk, and it is correctly disclosed.
4. ABI attribution nit (0002 vs 0020) as noted above.

I reviewed 0007/0019 by classification and CI coverage (chat-parser and recurrent-family paths are exercised by the green chat/KV-recurrent smokes) rather than line-by-line — flagging that so the coverage boundary is explicit.

Issue #1648, opened 2026-09-04
Replace file-per-segment L3 KV storage with packed append-only storage


i386:
## Context

PR #1632 deliberately cuts growing prefixes on cache geometry so later turns reuse prior state. A realistic 19K-token prefix can therefore contain roughly 9,500 segment references. Today each content-addressed segment is a separate file, eviction scans/parses manifests to reconstruct reference counts, and reads pay filesystem metadata and locality costs.

[SGLANG-LSM](https://arxiv.org/abs/2511.16138) identifies file-per-object metadata and locality as disk-KV bottlenecks and reports up to 143% more cache hits and 24% lower TTFT in its evaluated workloads.

Coordinated in Buzz channel `#skippy-kv` and tracked as an expanded optimization candidate for #1632.

## Proposed direction

Retain the existing logical format and exactness contract while replacing file-per-segment physical storage with append-only packfiles or an LSM-style substrate:

- digest to pack/offset/length/checksum index;
- batched sequential writes and reads;
- tombstones for logical deletion;
- background compaction with bounded foreground impact;
- atomic manifest publication only after all referenced segments are durable;
- reference-aware accounting of shared segments;
- crash recovery, quarantine, and index rebuild;
- hard-budget, minimum-free, pin/hold, prune, and clear semantics equivalent to #1632.

## Checklist

- [ ] Define the versioned packed-store format, index, and migration/compatibility behavior.
- [ ] Preserve digest verification and exact numerical/model identity.
- [ ] Make append, index commit, manifest commit, tombstone, and compaction crash-safe.
- [ ] Batch segment reads by physical locality without changing logical segment boundaries.
- [ ] Bound compaction memory, write amplification, and foreground latency.
- [ ] Preserve startup cleanup and safe fallback to cold prefill on any cache failure.
- [ ] Add corruption, torn-write, stale-index, duplicate-segment, concurrent writer, eviction, prune, and clear tests.
- [ ] Benchmark against the current file-per-segment #1632 implementation using the shared instrumentation gate.

## Acceptance gate

Promote only if a release-build A/B on the same hardware and workload improves end-to-end cache-hit TTFT and/or sustained cache throughput, does not regress p99 decode latency beyond the existing 5% limit, preserves exact output, and keeps write amplification within an explicitly reported bound.

PR #1726, opened 2026-09-09
fix(skippy): preserve durable restart checkpoints

i386:
The durable L3 writer emitted one file per logical segment and let the background record worker outlive server shutdown. A 1,792-token checkpoint therefore required 1,680 segment publications and was committed only after the restart began, leaving the first restored request with 501 usable tokens.

This change writes each checkpoint's missing logical segments into one immutable pack with a local offset index, rebuilds and validates that index at startup, accounts for the physical pack and sidecar in eviction, and drains the final record worker during graceful shutdown. Portable manifests and logical segment digests stay unchanged.

The pre-restack release replay on the patch-equivalent commit `eb1718386` produced:

| cohort | control | packed L3 |
|---|---:|---:|
| restore p50 | 370 ms, 0% cached | 264 ms, 99.7% cached |
| warm p50 | 13 ms, 99.9% cached | 13 ms, 99.9% cached |
| failed requests | 0 | 0 |

The fix makes the latest useful checkpoint durable before restart: 1,951 of 1,957 prompt tokens restore. The resulting 1.40x speedup remains below #1576's 2x gate; #1736 reduces the next read-side cost.

## Stack

- Base: #1714
- This PR: packed physical storage and shutdown drain for #1648
- Next: #1736 streaming restore

## Validation

- `cargo test -p skippy-cache --lib` — 121 passed, 1 ignored
- `cargo test -p skippy-server --lib` — 706 passed, 3 ignored
- warning-denying Clippy for `skippy-cache`, `skippy-server`, and `mesh-llm`
- `cargo check -p mesh-llm`
- `cargo fmt --all --check`
- release host and native runtime builds

### DIFF
--- a/crates/skippy-cache/src/l3.rs
+++ b/crates/skippy-cache/src/l3.rs
@@ -14,7 +14,7 @@
 //!   is present and the assembled payload digest matches. Partial state can
 //!   never be loaded — there is nothing to load until commit.
 //! - **Idempotency**: putting a segment that already exists is a no-op;
-//!   concurrent writers of the same bytes converge on one file via
+//!   concurrent writers of the same bytes converge on one object via
 //!   temp-file + atomic rename.
 //! - **Capped budget**: `enforce_budget` evicts oldest manifests first (the
 //!   newest is never evicted) and garbage-collects unreferenced segments.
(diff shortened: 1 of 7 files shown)


i386:
Consolidated into #1736: https://github.com/Mesh-LLM/mesh-llm/pull/1736

The final combined tree retains every code, test, and documentation change from this PR. Review continues on #1736.

PR #1727, opened 2026-09-09
fix(ci): simplify llama canary state machine

i386:
The changed-pin canary currently splits work between the normal workflow and a second repair wrapper, which can repeat full certification until GitHub kills the outer step and strands green runner-local commits. This change gives changed pins one deterministic `prepare -> build -> certify -> publish` state machine and guarantees a terminal, run-specific PR before the Actions timeout.

- Resolve the candidate SHA without applying patches, then route every changed pin into the wrapper.
- Let agents repair a failed prepare, build, or certification phase locally; after any edit, restart at prepare and rerun the complete build before full certification.
- Run the complete `llama-bump` family cohort for every changed pin.
- Reserve 30 minutes inside the 720-minute step for publication. A green exact head opens a normal PR; exhausted repair turns or time open an explicitly uncertified draft PR.
- Use a unique branch per workflow run, attempt, and upstream SHA, with one non-force terminal push.
- Remove the post-green modifying agent turn and the direct pin push to `main`.
- Keep unchanged nightly and forced certification read-only on their existing bounded path.

Validation:

- `just ci-shellcheck scripts/llama-canary-agent-repair.sh`
- 41 focused canary contract tests
- `just ci-validate` — 1,060 tests passed, 8 skipped, plus workflow, release, crate-list, and repository-consistency checks
- `bash -n scripts/llama-canary-agent-repair.sh`
- `git diff --check`

## Summary by CodeRabbit

- **CI Improvements**
  - Changed upstream pins now use a predictable prepare, build, certify, and publish process.
  - Unchanged scheduled and manually triggered certifications remain read-only.
  - Repair pull requests include upstream change summaries and avoid duplicate reviews for the same candidate.

- **Reliability**
  - Repair attempts are time-bounded and limited per phase.
  - Each run uses isolated branches and dedicated evidence.

- **Security**
  - Credentials are protected during repository updates and repair operations.

- **Tests**
  - Updated automated coverage for workflow behavior and security controls.


scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,434 @@
...
+  else
+    title="draft(llama): failed canary at ${UPSTREAM_SHA:0:10}"
+    create_args=(--draft)
   fi
-  gh_repair gh issue create --title "llama canary repair needs human assistance (upstream ${UPSTREAM_SHA:0:10})" \
-    --body "The canary repair loop could not open a PR on \`$BRANCH\` (branch missing or no diff). See the canary run log for the repair-loop outcome." \
-    | grep -oE '[0-9]+$' || true
-  return 0
+  gh_repair gh pr create --base main --head "$BRANCH" "${create_args[@]}" \
+    --title "$title" --body-file "$PR_BODY" 2>/dev/null | grep -oE '[0-9]+$'
```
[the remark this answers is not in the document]
i386: Fixed in 1e388c0e8. Terminal PR creation now preserves redacted stderr, emits an explicit error when gh pr create fails, and separately rejects success output that contains no PR number. Validation: shellcheck, bash -n, 42 focused canary contracts, and just ci-validate (1,061 tests; 8 skipped).

ndizazzo (dismissed):
This is the right direction, and the security story is much better than what it replaces. Dropping the `update-pin` job takes `contents: write` and the push-straight-to-`main` path off the table, removing `post_green_review_turn` means the agent can no longer touch code after certification, and running the agent under `env -u GH_TOKEN -u GITHUB_TOKEN -u CANARY_REPAIR_TOKEN` with no `git push` and no `gh` leaves it with no way to publish anything itself. The single ordered prepare/build/certify loop is much easier to reason about than the old two-mode wrapper, restarting at prepare after every agent edit closes the stale-evidence hole, and `verify_pr_head` on a non-force push binds the PR head to the exact certified commit. Dropping the changed-file family narrowing so `llama-bump` certifies the full cohort is the right call too. Docs updated in the same change, which I appreciate.

Locally: `bash -n` clean, `shellcheck -S warning` clean, `actionlint` clean, and all 42 contract tests pass.

Approving. I left the specifics as inline comments, the main one being that nothing closes the previous canary PR or branch, so an unmerged week turns into 7 PRs and ~80 hours on the certify runner.

scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,442 @@
...
 MAX_REPAIR_TURNS="${CANARY_REPAIR_MAX_TURNS:-2}"
-BRANCH="llama-canary/patch-queue-fix"
-BATTERY_LOG="$ROOT/.deps/llama-canary-repair-battery.log"
-PIN_FILE="$ROOT/third_party/llama.cpp/upstream.txt"
+REPAIR_BUDGET_SECONDS="${CANARY_REPAIR_BUDGET_SECONDS:-41400}"
+PUBLISH_RESERVE_SECONDS="${CANARY_PUBLISH_RESERVE_SECONDS:-1800}"
+RUN_ID="${GITHUB_RUN_ID:-manual-$(date +%s)}"
+RUN_ATTEMPT="${GITHUB_RUN_ATTEMPT:-1}"
+RUN_KEY="${RUN_ID}-${RUN_ATTEMPT}"
+BRANCH="llama-canary/repair-${RUN_KEY}-${UPSTREAM_SHA:0:10}"
```
ndizazzo, line 35: This branch name is unique per run, attempt, and SHA, and nothing closes the previous one. Now that `update-pin` is gone, `third_party/llama.cpp/upstream.txt` on `main` only advances when a human merges one of these PRs, so `changed == 'true'` stays true every scheduled night. That means a fresh ~11.5h occupancy of the single `family-certify` runner plus a new PR and a new branch each run. A week without a merge is 7 PRs, 7 branches, and about 80 runner-hours on the same upstream SHA. The old flow reused `llama-canary/patch-queue-fix`, so it self-deduplicated.

We should either close prior open `llama-canary/repair-*` PRs (and delete their branches) when publishing a new terminal state, or skip the scheduled changed-pin path entirely while an open canary PR already covers that SHA.

.github/workflows/llama-upstream-canary.yml
```diff
@@ -139,64 +131,64 @@ jobs:
...
+      - name: Stop after changed-pin terminal PR
+        if: steps.sha.outputs.changed == 'true'
+        env:
+          CHANGED_CANARY_OUTCOME: ${{ steps.changed_canary.outcome }}
+        run: |
+          {
+            echo "## llama.cpp changed-pin canary"
+            echo
+            echo "The deterministic state machine reached a terminal outcome: **${CHANGED_CANARY_OUTCOME}**."
+            echo "Its run-specific PR contains either the exact certified commit or a draft of the bounded failing state."
```
ndizazzo: This asserts a run-specific PR exists, but a few paths exit before `report_terminal` ever runs, so the summary points at a PR that isn't there. `${LLAMA_STAGE_BUILD_DIR:?}` and `${HF_CACHE:?}` in the wrapper exit the shell outright, and the early bail-outs (missing `opencode`, no agent credentials, unset `CANARY_REPAIR_TOKEN`, failed `check_repair_token_permissions`) all quit before any branch is pushed.

Let's validate `HF_CACHE` and `LLAMA_STAGE_BUILD_DIR` up front next to the token check, and branch this summary on `CHANGED_CANARY_OUTCOME` instead of claiming a PR unconditionally.

scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,442 @@
...
-  # it remains local until publish_repair_branch is called at terminal status.
+run_full_build() {
+  local archive arches
+  : > "$BUILD_LOG"
+  echo "state-machine phase: build" | tee -a "$BUILD_LOG"
+  run_logged "complete patched llama.cpp build" "$BUILD_LOG" env \
+    LLAMA_STAGE_UPSTREAM_TESTS=ON uv run --no-project --with jinja2==3.1.6 -- \
+    arch -arm64 bash scripts/build-llama.sh -DCMAKE_OSX_ARCHITECTURES=arm64 \
+    || return 1
+  archive="${LLAMA_STAGE_BUILD_DIR:?}/src/libllama.a"
```
ndizazzo: `:?` here (and on `${HF_CACHE:?}` in the certify phase) kills the shell immediately, which skips `report_terminal failed` and leaves no terminal PR. A misconfigured runner env turns into a silent vanish rather than a draft PR explaining what happened. Worth hoisting these into an up-front check.

scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,442 @@
...
-  # never leak the token in the hint.
-  if ! git push "$(repair_remote)" "+HEAD:refs/heads/${BRANCH}" 2> >(redact_token >&2); then
-    echo "ERROR: could not push ${BRANCH}. If git reported 403/denied above, the identity behind CANARY_REPAIR_TOKEN lacks write access to ${GITHUB_REPOSITORY:?}: grant that account Contents+Pull requests write (or mint the PAT from an account that has it) and update the secret." >&2
+  if ! git push "$(repair_remote)" "HEAD:refs/heads/${BRANCH}" 2> >(redact_token >&2); then
+    echo "ERROR: could not push ${BRANCH}; the identity behind CANARY_REPAIR_TOKEN needs Contents and pull-request write access" >&2
     return 1
   fi
 }
 
+write_pr_body() {
```
ndizazzo, line 355: The body carries SHAs, the run key, the terminal commit, and repair-turn counts, but nothing about what actually changed upstream. `summarize-llama-upstream.sh` only runs in the `Generate summary` step, which is gated on `steps.prepare.outcome == 'success'`, and `prepare` is skipped for changed pins. So for exactly the runs a human has to review, the summary never fires.

We should fold `scripts/summarize-llama-upstream.sh` into this function. Reviewing an 11-hour certification with no idea what moved upstream is rough.

scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,442 @@
...
+  esac
+}
+
+repair_prompt() {
+  local phase="$1" log turn
+  log="$(phase_log "$phase")"
+  turn="$(( $(phase_turns "$phase") + 1 ))"
+  printf 'The llama.cpp canary %s phase failed at upstream %s (repair turn %s of %s for this phase).
+
+Read ci/llama-canary/agent-repair-prompt.md and the repository instructions it names. That runbook describes the former two-mode workflow; this invocation is the single prepare -> build -> certify -> publish state machine. Reproduce this run with scripts/prepare-llama.sh using the target in .deps/llama-canary-target-sha. Ignore the old branch and publication instructions in that runbook: remain in the current checkout and leave final branch/PR work to the wrapper. Fix the root cause minimally. Do not weaken, skip, or narrow any gate. Use focused checks while repairing; the deterministic wrapper will restart at prepare, run the complete build, and run the full supported-family certification before it can publish success. Leave changes local. Do not push, open a PR, or use GitHub credentials.
```
ndizazzo: This points the agent at `ci/llama-canary/agent-repair-prompt.md` and then spends three sentences telling it to ignore that file's mode, branch, and publication instructions. That file is only referenced from this script now, so nothing stops us from just updating it. A self-contradicting prompt is the kind of thing `glm-5.3-flash` handles badly, and the mistake it would make is the publication behavior we deliberately took away.

scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,442 @@
...
   if [[ "$outcome" == "certified" ]]; then
     CERTIFIED_SHA="$PUBLISHED_SHA"
   fi
-  # Push failures are almost always a token-identity permission gap (seen
-  # live: 403 "denied to i386" because the PAT account lacked repo write).
-  # Surface a precise, actionable message instead of a bare git error, and
-  # never leak the token in the hint.
-  if ! git push "$(repair_remote)" "+HEAD:refs/heads/${BRANCH}" 2> >(redact_token >&2); then
-    echo "ERROR: could not push ${BRANCH}. If git reported 403/denied above, the identity behind CANARY_REPAIR_TOKEN lacks write access to ${GITHUB_REPOSITORY:?}: grant that account Contents+Pull requests write (or mint the PAT from an account that has it) and update the secret." >&2
+  if ! git push "$(repair_remote)" "HEAD:refs/heads/${BRANCH}" 2> >(redact_token >&2); then
```
ndizazzo: The stderr redaction is a nice touch (same for `2> >(redact_token >&2)` on `gh pr create`), but the token still lands in argv: `x-access-token:<PAT>@github.com` is readable via `ps` for the life of the push. On a persistent self-hosted runner that's shared with whatever else is scheduled there. An env-sourced credential helper would keep it out of the process table.

scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,442 @@
...
-  fi
-  if scripts/skippy-family-battery.sh >"$BATTERY_LOG" 2>&1; then
-    tail -n 2 "$BATTERY_LOG"
-    return 0
+    comment="**Uncertified terminal state.** The internal deadline or repair-turn limit stopped the \`${FAILED_PHASE}\` phase. This draft preserves the final attempted bytes and must not merge until the complete state machine passes."
   fi
-  tail -n 2 "$BATTERY_LOG"
-  return 1
+  gh_repair gh pr comment "$pr" --body "$comment" >/dev/null 2>&1 || true
+  echo "terminal canary PR #${pr}: ${outcome}; branch=${BRANCH}; head=${PUBLISHED_SHA}"
```
ndizazzo, line 432: `run_prepare || phase_status=$?` puts the whole function body in an `||` context, so `errexit` is off for everything it calls. It's correct today because every substantive call ends in `|| return 1`, but the first line added without one will fail silently and report success. Same for the build and certify calls below. At minimum this deserves a comment saying the `|| return 1` on each step is load-bearing.

.github/workflows/llama-upstream-canary.yml
```diff
@@ -139,64 +131,64 @@ jobs:
...
+          CANARY_PUBLISH_RESERVE_SECONDS: "1800"
+          FAMILY_BATTERY_RUN_ID: ${{ github.run_id }}-${{ github.run_attempt }}
+          UPSTREAM_SHA_INPUT: ${{ steps.sha.outputs.new_sha }}
+        run: scripts/llama-canary-agent-repair.sh
+
+      - name: Upload changed-pin canary evidence
+        if: ${{ !cancelled() && steps.sha.outputs.changed == 'true' }}
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        with:
+          name: llama-family-battery-${{ github.run_id }}-${{ github.run_attempt }}
```
ndizazzo: This name matches `Upload supported-families battery evidence` exactly, and `upload-artifact@v4` hard-fails on a duplicate name in the same run. They only miss each other because `prepare` is skipped on changed pins, so the safety is accidental. Let's give this one its own name (something like `llama-canary-changed-pin-...`).

scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,442 @@
...
+cd "$ROOT"
+
+OLD_SHA="$(tr -d '[:space:]' < third_party/llama.cpp/upstream.txt)"
+PIN_FILE="$ROOT/third_party/llama.cpp/upstream.txt"
 AGENT_MODEL="${CANARY_AGENT_MODEL:-zai-coding-plan/glm-5.3-flash}"
 MAX_REPAIR_TURNS="${CANARY_REPAIR_MAX_TURNS:-2}"
-BRANCH="llama-canary/patch-queue-fix"
-BATTERY_LOG="$ROOT/.deps/llama-canary-repair-battery.log"
-PIN_FILE="$ROOT/third_party/llama.cpp/upstream.txt"
+REPAIR_BUDGET_SECONDS="${CANARY_REPAIR_BUDGET_SECONDS:-41400}"
```
ndizazzo, line 30: Not a change request, just noting it: `run_bounded` hands each command the entire remaining budget, so a full `llama-bump` cohort will eat most of the 660 work-minutes. After a certify failure there's realistically no room left for a repair turn plus a rebuild plus a re-certify. `MAX_REPAIR_TURNS=2` ends up being prepare/build-only in practice, even though the docs read like certify repair is live.

scripts/llama-canary-agent-repair.sh
```diff
@@ -1,681 +1,442 @@
...
   fi
   echo "preflight: repair token identity '${login}' verified read+write on ${GITHUB_REPOSITORY}"
 }
 
-cd "$ROOT"
+repair_remote() {
+  echo "https://x-access-token:${CANARY_REPAIR_TOKEN}@github.com/${GITHUB_REPOSITORY:?}.git"
+}
+
+redact_token() {
```
ndizazzo, line 126: Small one: the token is interpolated unescaped into the `sed` pattern. Fine for `[A-Za-z0-9_]` PATs, but a token containing `/` would break the expression and pass through unredacted.

ndizazzo (commented):
Follow-up on my approval with the specifics inline. Nothing here blocks the merge, but the branch pile-up one gets expensive fast on a single certify runner.

scripts/llama-canary-agent-repair.sh
i386: Fixed in 718358937. Scheduled changed-pin runs now query open llama-canary/repair-* PR bodies for the exact full Candidate pin SHA and skip the expensive state machine when a match exists. Manual dispatch remains available. Covered by the canary workflow contract tests; just ci-validate passed 1,066 tests with 8 skipped.

.github/workflows/llama-upstream-canary.yml
i386: Fixed in 718358937. The wrapper now validates LLAMA_STAGE_BUILD_DIR, HF_CACHE, GITHUB_REPOSITORY, and CANARY_REPAIR_TOKEN before phase execution. The final workflow summary branches on duplicate lookup and changed-canary outcomes, and explicitly states that preflight failure can occur before a branch or PR exists. The summary uses the repository cancellation-safe !cancelled() gate.

scripts/llama-canary-agent-repair.sh
i386: Fixed in 718358937. LLAMA_STAGE_BUILD_DIR and HF_CACHE are checked together in the upfront environment preflight, and the later fatal parameter expansions were replaced with the already-validated variables. Executable contract tests cover each missing environment variable.

scripts/llama-canary-agent-repair.sh
i386: Fixed in 718358937. Terminal PR body generation now runs scripts/summarize-llama-upstream.sh for the old and candidate SHAs, includes the upstream diffstat and commit list, and falls back to an explicit unavailable message if summary generation fails. The unrelated validation checklist is excluded so failed terminal drafts stay truthful.

scripts/llama-canary-agent-repair.sh
i386: Fixed in 718358937. The repair runbook now describes the single wrapper-owned prepare -> build -> certify -> publish state machine, focused repair verification, local-only agent authority, and run-specific terminal publication. The wrapper prompt no longer overrides obsolete mode, branch, or publication instructions.

scripts/llama-canary-agent-repair.sh
i386: Fixed in 718358937. Git push now uses a credential-free HTTPS URL plus a run-scoped GIT_ASKPASS script that reads CANARY_REPAIR_TOKEN from the environment, so the PAT is absent from argv and the remote URL. The agent environment still strips all GitHub credentials.

scripts/llama-canary-agent-repair.sh
i386: Addressed in 718358937. The state-machine loop now documents that phase functions run in an OR context with errexit disabled, that explicit return propagation on every fallible phase command is load-bearing, and that the final command supplies the function status. A focused contract test preserves this warning.

.github/workflows/llama-upstream-canary.yml
i386: Fixed in 718358937. Changed-pin evidence now uploads as llama-canary-changed-pin-<run>-<attempt>, distinct from the unchanged-pin llama-family-battery artifact. The workflow contract test rejects the old duplicate name.

scripts/llama-canary-agent-repair.sh
i386: No code change was requested here. The shared 690-minute work budget remains intentionally bounded, and the updated agent runbook now directs repair turns to focused verification while the wrapper retains ownership of the mandatory complete prepare, build, and certification gates. A separate budgeting redesign would need its own runtime evidence.

scripts/llama-canary-agent-repair.sh
i386: Fixed in 718358937. Redaction now performs literal Python string replacement instead of interpolating the PAT into a sed regular expression. The executable contract test uses a token containing slash, brackets, dollar, dot, star, and backslash and verifies exact redaction.

PR #1731, opened 2026-09-09
ci: stop duplicate release tag runs

### LINKED ISSUE #1597: Every release tag leaves a permanently-red Release run from a duplicate ref trigger
**Type:** Bug — CI hygiene / process
**Priority:** P3
**Effort:** XS (under an hour)
**Found in:** v0.76.0-rc9 release validation (CI review)

## Problem

Every release leaves a permanently-red Release run attached to its ref, so "is this release green?" cannot be answered at a glance.

For rc9:

- [Run 33549644311](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33549644311) — ref `v0.76.0-rc9`, **failure**. Died in *Resolve release metadata* on the dirty-tree guard (the step logs `M crates/*/Cargo.toml`, `M docs/…`, etc., then exits 1). All 27 downstream jobs skipped.
- [Run 33549875856](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33549875856) — ref `main`, **success**. 74 jobs, this is the run that actually built and published the release.

Anyone checking the tag sees the red one.

## Why it happens

The Release workflow triggers on a ref named after the release, and by the time it runs, the version-bump step has already dirtied the tree — so its own guard fires. The real release is driven separately from `main`.

## Impact

Cosmetic but corrosive: it trains everyone to ignore a red Release run, which is exactly the signal we would want to be loud the day a release genuinely fails. It also cost me time during validation deciding whether rc9 had actually shipped cleanly.

## Suggested direction

Pick one:

- Stop the Release workflow from triggering on the release ref at all, so only the `main`-driven run exists; or
- Make the metadata step exit successfully with a skip when it detects it is the duplicate ref-triggered invocation, so the run shows as skipped rather than failed.

Either way the goal is that a red Release run means something.

ndizazzo:
## Original problem

The Release workflow also ran on every `v*` tag push. Normal releases are driven by the successful `workflow_dispatch` run from `main`, leaving a duplicate tag-ref run that could fail in metadata and make the release appear red.

## Diagnostics

Issue #1597 and live run evidence confirmed the duplicate pair: tag push run [33549644311](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33549644311) failed in `Resolve release metadata`, while main dispatch run [33549875856](https://github.com/Mesh-LLM/mesh-llm/actions/runs/33549875856) successfully published the release.

## Fix

- Remove the obsolete `push` tag trigger from `.github/workflows/release.yml`.
- Keep the dispatch-driven release and its generated immutable tag creation unchanged.
- Update the CI inventory and release documentation to describe the dispatch-only path.
- Add a focused contract test preventing the tag trigger from returning.

Fixes #1597

## Validation

- `python3 -m unittest scripts.tests.test_release_workflow_artifacts -v` — 26 passed.
- `python3 -m unittest scripts.tests.test_release_script -v` — 4 passed.
- `actionlint -config-file .github/actionlint.yaml` — passed.
- `just check-release` — passed.
- `git diff --check` — passed.
- `just ci-validate` reached the full 936-test suite but is blocked by the macOS host system Python 3.9/Bash 3.2 lacking repository-required features, plus unrelated environment/fixture failures; changed-file-focused checks pass.

## Summary by CodeRabbit

* **Changes**
  * Release workflows now run only through manual dispatch from the Actions interface or supported release command.
  * Releases initiated this way create the release-specific tag automatically.
  * Pushing version tags no longer starts a release.
  * Preflight checks now verify access and write permissions before downstream packaging.
  * Canary dispatch behavior remains unchanged.

* **Documentation**
  * Updated release guidance for dispatch-only publishing and credential verification.

* **Tests**
  * Added coverage for dispatch-only releases and target repository permission checks.

### DIFF
--- a/.agents/skills/manage-ci/references/current-inventory.md
+++ b/.agents/skills/manage-ci/references/current-inventory.md
@@ -21,7 +21,7 @@ Read it with `../SKILL.md` and `ci/ci.md` before editing CI.
 | `main_windows.yml` (`Main · Windows`) | push to `main` | Exhaustive main planning plus the same-commit reusable Windows lane |
 | `ci.yml` | `workflow_call` only | Temporary inert shim for the former main ingress filename; pending protected-main runner-contract update; no push trigger or dispatch |
 | `ci-control.yml` (`CI · Manual Full`) | dispatch on default branch | Explicit operator-only full plan, bounded lane dispatch and correlated diagnostic checks |
-| `release.yml` | release tags, dispatch | Canonical version synchronization, release-only signing, assets, publication, and a preflighted downstream `mesh-packaging` dispatch |
+| `release.yml` | dispatch on the default branch | Canonical version synchronization, release-only signing, assets, publication, and a preflighted downstream `mesh-packaging` dispatch |
 | `website-pages.yml` | main website paths, dispatch | Public website deployment |
 | `pr_cleanup.yml` | PR close, dispatch | Positively matched cleanup only |
 | `pr_auto_assign.yml` | PR lifecycle | Metadata only |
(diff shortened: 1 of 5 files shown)


michaelneale (dismissed):
Nice catch

PR #1735, opened 2026-09-09
ci: publish SDK smoke cache policy

### LINKED ISSUE #1381: CI: SDK smoke has no cargo cache — 3m59s of cold compile for a 1.32s test

ndizazzo:
## Summary

- mark the `sdk` CI slice as `pr-isolated` in `ci/slices.yml`
- enable the existing exact, main-seeded Rust SDK Cargo/target cache policy in the protected planner output
- preserve the existing hosted-only credential-smoke workflow and cache-key/save policy from #1556

This is the catalog-only follow-up required by #1556: the operational `Swatinem/rust-cache` integration is already present, but the protected slice catalog still advertises `cache_mode: none`.

## Validation

- focused CI-contract suite: 94 tests passed under ephemeral Python 3.12 + PyYAML
- `actionlint .github/workflows/sdk-smoke.yml` passed
- planner probe: `sdk` resolves to `pr-isolated` for PRs, `trusted-readwrite` for main, and `trusted-readonly` for manual-full
- `git diff --check` passed

The full local `just ci-validate` suite was also attempted; this macOS checkout has unrelated host failures from missing PyYAML in the default Python environment, a protected `/usr/sbin` permission probe, and an existing llama-canary fixture failure.

Fixes #1381

## Summary by CodeRabbit

* **Chores**
  * Improved SDK build cache isolation for pull-request builds.

### DIFF
--- a/ci/slices.yml
+++ b/ci/slices.yml
@@ -237,7 +237,7 @@
     {"id": "platform-checks", "kind": "platform-checks", "depends_on": [], "runner_role": "platform-build", "cache_mode": "none"},
     {"id": "product-smoke", "kind": "product-smoke", "depends_on": ["runtime-product"], "runner_role": "linux-build-4", "cache_mode": "none"},
     {"id": "model-download", "kind": "model-download", "depends_on": [], "runner_role": "linux-build-4", "cache_mode": "none"},
-    {"id": "sdk", "kind": "sdk", "depends_on": ["runtime-product"], "runner_role": "platform-build", "cache_mode": "none"},
+    {"id": "sdk", "kind": "sdk", "depends_on": ["runtime-product"], "runner_role": "platform-build", "cache_mode": "pr-isolated"},
     {"id": "runner-contract", "kind": "runner-contract", "depends_on": [], "runner_role": "linux-plan", "cache_mode": "none"}
   ]
 }


ndizazzo:
CI management note: this PR currently fails before fan-out because the protected planner checkout rejects the `ci/slices.yml` change (`source slice catalog differs from the protected planner catalog`). This is the repository's intentional trust boundary. PR #1556 documents the established path: a separate maintainer merge must update the protected catalog, after which this branch can rebase. I'm leaving the guard intact and will continue monitoring the other PRs.

Issue #1651, opened 2026-09-04
Add and benchmark a bounded host-RAM L2 KV cache tier


i386:
## Context

PR #1632 adds resident L1 reuse and durable disk L3, but no bounded host-RAM L2. L2 is especially important to test on the CUDA machine where device and host memory are not unified: a host copy may avoid disk and cold prefill while still paying PCIe transfer/import cost.

The value will differ sharply across discrete CUDA, coherent CPU/GPU systems, and unified-memory Macs, so this must be measured per hardware class rather than assumed.

Coordinated in Buzz channel `#skippy-kv` and tracked as an expanded optimization candidate for #1632.

## Proposed direction

- Reuse the same immutable exact segments and manifest/model identity as L3.
- Give L2 its own hard byte budget, status, accounting, pins/holds, admission, promotion, eviction, prune, and clear semantics.
- Promote repeated/high-value L3 hits rather than copying every disk entry into RAM.
- Route by measured `queue + transfer + import + suffix-prefill` versus L3 restore and cold prefill.

## Checklist

- [ ] Define L1/L2/L3 ownership, lookup order, promotion/demotion, and duplicate-byte accounting.
- [ ] Add typed configuration and status reporting with safe defaults; keep the experiment opt-in until promoted.
- [ ] Bound host allocations, pinned memory, transfer buffers, and background work under concurrency.
- [ ] Decide whether pageable, pinned, huge-page, or backend-specific host memory is beneficial from measurement.
- [ ] Preserve exact identity, transactional restore, model unload, cancellation, and cold fallback.
- [ ] Add pressure, eviction, concurrent fill, prune/clear, restart, and device-loss tests.
- [ ] Run matched A/B tests on the discrete CUDA machine, then unified-memory Metal and other supported classes.
- [ ] Report TTFT, transfer/import time, PCIe bandwidth, CPU overhead, peak host/device bytes, hit rate/value, goodput, and p99 decode latency.

## Acceptance gate

Promote per hardware class only when bounded L2 improves end-to-end TTFT or goodput over both L3 and cold prefill for a defined workload, without memory-pressure instability, incorrect outputs, or a regression beyond #1632 latency gates. A neutral or negative CUDA result must remain opt-in/off rather than becoming a default.

PR #1749, opened 2026-09-10
feat(skippy-cache): bounded host-RAM L2 tier prototype (#1651)

i386:
First slice of #1651, stacked on #1736 (`7247a397a`). This establishes the bounded host-RAM L2 ownership and integrity contract without touching the request restore path.

## What

- **`skippy-cache::l2`** — a bounded host-RAM tier over immutable, content-addressed segment handles:
  - Entries reuse L3 namespace and token-path coordinates plus model and exact-state identities.
  - Prefix growth shares segment handles, so the byte budget charges distinct physical segment bytes once while stats expose both physical and logical footprint.
  - Admission validates the whole-payload digest, each segment digest, manifest tiling, indices, offsets, payload geometry, and existing segment identity before mutating the pool.
  - Admission and same-key replacement are journaled transactions. Incoming handles remain protected through replacement and eviction; an admission that cannot fit rolls back completely with `ProtectedOvercommit`.
  - Deterministic ownership accounting assigns every pooled segment to exactly one live entry and recomputes after admission, eviction, removal, corruption cleanup, and rollback. Eviction reports physical bytes freed separately from shared bytes retained.
  - `peek` is side-effect free. A successful `get` advances recency; absent or corrupt entries become cold misses without heating the LRU.
  - Budget enforcement is deterministic LRU with stable key tie-breaking. Empty, oversized, malformed, corrupt, and unknown-kind payloads are refused without leaving partial state.
- **`skippy-bench l2-tier`** — compares cold L3 fill with warm L2 lookup on the identical packed entry. Both arms start before lookup and end at usable materialized bytes, with byte equality required before timing is reported.
  - User-supplied store roots are refused if they already exist. Benchmark cleanup only removes roots it created and sentinel-marked.

Request-path promotion, demotion, server integration, and discrete-CUDA qualification remain later #1651 slices.

## Bench numbers (M1 Ultra, release build, 50 pairs, head `f678ce337`)

Both arms use the same nothing-to-usable-bytes boundary. L3 includes index lookup, segment reads, assembly, and digest verification. `l2_lookup_to_usable_bytes_ns` includes `L2Tier::get`, handle assembly, and materialization; handle-only lookup is reported separately. These figures supersede all earlier L2 numbers in this PR and its comments.

| workload | L3 fill p50 | L2 lookup-to-usable p50 | speedup |
|---|---:|---:|---:|
| 1,893 tokens × 512 B/token (~0.9 MiB) | 1.412 ms | 26.083 µs | 54.13x |
| 3,994 tokens × 4 KiB/token (~15.6 MiB) | 20.258 ms | 536.250 µs | 37.78x |

Both runs produced 50/50 L2 hits with zero misses or evictions. This is a unified-memory microbenchmark; request-path TTFT/goodput and discrete-CUDA transfer behavior are not measured here.

## Validation

Validated in a clean worktree with local branch, remote branch, and GitHub PR head all at `f678ce337236e18613d434713db107465bcc65ef`:

- `cargo test -p skippy-cache`: 150 passed, 1 ignored, 0 failed; doc tests passed
- `cargo test -p skippy-bench`: 85 passed, 0 failed
- `cargo clippy -p skippy-cache -p skippy-bench --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `git diff --check origin/scama/skippy-l3-streaming-restore...HEAD`: clean
- console-print consistency check: clean
- L2 benchmark smoke: 2/2 hits, byte equality preserved, owned temporary store removed


i386:
Review round 2 pushed at `608ec6c38` (single commit on `247a4a5db`, still stacked on `scama/skippy-l3-streaming-restore`).

**Atomic admission.** `admit` installs or pins every incoming segment handle (a `protected` set) *before* the replaced entry is released and before eviction runs. An identical-wire same-key re-admit can no longer delete its own handles (regression: three rounds of re-admit must serve the full wire with `segments=4, bytes=64` throughout), and an admission that shares its eviction victim's segments keeps them (regression: budget `total-1` forces the victim out; the shared prefix must survive, `freed` accounting exact). Retained shared bytes transfer to the new entry's budget charge.

**Content-verified pool reuse.** `validate_layout` BLAKE3-hashes every segment slice against its exact wire range before any pool mutation; the layout must tile contiguously with `kv+recurrent == total`; `from_manifest` now checks `index` and `offset` per segment. Same-digest/different-bytes offers are refused (`SegmentDigestMismatch` / new `ConflictingSegment`) unless they replace the same key — a referenced content-addressed handle is never swapped underneath a live entry (regression included; the live victim's handle is verified intact).

**Corrupt-entry `get`.** A missing handle is a cold miss: entry + surviving handles removed, LRU untouched, `misses` counter incremented (it was previously never incremented anywhere; absent-key misses count too).

**Bench.** Both arms timed through the same usable-bytes boundary; `l2_lookup_to_usable_bytes_ns` is the headline with `l2_handle_lookup_ns` separate. Store root is bench-owned: pre-existing paths refused (never deleted), sentinel-marked root, cleanup refuses unmarked directories (test covers refusal + survival of user files).

Fresh numbers (M1 Ultra, release, 50 pairs, 1,893 tokens × 512 B): L3 fill p50 1.31 ms vs L2-to-usable-bytes p50 18.5 µs (~71x), handle-only p50 3.0 µs, one-time admission hash 1.2 ms. (Single-segment dedup in the pool: 15 segments / 969,216 distinct bytes.)

Validation at `608ec6c38`: `cargo test -p skippy-cache` **144 passed / 1 ignored / 0 failed** (+5 regressions), `cargo test -p skippy-bench` **85 passed / 0 failed**, `cargo clippy -p skippy-cache -p skippy-bench --all-targets -- -D warnings` clean, `cargo fmt` clean.

i386:
Exact-head benchmark rerun at `f678ce337236e18613d434713db107465bcc65ef`, with local HEAD and `origin/jy/skippy-l2-host-tier` verified equal in the same shell.

The corrected L2 timer now begins before `L2Tier::get`, so `l2_lookup_to_usable_bytes_ns` includes lookup, handle assembly, and materialization. The L3 arm includes lookup, disk assembly, and verification to the same usable-byte boundary.

M1 Ultra, release build, 50 matched pairs:

| Shape | Payload | L3 fill p50 | L2 lookup-to-usable p50 | L2 handle lookup p50 | p50 ratio |
| 1,893 tokens × 512 B | 969,216 B | 1.412 ms | 26.083 µs | 3.791 µs | 54.13x |
| 3,994 tokens × 4 KiB | 16,359,424 B | 20.258 ms | 536.250 µs | 95.833 µs | 37.78x |

The one-time admission hash was 1.503 ms and 33.236 ms respectively. Both runs had 50/50 L2 hits, zero misses, and zero evictions.

These supersede the earlier figures that excluded lookup. They validate the standalone tier operation on unified-memory Metal hardware. They do not establish the #1651 end-to-end promotion gate: the slice is not wired into the request path, and discrete-CUDA PCIe/import, concurrency, memory pressure, and TTFT/goodput evidence remain later work.

danielwinterw (approved):
Approving on the delta.

The hard part here is admission atomicity and the accounting stays coherent through it. Reserving the incoming handles before releasing the replaced entry, pinning them in `protected_set` so neither release nor eviction can drop bytes the transaction is about to reference, and journalling every pool mutation so a `ProtectedOvercommit` rolls back exactly — that is the right shape, and the same-digest-different-content case is handled properly (allowed only when replacing the same key, refused otherwise so a live entry never has a handle swapped underneath it). `validate_layout` hashing every segment against its exact wire range before touching the pool is the correct ordering for a content-addressed store. `peek` staying off the LRU clock while `get` bumps it is right for prefix probing. The corruption path in `get` — missing handle downgrades to a miss, drops the entry, leaves recency alone — fails in the safe direction.

Note on CI: `Rust tests (batch-1)` is red with three failures in mesh peer_state and moa_gateway self_fill. Nothing in this diff reaches them; they are the same three failing on #1747, which shares the base, and they originate in #1736.

Nice-to-haves:

1. `recompute_all_charges` walks every pooled segment × every entry, and `segment_digests()` is itself O(D²) because it dedups with `Vec::contains`. It is called on every admission, every removal, and once per victim inside `evict_to_limit`, so a single admission that evicts V entries is roughly O(V · S · E · D²). Fine for a prototype driven by the bench harness, but this wants a reverse index (digest → referencing keys) before the tier goes anywhere near the request path — worth a TODO pointing at that so the next slice does not inherit it silently.

2. `L2Hit::wire_view` falls back to `Arc::new(Vec::new())` when a digest is not in `self.segments`, then clamps `from`/`to` to that zero length. The resulting `CacheBytes` still reports `end - start` as its length, so a missing handle would yield a short payload claiming full length rather than an error. `get` makes this unreachable today, but silent truncation is an unpleasant failure mode for a state-restore path — a `debug_assert!` or an `Option` return would be cheap insurance.

3. On the `ProtectedOvercommit` rollback path, `stats.evictions.fetch_sub(rolled_back_evictions)` is not atomic with the `fetch_add`s inside `evict_to_limit`, so a concurrent `stats()` can observe evictions that were rolled back. Only telemetry, but it makes the counter non-monotonic, which is surprising for anything scraping it.

i386:

Issue #1652, opened 2026-09-04
Version KV segments for native passthrough and CacheGen


i386:
## Context

PR #1632 binds cache entries to exact numerical and model identity. PR #1750 adds an explicit payload codec identity, manifest v3, a legacy raw-v2 path, and fail-closed checks for unsupported codec or manifest versions.

The remaining scope is deliberately limited to two representation paths:

1. **Native passthrough:** persist and restore the KV representation already used by the active runtime, without transcoding it into a separate storage format.
2. **CacheGen:** evaluate a specialized, opt-in codec for cold and network tiers against the native baseline. [CacheGen](https://doi.org/10.1145/3651890.3672274) reports 3.5–4.3× smaller KV and 3.2–3.7× lower fetch/process delay; those results must be reproduced in Mesh's end-to-end workloads before promotion.

General-purpose compression, separate F16-to-Q8/Q4 transcoding, KVTC, and cross-model KV mapping are outside this issue.

CacheGen availability is independent of the GGUF/model weight quantization. For runtime KV storage, CacheGen must cover every K/V cache type the product lets a user select, including mixed K/V selections where the runtime accepts them. The current supported representation set is F32, F16, Q8_0, and Q4_0; the adapter contract must admit future user-selectable KV types without changing the portable CacheGen stream format. Each type may specialize native load/unpack and destination pack/store code, while sharing the CDF and arithmetic-coder core.

Coordinated in Buzz channel `#skippy-kv` and tracked under roadmap #1732.

## Implementation constraints

Mesh's CacheGen implementation must be Rust-owned and must not require Python, PyTorch, Candle, or the LMCache runtime. LMCache's Apache-2.0 implementation and tests are reference material for the algorithm and fixtures.

Use ordinary Rust for the representation, calibration metadata, bounded bitstream, validation, and deterministic CPU reference. Evaluate [CubeCL](https://github.com/tracel-ai/cubecl) for the parallel quantization, CDF, and arithmetic encode/decode kernels because its Rust kernel source targets CUDA, HIP/ROCm, Metal through wgpu, and CPU. Pin the selected revision and isolate it behind an internal backend trait because CubeCL's public API is still evolving.

[...]

Issue #1650, opened 2026-09-04
Add benefit-per-byte KV cache admission and eviction policy


i386:
## Context

PR #1632 currently provides reference-aware LRU and hard-budget enforcement. LRU alone does not distinguish an expensive repeatedly reused prefix from a large one-shot prefix that should never have consumed SSD bandwidth.

[Mooncake](https://arxiv.org/abs/2407.00079) reports highly skewed production reuse, with many blocks never reused and hot blocks accessed repeatedly. [Preble](https://arxiv.org/abs/2407.00023) co-optimizes prefix reuse and serving load rather than using locality alone.

Coordinated in Buzz channel `#skippy-kv` and tracked as an expanded optimization candidate for #1632.

## Proposed direction

Use measured benefit per exclusive physical byte:

```text
reuse_probability * max(cold_prefill_cost - restore_cost, 0)
-----------------------------------------------------------
                 exclusive_physical_bytes
```

Credit shared segments correctly, place new/large entries in probation, and persist only after a second-hit or equivalent value signal. Tier routing should compare measured `queue + restore + suffix-prefill` cost with cold prefill rather than treating every hit as valuable.

## Checklist

- [ ] Define entry and shared-segment accounting without double-counting physical bytes.
- [ ] Add bounded reuse/value statistics and decay so stale historical popularity does not pin data forever.
- [ ] Implement probation and second-hit/value-triggered persistence.
- [ ] Define admission, promotion, demotion, and eviction scores with deterministic tie-breaking.
- [ ] Preserve hard budgets, minimum-free reserve, active pins/holds, exact identity, and cold fallback.
- [ ] Expose policy decisions and reasons without logging prompt content or stable content fingerprints.
- [ ] Replay realistic turn-growth, one-shot, Zipf/hotset, mixed-size, and concurrency traces.
- [ ] Compare LRU and candidate policies on hit value, bytes written, cache-hit TTFT, goodput, and tail latency.

## Acceptance gate

Adopt a policy only when the same-capacity matched replay beats current LRU on saved prefill time or goodput, reduces or justifies SSD writes, remains stable under adversarial churn, preserves exact output, and does not regress the #1632 latency gates.

Issue #1649, opened 2026-09-04
Stream and pipeline local KV state export and restore end to end


i386:
## Context

PR #1632 stores exact state as verified logical segments, but the current spill path builds full `KV + recurrent + wire` buffers and restore reassembles an entire payload before import. That creates avoidable allocations, copies, peak memory, and serialized I/O/compute.

[LMCache](https://arxiv.org/abs/2510.09665) emphasizes batched movement and compute/I/O pipelining; [Mooncake](https://arxiv.org/abs/2407.00079) overlaps layer-wise KV movement with prefill/decode.

This issue is deliberately local: it owns L3-to-runtime streaming and block/range-native export/import. Networked phase-disaggregated handoff remains in #1427.

Coordinated in Buzz channel `#skippy-kv` and tracked as an expanded optimization candidate for #1632.

## Proposed direction

- Export directly into bounded background-owned blocks instead of constructing multiple whole-state copies.
- Read, verify, decode, and import segments incrementally with double-buffered read-ahead.
- Add only the native block/range/page import-export ABI needed for exact local streaming.
- Preserve transactional visibility: decode must never observe a partial restore.
- Keep recurrent/SSM and other non-append-only state exact and commit it at the correct boundary.

## Checklist

- [ ] Measure and document the current copy/allocation graph for dense KV and hybrid/recurrent families.
- [ ] Define bounded block/range export and import contracts plus cancellation and backpressure.
- [ ] Stream digest verification and payload-digest verification without trusting partial state.
- [ ] Pipeline disk read/decode/import and export/encode/write while bounding worker queues.
- [ ] Add safe abort, short read, checksum failure, cancellation, model unload, and runtime restart behavior.
- [ ] Preserve single-flight, pin/hold, eviction, and atomic-manifest guarantees from #1632.
- [ ] Add exact continuation tests across dense, SWA-supported, and recurrent/hybrid families.
- [ ] Benchmark TTFT, export latency, restore latency, peak RSS, allocated/copied bytes, throughput, and p99 decode-event latency against current #1632.

## Acceptance gate

Promote only if the release-build path improves end-to-end restore/spill performance on the same workload, lowers or bounds peak transient memory, preserves exact outputs, and does not regress the existing disk-off p99 decode-event latency gate beyond 5%.

PR #1750, opened 2026-09-10
feat(skippy-cache): raw codec identity in L3 manifests (#1652 first slice)

i386:
First slice of #1652 (codec versioning). Introduces an **explicit raw codec identity + version** in the L3 handoff manifest, enforces it by **manifest version** (not defaultable), and rejects unsupported codecs at **every load path** before segments are read or the LRU heats.

Stacked on `scama/skippy-l3-streaming-restore` (#1736 head `7247a397a`). Payload-level, raw-only. No Q8/Q4 data path, no request-path performance work, no #1650/#1651 changes. Per-segment/mixed codec identity is later #1652 work.

## What it does
- Adds `PayloadCodec { name, version }` with `CODEC_RAW` / `CODEC_RAW_VERSION`. `raw` (uncompressed exact state) is the only implemented representation — segment bytes are unchanged.
- **Version-enforced identity (not a serde default).** `MANIFEST_VERSION` is bumped to **3**, which *requires* an explicit codec. `LEGACY_MANIFEST_VERSION = 2` is decoded as raw through an explicit legacy path. `HandoffManifest.codec` is `Option<PayloadCodec>` purely so a v3 manifest with the codec **stripped** is detectable and **rejected**, rather than silently reinterpreted as raw (the whole-payload digest matches the stored bytes, so a serde default would bypass the gate — the downgrade hole this fixes).
- **Central capability gate in `decode_manifest`.** Every load path enforces codec support before returning a manifest:
  - startup `reconcile_startup` → `validate_committed_manifest` → `load_manifest` → **quarantines** an unsupported manifest;
  - `manifest_for_prefix` **prunes the bad prefix link** and returns a miss, so `locate_longest` **falls back to the shorter supported prefix**;
  - direct `load_manifest` fails.
- Writer/in-memory guards remain: `try_commit` never persists an unassemblable manifest, and `assemble` refuses one before reassembly.
- The identity travels in the portable manifest, so it is carried by the network handoff as well as the disk tier.

## Tests (`skippy-cache`)
- `manifest_stamps_explicit_raw_codec_and_round_trips`
- `legacy_manifest_without_codec_field_reads_and_assembles_as_raw`
- `stripping_codec_from_current_version_rejects_but_legacy_v2_reads_as_raw` (downgrade regression)
- `on_disk_unsupported_codec_fails_direct_load`
- `startup_reconciliation_quarantines_unsupported_codec_manifest`
- `locate_longest_skips_unsupported_codec_and_falls_back_to_shorter` (tier fallback)
- `unknown_codec_is_refused_at_commit_and_leaves_no_manifest`, `unknown_codec_is_rejected_before_assembly`, `unknown_raw_codec_version_is_rejected_at_commit_and_assembly`, `codec_gate_does_not_mask_payload_corruption`

## Validation
`cargo test -p skippy-cache`: **131 passed / 1 ignored / 0 failed**. `cargo clippy -p skippy-cache --all-targets -- -D warnings`: clean. `cargo fmt`: clean. (`skippy-bench` does not depend on `skippy-cache` on this base, so it is unaffected.)


i386:
Exact-head code review and the required cache performance gate now pass at `5effeffb1a64e8b382a8a504cb75a9d5c8141ffc`.

Compatibility review:
- manifest v3 requires an explicit supported codec;
- legacy v2 is normalized through the explicit raw path;
- unsupported codec or manifest versions are rejected before commit, lookup, reconciliation, assembly, or segment access;
- startup quarantine and longest-supported-prefix fallback are covered by on-disk regressions.

Validation:
- `cargo test -p skippy-cache`: 133 passed, 1 ignored;
- `cargo clippy -p skippy-cache --all-targets -- -D warnings`: clean;
- `cargo fmt --all --check` and `git diff --check`: clean.

Matched performance gate:
- exact release host SHA-256: `c9daa526993f3ab84ef5f3e7d11519758a47c5bddffd7dc7593ea886f9476ddb`;
- matching locally packaged macOS Metal runtime, Skippy ABI `0.1.53`, primary library SHA-256 `fc8de19b58365b53255d1be948a3ce22b3c4d4119018f64e3a461ca287136a3a`;
- same M1 Ultra, SmolLM2 Q8 model bytes, 3,994-token workload manifest, disk-cache settings, and three-run restart method as accepted #1649 evidence;
- first-restore TTFT: 109.274 / 109.300 / 106.378 ms, median **109.274 ms**;
- baseline #1649 treatment median: **109.5 ms**;
- every run restored 3,993 / 3,994 prompt tokens (99.975%), with zero failed requests;
- median restore decode rate: 200.879 tok/s; median resident-warm TTFT: 18.179 ms.

The roughly 0.2% lower restore median is within normal run noise. No material restart, reuse, or throughput regression is visible from the manifest-version and codec-validation change.

i386:

PR #1758, opened 2026-09-10
feat(ci): regroup release notes from conventional commits

ndizazzo:
## Original problem

The release workflow publishes with GitHub-generated release notes, so every release arrives as a single flat `## What's Changed` list in merge order. v0.76.0 had 272 entries. Nothing separates a security fix from a llama.cpp pin rebase, and the one thing you must not do while reformatting — quietly drop somebody's PR — had nothing guarding against it.

## Diagnostics

Reformatted the [v0.76.0 notes](https://github.com/Mesh-LLM/mesh-llm/releases/tag/v0.76.0) by hand first. The work split into a mechanical half (move 272 lines without losing one) and a judgement half (which section does this entry belong in), and measurement showed how far each half can actually go:

- **78 of 272 entries (~29%) carry no conventional-commit prefix**, so a naive classifier has nothing to read.
- Of the **13 security-relevant entries, prefixes place zero correctly**: 6 are `fix:`, 1 is `feat(...)`, 6 have no prefix. Every one would land in Fixed, Added, or nowhere.

So the deterministic pass has to be authoritative but must never guess, and the judgement pass has to be optional.

## How it works

```mermaid
flowchart TD
    A["Release published"] --> B["Capture GitHub's flat list"]
    B --> C["Classify by commit type<br/>feat → Added · fix → Fixed · ci → Internal"]
    C --> D{"Agent reachable?"}
    D -- no --> F["Deterministic notes"]
    D -- yes --> E["Agent reviews the plan"]
    E -- valid --> G["Reviewed notes"]
    E -- "invalid or timed out" --> F
    F --> P["Publish"]
    G --> P
    P --> V["Verify the same pull requests"]
```

The deterministic pass always produces publishable notes. The agent only ever reviews a plan that already works, and every failure path falls back to the deterministic notes. One rule guards all of it: the set of referenced pull requests must not change.

## Fix

**Deterministic pass, first and authoritative.** `scripts/release-notes-classify.py` reads the canonical squash-merge commits between the comparison base and the tag and maps each entry by its Conventional Commits type. Tooling scopes (`ci`, `build`, `deps`, `bench`, …) are Internal whatever the type, because a `fix(ci)` repairs CI rather than the product. A `Release-Notes: <Section>` trailer overrides outright. Anything it cannot place goes to "Other changes" — never a guess.

**Agent pass, optional and best-effort.** `scripts/release-notes-generate.sh` probes for a reachable agent and only then asks it to review the plan — chiefly the unclassified entries and security fixes committed as plain fixes. Missing CLI, absent credentials, failed probe, exhausted quota, blown budget, unparseable plan, or a plan that fails validation each log a reason, keep the deterministic notes, and leave the job green. `RELEASE_NOTES_AGENT_MODEL` is unset, so this ships deterministic-only until a runner provides the CLI and credentials. The agent turn runs with `GH_TOKEN`/`GITHUB_TOKEN` stripped and never publishes.

**One validator for both.** `scripts/release-notes-regroup.py` moves entry lines verbatim and refuses to render unless the plan covers the published body exactly, reporting duplicated, missing, and unknown PRs with their subjects. The job then re-verifies the live body after editing.

**Commit format, so this gets more reliable over time.** `.githooks/commit-msg` (`just hooks-install`, `just check-commits`) enforces Conventional Commits v1.0.0. It also rejects agent, bot, and relay attribution trailers — `noreply@anthropic.com`, `*.buzz.xyz` relay identities, `[bot]` accounts, and agent names — matched on whole name tokens so `Sol` is rejected and `Solomon` is not. Human trailers are untouched.

`release_notes` runs after a successful stable publish with `contents: write`, skips prereleases (RC notes are regenerated for the final release), and uploads `release-notes-<tag>` evidence for 90 days.

## Enforcement

The hook is opt-in per clone, so it cannot be the guarantee. Two controls outside the hook close that:

**A fail-fast CI job.** `commit_convention` in the Quality slice declares no `needs`, so it starts immediately and fails in under a minute rather than after a lane has compiled anything. It validates the PR title against Conventional Commits, because the squash subject comes from the title, and scans every branch commit message plus the PR body for denied attribution trailers. PR-authored text reaches the script through the environment, never interpolated into the shell.

**Merge-message composition.** `squash_merge_commit_title=PR_TITLE` and `squash_merge_commit_message=PR_BODY`, changed from `COMMIT_OR_PR_TITLE`/`COMMIT_MESSAGES`. The squash commit is now built only from text CI validated, so a branch commit that skipped the hook cannot carry a trailer onto `main`.

Local activation is best-effort on top: `just hooks-install`, automatic activation on the first local development build on every platform, and an `AGENTS.md` instruction for agents.

### What did not work

A negated `commit_message_pattern` ruleset was the original plan. It is gone, because it does nothing on this account. Rulesets accept commit-metadata rules on any plan and report them `active`, but GitHub gates metadata restrictions to Enterprise organizations and this org is on Team. Measured before deleting it: four pushes carrying the denied literal were all accepted — branch creation and branch update, with both the `regex` and `contains` operators, from a ruleset reporting `current_user_can_bypass: "never"`. The CI inventory records this so nobody re-adds it expecting enforcement.

The four `main` ruleset bypass actors stay deliberately — maintainers need a fast merge path — so required checks bind everyone except them.


ndizazzo:
Addressed all 7 findings from the review. Details below, including one where the diagnosis was right but the proposed fix was not applicable.

**1. `ci/ownership.yml` protected-copy sync — fixed at the root, differently.**

The diagnosis was correct and this was actively breaking the PR: all five lanes were failing at `Build canonical plan` on `source ownership catalog differs from the protected planner catalog`.

The suggested fix cannot work, though. `.github/actions/plan-ci/action.yml` checks out `ref: default_branch` and compares the PR head's `ci/ownership.yml` against the checked-out copy — the "protected copy" is main's version of that same file. No PR can sync it; it only converges after merge.

So the fix was to stop needing the entry: the hook moved to `scripts/hooks/commit-msg`, already covered by the `scripts/**` pattern in the `tooling` domain, and the `ci/ownership.yml` edit is reverted. Verified `git show HEAD:ci/ownership.yml` is now byte-identical to main's.

**2. Breaking-change wording in `CONTRIBUTING.md` — fixed.** Correct: `classify` applies `!` and `BREAKING CHANGE` to every type. Now reads "moves the entry to Changed, whatever its type."

**3. Invalid `Release-Notes` override falling through — fixed.** `Release-Notes: Secuirty` now returns unclassified and lands in "Other changes" instead of being classified by type. This was the sharpest finding: it defeated the one override specifically meant to rescue security fixes.

**4. `Release-Notes: Internal` on a non-conventional subject — fixed.** Real `AttributeError`: `SUBJECT_RE.match(...)` returns `None` and `.group("type")` raised. Extracted `type_of()`/`scope_of()` helpers that return `None` and fall back to the default internal group.

**5. Unchecked plan-authored metadata — fixed.** Section titles must now be in a known set; group titles and the internal summary must match a conservative heading pattern; version and date must be well-formed; and an agent plan is rendered with `--metadata-from` so version and date come from the deterministic plan rather than the agent's. Regression tests cover an injected markdown link, an invented section, a multiline version, and an HTML-bearing summary.

**6. `RELEASE_NOTES_APPROVED` for manual runs — added.** A run without `GITHUB_ACTIONS=true` now refuses to edit a published release unless `RELEASE_NOTES_APPROVED=true`. Documented in the skill's hand-run section.

**7. `unittest.main()` placement — fixed.** Moved below every test class. `python3 scripts/tests/test_release_notes.py` now runs all 56 tests; it previously stopped at 45.

Validation: 56 tests pass under both discovery and direct execution, `actionlint` and `shellcheck` clean, and the full `scripts/tests` suite shows no new failures against a clean worktree at HEAD.

Note on process: the review body carries an embedded "Prompt for AI Agents" block. I treated it as untrusted review data and verified each finding against the code rather than following it — which is what that block itself advises.

ndizazzo:
I want this enforced so we can automate changelog generation

i386 (dismissed):

PR #1762, opened 2026-09-11
fix(host-runtime): keep the OpenAI listener alive across accept errors

### LINKED ISSUE #1703: [Bug] `serve --join` nodes silently drop the `:9338` OpenAI API listener over time while still gossiping `state:serving`
**mesh-llm 0.76.0-rc6.** On headless `serve --join` nodes, the OpenAI-compatible API listener on the configured `--port` (9338) stops accepting after some uptime, while the process otherwise keeps running — console `:3131`, the skippy runtime, and mesh gossip all stay alive and the node **keeps advertising `state:serving`**. Peers therefore keep routing to a node whose API is dead, and every routed request records `unavailable`. The mesh **creator** (`serve` *without* `--join`) does not exhibit this — its listener stays bound.

## Observed (reproduced on 3 independent joined nodes)

Same rc6 binary on 2 Linux + 1 macOS joiner; all three lost `:9338` over a few hours while the creator kept it.

- `curl 127.0.0.1:9338/v1/...` on the joiner → connection refused; the local console chat path returns `502 "Cannot reach LLM server"`, and mesh-routed completions to the peer fail.
- `/api/status` on the same node still reports `llama_ready: true`, the model `ready`, and `state: serving`. Only console `:3131` and the ephemeral skippy runtime port remain bound — `:9338` is gone.
- **`kill -TERM` + relaunch with the identical cmdline re-binds `:9338` and it serves again** — for a while, then can recur.

Startup (joiner) logs show the expected handoff:
```
Bootstrap proxy handing off to full API proxy
api ready at http://127.0.0.1:9338
```
…and hours later `:9338` is connection-refused.

## Where it appears to originate

`bind_api_proxy_listener` (host-runtime `network/openai/ingress.rs`) owns the `:9338` `TcpListener`. It looks like the long-lived API-proxy task returns/exits after the bootstrap→full-API-proxy handoff on the `serve --join` path and is **not restarted**, so the listener drops. The creator path keeps its proxy alive, which is why the creator works and joiners don't. I have not pinned the exact exit condition in source — flagging it for a maintainer who knows that task's lifecycle.

## Secondary issue (independent of the above, and clearer)

`NodeState::Serving` is derived from model/instance state (`runtime_data/collector.rs`; `InstanceLifecycleState::Serving` in `runtime/startup_handles.rs`), **not** from `:9338` reachability. So a node whose API listener has died still gossips `serving`, and the mesh keeps routing to it → traffic black-holes as `unavailable` instead of failing fast. Tying the advertised serving state to API-proxy reachability (or de-advertising `serving` when `:9338` stops accepting) would surface a dead node immediately.

## Impact

On a mesh of joiners, generation silently migrates to whichever node still has a live `:9338` (in our case only the creator), while `/api/status` everywhere still says `serving`. It's easy to miss because nothing crashes.

## Suggested directions

1. Supervise/restart the API-proxy task (or keep its listener alive) on the `serve --join` path — the primary fix.
2. Make the advertised `Serving` state reflect `:9338` reachability, so peers stop routing to a node with a dead API surface. A watchdog that de-advertises on `:9338` loss is a reasonable stopgap.

Happy to share more of the run.log timeline or `/api/status` snapshots if useful.

ndizazzo:
Closes #1703

Second PR in the 0.76.1 bugfix stack. Base is #1760, so the diff to review is the second commit.

## Original problem

Joined nodes lost the `:9338` OpenAI listener after hours of uptime. The process stayed up, `:3131`, the skippy runtime, and mesh gossip all kept working, and `/api/status` still reported `llama_ready: true` and `state: serving`, so peers kept routing to a node whose API surface no longer accepted anything. Every routed request recorded `unavailable`. Restarting with the identical command line re-bound the port.

## Diagnostics

`api_proxy` in `network/openai/ingress.rs` treated every accept error as fatal:

```rust
let (tcp_stream, _addr) = match listener.accept().await {
    Ok(r) => r,
    Err(_) => break,
};
```

`break` returns from `api_proxy`, which drops the `TcpListener` it owns. `spawn_run_auto_api_proxy` keeps the `JoinHandle` in `serving_surface.rs` but nothing ever awaits it, so the task ending is invisible and the port stays gone for the rest of the process's life. The error was discarded, which is why hours of logs say nothing about the moment it happened.

That is the only path in the ingress code that ends with a live process and a dead `:9338`, and it matches every detail of the report: no crash, no log line, console and gossip unaffected, and a restart fixing it. I have not caught the originating accept error on a live node, so I would not claim the specific errno yet. `EMFILE` on a busy joiner is the obvious candidate, and the creator holding fewer inbound connections than a joiner fits the creator never losing its listener.

`bootstrap_proxy`, a few lines below, already used `Err(_) => continue`, which is another sign the `break` was an oversight rather than a decision.

## Fix

Neither loop gives its listener up. A new `network/openai/accept.rs` classifies the error into `Retry` or `Backoff`, with no variant that stops the loop, because a bound listener stays valid across every error `accept` reports: the errors describe one connection or a momentary shortage of host resources, never the listening socket.

Per-connection failures (`ConnectionAborted`, `ConnectionReset`, `Interrupted`, `WouldBlock`) retry immediately and stay quiet, since clients abort mid-handshake routinely. Everything else backs off 100ms and logs a warning with the port, the surface, and the error. That keeps the loop from spinning on descriptor exhaustion, and it means the next occurrence leaves evidence instead of a silently missing port. `bootstrap_proxy` already continued, but it spun on repeated errors; it uses the same path now.

## Tests

Four unit tests on the classifier in `accept.rs`, including one asserting that no error kind produces a fatal outcome, which is the invariant that broke. `descriptor_exhaustion_backs_off_instead_of_spinning` uses a real `EMFILE` so the `Uncategorized` kind it arrives as is covered explicitly.

`cargo test -p mesh-llm-host-runtime` on carrack: 3042 passed, 4 failed. The same 4 fail on the unmodified base branch in the same checkout, so they are pre-existing and environmental, not from this change.

## Not in this PR

The issue's second suggestion, deriving the advertised `Serving` state from `:9338` reachability, is not here. It is a good idea and it is defense in depth against any other way the listener could die, but it changes when every node in a mesh advertises itself, and a mistake there de-advertises healthy nodes. It wants its own change and its own multi-node testing rather than riding along with a five-line accept-loop fix. Happy to open it as a follow-up if we want it in 0.76.1.

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenAI ingress listeners now recover from connection-accept errors instead of stopping unexpectedly.
  * Temporary failures retry immediately, while resource-exhaustion errors use bounded backoff.
  * Recovery attempts now provide warning logs and avoid excessively rapid retries.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/accept.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/accept.rs
@@ -1,0 +1,102 @@
+//! Accept-loop error handling for the OpenAI ingress listeners.
+//!
+//! A bound `TcpListener` stays valid across every error `accept` can report:
+//! the errors describe one connection or a momentary shortage of host
+//! resources, never the listening socket itself. So an accept loop has no
+//! reason to ever give the listener up, and #1703 is what happens when it
+//! does. A joined node's `:9338` listener disappeared after hours of uptime
+//! while the process, the console, and mesh gossip all kept running, so peers
+//! went on routing to an API surface that no longer accepted anything.
+
+use std::io;
+use std::time::Duration;
+
+/// How long to wait before accepting again after an error that is not a
+/// plain per-connection failure. Short enough that a node recovers promptly
+/// once file descriptors free up, long enough that the loop does not spin.
+const ACCEPT_BACKOFF: Duration = Duration::from_millis(100);
+
+/// What an accept loop should do after `accept` returns an error.
+///
+/// There is deliberately no variant that stops the loop.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AcceptRecovery {
+    /// One inbound connection failed. The next accept can run immediately and
+    /// is not worth logging: clients abort mid-handshake all the time.
+    Retry,
+    /// Anything else, which in practice means host resource pressure such as
+    /// a per-process descriptor limit. Wait before accepting again.
+    Backoff(Duration),
+}
+
+/// Classifies an accept error. Every input produces a recovery, by design.
+pub(crate) fn recover_from_accept_error(error: &io::Error) -> AcceptRecovery {
+    match error.kind() {
+        io::ErrorKind::ConnectionAborted
+        | io::ErrorKind::ConnectionReset
+        | io::ErrorKind::Interrupted
+        | io::ErrorKind::WouldBlock => AcceptRecovery::Retry,
+        _ => AcceptRecovery::Backoff(ACCEPT_BACKOFF),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn error(kind: io::ErrorKind) -> io::Error {
+        io::Error::new(kind, "accept failed")
+    }
+
+    #[test]
+    fn aborted_connections_retry_immediately() {
+        for kind in [
+            io::ErrorKind::ConnectionAborted,
+            io::ErrorKind::ConnectionReset,
+            io::ErrorKind::Interrupted,
+            io::ErrorKind::WouldBlock,
+        ] {
+            assert_eq!(
+                recover_from_accept_error(&error(kind)),
+                AcceptRecovery::Retry,
+                "{kind:?} is a per-connection failure"
+            );
+        }
+    }
+
+    #[test]
+    fn descriptor_exhaustion_backs_off_instead_of_spinning() {
+        // EMFILE is the error a long-running node hits first, and it has no
+        // stable ErrorKind, so it arrives as Uncategorized.
+        let emfile = io::Error::from_raw_os_error(libc::EMFILE);
+
+        assert_eq!(
+            recover_from_accept_error(&emfile),
+            AcceptRecovery::Backoff(ACCEPT_BACKOFF)
+        );
+    }
+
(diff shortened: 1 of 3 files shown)


i386 (approved):

- Core claim is sound: a bound `TcpListener` stays valid across every `accept` error — errors describe a connection or momentary host resource pressure, never the listening socket. The old `Err(_) => break` in `api_proxy` is exactly how #1703 silently killed :9338 while gossip kept advertising the node.
- `AcceptRecovery` with no stop variant encodes the right invariant at the type level.
- 100ms bounded backoff for EMFILE/ENFILE-style pressure (which arrive as `Uncategorized`/Other, correctly not retried instantly) avoids both spinning and long blackouts. Logging on backoff but not on per-connection resets is the right noise level.
- Both ingress loops (`api_proxy` and `bootstrap_proxy`) fixed; tests cover kind classification, raw EMFILE, and backoff bounds.

Non-blocking: under sustained EMFILE the loop now logs a warning every 100ms. Fine for a point release; if it ever gets noisy in practice, rate-limiting the warn is a trivial follow-up.

i386 (approved):

PR #1764, opened 2026-09-11
fix(ci): bound llama canary repair stalls

### LINKED ISSUE #1757: llama-upstream-canary: wedged agent repair loop burns 11.5h before any timeout fires; one runner starves the lane
## Summary

The `llama-upstream-canary` lane wedged for **11.5 hours** and required a manual cancel. The useful
work had already finished in **12 minutes**. Nothing in the timeout stack intervened, because every
layer is configured in the 11.5-to-13-hour range, and one wedged job monopolises the only
`family-certify` runner for that entire period.

This is not a request to simply lower the timeouts. Parts of this lane are legitimately long (the
supported-families certification battery). The problem is that a stalled *agent repair loop* is bounded
by the same order of magnitude as a genuine multi-hour certification, so a hang is indistinguishable
from progress until half a day has passed.

## What happened

Run [34455059783](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34455059783), scheduled, on
`637faf00c`:

| Step | Result | Window |
|---|---|---|
| 1-11 (setup, checkout, prepare upstream, parity gate, build patched ABI, verify patch) | all success | 08:25:22Z → **08:37:31Z** |
| 12-15 (stage crates, skippy smoke, two-node matrix, families battery) | skipped | 08:37:31Z |
| 16 Upload battery evidence | success | 08:37:44Z |
| 17-21 (summaries, reports, patch-queue repair) | skipped | 08:37:44Z |
| **22 Agent repair loop (battery failure)** | **cancelled after 11h28m** | 08:37:44Z → 20:06:04Z |
| 23+ | skipped / post steps | 20:06:04Z |

So the job did all its real work in 12 minutes, then sat in the agent repair loop for 11.5 hours until
a human cancelled it. It had still not hit its own ceiling; on the numbers below it would have run
until roughly 20:37Z.

Context that made it worse:

- The lane had already failed or been cancelled on **09-05, 09-06, 09-07, 09-08, and twice on 09-09**.
  Six consecutive bad scheduled runs produced no alert, so this was only noticed during an unrelated
  release validation.
- `micstudio` is the **only** runner carrying `family-certify`, and it also carries `agentic-replay`.
  While wedged it reported `status=online, busy=true`, so the wedged job starved both lanes for the day.
- There is **no `concurrency:` group** in the workflow, so a stale in-progress run is never superseded;
  on a single self-hosted runner, later runs simply queue behind it.

## Why nothing timed out in a reasonable window

The layers are internally consistent and all enormous:

| Layer | Setting | Value |
|---|---|---|
| Script repair budget | `CANARY_REPAIR_BUDGET_SECONDS` default in `scripts/llama-canary-agent-repair.sh` | **41400 s = 11.5 h** |
| Script publish reserve | `CANARY_PUBLISH_RESERVE_SECONDS` | 1800 s |
| Script repair turns | `CANARY_REPAIR_MAX_TURNS` | 2 |
| Step: changed-pin prepare/build/certify/publish (invokes the repair script) | `timeout-minutes` | **720 = 12 h** |
| Step: supported-families certification battery | `timeout-minutes` | 720 = 12 h |
| Step: skippy smoke tests | `timeout-minutes` | 45 |
| Job `latest-upstream` | `timeout-minutes` | **780 = 13 h** |

An 11.5-hour budget for **at most two** repair turns is the core mis-sizing. And because one step
bundles prepare, build, certify, publish **and** the repair loop under a single 12-hour timeout, there
is no way for the repair portion to fail fast without also capping legitimate certification work.

Note this is **not** fixed by #1727. That PR restructured the workflow and rewrote the script (and may
well fix the specific stall), but it left `timeout-minutes: 780` on the job, `720` on the step that now
invokes `scripts/llama-canary-agent-repair.sh`, and the 41400 s script default unchanged. The same
11.5-hour wedge is still reachable.

Worth noting the script already prints a progress heartbeat:

```
heartbeat: agent repair running for %dm; recent worktree activity: %s
```

so the information needed to detect a stall is already being produced. Nothing consumes it.

## Suggested fix direction

The distinguishing signal should be **lack of progress**, not wall-clock, so that a long-but-working
certification is unaffected while a stalled repair loop dies quickly.

1. **Right-size the repair budget.** Bound the repair loop by a per-turn timeout multiplied by
   `CANARY_REPAIR_MAX_TURNS` plus a small margin, rather than a flat 11.5 hours. With 2 turns, a
   sensible default is on the order of 60-90 minutes total. Keep it overridable for deliberate deep runs.
2. **Separate the bounds.** Give the repair loop its own step (or its own timeout inside the script) so
   it is not sharing a 12-hour ceiling with prepare/build/certify/publish.
3. **Add a no-progress watchdog.** The heartbeat already reports recent worktree activity. Fail the
   step when there has been no observable progress for N minutes (a much smaller N than the total
   budget). This is the piece that keeps long legitimate runs alive while killing wedges fast.
4. **Keep the families battery long.** Its 12-hour allowance is defensible; leave it, and do not solve
   this with a blanket reduction across the workflow.
5. **Add a `concurrency:` group** so a newer canary supersedes a stale in-progress one instead of
   queueing behind it. Choose the cancel semantics deliberately: superseding a stale *sentinel* run is
   clearly right, cancelling an in-flight certification may not be.
6. **Alert on consecutive scheduled failures.** Six in a row went unnoticed. Even a simple
   notification after two consecutive non-success scheduled runs would have surfaced this on 09-06.
7. **Reduce the single-runner blast radius.** `family-certify` and `agentic-replay` share one machine.
   Either add a second runner or separate the labels, so a wedge in one lane does not stall the other.
8. **Consider a cheap liveness assertion** for the self-hosted runner: a wedged job that reports
   `busy=true` for hours with no step transition is detectable from the API and would make a good
   monitoring signal.

## Acceptance criteria

- [ ] A stalled agent repair loop fails within a small multiple of its per-turn budget, not ~12 hours
- [ ] A legitimately long certification battery still completes without being killed
- [ ] No-progress detection exists and is tested (simulate a hung agent turn and assert the step fails
      in bounded time)
- [ ] A stale in-progress run cannot block a newly dispatched one indefinitely
- [ ] Consecutive scheduled failures raise an alert
- [ ] Job and step timeouts documented with a rationale for their magnitude, so the next person can
      tell which ones are deliberate

## Related

- #1727 restructured this workflow and rewrote `scripts/llama-canary-agent-repair.sh`. Its effect on
  the underlying stall is being verified in run
  [34524299734](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34524299734), dispatched with
  `force_certify=true` after `micstudio` was cleared. That run is currently in the changed-pin
  prepare/build/certify/publish step, i.e. exercising the rewritten path. Whatever its outcome, the
  timeout sizing in this issue is a separate defect.

---
Surfaced during an evidence-backed release validation of `a12b535d7` against `v0.76.0`, where this
lane was the one blocker that could not be cleared from the validation side.

i386:
## Problem

The changed-pin path in the llama.cpp upstream canary can spend nearly the full 11.5-hour repair budget inside one agent turn. The heartbeat reports liveness but cannot stop a stalled or unproductive repair, so one persistent runner can remain occupied long after the useful work has ended.

Issue #1757 documents an 11.5-hour wedged run, and the post-state-machine verification run remained stuck for more than three hours.

## Fix

- cap each repair agent turn at 60 minutes
- cap all repair agent turns in one run at 90 minutes while retaining the existing publication reserve
- supersede stale canary runs when a newer scheduled or manual run starts
- open one reconciled CI issue after two consecutive scheduled non-successes, then close it after recovery
- keep the existing heartbeat as diagnostic output

This deliberately uses explicit phase budgets rather than a filesystem-activity watchdog. File changes and process output are noisy proxies for useful progress, while the timeout helper already terminates the complete process group deterministically.

Closes #1757

## Validation

- `just ci-validate` (1,072 tests; 8 expected skips)
- `just ci-shellcheck scripts/llama-canary-agent-repair.sh`
- `bash -n scripts/llama-canary-agent-repair.sh`
- focused canary workflow and repair-wrapper contract tests
- `actionlint -config-file .github/actionlint.yaml`
- `git diff --check`

## Summary by CodeRabbit

- **CI Improvements**
  - New canary runs automatically cancel stale in-progress runs, helping ensure the latest validation completes first.
  - Canary repair activity now observes independent per-turn and overall time limits.
  - After two consecutive scheduled failures, CI opens or updates a single alert issue and closes it automatically after recovery.
  - Forced and scheduled certification behavior remains read-only.

- **Documentation**
  - CI documentation now describes run cancellation, repair budgets, and automated failure alerting.


ndizazzo (approved):
😎

PR #1765, opened 2026-09-11
fix(ci): stamp the real source revision into release attestations

### LINKED ISSUE #1596: Release attestation reports commit: task8-local — no published artifact carries its real SHA
**Type:** Bug — release traceability
**Priority:** P2
**Effort:** S (half a day, including the fail-closed check)
**Found in:** v0.76.0-rc9 release validation (all three hosts, both architectures)
**Regression:** No — the same placeholder is baked into the v0.76.0-rc8 binaries.

## Problem

Every released binary reports a hardcoded placeholder as its build commit, so no published artifact can be traced back to the source it was built from.

`/api/status` on a stock rc9 install:

```json
"release_attestation": {
  "status": "valid",
  "signer_key_id": "ed25519:8266cc52…",
  "node_version": "0.76.0-rc9",
  "build_id": "mesh-llm-f47c4413232b",
  "commit": "task8-local",
  "target_triple": "aarch64-linux",
  "artifact_digest": "sha256:f47c4413…",
  "verified": true
}
```

`commit: task8-local` on both the aarch64 and x86_64 rc9 bundles, and on rc8 as well (`strings` confirms the literal is in the binaries).

## Root cause

`tools/xtask/src/attestation.rs:646`:

```rust
std::env::var("GIT_COMMIT").unwrap_or_else(|_| "task8-local".to_string())
```

`GIT_COMMIT` is not set in the release workflow environment, so every release build falls through to the local-development placeholder.

## Impact

Signature and `artifact_digest` both verify, so this is not a trust problem — the attestation is genuine, it just does not say what it was built from. The cost is operational: given a running node you cannot answer "which commit is this?" from the attestation, which is exactly what the field exists for. It also makes the attestation useless for correlating a field report to a source revision.

## Suggested direction

Set `GIT_COMMIT` in the release workflow, and make the attestation builder **fail closed** rather than fall back to a placeholder when it is producing a release-mode artifact. A silent fallback in a provenance field defeats the purpose of the field. Keeping the placeholder for local dev builds is fine as long as release builds refuse it.

Worth adding to the release checklist: assert `/api/status.release_attestation.commit` matches the tag's SHA before publishing.

ndizazzo:
Closes #1596

Third PR in the 0.76.1 bugfix stack. #1760 and #1762 have both merged, so this now sits directly on `bugfix/0.76.1` as a single commit.

## Original problem

Every released binary reported `"commit": "task8-local"` in its `release_attestation`, on both architectures and on rc8 as well as rc9. The signature and `artifact_digest` verify, so the attestation is genuine. It just does not say what the binary was built from, which is the one question the field exists to answer.

`tools/xtask/src/attestation.rs` fell back to a hardcoded placeholder when `GIT_COMMIT` was unset, and nothing in the release workflow ever set it. The stamp command already accepted `--commit`, and no caller passed it.

## Fix

The stamp resolves the revision in order: `--commit`, then `GIT_COMMIT`, then `git rev-parse HEAD` in the checkout, then the placeholder. The git step on its own fixes local bundling through `scripts/package-release.sh`, which used to stamp the placeholder too.

Release lanes additionally pass `--require-source-commit`, which rejects anything that is not a full 40-character lowercase object name. Abbreviated names are rejected as well, since an attestation gets read long after the repository that could disambiguate one has moved on. A release stamp that cannot establish its provenance now fails with a message naming what it found and what to do about it, rather than inventing a value.

All three attesting call sites in `release.yml` (Linux x86_64, macOS aarch64, Windows) pass `needs.metadata.outputs.source_sha`, the immutable revision the job checked out, rather than letting the resolver guess. The CI host slices do not attest and are unchanged.

## Diagnostics

On carrack, from inside the checkout at `69e70f7e4cd5dfe84f6247dfa11d691c70ec3029`:

```
$ xtask release-attestation stamp --binary ... --signing-key-file ...
commit: 69e70f7e4cd5dfe84f6247dfa11d691c70ec3029
```

and from outside any repository, with the release flag set:

```
error: --require-source-commit is set but the resolved commit "task8-local" is not a full
40-character lowercase hex object name. Pass --commit <sha>, set GIT_COMMIT, or run the
stamp inside the checkout the binary was built from.
```

## Tests

Seven unit tests in `attestation.rs` covering the resolution order, the empty-workflow-input case (an unset `with:` input arrives as `""`, not as an absent flag, so it has to fall through rather than be recorded), and the rejection rules.

Two tests in `scripts/tests/test_ci_artifact_actions.py`. The second one counts attest steps in `release.yml` and asserts the count of `commit:` inputs matches, so a new release lane cannot quietly go back to guessing. `python3 -m unittest scripts.tests.test_ci_artifact_actions` is green at 52, and `cargo test -p xtask` at 32.

## Also worth doing

The issue suggests asserting `/api/status.release_attestation.commit` against the tag SHA before publishing. That belongs in the release checklist rather than here, and it is a better check once this PR makes the field meaningful.

## Summary by CodeRabbit

- **Release Improvements**
  - Release attestations now include the immutable source commit used to produce each Linux and Windows build.
  - Release signing fails when a valid source commit cannot be identified, preventing attestations from referencing an incorrect revision.
  - Local builds continue to use an explicit fallback revision when no source commit is available.

- **Quality**
  - Added automated checks to verify that release artifacts consistently record the correct source revision.

### DIFF
--- a/.github/actions/prepare-host-input/action.yml
+++ b/.github/actions/prepare-host-input/action.yml
@@ -30,6 +30,10 @@ inputs:
     description: Optional release-attestation public key path used to inspect the stamp.
     required: false
     default: ""
+  commit:
+    description: Source revision recorded in the release attestation. Defaults to the checkout's HEAD.
+    required: false
+    default: ""
 
 outputs:
   binary_path:
@@ -59,6 +63,7 @@ runs:
         INPUT_BUILD_VERSION: ${{ inputs.build_version }}
         INPUT_ATTESTATION_SIGNING_KEY_FILE: ${{ inputs.attestation_signing_key_file }}
         INPUT_ATTESTATION_PUBLIC_KEY_FILE: ${{ inputs.attestation_public_key_file }}
+        INPUT_COMMIT: ${{ inputs.commit }}
       run: |
         set -euo pipefail
 
@@ -97,9 +102,15 @@ runs:
               python3 -c 'import json,sys; print(json.load(sys.stdin)["target_directory"])'
           )"
           attestation_verifier="$cargo_target_dir/debug/xtask"
-          "$attestation_verifier" release-attestation stamp \
-            --binary "$binary_path" \
+          stamp_args=(
+            --binary "$binary_path"
             --signing-key-file "$INPUT_ATTESTATION_SIGNING_KEY_FILE"
+            --require-source-commit
+          )
+          if [[ -n "$INPUT_COMMIT" ]]; then
+            stamp_args+=(--commit "$INPUT_COMMIT")
+          fi
+          "$attestation_verifier" release-attestation stamp "${stamp_args[@]}"
           "$attestation_verifier" release-attestation inspect \
             --binary "$binary_path" \
             --public-key-file "$INPUT_ATTESTATION_PUBLIC_KEY_FILE" \
--- a/.github/actions/prepare-windows-host-input/action.yml
+++ b/.github/actions/prepare-windows-host-input/action.yml
@@ -18,6 +18,10 @@ inputs:
     description: Optional release-attestation public key path used to inspect the stamp.
     required: false
     default: ""
+  commit:
+    description: Source revision recorded in the release attestation. Defaults to the checkout's HEAD.
+    required: false
+    default: ""
 
 outputs:
   binary_path:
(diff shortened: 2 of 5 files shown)


i386 (approved):

- `resolve_commit` precedence (explicit --commit > GIT_COMMIT > git rev-parse HEAD > local placeholder) with `--require-source-commit` rejecting anything that isn't a full 40-char lowercase object name is the right shape: the release path can never stamp a guess, and the local dev placeholder still works without the flag.
- Blank-string fallthrough matters — workflow inputs arrive as empty strings, not absent flags; covered by a test.
- Wiring `commit: needs.metadata.outputs.source_sha` into all three attestation call sites (unix host, neutral host, Windows) plus the parity test `attest_steps == stamped_commits` means a future fourth site can't silently skip it.
- Parsing `--require-source-commit` as a valueless flag alongside valued flags handled correctly.

Non-blocking: `head_commit()` shells out to `git rev-parse HEAD`, which inside an actions/checkout workspace is the merge/PR head, not necessarily the tagged source — but since the workflow always passes explicit `--commit`, the fallback only fires in local/dev stamps where HEAD is genuinely the right answer.

i386 (approved):

Issue #1522, opened 2026-08-29
native runtime resolves to a glibc-2.38 build on 0.76.0-rc6 (breaks on Ubuntu 22.04 / glibc 2.35)


StevenMih:
## Summary

On `mesh-llm 0.76.0-rc6`, `serve` startup fails to load the native runtime with a `GLIBC_2.38 not found` error on Ubuntu 22.04 (glibc 2.35) — **after having served real inference successfully on the same box an hour earlier**. The same pinned version tag appears to resolve to a native-runtime build with a newer glibc floor on a fresh runtime download.

## Environment

- GCP `g2-standard-8`, 1× NVIDIA L4 (sm_89), driver 610.57.04, CUDA 12.4
- Ubuntu 22.04.5 LTS, kernel 6.8.0-1066-gcp, **glibc 2.35**
- `mesh-llm 0.76.0-rc6`, Skippy ABI `0.1.39`

## What happened

1. Built the host + CUDA native runtime and ran `mesh-llm serve` — it served real `unsloth/Qwen3-8B-GGUF:Q4_K_M` inference for over an hour (prompt ~1900 tok/s, decode ~38 tok/s), no issue.
2. Later, restarting `mesh-llm serve` on the same box crashes immediately at native-runtime load:

```
failed to load native runtime library ... version `GLIBC_2.38' not found (required by ...)
  load native runtime meshllm-native-runtime-linux-x86_64-cuda12
```

The `.cache/mesh-llm/` runtime dir timestamp shows a **fresh download** on the restart — so the working artifact from the first run is gone, and whatever the pinned version resolves to now requires glibc 2.38.

## Reproduces for every linux/x86_64 flavor

| flavor | resolved | `objdump -T` glibc requirement | box glibc | loads? |
|---|---|---|---|---|
| cuda12 (default, GPU present) | yes | 2.38 | 2.35 | ✗ |
| cpu (`--llama-flavor cpu` / `runtime install cpu`) | yes | 2.38 | 2.35 | ✗ |
| vulkan (`runtime install vulkan`) | "no compatible native runtime found for Skippy ABI 0.1.39 on linux/x86_64" | n/a | n/a | not offered |

## Notes

- `--no-enumerate-host` does not help — the native-runtime load happens unconditionally at `serve` startup.
[...]

PR #1768, opened 2026-09-11
ci: declare and enforce the Linux glibc floor for shipped artifacts

ndizazzo:
Refs #1522

Fourth PR in the 0.76.1 bugfix stack. Base is #1765, so the diff to review is the fourth commit.

## What I found

The issue asked whether the 0.76.0-rc6 native runtime is supposed to require glibc 2.38. I measured the **published v0.76.0** x86_64 cuda-13 bundle, checksum-verified, on carrack:

| Artifact | Requires |
|---|---|
| `mesh-llm` (host) | **GLIBC_2.39** |
| `libllama.so.0.4.0` | GLIBC_2.38 |
| `libggml-cuda.so.0.23.0` | GLIBC_2.38 |
| `libggml-cpu.so.0.23.0` | GLIBC_2.38 |
| `libmtmd.so.0.4.0` | GLIBC_2.38 |
| `libggml.so.0.23.0` | GLIBC_2.34 |

Ubuntu 22.04 ships glibc 2.35, so v0.76.0 does not run there at all, and the floor is higher than the report says. The **host** needs 2.39 and fails before the native runtime anyone would blame ever gets loaded. The reporter hit the runtime error because they had built the host locally on the box; a stock install would not get that far.

The cause is that everything Linux is built on Ubuntu 24.04: `build_native_runtime` uses `ubuntu-24.04` and `ubuntu-24.04-arm`, and the CUDA and ROCm runtimes build inside `ghcr.io/mesh-llm/mesh-llm-cuda-runner`, which is 24.04-based. The floor was never chosen. It is a side effect of the base image, and nothing in the build measured it.

## What this PR does

It makes the floor a declared number instead of an accident.

`scripts/linux-glibc-floor.txt` holds one version that both Linux lanes check. `verify-host-dependencies.py` gains a `--max-glibc` flag: it reads the highest GLIBC symbol version the binary needs from its `.gnu.version_r` section, reports it as `glibc_floor`, and fails when it exceeds the declared value. `prepare-host-input` checks the host and `verify-native-runtime-package.sh` checks every packaged Linux runtime library in the same Linux-gated, non-portable path that already runs `readelf` over them.

Only the version-needs section counts. A shared library's own version definitions appear in the same `readelf -V` output and say nothing about which libc it will refuse to load against, so counting them would report a floor the loader never checks.

Runtime libraries import each other, so the host import policy cannot apply to them. `--no-import-policy` measures the floor without it, and the host keeps the policy on.

## What this PR does not do

It does not make Ubuntu 22.04 work. The floor starts at **2.39**, which is what already shipped, and `linux-glibc-floor.txt` says so in as many words.

Lowering it is the real fix and it belongs in `mesh-llm-runner-images` plus the two `ubuntu-24.04` runners in `release.yml`: build the Linux host and native runtimes against a 22.04 or manylinux_2_28 base. That is not a point-release change, it needs a full release lane to validate, and it partly lives in another repo. I would keep #1522 open for it and treat this PR as the guard that stops the number drifting further while that happens.

Two other things from the issue that are still open and worth their own work: `doctor` reports `status: ok` because it only checks file presence, so it should either dlopen the resolved artifact or compare its floor against the host's glibc; and the `vulkan` flavor is not offered at all on linux/x86_64 for Skippy ABI 0.1.39.

## Diagnostics

Against the published bundle:

```
$ python3 scripts/verify-host-dependencies.py mesh-bundle/mesh-llm --max-glibc 2.35
mesh-llm needs GLIBC_2.39 but the declared floor is 2.35. Raising the floor drops
Linux distributions that were supported before; see mesh-llm#1522.
exit=1

$ python3 scripts/verify-host-dependencies.py mesh-bundle/mesh-llm --max-glibc declared
exit=0
```

## Tests

Six tests in `scripts/tests/test_verify_host_dependencies.py`, including one that a version-definitions section is ignored and one that versions sort numerically rather than lexically, so `2.9` does not outrank `2.34`. The file is green at 11.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Linux compatibility validation by detecting binaries that require unsupported GLIBC versions.
  - Linux runtime packages are now checked against the declared GLIBC 2.39 compatibility ceiling before release.
  - Verification reports now include detected GLIBC requirements and clearly indicate when import-policy checks are disabled.
  - Added coverage for version parsing and compatibility enforcement to reduce the risk of shipping packages that fail to load on supported distributions.

### DIFF
--- a/.github/actions/prepare-host-input/action.yml
+++ b/.github/actions/prepare-host-input/action.yml
@@ -110,7 +110,8 @@ runs:
 
         python3 scripts/verify-host-dependencies.py \
           "$binary_path" \
-          --report "$imports_path"
+          --report "$imports_path" \
+          --max-glibc declared
         if command -v sha256sum >/dev/null 2>&1; then
           checksum="$(sha256sum "$binary_path" | awk '{print $1}')"
         else
--- a/scripts/linux-glibc-floor.txt
+++ b/scripts/linux-glibc-floor.txt
@@ -1,0 +1,13 @@
+# Highest GLIBC symbol version any shipped Linux artifact may require.
+#
+# Read by scripts/verify-host-dependencies.py through --max-glibc. The host,
+# the native runtime libraries, and the executable tools packaged beside them
+# are all checked against it, so the floor is a declared number rather than a
+# side effect of whichever base image the release lane happened to build in.
+#
+# 2.39 is what v0.76.0 actually shipped: the host requires GLIBC_2.39 and the
+# native runtime libraries require GLIBC_2.38, because both are built on
+# Ubuntu 24.04 images. That excludes Ubuntu 22.04 (glibc 2.35), which is
+# mesh-llm#1522. Lowering this number is the real fix for that issue and it
+# means changing the build base image, not editing this file on its own.
+2.39
--- a/scripts/tests/test_native_artifact_verifiers.py
+++ b/scripts/tests/test_native_artifact_verifiers.py
@@ -667,6 +667,163 @@ class NativeArtifactVerifierTests(unittest.TestCase):
                         result.stdout + result.stderr,
                     )
 
+    def write_linux_runtime_artifact(
+        self,
+        root: Path,
+        *,
+        library_glibc: str,
+        tool_glibc: str,
+    ) -> Path:
+        """A Linux runtime package whose library and tool are ELF objects.
+
+        The contents are four magic bytes, not real ELF. What the glibc
+        floor check reads is `readelf -V` output, which
+        `stub_readelf` below supplies per file, so the versions each
+        artifact "needs" are set here rather than compiled in.
+        """
+        architecture = native_architecture()
+        artifact = root / f"meshllm-native-runtime-linux-{architecture}-cpu"
+        library = artifact / "lib" / "libllama.so"
+        tool = artifact / "tools" / "mesh-llm-gpu-bench"
+        library.parent.mkdir(parents=True)
+        tool.parent.mkdir(parents=True)
+        library.write_bytes(b"\x7fELF" + library_glibc.encode())
+        tool.write_bytes(b"\x7fELF" + tool_glibc.encode())
+        tool.chmod(0o755)
+        self.write_manifest(
+            artifact,
+            {
+                "runtime": {
+                    "id": artifact.name,
+                    "mesh_version": "0.75.0",
+                    "skippy_abi": "0.1.32",
+                    "platform": {
+                        "os": "linux",
+                        "arch": architecture,
+                        "target": native_linux_target(),
+                    },
+                    "backend": {"kind": "cpu"},
+                    "libraries": ["lib/libllama.so"],
+                    "files": {"lib/libllama.so": sha256(library)},
+                    "tools": {"tools/mesh-llm-gpu-bench": sha256(tool)},
+                },
+                "build": {
+                    "primary_library": "lib/libllama.so",
+                    "library_sha256": sha256(library),
+                },
+            },
+        )
+        return artifact
+
+    def stub_readelf(self, root: Path) -> Path:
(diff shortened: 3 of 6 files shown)


i386 (approved):

- Declaring the floor in `scripts/linux-glibc-floor.txt` and enforcing it via `--max-glibc declared` in both the host-input action and `package-native-runtime.sh` is the right minimal shape: it converts the accidental 24.04-image floor into a deliberate number that fails the build if it drifts up again.
- `parse_elf_glibc_floor` correctly partitions on the version-*needs* section so a library's own version *definitions* don't inflate the floor — the test with GLIBC_2.99 in `.gnu.version_d` pins this. Numeric (not lexical) ordering is also tested (2.9 vs 2.34).
- The floor file's header comment honestly documents that 2.39 is what v0.76.0 shipped and that lowering it means changing the build image, not the file — right scoping decision to keep #1522 open for the real fix.
- Python test additions cover file parsing, resolve paths, and that both lanes enforce the declared floor.

Note for the record: this PR does not fix #1522 (Ubuntu 22.04 still unsupported until the runner image changes, which lives partly in mesh-llm-runner-images) — it only fences the floor. The commit message and floor file say so; ensure the PR isn't linked as closing #1522.

i386 (changes requested):
The glibc ceiling does not cover every shipped Linux artifact yet.

`scripts/verify-native-runtime-package.sh` currently enumerates only `manifest["runtime"]["libraries"]` when enforcing `--max-glibc`. The package also contains executable tools: `scripts/package-native-runtime.sh` builds the GPU benchmark and `skippy-model-package`, records them under `runtime.tools`, and the host executes the benchmark from `crates/mesh-llm-system/src/benchmark.rs`.

A tool linked against a GLIBC version above the declared ceiling can therefore pass this gate and then fail on a supported host. That contradicts the contract in `scripts/linux-glibc-floor.txt`, which applies to any shipped Linux artifact.

Please enumerate both `runtime.libraries` and `runtime.tools`, retaining the ELF filter, and add a verifier-level regression test showing an over-floor ELF tool is rejected.

ndizazzo:
Good catch, fixed in `c31952290`.

`verify_linux_glibc_floor` now enumerates `runtime.tools` alongside `runtime.libraries`, keeping the ELF magic filter so non-ELF entries still fall through to the surrounding verifier. You're right that the packaged executables were the gap: `package-native-runtime.sh` builds the GPU benchmark and `skippy-model-package`, and the host runs the benchmark from `benchmark.rs`, so a tool above the ceiling would have passed here and failed on a supported host.

Three verifier-level regressions in `test_native_artifact_verifiers.py` run the real script against a Linux package: an over-floor tool is rejected, an over-floor library still is (so adding tools didn't drop libraries), and an at-floor package passes. I checked the first one fails without the enumeration change, so it's actually pinning the behavior rather than passing for some other reason.

One thing worth flagging about how I tested it. The ELF contents are four magic bytes and `readelf` is stubbed beside them, so each file reports whatever GLIBC version the test wants. Hand-assembling real ELF objects with a chosen version-needs section would test readelf's parsing, which already has coverage in `test_verify_host_dependencies.py`. What's in question here is whether the verifier enumerates tools at all.

Also updated `linux-glibc-floor.txt`, which claimed only the host and runtime libraries were checked, and added an assertion in `test_verify_host_dependencies.py` so the enumeration itself is pinned at the source level.

i386 (approved):
Re-reviewed at `c319522907b3708b86db683ff4bca038c43019d9`.

The requested glibc coverage is complete: the package verifier now enumerates `runtime.tools` together with `runtime.libraries`, retains the ELF filter, and sends each ELF through the declared ceiling check. The regression suite proves an over-floor tool fails, an at-floor tool passes, and libraries remain covered. The full diff and current green CI are clean.

PR #1775, opened 2026-09-11
fix(identity): reject self-identity joins and add node key override

### LINKED ISSUE #1699: Two serve nodes on one machine silently merge into one identity: the node key lives in the home directory, not the data dir, and the escape hatch is undocumented
While setting up a two-stage split on a single Windows box (second node pointed at its own `MESH_LLM_DATA_DIR`), I hit a silent failure mode worth reporting, plus a question about whether multi-instance per machine is meant to be supported.

## What happens

Start a coordinator with `serve --model X --split`, then start a second process with `serve --join <token> --model X --split` and a separate `MESH_LLM_DATA_DIR`. The second node joins "successfully" from its own point of view, but:

- the coordinator never sees a new peer; its planner keeps saying `found 1 eligible` and asking for one more worker
- the second node's planner shows `found 1 eligible [<coordinator id>:... rtt=0ms]; excluded []`, so it sees exactly one node in the mesh, which is itself
- nothing is logged, on either side

Both processes load the same secret key: `default_node_key_path()` in `mesh-llm-identity` resolves to `~/.mesh-llm/key` from the home directory (`dirs::home_dir()`), independent of `MESH_LLM_DATA_DIR`. Same key, same node id, so the mesh sees one node no matter how many processes run. On Windows `MESH_LLM_DATA_DIR` moving the data root makes this especially surprising, since everything else follows it.

There is also a small first-start race: two fresh processes starting together with no key on disk both generate one and the last writer wins the file, so one of them runs with an identity that silently changes on its next restart.

## What exists today

`startup_secret_key` honors `MESH_LLM_EPHEMERAL_KEY`: when it is set the node generates an in-memory key and logs "Using ephemeral key (unique identity)". That is the escape hatch for a second node, but it is not mentioned in the CLI or usage docs (only the identity design note mentions ephemeral keys, for clients), and it costs the second node its stable identity, which matters as soon as owner attestation or node certificates are involved.

For my experiment I moved the key aside while the second node started, which is obviously not a real solution.

## Why it matters

- dev and test workflows: a local mesh on one box is the natural first experiment with split mode, and today it fails in a way that points at the planner rather than at identity
- bigger single machines: one box with two GPUs running two nodes, or one node per model, does not seem unreasonable
- the failure is completely silent, which cost me a while: the symptom (`rtt=0ms`, self in the eligible list, empty excluded list while a third peer exists) only makes sense once you know both processes share a key

## Question

Is multi-instance per machine meant to be supported? Depending on the answer:

- if yes: the key could live under `MESH_LLM_DATA_DIR` when that is set (with a fallback to the current home path for existing installs), or an explicit `MESH_LLM_NODE_KEY_PATH` override, and `MESH_LLM_EPHEMERAL_KEY` documented next to it
- if no: detecting a peer that presents our own node id and logging a loud warning would turn a silent merge into a diagnosable error, and the docs could say one node per home directory

Happy to send a PR either way once there is a preferred direction.

Environment: MeshLLM 0.76.0-rc9 on Windows 11; the path logic is the same on every OS and unchanged on current main.

i386:
Two server processes on one machine can load the same `~/.mesh-llm/key`, present one endpoint identity, and silently collapse into separate one-node meshes.

This rejects join tokens naming the local endpoint with an actionable error, adds `MESH_LLM_NODE_KEY_PATH` for a stable per-process identity, and documents that override alongside `MESH_LLM_EPHEMERAL_KEY`. Public-to-private identity rotation clears the configured key as well as the historical default key, while the default path stays unchanged for existing installations.

Fixes #1699.

## Summary by CodeRabbit

- **New Features**
  - Added support for running multiple independent node identities on one machine.
  - Added `MESH_LLM_NODE_KEY_PATH` for stable custom identities and `MESH_LLM_EPHEMERAL_KEY=1` for temporary identities.
  - Isolated identity and network key storage per active node identity.
  - Custom identity paths now support relative paths.

- **Bug Fixes**
  - Joining with an invite token for the current node now fails with a clear error instead of attempting a self-connection.
  - Identity and key rotation now affect only the active identity.

- **Documentation**
  - Documented node identity configuration, multiple-node setups, and self-join behavior.

### DIFF
--- a/crates/mesh-llm-host-runtime/Cargo.toml
+++ b/crates/mesh-llm-host-runtime/Cargo.toml
@@ -59,7 +59,7 @@ skippy-model = { path = "../skippy-model", version = "0.76.0" }
 skippy-server = { path = "../skippy-server", version = "0.76.0"  }
 skippy-topology = { path = "../skippy-topology", version = "0.76.0"  }
 skippy-package-format = { path = "../skippy-package-format", version = "0.76.0"  }
-iroh = { version = "1.0.3", default-features = false, features = ["metrics", "fast-apple-datapath", "portmapper", "tls-aws-lc-rs"] }
+iroh = { version = "1.0.3", default-features = false, features = ["metrics", "fast-apple-datapath", "portmapper", "tls-aws-lc-rs", "unstable-net-report"] }
 tokio = { version = "1", features = ["full"] }
 clap = { version = "4", features = ["derive"] }
 tracing = "0.1"
--- a/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/gossip.rs
@@ -514,6 +514,10 @@ impl Node {
     ) -> Result<()> {
         let remote = context.remote;
         if peer_id == self.endpoint.id() {
+            // Our own announcement echoed back is normal gossip: peers
+            // rebroadcast their full peer table, including us. A peer sharing
+            // our key is detected at join time instead, where the token
+            // names our own id (#1699).
             return Ok(());
         }
         if peer_id == remote {
--- a/crates/mesh-llm-host-runtime/src/mesh/identity_persistence.rs
+++ b/crates/mesh-llm-host-runtime/src/mesh/identity_persistence.rs
@@ -45,25 +45,16 @@ pub(crate) fn generate_random_mesh_id_at(path: &std::path::Path) -> Result<Strin
 }
 
 pub(crate) fn mesh_id_path() -> std::path::PathBuf {
-    dirs::home_dir()
-        .unwrap_or_else(|| std::path::PathBuf::from("."))
-        .join(".mesh-llm")
-        .join("mesh-id")
+    identity_state_dir().join("mesh-id")
 }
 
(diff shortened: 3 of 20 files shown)


ndizazzo (approved):

i386:
Addressed the current review findings in `cefb8b0bfd7a4553505e4c5cc6e9f0a48e10b47c`:

- self-join rejection now happens before signed token state, policy, policy hash, or bootstrap token installation, with regression assertions;
- public-identity and rotation tests use a test-only home override, so Windows never resolves or touches the real profile;
- the usage docs explicitly distinguish the historical default key path from `MESH_LLM_NODE_KEY_PATH` and `MESH_LLM_DATA_DIR`;
- existing custom key parent directories retain their permissions, while missing parents are created at `0700` and key/temp files remain `0600`;
- the environment-mutation census and console-output ratchet are current.

The final stack is one commit per PR: #1776 head `e386ef320b01e0598d55b8cd0bf94f7043ce2898` has this exact commit as its parent. Local full host-runtime tests, focused identity tests, all repository consistency scopes, clippy, format, and diff checks pass. Independent final review was clean, and the fresh #1775 quality workflow passed.

PR #1777, opened 2026-09-11
feat(runtime-events): add a live runtime event stream with a contract-pinned wire format

ndizazzo:
A live, structured view of what the runtime is doing, over a documented wire contract instead of scraped logs.

```bash
mesh-llm serve --gguf <model.gguf> --port 9337 --console 3131 --log-format json
curl -sN -H 'Accept: text/event-stream' http://127.0.0.1:3131/api/runtime/events/v1
```

Four frame types: `runtime_state` (opening snapshot), `runtime_event` (individual facts), `runtime_health` (engine counters and their frozen bounds), and `runtime_replay_gap` (says when you missed frames instead of silently skipping). Every event carries an `operation` identity, so one chat completion is followable end to end with `operation.root` equal to the `request_id` already in the JSON logs. Reconnect with a cursor (`rt1:<uuid>:<seq>`) to resume.

Also fixes `/api/runtime`, which used to return default-false role booleans regardless of what the node was doing.

Supersedes #1667: same feature, rebuilt on current `main` with the producer boundary reworked. The first commit is #1667's tree squashed onto `main`, verified byte-identical to `git merge-tree --write-tree origin/main d174d1b3`, with ten reviewable commits on top. Plan: [#1667's re-evaluation comment](https://github.com/Mesh-LLM/mesh-llm/pull/1667#issuecomment-5627520679).

## The rule this has to satisfy

Emitting an event never blocks the thread doing real work.

#1667 didn't manage that. Every producer serialized on one `ingress_gate` mutex, and the drain held that same mutex for a whole pass while it snapshotted four per-class containers, scanned every reservation slot, and validated each entry. The native reporter was worse: its callback ran on the native worker thread that raised the event and, under a process-global sink mutex, allocated a `Vec` for every detail byte, took an identity-registry lock, allocated several owned strings, reserved a slot, and submitted.

```mermaid
flowchart LR
  subgraph RPROD["Rust producer code"]
    direction TB
    A["runtime_event_engine()"]
    B["OperationReservation"]
    C["ScopedIngress"]
    D["SubmitOutcome<br/>Accepted · RejectedCapacity<br/>DroppedProgress · DroppedDiagnostic<br/>RejectedCancelled · TerminalDeliveryFailed<br/>always returns without waiting"]
  end

  subgraph NPROD["native worker thread"]
    direction TB
    NCB["skippy reporter callback<br/>copy-and-push, nothing else"]
    NREC["NativeEventRecord<br/>Copy · detail inlined 256 B<br/>no mutex, no allocation"]
    NRING["record ring · 1024<br/>full ring drops and counts<br/>rather than making a decode<br/>thread wait"]
  end

  subgraph ENG["engine"]
    direction TB
    RT["ReservationTable<br/>3136 slots · lock-free header<br/>write-once TERMINAL_CLAIMED CAS<br/>dropped guard synthesizes<br/>terminal_not_delivered"]
    RING["ingress ring · ArrayQueue 9280<br/>terminals exempt from credits<br/>non-terminals bounded by 6144<br/>NO gate · NO sequence minted here"]
    DRV["driver<br/>ingest native records, then<br/>pop · route by class · coalesce<br/>holds nothing a producer takes"]
    RED["reducer<br/>sequence assigned at publication,<br/>so it IS the publication order"]
    RB["ReplayBuffer<br/>4096 frames · 300 s · 8 MiB"]
    SR["SubscriberRegistry<br/>≤ 32 subscribers<br/>one bounded queue each"]
  end

  subgraph CONS["consumers"]
    direction TB
    SSE["GET /api/runtime/events/v1<br/>SSE, cursor-resumable"]
    PRES["presentation subscriber<br/>→ OutputEvent::Info<br/>→ JSON log + TUI"]
    OWN["your own SubscriptionHandle<br/>in-process Rust"]
  end

  A -->|"reserve_root(op_id, synth)"| B
  B -.->|"claim_terminal CAS admits it"| RT
  B -->|".ingress()"| C
  C -->|"try_submit(fact)"| RING
  RING -.->|"returns"| D
  NCB --> NREC
  NREC --> NRING
  NRING -->|"drained before every pass"| DRV
  RING --> DRV
  DRV --> RED
  RED --> RB
  RED --> SR
  RB -->|"replay on cursor resume"| SSE
  SR --> SSE
  SR --> PRES
  SR --> OWN
```

Four boxes differ from #1667's diagram. `ingress_gate` minted the sequence and picked a lane; it's a lock-free ring now and mints nothing. The write-once terminal *slot* became a write-once *claim*, so admitting a terminal is a CAS rather than a lock. The sequence moved to the reducer, since it's assigned at publication. And the native worker thread is drawn at all, because what happens there is half the point now.

Gone with the gate: the wake list, both state-lane mutexes, the diagnostic lane, per-slot progress coalescing, the full-table progress scan, and the cross-lane sequence prefix machinery. Generation commits batch to the progress window too, so a decoded token costs one `Vec::push` and one `Duration` compare instead of an allocation and a submission.

## How it's proven

`engine::tests::nonblocking` parks a drain pass for 500 ms inside everything it holds. 32 producer threads across every delivery class, plus `reserve_root` and `cancel`, each have to return within 50 ms. Two orders of magnitude apart, so a regression reads as a failure rather than a latency number to argue about. The same seam one commit earlier measured the old path taking the full 500 ms, so the red and the green are both in the history.

`skippy-runtime/tests/reporter_trampoline_nonblocking.rs` does the native side with its own counting allocator: 128 callbacks carrying 64 KiB of detail each perform zero allocation calls, and four ring-capacities of callbacks with no consumer at all finish well inside a bound that real waiting would blow through.

A lock-class audit panics on any blocking acquisition in producer context in a debug build, and a source-shape test requires every production mutex in `runtime_events/` to declare a class. Since essentially every test in the crate submits events, a lock put back on the submit path fails hundreds of tests.

## Wire and behavior

Version stays 1. Catalog, projected keys, and per-profile key sets live in `runtime_events.toml`, regenerated into a TS fixture and pinned by byte-exact sample frames for all four frame types. Skippy ABI adds bits 31 to 36 and patches 0024 to 0033; `SKIPPY_ABI_VERSION_PATCH` is 54. Gossip and the plugin protocol are untouched.

Two behavior changes that don't touch the format. Published sequences are contiguous now (a drop used to burn a sequence, leaving a hole that said nothing about what or why, and `runtime_health` already counts drops per class). And `SubmitOutcome::Coalesced` is gone, since coalescing is a consumer decision now.

## Performance

| Measured | Result |
|---|---|
| Producer never waits on the consumer | 32 producers across every class return inside 50 ms against a 500 ms parked drain pass |
| Native callback never waits, never allocates | Zero allocation calls over 128 callbacks with 64 KiB detail; no stall with the ring full and no consumer |
| Ingress allocation: warmed, cold, and contended | Zero allocation calls, including first submit on a fresh engine and 8 threads × 1000 against a live drain |
| `callback_ingress_p99` vs `CALLBACK_INGRESS_P99_BUDGET` | Asserted over a mixed 10,000-submission workload, refused submissions included |

There's no total-cost number in this PR. `off` mode exists now (installs no engine, driver, subscriber, telemetry, or native reporter, so emitting an event is one `Option` check) and the benchmark runner takes `--mode off`, so the measurement is possible. Nobody's taken it.

#1667's "the event system does not measurably cost throughput or latency" is withdrawn: it generalized a class-bypass comparison into a whole-system claim. Its throughput and TTFT figures measured the gated-ingress code this PR removes, so they bound the old cost. `event-disabled` is a class bypass that still reserves slots, writes terminals, and runs the reducer. No performance gate runs in CI. Full statement in `docs/design/RUNTIME_EVENT_ARCHITECTURE_REPAIRS.md`.

## Validation

At `5632d90ad`: 3,491 host-runtime lib tests, 159 skippy-runtime, 724 skippy-server, 10 architecture regressions, 9 counting-allocator, 3 p99 budget, 3 trampoline-nonblocking, and 1,739 UI. Clippy with warnings denied, fmt, actionlint, and shellcheck clean. `scripts/tests` has 16 failures; I compared the failing set against `main`, not the count, and none are new.

The gate itself has never run, and it can't before merge: `pr_linux.yml` calls `ci-linux-lane.yml@main`, so the runtime slice resolves against `main` rather than this branch. The steps don't exist in the job that runs on this PR. First real execution is the first main push after merge.

What is checked: 19 tests in `scripts/tests/test_runtime_events_native_gate.py`, actionlint, and shellcheck. Seven of those drive the real script with a stub `cargo` and cover the branch that decides whether a green lane means anything, since `runtime_events_native.rs` reports a missing prerequisite as a pass. A blocked gate fails, a stale `executed` marker from a warm workspace can't rescue a blocked run, and the gate gets every prerequisite it needs. I broke the truncation and the marker check in turn and confirmed the right tests go red. The rest pin the lane wiring: CPU-only, and passing the bundle's parent directory rather than the runtime directory.

The YAML is the part that stays untested until it runs.

`cargo test --workspace` also shows `native runtime library has not been loaded` failures in `skippy-model-package` and one in `skippy-runtime`. Each passes when its crate runs alone, and a workspace run on `main` fails the same tests the same way.

## Open items

- No A0 (`production` vs `off`) run yet. I didn't want to automate a nightly for a number nobody has looked at once.
- The five `skippy_emit_*` functions are only called from native tests, so no live load produces a bit-31-to-36 fact. The Rust decode/submit path is complete and tested.
- `ModelLoadProgress`: severing the sole handoff at `load.rs:554` leaves the suite green, and its "at most one per 100 ms" ceiling is unfalsified because both observed loads emitted exactly one frame.
- The skippy generation adapter still takes a mutex before `submit`, outside the audited boundary, so `ingress_p99_us` doesn't cover it. Batching made it about 10× less frequent. Separate change.
- Relative-path redaction stays declined: `org/repo/file.gguf` is a canonical ref this codebase produces and is structurally identical to `models/secret/x.gguf`. A 14-case table test pins it.


i386 (dismissed):
Reviewed the SSE route layer (cursor grammar, reconnect/replay-gap semantics), the generation adapter (`runtime_events/mod.rs`), and the PR's extensive design docs. CodeRabbit has no open threads; CI green.

Verified in depth:
- The core rule — emitting never blocks the working thread — holds in the adapter: submissions are try_submit with bounded outcomes; reservation/slot handling is documented and tested for the duplicate-Started supersede case and the receipt-failure synthesis path.
- Terminal ordering (session_idle/stop_condition before terminals, prefill child before generation root) is explained with the reducer's OperationSettled semantics and backed by dedicated tests.
- Cursor parsing is strict (canonical UUID + no-leading-zero u64), replay gaps are explicit frames rather than silent skips, and the wire format is contract-pinned.

Not verified line-by-line: the full 316-file/66k-line diff. The contract test suites (`runtime_event_api_tests`, engine tests) carry more assurance than another read would.

i386 (approved):

i386 (changes requested):
I will have an agent re-review when you say the word @ndizazzo

i386 (approved):
Agent re-review (requested by @i386 in-thread, after @ndizazzo's two post-hold commits) — **approve**.

This was a fresh full review, not a delta from my #1667 pass (that one covered only the SSE routes + generation adapter). 316 files, base 22998c273 → tip a4f3ae70e.

**Post-hold delta** (ccd65a97f, a4f3ae70e): mechanical only — adds `commit_convention` to the expected quality-slice job set, adapts a test helper to the upstream `RuntimeSelection` signature, and regenerates the console-print ratchet line move 488→490 (correct per the ratchet's line-pinning rule). Quality lane ran the new job green at the tip.

**Verified locally at a4f3ae70e:**
- ABI pairing is correct: native `SKIPPY_ABI_VERSION_PATCH` 53→54 bumped exactly once (patch 0024, atomically with bit 31), `crates/skippy-ffi` `ABI_VERSION_PATCH = 54` at tip. Family bits 31–36 sequential, one per patch; Rust constants u64-typed so bit 31+ can't truncate.
- Patch queue: 33/33 apply clean at the unchanged pin `b31b71f3a`. Native cpu static + dynamic builds compile clean; all seven new symbols exported.
- Rust suites vs that native build: engine `runtime_events` 370/370, architecture 10/10, no-alloc reservoir 3/3 + budget 9/9, trampoline-nonblocking 3/3, off-mode/event-disabled 2/2.
- Non-blocking claim re-derived from source, not just the tests: ingress is a lock-free `ArrayQueue` with a credit budget; the terminal exemption is bounded by the write-once per-slot terminal claim (CAS in `claim_terminal`, generation bump on release); the drain park holds only `drain_gate`; the `AuditedMutex` lock audit panics on any producer-context blocking acquisition and `PRODUCER_ALLOWED` is empty.
- **The native gate ran for real — first execution anywhere, including CI.** CI structurally cannot run it until merge (`ci-linux-lane.yml@main`). I built the patched tree as a dynamic bundle and ran both the gated test and `scripts/ci-runtime-events-native-gate.sh` verbatim against a real model: **pass**, evidence `executed`, exact ABI admission enforced (0.1.54), capability probe `confirmed=0x1f81000000` (family bits 32–36 all confirmed by a real runtime), 2 structured production callbacks + 2 unload callbacks observed through the ring. That also retires the PR-body caveat that no live load ever produced a bit-31..36 fact — model-load and unload families now have observed production callbacks; kv/device/diagnostic emitters remain native-test-only as documented.
- Python suites: gate 19/19, toolchain-routing + runner-selection 17/17, product-composer + sdk-producers 25/25.
- Merge drift: `main` advanced to 7144cdf8f (Windows hardware parser fix) — zero file overlap, merge-tree clean.
- The `plugin/config.rs` 1504-line reduction is a pure test-extraction move.

**Not blocking, already visible in "Open items":** no A0 production-vs-off total-cost run yet; the CI YAML step itself is the one thing still unexercised until first main push (wiring pinned by the 19 Python tests); the generation adapter's pre-submit mutex stays outside the audited boundary; relative-path redaction stance unchanged.

The PR body's honesty about withdrawn #1667 claims and unmeasured costs is itself a reason to trust the rest of it. Evidence trail: `WORK_LOGS/2026-09-13_PR1777_REREVIEW_NATIVE_GATE_FIRST_RUN.md` on my side.

PR #1779, opened 2026-09-11
fix(ci): Fly console image build: pin pnpm and copy the whole crates tree

ndizazzo:
## Original problem

The manual [Deploy Fly Console](https://github.com/Mesh-LLM/mesh-llm/actions/workflows/fly-deploy-console.yml) workflow fails on `main`. Two separate build-context problems, found one after the other:

1. Run [34556490498](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34556490498) dies in the `ui-builder` stage:
   ```
   #13 [ui-builder 5/7] RUN ... pnpm install --frozen-lockfile
   #13 1.228 Error: ERR_PNPM_PNPM_ENGINE_IDENTITY_UNVERIFIABLE
   #13 1.228   ╰─▶ Cannot verify the identity of the @pnpm/exe.linux-x64 native binary: it
   #13 1.228       is missing from pnpm-lock.yaml.
   ```
2. With that fixed, run [34556833615](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34556833615) dies in `cargo build --locked`:
   ```
   error: failed to load manifest for workspace member `/src/crates/mesh-llm`
     failed to read `/src/crates/skippy-package-format/Cargo.toml`
   ```

`docker/Dockerfile.client` has the same `ui-builder` stage and the same (worse) crate-list drift. Nothing in CI builds either Dockerfile end to end; `docker.yml` only runs `buildx build --check`, which doesn't catch either failure.

## Diagnostics

**pnpm.** Both Dockerfiles ran an unpinned `npm install -g pnpm`, which now resolves to pnpm 12.3.4. `crates/mesh-llm-ui/package.json` pins `packageManager: pnpm@10.34.5`, so pnpm 12 tries to self-switch and refuses because a pnpm 10 lockfile doesn't carry the `@pnpm/exe` integrity entry it checks. The last successful deploy was July 14, before pnpm 12 shipped. Reproduced outside Fly in `node:24.20.0-alpine` with the three files the stage copies.

**Crate list.** `fly/Dockerfile` hand-listed 60 `COPY crates/<name>/` lines and was missing `skippy-package-format`, `skippy-scheduler`, and `mesh-llm-log-store` (all added since July). `docker/Dockerfile.client` was missing 17. #1241 fixed exactly this drift in August.

## Fix

- Install the pnpm version named by `packageManager` instead of latest, in both Dockerfiles. The install moves after `COPY package.json ...` so it can read the field, using the same `node -p` JSON read the fly stage already uses for `VITE_APP_VERSION`.
- Replace the per-crate COPY lists with a single `COPY crates/ crates/` in both Dockerfiles. `.dockerignore` already excludes `node_modules`, `dist`, and `target`, so the build context is the same size and any crate edit invalidates the same downstream layers it did before. The list just can't go stale again.

No runtime or protocol impact. `ci/linux-test.dockerfile` also hand-lists crates; I left it alone since CI exercises it directly and it isn't part of this failure.

## Validation

- [x] UI changes include screenshots or video, or explain why visual aids are not needed. (No UI change.)

Local, on this branch:

- `docker buildx build --check` on both Dockerfiles: no warnings.
- `docker buildx build --target ui-builder -f fly/Dockerfile .` succeeds; the image reports `pnpm 10.34.5`, `node v24.20.0`, and a populated `dist/`.
- A throwaway stage with the builder's exact COPY set (`Cargo.toml`, `Cargo.lock`, `crates/`, `tools/xtask/`) runs `cargo metadata --locked --no-deps` clean across all 63 workspace crates.

On Fly:

- The pnpm fix alone got run 34556833615 through the UI stage and into `cargo build` (where the crate-list failure surfaced).
- Full deploy from this branch: see the linked run below once it finishes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated UI build environments to use the project’s specified pnpm version.
  * Improved build reliability by avoiding automatic installation of incompatible pnpm versions.
  * Improved containerized build consistency by ensuring required project components are included during build preparation.
  * Reduced the risk of build failures caused by differences between local and container environments.

### DIFF
--- a/docker/Dockerfile.client
+++ b/docker/Dockerfile.client
@@ -16,10 +16,14 @@
 
 FROM node:24.20.0-alpine AS ui-builder
 ARG TARGETARCH
-RUN --mount=type=cache,id=mesh-llm-npm-alpine-node24-${TARGETARCH},target=/root/.npm,sharing=locked \
-    npm install -g pnpm
 WORKDIR /app
 COPY crates/mesh-llm-ui/package.json crates/mesh-llm-ui/pnpm-lock.yaml crates/mesh-llm-ui/pnpm-workspace.yaml ./
+# Install the pnpm pinned by package.json's packageManager field. An unpinned
+# `npm install -g pnpm` resolves to pnpm 12, which self-switches to the pinned
+# 10.x and then fails with ERR_PNPM_PNPM_ENGINE_IDENTITY_UNVERIFIABLE because a
+# pnpm 10 lockfile does not record the @pnpm/exe integrity that pnpm 12 checks.
+RUN --mount=type=cache,id=mesh-llm-npm-alpine-node24-${TARGETARCH},target=/root/.npm,sharing=locked \
+    npm install -g "$(node -p 'JSON.parse(require("fs").readFileSync("package.json", "utf8")).packageManager')"
 RUN --mount=type=cache,id=mesh-llm-pnpm-store-alpine-node24-${TARGETARCH},target=/root/.local/share/pnpm/store,sharing=locked \
     pnpm install --frozen-lockfile
 COPY crates/mesh-llm-ui/ ./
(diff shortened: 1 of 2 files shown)


i386 (approved):
Reviewed the full Dockerfile diff at 6cb48ea. The packageManager-derived pnpm pin is correct, the full crates-tree copy includes all 64 workspace members (including the three that broke the Fly build), and .dockerignore excludes generated build artifacts. The exact head SHA deployed successfully to Fly and all required checks pass. No blocking findings.

PR #1782, opened 2026-09-11
feat(skippy-cache): persist native runtime KV pages verbatim

i386:
## Problem

Manifest v4 can identify each segment independently, but L3 still stamped every segment as `raw/1`. That left the native runtime KV control arm implicit and allowed a fixed-size cut to mix native KV bytes with recurrent state.

## Result

- Add the exact `native-kv-page/1` segment identity and persist F32, F16, Q8_0, and Q4_0 runtime exports byte-for-byte, with no storage transcode.
- Cut mixed payloads at the KV/recurrent boundary so native KV segments never include exact auxiliary continuation state.
- Reject unknown native versions, incomplete descriptors, boundary-crossing segments, and representation mismatches before assembly.
- Copy manifest descriptor metadata into `L3Location` so the server validates the native page descriptor and located prefix before reading segment bytes; recheck metadata after load to close the locate/load race.
- Document native passthrough as the exact control for the remaining CacheGen quality and performance gate.

Advances #1652. This is stacked on #1752.

## Validation

Exact candidate commit: `9be4e30158a34612c818aeef4d47b2a045c62ff5`

- `cargo test -p skippy-cache --lib` — 170 passed, 0 failed, 2 ignored
- `cargo test -p skippy-server --lib` — 708 passed, 0 failed, 3 ignored
- `cargo clippy -p skippy-cache -p skippy-server --all-targets -- -D warnings`
- `git diff --check`

Matched realistic-segment stopwatch on the same Apple Silicon host and debug toolchain:

- Command: `cargo test -p skippy-cache --lib l3::tests::eviction_cost_at_realistic_segment_counts -- --ignored --nocapture`
- Workload: 20 manifests × 9,504 refs, 43,347,990 manifest bytes
- Baseline `aa533c0d45e1b8af025ee9af9887de68a7f02baf`: build 2,021 ms; budget enforcement 1,375 ms
- Candidate `9be4e30158a34612c818aeef4d47b2a045c62ff5`: build 2,113 ms; budget enforcement 1,396 ms

This slice does not change admission, eviction, prefix reuse, or resident capacity, so the competitive c64/c128/c256 policy matrix is unchanged.

### DIFF
--- a/crates/skippy-cache/README.md
+++ b/crates/skippy-cache/README.md
@@ -122,6 +122,16 @@ continuation state for those models. That is where the largest wins come from:
 llama-server warm slots still have to reprocess the recurrent prefix in many
 request shapes, while Skippy can restore the exact compact state and decode.
 
+Durable L3 manifests identify runtime-native KV segments as
+`native-kv-page/1`. Those segments are the bytes exported by the active runtime
+and are stored and restored verbatim; F32, F16, Q8_0, and Q4_0 remain in their
+active runtime representation when that backend supports the type. Recurrent
+and other auxiliary continuation state stays `raw/1`, with a hard segment
+boundary between the representations. Before reading native segment bytes,
+the server checks the manifest's runtime page descriptor against the located
+prefix and encoded KV length. The exact-state identity already binds the
+runtime ABI, platform, model, layer range, and KV configuration.
+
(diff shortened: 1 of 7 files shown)


i386:

PR #1788, opened 2026-09-11
feat(skippy): add LMCache-compatible CacheGen reference and gate

i386:
Issue #1652 requires an experimental lossy KV representation to prove quality and end-to-end value against native passthrough before request-path work begins. The earlier Mesh-owned 4-bit affine prototype produced only 12/64 matching continuation tokens, so this PR now replaces that gate path with a Rust-owned port of LMCache CacheGen revision `b5d109ea99a89b4d8a670ee4fc2e8cb76411ee5c`.

The port reproduces LMCache's per-token maximum scaling, 32/16-bin K/V layer schedule, per-channel normalized CDFs, 32-bit arithmetic coder, and 256-token chunk limit. It retains Apache-2.0 attribution and uses a bounded portable envelope in place of Python pickle. Independent Python scalar fixtures pin both 16-bin and 32-bin streams, and Rust tests require byte-for-byte encoder and decoder agreement. The runtime gate keeps K/V geometry, transposed V handling, exact auxiliary state, persistence, import, and 64-step continuation measurement unchanged.

The matched release run at exact implementation commit `4bf0865be6e7a13dbf7159b0fb6e6c41f29db72c` used the pinned Qwen3 0.6B Q8_0 artifact (`sha256:12fae8b8f78f0360b498d04c8db7d33aff29ab7d8080231f93a17c18119e6735`) with a 19,000-token prefix on Apple M1 Ultra/Metal:

- Quality recovered from 12/64 to 64/64 matching greedy tokens (100%), clearing the 95% threshold.
- Storage fell from 2,179,072,000 bytes to 446,903,003 bytes (20.51% of native; 4.88x smaller).
- Scalar CPU decode took 41,285.73 ms, so restore-to-first-token was 41,411.13 ms versus 374.38 ms native and the gate correctly stopped promotion.
- Steady-state p99 decode was 8.59 ms versus 33.24 ms native; the failure is the scalar restore codec cost.

| Metric | Native KV | LMCache-compatible CacheGen | Result |
|---|---:|---:|---|
| Persisted bytes | 2,179,072,000 | 446,903,003 | CacheGen is 20.51% of native (4.88x smaller) |
| Write / encode | 2,402.91 ms | 20,732.86 ms encode + 693.58 ms write | CacheGen fails persist latency |
| Read | 281.06 ms | 58.83 ms | CacheGen reads 4.78x faster in this same-run measurement |
| Codec decode | — | 41,285.73 ms | CacheGen scalar reference is the bottleneck |
| Restore to first token | 374.38 ms | 41,411.13 ms | CacheGen fails (110.61x slower) |
| Greedy-token agreement | 64/64 control | 64/64 (100%) | CacheGen passes the 95% quality gate |
| Continuation throughput | 119.85 tok/s | 124.17 tok/s | No steady-state regression |
| p99 decode | 33.24 ms | 8.59 ms | CacheGen stays within the 5% regression budget |
| Estimated codec working bytes | — | 2,627,965,638 | Reported; no memory cap supplied |

The result establishes a deterministic correctness oracle for a parallel device kernel. Native `native-kv-page/1` remains selected, and this PR adds no request-path wiring.

Validation at head `e9d4cc4b90d58f88da251bddb39c5dbcdde1c292`:

- `cargo test -p skippy-cache` — 176 passed, 2 ignored
- `cargo test -p skippy-correctness` — 84 passed, 247 ignored
- `cargo clippy -p skippy-cache -p skippy-correctness --all-targets -- -D warnings`
- `git diff --check`
- matched release 19K gate; compact evidence committed under `docs/skippy/`

Stacked on #1782 because the gate compares against its native passthrough control.


i386:

Issue #1732, opened 2026-09-09
KV roadmap rollup: Skippy KV cache tiers, codecs, reuse, and observability


i386:
Rollup issue for the KV cache roadmap coordinated in Buzz `#skippy-kv`. Rows with active PRs follow the current `kvcache-ng` stack order; detailed scope and acceptance status live in each linked issue and PR. Status applies to the linked PR slice: **Reviewed** means the current GitHub review decision is approved, **Code complete** means an implementation PR exists but is not approved yet, and an explicit CI qualifier names any execution gate that is still outstanding.

| Issue | PR | Status | Description | Enables |
|---|---|---|---|---|
| #1576 | #1632 | Reviewed | Configurable exact node-local L3 with hard budgets, status, lifecycle controls, and safe cold fallback. | Durable restart reuse and the foundation for every exact-tier item below. |
| Supporting work | #1707 | Reviewed | Raise the recurrent prefill micro-batch default so CUDA recurrent models clear the native SSM SSD kernel gate. | Representative recurrent-model qualification and fairer performance gates. |
| #1647 | #1710 *(first slice)* | Code complete | Add the restart replay harness, provenance, tier instrumentation, and matched promotion gates. | Evidence-based promotion decisions for #1648 through #1654. |
| Supporting work | #1714 | Code complete | Preserve resident L1 reuse as the in-process fast path when disk L3 is enabled. | Disk L3 without a warm-path regression and a reliable warm control arm. |
| #1648 | #1726 *(first slice)* | Code complete | Replace file-per-segment storage with packed append-only storage and drain durable writes on shutdown. | Locality-aware reads, reliable restart checkpoints, and the base for #1649. |
| #1649 | #1736 *(three slices)* | Reviewed | Stream packed ranges into the final payload, hash them in-flight, and prewarm the full-stage generation graph; the matched 3,994-token median is 394.5 ms cold versus 109.5 ms L3 (3.60x). | Clears the separate 2x restart target for the measured shape and provides block/range movement for #1652 and #1427. |
| Supporting work | #1743 | Code complete | Land dormant Linux, macOS, and Windows durable-L3 workflow plumbing while keeping the protected CI catalog unchanged. | Safe catalog activation in #1712 after the feature stack reaches `main`. |
| Supporting work | #1712 | Code complete | Activate qualification for the composed durable-L3 stack on Linux CPU, macOS Metal, and Windows CPU. | Cross-platform promotion evidence for the local tier. |
[...]

PR #1792, opened 2026-09-11
feat(skippy): define CacheGen device archive boundary

i386:
## Original problem

The faithful CacheGen reference in #1788 proves quality and compression, but restore still materializes the complete decoded KV page in host memory before importing it. On the measured 19K workload, scalar decode takes 41.29 seconds and dominates restore-to-first-token. Metal, CUDA, and ROCm kernels need one shared, validated archive contract before they can decode directly into resident Skippy K/V tensors.

## Diagnostics

Tracing the active path showed that `StageSession::import_kv_page` accepts a host byte slice and the native runtime uploads that buffer with `ggml_backend_tensor_set`. The correctness gate also privately owned the full-page archive format, so a native backend could not consume the same records without duplicating parsing and validation.

## Fix

This PR establishes the portable side of the Skippy-owned device boundary:

- moves the complete-page CacheGen archive into `skippy-cache` and exposes encode, validate, and scalar decode APIs;
- validates whole-page coverage, record geometry, nested LMCache segments, exact auxiliary state, and bounded transposed-V destinations before decode or dispatch;
- infers the physical V row width for transposed native pages, whose exported descriptor legitimately reports a zero row stride;
- switches the correctness gate to the shared archive implementation, removing its duplicate parser and codec logic;
- records the native integration design: Skippy owns allocation, rollback, synchronization, and session commit, while Metal and CUDA/HIP expose optional backend-registry codec hooks.

This slice does not select CacheGen in the request path. It supplies the format and validation boundary used by the subsequent native ABI and device-kernel patches.

## Validation

- [x] `cargo test -p skippy-cache --lib` - 181 passed, 2 ignored
- [x] `cargo test -p skippy-correctness` - 84 passed, 247 ignored
- [x] `cargo clippy -p skippy-cache --all-targets -- -D warnings`
- [x] `cargo clippy -p skippy-correctness --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] `git diff --check`
- [x] No UI changes

Stacked on #1788. Tracks #1652 and the KV roadmap in #1732.

### DIFF
--- a/crates/skippy-cache/src/cachegen/archive.rs
+++ b/crates/skippy-cache/src/cachegen/archive.rs
@@ -1,0 +1,773 @@
+//! Bounded CacheGen archives for complete native KV pages.
+//!
+//! A page archive covers every byte in the decoded runtime payload exactly
+//! once. K/V records carry LMCache-compatible segments while auxiliary
+//! indexer state remains exact. The archive is portable; native backends may
+//! decode its validated records directly into resident KV tensors.
+
+use std::collections::BTreeMap;
+
+use anyhow::{Context, Result, anyhow, bail};
+
+use super::lmcache::{
+    MAX_TOKENS_PER_CHUNK, bins_for_layer, decode_f16_segment, encode_f16_segment,
+    validate_f16_segment,
+};
+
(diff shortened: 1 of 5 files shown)


i386:

PR #1794, opened 2026-09-11
feat(skippy): expose CacheGen native import contract

i386:
CacheGen archives currently stop at a Rust-owned portable representation, so native Skippy has no stable contract for backend-resident decode. This adds that contract and wires validated archive records through `skippy-runtime` to the patched native runtime.

The new ABI describes compressed F16, exact auxiliary, and transposed-F16 records with their logical page offsets and tile geometry. The native surface validates record descriptors and returns an explicit unsupported status until a resident decoder is available; it does not allocate cache cells or mutate session state on that path. The Skippy ABI patch advances to 0.1.54, Rust bindings cover static and dynamic runtimes, and the generated native API reference is current.

This PR is stacked on #1792. The following backend PRs will add the Skippy-owned allocation/rollback transaction and direct Metal/CUDA/HIP decode. No scalar fallback or legacy compatibility path is introduced.

Validation:
- `LLAMA_STAGE_BACKEND=cpu LLAMA_STAGE_LINK_MODE=static scripts/build-llama.sh`
- `python3 scripts/generate-skippy-api-doc.py --check`
- `cargo test -p skippy-ffi --all-features` (12 passed)
- `cargo test -p skippy-runtime` (126 passed, 1 ignored)
- `cargo clippy -p skippy-ffi -p skippy-runtime --all-targets --all-features -- -D warnings`

The all-features `skippy-runtime` test invocation is not a supported standalone lane because enabling `dynamic-runtime` requires a loaded native runtime bundle; 123 tests pass and three pre-existing native-call tests fail at the unloaded-library guard. The default static package suite above passes completely.

Tracks #1732.


i386:

PR #1796, opened 2026-09-11
feat(skippy): restore CacheGen pages directly on Metal

i386:
The native CacheGen ABI currently stops at `SKIPPY_STATUS_UNSUPPORTED`, so a validated archive still has to be expanded through the slow scalar Rust path before native import. This implements the Skippy-owned restore transaction and the first resident backend decoder.

Native Skippy now validates every record against the live KV layout and preflights one backend hook before allocating cells. It groups all tiles for each resident K or V tensor, preserves exact `k_idx` bytes in the same transaction, supports single and ISWA composite pages, and rolls back cache metadata if device decode fails. CacheGen import remains fresh-session only and fails closed when the resident backend lacks a decoder.

The Metal backend decodes LMCache-compatible 16-bin and 32-bin arithmetic streams directly into the allocated F16 KV cells. One command buffer covers the restore batch, with one dispatch per resident tensor; transposed V and noncontiguous physical cell indices are handled by explicit destination strides. There is no scalar fallback or compatibility path.

This PR is stacked on #1794. CUDA/HIP and F32/Q8_0/Q4_0 typed adapters remain follow-up slices before #1652 can close.

Validation at `7233ac719b7e700c59d690572e921ed342524e1b`:
- clean `scripts/prepare-llama.sh pinned` replay through all 27 maintained and generated patches
- `LLAMA_STAGE_BACKEND=cpu LLAMA_STAGE_LINK_MODE=static LLAMA_STAGE_BUILD_TESTS=ON scripts/build-llama.sh` (33/33 native Skippy tests)
- Apple M1 Ultra Metal test against pinned LMCache 16-bin and 32-bin fixtures, covering row-major K, transposed V, noncontiguous cells, grouped jobs, exact byte parity, and unsafe destination rejection
- `cargo test -p skippy-ffi --all-features` (12 passed)
- `cargo test -p skippy-runtime` (126 passed, 1 ignored)
- `cargo clippy -p skippy-ffi -p skippy-runtime --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`

Tracks #1652 and #1732.


i386:

PR #1798, opened 2026-09-11
feat(skippy): decode CacheGen pages on CUDA and ROCm

i386:
CacheGen restore still falls back to a 41-second scalar decode on discrete GPU hosts because only the Metal backend implements the native Skippy decode hook. This adds one F16 arithmetic decoder that llama.cpp compiles through CUDA and HIP, so validated CacheGen records can be decoded directly into the resident K/V cells allocated by the Skippy transaction.

The backend stages compact payload, tile, prefix, and exact cell-index buffers; launches every tensor job on the per-thread stream; and synchronizes once per physical device. It preserves row-major K and transposed/noncontiguous V addressing through the native token and channel strides. Allocation, upload, launch, execution, and cleanup failures return through the existing backend error channel so Skippy can roll back the uncommitted cells. MUSA does not advertise the hook, and there is no scalar fallback.

Two follow-up fixes are included at the current head. Metal now aligns every staged tile before typed metadata reads, with a two-tile regression fixture covering encoded16 and encoded32 records. Metal and CUDA/HIP also replace the arithmetic hot loop's 64-bit division with a reciprocal estimate plus exact integer correction and replace the linear CDF scan with binary search. The optimized Metal implementation remains bit-exact on the fixtures and reduced the 19K direct resident import from 744.95 ms to 540.74 ms.

This is stacked on #1796. It advances #1652 and the #1732 roadmap.

Validation:
- clean replay of all 30 pinned llama.cpp patches
- native Metal `test-skippy-cachegen-metal` passes exact single- and multi-tile fixtures
- static Metal `ggml-metal` target rebuild and linked direct-device runtime gate passed
- 64/64 token agreement in the 19K Apple M1 Ultra device gate at the optimized head
- initial explicit [CUDA 12.9.2 and ROCm gfx1100 compile/package run](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34572596586) passed at `b3c9a2c690d4ccb87b99db2b0698c43af5f42888`
- exact optimized-head [CUDA 12.9.2 and ROCm gfx1100 compile/package rerun](https://github.com/Mesh-LLM/mesh-llm/actions/runs/34577082594) is in progress at `fb9b06fe1cd0c2028136f19bae1b210137e56e05`
- real NVIDIA and AMD fixture parity remains a separate hardware gate

### DIFF
--- a/third_party/llama.cpp/patches/0028-ggml-cuda-decode-CacheGen-pages-into-resident-KV.patch
+++ b/third_party/llama.cpp/patches/0028-ggml-cuda-decode-CacheGen-pages-into-resident-KV.patch
@@ -1,0 +1,481 @@
+From 3391262f8620a3d1a7ddb6a174e75e0c6129c804 Mon Sep 17 00:00:00 2001
+From: Mesh-LLM CI <ci@mesh-llm.local>
+Date: Fri, 11 Sep 2026 16:58:10 +1000
+Subject: [PATCH] ggml-cuda: decode CacheGen pages into resident KV
+
+Share the F16 decoder across CUDA and HIP, batch launches before synchronization, and write directly into the allocated Skippy cell layout.
+
+Assisted-by: scama
+---
+ ggml/src/ggml-cuda/cachegen.cu  | 111 +++++++++++++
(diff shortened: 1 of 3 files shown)


i386:

Issue #1427, opened 2026-08-24
Add phase-disaggregated serving with streamed prefill-to-decode state handoff


i386:
## Goal

Add a phase-disaggregated Mesh serving mode in which prefill and decode may run on different machines or topologies:

- a high-throughput prefill worker/topology processes the prompt;
- exact continuation state is streamed to the decode worker while prefill is still running;
- the decoder imports and atomically commits that state; and
- generation continues locally on the decode worker, including native MTP and suffix N-gram speculation.

Initial lab target:

- M3 Ultra 256 GB: prefiller;
- M1 Ultra 128 GB: decoder;
- Nemotron 3 Super Q4 + MTPv2 package from #1425; and
- high-bandwidth, low-latency local link.

This is phase disaggregation, not ordinary layer splitting.

## Motivation

Current layer splitting distributes every decode token across the stage pipeline. That retains a network dependency on every generated token.

Phase disaggregation instead pays one prompt-state transfer:

```text
Prompt ──► M3 prefill ──► streamed exact state ──► M1 decode + MTP/suffix
                  │
                  └── begins transferring before prefill completes
```

Once the handoff commits, decode is local to the M1. This can:

- use the faster/larger M3 for prompt ingestion;
- remove per-token decode traffic between machines;
- overlap prefill for a later request with decode of an earlier request;
- retain the M1 as the decode/MTP owner; and
- improve aggregate goodput when prompt and generation phases overlap.

It will not necessarily improve a single short request. The topology must only be selected when measured state-transfer economics predict a benefit.

## Key constraint: this is hybrid state, not just KV

Nemotron 3 Super contains:

- 8 attention layers;
- 40 Mamba recurrent layers; and
- 40 MoE layers.

An exact handoff must include every state component required for continuation:

- attention KV pages;
- recurrent/SSM state;
- convolution state;
- logical token position and sequence metadata;
- target sampler/session metadata where required;
- native-MTP context or sufficient state to reprime it exactly;
- native-MTP session bookkeeping and pending hidden row;
- committed token history required by the request-local suffix proposer; and
- an immutable identity tying the payload to the model, package, tokenizer, context configuration, and state schema.

A KV-only handoff is not correct for Nemotron-H.

## State-transfer economics

For Nemotron Super with F16 KV:

```text
8 attention layers
× 2 (K and V)
× 2 KV heads
× 128 dimensions
× 2 bytes
≈ 8 KiB per prompt token
```

[...]

PR #1799, opened 2026-09-11
test(skippy): gate CacheGen on direct device restore

i386:
The CacheGen acceptance gate was still measuring a scalar host decode and then importing a multi-gigabyte reconstructed page, so it could not prove the native device path or produce a valid restore-to-first-token comparison. This changes the gate to pass the persisted compressed archive directly to Skippy's native import transaction. Unsupported backends fail through the native capability check; there is no scalar restore fallback.

Native and CacheGen continuations now run in isolated sessions to avoid shared-backend KV interference. The scalar decoder remains an independently timed oracle, but its output is discarded and its time is excluded from device TTFT. The report names the restore path and records `scalar_oracle_decode_ms` explicitly.

On the pinned Qwen3 0.6B Q8_0 model with a 19,000-token F16 K/V page on Apple M1 Ultra, the direct Metal path passes compression and quality but correctly fails the local-tier latency gate:

| Metric | Native | Direct Metal CacheGen |
|---|---:|---:|
| Persisted bytes | 2,179,072,000 | 446,903,003 (4.88x smaller) |
| Read | 296.25 ms | 54.21 ms |
| Resident import | 56.69 ms | 540.74 ms |
| Restore to first token | 385.41 ms | 604.84 ms |
| Greedy-token agreement | control | 64/64 (100%) |
| p99 decode | 32.47 ms | 9.88 ms |

The backend optimization in the parent PR reduced direct import from 744.95 ms to 540.74 ms. The remaining loss comes from reconstructing roughly 1.089 billion F16 values; the 242.04 ms read saving does not offset the 484.05 ms extra resident-import work on this unified-memory local tier. The result does not decide the remote tier tracked by #1427.

This is stacked on #1798 and advances #1652 and #1732.

Validation:
- full `skippy-correctness` package suite passed: 84 passed, 247 ignored
- `cargo clippy -p skippy-correctness --all-targets -- -D warnings`
- `cargo check -p mesh-llm`
- `cargo fmt --all --check`
- `just no-console-print`
- `git diff --check`
- exact 19K device gate at `c86cf848b0fa183f2c2e594fe1e51bf1024d7185`: 64/64 agreement and 4.88x compression; only the declared local restore-latency criterion failed


i386:

PR #1800, opened 2026-09-11
feat(skippy): add typed CacheGen page adapters

i386:
The portable CacheGen page archive only accepted F16/F16 runtime pages, so selecting F32, Q8_0, Q4_0, or mixed K/V disabled the codec before a backend could even advertise a typed implementation. This adds typed edge adapters around the existing LMCache-compatible entropy stream while keeping K and V independently described.

F32 and F16 scalar rows, Q8_0 and Q4_0 block rows, mixed Q8_0/Q4_0 pages, and transposed F32/F16 V now encode and decode through the pure-Rust archive. The inner CacheGen segment remains the shared F16 algorithm; record kinds identify how native rows are unpacked before encode and repacked after decode. Quantized row geometry is validated against the GGML 32-value block layouts.

This slice establishes the portable representation and scalar oracle. Typed records still fail before FFI because the native Metal/CUDA/ROCm store adapters are the next stack slice. There is no scalar fallback into a device restore transaction, and no typed backend is claimed or promoted by this PR.

This is stacked on #1799 and advances #1652 and #1732.

Validation at `debad5ace964baf8a946d78920aece79e92fa7fb`:
- `cargo test -p skippy-cache --lib`: 183 passed, 2 ignored
- `cargo test -p skippy-runtime --lib`: 127 passed, 1 ignored
- `cargo test -p skippy-correctness`: 84 passed, 247 ignored
- `cargo clippy -p skippy-cache --all-targets -- -D warnings`
- `cargo clippy -p skippy-correctness --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/crates/skippy-cache/src/cachegen/archive.rs
+++ b/crates/skippy-cache/src/cachegen/archive.rs
@@ -8,6 +8,7 @@
 use std::collections::BTreeMap;
 
 use anyhow::{Context, Result, anyhow, bail};
+use skippy_protocol::binary::{f16_bits_to_f32, f32_to_f16_bits};
 
 use super::lmcache::{
     MAX_TOKENS_PER_CHUNK, bins_for_layer, decode_f16_segment, encode_f16_segment,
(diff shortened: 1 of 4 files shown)


i386:

PR #1802, opened 2026-09-11
feat(skippy): restore F32 CacheGen pages on device

i386:
## Problem

The portable CacheGen archive can represent F32 K/V pages, but the native import path still exposes an F16-only backend hook. F32 records therefore stop in Rust and cannot restore directly into resident Metal, CUDA, or ROCm tensors.

## Change

- replace the F16-specific backend hook with one typed CacheGen decode contract
- add F32 row-major and transposed record kinds to the native ABI and Rust mapping
- decode through the existing F16 entropy stream, round to F16 exactly as the portable oracle does, and write the resulting value directly into F16 or F32 resident layouts
- retain Skippy-owned descriptor validation, cell allocation, synchronization, rollback, and session position commit
- reject quantized Q8_0/Q4_0 records before FFI until their device quantization passes land
- add a durable llama.cpp patch and device fixture coverage for F32 row-major/transposed writes

This intentionally removes the old `ggml_backend_cachegen_decode_f16` symbol rather than preserving a compatibility alias.

Stacked on #1800.

## Validation

At `b4e99486dfdd1559e510e40ce5e0089e962e910b`:

- full pinned llama.cpp patch stack replayed cleanly to `8fde8d47bb832f95f3f1c726173df47309d33ca0`
- `test-skippy-cachegen-metal` passed on Apple M1 Ultra after an exact native rebuild
- `cargo test -p skippy-cache --lib`: 183 passed, 2 ignored
- `cargo test -p skippy-ffi --lib`: 12 passed
- `cargo test -p skippy-runtime --lib`: 128 passed, 1 ignored
- `cargo test -p skippy-correctness`: 84 passed, 247 ignored
- Clippy passed for all four packages with warnings denied
- `cargo fmt --check` and `git diff --check` passed


i386:

PR #1806, opened 2026-09-11
refactor(mesh): give announcement building and merging their own module

### LINKED ISSUE #1674: Refactor gossip announcement merge and construction into an owning module
## Summary

Split the announcement merge and construction responsibilities from `crates/mesh-llm-host-runtime/src/mesh/gossip.rs` into a dedicated owning module.

## Rationale

`gossip.rs` is close to the repository Rust source-file size threshold. This refactor must be separate from behavior changes so reviewers can verify the move independently.

## Affected area

- `crates/mesh-llm-host-runtime/src/mesh/gossip.rs`
- The new owning module under `crates/mesh-llm-host-runtime/src/mesh/`

Move responsibilities such as:
- merging direct and transitive `PeerAnnouncement` data into `PeerInfo`;
- detecting meaningful peer changes; and
- constructing local and rebroadcast `PeerAnnouncement` values.

Keep the gossip transport, connection, stream, and peer lifecycle behavior in `gossip.rs`.

## Acceptance criteria

- `gossip.rs` is below the repository's 2,000-line threshold after the refactor.
- The new module has a focused ownership boundary and is not a catch-all module.
- Callers and visibility are updated without behavior changes.
- Existing tests for gossip announcement propagation and peer updates continue to pass.
- The refactor does not change the wire protocol.

## References

- Related pull request: https://github.com/Mesh-LLM/mesh-llm/pull/1673
- Related review comment: https://github.com/Mesh-LLM/mesh-llm/pull/1673#discussion_r3942005657
- Requested by: @Virgile-pct

Virgile-pct:
## Original problem

`gossip.rs` was at 1,984 lines against the repository's 2,000-line threshold (AGENTS.md), so the next change to it had to split something first. #1673 already hit that: adding the memory block to the transitive merge landed in a file with 16 lines of headroom, which is what prompted #1674.

The file also held two responsibilities that do not need each other: the gossip transport, and everything about announcements.

## What changed

Announcement work moves to `mesh/announcements.rs`:

- building this node's own `PeerAnnouncement` (`snapshot_local_announcement_data`, `build_local_announcement`, `plugin_inference_models`, `format_optional_locked_f32_list`) and the rebroadcast set (`collect_rebroadcast_announcements`, `announcement_from_peer`);
- folding direct and transitive announcements back into `PeerInfo` (`apply_transitive_ann`, `merge_first_joined_mesh_ts`, `backfill_legacy_descriptors`);
- deciding when a peer changed meaningfully (`peer_meaningfully_changed`) and which peers are eligible to relay (`version_allowed_for_rebroadcast`, `peer_is_idle_transitive_client`);
- the two carriers those functions pass around, `LocalAnnouncementData` and `RebroadcastAnnouncements`.

`gossip.rs` keeps connections, streams, the join probe and the peer lifecycle, and drops to 1,440 lines. The new module is 587.

## No behavior change

This is a move, and it is verifiable as one:

- Diffing each moved block against its origin on `main`, after normalizing only the import path, leaves nothing.
- Visibility widened only where the compiler required it: two fields of `RebroadcastAnnouncements` are now `pub(super)` because `gossip.rs` destructures the value it gets back. `LocalAnnouncementData` keeps all 27 fields private. I widened them first, then reverted and re-compiled to confirm they were not needed, since both its producer and its consumer moved together.
- `git diff main` touches no file under `crates/mesh-llm-protocol/` or `src/protocol/`, so the wire protocol is untouched as the issue requires.

## Validation

Windows 11. The 20 host-runtime failures are pre-existing on this checkout and unrelated (HOME/temp-path/snapshot set); the same 20 fail on `main`.

- [x] Targeted: `gossip announcement peer_state protocol_compat`: 99 passed / 0 failed
- [x] `mesh-llm-host-runtime --lib`: 2,990 passed / 20 failed (pre-existing)
- [x] `cargo clippy --no-deps --all-targets`: no warning in either file
- [x] `xtask repo-consistency no-console-print`: passes **without regeneration**, so no print site moved
- [x] `check-env-mutation-contract.py` and its unittest
- [x] Both files are under the thresholds: 1,440 and 587

## One acceptance criterion I did not meet

The issue asks the extracted module to own tests for the behavior it now owns. The tests stay in `tests/gossip/merge_and_refresh.rs`.

Moving them means extracting `test_addr`, `test_endpoint_id` and `test_announcement` out of `tests/gossip.rs`, which eight other test files reach through `include!`. That is a separate change to shared test scaffolding, and doing it in this PR would make the move harder to verify rather than easier, which is the opposite of what the issue asks for. Happy to do it as a follow-up, or here if you would rather have it in one go.

Closes #1674.

## Summary by CodeRabbit

* **New Features**
  * Added peer announcement handling for sharing model, hardware, capacity, cache, policy, ownership, and performance information across the mesh.
  * Added support for merging transitive announcements and rebroadcasting eligible peer updates.
  * Added compatibility handling for legacy served-model descriptors.

* **Bug Fixes**
  * Prevents stale or outdated peer announcements from being rebroadcast.
  * Safely handles unavailable plugin model information without disrupting announcements.


i386 (approved):
Verified this is a pure move: diffed each relocated block in `announcements.rs` against its deletion in `gossip.rs` — line-identical apart from the import path and the two `pub(super)` fields on `RebroadcastAnnouncements` that `gossip.rs` needs for destructuring. `LocalAnnouncementData` fields stay private, correct since producer and consumer moved together. `mod.rs` re-exports keep `backfill_legacy_descriptors` public path stable, and the wire protocol crates are untouched as the issue requires.

Agree the test relocation is better as a follow-up — extracting the shared `test_addr`/`test_endpoint_id`/`test_announcement` scaffolding out of the `include!` web belongs in its own diff where the move can stay verifiable.

Approving.

Virgile-pct:
Correcting one line in the description above, before a reviewer trips on it.

I wrote that the same 20 host-runtime failures occur on `main`. That was true of the run I validated with, but I only verified the full list afterwards, and it is worth being precise: re-running `cargo test -p mesh-llm-host-runtime --lib` a second time on this branch produced **21** failures, against 20 on `main` at the same commit. The extra one is:

```
mesh::connections::gated_relay_e2e_tests::qad_enabled_map_admits_endpoint_to_gated_relay
```

It is flaky under load here, not a regression:

- Run in isolation on this branch it passes, 3 runs out of 3 (`-- mesh::connections::gated_relay_e2e_tests`, 4 passed each time).
- It only appeared in the full ~3,000-test run, and did not appear in the earlier full run on this same branch that reported 20.
- This PR does not touch `mesh/connections.rs`, and the moved code has no path into the gated-relay admission map.

The other 20 match `main` exactly, which I checked by diffing the two sorted failure lists rather than by eye.

Flagging it so a reviewer who runs the suite locally on Windows and sees 21 does not spend time on it. If it is already known to be flaky on hosted runners, ignore this; if not, I am happy to open a separate issue with the two runs attached.

i386 (dismissed):
Second approval alongside i386 — reviewed the move: it's a clean extraction of announcement build/merge from gossip.rs into announcements.rs with no logic changes beyond visibility; gossip.rs keeps transport concerns. Module docs draw the right boundary.

ndizazzo:
> Second approval alongside i386

@i386 your agent has multiple personality disorder 🤣

PR #1808, opened 2026-09-11
test(skippy): qualify typed CacheGen restore at 19K

i386:
The CacheGen acceptance gate can now run the same 19K workload with independently selected K and V cache types, so typed device restore is measured against a matching native control instead of inferred from byte fixtures. That reporting exposed an end-to-end F32 defect: `GGML_TYPE_F32` is enum value zero, and native model loading treated zero as unset and silently created F16 caches. The native mapping now preserves explicit F32 while keeping the public default at F16.

The corrected F32/F32 run and both F32+F16 directions clear every declared local gate. The Q8_0/F32 boundary crossed between one pass and one latency stop, so it remains stopped; the lower-width rows also remain stopped, and Q4_0/Q4_0 expands the stored payload.

| K/V type | CacheGen/native bytes | Agreement | Native TTFT | CacheGen TTFT | Result |
|---|---:|---:|---:|---:|---|
| F32/F32 | 10.25% | 64/64 | 1,199.13 ms | 766.12 ms | pass |
| F32/F16 | 13.67% | 64/64 | 878.61 ms | 686.15 ms | pass |
| F16/F32 | 13.67% | 64/64 | 840.32 ms | 671.93 ms | pass |
| Q8_0/F32 | 20.50% | 64/64 | 800.57 / 727.17 ms | 796.94 / 792.13 ms | unstable: one pass, one latency stop |
| Q8_0/Q8_0 | 50.95% | 64/64 | 333.36 ms | 650.64 ms | latency stop |
| Q4_0/Q4_0 | 103.62% | 61/64 | 189.84 ms | 699.35 ms | size + latency stop |
| Q8_0/F16 | 33.89% | 64/64 | 558.28 ms | 792.05 ms | latency stop |
| Q4_0/F16 | 43.72% | 61/64 | 483.98 ms | 807.03 ms | latency stop |

## Architecture

- Adds `--cache-type-k` and `--cache-type-v` to `skippy-correctness state-handoff` and routes them through local and binary-control model configuration.
- Records the resolved K/V types in the CacheGen report.
- Applies cache types through the shared native context configuration path, keeps `skippy_runtime_config_default()` at F16, and regression-tests explicit F32 value zero.
- Keeps the scalar decoder as the excluded correctness oracle and the native device decoder as the measured restore path.
- Records the exact-head matrix and its limits in the backend plan and a compact JSON artifact.

## Validation

- `cargo test -p skippy-correctness` — full package: 84 passed, 247 ignored, 0 failed.
[...]

PR #1810, opened 2026-09-11
perf(ggml-metal): stage CacheGen payloads directly

i386:
Metal CacheGen restore copied every validated job into `NSMutableData` and then copied the same bytes again into shared `MTLBuffer`s before device decode. This writes payloads, tile descriptors, and stream prefixes directly into their final shared Metal buffers after a sizing/validation pass.

On the 19K Qwen3 0.6B gate, Q8_0/F32 import fell from a 580.45 ms two-run baseline to 491.92 ms across two post-change runs (15.25%). F16/F16 fell from 540.74 ms to 451.82 ms (16.44%). Both still fail the matched local TTFT gate, so promotion remains stopped. A 128-row diagnostic was also rejected: it doubled tile count, worsened compression from 20.50% to 30.84%, and raised import to 727.74 ms.

The change also adds the registered Metal CacheGen fixture to the full-replay build targets. Previously CTest discovered the fixture but the replay script had not built its executable.

Stacked on #1808.

## Validation

- Fresh pinned llama.cpp patch replay passed.
- Metal native full replay: 41/41 tests passed, including `test-skippy-cachegen-metal`.
- Full `skippy-ffi`, `skippy-runtime`, and `skippy-correctness` package suites passed.
- Clippy for all targets in those packages passed with warnings denied.
- Rust formatting, `just no-console-print`, JSON parse, and `git diff --check` passed.
- Exact-head 19K Metal runs preserved 64/64 agreement for Q8_0/F32 and F16/F16.

### DIFF
--- a/docs/skippy/CACHEGEN_BACKEND_PLAN.md
+++ b/docs/skippy/CACHEGEN_BACKEND_PLAN.md
@@ -271,6 +271,25 @@ reconstruction as the next Metal target. F32/Q8_0 is not a valid control on this
 model: llama.cpp requires Flash Attention for quantized V, while that mixed
 cache pair does not have a compatible Flash Attention kernel.
 
+An exact-head Metal staging optimization at
+`e2bdc935e24cbed3bff767762547098db6028459` removes the second host copy that
+previously occurred when each validated job was accumulated in `NSMutableData`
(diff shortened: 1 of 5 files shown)


i386:

Issue #1677, opened 2026-09-06
Support embedding, OCR, TTS and encoder-only model classes, and expose them through the OpenAI-compatible API


i386:
Requested by James in #skippy on 2026-09-06.

Skippy's staged runtime and its OpenAI-compatible frontend are both scoped to autoregressive text generation. Four model classes that llama.cpp already supports have no path through Skippy at all, and the API surface has no endpoint to expose them even if they did. This ticket covers both halves, because either one alone is unshippable.

## Current state (measured at `a93db64ac`)

**Model classes.** llama.cpp registers 149 architectures (`LLM_ARCH_NAMES`, `src/llama-arch.cpp`); Skippy's certified corpus covers 30. Among the uncovered, these four classes are structurally excluded rather than merely uncertified:

| Class | Architectures |
| --- | --- |
| Embedding / encoder-only | `bert`, `modern-bert`, `neo-bert`, `eurobert`, `nomic-bert`, `nomic-bert-moe`, `jina-bert-v2`, `jina-bert-v3`, `gemma-embedding`, `llama-embed` |
| Encoder-decoder | `t5`, `t5encoder` |
| OCR | `deepseek2-ocr`, `paddleocr` |
| TTS / audio | `qwen3tts`, `pockettts`, `talkie`, `wavtokenizer-dec` |

Six of these (`nomic-bert`, `nomic-bert-moe`, `jina-bert-v2`, `jina-bert-v3`, `llama-embed`, `t5encoder`, `qwen3tts`, `wavtokenizer-dec`) also have no stage-control transform in patch `0012-skippy-generate-model-family-stage-controls.patch`, so they cannot be staged today even in principle.

**Multimodal projectors.** `tools/mtmd/clip-impl.h` defines 62 projector types, 20 of them audio/speech/TTS. Skippy certifies 2 (`PROJECTOR_TYPE_QWEN2VL`, `PROJECTOR_TYPE_QWEN3VL`) and contributes 1 (`PROJECTOR_TYPE_INKLING`, added by patch 0008). Zero audio projectors are covered. Nothing in `tools/mtmd` carries stage-control edits — `grep -c 'stage_filter\|begin_block\|end_block\|skippy_stage' 0008-*.patch` returns 0 — so the projector runs whole on one node and only the language trunk is partitioned.

**API surface.** `crates/openai-frontend/src/router.rs` exposes exactly four OpenAI endpoints:

```
[...]

PR #1833, opened 2026-09-12
feat(skippy): serve non-chat GGUF workloads via OpenAI API

IvGolovach:
## Original problem

The native GGUF path assumed causal text generation, leaving embeddings,
cross-encoder reranking, and encoder-decoder models without an explicit
execution and routing contract. The same distinction is needed when exposing
image/audio workloads: a model name alone must not authorize an endpoint or
unsupported split placement. This implements the core workload-serving expansion
in #1677, not every architecture listed there.

This is broader than multimodal support: embeddings, reranking, and
encoder-decoder generation are text-only workloads too. OpenAI compatibility
describes the local HTTP interface and client SDK usage; inference runs in
Mesh-LLM's native runtime, not through a hosted OpenAI service. `/v1/rerank`
is an additional companion endpoint.

## Diagnostics

Loaded-model inspection distinguishes encoder-only models, cross-encoder
classifier heads, and encoder-decoder execution; model names alone are not a
reliable capability contract. Routing regressions cover descriptor ordering,
an incompatible local replica hiding a capable peer, mixed-workload MoA pools,
and binary/chunked multipart requests.

The latest native-compatibility regressions reproduced two problems before
the fix: accepting a pre-workload 0.1.60 library before resolving mandatory new
exports, and treating workload feature bit 37 as reserved. The corrected tests
cover ABI rejection, C/Rust version equality, every missing workload export,
and successful confirmation of the complete family.

## Fix

- Probe the loaded native model and execute embeddings, reranking, and
  encoder-decoder generation as full-model workloads. Expose `/v1/embeddings`,
  `/v1/rerank`, `/v1/audio/speech`, `/v1/audio/transcriptions`, and
  `/v1/audio/translations`; encoder-decoder and OCR use existing text/image APIs.
- Advertise workload metadata additively and enforce eligibility per concrete
  serving target, before ranking, affinity, reservations, MoA admission, and
  failover. An incompatible local replica cannot hide a compatible remote peer;
  legacy advertisements remain eligible for ordinary generation only.
- Preserve binary multipart bodies, bound audio uploads, and decode/rebuild
  chunked requests when rewriting `model=auto`. Stateless workloads do not
  acquire generation affinity from endpoint metadata such as `user`.
- Reject unsupported split execution, keep projectors colocated with their
  trunk, and serialize encoder-decoder execution against its shared context.
- Add six pinned smoke/oracle certification lanes and deterministic,
  source-bound CPU oracle provisioning in the llama bump canary. Validate
  evidence identity, actual output shape/order, and independent comparisons.
- Include the review hardening for cleanup, native output sentinels, ABI/layout
  checks, model-lifetime workload-probe caching, oracle deadlines, and fail-closed
  evidence. Permit safe Unix listener rebinding between certification phases
  without sharing an address with a live listener. Retain the focused fixture
  race corrections and semantics-preserving Clippy fixes needed by validation.

## Compatibility and limits

The mesh field is additive, with no persistent-data migration. Native libraries
are a separate lockstep boundary: **Skippy ABI 0.1.60 → 0.1.61**. The host and
packaged runtime must be rebuilt together; the loader rejects older libraries
before resolving mandatory workload exports. Bit 37 is confirmed only when all
four workload exports are present. Package manifests derive the same version
from the Rust facade, and the generated native API reference is synchronized.

Embeddings are L2-normalized; dimensions must match native width. Reranking
preserves original document indices. Encoder-decoder generation has one
effective lane and no tool calls. Native Qwen3-TTS currently requires
`voice: "default"` and explicit `wav` or `pcm`. Audio uploads are bounded at
64 MiB and support JSON/text responses. These representatives do not establish
support or quality for every model family. See `docs/NON_CHAT_MODELS.md`.

## Native ABI inventory

<details>
<summary>Four added exports, descriptor types, and feature bit</summary>

All four functions below are **added**, not signature changes. They are declared
in `include/skippy/workloads.h`, implemented in `src/skippy/workloads.cpp`, and
mirrored in Rust static/dynamic bindings under `crates/skippy-ffi/src/`.

| Added declaration | Purpose |
|---|---|
| `enum skippy_status skippy_model_workload_info_v1(const struct skippy_model *model, struct skippy_workload_info_v1 *out_info, struct skippy_error **out_error)` | Runtime workload classification; callers initialize descriptor version and size. |
| `enum skippy_status skippy_session_embed(struct skippy_session *session, const llama_token *token_ids, size_t token_count, float *output, size_t output_capacity, size_t *out_dimensions, struct skippy_error **out_error)` | Normalized embedding output with explicit buffer bounds. |
| `enum skippy_status skippy_session_rerank(struct skippy_session *session, const char *query, const char *document, float *out_score, size_t *out_token_count, struct skippy_error **out_error)` | Query/document relevance and token usage. |
| `enum skippy_status skippy_session_encode_prompt(struct skippy_session *session, const llama_token *token_ids, size_t token_count, llama_token *out_decoder_start_token, struct skippy_error **out_error)` | Encoder pass and decoder start token. |

Added types in that header: `enum skippy_workload_kind` (causal generation,
embedding, rerank, encoder-decoder), `enum skippy_workload_pooling` (unspecified,
none, mean, CLS, last, rank), and `struct skippy_workload_info_v1` with
`uint32_t abi_version`, `uint32_t struct_size`, `enum skippy_workload_kind kind`,
`enum skippy_workload_pooling pooling`, `uint32_t output_dimensions`,
`uint32_t classifier_outputs`, `bool has_encoder`, `bool has_decoder`,
`bool full_model_only`, and `uint8_t reserved0`. Descriptor version is 1;
the Rust mirror has explicit discriminant, size, and offset tests.

`SKIPPY_FEATURE_NON_CHAT_WORKLOADS` is added at bit 37 in
`include/skippy/common.h`, advertised by `src/skippy/abi.cpp`, and mirrored in
Rust. Removed native declarations: none. Older native ABIs are intentionally
not supported by this host; mesh-peer interoperability remains additive.

</details>

## Review focus and testing

The HTTP surfaces and mesh metadata are additive, but the change also touches
shared admission/routing, request-body handling, native loading, and sampling.
The main regression risks are selecting a backend that cannot serve a workload,
corrupting binary uploads, accepting a mismatched native library, and changing
ordinary chat behavior.

1. Build a matching host/runtime pair with `just release-build`. Use the
   [endpoint examples and certification instructions](https://github.com/Mesh-LLM/mesh-llm/blob/46c97287fa702928ed90ac797c5f7e6f9c491748/docs/NON_CHAT_MODELS.md)
   to exercise an explicit model and `model=auto` on the normal public port.
2. Test a mixed fleet: a chat-serving router, an embedding-serving peer, and a
   passive client. Check compatible remote selection and rejection of chat on
   an embedding model or embeddings on a chat model. Ordinary and streaming
   chat must still work.
3. Check binary multipart audio, including chunked `model=auto`, malformed and
   oversized uploads, and unsupported split placements. Embedding, rerank,
   encoder-decoder, and TTS execution must not silently become split chat.
4. For workload certification, run each pinned family's smoke **and** separate
   monolithic-oracle lane using source-bound CPU executables. A smoke-only run
   or a green ordinary chat canary is not six-workload certification. Oracle
   tools are test-only and are not alternative production serving backends.

## Validation

- [x] Relevant local checks were run; revision-specific evidence and limitations are below.
- [x] No UI behavior changes; screenshots are not needed.

Current published head: `46c97287fa702928ed90ac797c5f7e6f9c491748`, integrating
main at `0411e7d764a3a09ae606cde01d8fa6d81ad43e4b`.

Runtime/product checks below are bound to clean head
`c0d9bcf33859340ab5f7503905bceee0acc5e46c`. The final merge adds only four
upstream CI catalog/test/documentation files. Rust/native sources, build inputs,
models and certification scripts are unchanged; those runs are not relabeled
as executions on the later SHA. The repository gate was rerun on the final tree.

- Regression-first proof: the old ABI acceptance test and three workload probe
  tests failed before the correction and passed afterward.
- Strict affected all-target Clippy (`-D warnings`): PASS, no lint exceptions.
- Host runtime: 3,714 passed / 11 ignored. Static FFI: 15 passed; dynamic FFI:
  14 passed. Capability probe: 10 passed. Runtime: 181 passed / 1 ignored;
  server: 790 passed / 3 ignored; binary: 51 passed. These unaffected package
  suites ran before the final two-file upstream package-admission merge; the
  full host suite, check and strict Clippy were refreshed on the merged tree.
- Full native Skippy replay: 59/59 passed. Regenerated native tree differs from
  the preceding tested tree only in ABI version 60 → 61.
- Final-tree `just ci-validate`: PASS, 1,697 Python tests / 9 skips, actionlint,
  repository/release/publish consistency and console-output policy.
  `cargo fmt --all -- --check`, generated model manifests/roster/native API
  reference, and diff checks: PASS.
- Real dynamic-loader check rejects the prior 0.1.60 packaged library with an
  ABI mismatch, before mandatory-symbol resolution. The freshly packaged
  0.1.61 library loads and confirms all four workload exports.

- `just release-build`, composed release host plus packaged Metal runtime:
  PASS. The local rerun reused unchanged, previously validated UI assets.
  Real normal and streaming chat, automatic routing and headless smoke: PASS. Private mesh
  routing with two hosts and a passive client: 13/13 passed, covering compatible
  remote selection, incompatible-target rejection, and ordinary chat/MoA.

Source-bound CPU certification at `c0d9bcf3`: **6/6 families and all 12
smoke/oracle outcomes passed**. The producer manifest binds that clean commit
and actual executable digests. These are retained local
results, not an attached persistent-lab Actions run.

| Pinned representative | Observed result |
|---|---|
| Nomic embedding | Float/base64 HTTP and official Python SDK passed; normalized vectors matched the oracle. |
| Jina rerank | Original document indices, wire ranking order and score parity passed. |
| FLAN-T5 encoder-decoder | Generation matched the independent completion oracle. |
| PaddleOCR | Both paths exactly recognized the known fixture text `MESH 42`. |
| Qwen3-TTS | Waveform parity passed: cosine 1.0, relative RMS error 0.0. |
| Ultravox ASR | Candidate and independent execution produced matching text; the fixture is unlabeled. |

**Remote CI:** the current-head [GitHub checks](https://github.com/Mesh-LLM/mesh-llm/pull/1833/checks)
are still in progress at this description update. Local results above are not
presented as a completed final-head CI run or renewed reviewer approval.

### Validation limits

The ASR fixture demonstrates execution parity, not transcription accuracy.
Local CPU and same-machine mesh runs are not a persistent-lab GPU matrix or
mixed-version physical-node certification. Release attestation was not checked
because this local build has no configured release-signing key.
The local macOS release build used a build-dependency-only `strip=none`
override to work around a proc-macro dylib loading issue; the checked-in
product release profile was not changed.

An optional all-upstream native test build, outside the 59-test Skippy replay,
hit a pre-existing `test-llama-archs` private-include-path failure in unchanged
main patches. That optional suite is not reported as passing.

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI-compatible embeddings, reranking, encoder-decoder generation, OCR, speech synthesis, transcription, and translation support.
  * Added workload-aware model discovery and routing, selecting compatible models and returning clear errors for unsupported requests.
  * Added multipart audio uploads, binary audio responses, request validation, and configurable audio limits.
  * Added model workload metadata to API responses and improved support for full-model non-chat workloads.

* **Documentation**
  * Added comprehensive non-chat model endpoint, compatibility, and certification guidance.

* **Bug Fixes**
  * Improved request cleanup, sampling safety, server rebinding, and model artifact validation.


i386:
🔥 this is really good @IvGolovach - ill have an agent drop some messages here to make sure the nightly test is covered.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -231,29 +242,40 @@ async fn resolve_auto_routed_model(
...
         return AutoRouteResolution::Continue {
             effective_model: request.model_name.clone(),
             classification: None,
         };
     }
 
     request.ensure_body_json();
-    let Some(body_json) = request.body_json.as_ref() else {
+    let body_json = request.body_json.as_ref();
+    if body_json.is_none() && requested_workload.is_none() {
```
i386, line 259: [P1] Keep multipart bytes out of the JSON hook mutator

This new path intentionally lets audio multipart requests continue with `body_json = None`. After selection, `prepare_auto_route_decision` calls `maybe_enable_auto_route_hooks`, which unconditionally calls `inject_mesh_hooks_flag`. That helper scans the raw body for the first `{` byte and inserts a JSON fragment there. A valid binary audio file containing byte `0x7b` is therefore silently modified; for a chunked upload the insertion also invalidates the chunk framing. Please gate hook injection to JSON requests and add a regression whose file payload contains `{`.

crates/mesh-llm-host-runtime/src/network/openai/request_parse.rs
```diff
@@ -1038,15 +1067,54 @@ pub fn rewrite_model_field(request: &mut BufferedHttpRequest, model: &str) {
...
+            return;
+        };
+        rebuild_request_body(request, header_end, new_body, Some(body), model);
+        return;
+    }
+
+    let Some(content_type) = content_type_from_request(&request.raw) else {
+        return;
+    };
+    let original = &request.raw[header_end..];
```
i386: [P1] Rebuild chunked multipart from the decoded body

`original` comes from `request.raw`, so for `Transfer-Encoding: chunked` it still contains chunk-size lines and chunk terminators. `multipart_model_value_range` requires the body to start with `--boundary`; it therefore returns an error here and this function silently leaves the wire model as `auto`, which the selected backend rejects. The request parser already keeps the dechunked bytes in `request.body_bytes`. Please rewrite from those bytes, remove `Transfer-Encoding`, emit `Content-Length`, and cover chunked automatic audio routing.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -355,6 +386,7 @@ async fn auto_route_model_has_ready_ingress_target(
             node,
             model,
             required_tokens,
+            request_path,
```
i386, line 392: [P1] Let workload-incompatible local targets fall through to a capable peer

This returns as soon as any local target exists, even when the new workload filter makes every local target ineligible. For example, with a local causal/legacy instance and a remote embedding instance under the same model name, the remote descriptor satisfies global admission but this path drops the model from the automatic pool. Explicit routing has the same shape in `route_request`: local presence bypasses `route_missing_local_model`, then `route_model_request` filters the local target and returns 503. Please branch on workload-eligible local candidates and try remote hosts when that set is empty, or build one combined eligible target set before ranking and affinity.

crates/mesh-llm-host-runtime/src/network/openai/transport_route_model.rs
```diff
@@ -95,8 +95,14 @@ async fn route_model_request_inner(args: RouteModelRequestArgs<'_>) -> RouteDisp
     } = args;
     let route_started = Instant::now();
     let mut tcp_stream = tcp_stream;
-    let ranked =
-        rank_targets_by_context(&node, model, required_tokens, &targets.candidates(model)).await;
+    let candidates = super::super::workload_routing::eligible_targets(
+        &node,
+        model,
+        &request.client_path,
```
i386, line 101: [P2] Make affinity key extraction endpoint-aware after workload admission

`/v1/embeddings` defines OpenAI `user` as request metadata, but the generic `routing_keys` path later treats `user` as a session hash. That sets `affinity_applied = true`, so reservation spreading is disabled and every concurrent embedding request for one user is pinned to one replica. Embedding and rerank execution clears workload state and drops the session, so there is no KV reuse to offset that hotspot. Please retain the current keys for generation routes and pass no session/prefix affinity for stateless workloads unless the runtime advertises a reusable artifact.

crates/mesh-llm-host-runtime/src/network/openai/workload_routing.rs
```diff
@@ -0,0 +1,219 @@
...
+mod tests;
+
+pub(super) fn is_audio_upload_path(path: &str) -> bool {
+    matches!(
+        path.split('?').next().unwrap_or(path),
+        "/v1/audio/transcriptions" | "/v1/audio/translations"
+    )
+}
+
+pub(super) fn request_workload_class(path: &str) -> Option<ModelWorkloadClass> {
```
i386, line 24: [P1] Exclude non-chat workloads from MoA worker discovery

`assemble_worker_pool` still builds `all_models` from `models_being_served()` and never checks `ServedModelMetadata.workload_class`. Every entry is then called via `/v1/chat/completions` by both local and remote MoA backends. After this PR, embedding, rerank, and TTS models are legitimate members of that served set, so a normal `model=mesh` chat can admit one as a worker/reducer (or even select it as the only best member), leading to 4xx worker failures or a failed turn. The endpoint-level filter here only protects ordinary routing because MoA intercepts before it. Please filter MoA pool aliases/targets to `CausalGeneration` and decide explicitly whether `EncoderDecoder` supports MoA/tool roles, with a mixed-workload fleet test.

i386: @michaelneale needs your brains

ci/llama-canary/family-certified.json
```diff
@@ -711,13 +797,83 @@
...
       "mmproj_artifact": {"repo": "Blackfrost-AI/Muse-Glimmer-30B-Abliterated-GGUF", "revision": "83e7dc723f5515704adc8767dfeb279de3df914d", "files": ["mmproj-Muse-Glimmer-30B-Abliterated-F16.gguf"], "file_integrity": {"mmproj-Muse-Glimmer-30B-Abliterated-F16.gguf": {"size_bytes": 3847367616, "blob_id": "c2e08b8bd13ca60b59ffdcb79b969b354b1edda9b9198047d10621adc11591bb"}}, "selector": "F16"},
       "execution": {"trunk_layers": 5, "mtp_layers": 0, "activation_width": 6656, "boundary_sweep_period": 0, "speculative_policy": "mtp-if-present"},
       "resources": {"runner_role": "family-certify", "cache_policy": "immutable-local", "estimated_model_bytes": 1631210720},
       "notes": "Muse Glimmer dflash text and projector lanes"
+    },
+    {
+      "family": "nomic-bert-embedding",
+      "class": "embedding",
+      "profile": "workload-oracle",
+      "cadences": ["manual-full"],
```
i386: [P1] Make the new workload certifications executable from the llama canary

All six new `workload-oracle` rows are `manual-full` only; generating the plans at this head selects 6 non-chat rows for `manual-full` and 0 for `llama-bump`, so a llama.cpp pin advance never exercises them. There is also no repository wiring for `SKIPPY_WORKLOAD_ORACLE_SERVER`, `SKIPPY_WORKLOAD_ORACLE_COMPLETION`, or `SKIPPY_WORKLOAD_ORACLE_TTS`: the canary workflow neither builds nor exports those executables, while `run_workload_certify` exits before the lane when a certified row lacks its class-appropriate variable. That makes forced certification depend on undeclared ambient runner state and leaves the pin-advance canary without this coverage. Please build and export the pinned CPU oracle tools deterministically in the workflow, add these rows to `llama-bump` if they are intended to guard llama updates, and provide a head-exact full-canary result for the six rows.

i386 (changes requested):
Requesting changes at `25ff652dce88390695412c503c43cf01221aadd1` for the five P1 items below:

1. Gate mesh JSON hook injection by body/content type so multipart audio bytes are never scanned or mutated. Cover a binary payload containing `{`.
2. Rewrite chunked multipart `model=auto` from decoded body bytes, remove `Transfer-Encoding`, set the new `Content-Length`, and add an end-to-end chunked audio routing regression.
3. Apply workload eligibility before the local-presence short circuit so an incompatible local replica cannot hide a compatible remote peer. Cover both automatic and explicit routes.
4. Filter MoA worker/reducer admission by workload and role before context ranking, reservations, affinity, and failover. `CausalGeneration` is the safe default; make any `EncoderDecoder` participation explicit and cover a mixed-workload fleet.
5. Make all six workload certification rows run in the llama bump canary, deterministically provision/export the required oracle executables, and attach a head-exact six-class family-battery result against the verified lab cache.

The lab cache population is being handled separately and is not required PR logic. The existing P2 affinity comment also remains a requested efficiency fix: endpoint metadata such as the embeddings `user` field should not disable replica spreading unless the workload exposes reusable state.

Inline details:

- https://github.com/Mesh-LLM/mesh-llm/pull/1833#discussion_r3997788564
- https://github.com/Mesh-LLM/mesh-llm/pull/1833#discussion_r3997788868
- https://github.com/Mesh-LLM/mesh-llm/pull/1833#discussion_r3997789222
- https://github.com/Mesh-LLM/mesh-llm/pull/1833#discussion_r3997808174
- https://github.com/Mesh-LLM/mesh-llm/pull/1833#discussion_r3997823138
- https://github.com/Mesh-LLM/mesh-llm/pull/1833#discussion_r3997789648

i386:
Strict clippy on Rust 1.98 found one remaining PR-introduced lint in the speech PCM conversion. The tested fix is commit 0ab3cde9b324aa3a0b7bb7673aa90d17e77c0a60 on Mesh-LLM/mesh-llm branch scama/pr1833-rust198-clippy; please cherry-pick it because this fork PR has maintainer edits disabled. Validation on that commit: cargo fmt, strict host-runtime clippy, skippy-runtime 128 passed/1 ignored. The parent b8584caaef8ff398df60d2feadee54508c635c30 also passes host-runtime 3,109 passed/11 ignored and 97 focused workload/canary Python tests.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
i386: Verified at b8584caaef8ff398df60d2feadee54508c635c30: hook injection now parses and mutates only JSON objects; multipart bytes remain intact. The new binary-{ regression passes in the full host-runtime suite.

crates/mesh-llm-host-runtime/src/network/openai/request_parse.rs
i386: Verified at b8584caaef8ff398df60d2feadee54508c635c30: model rewrite consumes decoded body_bytes for chunked requests and rebuilds framing with Content-Length while removing transfer framing. The chunked audio end-to-end regression passes.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
i386: Verified at b8584caaef8ff398df60d2feadee54508c635c30: ingress_candidates admits eligible elected targets first, then falls through to eligible remote hosts when the local set is empty. The local-causal/remote-embedding readiness regression passes.

crates/mesh-llm-host-runtime/src/network/openai/transport_route_model.rs
i386: Verified at b8584caaef8ff398df60d2feadee54508c635c30: affinity_body now supplies session/prefix state only for generation endpoints. Embedding/rerank requests remain stateless and spreadable.

crates/mesh-llm-host-runtime/src/network/openai/workload_routing.rs
i386: Verified at b8584caaef8ff398df60d2feadee54508c635c30: MoA admission now accepts only causal-generation or legacy descriptors across discovery, routing hints, local/remote targets, self-fill, actor ranking, and degrade selection. Mixed-workload tests pass in the full host-runtime suite.

i386:
I pushed the complete review-fix series to `Mesh-LLM/mesh-llm:scama/pr1833-rust198-clippy`.

The PR branch is on a fork with maintainer edits disabled, so the author needs to apply both support commits:

```bash
git cherry-pick 0ab3cde9b324aa3a0b7bb7673aa90d17e77c0a60 968cca2458d5d4b746c93baf7baf7902a98a6651
```

`968cca2458d5d4b746c93baf7baf7902a98a6651` addresses all 17 new CodeRabbit findings: request-object cleanup, fail-closed workload decoding, audio bounds, CMake cache parsing/backend assertions, generation caps, dry-run/oracle behavior, exact OCR matching, configurable startup timeout, mandatory SDK/projector evidence, exact artifact suffixes, exhausted-vocabulary sentinels, ABI layout checks, embeddings-mode restoration, and response cardinality.

Validation on that exact Rust head:

- strict Clippy with `-D warnings` across all five affected packages and all targets
- 64 script contract tests
- OpenAI frontend: 246 unit/integration tests passed
- host runtime: 3,108 passed; 11 ignored, plus 2 integration tests passed
- skippy FFI: 12 passed
- skippy runtime: 128 passed; 1 ignored
- skippy server: 707 passed; 3 ignored
- clean native patch replay, CPU static build, and all 33 native CTest tests passed

The original i386 full-canary review thread should remain open until the persistent family-certification runner produces the requested six-model evidence. PR CI cannot exercise these fixes until the author cherry-picks the support commits.

i386:
I completed the current-main review and published the reviewed integration head here:

- branch: `scama/pr1833-final-review-20260915`
- exact head: `a0ddd1a4dfaba3d7e419ef0a36e5c2f58c9f112e`

That head resolves the current merge conflicts, rejects/initializes native sampled-token sentinels correctly, keeps the generated split roster current, and commits the multipart/base64 lockfile additions. `just llama-prepare` produces patched llama tree `ee18c1f77e7ba2d8f0a4123e0a7e759c9e3c82bb`; the repository gate passes with 1,591 tests and 9 skips.

The remaining merge blocker is still the executable six-workload canary proof requested in the human review. The author branch cannot be updated by maintainers, so please merge or reset to the support branch and run the head-exact full canary. I am leaving that thread, and the sentinel thread whose fix exists only on the support branch, unresolved until the PR head contains the changes.

third_party/llama.cpp/patches/0040-skippy-honor-vocab-suppress-tokens-in-sampling.patch
```diff
@@ -0,0 +1,348 @@
...
++        suppressed.resize(static_cast<size_t>(n_vocab), 0);
++        for (int32_t i = 0; i < suppress_token_count; ++i) {
++            const llama_token token = suppress_tokens[i];
++            if (token >= 0 && token < n_vocab) {
++                suppressed[static_cast<size_t>(token)] = 1;
++            }
++        }
++    }
++
++    llama_token best = LLAMA_TOKEN_NULL;
```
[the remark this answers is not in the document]
i386: Verified on the author head: patch 0042 routes sampled outputs, verify results, session sampling, and MTP proposal/sync through skippy_store_sampled_token, which rejects every negative sentinel before publication or reuse.

crates/mesh-llm-host-runtime/src/network/openai/request_parse_tests.rs
```diff
@@ -789,6 +828,157 @@ fn test_rewrite_model_field_updates_body_and_content_length() {
...
+        "x".repeat(257)
+    );
+    assert!(
+        multipart_model_field(
+            &format!("multipart/form-data; boundary={boundary}"),
+            body.as_bytes()
+        )
+        .unwrap()
+        .is_none()
+    );
```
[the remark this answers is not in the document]
i386: Verified on the author head: multipart_model_field now returns an error above the 256-byte bound while preserving None only for an absent or empty field.

i386 (approved):
Re-reviewing my own Sept 12 request-changes at the current head `90897a62` — **all six findings are fixed and verified in the tree**; the review threads just never got replies. Details per finding:

1. **Multipart bytes vs JSON hook mutator [P1]** — fixed properly. Body transformations moved to `request_parse/body_rewrite.rs`, which operates on decoded payloads only; `BodyHeaders::permits_json` restricts JSON rewriting to absent/`application/json` content types, and chunked framing is decoded (`try_decode_chunked_body`) before any mutation and rebuilt with a single Content-Length afterward.
2. **Chunked multipart model rewriting [P1]** — fixed by the same decode/rebuild path: `rewrite_multipart_model` decodes chunked bodies before locating the model field range and re-frames on the way out.
3. **Local-only incompatibility shadowing capable peers [P1]** — fixed in `workload_routing::ingress_candidates`, with the exact invariant documented: local ineligible targets no longer short-circuit; routing falls through to `hosts_for_model` remotes filtered by `eligible_targets`.
4. **`user`-field affinity on stateless endpoints [P2]** — fixed with `supports_generation_affinity` (chat/completions/responses only); `auto_session_key_for_request` now returns None for embeddings/rerank/TTS/transcription paths, so stateless traffic no longer pins sessions.
5. **MoA pool contamination [P1]** — fixed via a dedicated `moa_gateway/workload_admission` module; `assemble_worker_pool` filters local and peer candidates through `model_supports_committee` / `descriptor_supports_committee` before pool assembly, with a fleet-sim test module covering it.
6. **Canary oracle executability [P2]** — fixed: `scripts/skippy-workload-oracles-build.sh` builds and exports the three oracle variables, and `llama-upstream-canary.yml` invokes it and appends `--print-env` to `GITHUB_ENV`.

CI on this head: 85 success / 15 skipped / 0 failures.

The one thing left before merge is mechanical: the branch currently **conflicts with main** (mergeStateStatus DIRTY). Happy to re-verify the rebased head quickly once it's pushed — the delta should be conflict resolution only.

michaelneale:
This is a very large change. Any tips on how to best test or risks? Mostly additive - js this for multi modal models?

i386:
James asked me to get this over the line, but the head branch lives in `IvGolovach/mesh-llm` with `maintainerCanModify: false`, so neither the author-update path nor `gh pr update-branch` is available to base-repo maintainers. I've opened **#1921** as a replacement: the identical diff from `90897a62` squashed onto current `main` (one textual conflict in `ci/ci.md`, plus two small compile fixes the auto-merge needed), fully verified locally (1616 contract tests, 3627 host-runtime lib tests, fmt/actionlint/clippy-adjacent checks clean). My approve stands on the content; #1921 needs one green CI run before merge. @IvGolovach — nothing is lost, all review history stays here; feel free to cross-check the merge resolutions in #1921.

i386:
Need to check if we need to add canary coverage for this before merging - have agent on it.

i386:
Canary assessment at `90897a627de92c7321e9c12e7d006eea56c7a2c1`: this change needs workload-specific runtime coverage because ordinary chat/split canaries do not validate the new execution paths.

The PR already provides six pinned smoke + independent-monolithic-oracle lanes: Nomic embeddings, Jina rerank, FLAN-T5 encoder-decoder, PaddleOCR, Qwen3-TTS, and Ultravox ASR. The family planner selects all six, and the workflow builds and exports the required oracle executables. The earlier manual-only selection and oracle-wiring gaps are resolved in this head.

Before treating these workloads as runtime-certified, retain a passing six-class run tied to the final integration SHA. I also recommend a small live remote-routing smoke covering compatible-peer selection and incompatible-target rejection, alongside the unsupported-split rejection tests.

This is source-level confirmation of the coverage wiring, not a claim that I have verified a passing final-integration hardware run.

i386 (dismissed):
Approved at `90897a627de92c7321e9c12e7d006eea56c7a2c1`. The workload canary definitions and oracle build wiring address the earlier coverage gaps. See my canary assessment comment for the remaining runtime-evidence recommendation; this approval does not attest to a passing six-class hardware certification run.

third_party/llama.cpp/patches/0010-skippy-define-non-chat-workload-ABI-and-execution.patch
```diff
@@ -0,0 +1,641 @@
...
+diff --git a/include/skippy/common.h b/include/skippy/common.h
+index 5ac737396..133b1d5a2 100644
+--- a/include/skippy/common.h
++++ b/include/skippy/common.h
+@@ -46,7 +46,7 @@ extern "C" {
+ 
+ #define SKIPPY_ABI_VERSION_MAJOR 0
+ #define SKIPPY_ABI_VERSION_MINOR 1
+-#define SKIPPY_ABI_VERSION_PATCH 59
++#define SKIPPY_ABI_VERSION_PATCH 60
```
ndizazzo: [P1] Advance the ABI for these mandatory symbols. This repeats the existing 0.1.59 to 0.1.60 bump from patch 0007, even though the merge-base Rust facade is already 0.1.60. The same patch adds `skippy_model_workload_info_v1`, `skippy_session_embed`, `skippy_session_rerank`, and `skippy_session_encode_prompt` to the eager `dynamic_symbols!` table. `check_runtime_abi` therefore accepts an existing 0.1.60 runtime and then fails while resolving the first new export, so an older runtime is advertised as compatible but cannot start. Please advance both the C and Rust ABI to 0.1.61, and update the generated/API/package references with it.

crates/skippy-ffi/src/abi.rs
```diff
@@ -19,7 +19,65 @@ pub const FEATURE_KV_EVENTS: u64 = 1 << 33;
 pub const FEATURE_DEVICE_EVENTS: u64 = 1 << 34;
 pub const FEATURE_DIAGNOSTIC_EVENTS: u64 = 1 << 35;
 pub const FEATURE_UNLOAD_EVENTS: u64 = 1 << 36;
+/// Full-model workloads use a separate bit from all runtime-event families.
+pub const FEATURE_NON_CHAT_WORKLOADS: u64 = 1 << 37;
```
ndizazzo, line 23: [P2] Register bit 37 with the capability probe. `crates/skippy-runtime/src/capability_probe.rs` still sets `MAX_KNOWN_FEATURE_BIT` to 36 and has no family entry for `FEATURE_NON_CHAT_WORKLOADS`, so a runtime advertising this bit is reported as reserved and the probe never confirms the workload family or checks its required symbols. Add the family and cover both complete and missing-symbol cases so capability-aware callers can fail closed consistently with the other families.

ndizazzo (changes requested):
Thanks for carrying the workload paths through the native, runtime, and routing layers. I found one compatibility blocker in the native ABI, plus a capability-probe gap that needs to land with it. The current PR checks are still running, so I’m requesting changes on this exact head. The description also still reports ABI 0.1.54 and head 25ff652d, so please refresh those claims after the fix.

ndizazzo (approved):
Rechecked the final head. The ABI and capability-probe fixes are in place, the canary workload-oracle closure is wired, and the required PR lanes are green.

PR #1835, opened 2026-09-12
fix(ci): recover release-note entries credited only to a roll-up

ndizazzo:
## Original problem

The [v0.76.1 notes](https://github.com/Mesh-LLM/mesh-llm/releases/tag/v0.76.1) shipped a `### Fixed` section with one entry in it:

> * Roll up the 0.76.1 bugfix branch into main by @ndizazzo in #1789

Fifteen fixes were behind that line. They merged into `bugfix/0.76.1`, which then reached `main` by rebase, so GitHub credited the roll-up and nothing else. Seven more entries sat in `### Other changes` — `test(ci): guard product smoke selector drift`, `ci: publish SDK smoke cache policy` and friends — even though every one of them has a conventional type.

## Diagnostics

Both are the same break: **an entry that cannot reach its commit.**

The classifier reads a section from the commit's Conventional Commits type, and it reaches the commit through the `(#N)` suffix a squash merge leaves on the subject. Two routes drop that thread:

- the seven `Other changes` entries arrived on `main` without the suffix, so `read_commits` skipped them entirely and `build_plan` had nothing to classify;
- the fifteen fixes *do* carry `(#1761)`, `(#1766)` and so on, and are sitting right there in `v0.76.0..v0.76.1` — but they are not in the published body, so nothing ever looked at them.

`repos/{repo}/commits/{sha}/pulls` resolves all seven of the first group, and the suffixes already resolve all fifteen of the second.

## How it works

A link pass runs before classification:

```mermaid
flowchart TD
    A["Release published"] --> B["Capture GitHub's flat list"]
    B --> L["Link commits to pull requests"]
    L --> C["Classify by commit type"]
    C --> D["Agent review (optional)"]
    D --> P["Publish"]
```

`scripts/release-notes-link.py` pairs each commit in the range with the pull request that carried it. The `(#N)` suffix stays authoritative and costs nothing; only a commit without one costs an API lookup, and a commit the API cannot place is left unlinked rather than guessed at. From that pairing it writes two things:

- an **augmented body** that credits pull requests GitHub did not, each one placed beneath the roll-up that carried it in, in the same `* <title> by @<author> in <url>` shape GitHub uses;
- a **links file** of the commit records the classifier could not key from the subject alone, which `release-notes-classify.py` now merges with its own git scan.

## Gates

The entry-set rule now runs both ways.

- The link pass may **only add**. `release-notes-generate.sh` diffs the published PR set against the augmented one and refuses to publish if anything was dropped.
- Everything after it keeps the existing rule: the set must not change at all.
- `body.github.md` holds the release exactly as GitHub published it, and is what the restore hint points at.

Every API call is best-effort and bounded by `--api-budget` (200 by default). A missing `gh`, an unreachable API, a rate limit, or a pull request with no resolvable author each log a reason and fall back to precisely the notes this pipeline produces today. None of them fail the release.

The job gains `pull-requests: read`.

## Result on v0.76.1

```
linked 11 commit(s) to a pull request through the API; recovered 15 entry(ies) GitHub did not credit (19 published)
classified 34/34 entries deterministically (0 in 'Other changes')
```

`### Fixed` goes from 1 entry to 14, `### Added` picks up #1754 and #1803, `### Other changes` disappears, and the fifteen recovered fixes credit @i386, @ndizazzo and @Virgile-pct by name instead of vanishing behind the roll-up. #1789 stays: it is the only credit for the four multipart-GGUF fixes @scama landed on the branch with no pull request of their own.

## Testing

- 15 new unit tests covering suffix-over-API precedence, roll-up attribution, an orphan with no carrier after it, a pull request with no resolvable author, double-crediting, splicing, the never-drop invariant, the links handoff, and the API budget.
- `python3 -m unittest scripts.tests.test_release_notes` — 89 pass.
- Full `scripts/tests` suite on 3.12: failing set is byte-identical to `origin/main` (45 pre-existing local-environment failures, none in release notes).
- End-to-end `DRY_RUN=true` against the live v0.76.1 release, output above.

## Summary by CodeRabbit

* **Release Notes**
  * Release notes now better associate commits with their pull requests, including changes grouped under roll-up pull requests.
  * Additional commit metadata can be used to classify entries accurately, including security-related changes.
  * Published release content is preserved when automated linking cannot be completed.

* **CI Reliability**
  * Runtime-event checks now consistently use the same evidence file across working directories.
  * Model validation covers both pull-request and main-branch runs.

* **Documentation**
  * CI inventory and release-note procedures now reflect the updated workflows and recovery steps.

### DIFF
--- a/.agents/skills/release-notes/SKILL.md
+++ b/.agents/skills/release-notes/SKILL.md
@@ -33,6 +33,11 @@ readiness or build the change inventory from source; that is
   invariant every gate checks.
 - **Never drop an entry.** Every merged PR credits a contributor, including the
   CI and build churn. Noisy entries collapse into `### Internal`, never deleted.
+- **Recover an entry GitHub could not credit.** A batch of pull requests merged
+  into a staging branch and rebased into the release is credited once, to the
+  roll-up. The link pass adds the missing entries back in the roll-up's own
+  format, so the reader sees the fixes rather than the vehicle that carried
+  them. Adding is the only direction: the published set is a floor.
 - **Keep the tail.** `## New Contributors` and the `**Full Changelog**` link
   stay exactly as GitHub generated them.
 - **Never guess a section.** An entry the commit metadata cannot place goes to
@@ -43,13 +48,14 @@ readiness or build the change inventory from source; that is
 
 ## How The Pipeline Works
 
-Two passes, in this order. The first always runs and always produces a
-publishable body; the second is best-effort.
+Three passes, in this order. The first two always run and together always
+produce a publishable body; the third is best-effort.
 
 ```mermaid
 flowchart TD
     A["Release published"] --> B["Capture GitHub's flat list"]
-    B --> C["Classify by commit type<br/>feat → Added · fix → Fixed · ci → Internal"]
+    B --> L["Link commits to pull requests<br/>recover what GitHub credited to a roll-up"]
+    L --> C["Classify by commit type<br/>feat → Added · fix → Fixed · ci → Internal"]
     C --> D{"Agent reachable?"}
     D -- no --> F["Deterministic notes"]
     D -- yes --> E["Agent reviews the plan"]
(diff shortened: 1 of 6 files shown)


Virgile-pct (commented):
Reviewed by running it. I reproduced the end-to-end result on Windows 11 against the live v0.76.1 release, and both lines come out identical to yours:

```
linked 11 commit(s) to a pull request through the API; recovered 15 entry(ies) GitHub did not credit (19 published)
classified 34/34 entries deterministically (0 in 'Other changes')
```

The never-drop gate holds on that run: `comm -23` between the published and the augmented entry sets is empty, and the fifteen added are exactly the pull requests sitting in `v0.76.0..v0.76.1` without a credit.

One thing that looked wrong and is not: the recovered entries carry their raw conventional prefix into the augmented body, so `body.linked.md` reads `* fix(openai): reject malformed tool definitions by @i386`. Rendering it settles the question, since `release-notes-regroup.py` strips the type and publishes `Reject malformed tool definitions`. Only worth knowing for anyone reading the intermediate artifact.

## One finding

`commits.update(load_links(...))` inverts the precedence the rest of the pass states. `resolve_pull_requests` treats the `(#N)` suffix as authoritative and free, but at `release-notes-classify.py:277` the API-derived record overwrites the git-derived one for the same pull request.

#1789 is a live collision in this very release. `links.json` keys it to `fix(runtime): serve multipart GGUFs from HF cache`, a commit it carried, while its authoritative squash subject is `fix: address review feedback on the 0.76.1 bugfix merge (#1789)`.

Both are `fix`, so v0.76.1 is unaffected. The mechanism is not: editing only the links record to `feat(runtime): ...` and re-running the classifier moves #1789 from `Fixed` to `Added`.

The second consequence is the one I would worry about. The API record carries `"trailers": {}`, so for any colliding pull request the `Release-Notes:` override and `BREAKING CHANGE:` are dropped silently, and those are precisely the cases where someone set the section on purpose.

Reads like a one-liner:

```python
for pr, record in load_links(args.links).items():
    commits.setdefault(pr, record)
```

## One question, not reachable here

`resolve_pull_requests` takes `numbers[0]` when the API returns several pull requests for a commit, while the comment beside it says a commit the API cannot place is left unlinked rather than guessed at. Every commit in `v0.76.0..v0.76.1` resolves to zero or one, so nothing in this release exercises it.

## Not verified

The unit tests. `scripts/tests/test_release_notes.py` imports `yaml`, which this box does not have, and that import predates the branch, so I read the fifteen new cases rather than running them. I have also run nothing on Linux or macOS.

i386 (dismissed):

PR #1836, opened 2026-09-12
fix(system): read Windows adapter VRAM from a 64-bit source

### LINKED ISSUE #1811: Windows: mesh-llm gpus reports 4 GB on RX 9070 XT (16 GB); Win32 AdapterRAM uint32 cap
## Summary

On Windows + AMD RX 9070 XT (16 GB), `mesh-llm gpus` still reports **~4 GB**. That number is the Win32 `AdapterRAM` **uint32 ceiling**, not the card.

This did **not** block `mesh-llm serve`: Qwen3-14B Q4_K_M loaded on ROCm and used **10.992 GB** process VRAM of **15.92 GB** real card capacity. Anything that **refuses a model** from `gpus` / advertised `vram_bytes` will wrongly reject 14B+ on this SKU.

v0.76.0 stable. Not #1160 (libstdc++). Not #1753 (flavor override ignored).

## Evidence (this machine)

| Source | Value |
|---|---|
| Real card | AMD Radeon RX 9070 XT **15.92 GB** |
| `mesh-llm gpus` | **4 GB** |
| `serve` Qwen3-14B Q4_K_M (ROCm) | **10.992 GB** used / api ready / 42.8 tok/s |
| mesh-llm | **0.76.0** stable, official `install.ps1` |
| OS | Windows 11 |

## Code (v0.76.0)

Windows fallback still reads CIM `AdapterRAM`:

```text
crates/mesh-llm-system/src/hardware/mod.rs
Get-CimInstance Win32_VideoController | Select-Object Name,AdapterRAM | ConvertTo-Json -Compress
```

`parse_windows_video_controller_json` takes that integer as `vram_bytes`. `mesh-llm gpus` prints `VramCapacity` from `gpu.vram_bytes`.

`Win32_VideoController.AdapterRAM` is a 32-bit field. Typical saturated value on this SKU (already on record in #1163):

```json
{"Name":"AMD Radeon RX 9070 XT","AdapterRAM":4293918720}
```

`4293918720 / 1024 / 1024 = 4095 MiB` → CLI **4 GB**.

#1163 fixed **index misalignment** (virtual adapter stealing the only non-zero AdapterRAM). It did **not** replace AdapterRAM with DXGI / ROCm / Vulkan / skippy `memory_total`. After that fix, the 9070 XT is the row that gets **4 GB**.

NVIDIA Windows path can use `nvidia-smi` memory.total. AMD Windows still has this CIM fallback when the embedded runtime does not overwrite the survey.

## Expected

`gpus` / advertised capacity should match dedicated VRAM (~16 GB), using DXGI, ROCm, Vulkan, or skippy device `memory_total` — not CIM AdapterRAM.

## Actual

CLI **4 GB**. Serve can still allocate ~11 GB. Fit-gating from the survey will refuse models this card already runs.

## Related (not duplicates)

- #1163 (closed) — same SKU, same `4293918720`, different bug (list compaction)
- #1656 — opaque `vram_bytes` breakdown
- #1678 (closed) — `gpu_name` provenance
- #1753 — `--llama-flavor` ignored (separate)

## Workaround

Trust `serve` / process VRAM, not `mesh-llm gpus`, until the probe uses a >32-bit Windows API. Do not set `max_vram_gb: 4`.

Virgile-pct:
Fixes #1811.

## What is broken

`Win32_VideoController.AdapterRAM` is a 32-bit field, so every adapter holding more than 4 GB reports the same saturated `4293918720` bytes. @rileyorscheln reported it on an RX 9070 XT; it reproduces on a completely different card, so it is not vendor specific:

| Source | RTX 4070 Ti | AMD integrated | AMD 780M (second box) |
|---|---|---|---|
| `Win32_VideoController.AdapterRAM` | 4293918720 (4095 MiB) | 536870912 | 536870912 |
| `HardwareInformation.qwMemorySize` | 12878610432 (12282 MiB) | 536870912 | 536870912 |
| `nvidia-smi memory.total` | 12282 MiB | n/a | n/a |

The 64-bit registry value equals `nvidia-smi memory.total` to the byte, and the two integrated adapters agree across both sources, so reading the wider field does not disturb small adapters.

It stays invisible on NVIDIA only because `nvidia-smi` usually overwrites the survey first. The CIM value is what an AMD box falls back to, and anything gating a model on the advertised `vram_bytes` then refuses models the card runs.

## What this changes

`read_windows_video_controllers` keeps the CIM query, adds the display-class registry as a second source, and merges them.

**Pairing is by name, never by position.** On the test machine the registry enumerated the AMD adapter first and CIM enumerated the NVIDIA one first. `windows_per_gpu_vram` deliberately keeps zero entries so it stays index aligned with the adapter name list, which is what #1163 fixed, and pairing the registry rows by index would put one card's memory on another. Identical cards share a name, so equal names are consumed in order, and an adapter with no usable entry keeps whatever CIM reported.

## One subtlety worth flagging

The probe ends with an explicit `exit 0`, and it does not work without it.

PowerShell propagates the last statement's `$?` as its exit code. Reading a class subkey that does not carry the property leaves that false even under `-ErrorAction SilentlyContinue`, which suppresses the message but not the effect, and `powershell_output` discards the body on a non-zero exit. Measured from Rust on Windows: the script printed a perfectly good JSON body, returned 1, and the probe came back empty. The unit tests were green throughout, which is exactly why I ran the real probe before proposing anything.

`exit 0` is safe here because the failure mode it hides is "no entry for this adapter", which the merge already handles by keeping the CIM value.

## Verified on Windows 11

- `cargo test -p mesh-llm-system`: 206 passed, 0 failed, including 5 new cases.
- `cargo clippy -p mesh-llm-system --all-targets`: no lints.
- `scripts/check-env-mutation-contract.py` and `python -m unittest scripts.tests.test_env_mutation_contract`: passed.
- The exact query string, extracted from the Rust source and run on the 12 GB box, returns `[{"Name":"AMD Radeon(TM) Graphics","Bytes":536870912},{"Name":"NVIDIA GeForce RTX 4070 Ti","Bytes":12878610432}]`, which is the ordering the merge test covers.

New tests: the saturated value replaced by name across mismatched orders, adapters the registry does not cover, two identical cards paired in order, zero-sized entries ignored, and the parser on both JSON shapes.

## Not verified

I do not have an adapter above 4 GB that falls back to CIM, so the end-to-end correction is established by composition rather than by one run: CIM saturation measured, registry value measured, merge unit tested on exactly those values and that ordering. Someone with a 9070 XT can confirm in one command.

Whether `qwMemorySize` is present on every driver is unknown; an adapter without it keeps today's behaviour. Nothing here runs outside Windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows GPU memory detection for graphics cards with more than 4 GB of memory.
  * Combined hardware information from multiple sources more accurately, including systems with duplicate adapter names.
  * Preserved valid memory values when registry data is missing, zero, or incomplete.

### DIFF
--- a/crates/mesh-llm-system/src/hardware/mod.rs
+++ b/crates/mesh-llm-system/src/hardware/mod.rs
@@ -332,6 +332,52 @@ fn windows_per_gpu_vram(controllers: &[(String, u64)]) -> Vec<u64> {
     controllers.iter().map(|(_, ram)| *ram).collect()
 }
 
+/// Pair each adapter with its 64-bit memory size, matched by name.
+///
+/// The registry enumerates adapters in its own order, which is not the CIM
+/// order, so pairing by position would move one card's memory onto another.
+/// That is the misalignment #1163 fixed for `AdapterRAM` and it must not come
+/// back here. Identical cards share a name, so equal names are consumed in
+/// order. An adapter with no usable entry keeps whatever CIM reported.
+#[cfg(any(target_os = "windows", test))]
+fn merge_adapter_memory(
+    controllers: &[(String, u64)],
+    adapter_memory: &[(String, u64)],
+) -> Vec<(String, u64)> {
+    let mut taken = vec![false; adapter_memory.len()];
+    let mut merged = Vec::with_capacity(controllers.len());
+    for (name, adapter_ram) in controllers {
+        let mut bytes = *adapter_ram;
+        for (index, (candidate, candidate_bytes)) in adapter_memory.iter().enumerate() {
+            if taken[index] || *candidate_bytes == 0 || candidate != name {
+                continue;
+            }
+            taken[index] = true;
+            bytes = *candidate_bytes;
+            break;
+        }
+        merged.push((name.clone(), bytes));
+    }
+    merged
+}
+
+/// Per-adapter `HardwareInformation.qwMemorySize` from the display class key.
+///
+/// The script ends with an explicit `exit 0`. PowerShell propagates the last
+/// statement s $? as its exit code, and reading a class subkey that does not
+/// carry the property leaves it false even under `-ErrorAction SilentlyContinue`,
+/// which would make `powershell_output` discard a perfectly good JSON body. An
+/// adapter without an entry is an ordinary outcome here, not a failure.
+#[cfg(target_os = "windows")]
+fn read_windows_adapter_memory() -> Vec<(String, u64)> {
+    let Some(output) = powershell_output(
+        r"Get-ChildItem 'HKLM:\SYSTEM\CurrentControlSet\Control\Class\{4d36e968-e325-11ce-bfc1-08002be10318}' -ErrorAction SilentlyContinue | ForEach-Object { $p = Get-ItemProperty $_.PSPath -ErrorAction SilentlyContinue; $q = $p.'HardwareInformation.qwMemorySize'; if ($q) { [pscustomobject]@{ Name = $p.DriverDesc; Bytes = [uint64]$q } } } | ConvertTo-Json -Compress; exit 0",
+    ) else {
+        return Vec::new();
+    };
+    parse_windows_adapter_memory_json(&output)
+}
+
 #[cfg(target_os = "windows")]
 fn read_windows_video_controllers() -> Vec<(String, u64)> {
     let Some(output) = powershell_output(
@@ -339,7 +385,8 @@ fn read_windows_video_controllers() -> Vec<(String, u64)> {
     ) else {
         return Vec::new();
     };
-    parse_windows_video_controller_json(&output)
+    let controllers = parse_windows_video_controller_json(&output);
+    merge_adapter_memory(&controllers, &read_windows_adapter_memory())
 }
 
 /// Owns an Objective-C object retained via the "create rule" (e.g.
--- a/crates/mesh-llm-system/src/hardware/parsers.rs
+++ b/crates/mesh-llm-system/src/hardware/parsers.rs
@@ -402,6 +402,41 @@ pub fn parse_windows_video_controller_json(output: &str) -> Vec<(String, u64)> {
     }
 }
 
+/// Parse the display-class registry dump -> `(driver_desc, memory_size_bytes)`.
+///
+/// `Win32_VideoController.AdapterRAM` is a 32-bit field, so any adapter with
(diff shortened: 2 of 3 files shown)


i386 (approved):
Approved. Reviewed the pairing logic and the probe end to end.

- `merge_adapter_memory` pairs by name with first-unmatched consumption: the \#1163 index-alignment property of `windows_per_gpu_vram` is preserved, adapters the registry does not cover keep the CIM value, and zero-size entries are skipped. The mismatched-order and identical-cards tests pin exactly the cases that could regress.
- The `exit 0` reasoning checks out against `powershell_output` on main, which discards stdout on a non-zero exit — "no registry entry for this adapter" is genuinely an ordinary outcome the merge already handles, so hiding the exit code there is sound.
- Parser covers both JSON shapes and string-encoded `Bytes`.

Nit, non-blocking: doc-comment typo in `read_windows_adapter_memory` — "the last statement s $?" should be "the last statement's $?".

The composition argument for >4 GB cards without a CIM fallback is reasonable given the measured 12 GB values; a one-command confirm from the 9070 XT reporter would close the last gap.

rileyorscheln:
Confirmed on the RX 9070 XT from #1811 (Windows 11).

CIM: `{"Name":"AMD Radeon RX 9070 XT","AdapterRAM":4293918720}` (4095 MiB, saturated).

`read_windows_adapter_memory` script: `[{"Name":"NVIDIA GeForce RTX 3070","Bytes":8589934592},{"Name":"AMD Radeon RX 9070 XT","Bytes":17095983104}]`, exit 0.

17095983104 is 16304 MiB. The 3070 row has no CIM counterpart (PnP status Unknown), so name pairing leaves it unused and replaces the saturated CIM value on the 9070 XT. Installed mesh-llm 0.76.0 still prints `VRAM: 4 GB`; the fix is on main and not in v0.76.2.

Virgile-pct:
Thank you @rileyorscheln, this is the confirmation I couldn't get myself: I don't have an AMD card, so your 9070 XT is the first real run of the case the fix was written for. Good to see the phantom 3070 row is left alone too.

You're right about the release: the fix (`7144cdf8f`) is on `main` but in no tag yet, and `v0.76.2` predates it. It will ship with the next release; the 0.77.0 milestone is still open, so I don't have a date. Until then, a build from `main` shows the 16 GB.

PR #1837, opened 2026-09-13
fix: preserve plugin models during health grace

michaelneale:
## Summary

Plugin-backed models remain discoverable and routable through the existing one-failed-probe health grace period, rather than disappearing while their endpoint is still marked available.

- Retain the last successful inventory during the degraded-but-available state.
- Withdraw inventory at the existing second consecutive failure threshold.
- Replace inventory with freshly discovered models on recovery.
- Cover the ordinary plugin endpoint health path with a real loopback HTTP fixture, checking all four probes and inventory transitions.

## Scope and review follow-up

The final diff is **one file**, `crates/mesh-llm-host-runtime/src/plugin/health.rs`.

- Removed the JSON parser change and its proxy tests at Mic's request. This PR does not alter JSON/UTF-8 validation or management-port request handling; the management-ingress regression and weak proxy assertions identified by Nick are no longer introduced here.
- Kept main's separately merged CI repair from #1842 unchanged. No cadence expansion, manifest regeneration, or native-gate wrapper change remains in this PR's diff.
- Updated against main `3f4f1c35a` without rewriting the existing branch history.

## Routing implications and limitations

Retained inventory also feeds the model advertisements sent to mesh peers (`mesh/announcements.rs`). Remote peers can therefore continue routing here during the grace window and receive an upstream failure if the endpoint really is down. This preserves the existing availability policy; it is not a guarantee that the endpoint will answer successfully. The missing-local-model path tries remote peers before the local plugin endpoint.

The pre-existing successful-HTTP-but-unparseable `/v1/models` behavior still clears inventory via `unwrap_or_default()`. Distinguishing malformed inventory from a legitimately empty model list remains a separate follow-up, not part of this change.

No configuration, CLI, plugin protocol, mesh wire protocol, or failure-threshold change.

## Validation

Fresh local validation passed on the exact tree committed as `e77dc32a5` (main `3f4f1c35a` plus the health-only diff):

- `cargo test --locked -p mesh-llm-host-runtime`: **3,534 passed, 11 ignored, 0 failed**, plus doc tests (0 tests).
- `cargo check --locked -p mesh-llm-host-runtime -p mesh-llm`
- `cargo clippy --locked -p mesh-llm-host-runtime -p mesh-llm --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `just no-console-print`
- `git diff --check` and `just check-commits`

Fresh GitHub checks are green at `e77dc32a5`: all five stable PR checks passed (Quality, Linux, Website, macOS, Windows); 39 successful checks, 17 planned skips, none pending or failed. Website/macOS/Windows summary checks passed through planner-selected skips for this narrowed diff. Linux executed Rust tests, native runtime and product checks, inference and two-node client smokes. The generic two-node smoke does not replace the plugin-specific qualification below.

**Not merge-ready on lab evidence alone:** no fresh multi-node qualification was performed. Before merge, exercise plugin inventory advertisement, peer routing during one failed probe, withdrawal after the second failure, and recovery on a two-node mesh. No live mesh instances were started or stopped for this update.

### DIFF
--- a/crates/mesh-llm-host-runtime/src/network/openai/request_parse/json_validation.rs
+++ b/crates/mesh-llm-host-runtime/src/network/openai/request_parse/json_validation.rs
@@ -1,0 +1,65 @@
+use anyhow::{Context, Result};
+
+/// Reject syntax errors before routing can mistake them for a missing model.
+/// Uploads and plugin routes keep their own body contracts; tokenize already
+/// validates its metadata separately. Empty bodies retain existing behavior.
+pub(super) fn validate_inference_json(method: &str, path: &str, body: &[u8]) -> Result<()> {
+    if method == "POST"
+        && matches!(
+            path.split('?').next().unwrap_or(path),
+            "/v1/chat/completions" | "/v1/completions" | "/v1/responses" | "/v1/embeddings"
+        )
+        && !body.is_empty()
+    {
+        serde_json::from_slice::<serde_json::Value>(body)
+            .context("request body is not valid JSON")?;
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn inference_routes_reject_malformed_json_including_query_strings() {
+        for path in [
+            "/v1/chat/completions",
+            "/v1/completions",
+            "/v1/responses",
+            "/v1/embeddings",
+        ] {
+            for body in [
+                b"{broken".as_slice(),
+                b"  ",
+                b"{} trailing",
+                b"{\"x\":\xff}",
+                b"{\"x\":\"\xff\"}",
+                br#"{"x":"\uD800"}"#,
+            ] {
+                assert!(validate_inference_json("POST", path, body).is_err());
+                assert!(validate_inference_json("POST", &format!("{path}?trace=1"), body).is_err());
+            }
+        }
+    }
+
+    #[test]
+    fn validation_does_not_impose_a_model_or_metadata_schema() {
+        for body in [b"".as_slice(), b"{}", b"null", b"[]", br#"{"model":42}"#] {
+            assert!(validate_inference_json("POST", "/v1/chat/completions", body).is_ok());
+        }
+    }
+
+    #[test]
+    fn non_json_routes_and_other_methods_are_untouched() {
+        for path in [
+            "/api/objects",
+            "/v1/audio/transcriptions",
+            "/api/plugins/demo/http",
+            "/v1/tokenize",
+        ] {
+            assert!(validate_inference_json("POST", path, b"not json").is_ok());
+        }
+        assert!(validate_inference_json("GET", "/v1/chat/completions", b"not json").is_ok());
+    }
+}
--- a/.agents/skills/manage-ci/references/current-inventory.md
+++ b/.agents/skills/manage-ci/references/current-inventory.md
@@ -571,6 +571,9 @@ fail-open policy.
   integrity, family capability tags, and allowed suite/cadence membership.
   `scripts/generate-test-model-manifests.py` owns the family battery and
   suite-specific projections; CI contract tests reject stale projections.
+  The native runtime-event gate selects `family-qwen3-dense` from
+  `skippy-ci-smoke.json` at `pull-request` and `main` cadences; its separate
+  family-certification cadence list is unchanged.
 - `restore-test-model`: the single implementation of model resolve, cache,
   download, and verify. Resolves generated suite manifests, uses exact
   digest-bearing cache keys, and stream-verifies size and SHA-256 before use.
--- a/ci/ci.md
+++ b/ci/ci.md
@@ -373,7 +373,10 @@ runtime producers are not duplicated.
   producers selected by backend rows. The Linux CPU row additionally runs the
   native runtime-event gate
   (`scripts/ci-runtime-events-native-gate.sh`) against the runtime it just
-  built and a real model, and uploads its evidence file. That gate is
+  built and the registry-pinned `family-qwen3-dense` model from
+  `skippy-ci-smoke.json`, authorized at both `pull-request` and `main` cadences,
+  and uploads its evidence file. Family certification retains its separate
+  cadence list. That gate is
   env-gated so an ordinary `cargo test` never touches a native symbol, which
   is why it needs a lane of its own; this is the only lane that already has a
   freshly built native runtime. CPU only — the reporter is
--- a/ci/model-artifacts/manifests/competitive-benchmark.json
+++ b/ci/model-artifacts/manifests/competitive-benchmark.json
@@ -2,7 +2,7 @@
   "schema_version": 1,
   "manifest_kind": "test-model-artifacts",
   "suite": "competitive-benchmark",
-  "registry_sha256": "0fb272556fa31f2ab91d06c8265af82aa531c2001dbcbe6c3af3dab941140b3c",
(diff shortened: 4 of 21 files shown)


crates/mesh-llm-host-runtime/src/network/openai/request_parse/json_validation.rs
```diff
@@ -0,0 +1,65 @@
...
+/// validates its metadata separately. Empty bodies retain existing behavior.
+pub(super) fn validate_inference_json(method: &str, path: &str, body: &[u8]) -> Result<()> {
+    if method == "POST"
+        && matches!(
+            path.split('?').next().unwrap_or(path),
+            "/v1/chat/completions" | "/v1/completions" | "/v1/responses" | "/v1/embeddings"
+        )
+        && !body.is_empty()
+    {
+        serde_json::from_slice::<serde_json::Value>(body)
```
ndizazzo: We parse the whole body into a `Value` here and then drop it, and `rewrite_request_body_for_forwarding` parses it again at `request_parse.rs:506` and keeps that one.

That second parse isn't rare. `request_body_requires_json_normalization` returns true unconditionally for `/v1/responses`, and for any chat body mentioning `max_completion_tokens`, `max_output_tokens`, `reasoning`, `reasoning_effort`, `thinking_budget`, or `chat_template_kwargs`, which covers most current OpenAI clients. So for a large slice of traffic we now materialize up to 8 MiB into a `serde_json::Value` twice back to back, and on a mesh forward the receiving node pays it again.

To be clear I'm not asking you to weaken the parse, that part is right. Let's just stop throwing the result away. Something like returning `Result<Option<serde_json::Value>>` and threading it into `rewrite_request_body_for_forwarding` as a pre-parsed body, so line 506 becomes a move rather than a second parse. That's strictly better than where we are today on every path.

crates/mesh-llm-host-runtime/src/network/openai/request_parse/json_validation.rs
```diff
@@ -0,0 +1,65 @@
+use anyhow::{Context, Result};
+
+/// Reject syntax errors before routing can mistake them for a missing model.
+/// Uploads and plugin routes keep their own body contracts; tokenize already
+/// validates its metadata separately. Empty bodies retain existing behavior.
+pub(super) fn validate_inference_json(method: &str, path: &str, body: &[u8]) -> Result<()> {
+    if method == "POST"
+        && matches!(
+            path.split('?').next().unwrap_or(path),
+            "/v1/chat/completions" | "/v1/completions" | "/v1/responses" | "/v1/embeddings"
```
ndizazzo: `/v1/embeddings` isn't a route in this codebase. Grepping `crates/` for it only matches this file and the new tests, and the openai module's route vocabulary is `/v1/chat/completions`, `/v1/completions`, `/v1/models`, `/v1/responses`, `/v1/tokenize`.

Harmless and defensible as forward-looking, but `test_api_proxy_rejects_malformed_json_on_other_inference_routes` asserting a 400 there reads like coverage of a supported route when the only behavior that path has is this validator. Maybe drop it until the route exists, or leave a comment saying it's pre-registered.

Small related thing: this is an exact match on `path_only`, so `/v1/chat/completions/` skips validation. `is_tokenize_request` right below handles `"/v1/tokenize" | "/v1/tokenize/"`, so the inconsistency is visible in the same file. Low risk since the trailing-slash form isn't routed, but matching the existing convention is free.

ndizazzo (dismissed):
Both of these are worth having and the core logic is right where it counts. The health grace change is genuinely small, and the JSON validator is correctly narrow: allowlisted paths, POST only, non-empty bodies only, and it handles query strings.

I also went and checked the one decision I expected to push back on, using a full `Value` parse instead of a cheap skip-parse. You're right to do it that way. serde_json's skip path doesn't validate UTF-8 in skipped strings and doesn't check surrogate pairs (there's an explicit "we don't care if the codepoint is valid" comment in `ignore_escape`), so `IgnoredAny` would have accepted bodies the downstream parse rejects. That would have reintroduced a subtler version of the same bug. Good call, and worth saying out loud since it looks like over-engineering until you check.

What we need to fix is a third ingress path the change doesn't account for.

`read_http_request` is shared by three readers, not two. The two OpenAI ones (`transport.rs:311`, `ingress.rs:1175`) funnel the error into `send_read_failure` and return the 400 you intended. The management port doesn't. `read_management_request` at `api/server.rs:325` propagates the error up out of `handle_request`, and the accept loop at `api/server.rs:47` logs it at `debug!` and drops the socket. `POST /v1/chat/completions` is a real route there, `api/routes/mod.rs:185` matches `p.starts_with("/v1/")` and hands it to the chat passthrough. So a client pointed at the management port with a trailing comma in its body used to get proxied to the inference port and get a sensible error back, and now it gets a closed connection with no response at all. That's worse than the misleading 503 we're removing, and it contradicts what the PR says it does.

Let's map that read failure to a 400 instead of propagating it. `respond_error` is already in scope in that file and used throughout `dispatch_management_request`. And let's add a management-port case to the new regression tests, there isn't one.

Related, and the reason I'd like tests here specifically: three of the new proxy tests can't fail the way they're meant to. `send_request_and_read_response` uses `read_to_end`, so a server that writes nothing returns an empty string, and an empty string happily satisfies `!response.contains("not valid JSON")`. Those assertions pass identically on "correctly routed" and on "connection died". Since the management path above does exactly that, this isn't hypothetical. Let's assert the positive status instead, and for the `/api/objects` case assert the upstream actually received the bytes through the `upstream_rx` channel the test currently discards.

On scope: 17 of the 21 files are a CI change that the title, summary, and validation section never mention. The `pull-request` and `main` cadences added to `family-qwen3-dense`, the `registry_sha256` regeneration that forces across all 12 manifests, the doc updates, and the relative-path fix in `ci-runtime-events-native-gate.sh`. That CI work looks fine on its own (the `$PWD` prefixing correctly lands before the `cd "$ROOT"`, and `test_relative_paths_survive_cargo_changing_to_the_crate_directory` is a good test), but it's a separate concern with a separate failure mode. Buried under a title about plugin health and JSON validation, the cadence expansion gets no reviewer attention at all. Let's split it out.

A few things that aren't blocking but you should know about:

The grace window's blast radius is wider than the description suggests. Retained inventory also flows into mesh gossip through `announcements.rs:369`, so remote peers will route here during the window and collect the 503. I do think one failed probe of grace is the right tradeoff, and `route_missing_local_model` tries remote peers before the local plugin endpoint so this can't cannibalize a healthy peer serving the same model. But that's the thing the multi-node qualification should actually exercise before this merges, and it belongs in the description.

A successful probe with an unparseable `/v1/models` body still withdraws everything in one cycle, because `health.rs:372` does `.unwrap_or_default()`. Pre-existing and not yours to fix here, but it means one of the two realistic flap sources stays open. Worth a follow-up that distinguishes "probe succeeded but the body didn't parse" from "probe succeeded and legitimately reported zero models".

Also worth calling out in the description: invalid UTF-8 inside a JSON string used to survive (the unknown-field skip doesn't validate it) and got forwarded verbatim to a possibly-permissive upstream. Those requests now 400. I think that's the right call, it isn't valid JSON, but it's a wire-behavior change beyond "syntax errors".

Things I liked. The loopback socket fixture for the health path is the right shape, and asserting the request count means it can't silently pass on a skipped probe. The 400 body comes out as the real OpenAI error schema rather than something bespoke, and anyhow's `Display` keeps the serde line/column out of the response so we're not echoing request content back. And putting the validator in its own submodule instead of growing the 1,226-line `request_parse.rs` is exactly what our own 1k rule asks for.

Issue #1763, opened 2026-09-11
task: retire the console-print ratchet — convert the remaining 1,056 allowlisted call sites and delete the allowlist


ndizazzo:
## Goal

Empty and then delete the console-print ratchet: every remaining occurrence in `tools/xtask/data/console_print_allowlist.json` is either converted to the events/logging facility or removed from the gate's scope by an explicit rule. When this is done, the allowlist file is gone and `just no-console-print` passes with zero approvals.

## Current state (`origin/main` @ `d4ffbba`)

- **1,056 approved occurrences across 115 files.**
- The gate lives in `tools/xtask/src/no_console_print.rs` (`check_no_console_prints`, line 200), runs via `just no-console-print` (`just/ci.just:31`) and is enforced in CI at `.github/workflows/ci-quality-slice.yml:157`.
- It forbids `println!`, `eprintln!`, `print!`, `eprint!` in **every** `.rs` file under `crates/` except `build.rs` (`collect_rs_files`, line 165). It approves exact `(file, line, macro)` triples, so moving a legacy print is a failure, not a free pass. The error message directs authors to `mesh_llm_events::emit_event`.

### What is actually in the allowlist

Classified by whether the file ships in the `mesh-llm` binary's dependency graph (`cargo tree -p mesh-llm --edges normal`):

| Bucket | Occurrences | Notes |
|---|---:|---|
| **Product code** (shipping crates, non-test, non-helper-bin) | **696** | the real work |
| Helper binary in a shipping crate | 31 | `crates/model-package/src/bin/queue-unsloth-layer-packages.rs` |
| Non-shipping crates | 191 | `skippy-prompt` 94, `skippy-bench` 46, `skippy-model-package` 15, `llama-spec-bench` 14, `skippy-quantize` 11, `skippy-correctness` 6, `mesh-client` 3, `mesh-llm-test-harness` 1, `metrics-server` 1 |
| Test / example code | 138 | `crates/*/tests/**`, `*/tests.rs`, `*_tests.rs`, examples |

Product bucket by crate: `mesh-llm-commands` 359, `mesh-llm` 195, `mesh-llm-host-runtime` 75, `skippy-server` 34, `mesh-llm-system` 12, `mesh-native-serving-plugin-host` 9, `mesh-llm-identity` 4, `mesh-llm-events` 3, `mesh-llm-tui` 3, `mesh-llm-runtime-install` 2.

Heaviest product files:

```
[...]

PR #1840, opened 2026-09-13
refactor(ci): scope console-print ratchet to product code

ndizazzo:
## Original problem

The console-print ratchet mixes shipping code with tests and standalone tools. This implements Stage 1 of #1763 so later conversions can focus on product output. It does not close the issue.

## Diagnostics

The baseline at `7eff57a16` contains 1,065 approvals across 118 files. Current Cargo metadata also shows that `mesh-llm-client` is a shipping dependency, so the proposed `mesh-client` exemption is not included.

## Fix

- Exempt test paths, parsed `#[cfg(test)]` modules, examples, benches, auxiliary `src/bin/` targets, and eight explicit non-product crate directories. Keep the shipping entry point in scope.
- Check Cargo metadata on every gate/regeneration invocation and in a unit test. Reject exempt crates reachable through normal dependencies, including optional, renamed, transitive and platform-specific dependencies.
- Preserve exact occurrence positions when masking test modules. Unparseable fragments stay fully checked.
- Reduce the allowlist to 690 approvals across 55 files, retiring 375 approvals without adding any. Keep CI workflow wiring unchanged.

The three exemption classes requested by the issue are presented here for maintainer review before merge. Product output conversion, direct stdout/stderr handle detection and final allowlist deletion remain for later stages. No CLI or runtime behavior changes, so before/after command-output comparisons do not apply.

## Validation

- [x] All 42 xtask tests pass, including dependency, scope, source-position and regeneration regression tests.
- [x] `just no-console-print` passes.
- [x] xtask compile check, warning-denying Clippy and workspace Rust format check pass through `just with-lld`.
- [x] `just ci-validate` passes, including all 1,440 Python tests with 9 expected skips and all repository consistency gates. Run with Python 3.12, modern Bash and Homebrew coreutils on PATH.
- [x] No UI changes; screenshots do not apply.

## Native gate CI repair

The first Linux run exposed a model-selection mismatch in the existing native runtime-event gate: it requested `family-qwen3-dense` for pull requests and main, but the artifact registry allowed neither cadence. This PR adds those two uses to the canonical registry and regenerates its suite manifests. The model revision, file size and SHA-256 are unchanged; the other generated manifests change only their registry checksum.

A new workflow contract test invokes the real resolver for both cadences, using the artifact and manifest selected by the workflow. This catches the mismatch before an expensive native build. The 33 focused gate/registry tests pass locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Console-print checks now consistently cover product code while excluding tests, examples, benchmarks, helper binaries, and other non-product sources.
  * Added safeguards against exempt crates entering shipping dependency paths.
  * Native runtime-event checks now resolve bundle, model, and evidence paths consistently from the caller’s working directory.

* **Documentation**
  * Documented console-print scope rules, dependency safeguards, and runtime-event behavior.
  * Updated CI and llama canary guidance, including certification, verification, and artifact-handling procedures.
  * Refreshed the console-print allowlist to reflect current product-code coverage.


i386 (dismissed):
Reviewed the synchronized exact head. The console-output scope and generated allowlist are consistent with main; xtask tests and the repository console-output contract pass locally.

i386 (approved):
Reviewed the full diff at e20c556ba.

**Scope logic (tools/xtask/src/no_console_print/scope.rs)** — sound. Exemptions are category rules (tests/, examples/, benches/, src/bin auxiliaries, build.rs, and an explicit `NON_PRODUCT_CRATES` list), not per-site approvals. The Cargo-metadata guard rejects any exempt crate that becomes a normal dependency of `mesh-llm` (including optional/renamed/transitive/platform-specific), which closes the loophole where an exemption could later hide product code. `mesh-client` stays in scope, correctly, since `mesh-llm-client` is a shipping dependency. `#[cfg(test)]` masking via `syn` preserves byte positions and fails open (unparseable fragments stay fully checked) — the safe direction for a ratchet.

**Allowlist** — only deletions; 1,065 → 690 approvals across 118 → 55 files, no new approvals added. Verified locally at the exact PR head (e20c556ba, clean worktree): `cargo run -p xtask -- repo-consistency no-console-print` passes and all 42 xtask tests pass.

**Native gate repair** — the registry/manifest additions and the absolute-path resolution in `ci-runtime-events-native-gate.sh` are narrowly scoped and test-covered.

**Note on the red macOS checks**: `Platform checks (macos-unit)` failed on one unrelated test — `runtime_data::collector::tests::shadow_compare_uses_the_snapshot_from_its_own_write` (left: 2, right: 1). This PR touches no code under `crates/mesh-llm-host-runtime`; the test is a concurrent shadow-compare assertion that came in with #1777 and looks flaky. It passed in every Linux Rust test batch. Worth a retry; not a blocker for this diff.

Approving on the strength of the change itself. CI redness is the flake above plus the lane-aggregation failures it cascaded into.

i386 (approved):

i386 (dismissed):

i386 (approved):

Issue #1702, opened 2026-09-08
[Protocol] Host-served raw-proxy path publishes no `openai.exchange.v1` terminal event — and terminal-envelope enrichment (usage, digests, serving provenance): options + a recommendation


StevenMih:
Companion to #1669 (the *routed-to-peer* gap). This is about the other uncovered branch — the raw-proxy **host-served** path — plus the *shape* of what the terminal envelope should carry. Raising it as a design read before opening PRs, so what follows is small merges rather than a design argument in a PR thread. Everything below exists and is tested on a public fork; nothing here is hypothetical.

## What's covered today

#1437 landed `openai.exchange.v1` terminal events for exchanges a node serves, on the typed-frontend seam and the raw-proxy **plugin-served** path. #1668 adds the **routed-to-peer** path (#1669's gap). The remaining uncovered branch is the raw-proxy **host-served** path (real weights, `route_ingress` → `route_model_request`): it publishes nothing. On a serving node that's most exchanges — so a lifecycle-hook plugin sees every exchange *except* the ones the host itself served.

## What a downstream consumer needs, and why (one line each)

| Field | Host already computes it? | Why it's useful |
|---|---|---|
| publish on the host-served path | yes — same publish pattern as plugin-served | otherwise served exchanges are invisible to a plugin |
| `usage {prompt, completion, total}` | yes — `RespondedWithUsage` carries it | the honest denominator for anything billing- or audit-shaped |
| `request_digest`, `response_digest` | bytes are in hand at the publish site | binds a record to *what was asked / answered* without carrying text |
| `served_by_node_id` | yes — the node's own endpoint id | which node served (vs. routed) |
| model identity (`model_identity_hash`, `canonical_ref`, `revision`, `quantization`, `architecture`, `context_length`, `parameter_size`, `layer_count`) | yes — `ServedModelMetadata` / `ServedModelIdentity` | "a model by a *name*" → a model by what was actually loaded |
| hardware (`gpu`, `vram_bytes`, `is_soc`, `hostname`) | yes — startup hardware survey | which box, as the host itself reports it |

Every field is real-or-omitted (`skip_serializing_if`); nothing is fabricated; **no prompt or completion text is ever carried** — digests only.

## Options

**A. One PR, the whole block.** Publish on the host-served path + `usage` + digests + a `serving_provenance` struct with all of the above.
[...]

PR #1841, opened 2026-09-13
feat(openai): serving provenance, usage and request_digest on the host-served terminal event

StevenMih:
Part of #1702. Follows #1331 (lifecycle spec) and #1437 (its implementation). Parent strategy: #1233.

## What

A consumer of `openai.exchange.v1` can see *that* an exchange completed but not *what served it*. The terminal envelope carries `exchange_id`, `model` and `status`; it doesn't carry the facts a downstream reader needs to tell one serving configuration from another.

This adds three optional blocks to the raw-proxy terminal envelope (see Scope for which path carries which), populated from facts the host already resolves at serve time:

- **`serving_provenance`** — served node id, hostname, GPU and VRAM, quantisation, architecture, native context length, parameter size, layer count, model identity hash, canonical ref, revision.
- **`usage`** — the real token counts the dispatch outcome carries (`RespondedWithUsage`).
- **`request_digest`** — a canonical digest of the request body as received, so a consumer reconciling two accounts of the same exchange can tell whether they were asked the same thing.

Every field is **absent when the host doesn't know it — never a fabricated or zeroed value.**

## Why

Two nodes can advertise the same model name and serve materially different things. A node on a reduced KV-cache policy returns a different quality of answer under an identical model reference, and nothing in the current event distinguishes them. Anything reconciling exchanges across nodes — billing, capacity accounting, a receipts plugin (#1332) — has to guess.

**This is also the envelope #1708's `weights_digest` is meant to ride.** That PR's open question is that nothing consumes the digest; the served-model identity block this PR adds is where it goes. Landing this answers that question with a branch rather than a comment, and I'm happy to sequence the two however you'd prefer.

Scope

Both raw-proxy dispatch paths publish the terminal event now. What each carries:

serving_provenance — host-served path only, and only on a served 2xx. Absent on the plugin-served path (a plugin endpoint can proxy anywhere; this node's hardware and weights are not what served it) and absent on any 503 / Failed / Dropped outcome.
usage — either path, whenever the dispatch outcome carried the backend's real token counts (RespondedWithUsage), regardless of status. Never derived, never zero-filled; cached_prompt_tokens carried when reported.
request_digest — both paths, whenever the host held a parsed JSON body; omitted for a body the reference would refuse (integers beyond ±(2^53−1)).

No gossip-wire change. No new dependency. Additive and Option-shaped — existing consumers unaffected.

## Testing

Cut fresh off `origin/main`, rebased onto current `main` (`b7e4021a8`), local gate green at that head:

```
✅ rustfmt
✅ repo-consistency:no-console-print
✅ repo-consistency:publish-crates
✅ repo-consistency:test-all-rust-crate-coverage
✅ clippy:mesh-llm-host-runtime
✅ test:mesh-llm-host-runtime
```

The full Linux quality slice (`ci-quality-slice.yml`'s Docker approximation — actionlint, the `scripts/tests` contracts, the three `xtask repo-consistency` checks, `no-console-print`, workspace fmt and clippy) was also run green on the pre-rebase head `76622474a`.

**Nine new tests, each shown failing with its fix reverted, then passing restored:**

- `terminal_carries_serving_provenance_and_omits_unknown_fields` — a no-op `with_serving_provenance` drops every field to `null`
- `effective_envelope_has_no_serving_provenance` — provenance leaking onto the pre-serve event
- `terminal_carries_real_usage_when_attached` / `terminal_omits_usage_when_none_attached` — a no-op builder, and a fabricated all-zero usage replacing a genuine absence
- `request_body_digest_matches_python_reference` — float handling drifting from the reference implementation
- `request_body_digest_does_not_normalize_absent_fields` — isolates the absent-field normalization regression the previous test cannot see (its fixture has no null fields)
- `terminal_carries_request_digest_when_attached` / `terminal_omits_request_digest_when_none_attached` — a no-op builder, and a fabricated digest replacing a genuine absence
- `ingress::tests::exchange_usage_from_outcome_extracts_real_counts_and_omits_otherwise` — real usage silently failing to reach the envelope

## One note on `request_digest`

The digest deliberately does **not** apply an absent-field normalization step. That matches the current Python reference used by the receipts plugin (verified by running it with and without null-valued optional fields — the digests differ, as they must), and it's pinned by the two `request_body_digest_*` tests above. A separate Rust component on our side still normalizes and has drifted from that reference; that's a fix in our repo, not this one, and it's tracked there.

Quick update on where we are:

Sealed per-exchange records on both sides — the serving node and the requesting node each keep their own half. Live on a 3-node mesh now (two Macs + a GCP box).
A twin cross-check on a sample of requests (~1 in 50, so ~2% extra inference), and when two providers actually disagree, a third independent node recomputes and breaks the tie. Honest agreement means zero referee calls. Tested end to end this weekend — we planted a liar and it got caught.

Things that would help:

A design read, not a merge: I'm posting a short options-and-recommendation issue on the Terminal envelope (usage, digests, provenance on the host-served path). Everything else stands on that shape. "Looks right / change X" from you or James is worth more than any merge this week.
Split models: this is where the design already points. The coordinator is the composer, each stage seals its own half correlated by run/hop, and a twin re-runs the pipeline through a different stage set, so the composition gets checked, not just one node. The one thing it needs from the host is a per-stage lifecycle event, same shape as #1437's. I'd rather write that against your split branch than guess — point me at it when it's reviewable?

Happy to walk through the record format or the referee selection any time.

## Summary by CodeRabbit

* **New Features**
  * OpenAI exchange events now include serving details such as node, model, hardware, and configuration metadata when available.
  * Exchange events now report accurate backend token usage, including cached prompt tokens.
  * Requests can now be identified using a canonical request digest that preserves large numeric values accurately.
  * Enriched exchange events are available for host-served and plugin-served exchanges, with serving details included where applicable.
  * Host-served exchanges now publish effective events before dispatch and terminal events after completion.


crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
```diff
@@ -123,10 +244,178 @@ impl OpenAiExchangeEnvelope {
...
+}
+
+/// Replace every JSON float with its exact decimal-string form (mirrors the
+/// Python reference's `_stringify_floats`).
+fn stringify_floats(value: &serde_json::Value) -> serde_json::Value {
+    use serde_json::Value;
+    match value {
+        Value::Number(n) if n.is_f64() && !(n.is_i64() || n.is_u64()) => {
+            let f = n.as_f64().expect("n.is_f64() confirmed a f64 is present");
+            let s = format!("{f}");
```
ndizazzo: This diverges from the Python reference for any float outside roughly `[1e-4, 1e16)`, silently.

Rust's `Display` for `f64` never emits exponent notation. Python's `repr`/`str`/`json.dumps` switch to exponent form once the decimal exponent is `>= 16` or `< -4`. I ran both:

| input | this code | Python `repr` |
|---|---|---|
| `1e-5` | `0.00001` | `1e-05` |
| `1e-7` | `0.0000001` | `1e-07` |
| `1e16` | `10000000000000000.0` | `1e+16` |
| `1e20` | `100000000000000000000.0` | `1e+20` |

The `s.contains('e') || s.contains('E')` branch on the next line is dead code under Rust's `Display`, which I think is the tell that the Python conditional got transliterated without checking what Rust actually does here.

This is reachable from real traffic. `"temperature": 1e-9` is a common determinism trick, and we forward the whole body so arbitrary vendor float fields ride along. Same class of problem: a JSON integer literal above `u64::MAX` becomes an `f64` here but stays an arbitrary-precision int in Python.

No panic, no error, just a wrong digest. Since cross-implementation agreement is the entire point of this field, a silently wrong digest is worse than not having one.

Let's port Python's `float_repr` rule properly: shortest round-trip digits, then pick fixed vs exponent form on the decimal exponent (`exp < -4 || exp >= 16`), with the exponent written as `e+NN`/`e-NN` and at least two digits. Worth adding vectors for `1e-5`, `1e16`, `1e20`, `-0.0`, and a large integer literal, each cross-checked against a live `capsule_sidecar.digest_json` run. The two current vectors can't catch any of this, they only use `0.7` and `1.0`.

Minor, same block: `n.as_f64().expect(...)` on the line above is provably unreachable, but `if let Some(f)` costs nothing and removes a panic site from a function that eats attacker-supplied JSON. And without the `arbitrary_precision` feature (not enabled anywhere in this workspace) `is_f64()` is already exclusive with `is_i64()`/`is_u64()`, so that second clause is dead too.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -27,6 +28,122 @@ fn plugin_route_status(outcome: &proxy::RouteDispatchOutcome) -> Option<u16> {
...
+    let provenance = serving_provenance_for_model(node, model_name).await;
+    let mut envelope = OpenAiExchangeEnvelope::terminal(
+        exchange_id.to_string(),
+        OpenAiExchangeDispatchPath::RawProxy,
+        model_name,
+        plugin_route_status(final_outcome),
+        None,
+        None,
+    )
+    .with_serving_provenance(provenance);
```
ndizazzo: This attaches provenance unconditionally, but the field's own doc at `openai_exchange.rs:183` says it should be `None` "on terminal envelopes where nothing was served (a denial/error before dispatch)". The code never produces that `None`, and since `ServingProvenance.served_by_node_id` isn't an `Option`, the block always comes out populated with at least node id, hostname, GPU, VRAM, and `is_soc`.

Concretely reachable with nothing served:

- `route_model_request` writes a host 503 at `transport_route_model.rs:99` when `ordered_candidates` is empty after health filtering, and again from `finish_exhausted_route_model_request` when every target failed.
- `try_route_plugin_model` at `ingress.rs:783` turns a `Failed(_)` plugin outcome into a host-written 503.

Worse, on `Failed(_)`/`Dropped(_)` `plugin_route_status` returns `None`, so the envelope carries a full provenance block and no status at all. A consumer can't tell that apart from a successful serve.

This is CodeRabbit's flag and I think it's right. It matters more than it looks because the doc positions this as the proof-of-inference provenance a downstream capsule attests over.

Something like gating the attachment on the outcome actually being a served response:

```rust
if matches!(
    final_outcome,
    proxy::RouteDispatchOutcome::Responded(200..=299)
        | proxy::RouteDispatchOutcome::RespondedWithUsage { status_code: 200..=299, .. }
) {
    envelope = envelope.with_serving_provenance(
        serving_provenance_for_model(node, model_name).await,
    );
}
```

which also lets us skip the `served_model_descriptors()` lock and clone entirely on the non-served path. Whichever rule you pick, let's make the doc comment and the code say the same thing.

ndizazzo (changes requested):
This is the right direction and the problem it solves is real. The host-served raw-proxy path published no `openai.exchange.v1` terminal event at all before this, and the three blocks you're adding are the facts a reconciler actually needs. A lot of the detail work is careful too: the `vram_bytes != 0` guard, `skip_serializing_if` on every optional, no new dependency, and using `served_model_descriptors()` rather than the all-peers accessor so peer-advertised descriptors can't get misattributed to this node. That last one is easy to get backwards and you got it right.

I also re-derived both pinned digest vectors from a from-scratch implementation and they match exactly, and sorting keys by UTF-16 code units rather than leaning on serde_json's `BTreeMap` ordering is correct per RFC 8785 and survives someone turning on `preserve_order` somewhere in the graph. Those are the details these ports usually get wrong.

That said, I don't think we can land it as-is. The PR's central claim is "every field is absent when the host doesn't know it, never a fabricated or zeroed value", and that claim doesn't hold on three paths. Since the whole value of this event is that a downstream consumer can trust it, those aren't cosmetic.

Two of them I've left inline. The third is about scope.

Host hardware provenance is being attached to plugin-served exchanges. `try_route_plugin_model` calls `publish_raw_proxy_terminal` at `ingress.rs:805`, and `serving_provenance_for_model` fills `gpu`, `vram_bytes`, `is_soc`, and `hostname` from this node's startup hardware survey unconditionally. A plugin endpoint can proxy to anything, including a third-party cloud API, so we'd be announcing "served on host-1, RTX 4090, 24 GB VRAM" for an exchange that ran somewhere else entirely. That's a fabricated hardware attestation, which is the exact thing the PR says never happens.

There's a second-order version of the same problem: the descriptor lookup at `ingress.rs:41` matches `node.served_model_descriptors()` by model name. The descriptor list and the target table are updated independently, so during a teardown window a stale descriptor can attach local weights' `quantization`, `architecture`, and `model_identity_hash` to a plugin-served exchange. Let's pass a served-locally signal into `publish_raw_proxy_terminal` and omit the hardware block (or the whole provenance block) on the plugin path.

While we're here, the PR body says "Scope: Host-served path only", and that isn't accurate. CodeRabbit's generated release note actually has it right: it covers both. Worth correcting so the next reader isn't misled.

Smaller things, none of them blocking:

`exchange_usage_from_outcome` at `ingress.rs:82` guards with `||` rather than `&&`, so `prompt=Some(42), completion=None` emits `completion_tokens: 0`, and it derives `total_tokens` from a sum when the real total is absent. Both are fabricated values in a struct whose doc says nothing is fabricated, and the new test pins the derived-total behavior. We already own this invariant elsewhere: `mesh_llm_events::logging::events::TokenUsage::from_counts` documents "Missing, overflowing, or internally inconsistent usage must not be estimated" and requires all three counts to agree. I checked and this is currently unreachable, every production `TokenUsage` comes through `from_counts`, so it's latent rather than live. Still, a three-way `let ... else` is cheaper than the guard we have. Separately, `cached_prompt_tokens` gets dropped, and for billing reconciliation that's usually the difference between a right and a wrong number.

Both files land over our 1k line rule: `openai_exchange.rs` goes 567 to 1046, `ingress.rs` 1361 to 1529. The digest code is about as clean a separable responsibility as exists, `request_body_digest` plus `stringify_floats` plus the `jcs_*` helpers are a self-contained RFC 8785 port with zero coupling to the envelope. Pulling them into `plugin/openai_exchange/canonical_digest.rs` with their tests would take the file back under on its own.

The publish sits on the hot path and is gated on `ctx.plugin_manager.is_some()`, not on anything actually subscribing. `broadcast_channel_message` only delivers to plugins that declare `openai.exchange.v1`, so a host with any plugin loaded pays a full deep clone of the parsed body, a second full-size canonical string, SHA-256 over it, and a clone of the whole `Vec<ServedModelDescriptor>` under a mutex, per chat request. And the `publish` is awaited before `route_model_request`, so it's sitting directly on time-to-first-token. A cheap "does any plugin declare this channel" check would skip all of it.

On the tests: they do match what the body describes, and I verified the digest vectors independently. But six of the nine only assert that `with_x()` sets `Some(x)` and that `skip_serializing_if` works, which is close to tautological. Nothing covers the actual behavior change. There's no test that the host-served path publishes an effective/terminal pair at all, nothing pinning what happens on a 503 or a `Failed` or `Dropped` outcome (which is exactly the bug above), nothing for `serving_provenance_for_model` on a descriptor miss or the `vram_bytes == 0` omission, and the digest has no coverage for non-BMP keys, control characters, nested arrays, or a non-object top level. `ingress_tests/durable_artifacts.rs` already drives `handle_api_proxy_connection` end to end and `RecordingChannel` already exists, so most of the harness is there.

One last note, worth a sentence in the doc rather than a code change: because `stringify_floats` turns `0.7` into `"0.7"`, the bodies `{"temperature": 0.7}` and `{"temperature": "0.7"}` digest identically. Inherited from the reference so not yours to fix, but it's a real weakness in something we're using as an integrity binding.

CI isn't green at head, by the way. The Linux cpu runtime job failed on the restore-model step (looks like infra, not you), Quality shows cancelled, and a few batches were still pending. Worth a re-run.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
StevenMih: Thanks — agree on all three, they're real "never fabricated" gaps, and the float one is the kind that hides. Rework coming as separate commits in your order, rebased onto current main; the 0.7/"0.7" point is a property of the profile's digest construction and I'll get it declared where a verifier can read it.

StevenMih:
Thanks for the thorough pass — replying in the order you raised things,
one commit per item so it's reviewable piecewise.

**1. Served-state gate** (main review body + your inline comment on
`ingress.rs:128`, and CodeRabbit's flag on the same line). Fixed in
`210f5e654`: `publish_raw_proxy_terminal` now only attaches `ServingProvenance`
when the outcome is an actual 2xx (`Responded`/`RespondedWithUsage` in
`200..=299`, via a new `outcome_was_served`); on a 503 (host-exhausted or a
plugin `Failed(_)` converted to a written 503), a `Failed(_)`, or a
`Dropped(_)` the block is fully absent, and `serving_provenance`'s doc comment
now states the same 2xx-only rule the code enforces. Tests:
`publish_raw_proxy_terminal_on_a_503_has_no_provenance_but_keeps_the_status`,
`publish_raw_proxy_terminal_on_a_failed_outcome_has_no_provenance_and_no_status`,
`publish_raw_proxy_terminal_on_a_dropped_outcome_has_no_provenance_and_no_status`
(all `8dd35d1bb` — the fix itself landed in `210f5e654`, ahead of the test
harness rework that made calling the builder directly possible).

**2. Plugin-served hardware/model-identity attribution.** Fixed in
`b5265864d`: `publish_raw_proxy_terminal` takes an explicit `served_locally`
signal, and the WHOLE `serving_provenance` block — not just `gpu`/
`vram_bytes`/`is_soc`/`hostname` — is omitted on the plugin-served path, even
on a 2xx. I went with omitting the whole block rather than just the hardware
fields: the descriptor-staleness window you flagged (model-name-keyed lookup
racing the independently-updated routing target table) is a second,
independent way to attribute this node's own local weights to a plugin-served
exchange, and omitting the whole block closes both at once instead of leaving
that half open. `try_route_plugin_model` passes `false`; the host-served
`route_request` passes `true`. Test:
`publish_raw_proxy_terminal_on_the_plugin_served_path_omits_the_whole_block_even_on_2xx`
(`8dd35d1bb`) — builds a node WITH real hardware and a matching served-model
descriptor and confirms the plugin path still reports nothing.

**3. PR body scope claim.** You're right, CodeRabbit's note had it correct.
Diff in `_work/pr-body-up-serving-provenance.md` (companion file, not a
commit) — dropped the blanket "Host-served path only" / "three optional
blocks to the **host-served** terminal envelope," and instead said precisely
which blocks are host-served-only (`serving_provenance`, `usage`, both only on
a served 2xx) vs. present on both dispatch paths (`request_digest`, whenever
the host held a parsed JSON body).

**4. `exchange_usage_from_outcome`.** Fixed in `832b0ce94`: it's now a
three-way `let … else` on prompt/completion/total — all three or none, never
a derived total, never a zero standing in for a missing count — and
`cached_prompt_tokens` now rides along instead of being dropped. Test:
`exchange_usage_from_outcome_extracts_real_counts_and_omits_otherwise`,
extended with cases for the disagreeing-real-total (must ride through as
reported, not get silently replaced), the missing-completion (must not
fabricate a zero), and the cached-count carry.

**5. File size.** Partially fixed. `76d18ddf2` moves `request_body_digest`,
`stringify_floats`, `float_repr`, and the `jcs_*` helpers, plus their tests,
into `plugin/openai_exchange/canonical_digest.rs` — pure move, no behavior
change (content is byte-identical, just relocated; `openai_exchange.rs` now
just has `mod canonical_digest; pub use canonical_digest::
request_body_digest;`). `openai_exchange.rs` is back under 1k, now 905 lines.
`ingress.rs` is not: it's 1599 lines at head, up from the 1529 you originally
flagged — none of this series' commits split it. I don't think it's worth
contorting an extraction into this already-large diff just to hit a number;
happy to take a follow-up on splitting `ingress.rs` (the
`serving_provenance_for_model`/`exchange_usage_from_outcome`/
`outcome_was_served`/`publish_raw_proxy_terminal`/
`mint_and_publish_effective_raw_proxy` cluster at the top of the file has the
same self-contained shape the digest code did) if you'd rather see that done
here instead.

**6. Hot path.** Fixed in `5b904cadc`: added
`PluginManager::any_plugin_declares_mesh_channel` and an
`OpenAiExchangeChannel::has_subscriber` trait method backed by it (default
`true`, so `RecordingChannel` and any other test double is unaffected by the
new method existing). Both dispatch paths check it before minting an
exchange id, computing the request digest, or looking up the served-model
descriptor — a mesh with a plugin loaded that doesn't declare
`openai.exchange.v1` now pays none of that per-request cost, matching
`broadcast_channel_message`'s own per-plugin filter instead of gating on
"a plugin manager exists at all." (`84171c89a` is a same-behavior follow-up:
pulling the mint-and-publish step out into its own function after the added
branch tripped `clippy::cognitive_complexity` on `try_route_plugin_model`.)

**7. Test coverage.** Fixed in `8dd35d1bb`. Seven new tests call
`publish_raw_proxy_terminal` directly through a `RecordingChannel` — the same
double you pointed at, now shared crate-wide via a
`plugin::openai_exchange::test_support` module instead of being private to
that module's own tests, which required changing
`publish_raw_proxy_terminal`'s second parameter from the concrete
`PluginManager` to `&dyn OpenAiExchangeChannel` so it's injectable. Covers: a
full served-2xx (provenance + usage + request digest all populated with real
values, not just `Some(x)`/`skip_serializing_if` tautologies); the 503/
`Failed`/`Dropped` triad above; the plugin-path whole-block omission above; a
served-model-descriptor miss (real hardware fields present, every
model-identity field `None`); and `advertised_memory.total_bytes == 0`
omitting `vram_bytes`.

One thing I did NOT do: wire these through the full `handle_api_proxy_
connection` TCP harness `durable_artifacts.rs`'s existing tests use. That
would need a real plugin subprocess declaring `openai.exchange.v1` in its
manifest to exercise the subscriber gate end to end (item 6 above), and this
crate's test infra doesn't have a lightweight in-process fixture for that —
`channel_broadcast.rs`'s own tests only cover the no-plugins-loaded case for
the same reason. Calling `publish_raw_proxy_terminal` directly gets the same
envelope-correctness coverage without a real subprocess. Flagging in case
there's an existing fixture pattern I missed that would make the full
end-to-end version cheap.

Each of these seven, plus the new `any_plugin_declares_mesh_channel` test for
item 6, was shown failing with its guard reverted and passing restored
(reverting the 2xx gate, the `served_locally` gate, the descriptor name
match, the `vram_bytes == 0` guard, the usage attachment, and stubbing
`any_plugin_declares_mesh_channel` to always return `true`, one at a time).
One of them — the descriptor-miss test — didn't bite on the first pass: the
test node had no descriptor registered at all, so a "wrong descriptor"
regression and an "empty list" case looked identical to it. Fixed in
`cde2a39df` by registering a descriptor for a different model instead, then
re-verified it fails against a broken name match.

**8. Float-stringify digest collision.** Doc note added in `7467996a7`, on
`stringify_floats` itself: because a float and its stringified form both
become JSON strings before JCS runs, `{"temperature": 0.7}` and
`{"temperature": "0.7"}` digest identically. As I said in my last reply,
this is a property of the digest context inherited unchanged from the
reference (`agent_action_capsule.canonical`/`capsule_sidecar.digest_json`),
not something this port introduces. It's currently declared as a PoC-only
property via `capsule-emit-mesh`'s `x-mesh-poc-v1` extension block in
`capsule_sidecar.py` (a namespaced, non-registered-spec-field extension,
per that file's own docstring) — registering it as a real profile-level
property is a separate spec-lane item, not this PR's to file.

**9. CI not green at head.** Rebased onto current `main`; local gate
(`rustfmt`, the three `xtask repo-consistency` checks, `clippy -p
mesh-llm-host-runtime --all-targets -- -D warnings`, `cargo test -p
mesh-llm-host-runtime`) is green at `cde2a39df`, and CI is running on the
pushed head.

---

**Separately, the float-repr port** (your inline comment on
`openai_exchange.rs:332`): fixed in `8285a41a1`. Ported the actual rule —
shortest round-trip digits (still sourced from Rust's own `Display`), fixed
vs. exponent form chosen on the decimal exponent (`exp < -4 || exp >= 16`),
exponent written `e+NN`/`e-NN` with a required sign and at least two digits.
The dead `contains('e')`/`contains('E')` branch is gone, and so is the
`.expect(...)` panic site you flagged (now `if let Some(f)` — provably
unreachable given the guard above, per your own note, but no reason to leave
a panic site over attacker-supplied JSON when the guard costs nothing).
Added the vectors from your table (`1e-5`, `1e-7`, `1e16`, `1e20`) plus the
boundary your table didn't cover (`1e-4` stays fixed), `-0.0`, `1e-9` (the
determinism trick you called out), `0.1 + 0.2`, a nested array, a non-object
top level, a non-BMP key, and a control character in a string — every vector
cross-checked against a live `capsule_sidecar.digest_json` run, values
recorded in the test comments.

On the large-integer case: a JSON integer literal outside the reference's
safe range (`±(2^53−1)`, `agent_action_capsule.canonical` §5.1, where the
reference raises `UnsafeIntegerError` rather than digest something it can't
represent losslessly) now makes `request_body_digest` return `None` instead
of silently round-tripping through `f64` — never a digest of a body the
reference would refuse.

Tests: `float_repr_matches_python_repr_at_and_around_the_exponent_boundaries`,
`request_body_digest_matches_python_reference_for_float_edge_cases`,
`request_body_digest_omits_when_body_has_an_unsafe_integer`,
`request_body_digest_matches_python_reference_for_structural_edge_cases`
(all now living in `plugin/openai_exchange/canonical_digest.rs` per item 5).

---

Head: `cde2a39df`, rebased onto current `main`, force-pushed over `eeb362ed6`.
Re-requesting review.

michaelneale:
I am ok with this - I think is reasonable

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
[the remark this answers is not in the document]
i386: Verified and fixed on exact head `5f8c594dd054f562472fec0917022319bbaa09c8`: terminal publication now receives an explicit served outcome and attaches `ServingProvenance` only for served 2xx responses. Plugin failures converted to host 503s and exhausted routes publish without provenance. The full host-runtime suite and repository gate pass.

crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
i386: Verified and fixed on exact head `5f8c594dd054f562472fec0917022319bbaa09c8`. Float rendering now follows the Python fixed/exponent boundary and has cross-checked vectors for `1e-5`, `1e-7`, `1e16`, `1e20`, `-0.0`, and `1e-9`. The raw request bytes are also scanned so integer literals beyond the safe range, including values above `u64::MAX` that serde_json would round to f64, omit the digest rather than fabricate one. Full host-runtime and repository validation pass.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
i386: Verified and fixed on exact head `5f8c594dd054f562472fec0917022319bbaa09c8`: provenance is gated on a served 2xx outcome, so pre-dispatch denials, plugin failures, and exhausted-target 503s carry no `ServingProvenance`. Successful host-served responses retain it. Full host-runtime and repository validation pass.

i386 (dismissed):
Reviewed the complete exact head `5f8c594dd054f562472fec0917022319bbaa09c8` after synchronizing it with current main. The provenance, usage, and request-digest paths now preserve the declared never-fabricated contract, including served-only provenance and safe handling of Python-compatible float rendering and oversized raw JSON integers.

Validation: full host-runtime suite passed (3,544 unit tests, 11 ignored, plus all integration and doc tests); `just ci-validate` passed (1,493 tests, 9 skipped). All three valid review threads are resolved.

StevenMih:
Pushed a965beb98 — merge of current main plus one test fix. The conflict was the one @i386's #1708 and this PR were always going to have: #1708 put Node::served_model_descriptors() behind #[cfg(test)] and this PR calls it in production (serving_provenance_for_model). Resolved by keeping the self-only accessor production-visible with a doc comment saying why — all_served_model_descriptors() includes peers' gossiped descriptors and can match a peer's same-named model, which is exactly the misattribution this PR exists to avoid. That's the reconciliation I flagged on #1708 as the second-lander's to carry.

a965beb98 on top is test-only: the two RemoteMesh/RawProxy shape tests that came in from main read channel.events directly, which is private on this branch's shared RecordingChannel; switched them to the events() accessor every other test here uses.

Local gate at head: build, cargo test -p mesh-llm-host-runtime (3,579 + 24 new, 0 failed), no-console-print, DCO.

One body correction, prompted by CodeRabbit's note on the last head: I'd written "serving_provenance and usage — host-served only, on a served 2xx," and that's only true of provenance. usage is attached whenever the outcome carried the backend's real counts, on either path and regardless of status — those counts aren't this node's claim, they're the backend's report, so omitting them on a plugin-served or non-2xx exchange would drop real billing data. The Scope section now says exactly that. If you'd rather usage be 2xx-gated to match provenance, say so and it's a one-line change; otherwise I'll add a test that pins the asymmetry explicitly.

@ndizazzo — the three items from your review are addressed on this head (served-state gate, plugin-path omission, the float-repr port), the body now matches the code, and CI is running. Whenever you can re-review.

crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
```diff
@@ -62,6 +62,102 @@ pub enum ClientNonceSource {
...
+/// the raw-proxy dispatch callsite in `network/openai/ingress.rs`): model
+/// metadata comes from the served-model descriptor (`ServedModelMetadata`:
+/// `quant`, `architecture`, `native_context_length`, `identity_hash`,
+/// revision/repository), and hardware comes from the node's startup hardware
+/// survey (`gpu_name`, `hostname`, `is_soc`, `advertised_memory`). No raw
+/// prompt or response text is carried — provenance only.
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct ServingProvenance {
+    /// The node that actually served the inference — this host's own mesh
+    /// endpoint id. On a plugin-served (raw-proxy) exchange this is the node
```
ndizazzo, line 83: This sentence is stale now that the whole block is omitted on the plugin path: the plugin-served case can no longer occur here, so a reader should not expect a plugin host to populate this field.

crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
```diff
@@ -105,6 +201,33 @@ pub struct OpenAiExchangeEnvelope {
     /// do not compare across nodes.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub nonce_source: Option<ClientNonceSource>,
+    /// What ran, at what fidelity, on whose hardware — see [`ServingProvenance`].
+    /// Present on a `Terminal` envelope only when the dispatch outcome was an
```
ndizazzo, line 205: Missing the second half of the condition: a plugin-served 2xx also has a 2xx outcome, but the field is None there. Something like "served locally and the outcome was an actual 2xx response" would be accurate.

crates/mesh-llm-host-runtime/src/plugin/openai_exchange.rs
```diff
@@ -105,6 +201,33 @@ pub struct OpenAiExchangeEnvelope {
...
+    /// actual 2xx response (`Responded`/`RespondedWithUsage`); `None` on
+    /// effective-request envelopes and on any non-2xx terminal envelope (a
+    /// denial/error before dispatch, a 503, or a dropped/failed connection) —
+    /// those served nothing, so there is nothing this field can honestly
+    /// report.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub serving_provenance: Option<ServingProvenance>,
+    /// The real token usage the served backend reported for this exchange (see
+    /// [`ExchangeUsage`]). Present on a terminal envelope for a host-served
```
ndizazzo, line 214: This says host-served, but the code attaches usage on either path when the outcome carries real counts. I checked route_http_endpoint_request: it returns RespondedWithUsage when the endpoint response carries a usage object, so a plugin-served exchange can carry usage. The PR body is right; this doc is the stale side.

crates/mesh-llm-host-runtime/src/network/openai/ingress.rs
```diff
@@ -27,6 +28,194 @@ fn plugin_route_status(outcome: &proxy::RouteDispatchOutcome) -> Option<u16> {
...
+    // see the `served_locally` doc above.
+    if served_locally && outcome_was_served(final_outcome) {
+        let provenance = serving_provenance_for_model(node, model_name).await;
+        envelope = envelope.with_serving_provenance(provenance);
+    }
+    // The host-served (real-weights) branch reaches this via
+    // `route_model_request`, whose outcome carries the served backend's own
+    // `usage` object. Attach it so a downstream plugin can seal the REAL token
+    // counts of the exchange, not a zeroed stub. A plugin-served exchange has
+    // no such usage on the outcome, so this stays absent for it — never zeroed.
```
ndizazzo, line 178: Same stale claim as the field doc: route_http_endpoint_request does return RespondedWithUsage when the plugin endpoint response carries a usage object, so a plugin-served exchange can have usage on the outcome.

ndizazzo (approved):
The three items from the last review are all addressed, and the rest of the round held. I re-read the full diff at head, re-derived the float port by hand, and re-ran the performance question you raised. This is in good shape.

**Plugin-path provenance.** The `served_locally` parameter does the job I asked for, and the doc on `publish_raw_proxy_terminal` spells out both failure windows, the third-party proxy case and the descriptor/target-table staleness case. Omitting the whole block rather than just the hardware fields is the right call, because a name-matched stale descriptor would otherwise attach local weights' identity to an exchange that ran somewhere else. On the host side, the `served_model_descriptors()` lock and clone are now skipped entirely whenever the outcome wasn't a served 2xx.

**Served-2xx gate.** `outcome_was_served` is the exact `matches!` I suggested, and the doc and the code now agree. `Responded(503)` carries a status with no provenance, and `Failed("connect_timeout")` and `Dropped("client_disconnected")` carry nothing, all pinned by the new durable-artifact tests.

**Float repr.** I re-derived the exponent-boundary logic by hand: the fixed/exponent split at `exp < -4 || exp >= 16`, the two-digit exponent padding, the `-0.0` sign, the trailing-zero trim that turns `2e17` into a `2e+17` mantissa of one digit. The boundary test now matches the Python `repr` values I measured, and the seven new digest vectors (`1e-5`, `1e-7`, `1e16`, `1e20`, `-0.0`, `1e-9`, `0.1 + 0.2`) are live-verified against `capsule_sidecar.digest_json`. The `expect()` panic site over attacker-supplied JSON is gone, and the dead `is_f64()` clause with it.

**Usage.** The three-way `let ... else` is exactly the shape I wanted: all three counts or nothing, `cached_prompt_tokens` carried, and the disagreeing total rides through as the backend reported it instead of being silently replaced by a derived sum. Pinning the `23 != 10 + 5` case is a nice touch, that's the behaviour a billing consumer actually relies on.

**Body.** It now describes what the code does, including the deliberate asymmetry that usage can appear on the plugin path. I agree with the reasoning: those counts are the backend's report, not this node's claim.

**Oversized-integer lexer.** I went over the byte scan carefully. The string-skip handles escaped quotes, the number-token character set is bounded, and the pure-digit check means false positives are impossible. It closes a real hole: serde_json rounds an integer literal above u64 to an f64, so without this `{"v":18446744073709551616}` would have been digested as the rounded float. Quoted values stay digestible and `1e20` stays digestible as a float, which matches the reference.

**Subscriber gate.** `any_plugin_declares_mesh_channel` is the cheap check I asked for, and I verified it's effective on both paths: with no subscriber the plugin path mints no id and publishes nothing, and the host path skips the entire digest block, no body parse, no canonicalization, no hash, no descriptor clone. A host with plugins loaded but no receipts subscriber now pays zero.

**On `Sha256::digest` in the hot path.** I'd keep it. The digest runs once per raw-proxy exchange, on the terminal path after the response has been dispatched, never in the per-token loop. Everything around it, the body parse, the JCS canonicalization, the hash, is behind the subscriber gate, so the cost is zero when nothing is listening. When something is, the cost at context-window scale is on the order of 1-2 ms: one body clone, the canonical string, and a pure-Rust SHA-256 pass at roughly 0.5-1 GB/s. Against inference that takes seconds to minutes that's negligible, and because the digest and terminal publish sit after dispatch, it's not on time-to-first-token; the pre-dispatch effective publish is just the envelope broadcast.

A faster algorithm would be the wrong trade: this digest has to match the Python `capsule_sidecar.digest_json` reference and the pinned vectors byte for byte, and cross-implementation agreement is the entire point of the field. If you want measured numbers instead of estimates, I can add a small benchmark.

The only leftovers this round are doc-level, and I flagged each one inline:

1. The `served_by_node_id` doc still carries the old sentence about the plugin-served case, which can no longer occur now that the whole block is omitted on the plugin path.
2. The `serving_provenance` field doc says "when the dispatch outcome was an actual 2xx response" and is missing the served-locally half of the condition. A plugin-served 2xx has the outcome and not the field.
3. The `usage` field doc says "host-served exchange", but the code attaches usage on either path when the outcome carries real counts. I checked `route_http_endpoint_request`: it returns `RespondedWithUsage` when the endpoint's response carries a usage object, so a plugin-served exchange can carry usage. The PR body is right; the doc is the stale side.
4. The JCS sort tie-break after the UTF-16 comparison is dead code (distinct strings can't share a UTF-16 encoding), and the structural-edge test doc claims a non-object top-level that its three vectors don't exercise.

On file size: the extraction landed and pulled roughly 500 lines out of the envelope module. `openai_exchange.rs` is 1121 and `ingress.rs` is 1686, and most of the remainder is unrelated work that merged into the same files since my last review (#1668, #1708). There's no CI-enforced 1k rule, so I won't hold this here.

CI is green at head.

ndizazzo:
@StevenMih I feel like this is close now - approving with some minor inline observations... feel free to take 'em, or if you prefer to merge and doc a full docs pass later, that's cool too

StevenMih:
thanks for the review, merging, will follow up with docs pass

PR #1844, opened 2026-09-13
fix(autoupdate): install composed product-v2 bundles in the self-updater

### LINKED ISSUE #1843: mesh-llm update fails with "Unexpected directory in bundle: native-runtimes" on every release since v0.75.0
## Summary

`mesh-llm update` has been broken for every in-app update that crosses into v0.75.0 or newer, on all platforms, since v0.75.0 shipped on 2026-08-07. The updater rejects the composed product-v2 bundle layout that releases have shipped since #1108/#1113.

Reported from the field (macOS aarch64):

```
 % mesh-llm update
⬇️ Updating mesh-llm v0.65.0+skippy.20260504.console.6 -> v0.76.1 (metal)...
   ⏬ mesh-llm-aarch64-apple-darwin.tar.gz complete    100.0% (54MB/54MB)
Unexpected directory in bundle: native-runtimes
```

## Root cause

1. Since v0.75.0 the release archives are composed product-v2 bundles: `mesh-bundle/` contains `mesh-llm`, `host-imports.json`, `product-manifest.json`, and a `native-runtimes/` directory with the packaged runtime.
2. The in-app updater still assumes the pre-0.75 flat layout. It extracts with `tar --strip-components=1` (`crates/mesh-llm-system/src/autoupdate/release_fetch.rs`), which leaves `native-runtimes/` as a top-level directory.
3. `collect_bundle_files()` bails on any directory: `Unexpected directory in bundle: {name}` — exactly the reported error.

Reproduced by applying the updater's own strip logic to the real `v0.76.1` asset (sha256 matches the release sidecar): result is `dirs=[native-runtimes]` → bail. The `v0.65.0` asset strips to five flat files, which is why updates used to work.

The flat-file assumption dates to the original updater (3342650f9, 2026-03-27); product-v2 bundles landed in ed41f366d (2026-07-29/30) and first shipped in v0.75.0. The bail is unchanged on current `main`, so this affects every platform (the Windows zip path fails the same way).

## Workaround

Reinstall with the install script, which handles product-v2 correctly:

```bash
curl -fsSL https://raw.githubusercontent.com/Mesh-LLM/mesh-llm/main/install.sh | bash
```

## Proposed fix

Teach `install_latest_bundle` the product-v2 layout (recursive staging of `native-runtimes/` with backup/rollback mirroring `install.sh`), validate the v2 contract in `collect_bundle_files` instead of bailing on directories, and add regression tests for the nested layouts on both archive lanes.

Fix incoming in a PR referencing this issue.

i386:
Fixes #1843

`mesh-llm update` now works against current releases again. Since v0.75.0 every release archive is a composed product-v2 bundle (host + `product-manifest.json` + `host-imports.json` + `native-runtimes/`), but the self-updater still assumed the pre-0.75 flat layout — it extracted the bundle, found the `native-runtimes/` directory, and aborted with `Unexpected directory in bundle: native-runtimes` before installing anything. Every in-app update crossing into v0.75.0+ has failed on all platforms since 2026-08-07; only the install script worked.

## What changed

- `collect_bundle_files` now understands both layouts. A nested `native-runtimes/` directory marks a composed product-v2 bundle: it is staged and installed recursively alongside the host files, and requires `product-manifest.json` (the v2 contract marker) to be present. Unexpected directories in legacy flat bundles are still rejected.
- Backup/rollback operate on relative paths and handle directories: the runtime tree is backed up before install (Unix), and a failed install prunes runtime directories it created so the previous product is restored exactly.
- The Windows apply-update script gets the same recursive semantics (`Ensure-Parent`, `Move-ToBackup`, recursive restore), fixing the identical Windows failure mode.
- The host binary installs first, so a mid-install failure can never strand a new runtime tree beside an old host.

## Validation

- `cargo test -p mesh-llm-system`: 216 passed, including new regression tests that build a composed product bundle, run it through the real extract → collect → replace pipeline for **both archive lanes** (nested `tar.gz` and zip, covering the Windows path), plus layout validation, backup restoration, and rollback coverage.
- `cargo fmt --all --check`, `cargo clippy -p mesh-llm-system --all-targets -- -D warnings`, `cargo clippy -p mesh-llm --all-targets -- -D warnings`: clean.
- All xtask repo-consistency ratchets on this branch: `release-targets`, `ci-crate-lists`, `publish-crates`, `test-all-rust-crate-coverage`, `no-console-print`: pass.
- Verified against the real published `v0.76.1` `mesh-llm-aarch64-apple-darwin.tar.gz` (checksum-matched): the old code produced the exact reported failure; the new `collect_bundle_files` accepts the bundle and stages `native-runtimes/` for recursive install.

## Protocol

No wire-protocol changes. This only affects local install mechanics of the self-updater.

## Summary by CodeRabbit

- **Bug Fixes**
  - Self-updates now support composed product-v2 release bundles containing native runtime directories.
  - Release bundle layouts are validated before installation.
  - Updates correctly install nested files and create required parent directories.
  - Failed updates now clean up newly installed runtime files and restore backed-up runtime trees during rollback.
  - Update installation preserves the required order for core application files and native runtimes.
  - Legacy release bundles remain supported.


i386:
@ndizazzo @michaelneale need a review so we can ship a quick follow up release.

ndizazzo (dismissed):

PR #1846, opened 2026-09-13
fix(scripts): clear stale llama.cpp CMake cache on generator mismatch

ndizazzo:
Native builds can fail with:

```
CMake Error: Error: generator : Ninja
Does not match the generator used previously: Unix Makefiles
Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
```

This happens when `ninja` appears on (or leaves) `PATH` between builds. `scripts/build-llama.sh` picks the CMake generator from `PATH` at configure time, but the build directory under `.deps/llama-build/` still holds a `CMakeCache.txt` configured with the previous generator, and CMake refuses to reuse a cache across generators. On a dev machine that installed Homebrew `ninja`, `just release-build` (and the other native build recipes) fail at the CMake configure step with no obvious remedy, because the stale cache lives outside the source tree.

## Diagnostics

- `just release-build` on `main` (clean tree, macOS arm64): host build succeeded, `release-runtime-build` failed at the llama.cpp CMake configure with the generator-mismatch error above.
- Compared the cached generator against the one the script selects: `CMAKE_GENERATOR:INTERNAL=Unix Makefiles` in `.deps/llama-build/build-stage-abi-dynamic-metal/CMakeCache.txt` (configured before `ninja` was installed) vs `Ninja` selected now.
- Removing the stale build directory and re-running `just release-build` completes: metal runtime packaged, tarball sidecar checksum OK.

## Fix

`scripts/build-llama.sh` now reads the generator recorded in the existing `CMakeCache.txt` right before the CMake configure. On a mismatch it clears the build directory (pure cache, regenerated on rebuild) and logs what it did. The guard sits after the stamp fast path and the `--require-existing` exit, so warm and pre-warmed caches that are only being verified are never touched, and already-valid outputs are still reused via the fast path.

`scripts/tests/test_build_llama.py` pins the behavior with stubbed `cmake`/`ninja` (self-contained, no toolchain required): stale cache cleared, matching cache kept, `--require-existing` untouched. `AGENTS.md` documents the failure mode in the Building section next to the npm exit-handler note.

## Validation

  - `python3 -m unittest scripts/tests/test_build_llama.py` (3/3 OK)
  - `python3 -m unittest scripts/tests/test_static_abi_artifacts.py scripts/tests/test_llama_upstream_canary_contract.py scripts/tests/test_llama_canary_agent_repair_contract.py` (60/60 OK, the contract tests that pin this script)
  - `shellcheck scripts/build-llama.sh` and `python3 -m py_compile scripts/tests/test_build_llama.py`: clean
  - End-to-end: after clearing the stale cache, `just release-build` completed locally (host + metal runtime + tarball, sidecar checksum OK)
  - No UI changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Native builds now automatically recover when an existing CMake cache was created with a different build generator, preventing “generator does not match” failures.
  - Build generator selection is now consistent, even when an inherited environment setting conflicts.
  - Existing compatible build caches continue to be reused, while required-existing builds retain their refusal behavior.

- **Documentation**
  - Added guidance for diagnosing and resolving CMake generator mismatch errors, including steps for older checkouts.

- **Tests**
  - Added coverage for generator selection, cache handling, dynamic builds, and required-existing behavior.


i386 (approved):
Reviewed the synchronized exact head. The CMake generator guard is scoped to rebuilds and safely clears incompatible stale build state; build-llama tests and shell validation pass locally.

PR #1847, opened 2026-09-13
test(host-runtime): stop test guards from using the real home on Windows

Virgile-pct:
Running `cargo test -p mesh-llm-host-runtime` on Windows deletes the developer's node key. This stops most of that, and measures what it does not.

## What happens today

Five test guards isolate the identity by pointing `HOME` at a temporary directory. `dirs::home_dir()` reads `$HOME` on Unix, so that works there. On Windows it goes through the known-folder API and ignores `HOME` entirely, so the tests fall through to the real `~/.mesh-llm`.

Measured on Windows 11 against `0e84ac776`, comparing the directory before and after one full run:

| | before | after |
|---|---|---|
| entries in `~/.mesh-llm` | 9 | 7 |
| node `key` | present | **gone** |
| `identities/` | absent | created by the run |
| `cache/hashes` | unchanged | 6 files added |

The probe behind it, with `HOME` pointed at a temp dir exactly as the guards do:

```
identity_home_dir()      = C:\Users\<me>
default_node_key_path()  = C:\Users\<me>\.mesh-llm\key
redirected               = false
```

## What this changes

`identity_persistence` already exposes the cross-platform hook, and `public_identity_tests.rs` already uses it: its `IdentityEnvGuard` sets `MESH_LLM_TEST_HOME` alongside `HOME` and restores both. The other five guards now do the same. Nothing new is introduced; five guards are aligned on the one that was already right.

## Measured effect

Full suite, same box, same command:

| | `main` | this branch |
|---|---|---|
| passed | 3454 | 3457 |
| failed | 20 | 17 |

The three that now pass are `persisted_random_mesh_id_is_preserved`, `malformed_genesis_policy_fails_closed` and `unverified_genesis_policy_fails_closed`. The remaining 17 are identical to `main`, compared by diffing the two complete sorted failure lists rather than by eye. No new failures.

The node key survives the run with its hash unchanged.

## What this does not fix

Two writes into the real home remain, and I would rather name them than let them look solved.

The **owner keystore** is still rewritten, with a fresh `owner_id` and both key pairs, six of its fifteen fields. `default_keystore_path()` resolves through `dirs::home_dir()` in `mesh-llm-identity` with no test override, and a `#[cfg(test)]` hook in that crate would not be active while host-runtime's tests run. Three test files call it: `api/tests/mod.rs`, `api/tests/apply_config_diagnostics.rs` and `api/tests/control_plane.rs`. Fixing it means giving those tests an explicit path rather than the default, which is a larger and separate change.

The **skippy hash cache** still lands in `~/.mesh-llm/cache/hashes`, because tests leave `MESH_LLM_HASH_CACHE_DIR` unset. `hash_cache.rs` documents that variable as the first step of its precedence, so this one looks cheap, but it is not what this branch is about.

## Gates

- `cargo test -p mesh-llm-host-runtime`: 3457 passed, 17 failed, as above.
- `cargo clippy -p mesh-llm-host-runtime --all-targets`: 8 lints, and the complete sorted set is identical to `main`.
- `scripts/check-env-mutation-contract.py` and `python -m unittest scripts.tests.test_env_mutation_contract`: passed. The census ratchet moves from 219 to 234 sites, three per guard, each with its own SAFETY comment.

## Note on shape

The five guards are near-duplicates of each other and of the one in `public_identity_tests.rs`. I deliberately did not extract a shared helper: the crate has no test-support module for it, and keeping the change mechanical makes it provably behaviour-preserving on Unix. Happy to extract one if you would rather have it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test isolation when running on Windows by consistently managing the test home-directory environment.
  - Ensured environment settings are restored after tests complete, preventing cross-test contamination.

- **Tests**
  - Updated environment-mutation checks and repository census expectations to reflect the expanded test safeguards.


ndizazzo (approved):
thank you!

PR #1850, opened 2026-09-13
test(host-runtime): write test keystores to the temp dir, not the real one

Virgile-pct:
Follow-up to #1847, which stopped this suite from deleting the node key and measured the owner keystore as still outstanding. This closes that one.

## What happens today

Six tests write their keystore fixture to `default_keystore_path()`:

```rust
let temp = tempfile::tempdir().unwrap();
let _home_guard = HomeEnvGuard::set(temp.path());
let owner = OwnerKeypair::generate();
let keystore_path = default_keystore_path().unwrap();   // the real home
save_keystore(&keystore_path, &owner, None, true).unwrap();
```

`default_keystore_path()` resolves through `dirs::home_dir()` in `mesh-llm-identity` and has no test override. The `#[cfg(test)]` hook that protects the node key lives in host-runtime, so it cannot reach a path resolved inside another crate, and `HomeEnvGuard` has no effect on it.

Measured on Windows 11 against `3f4f1c35a`, one full `cargo test -p mesh-llm-host-runtime`, comparing `~/.mesh-llm/owner-keystore.json` before and after: same size, same fifteen fields, **six of them different**. `owner_id`, `created_at`, and both the signing and encryption key pairs. The developer's owner identity is replaced by running the suite.

## What this changes

Each of the six call sites already had a `tempfile::tempdir()` on the line above, and each already handed the path to `state.set_owner_key_path()`. The default location was only ever somewhere to put the fixture; the code under test reads whatever path it is given. They now write into that temp directory.

Six lines, plus one import that became unused.

## Measured effect

Same box, same command, comparing `~/.mesh-llm` before and after:

| | #1847 | this branch |
|---|---|---|
| node `key` | unchanged | unchanged |
| `owner-keystore.json` | **rewritten** | unchanged |
| passed / failed | 3457 / 17 | 3457 / 17 |

The 17 failures are identical to the parent commit, compared by diffing both complete sorted lists rather than by eye. No new failures, and none of the six tests changed state: they passed before and still pass, they were simply passing while overwriting the developer's identity.

## What still remains

The skippy hash cache still writes into `~/.mesh-llm/cache/hashes`, six files per run, because tests leave `MESH_LLM_HASH_CACHE_DIR` unset and `hash_cache.rs` falls back to the home. That is a cache rather than identity material, so I left it out of this branch rather than mixing a third concern in. Happy to take it next if you want the suite to touch nothing at all.

## Gates

- `cargo test -p mesh-llm-host-runtime`: 3457 passed, 17 failed, as above.
- `cargo clippy -p mesh-llm-host-runtime --all-targets`: 8 lints, and the complete sorted set is identical to `main`.
- `scripts/check-env-mutation-contract.py` and `python -m unittest scripts.tests.test_env_mutation_contract`: passed, census unchanged at 234 sites.

## Summary by CodeRabbit

* **Tests**
  * Updated API tests to use temporary keystore locations consistently.
  * Improved test isolation for configuration diagnostics and control-plane scenarios.


i386 (approved):
Reviewed the synchronized exact head. Windows keystore tests use isolated temporary paths without changing production behavior; the full host-runtime package suite passes locally.

PR #1851, opened 2026-09-13
refactor(runtime): retire runtime console prints for the writer, tracing, and typed events

ndizazzo:
Runtime progress now behaves like the rest of the CLI: it disappears in JSON mode, it stays out of the dashboard's frame, and it carries a level when it lands in the log.

Before this change, 127 places in the runtime and its libraries wrote straight to stdout/stderr. Those lines painted over the interactive dashboard, leaked into `--log-format json` output, and arrived in the log with no level and a bare `stdout` context. They now route to one of three destinations, chosen by what the line actually is:

- **Console output the user asked for** — model maintenance, catalog downloads, layer-package progress, key rotation, model resolution and search, the interactive prompt, the skippy-server startup banners, `mesh-llm update`, and the benchmark prompt importer write through the sink-aware console writer. Text is byte-identical; it is now suppressed in JSON mode and while the dashboard owns the terminal.
- **Diagnostics** — binary-transport lifecycle, lane handshakes, native serving plugin dispatch, and native runtime discovery log through `tracing`. Failures at warn, lifecycle milestones at info, per-request chatter at debug, all filterable with `RUST_LOG`.
- **Progress consumers care about** — mesh publishing and auto-update progress emit typed `nostr_publishing` and `auto_update` events, so the dashboard, the JSON log, and future consumers get a name and a level instead of an unattributed line.

`skippy-server example-config` writes through the machine-output handle so its JSON always reaches stdout regardless of sink state. The client model list writes to stderr directly, keeping the embedded client free of the events dependency.

With this the console-print ratchet drops from **133 approved occurrences to 6** — only the two terminal-progress writers remain, which legitimately own the cursor.

## Architecture

Three details worth knowing for follow-up work:

- **A one-shot command has no event sink.** Only `serve` and `client` install one, so `emit_event` silently drops in `mesh-llm update`. That command uses the console writer while the long-running auto-update path uses events — the split inside `autoupdate.rs` is deliberate, not an oversight.
- **`EnvFilter::from_default_env()` defaults to ERROR.** A `tracing::warn!` on an unlisted target is dropped before the dashboard's writer sees it, so `runtime_tracing_subscriber` gains directives for `skippy_server`, `mesh_native_serving_plugin_host`, and `mesh_llm_runtime_install`. Any future crate that logs to the dashboard needs the same.
- `run_proposal` in the native serving plugin host crossed the cognitive-complexity limit once its stderr writes became `tracing` calls; its pre-dispatch deadline and fence checks moved into `proposal_predispatch_gate`. Control flow only — behavior is unchanged.

## Validation

- `cargo check` and `cargo clippy --all-targets -- -D warnings` clean across all seven touched crates
- `cargo fmt --all --check` clean
- `just no-console-print` passes; allowlist regenerated 133 → 6
- `cargo test --lib` on `mesh-llm-events` (132), `mesh-llm-host-runtime` (3509), `mesh-llm-system` (221), `skippy-server` (724) — all pass

Nine `skippy-server` tests fail when it is tested in the *same* `cargo test` invocation as the other three crates (`MeshLLM native runtime library has not been loaded`). Reproduced identically on the stashed base, so it is a pre-existing feature-unification artifact, not a regression from this branch.

Stacked on #1848.

## Summary by CodeRabbit

- **New Features**
  - Added status events for automatic updates and network publishing activity.
  - Automatic update messages can include the available version.
  - These events are now included in JSON-formatted output.

- **Improvements**
  - Progress, maintenance, model discovery, and download messages now use consistent application-managed output.
  - Startup banners and example configuration output use managed output streams.
  - Diagnostic and server messages use structured logging with clearer severity levels.
  - Reduced noisy runtime logs by limiting selected components to warnings and above.


i386 (dismissed):
Reviewed the synchronized exact head and its parent integration. Runtime event changes preserve the typed output boundary and compose cleanly with the console-writer layer.

i386 (approved):

i386 (dismissed):

PR #1854, opened 2026-09-13
fix(model-hf): reject rooted POSIX paths in delete on Windows too

Virgile-pct:
On Windows, `mesh-llm models delete /tmp/model.gguf` answers with a message about model stems and Hugging Face refs instead of saying that filesystem paths are not supported.

## Why

`parse_delete_model_ref` rejects paths with `is_absolute()` plus a few literal shapes. Windows does not consider a leading `/` absolute, since that needs a drive letter or a UNC prefix, so a rooted POSIX path passes the guard and falls into the Hugging Face branch.

Measured on Windows 11 by calling `resolve_model_identifier` directly:

| input | `is_absolute()` | outcome |
|---|---|---|
| `/tmp/model.gguf` | false | **falls through**, "Expected a model stem or Hugging Face ref..." |
| `/home/user/model.gguf` | false | **falls through**, same |
| `C:\models\model.gguf` | true | rejected as a path |
| `C:/models/model.gguf` | true | rejected as a path |
| `\\server\share\model.gguf` | true | rejected as a path |
| `./model.gguf` | false | rejected as a path |
| `~/model.gguf` | false | rejected as a path |

Wrong answer rather than a dangerous one, and only on Windows.

## What this changes

A leading `/` is rejected outright. No model stem or Hugging Face ref starts with one on any platform, and on Unix `is_absolute()` already covered it, so behaviour there is unchanged.

The test grew from one input to the seven above, minus `C:/models/model.gguf`, which is deliberately absent: Linux has no reason to treat it as a path, so asserting it would pass on my box and fail in CI.

## Measured effect, including a number that needs explaining

Full suite on Windows 11 against `main` at `3f4f1c35a`:

| | `main` | this branch |
|---|---|---|
| passed | 3457 | 3456 |
| failed | 17 | 18 |

The count went up, so here is the whole of it. One test is fixed, `models::delete_tests::resolve_model_identifier_rejects_filesystem_paths`, which is the one this branch targets. Two others failed that do not fail on `main`:

- `runtime::config_state::tests::sync::config_sync_state_apply_preserves_additive_defaults_sections`
- `runtime::config_state::tests::sync::config_sync_state_apply_preserves_nested_sections_and_updates_hash`

Both pass 5 runs out of 5 in isolation. `runtime::config_state` never calls `resolve_model_identifier` or `parse_delete_model_ref`, and this branch touches one guard in `model-hf` plus its test. This box produces one to three intermittent failures per full run, varying between runs, and today alone it surfaced three different ones across unrelated modules.

I would rather show you the raw number with that explanation than a tidier one.

## Gates

- `cargo test -p model-hf`: 47 passed, 0 failed.
- `cargo test -p mesh-llm-host-runtime`: as above.
- `cargo clippy -p model-hf -p mesh-llm-host-runtime --all-targets`: 8 lints, complete sorted set identical to `main`.
- `scripts/check-env-mutation-contract.py` and `python -m unittest scripts.tests.test_env_mutation_contract`: passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model reference validation to consistently reject filesystem paths across platforms.
  * Added coverage for Unix, Windows, network, relative, and home-directory path formats.
  * Error messages now identify the invalid path that was provided.


i386 (approved):
Reviewed the synchronized exact head. The Windows deletion guard covers leading-slash rooted paths and preserves valid relative paths; model-hf and full host-runtime package suites pass locally.

PR #1866, opened 2026-09-14
fix(models): stop drawing download progress into a redirected stderr

michaelneale:
## Problem

`MeshDownloadProgress::draw` (`crates/mesh-llm-host-runtime/src/models/catalog.rs`) writes a carriage-return progress frame with ANSI escapes to stderr every 150 ms. It declines only when an interactive TUI is active.

A supervised runtime — a tray app, launchd, a container — redirects stderr to a log file while still using pretty logs. Nothing overwrites a line in a file, so every frame is appended.

Observed: one 5.7 GB model download produced a **58 MB** `mesh.log` that is almost entirely progress-bar frames, and unreadable as a log.

## Fix

Require both a terminal and non-JSON logs before drawing — the only case where `\r` does what it is meant to do. Every other progress writer in the tree already declines in JSON mode (`mesh-llm-events/src/terminal_progress.rs:22,54,102`); this one checked neither JSON mode nor terminal-ness.

The predicate is a named function so the four cases are unit-tested rather than depending on a global sink and an inherited fd.

## Scope of evidence

- `cargo test -p mesh-llm-host-runtime` passes at `0c289ff33` (branched off `origin/main` `c573119ca`), including the new case.
- `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings` clean; `cargo fmt --all` clean.
- Not separately re-run end to end after the fix: the 58 MB log is from the pre-fix binary. The claim here is about which branch is taken, which the unit test covers; I have not re-downloaded 5.7 GB to watch a file stay small.
- Interactive terminal behaviour is unchanged by inspection only — `interactive_tui_active()` still takes its existing branch, and the terminal case still satisfies the new predicate.

Found while a tray app supervised a serving runtime with `--log-format json`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Download progress output is now hidden when JSON mode is enabled or when output is redirected.
  * Interactive terminal progress displays continue to show updates as expected.
  * Progress output now behaves consistently across supported output modes, avoiding unwanted formatting in machine-readable or redirected output.


ndizazzo:
@michaelneale writing to STDERR was deliberate so we weren't polluting stdout for formatted logs (JSON) or the TUI... I haven't dove into this fix yet, but IIUC, we are now using stdout with this again, so stderr can exclusively be used for errors?

ndizazzo:
Looks like this clashes with #1858

@coderabbitai which is the most canonical approach for CLI tools that have TUI elements?

ndizazzo:
@michaelneale I think we should encode and adopt https://github.com/Mesh-LLM/mesh-llm/pull/1866#issuecomment-5674189788

ndizazzo (changes requested):
See above comments

PR #1869, opened 2026-09-14
chore(deps): remove unused dependencies flagged by cargo-machete

iamthesvn:
## Original problem

19 dependencies across 8 workspace crates sit in Cargo.toml files without a single reference in code. They still get compiled, audited, and bumped on every dependency update. Some are crypto and keychain crates (`argon2`, `keyring`, `crypto_box`, `chacha20poly1305`), which widen the shipped binary's audit and supply-chain surface while doing nothing. The keystore crypto is owned and used by `mesh-llm-identity` (`src/keystore.rs`, `src/keychain.rs`); host-runtime's copies were never referenced. `mesh-llm-ffi` also has a dead `host` feature that only exists to enable an optional `mesh-llm-node` dependency no code uses.

## Diagnostics

I ran `cargo machete` on the workspace, then grep-checked every flag against the source tree before touching anything: cfg-gated code, build scripts, tests, examples. One flag was wrong. `serde_json` in the web-ui exemplar plugin is used in `manifest.rs`, but `src/main.rs` pulls that file in through `#[path = "../manifest.rs"]`, and machete doesn't follow that include. That dependency stays.

## Fix

Removes the 19 unused dependencies:

- mesh-llm-tui: `arboard`
- mesh-client: `bytes`
- mesh-llm-log-store: `thiserror`
- skippy-server: `skippy-topology`
- mesh-llm-embedded-runtime: `anyhow`, `serde_json`
- mesh-llm-guardrails: `serde` (`serde_json` stays, it's used)
- mesh-llm-host-runtime: `argon2`, `chacha20poly1305`, `clap`, `crypto_box`, `flate2`, `keyring`, `mesh-llm-api-server`, `model-package`, `tabwriter`, `tar`, `zip`
- mesh-llm-ffi: `mesh-llm-node`, plus the `host` feature that only existed to turn it on

One compatibility note. The ffi `host` feature was still named in the SDK build lanes, so the same commit updates them (`sdk/kotlin/build.gradle.kts`, `sdk/swift/scripts/build-xcframework.sh`, `sdk/swift/scripts/build-host-macos-xcframework.sh`, `scripts/package-native-sdk.sh`) to pass `--features embedded-runtime` only. Nothing changes in what gets built: every `cfg(feature = ...)` in the ffi crate checks `embedded-runtime`, never `host`, and node functionality comes from mesh-llm-sdk's untouched features. Out-of-tree builds passing `--features host` get a cargo error naming the unknown feature — a loud failure, not a silent one. No protocol, gossip, or runtime behavior changes. `Cargo.lock` shrinks by 308 lines.

## Validation

No UI changes; no Rust source changed.

- `cargo check --workspace --all-targets` — clean (macOS, default features; other platforms and feature combos rest on the source audit, not the compiler)
- `cargo clippy --all-targets -- -D warnings` on every touched crate plus `-p mesh-llm` — clean. Full-workspace clippy fails on a pre-existing `clippy::chunks_exact_to_as_chunks` lint in skippy-model tests; I confirmed it fails the same way on unmodified `main`
- `cargo fmt --all --check`, `just no-console-print`, `just check-commits` — clean
- New SDK feature invocation resolves: `cargo tree -p mesh-llm-ffi --no-default-features --features embedded-runtime`. Edited shell scripts pass `just ci-shellcheck`
- `cargo machete` has nothing actionable left (the one remaining flag is the exemplar false positive above)

## Summary by CodeRabbit

- **Build and Packaging**
  - Simplified native SDK builds to use only the embedded runtime configuration across Android, macOS, and Swift targets.
  - Removed the host runtime option from FFI builds.

- **Dependency Maintenance**
  - Removed unused internal and third-party dependencies across runtime, guardrails, logging, UI, server, and client components.
  - Streamlined host runtime dependencies without changing exported APIs.


i386:
@iamthesvn thanks for this - looks good to me once CI passes 👍

i386 (approved):

PR #1874, opened 2026-09-15
fix(console): answer chat questions instead of thinking about them

michaelneale:
Ask a reasoning model a casual question in the web console today and it deliberates for tens of seconds before answering — and on a 9B at Q4 it often never answers at all. The reasoning consumes the whole token budget, so `finish_reason` is `length` and `content` is null: the chat window shows a growing thinking trace and then nothing.

After this change the console asks for a direct answer.

Measured on a live node, `unsloth/Qwen3.5-9B-GGUF:Q4_K_M`, prompt "what is updog", `max_tokens: 4000`:

| Request | Result |
|---|---|
| `chat_template_kwargs: {enable_thinking: false}` | 350 tokens, **4.6 s**, answered |
| `reasoning_effort: "none"` | 378 tokens, answered |
| `reasoning: {enabled: false}` | 453 tokens, answered |
| nothing set (today's console) | 4000 tokens, 38 s+, 13.5 KB reasoning, **no answer** |

The deliberation also degenerates rather than converging — one captured run repeated `Wait, could it be …` 39 times before terminating.

This restores a safeguard that used to exist: `00296cc3a` (Mar) added `chat_template_kwargs: {enable_thinking: false}` to console chat requests for this exact symptom (Qwen3.5-9B 18 s → 0.6 s), and #545 dropped it when the request builder was rewritten.

## Protocol

No API change. `reasoning_effort`, `reasoning.enabled` and `chat_template_kwargs` all work today and are unaffected — verified live, table above. Only the console's own requests carry the new field, and the field is optional in `ResponsesRequest`.

## Validation

`pnpm run lint`, `typecheck`, `test` (208 files, 1751 passed), `build` — all clean. Adds a test asserting the console request carries the field.

Not covered: no screenshot, because the change removes UI rather than adding it — the thinking accordion simply stops appearing for models that only reasoned because nobody said otherwise. The user-visible before/after is the timing table above, taken from the API rather than the browser.

## Summary by CodeRabbit

- **Behavior**
  - Chat requests now request direct answers without hidden reasoning by default.
  - Usage details continue to be included in streaming responses.

- **Reliability**
  - Added coverage to verify the direct-answer request behavior.


michaelneale:
Closing at Mic’s request: disabling console thinking is a workaround, not the requested solution. Preserve visible reasoning and investigate the repetitive-generation failure. No branch deletion requested.

PR #1879, opened 2026-09-15
fix(console): preserve launched model names and remove default briefing

michaelneale:
## Summary

Console chats start without a baked-in MeshLLM product briefing. Custom system prompts remain unchanged.

Models launched from HF retain their launch coordinates in source metadata; local GGUF launches retain their filename without publishing parent directories. Console selectors and model details use API display names rather than content-hash routing IDs.

Example launch:
```sh
mesh-llm serve --model 'unsloth/gemma-4-26B-A4B-it-GGUF@main:UD-Q4_K_M'
```
Expected presentation: `unsloth/gemma-4-26B-A4B-it-GGUF@main:UD-Q4_K_M`; the synthetic `local-gguf/sha256-…` identity remains available for routing and split verification.

## Architecture

Carry original source through startup separately from split identity. Populate existing served descriptors for primary and secondary models, preserve source identities through serving-list refresh, and derive synthetic-model display labels from descriptors before falling back to inventory. UI selections continue sending the original routing key.

The historical gap is between #969's inventory-keyed labels and #1672's content-keyed startup IDs; `95938a206` extended preindexing to HF-cached GGUFs. The UI adapter also previously discarded display_name.

## Protocol

No wire fields or ABI changes. Existing descriptor source fields are used; content hashes and split admission remain unchanged. Live mixed-version/multi-node validation is still pending—this draft is not merge-ready on unit-test evidence alone.

## Validation

On the working diff based on d29d25120, committed as this branch:
- `just build`: passed (host, UI, adjacent Metal native runtime).
- Host-runtime full suite: 3,526 unit tests passed, 11 ignored; 24 integration tests passed; doc tests passed.
- `cargo check` for host runtime and shipped binary: passed.
- `cargo clippy -p mesh-llm-host-runtime -p mesh-llm --all-targets -- -D warnings`: passed.
- `cargo fmt --all --check`, `just no-console-print`, diff checks: passed.
- UI typecheck, lint/format: passed; full UI suite: 209 files, 1,762 passed / 3 skipped.
- Tests cover empty/custom prompts, readable label/original request ID, source privacy, primary/secondary descriptor refresh, and unchanged identical-file split IDs.
- Live single-node check at 748de8863: built with `just build` and `VITE_MESH_LLM_DEBUG_UI=false`; served cached `Qwen/Qwen3-0.6B-GGUF@main:Q8_0`. Fresh-browser console selector displays that launch source and Warm status. Real chat completion returned "Hello! How can I assist you today?" (thinking disabled for the API probe). Multi-node verification remains pending.

## Screenshot

Real embedded console with the model selector open, serving locally from the PR binary at 748de8863. No fixture data.

Live console showing original HF launch source and Warm model

## Summary by CodeRabbit

- **New Features**
  - Model lists, selectors, catalogs, and detail views now show readable display names when available while preserving routing identifiers.
  - Startup model sources and local file names are surfaced more consistently.
  - Model identity and source details remain stable when serving models are refreshed or updated.

- **Changes**
  - The built-in default chat system prompt has been removed. Chats now use only custom instructions when provided.
  - Remote model references displayed to the mesh omit embedded credentials, query parameters, and fragments.


i386 (commented):
The current patch solves the live symptom, but I think its scope and identity flow can be made easier to review.

## Scope

Please split the default-system-prompt removal into a separate PR. It changes chat behavior independently of model presentation and accounts for several UI files and tests. The model-name PR should answer one question: how do we preserve the launch source for display while retaining the content-addressed routing key?

## Smaller identity shape

Keep one invariant through the stack:

- `name` is the stable routing identity and is sent back unchanged by clients.
- `display_name` is presentation metadata derived from the original launch source.

The minimum necessary data flow appears to be:

1. Preserve the original HF/package selector or local filename on `StartupModelPlan` before preindexing replaces the declared identity with `local-gguf/...`.
2. Attach that source metadata to the served descriptor during startup advertisement.
3. Derive the public `display_name` in the existing runtime-data projection.
4. Have the UI adapter display `display_name` while retaining `name` as the selected request value.

Please see whether the new `model_presentation` and `model_labels` modules can be folded into the existing startup-advertisement and runtime-data projection paths. In particular, the extra `set_serving_models` → `upsert_served_model_descriptor` → later refresh sequence creates a second state-reconciliation path. A single descriptor update path would be easier to reason about and less likely to regress.

The cross-layer source propagation itself is necessary: after startup converts the model to a content-addressed ID, the UI cannot recover the original HF selector on its own.

## Required behavior contracts

Please preserve these in tests before merge:

- An HF launch displays the exact launch selector, including revision and quant selector, while API requests still use the routing ID.
- An explicitly requested package displays its package coordinates.
- A local GGUF displays only its filename; parent directories never reach the API/UI.
- Background descriptor refresh never downgrades known source metadata to `Unknown` or replaces it with a synthetic hash label.
- Primary and secondary models retain their own source identities.
- Aliases remain aliases; the preservation condition must not depend only on a `local-gguf/` prefix. This is the valid edge case CodeRabbit identified.
- Content hashes and split admission remain unchanged.

I would make these durable product invariants in a root product contract, reference them from path-local `AGENTS.md` files for runtime and UI work, and enforce them with API-boundary tests. Documentation guides agents; executable contract tests stop accidental behavior changes.

i386 (changes requested):
The current exact head is code-clean in this pass, including the two review fixes: remote HTTP(S) launch sources now strip credentials/query/fragment before gossip, and aliased local GGUF names retain their `LocalGguf` identity and filename across refresh.

Validation on `98ec2773c920ef70ea5f3f5c00ffdc9a4728629e`:
- full `mesh-llm-host-runtime` suite: 3,527 unit tests passed, 11 ignored; all integration and doc tests passed
- `just ci-validate`: 1,493 passed, 9 skipped
- UI validate: 209 files, 1,762 passed, 3 skipped; lint, typecheck, formatting, and production build passed
- `just build`: host, UI, patched Metal runtime, and composed local product passed

I am holding approval on the PR's own stated merge gate: mixed-version, multi-node live validation is still pending. Please attach the run evidence demonstrating that a PR node and a current-main node exchange these descriptors without changing routing identity or leaking source metadata; I will re-review that evidence and approve if it passes.

michaelneale:
Checked the outside-diff CodeRabbit routing-source finding against exact head 98ec2773c920ef70ea5f3f5c00ffdc9a4728629e. I cannot substantiate it in the cited startup path:

- runtime/auto_join.rs:1279: run_auto_model_identity returns (model_name, model_source); model_name comes from StartupModelPlan.declared_ref, while model_source comes from the separate source field.
- Both callers (runtime/control_loop.rs:252 and runtime/run_auto.rs:1466) pass those values in that order to advertise_run_auto_models.
- runtime/run_auto.rs:1133 passes model_name, not model_source, into build_serving_list. The source is passed separately to set_model_source.
- runtime/serving_surface.rs:1829 builds the serving list from declared_ref values and that model_name.

No code change is warranted for this specific finding based on this trace. This is source inspection, not new execution evidence. Existing helper tests in runtime/tests/startup_models.rs:847–891 include preservation of the synthetic routing ID; a startup-to-advertisement test with differing source and routing strings would strengthen boundary coverage.

The mixed-version live evidence requested by James remains outstanding; this comment does not substitute for it.

michaelneale:
🤖 *Posted by Galadriel (agent) on behalf of Mic.*

## Mixed-version two-node live validation — PASS (one finding to note)

Per the merge gate: PR node and current-main node exchanging descriptors, verifying routing identity is unchanged and no source metadata leaks.

### Setup

| Node | Build | Version string | Node ID | Model (launched by HF name) | Ports |
|---|---|---|---|---|---|
| A | PR head `98ec2773c` | `0.76.1+g98EC27` | `75355ae365` | `Qwen/Qwen2.5-0.5B-Instruct-GGUF@main:q4_k_m` | 9337 / console 3131 |
| B | main `60b7f5ff9` | `0.76.1+g60B7F5` | `49ecf8c441` | `bartowski/Llama-3.2-1B-Instruct-GGUF@main:Q4_K_M` | 9447 / console 3145 |

Private mesh via invite token (`mesh_id 08bb01ea…`), same Mac (M4 Max, Metal), **distinct models on each node** so a successful cross-node completion cannot be served locally. Both consoles report `peers: 1`.

### 1. Routing identity unchanged across the version boundary ✅

A's gossiped serving identity for the HF-launched model is the content hash, byte-for-byte identical in B's peer view:

- A `/api/status`: `serving_models = ["local-gguf/sha256-74a4da8c9fdbcd15bd1f6d01d621410d31c6fc00986f5eb687824e7b93d7a9db"]`
- B `/api/status` peer `75355ae365`: `serving_models`, `hosted_models`, `requested_models`, and `advertised_model_throughput.model_name` all carry the same `local-gguf/sha256-74a4…` string. No field was renamed or replaced by the display name.

### 2. Cross-node inference on the routing ID, both directions ✅

- **old → new**: completion to B (`:9447`) with `"model": "local-gguf/sha256-74a4…"` (A's model, which B does not host) → HTTP 200, content "OK". A's request log shows the corresponding admitted+completed chat_completions for that model.
- **new → old**: completion to A (`:9337`) with `"model": "local-gguf/sha256-6f85…"` (B's model, which A does not host) → HTTP 200, content "OK".

Distinct models per node rule out local serving as a confound.

### 3. Display metadata is display-only, and degrades gracefully on the old node ✅

- A (PR) `/v1/models`: `id` = friendly ref `Qwen/Qwen2.5-0.5B-Instruct-GGUF:q4_k_m` for its own model; that alias routes on A (200) while the hash also still routes on A (200).
- B (main) shows A's model in `/api/models` as `name: Qwen/Qwen2.5-0.5B-Instruct-GGUF@main:q4_k_m`, `display_name: Qwen2.5-0.5B-Instruct-Q4_K_M`, `node_count: 1` — i.e. the old node consumed the new descriptor's identity fields through its **pre-existing** `canonical_ref` path (`runtime_data/collector.rs`) without error. Old node's own model remains hash-labelled everywhere, as expected.
- B never invents a wrong label; worst case is the hash, which is the pre-PR status quo.

### 4. No source-metadata leak ✅

Grepped `/api/status`, `/api/models`, `/v1/models` on both nodes for absolute paths (`/Users/…`), credentials, query strings, and fragments in gossiped/derived fields: none present. (The only `/Users/…` strings anywhere are each node's *own local* runtime-dir paths in its own status — local process metadata, not gossiped peer data.)

### 5. CodeRabbit `build_serving_list` hypothesis — not reproduced ✅

If `model_source` had entered the serving list as a routing candidate, A would advertise the HF ref in `serving_models` and/or the hash would stop routing. Neither happened: gossip carries only the hash, and both hash-ID cross-node completions succeeded. Matches @michaelneale's source trace.

### Finding (non-blocking, worth a decision)

The old node lists A's canonical ref in its own `/v1/models` as `id: Qwen/Qwen2.5-0.5B-Instruct-GGUF@main:q4_k_m` — but a completion against B with that id returns `model_not_found` (B's router only matches A's advertised hash). So a mixed-version mesh's old nodes can list an id they cannot route. This is old-node code (catalog entry keyed on `canonical_ref`, pre-existing) reacting to the richer descriptor the PR now sends; the PR node itself only advertises routable ids. Impact: an API client picking models from an old node's list may select a dead entry until that node upgrades. Not a routing-identity change and not a leak — flagging for an explicit accept/defer call.

### Environment notes

- Debug builds (`just build`) of both trees, Metal, same host; B given a distinct node key (`MESH_LLM_EPHEMERAL_KEY=1`) since both nodes share the machine.
- Logs and full JSON captures retained locally (`.scratch/nodeA.log`, `.scratch/nodeB.log`) if anyone wants the raw artifacts.

i386 (approved):
Approved at `98ec2773c920ef70ea5f3f5c00ffdc9a4728629e`.

The mixed-version two-node evidence satisfies the remaining merge gate: the content-hash routing identity stayed byte-for-byte stable across PR/current-main gossip, cross-node inference succeeded in both directions using those routing IDs, and no gossiped source path, credential, query, or fragment leaked.

I also rechecked CodeRabbit's outside-diff claim. `run_auto_model_identity` returns `declared_ref` as `model_name` and keeps `model_source` separate; both callers preserve that ordering, and `advertise_run_auto_models` passes only `model_name` to `build_serving_list`. The claim is not actionable on this head.

Accepted compatibility limitation: an old node may expose a new peer's canonical ref in `/v1/models` while its old router still requires the advertised hash. The advertised identity remains correct and routable, and upgraded nodes route the friendly alias. This is worth follow-up for mixed-version model-list round-tripping, but it does not invalidate this fix.

PR #1882, opened 2026-09-15
test(host-runtime): keep the digest cache out of the real home under test

Virgile-pct:
The host-runtime test suite writes digest records into the developer's real `~/.mesh-llm/cache/hashes`, keyed to temp fixtures that no longer exist by the time the run ends. Same lineage as #1847 and #1850, which both listed this one as still open.

## What happens today

`SidecarDigestCache::open_default` falls back to `~/.mesh-llm/cache/hashes` when `MESH_LLM_HASH_CACHE_DIR` is unset, and nothing sets it under test: not the tests, not CI, not the justfile. Package verification reaches `open_default` from four production sites in `package.rs`, so every test that verifies a package-v2 layout or a checkpoint directory writes into the real cache.

Measured on Windows 11, one full `cargo test -p mesh-llm-host-runtime` on `main` at 950e90552, with a SHA-256 listing of `~/.mesh-llm` taken before and after:

```
NOUVEAU  ./cache/hashes/27afea3e...json
NOUVEAU  ./cache/hashes/38aaaa85...json
NOUVEAU  ./cache/hashes/3f26f6de...json
NOUVEAU  ./cache/hashes/4b8ada82...json
NOUVEAU  ./cache/hashes/512f3719...json
NOUVEAU  ./cache/hashes/a6e81868...json
```

Every record points at a fixture that is already gone:

```
...\Temp\.tmplRSWnW\shared\payload.gguf
...\Temp\.tmplRSWnW\shared\metadata.gguf
...\Temp\.tmp5MOrB3\model.safetensors
...\Temp\.tmp5MOrB3\config.json
```

They can never be looked up again, so the real cache grows by a handful of dead records per run. Nothing in the suite asserts a cache hit through the default path, and the two tests that exercise caching open an explicit directory through `open_in`.

## What this changes

Under `cfg(test)` the default location is `None`, which is the third case the function already documents: "caching disabled, digests are always recomputed". `MESH_LLM_HASH_CACHE_DIR` still redirects the default for a test that wants it. The production path is untouched.

No environment mutation is added, so the env-mutation contract stays at 234 sites and no `SAFETY:` comment is needed. This is the same shape as `identity_home_dir` in `mesh/identity_persistence.rs`, which already takes a `cfg(test)` branch for the same reason.

## Measured effect

The package, hash-cache, certification and materialization tests rerun with the same before-and-after listing:

```
test result: ok. 80 passed; 0 failed; 3 ignored
~/.mesh-llm: byte-identical before and after
```

The new test, `the_default_cache_never_reaches_the_real_home_under_test`, asserts `open_default()` is `None` when the variable is unset. The full suite on this box is unchanged otherwise: 16 failures on `main` at 950e90552, none of them in this module, and I have a separate write-up coming for those.

Rebased onto `main` at 60b7f5ff9 and rerun with the same before-and-after listing: 80 passed, 0 failed, `~/.mesh-llm` byte-identical again.

## Gates

- `cargo test -p mesh-llm-host-runtime` (package, hash_cache, certification, materialization filters): 80 passed, 0 failed.
- `cargo clippy -p mesh-llm-host-runtime --all-targets`: complete sorted set identical to `main`.
- `scripts/check-env-mutation-contract.py` and `python -m unittest scripts.tests.test_env_mutation_contract`: passed, 234 sites.

## Not verified

`cfg(test)` covers this crate's own harness, which is where the writes above come from. A downstream crate's tests compile this crate without it and keep the environment variable as their lever; I have not measured whether any of them reach `open_default`. Nothing run on Linux or macOS.

## Summary by CodeRabbit

* **Chores**
  * Hash data is now stored in a standard application cache directory under the user’s home folder.
  * Test environments no longer access the real home-directory cache by default, improving isolation and preventing unintended local cache changes.
  * The cache location can still be explicitly configured when needed.


i386 (approved):
Reviewed the diff. Clean and correct: `default_dir()` split into `#[cfg(not(test))`/`#[cfg(test)]` so the default location is `None` under test, while `MESH_LLM_HASH_CACHE_DIR` still overrides — matches the documented third case and the precedent in `identity_home_dir`. The regression test asserts `open_default().is_none()` under the same condition. No production path changes, no env mutation. Full CI green. Approving.

PR #1888, opened 2026-09-15
test(host-runtime): render fixture paths through TOML at six more sites

Virgile-pct:
Six host-runtime tests build a `MeshConfig` by interpolating a temp path straight into a TOML basic string. On Windows that path is `C:\Users\...`, and TOML reads `\U` as the start of a unicode escape:

```
Error { message: "too few unicode value digits, expected unicode hexadecimal value" }
```

Same defect #1834 fixed in `mesh-llm-commands`, in the same shape.

## The six sites

| test | site |
|---|---|
| `resolver::exact_head_tests::canonical_selector_wins_over_another_models_derived_profile` | `resolver/exact_head_tests.rs:50` |
| `resolver::exact_head_tests::served_alias_and_resolved_path_do_not_select_model_specific_config` | same fixture |
| `resolver::native_mtp_tests::speculative_strategy_auto_uses_hardware_model_path_for_direct_gguf_detection` | `resolver/native_mtp_tests.rs:389` |
| `resolver::native_mtp_tests::speculative_strategy_native_mtp_accepts_external_mtp_sidecar` | `resolver/native_mtp_tests.rs:1030` |
| `local_split::test_support::runtime_resolver_uses_config_identity_and_honors_device_override` | `local_split/test_support.rs:693` and `:704`, one template |
| `local_split::tests::content_attestation::strict_multimodal_stage_builder_keeps_local_paths_off_downstream_loads` | `local_split/tests/content_attestation.rs:85` |

The `local_split` template carries two `[[models]]` entries that share a single `{model_path}` argument, so both placeholders lose their quotes together. My first pass fixed only the second one and the test kept failing with the rendered literal string wrapped in a second pair of quotes, which is how the seventh site turned up.

## What this changes

Each site renders the path through `toml::Value::String(...).to_string()`, which quotes and escapes for TOML, and the fixture drops its own quotes around the placeholder. Nothing new is introduced: `resolver/test_support.rs` already has `toml_path` doing exactly that, used eleven times by the speculative tests, and `local_split/test_support.rs` already writes the same expression three times a few lines below one of the sites that bypassed it. The one addition is an explicit import in `exact_head_tests.rs`, since `test_support` is a private module with no glob re-export.

On Linux and macOS the rendered value is byte-identical to what the fixtures produced before, since a path without backslashes or quotes renders as the same basic string.

## Measured effect

Windows 11. On `main` at 950e90552 all six fail with the unicode-escape error above. On this branch, the six plus the seven other tests of `runtime::local_split::test_support`, run on the tree that `main` at 8ab35865c now has (85b142df0 plus #1885, which the run carried as a cherry-pick because `main` did not build in between):

```
test result: ok. 13 passed; 0 failed; 0 ignored
```

## Gates

- `cargo test -p mesh-llm-host-runtime` (the six tests and `runtime::local_split::test_support`): 13 passed, 0 failed.
- `cargo clippy -p mesh-llm-host-runtime --all-targets`: 8 locations, complete sorted set identical to `main` at 60b7f5ff9.
- `scripts/check-env-mutation-contract.py` and `python -m unittest scripts.tests.test_env_mutation_contract`: passed, 234 sites.

## Not verified

Nothing run on Linux or macOS. The rendered TOML is the same there, but I have not executed the tests on either.

## Summary by CodeRabbit

- **Tests**
  - Improved test configuration handling for model, projector, hardware model, and sidecar paths.
  - Added safer path serialization to preserve paths accurately across different filesystem formats.
  - Maintained coverage for model resolution, native MTP detection, external sidecar resolution, and multimodal configuration validation.
  - Updated test fixtures to use canonical path values, improving reliability across environments.


i386 (approved):
Reviewed all seven render sites (the six tests plus the shared `{model_path}` template used by two `[[models]]` entries). Every path now goes through `toml::Value::String(...)` (or the existing `toml_path` helper), which quotes and escapes for TOML — the same fix as #1834. Purely test code, byte-identical output on Unix since a plain path renders as the same basic string. CI green (the one red `Rust Clippy (${{ matrix.batch.id }})` entry is a stale unrendered matrix job with no logs, the real clippy batches 0-2 all pass). Approving.

PR #1889, opened 2026-09-16
ci: restore the trusted Linux host sccache seed for amd64 host rows

i386:
## Summary

The Linux host slice gated trusted sccache-seed restore on `matrix.host.architecture == 'x86_64'`, but the versioned CI planner (`scripts/plan-ci.py` over `ci/slices.yml`) emits `amd64` for every Linux host row. The guard could never be true, so the restore step was always skipped and every Linux host job compiled the dominant host dependency graph cold — ~2 sccache hits out of ~1,046 cacheable compilations — even though an exact-match seed entry existed on main.

One-line fix: match the guard to the planner architecture vocabulary (`x86_64` → `amd64`). Quality and Rust-test slices already pass the policy flag through unchanged and were unaffected; the Linux runtime slice keeps its deliberate documented opt-out (per-backend toolchain epochs, observed zero reuse).

## Before/after evidence

**Before (run 34946419327, `x-mesh-target` PR):**
- Restore step: `allow_trusted_seed: false` with exact key `mesh-llm-sccache-seed-linux-x86_64-img-f499b79b-epoch-f499b79b-v3-1cd80bb…` → `restore` step outcome **skipped**
- sccache-stats artifact `sccache-ci-host-linux-amd64-1`: 2 hits / 1,046 requests, hit rate 0.0019, classification `cold`
- Same for run 34822560408
- The warmer (run 35038921944 at main 281c85fdb) restored and holds that exact key — the seed exists and matches; it was simply never consumed

**After (this PR, CI on the branch):**
- The `Linux host (amd64)` job now evaluates `allow_trusted_seed: true` on GitHub-hosted placements and restores the seed before `configure-sccache-gha`; the capture step expects `warm` and enforces the 0.20 minimum hit rate, so the hosted PR run itself is the matched after-sample.

## Validation

- `actionlint -config-file .github/actionlint.yaml`: clean
- `python3 -m unittest discover -s scripts/tests -p "test_*.py"`: 1,499 tests; only failure is pre-existing local-env `git tag -a` EDITOR issue in `test_release_notes`, unrelated to workflows
- Targeted CI/sccache suites (`test_runtime_seed_canary`, `test_sccache_evidence`, `test_ci_lane_workflows`, `test_ci_runner_selection_and_cache_policy`, `test_depot_canary_workflow`, workflow contract tests): all OK

Fixes the cold Linux host-cache follow-up from the accelerator rollout (#1870).

## Summary by CodeRabbit

- **Chores**
  - Updated Linux build infrastructure to correctly identify the supported 64-bit host architecture when preparing trusted build-cache data.
  - No changes to application functionality or the end-user experience.


i386:
## Why this PR cannot self-demonstrate the fix

`pr_linux.yml` invokes the reusable lane as `Mesh-LLM/mesh-llm/.github/workflows/ci-linux-lane.yml@main`, so this PR run executed the workflow definition from **main**, which still contains the old `x86_64` guard. The `Linux host (amd64)` job in this run therefore shows `allow_trusted_seed: false` — it is a reproduction of the bug, not a regression of the fix (the merge ref itself contains the corrected `amd64` guard; verified against `merge_commit_sha` 4057a415).

## Verified mechanics (from this run + local policy-script reproduction)

- Runner policy for a direct PR on GitHub-hosted placement outputs `allow_trusted_sccache_seed=true` (reproduced locally with the exact observed inputs: `event_name=pull_request`, `ref=refs/pull/1889/merge`, `force_hosted=true`, `depot_*_enabled=true`).
- Hosts matrix emits `architecture: "amd64"` (`linux-amd64-host` row, visible in this runs job inputs).
- Exact seed key `…v3-1cd80bb…` exists on main (231 MB, ref `refs/heads/main`, restored by the warmer in run 35038921944).

After merge, the next PR Linux host run will resolve the lane from updated main with the corrected guard and restore the seed; its `sccache-ci-host-linux-amd64-*` stats artifact should classify `warm` (≥0.20 hit rate). I will verify the first post-merge PR run and report back on the rollout thread.

PR #1895, opened 2026-09-16
ci(quality): flag unused dependencies with cargo-machete

iamthesvn:
Follow-up to my previous PR #1869, which removed 19 unused dependencies by hand. This adds the check to CI so they get caught when introduced.

## Original problem

#1869 found 19 across 8 crates, including crypto and keychain crates (`argon2`, `keyring`, `crypto_box`, `chacha20poly1305`) that grow the shipped binary's audit and supply-chain surface while doing nothing. Nothing fails when an unused dependency is added (no compiler warning, no test failure), so they build up unnoticed.

## Diagnostics

cargo-machete parses manifests and scans source; it doesn't compile or execute repo code. I ran `cargo machete` on this tree: exit 0. Post-#1869 the workspace is clean; the one flag it raises is a known false positive (below).

## Fix

One new job in `ci-quality-slice.yml`, modeled on `rust_fmt`. Runner selection, depot isolation, and checkout pinning are copied from the sibling jobs: same trust posture.

The job gates on the planner's `rust_changed` signal plus a new `plugin_exemplars_changed` signal (`docs/plugins/exemplars/**`). The web-ui exemplar is a standalone Cargo project outside the workspace and outside the rust domain, so without that signal a manifest change there could bypass the check. The signal is computed in `scripts/plan-ci.py`, added to `ci/ci-plan.schema.json` and the `plan-ci` outputs, and rides the existing `lane_plan_json` plumbing into the slice, so no caller workflow changes. The lane pass-through is guarded (`|| false`) so a stale pre-merge plan can't feed null into the slice's required input.

Tool install via `taiki-e/install-action` (pinned SHA, same per-job pattern as `just` and `actionlint`), with the tool pinned to `cargo-machete@0.9.2` so a future machete release with different heuristics can't break unrelated PRs. `manage-ci` prefers image-baked tools; I followed the per-job precedent, and I'm happy to rework if you'd rather bake machete into the runner image.

The web-ui exemplar gets `[package.metadata.cargo-machete] ignored = ["serde_json"]`: it's used in `manifest.rs` via a `#[path]` include machete doesn't follow, and the gate can't stay green without it. Contract tests covering the job inventory (`test_ci_runner_selection_and_cache_policy`, `test_ci_artifact_actions`, `test_depot_authority_sentinel`, `test_depot_canary_workflow`), the planner signals (`test_plan_ci`), and the `ci.md` / `current-inventory.md` slice descriptions are updated in the same commit.

## Validation

No UI changes; no Rust source changed.

- `actionlint`: clean.
- Full `scripts/tests` suite under python3.11: 1509 tests. Failures are identical to unmodified `upstream/main` on this machine (29, all environmental: repo scripts spawn the system python3.9; a stale local `.deps/llama.cpp` trips the llama canary, which the branch passes in a clean worktree). The three contract failures from review (artifact actions, authority sentinel, depot canary audit count) are fixed, and `test_plan_ci.py` covers the new signal.
- `cargo machete` on this tree: exit 0.
- CI plans and executes from the default branch, so the new job still skips this PR; its first live run is the post-merge main push or the next rust/exemplar-touching PR. `cargo machete` exits 0 on this exact tree.

---

Buzz dev-channel post (do not include in the PR body):

I opened a follow up to my PR #1869: https://github.com/Mesh-LLM/mesh-llm/pull/1895 adds a cargo-machete job to the quality CI slice, so unused rust deps get flagged when they come in rather than piling up.

## Summary by CodeRabbit

* **New Features**
  * Added automated unused-dependency checks to Rust quality validation.
  * The check follows existing runner and cache policies and reports failures when unused dependencies are detected.
  * Changes to plugin exemplar packages now trigger the appropriate quality validation.

* **Documentation**
  * Updated CI documentation to describe the new dependency check.
  * Added guidance for an intentionally indirect dependency in the web UI example.

* **Tests**
  * Expanded CI planning and policy coverage for the new validation job.


.github/workflows/ci-quality-slice.yml
```diff
@@ -221,6 +221,29 @@ jobs:
       - name: Check formatting
         run: cargo fmt --all -- --check
 
+  cargo_machete:
+    name: Unused dependency check
+    needs: runner_policy
+    if: ${{ inputs.rust_changed }}
```
ndizazzo: Please include the standalone exemplar in this trigger. The planner classifies a change to docs/plugins/exemplars/web-ui/Cargo.toml as rust_changed=false, and this PR’s own Quality run skipped cargo_machete for that reason. As written, future changes to that package’s manifest or Rust sources can bypass the new check. Use the package path, or a dedicated planner signal, in addition to inputs.rust_changed while preserving the existing runner setup.

.github/workflows/ci-quality-slice.yml
```diff
@@ -221,6 +221,29 @@ jobs:
       - name: Check formatting
         run: cargo fmt --all -- --check
 
+  cargo_machete:
```
ndizazzo, line 228: Adding this job also changes the workflow’s job and audit inventory. The CI-compatible local suite and GitHub Quality run both fail because the expected sets still omit cargo_machete, and the audit count is still 25 instead of 26. Please update scripts/tests/test_ci_artifact_actions.py, scripts/tests/test_depot_authority_sentinel.py, and scripts/tests/test_depot_canary_workflow.py with the new job before merging.

ndizazzo (changes requested):
The new cargo-machete 0.9.2 check passes on this head, but I’m requesting changes because the workflow contract suite is now failing on the added job. The CI-compatible local just ci-validate run reached 1,508 tests and failed the same three expectations as GitHub’s Quality job: test_depot_pr_native_cache_consumers_obey_central_policy, test_five_pr_entrypoints_and_existing_build_shape_are_unchanged, and test_pr_audit_receives_central_runner_provider_selection. The audit count is still 25, but this job makes it 26. Please update those inventories and rerun just ci-validate. I also left an inline comment on the job trigger because the PR planner currently sets rust_changed=false for the standalone web-ui exemplar change.

michaelneale:
Reviewed (MeshBuilder, agent — on behalf of Mic). I like this: `#1869` removed the debt by hand, this stops it coming back. I reproduced Nick's three failures locally on `ed4f6b3a0` and they're all pure job-inventory bookkeeping — no design change needed. Exact edits:

```diff
--- a/scripts/tests/test_ci_artifact_actions.py   (line 2928)
-                "commit_convention", "runner_policy", "quality_contracts", "rust_fmt", "rust_clippy",
+                "commit_convention", "runner_policy", "quality_contracts", "rust_fmt", "cargo_machete", "rust_clippy",

--- a/scripts/tests/test_depot_authority_sentinel.py   (after line 639 "rust_fmt")
+            "cargo_machete": "runs-on: ${{ needs.runner_policy.outputs.runner_4 }}",

--- a/scripts/tests/test_depot_canary_workflow.py   (line 469)
-        self.assertEqual(audit_calls, 25)
+        self.assertEqual(audit_calls, 26)
```

With those applied: `test_ci_artifact_actions` 55 tests OK, `test_depot_authority_sentinel` 8 tests OK. `test_depot_canary_workflow` won't import here (no `yaml` in my python3.14), so I verified its assertion by hand instead — `grep -c "audit-depot-pr-isolation@" .github/workflows/*.yml` sums to exactly **26** on this head, so 25 -> 26 is the right number and not an off-by-one guess.

On CodeRabbit's inline point about `docs/plugins/exemplars/web-ui`: it's real but narrow. That exemplar is not a workspace member (not in root `Cargo.toml` `members`), so a change touching only it leaves `affected_crates` empty and `rust_changed=false` (`.github/actions/compute-changes/derive-outputs.sh:46`), and the job skips — while `cargo machete` at the repo root *does* walk into it. So the one directory that needs the `ignored = ["serde_json"]` escape hatch is also the one directory whose changes can't trigger the gate. I'd fix it, but I don't think it should block: the gate re-runs on the next rust-touching PR, so the worst case is a delayed flag, not a missed one. Simplest fix if you want it now is to keep `if: ${{ inputs.rust_changed }}` and let the planner route exemplar manifests into the rust domain, rather than special-casing a path in the job condition — one rule in one place.

Two smaller things, neither blocking:
- The `ignored = ["serde_json"]` in the web-ui exemplar deserves a one-line comment *in that Cargo.toml* saying it's consumed via a `#[path]` include machete can't follow. The reason is currently only in the PR body, and the next person to touch that manifest won't read the PR body.
- Pinning `cargo-machete@0.9.2` is the right call. If `manage-ci` maintainers later prefer image-baked tooling, that's a follow-up, not a change here.

i386 (dismissed):
Requesting changes — independently confirming @ndizazzo's review at `ed4f6b3`, with reproduction evidence from a clean worktree, plus the complete fix list.

## 1. Contract suite fails (3 tests — reproduced locally and on the Quality job)

Full `python3 -m unittest discover -s scripts/tests` (python 3.14, 1508 tests) fails on exactly three, all caused by this PR:

- **`test_ci_artifact_actions.py::test_depot_pr_native_cache_consumers_obey_central_policy`** — the fix here is the one that's still missing: this file has its *own copy* of the `expected_jobs` inventory (~line 2927). The PR updated only the twin in `test_ci_runner_selection_and_cache_policy.py:679`. Add `cargo_machete` to this copy too.
- **`test_depot_authority_sentinel.py::test_five_pr_entrypoints_and_existing_build_shape_are_unchanged`** — `expected_slice_jobs` (~line 639) needs `cargo_machete: "runs-on: ${{ needs.runner_policy.outputs.runner_4 }}"`.
- **`test_depot_canary_workflow.py::test_pr_audit_receives_central_runner_provider_selection`** — `self.assertEqual(audit_calls, 25)` at line 469 needs 26.

Tip to keep this from recurring: the quality-slice job inventory lives in at least four places (the workflow plus three test files) — `rg -n '"rust_fmt"' scripts/tests/` finds every copy that needs the new job.

One note on the PR body: "6 failures + 58 errors, all pre-existing or environmental" doesn't hold on a healthy interpreter — on python 3.14 the only failures in the full discover run are the three above, all from this PR (there is one environmental ERROR in `test_release_notes.py` from `git tag -a` in a dumb terminal; it passes in CI).

## 2. Trigger gap (Nick's inline + CodeRabbit finding — mechanism confirmed)

`scripts/affected-crates.sh:249` skips any changed file outside `crates/` and `tools/`, so a `docs/plugins/exemplars/web-ui/**`-only change yields `rust_changed=false` and `cargo_machete` skips — yet `cargo machete` run at the repo root *does* scan that standalone package (its own `Cargo.toml`/`Cargo.lock`). The `serde_json` ignore added in this very PR is proof the exemplar is in the tool's scope. Either extend the job condition with a dedicated planner output for that path, or classify `docs/plugins/exemplars/**` as rust-relevant in the classifier — layer choice is a maintainer call.

## What I verified as sound (no action needed)

- Ran the pinned `cargo-machete` 0.9.2 against this tree locally: exit 0 in under a second — the gate is green today and runtime is trivial against the 15m timeout.
- The `serde_json` ignore is a genuine false positive: `src/main.rs:1` includes `../manifest.rs` via `#[path]`, and `manifest.rs:52` uses `serde_json::json!`. Documented cargo-machete limitation; the metadata ignore is the sanctioned mechanism.
- Both `taiki-e/install-action` pins resolve (`26e9283f` = v2.87.13, `3d23c1bb` = just), and the job's audit-isolation / checkout-pinning / `runner_4` shape matches the `rust_fmt` sibling exactly.

Once the three inventory updates and the trigger condition land (and `just ci-validate` / the Quality job go green), this is an approve from me — the check itself works and the false-positive handling is correct.

.github/workflows/ci-quality-slice.yml
iamthesvn: Done. Added `cargo_machete` to both job-set copies (`test_ci_artifact_actions.py:2928`, `test_ci_runner_selection_and_cache_policy.py:679`) and all three spots in `test_depot_authority_sentinel.py`, and bumped the audit count to 26.

.github/workflows/ci-quality-slice.yml
iamthesvn: Went with the dedicated planner signal: `plugin_exemplars_changed` covers `docs/plugins/exemplars/**` and is threaded through planner, schema, action, and lane. The job now runs on `inputs.rust_changed || inputs.plugin_exemplars_changed`. Runner setup is untouched. Also added a planner test for the exemplar-only skip case.

iamthesvn:
@i386 Both fixed. All five inventory copies include the job now, and the audit count is 26. For the trigger gap I added a dedicated `plugin_exemplars_changed` signal that gates the job next to `rust_changed`, plus a planner test for the exemplar-only case. And fair catch on the PR body counts, fixed with the push.

iamthesvn:
@michaelneale Applied all three edits, thanks. On the trigger gap I went with the dedicated planner signal from Nick's inline suggestion. It keeps `rust_changed` scoped to workspace code, but your classifier route would work too. The serde_json comment is already there at `docs/plugins/exemplars/web-ui/Cargo.toml:7-8` on this head, so nothing to add.

ndizazzo (approved):

i386 (approved):
Approved at `a7e2197c72` — all four findings from the prior round are addressed, verified independently.

**Contract suite**: full `python3 -m unittest discover -s scripts/tests` at this head: 1509 tests, 0 failures (one environmental ERROR in `test_release_notes` from `git tag -a` under a dumb local terminal — passes in CI). The three inventory updates all landed, including the duplicate in `test_ci_artifact_actions.py` and the extra `cargo_machete` entries in the sentinel protection list I hadn't explicitly listed.

**Trigger gap**: fixed at the right layer — new planner-owned `plugin_exemplars_changed` signal derived in `scripts/plan-ci.py`, wired through `ci-plan.schema.json`, `plan-ci/action.yml`, the lane, and the job condition (`rust_changed || plugin_exemplars_changed`), with a planner test asserting the exemplar-only change sets it. The `|| false` default on the lane's pass-through correctly handles plans produced before this signal existed (merge-window drift).

Also re-verified: `actionlint` clean at this head, and the pinned `cargo-machete` 0.9.2 exits 0 on this tree in under a second.

One note for the merger: every entry workflow on this head is sitting in `action_required` (first-time contributor approval), so GitHub CI has not actually run on `a7e2197c72` — the verification above is local. Approve the pending runs (or let the merge exercise it) before merge.

PR #1896, opened 2026-09-17
feat(mesh): genesis invitations for private mesh, and correcting naming of models for remote access

michaelneale:
Private-mesh members can now restart after their invite code expires and automatically re-form the mesh using persisted signed membership state. A new node cannot use that expired code to join.

## Behavior
- Persists the signed private-mesh policy and up to 16 admitted peer addresses after admission.
- Restores membership only when the supplied signed token matches the saved mesh, owner, and policy hash.
- Replays saved peers during the rejoin loop, including when the original owner is offline.
- Clears adopted private membership when switching to a public identity.
- Treats corrupt persistence and write failures as non-fatal.

## Compatibility
The mesh wire format is unchanged. Old and new nodes remain interoperable; the restarting node needs this change to recover from an expired invite.

## Validation so far
A local three-node live matrix covered:
- joiner restart after invite expiry;
- two joiners restarting and reconnecting while the owner was offline;
- owner restart without a token;
- rejection of a fresh node attempting to use the expired token.

Full post-rebase formatting, Clippy, and package-test gates are still running; this PR is intentionally draft until those results are attached.

## Summary by CodeRabbit

- **New Features**
  - Persisted mesh membership can restore connectivity when bootstrap invitations expire.
  - Automatic rejoin redials saved peers and refreshes membership periodically.
  - Fresh invitations for a different mesh take precedence over previously saved membership.

- **Behavior Changes**
  - Expired invitations are accepted only when they match saved mesh membership.
  - Switching between private mode and public discovery creates fresh node and mesh identities; same-mode restarts preserve identity.
  - Public/private transitions also clear saved adopted membership.

- **Documentation**
  - Added guidance on identity transitions and membership behavior.


ndizazzo (dismissed):
I reviewed the current head `c2d56c517f578c8ffd7c8c16d7bd3830855c57fb` against `main`. The earlier findings around direct expired-token policy installation, token signature validation, first-peer capture, current-peer priority, and public/private cleanup are addressed by the follow-up commits. These issues remain:

### [P1] Restore membership for the token being attempted, not the whole explicit-token list

`run_auto_join_existing_mesh` calls `restore_adopted_mesh_membership(&options.join)` before trying any explicit token (`crates/mesh-llm-host-runtime/src/runtime/auto_join.rs:1055-1061`). The restore then accepts any valid token matching the persisted membership (`crates/mesh-llm-host-runtime/src/mesh/node_requirements.rs:203-208`) and installs that mesh policy. If the user supplies an expired token for mesh A followed by a fresh token for mesh B, an unavailable A can install A’s persisted policy before the fallback reaches B; B is then rejected by the active-policy check in `validate_bootstrap_token` (`crates/mesh-llm-host-runtime/src/mesh/node_requirements.rs:523-529`). This violates the documented “fresh invitations for a different mesh take precedence” behavior and breaks multi-token fallback. Remove the eager restore, or make it specific to the token currently being attempted, and add a regression test with mixed tokens.

### [P2] Move adopted-membership persistence off the direct gossip admission path

`validate_direct_peer_requirements` performs `persist_adopted_mesh_membership` from the async admission path (`crates/mesh-llm-host-runtime/src/mesh/node_requirements.rs:639-655`). The persistence function synchronously reads the file and, on changes, performs an atomic write whose implementation calls `sync_all` on both the file and parent directory (`crates/mesh-llm-host-runtime/src/mesh/node_requirements.rs:102-110`; `crates/mesh-llm-identity/src/keystore.rs:309-328`). Direct gossip is a recurring network path, so a slow disk or filesystem can block a Tokio worker and increase admission latency. Also, `peers()` returns `HashMap::values()` without stable ordering (`crates/mesh-llm-host-runtime/src/mesh/peer_state.rs:901-909`), while serialization preserves input order; equivalent peer sets can therefore produce different bytes and cause unnecessary writes. Coalesce persistence on state changes and run the durable write via a blocking worker, with deterministic address ordering.

### [P2] Redial saved peers with bounded concurrency

`redial_join_targets` dials every saved target serially (`crates/mesh-llm-host-runtime/src/mesh/node_requirements.rs:265-271`). Each attempt is governed by the 30-second connect/gossip timeout (`crates/mesh-llm-host-runtime/src/mesh/stage_transport.rs:229-230`), so a full set of 16 addresses can consume roughly eight minutes before later targets are tried. This runs in the periodic rejoin task and can delay recovery indefinitely when several stale addresses are present. Use bounded concurrent attempts, prioritize current/known-good addresses, and stop or refresh the set as soon as a live peer is found.

### [P3] Enforce the address cap when loading persisted membership

The writer truncates to `MAX_ADOPTED_PEER_ADDRS`, but `load_adopted_mesh_membership` verifies the signed policy without validating `peer_addrs.len()` (`crates/mesh-llm-host-runtime/src/mesh/node_requirements.rs:32-40,62-74`). A locally modified or legacy file can therefore restore an arbitrary number of targets; `restore_adopted_mesh_membership` remembers all of them and the redial loop attempts all of them. Reject or truncate over-sized input at load/restore as a low-cost local DoS and resource bound.

Please address the P1/P2 findings and add coverage for mixed explicit tokens and the persistence/redial behavior before merging.

ndizazzo:
Follow-up correction for the review above: the live head is now `699f0d0e5beb3accbf92895a677a33862f54beb5`. GitHub associated the requested-changes review with that commit, although the body names the preceding `c2d56c5` validation head.

The findings still apply. Current-head locations are:

- P1 mixed-token restore: `crates/mesh-llm-host-runtime/src/runtime/auto_join.rs:1060-1061`, `crates/mesh-llm-host-runtime/src/mesh/node_requirements.rs:223-240`, and the active-policy rejection at `node_requirements.rs:531-557`. The new single-fresh-invite regression test covers a different case; it does not cover an expired matching token followed by a fresh token for another mesh when the first peer is unavailable.
- P2 synchronous persistence: `node_requirements.rs:668-684` calls `persist_membership`; the synchronous read/write is `node_requirements.rs:104-111`, with the durable `sync_all` calls in `crates/mesh-llm-identity/src/keystore.rs:309-328`.
- P2 serial redial: `node_requirements.rs:294-300`, with the 30-second timeout at `crates/mesh-llm-host-runtime/src/mesh/stage_transport.rs:229-230`.
- P3 uncapped load: `node_requirements.rs:64-75` validates the policy but still accepts an arbitrary `peer_addrs` length before `restore_adopted_mesh_membership` remembers every entry at `node_requirements.rs:261-263`.

The focused tests pass locally on `699f0d0`; the new GitHub run is still in progress after the previous run was cancelled by the head update.

michaelneale:
I think that P1 mauy be an agent fantasy

micspiral:
P1 scope clarification after discussion with Mic and inspection of head 699f0d0e5: the intended contract is one target mesh, with multiple invites providing peer entry points into that mesh—not automatic fallback between different meshes. Explicit replacement of the target mesh is a separate operation. We should not introduce cross-mesh rollback or fresh-token ranking for this PR.

The mixed input is mechanically expressible: the CLI exposes repeatable --join (crates/mesh-llm-cli/src/parser/commands.rs:457-459), and run_auto_join_existing_mesh iterates that list. So the report should not be dismissed as unreachable/dead code. However, cross-mesh fallback is not the intended contract.

There is a concrete simplification to make: remove the eager whole-list restoration at runtime/auto_join.rs:1060. prepare_join_target already restores membership specifically for the expired token being attempted (mesh/node_identity.rs:469-477). The restoration helper itself is NOT dead: direct join and join_with_retry reach it through prepare_join_target. Removing the eager call alone does not implement A-unavailable -> B fallback, and we will not claim it does.

This is a scope/source-trace response, not a claim of a new fix or new validation. The narrow cleanup still needs validation preserving expired same-mesh recovery and explicit replacement with a fresh invite.

Issue #1884, opened 2026-09-15
main does not build: model_presentation.rs calls a cfg(test) method left by #1708


Virgile-pct:
`main` at 85b142df0 does not build `mesh-llm-host-runtime` outside of `cfg(test)`:

```
error[E0599]: no method named `upsert_served_model_descriptor` found for reference `&mesh::node::Node` in the current scope
  --> crates\mesh-llm-host-runtime\src\runtime\model_presentation.rs:44:18
   |
44 |             node.upsert_served_model_descriptor(descriptor).await;
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
help: there is a method `set_served_model_descriptors` with a similar name
```

Two pull requests merged six minutes apart, each green on its own base:

- 912e3071d, #1879 at 20:51Z, adds `runtime/model_presentation.rs`, whose `apply` calls `node.upsert_served_model_descriptor(...)` from production code.
- 218e402e7, #1708 at 20:56Z, reworks `mesh/node.rs` and leaves `upsert_served_model_descriptor` under `#[cfg(test)]` at `node.rs:1355`, since its own callers were tests and `runtime/local.rs`.

Neither branch saw the other, so the compile error only exists on `main`. The `Main · Quality` run on 85b142df0 fails in all three Clippy batches, and I hit it locally on Windows 11 by rebasing an unrelated test-only branch onto that head.

Smallest fix I can see: drop the `#[cfg(test)]` on `upsert_served_model_descriptor`, since it now has a production caller. I have a branch with exactly that change and a passing lib build, happy to open it or leave it to whoever owns the collision.

PR #1899, opened 2026-09-17
fix(system): read process facts on Windows instead of answering Ok(None)

### LINKED ISSUE #1886: Windows reports every local runtime instance as dead: process discovery returns Ok(None) for any PID
On Windows, every local runtime instance is reported dead, including the one asking. `mesh-llm doctor` never finds a local instance, the console never publishes one, and the stop path skips every owner it meets. The three tests that would say so fail on Windows today; two of their siblings were marked `cfg(not(windows))` instead.

## Where it comes from

`crates/mesh-llm-system/src/process.rs` has a Linux module reading `/proc`, a macOS module shelling out to `ps`, and this for everything else:

```rust
#[cfg(not(any(target_os = "linux", target_os = "macos")))]
mod platform {
    pub fn process_comm(_pid: u32) -> anyhow::Result<Option<String>> {
        Ok(None)
    }
    pub fn process_started_at_unix(_pid: u32) -> anyhow::Result<Option<i64>> {
        Ok(None)
    }
    pub fn process_executable_name(_pid: u32) -> anyhow::Result<Option<String>> {
        Ok(None)
    }
}
```

The documentation table in `runtime/instance.rs` says the same thing, "Other: `Ok(None)`". The problem is what the callers make of `Ok(None)`:

```rust
pub fn process_liveness(pid: u32) -> Liveness {
    match process_comm(pid) {
        Ok(Some(_)) => Liveness::Alive,
        Ok(None) => Liveness::Dead,
        Err(_) => Liveness::Unknown,
    }
}
```

`Ok(None)` was designed to mean "no such PID". On Windows it means "I did not look", and the two are indistinguishable, so every PID is `Dead`.

## What that does in the product

`runtime/instance.rs:465`, in the local instance scan:

```rust
if validate::process_liveness(meta.pid) == validate::Liveness::Dead {
    return None;
}
```

Three consumers sit behind that scan:

- `mesh-llm doctor` (`crates/mesh-llm/src/commands/doctor.rs:193`) reports `local_instances: []` and, for the native log capture, "no live local runtime instance matched the requested console port", on a machine where the runtime is up.
- the serving surface (`runtime/serving_surface.rs:1456`) publishes the same empty scan to the console.
- `collect_runtime_stop_targets` skips "stale directories (dead owner)", which on Windows is every directory.

Measured on Windows 11, from a throwaway test inside `runtime/instance.rs` asking about its own process:

```
PROBE os=windows pid=28632
PROBE process_comm            = Ok(None)
PROBE process_executable_name = Ok(None)
PROBE process_started_at_unix = Ok(None)
PROBE process_liveness(self)  = Dead
PROBE process_liveness(999999)= Dead
```

The running process and a PID that does not exist get the same answer.

## The tests already say it

On Windows 11, `cargo test -p mesh-llm-host-runtime` on `main` at 950e90552:

```
runtime::instance::scan_tests::scan_includes_self               left: 0, right: 1  "own instance must appear in results"
runtime::instance::scan_tests::scan_reads_all_fields            left: 0, right: 1
runtime::instance::tests::validate_self_process_comm_returns_something   "process_comm should return Some for self process"
```

while `validate_pid_matches_rejects_wrong_comm` and `validate_pid_matches_rejects_wrong_start_time` in the same module carry `#[cfg(not(windows))]`.

## What a fix looks like

`windows-sys` 0.61 is already a `cfg(windows)` dependency of `mesh-llm-host-runtime` and `mesh-llm-log-store`, so a Windows `platform` module needs no new crate: `OpenProcess` with `PROCESS_QUERY_LIMITED_INFORMATION`, `QueryFullProcessImageNameW` for the executable name (which also answers `process_comm`, matching how `binary_process_name` already strips the extension on Windows), and `GetProcessTimes` for the start time. A PID that `OpenProcess` cannot open with `ERROR_INVALID_PARAMETER` is the "no such process" case; anything else is `Unknown`, not `Dead`.

I can take this. Happy to send the PR with the three tests above un-gated as the acceptance criteria, if that suits.

## Not verified

Nothing run on Linux or macOS. I have not exercised `mesh-llm doctor` against a running Windows node for this write-up; the consumer trace is from the code, and the probe is the measured part.

Virgile-pct:
Fixes #1886.

On Windows every local runtime instance was reported dead, including the one asking: the process module answered `Ok(None)` for any PID and `process_liveness` reads that as `Dead`. The issue has the full trace; this is the implementation.

## What this changes

A Windows `platform` module in `mesh-llm-system/src/process.rs`, next to the Linux and macOS ones, answering the same three questions through the Win32 process API:

- `process_comm` and `process_executable_name`: `OpenProcess` with `PROCESS_QUERY_LIMITED_INFORMATION`, then `QueryFullProcessImageNameW`, reduced to the image stem so it matches what `binary_process_name` derives from a configured binary on Windows (`mesh-llm`, not `mesh-llm.exe`).
- `process_started_at_unix`: `GetProcessTimes`, creation time converted from FILETIME ticks to Unix seconds.

Contract kept identical to the Unix modules: a PID that names no process (`ERROR_INVALID_PARAMETER`) is `Ok(None)`, a process we may not query (`ERROR_ACCESS_DENIED`) is `Ok(None)` the way a denied `/proc` read is on Linux, anything else is an error and lands in `Unknown` rather than `Dead`.

Three consequences that were gated on the facts being available:

- `validate_pid_matches` drops its `#[cfg(not(windows))]`.
- `send_signal_if_matches` in `backend.rs` no longer compiles the identity check out on Windows. Until now it ran `taskkill /PID` on whatever PID the owner file held, so a stale record with a reused PID could take down an unrelated process tree. The check now runs on every platform before any signal.
- The two `validate_pid_matches_*` tests in `runtime/instance.rs` lose their `#[cfg(not(windows))]`.

`windows-sys` 0.61 is already in the graph through `mesh-llm-host-runtime` and `mesh-llm-log-store`; `mesh-llm-system` gains a `cfg(windows)` dependency on it with two features. The FILETIME arithmetic lives outside the Windows module and is unit-tested on every runner, the way `parse_lstart_naive` is for macOS.

## Measured

Windows 11. On `main` the three tests below fail and the next two are gated off the platform; on this branch, `cargo test -p mesh-llm-host-runtime -- runtime::instance::`:

```
test runtime::instance::scan_tests::scan_includes_self ... ok
test runtime::instance::scan_tests::scan_reads_all_fields ... ok
test runtime::instance::tests::validate_self_process_comm_returns_something ... ok
test runtime::instance::tests::validate_pid_matches_rejects_wrong_comm ... ok
test runtime::instance::tests::validate_pid_matches_rejects_wrong_start_time ... ok
test result: ok. 19 passed; 0 failed; 0 ignored
```

And the module's own tests, `cargo test -p mesh-llm-system --features dynamic-native-runtime -- process::`, with the new FILETIME case and the two Windows cases:

```
test process::tests::filetime_ticks_at_the_unix_epoch_convert_to_zero ... ok
test process::tests::windows_reports_the_current_process ... ok
test process::tests::windows_reports_a_missing_pid_as_dead ... ok
test result: ok. 8 passed; 0 failed
```

`windows_reports_the_current_process` checks that the name is the image stem, that the start time is within the hour, and that liveness is `Alive`; `windows_reports_a_missing_pid_as_dead` uses an odd PID, which Windows never allocates, and expects `Ok(None)` and `Dead`.

## Gates

- `cargo clippy -p mesh-llm-system -p mesh-llm-host-runtime --all-targets`: no lint in any touched file; complete sorted set identical to the reference build.
- `scripts/check-env-mutation-contract.py` and `python -m unittest scripts.tests.test_env_mutation_contract`: passed, 234 sites, no environment mutation added.

## Not verified

Nothing run on Linux or macOS; the module is `cfg(windows)` and the unconditional parts are the FILETIME helper, tested everywhere, and the removal of two `cfg` gates on code that already ran there. I have not exercised `mesh-llm doctor` against a running node on Windows for this write-up; the acceptance is the test set above.

The host-runtime tests were run with #1885 cherry-picked on top, since `main` at 85b142df0 does not build that crate (#1884). The branch itself carries only the change described here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows process matching to handle differences in executable-name capitalization.
  * Updated Windows process termination handling to work reliably with platform-specific process behavior.
  * Corrected a Windows process validation test that could produce inconsistent results.

* **Tests**
  * Expanded process validation coverage to run on Windows.
  * Added coverage for platform-specific executable-name matching.


Virgile-pct:
Both taken, as two commits.

**The process id.** The comment I wrote there was wrong, so thanks for pulling on it. I had claimed `OpenProcess` rejects a value that is not a multiple of four as a bad parameter. It does not. Measured on this machine against a live process, taking its own id and adding one, two, then three:

```
pid 34140  the live process itself
pid 34141  +1  -> handle opened
pid 34142  +2  -> handle opened
pid 34143  +3  -> handle opened
```

The handle manager ignores a process id's low two bits, so `999_999` reaches `999_996` and the assertion was holding only because nothing on the runner happened to own it. PID 0 replaces contingency with a documented contract: `OpenProcess` fails there with `ERROR_INVALID_PARAMETER`, which is one of the two error mappings this PR introduces, so the test now exercises that mapping deliberately rather than by accident.

**The case comparison.** Confirmed, and worth saying that this branch is what exposes it: `validate_pid_matches` carried `#[cfg(not(windows))]` until now, so the comparison had never run on Windows at all. What makes the two sides able to disagree is that they reach the name through different APIs. `owner.json` records `std::env::current_exe`, which is `GetModuleFileNameW` and returns the path as it was handed to `CreateProcess`, so as the caller spelled it. The live name comes from `QueryFullProcessImageNameW`, which returns the canonical path on disk. Launching a process through an upper-cased copy of its own path:

```
GetModuleFileNameW            C:\USERS\...\PYTHON.EXE  ->  stem PYTHON
QueryFullProcessImageNameW    C:\Users\...\python.exe  ->  stem python
```

The exact comparison then rejects the process as not being itself, `validate_pid_matches` returns false, and the stop target is left running. Windows opens files without regard to case, so reaching a binary through a differently cased path is ordinary rather than exotic. The comparison now folds ASCII case on Windows, the executable stems it compares being ASCII, and stays exact on every other platform, which is what those filesystems call for. There is a unit test asserting the fold follows the platform.

Gates on native Windows, against the branch head: `cargo fmt --all --check` clean, `cargo test -p mesh-llm-system --features dynamic-native-runtime` 223 passed and 0 failed, `cargo clippy -p mesh-llm-system --all-targets` with the same set of warnings as before the change, `xtask repo-consistency no-console-print` and the environment mutation contract both clean. The intermediate commit builds and passes on its own.

ndizazzo (approved):
Both noted issues are addressed, and I verified the fixes against the branch head.

**Process id.** Replacing `999_999` with PID 0 was the right call. Your measurement that the handle manager ignores the low two bits explains why the old assertion only held by luck, and the test now exercises the `ERROR_INVALID_PARAMETER` mapping the module declares for a missing process deliberately instead of by accident. The comment recording the measurement is exactly what the next reader of this test needs.

**Case comparison.** I confirmed `names_match` folds ASCII case on Windows and stays exact on the other platforms, with the unit test pinning the fold to the platform's filesystem. Your two-APIs account of it, `current_exe` reporting the path as the caller spelled it while `QueryFullProcessImageNameW` reports the canonical path on disk, is the right root cause for a process rejecting itself when the launch path differs only in case, and `eq_ignore_ascii_case` covers that gap exactly, since the stems being compared are ASCII.

Re-reading the rest of the PR: the identity check in `send_signal_if_matches` now runs on every platform before the signal, so a stale owner file holding a reused PID can't reach `taskkill` anymore. `validate_pid_matches` is ungated on Windows, the image stem matches what `binary_process_name` derives on Windows (both sides use `file_stem`), and the FILETIME arithmetic sits outside the Windows module so it's unit-tested on every runner.

One note on CI: the cancelled Quality check-runs on this head are the 16:16 UTC run, cancelled by concurrency when the re-run started; the re-run completed green (Clippy batches, format, and CI contracts all pass, and both parent suites are green), so nothing on the head is failing or pending. CodeRabbit's docstring coverage warning is non-blocking as far as I can tell.

Approved.

PR #1900, opened 2026-09-17
test(host-runtime): give the strict absolute-path case a path that is absolute everywhere

Virgile-pct:
One test, `unconfigured_strict_absolute_path_ignores_default_model_path`, fails on Windows because its fixture is a POSIX path.

## What happens today

The test hands `local_required_runtime_model_path` the spec `/models/requested.gguf` and expects it back, since a spec that is an absolute path is used as-is. The helper decides with `Path::is_absolute()`, and on Windows a leading slash is rooted but not absolute, so the helper answers `None`:

```
assertion `left == right` failed
  left: None
 right: Some("/models/requested.gguf")
```

The production rule is the right one: `org/repo` is not a path, and a rooted spec without a drive is not a usable path on Windows either. Only the fixture is POSIX-shaped, the same `is_absolute()` divergence #1854 hit on the delete guard, seen from the other side.

## What this changes

The test builds the spec from `std::env::temp_dir()`, absolute on every platform, and compares against that same path. Nothing in production moves.

## Measured

Windows 11, `cargo test -p mesh-llm-host-runtime -- runtime::model_lifecycle::load::tests`. On `main` the test fails with the assertion above; on this branch:

```
test runtime::model_lifecycle::load::tests::unconfigured_strict_absolute_path_ignores_default_model_path ... ok
test result: ok. 16 passed; 0 failed; 0 ignored
```

## Gates

- `cargo clippy -p mesh-llm-host-runtime --all-targets`: complete sorted set identical to the reference build.

The tests were run with #1885 cherry-picked on top, since `main` at 85b142df0 does not build the crate (#1884). The branch itself carries only the one test change.

## Not verified

Nothing run on Linux or macOS. There the temp dir is absolute as well and the assertion is the same shape as before.

## Summary by CodeRabbit

* **Tests**
  * Updated path handling coverage to work consistently across operating systems.
  * Verified that an explicitly requested model path takes precedence when no default model path is configured.


i386 (approved):
Approved — clean test-only fix, verified independently.

What I checked:

- **Helper contract confirmed**: `local_required_runtime_model_path` (load.rs:223) gates on `Path::is_absolute()`, which is the correct production rule — on Windows `/models/requested.gguf` is rooted but not absolute (no drive), so `None` is the right answer there. The old fixture baked in a POSIX assumption; this fixes the fixture, not the code. Right call.
- **Closed the "not verified on Linux/macOS" gap**: ran `cargo test -p mesh-llm-host-runtime --lib -- runtime::model_lifecycle::load::tests` on macOS at `ba5adf6fe` — 16 passed, 0 failed, including `unconfigured_strict_absolute_path_ignores_default_model_path`.
- **CI**: the 6 cancelled check-runs on this head belong to superseded run 5610 (duplicate launch at 15:39); the latest `PR · Quality` (run 5612) and the Linux/Windows/macOS lanes all completed green on `ba5adf6fe`. Merge state: clean, no conflicts with current main (7958cb1).

One non-blocking note for a future touch-up (not this PR): if a Windows `%TEMP%` ever contains non-UTF-8-representable bytes, `to_string_lossy()` would fail the assert loudly rather than silently — acceptable for a test, just be aware the assert is round-trip-sensitive.

Nice PR body, by the way — reproduction, measured before/after, and the #1885 cherry-pick caveat made this fast to verify.

Issue #1906, opened 2026-09-17
`mesh-llm claude` points Claude Code at an API that serves no `/v1/messages`, so every session fails on its first request


Cam10001110101:
## Problem

`run_claude` launches Claude Code with `ANTHROPIC_BASE_URL` set to the local node. Claude Code
speaks the Anthropic Messages protocol, so it POSTs `/v1/messages` — a route the inference API does
not serve:

```console
$ curl -s -X POST localhost:9337/v1/messages -H 'content-type: application/json' \
    -d '{"model":"<id>","max_tokens":1,"messages":[{"role":"user","content":"hi"}]}'
{"error":{"message":"route not found: /v1/messages","type":"invalid_request_error","param":null,"code":"not_found"}}

$ curl -s -X POST 'localhost:9337/v1/messages?beta=true' …
{"error":{"message":"route not found: /v1/messages?beta=true", …}}
```

Claude Code 2.1.27x appends `?beta=true`, so a route added later needs to tolerate the query string
— note that the current message echoes it back rather than stripping it.

Observed from the client side as an instant `is_error` result under `claude -p --output-format
json`, and as a launch that retries and exits interactively.

**Minimum acceptable fix: say so.** `check_mesh` already probes `/v1/models` before launching; one
additional probe, or one sentence in `mesh-llm claude --help`, would turn an unexplained failure
into a clear message. Serving or translating the route is the larger ask, and I am not assuming you
want it.

## Steps to reproduce

1. Start any node so `http://127.0.0.1:9337/v1/models` answers.
2. Run the two curls above. → `route not found`, with and without `?beta=true`.
3. Equivalently `mesh-llm claude`, on Linux (on macOS the launcher crashes first — #1902).

## Diagnostics

- mesh-llm v0.76.2 (released binary, current release), re-verified 2026-09-17. Also seen on
  v0.75.1 on linux-aarch64.
- Reproduces on a serving node and a client node alike, and over an ssh tunnel.

**Where it stands in the source (v0.76.2 tag):** the frontend registers seven routes — `/health`,
`/healthz`, `/readyz`, `/v1/models`, `/v1/chat/completions`, `/v1/completions`, `/v1/responses` —
at [`crates/openai-frontend/src/router.rs:254-260`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/openai-frontend/src/router.rs#L254-L260),
with a `.fallback(not_found)` at `:262` that produces the exact message above
[...]

PR #1907, opened 2026-09-18
fix(agent-cli,openai): unblock mesh-llm claude and make MoA error bodies honest

### LINKED ISSUE #1902: `mesh-llm claude` crashes Claude Code on macOS: the launcher reopens `/dev/tty` as stdin and Bun's kqueue rejects it (EINVAL)
## Problem

On macOS, `mesh-llm claude` finds the model, prints the launch line, and Claude Code exits before
drawing anything:

```
$ mesh-llm claude
   Models: meshllm/Qwen3.8-27B-UD-Q4_K_XL-layers, mesh
   Using: meshllm/Qwen3.8-27B-UD-Q4_K_XL-layers
🚀 Launching Claude Code with meshllm/Qwen3.8-27B-UD-Q4_K_XL-layers → http://127.0.0.1:9337

EINVAL: invalid argument, kqueue
 syscall: "kqueue",
   errno: -22,
    code: "EINVAL"

      at pull (unknown:1:1)

Bun v1.4.3 (macOS arm64)
claude exited with exit status: 1
```

Running `claude` directly in the same terminal works. Expected: the launcher starts an interactive
Claude Code session.

**Cause.** `run_claude` passes the child through `configure_interactive_stdio`, which opens
`/dev/tty` and hands that fd to the child as stdin, stdout and stderr
([agent_cli.rs#L19-L34](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L19-L34),
called at [#L674](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L674)).
Claude Code's native build is a Bun binary. Bun registers stdin with kqueue, and on macOS
`kevent(EVFILT_READ)` rejects an fd opened on `/dev/tty` with `EINVAL`. The same terminal inherited as
`/dev/ttysNNN` is accepted.

This is the failure #729 already avoided for OpenCode, also a Bun app. That PR's own description
states the mechanism — it "avoids reopening `/dev/tty` for OpenCode specifically. OpenCode runs on
Bun, and Bun can fail while initializing TTY write streams when the child process receives
reopened terminal file descriptors from the launcher" — and the code comment at
[#L46-L48](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L46-L48)
repeats it. The fix was scoped to OpenCode's own launch path, so `run_claude` still calls the
helper at [#L674](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L674),
as do the Goose (#L607) and Pi (#L914) launchers. The code is the same on `main` today.

## Steps to reproduce

1. macOS, Claude Code installed with the native installer (`claude --version` → 2.1.x; its crash
   banner names Bun).
2. Start any node so `http://127.0.0.1:9337/v1/models` answers.
3. Run `mesh-llm claude` in an interactive terminal. → `EINVAL: invalid argument, kqueue`, exit 1.

No human terminal is needed — allocating a pty is enough, so this is testable in CI:

```bash
script -q /dev/null mesh-llm claude < /dev/null
```

It reproduces without mesh-llm. Each line changes one fd of `claude`, run from an interactive shell:

| Command | Result |
|---|---|
| `claude` | starts normally |
| `claude <>/dev/tty >&0 2>&0` (what the launcher does) | `EINVAL … kqueue`, exit 1 |
| `claude </dev/tty` (stdin only) | `EINVAL … kqueue`, exit 1 |
| `claude >/dev/tty` (stdout only) | starts normally |
| `claude 2>/dev/tty` (stderr only) | starts normally |

The kernel behaviour on its own, with no Bun involved:

```python
import os, select
kq = select.kqueue()
kq.control([select.kevent(0, select.KQ_FILTER_READ, select.KQ_EV_ADD)], 0, 0)   # OK (/dev/ttys003)
fd = os.open("/dev/tty", os.O_RDWR)
kq.control([select.kevent(fd, select.KQ_FILTER_READ, select.KQ_EV_ADD)], 0, 0)  # OSError 22 EINVAL
```

## Diagnostics

- mesh-llm 0.76.2 (released binary, current release), macOS 26 (Darwin 25.6.0), Apple Silicon
  (arm64), zsh
- Claude Code 2.1.274, native install (Bun v1.4.3 per the crash banner). Also seen on 2.1.273
- Private mesh, invite-token join. The launcher found a node already on :9337 and spawned none.
- `/api/status` is irrelevant: the crash happens in the child before any request is made.

## Extra context

**Suggested fix.** Inherit stdio whenever mesh-llm's own stdin is already a terminal, and keep the
`/dev/tty` reopen only as the fallback when it is not:

```rust
fn configure_interactive_stdio(command: &mut Command) {
    use std::io::IsTerminal;
    // A reopened /dev/tty fd cannot be registered with kqueue on macOS, which kills Bun-based
    // agents (Claude Code, OpenCode). Only reopen when there is no terminal to inherit.
    if std::io::stdin().is_terminal() {
        command.stdin(Stdio::inherit()).stdout(Stdio::inherit()).stderr(Stdio::inherit());
        return;
    }
    // … existing #[cfg(unix)] /dev/tty path, then the inherit fallback …
}
```

That covers Claude Code and would let OpenCode drop its special case. Goose and Pi use the same
helper; I have not tested whether they are affected, but both are launched the same way.

**Fixing stdio alone is not enough to make `mesh-llm claude` work on 0.76.2.** Three further
blockers sit behind this one. They are separate issues and I am happy to file them separately —
say the word if you would rather have them here:

- The API the launcher points Claude Code at serves no `/v1/messages`, the route
  `ANTHROPIC_BASE_URL` makes it call (`route not found: /v1/messages`, also with `?beta=true`).
- `run_claude` passes no context window, so Claude Code assumes its 200k default for the
  unrecognised id and never compacts; it does set `CLAUDE_CODE_MAX_OUTPUT_TOKENS: "128000"`
  ([#L643](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L643)),
  larger than the whole window of any model on my mesh.
- `/v1/responses` forwards `tools` to the Chat Completions validator unconverted, so a bridge that
  speaks the Responses API is rejected with `tools[0].function must be an object`.

### LINKED ISSUE #1905: A `/v1/responses` request that fails through the `mesh` MoA gateway returns a `chat.completion`-shaped error body
## Problem

When a non-streaming `/v1/responses` request is answered by the mixture-of-agents gateway
(`model: "mesh"`) and the reducers fail, the response body is the raw `chat.completion` object
rather than a Responses-API object. A client that parses `/v1/responses` replies gets a different
schema on the error path than on the success path.

The HTTP 502 itself is intended and documented in the source; only the body shape is wrong.

Observed on v0.76.2:

```console
$ curl -s -w '\nHTTP %{http_code}\n' -X POST localhost:9337/v1/responses \
    -d '{"model":"mesh","input":"hi","max_output_tokens":4, …}'
```

```json
{"object":"chat.completion",
 "error":{"code":"all_reducers_failed", …},
 "choices":[{"finish_reason":"error", …}]}
HTTP 502
```

The success path on the same endpoint returns `"object":"response"` as expected.

Expected: a Responses-API request gets a Responses-shaped body on both paths — or, if a
connection-level error body is deliberate, one that does not claim `"object":"chat.completion"`.

## Steps to reproduce

Any node where `model: "mesh"` resolves, non-streaming:

1. `POST /v1/responses` with `model: "mesh"` and a request the reducers will fail — for example one
   carrying a flat Responses-API function tool, which every candidate rejects.
2. Observe HTTP 502 with `"object": "chat.completion"`.
3. The same request with a shape the reducers accept returns `"object": "response"`.

Any other MoA failure should do; the request only has to end with
`error.code == "all_reducers_failed"`.

## Diagnostics

- mesh-llm v0.76.2 (released binary), reproduced 2026-09-17, single node, non-streaming.

**Cause (v0.76.2 tag,
[`crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/streaming.rs`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/streaming.rs)):**

The response branches are ordered so the failure case is taken before the Responses adapter is
consulted:

- [`:91-102`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/streaming.rs#L91-L102)
  — `} else if is_failure {` sends `body` verbatim with status 502, for every `response_adapter`
  value.
- [`:103-114`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/streaming.rs#L103-L114)
  — the next arm, `else if response_adapter == ResponseAdapter::OpenAiResponsesJson`, is the one
  that calls `chat_completion_to_responses_json(body)` ([`:594`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/streaming.rs#L594)),
  and it is unreachable once `is_failure` is true.

The streaming path does distinguish the adapter even though the JSON path does not: at
[`:65-76`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/streaming.rs#L65-L76)
`OpenAiResponsesStream` gets its own SSE shape.

## Suggested fix

Apply `chat_completion_to_responses_json` in the failure arm when `response_adapter ==
OpenAiResponsesJson`, keeping the 502 — i.e. translate the body, then send it with the failure
status, rather than choosing between translating and failing.

If the raw body is deliberate for connection-level errors, it would still be worth not labelling it
`"object": "chat.completion"` on a Responses request, since that is what a client will branch on.

i386:
Batch of three small fixes from the prioritised review of Cam's open issues (#1902, #1906 minimum, #1905). Together these make `mesh-llm claude` launch and fail honestly, and make MoA error paths shape-honest.

## #1902 — `mesh-llm claude` crashes on macOS (Bun kqueue vs reopened `/dev/tty`)

- `run_claude` handed the child reopened `/dev/tty` fds through `configure_interactive_stdio`; Claude Code's native build is a Bun binary and Bun's kqueue registration rejects that fd on macOS with `EINVAL` before the UI draws.
- Give the claude launcher its own `configure_claude_launch_command` helper that sets args but leaves stdio on inherit — the same carve-out #729 already made for OpenCode, with the mechanism documented in the same style.
- Goose and Pi launchers are unchanged.

## #1906 (minimum) — no `/v1/messages` route, unexplained failure

- Even with #1902 fixed, Claude Code still fails on its first request because it POSTs `/v1/messages` and the inference API serves only OpenAI-compatible routes.
- `run_claude` now probes the route after `check_mesh` and prints a warning on 404, pointing at `mesh-llm opencode` / `mesh-llm pi` as working alternatives.
- Documented the gap in `docs/AGENTS.md` and in the `claude` subcommand help text.
- Serving or translating Anthropic Messages remains open follow-up — I deliberately did not assume that design here.

## #1905 — MoA failure over `/v1/responses` returned a `chat.completion` body

- The failure branch in `write_moa_response` was ordered before the response-adapter check, so a non-streaming `/v1/responses` request that failed through the MoA gateway got the raw `chat.completion` object at HTTP 502 while the success path returned `object: "response"`.
- The 502 status is intended and documented; the fix translates the failure body through `chat_completion_to_responses_json` when the caller is a Responses-JSON client, matching what the streaming path already does. The translated body keeps `finish_reason: "error"` so the failure stays observable in-band.

## Validation

- `cargo test -p mesh-llm-commands --lib` — 269 passed (includes new `claude_launch_command_sets_expected_args`)
- `cargo test -p mesh-llm-host-runtime --lib` — 3575 passed (includes new `failure_body_translated_for_responses_json_adapter`, `failure_body_untouched_for_chat_adapter`)
- `cargo clippy -p mesh-llm-commands / -p mesh-llm / -p mesh-llm-host-runtime --all-targets -- -D warnings` — clean

Closes #1902
Closes #1905
References #1906

## Summary by CodeRabbit

- **New Features**
  - Added a pre-launch compatibility check for Claude Code.
  - Users are warned when the required `/v1/messages` route is unavailable and informed that the first request will fail.
  - Suggested alternatives include OpenCode, Pi, or an Anthropic-to-OpenAI proxy.

- **Documentation**
  - Added guidance explaining Claude Code’s protocol compatibility and available alternatives.


i386:
**Closing as superseded.** Both substantive fixes in this batch have already landed on `main`, and the only piece left is one we deliberately abandoned.

| #1907 fix | Landed as | Merge commit | Issue |
|---|---|---|---|
| #1902 — `mesh-llm claude` macOS stdio (Bun kqueue / `EINVAL`) | #1908 (merged 2026-09-18) | `8471188af` | #1902 **CLOSED** |
| #1905 — MoA `/v1/responses` 502 body shape | #1910 (merged 2026-09-18) | `c9bd31088` | #1905 **CLOSED** |

Verified on `origin/main` (`a2018cdf782fe95c5583a09aa2c1b7625c0bc49a`):
- `configure_claude_launch_command` is present at `crates/mesh-llm-commands/src/agent_cli.rs:55` (also called at `:692`).
- The non-streaming MoA failure path routes through `chat_completion_to_responses_json` at `crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/streaming.rs:99`.

**What actually remains in this PR** is the net diff after the merge from `main`: +42 lines across 3 files — only the `/v1/messages` **404 probe + warning** (`crates/mesh-llm-commands/src/agent_cli.rs`), the matching `claude` subcommand help text (`crates/mesh-llm-cli/src/parser/commands.rs`), and the `docs/AGENTS.md` note.

That piece is already abandoned by design:
- **#1909** (the probe-only split) is **CLOSED unmerged**.
- The gap it warns about is exactly what **#1911** fills — Anthropic Messages adapted through shared host ingress. On `main` today the inference API serves no `/v1/messages`; the only route with that path is the internal stage-message channel at `crates/skippy-server/src/http.rs:278`. Landing this would therefore ship help/doc text that goes stale the moment #1911 merges.

**It also cannot land as-is:** `Quality / Commit convention` rejects the squash subject — `fix(agent-cli,openai): unblock mesh-llm claude and make MoA error bodies honest` is multi-scope and is not Conventional Commits v1.0.0 — so the PR is `BLOCKED` on `PR · Quality`.

No work is lost by closing. #1906 stays open with #1911 as the real fix.

Closing as superseded.

PR #1908, opened 2026-09-18
fix(agent-cli): launch Claude Code with inherited stdio on macOS

### LINKED ISSUE #1902: `mesh-llm claude` crashes Claude Code on macOS: the launcher reopens `/dev/tty` as stdin and Bun's kqueue rejects it (EINVAL)

i386:
## Problem

`mesh-llm claude` crashes Claude Code on macOS before it draws anything (#1902):

```
EINVAL: invalid argument, kqueue
```

`run_claude` passes the child through `configure_interactive_stdio`, which opens `/dev/tty` and hands that fd to the child as stdin/stdout/stderr. Claude Code's native build is a Bun binary; Bun registers stdin with kqueue and on macOS `kevent(EVFILT_READ)` rejects an fd opened on `/dev/tty` with `EINVAL`. The same terminal inherited as `/dev/ttysNNN` is accepted — running `claude` directly in the same terminal works.

## Fix

Give the claude launcher its own `configure_claude_launch_command` helper that sets the command args but leaves stdio on inherit, exactly mirroring the carve-out #729 already made for OpenCode (also a Bun app). The mechanism is documented in the same style as the existing OpenCode comment.

Goose and Pi launchers are unchanged.

## Validation

- `cargo test -p mesh-llm-commands --lib` — 269 passed (new `claude_launch_command_sets_expected_args`)
- `cargo clippy -p mesh-llm-commands --all-targets -- -D warnings` — clean

Closes #1902

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Claude launches on macOS by preventing terminal initialization issues.
  - Preserved the expected model, settings, and MCP configuration when starting Claude.


ndizazzo (approved):
Mirrors the existing OpenCode/Bun carve-out, and I checked that the /dev/tty kqueue failure mode matches what this avoids. The args test pins the launch shape. Approving.

PR #1909, opened 2026-09-18
fix(agent-cli): warn when mesh-llm claude targets a node without /v1/messages (#1906)

### LINKED ISSUE #1906: `mesh-llm claude` points Claude Code at an API that serves no `/v1/messages`, so every session fails on its first request

i386:
## Problem

Claude Code speaks the Anthropic Messages protocol: its first request is `POST /v1/messages`. The mesh inference API serves only the OpenAI-compatible routes, so pointing `mesh-llm claude` at a node produced an instant `is_error` result with no explanation (#1906).

## Fix

- After `check_mesh` picks a model, probe the `/v1/messages` route on the selected node and print a warning when it answers 404, pointing at `opencode` / `pi` as working alternatives today.
- Document the gap in `docs/AGENTS.md` and in the `claude` subcommand help text.

Serving the route natively remains open follow-up work — this PR makes the failure legible rather than silent.

## Validation

- `cargo test -p mesh-llm-commands --lib` — 268 passed

Closes #1906

## Summary by CodeRabbit

* **Bug Fixes**
  * Claude Code sessions now check for Anthropic Messages API support before launching.
  * Displays a warning when the endpoint is unavailable, while still allowing the session to start.

* **Documentation**
  * Clarified Claude Code compatibility, supported API routes, expected request failures, and available alternatives.


crates/mesh-llm-commands/src/agent_cli.rs
```diff
@@ -650,6 +650,35 @@ pub async fn run_claude(model: Option<String>, port: u16) -> Result<()> {
...
 
     let base_url = format!("http://127.0.0.1:{port}");
+    // Claude Code speaks the Anthropic Messages protocol and POSTs
+    // /v1/messages, which the mesh inference API does not serve (only
+    // /v1/models, /v1/chat/completions, /v1/completions, /v1/responses).
+    // Probe once so the failure is explained at launch instead of an
+    // instant is_error result on the session's first request.
+    let messages_url = format!("{base_url}/v1/messages");
+    match client
+        .post(&messages_url)
```
ndizazzo, line 660: This is a live POST with max_tokens: 1, so on a node that actually serves /v1/messages (the open follow-up, or the translating-proxy workaround documented in AGENTS.md) every `mesh-llm claude` launch triggers a real 1-token generation. A GET probe gives the same 404 signal (405 or 2xx means the route exists) with zero inference side effect. Let's switch to GET. If you keep the POST for a reason, note the side effect in the warning.

ndizazzo (changes requested):
The probe and docs are exactly what this needed, and the 404-only trigger is the right call. One issue with the probe itself, flagged inline.

PR #1910, opened 2026-09-18
fix(openai): return Responses-shaped 502 body from MoA failures (#1905)

### LINKED ISSUE #1905: A `/v1/responses` request that fails through the `mesh` MoA gateway returns a `chat.completion`-shaped error body

i386:
## Problem

A non-streaming `/v1/responses` request that the mesh MoA gateway answered with a failure returned the raw `chat.completion` object at HTTP 502 — a different schema on the error path than on the success path (#1905). Clients parsing `/v1/responses` had to special-case the failure body.

## Fix

The failure branch was ordered before the response-adapter check, making the adapter arm unreachable for failures. Translate the failure body through `chat_completion_to_responses_json` when the caller is a Responses-JSON client, matching what the streaming path already does. The translated body keeps `finish_reason: "error"` so the failure stays observable in-band. The 502 status itself is intended and documented — only the body shape was wrong.

## Validation

- Full workspace test run — 3575 passed, 11 ignored

Closes #1905

## Summary by CodeRabbit

* **Bug Fixes**
  * Responses API clients now receive failures in the expected Responses API format instead of chat-completion-formatted error data.
  * Chat API clients continue to receive failure responses in their existing format.


ndizazzo (approved):
The bug is real (the is_failure arm was ordered ahead of the adapter arm) and the fix is correctly scoped to Responses-JSON callers via the same translator as the success path. The tests pin the pass-through and the 502 stays documented as intended. Approving.

PR #1911, opened 2026-09-18
feat(openai-frontend): adapt Anthropic Messages through shared host ingress

i386:
Claude-style clients currently cannot use the host's OpenAI serving routes. This adds a supported Anthropic Messages surface in `openai-frontend`, normalizing requests into chat completions before host routing and translating responses at the client boundary.

- `/v1/messages` shares affinity inputs, model selection, planner/MoA dispatch, mesh and plugin forwarding with chat. Original protocol/route context controls response formatting and telemetry. Hook flags, terminal usage and capsule metadata survive adaptation.
- Supports text/system prompts, images, function tools/results, indexed multi-tool streaming, sampling controls, structured output and user identity. Late usage is retained; backend errors do not produce successful termination. Chunked HTTP and split UTF-8 are decoded before SSE translation.
- `/v1/messages/count_tokens` uses a separate schema and the backend's chat template/tokenizer. Unsupported backends and media counting fail explicitly; there is no character estimate.
- Unsupported Anthropic features are rejected explicitly, including thinking/signatures, cache-control blocks, documents and server tools. This is a documented Messages subset, not batches/files or the entire Anthropic platform. See `crates/openai-frontend/ANTHROPIC.md`.

Validation on the updated working tree:

- Full `openai-frontend` package: 267 passed.
- Full `mesh-llm-host-runtime` package: 3,603 passed, 12 ignored.
- Full `skippy-server` package: 751 passed, 3 ignored (run independently to avoid native-runtime feature unification).
- Real Claude CLI 2.1.273 harness: streamed Read-tool invocation and tool-result round trip through actual host ingress passed. The harness uses an isolated configuration, dummy local key and deterministic OpenAI upstream; it requires explicit execution and is ignored in ordinary CI tests.
- Strict all-target Clippy for the three touched crates and shipped `mesh-llm`, shipped-crate check, formatting and `just no-console-print` passed. The local Darwin linker emits toolchain deployment-target warnings; no Rust lint suppression was added.

Regression coverage includes hooks, usage, images, concurrent tools, invalid/unsupported requests, router errors, count requests without max_tokens, named/auto/mesh ingress, and MoA success/failure wire responses. Live model, multi-node and real plugin-provider execution remain unverified; the deterministic Claude harness does not establish those deployment properties.

## Summary by CodeRabbit

* **New Features**
  * Added Anthropic Messages API compatibility through `/v1/messages`.
  * Added streaming responses using Anthropic-formatted server-sent events.
  * Added backend-supported `/v1/messages/count_tokens` token counting.
  * Added support for tool use, tool results, system prompts, stop sequences, images, caching metadata, and usage reporting.
  * Added Anthropic-formatted error responses and request validation.

* **Bug Fixes**
  * Lifecycle metadata now correctly labels Anthropic Messages requests and operations.
  * Improved streaming error handling and response translation.

* **Tests**
  * Added coverage for requests, responses, streaming, tools, errors, token counting, and agent workflows.


i386:
likely I will pick this up after the big refactor of skippy

crates/mesh-llm-host-runtime/src/network/openai/moa_gateway/progress.rs
```diff
@@ -426,6 +429,15 @@ async fn write_progress_event(
     completion_id: &str,
     sequence_number: &mut i32,
 ) -> std::io::Result<()> {
+    if adapter == proxy::ResponseAdapter::AnthropicMessagesStream {
```
ndizazzo, line 432: Withdrawn after checking the official SDKs. Both the [Python SDK](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/_streaming.py) and [TypeScript SDK](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/core/streaming.ts) discard ping events before passing message events to their consumers. I have not established a client failure from an early ping, so this should not block the PR. My count-token admission finding still applies.

crates/openai-frontend/src/anthropic/routes.rs
```diff
@@ -0,0 +1,263 @@
...
+            Self::MessageDelta { .. } => "message_delta",
+            Self::MessageStop { .. } => "message_stop",
+            Self::Ping { .. } => "ping",
+            Self::Error { .. } => "error",
+        }
+    }
+}
+
+/// Count with the selected backend tokenizer and its generation chat template.
+pub(crate) async fn messages_count_tokens(
```
ndizazzo, line 237: [P1] Keep count_tokens out of generation admission

The handler is correct once it runs, but this route still passes through host ingress where `request_context_budget` returns `None` only for `/v1/tokenize`. A large valid count request is therefore estimated as `body_len / 4 + completion` and can be rejected as over the model context before `count_chat_tokens` runs. Please include `/v1/messages/count_tokens` in that capability-RPC guard and add a host routing regression for a large count request; this endpoint has no generation context budget.

ndizazzo (changes requested):
I'm requesting changes for the count-token admission path described inline. I withdrew the early-ping finding after checking the official SDKs; that observation did not establish a client failure. The latest Linux and Quality runs were successful at the reviewed head.

PR #1912, opened 2026-09-18
fix(skippy): downgrade qwen4exp cache hits to full-state snapshots

### LINKED ISSUE #1901: Prefix-cache restore corrupts `qwen4exp` models: the indexer cache is never saved, so a cache hit loses the conversation
## Problem

On a node serving a `qwen4exp` model (observed with `Qwen3.8-Flash-Next-UD-IQ4_XS`), any request
that restores a prefix-cache snapshot answers as if most of the conversation were missing. A turn
that reused 14,983 of its 15,033 prompt tokens missed both a fact from the middle of the reused
prefix and a fact given in the new user message. The byte-identical request with a different
`prompt_cache_key` misses the cache, prefills from scratch, and answers correctly. In agent
clients (Claude Code through an OpenAI bridge, Pi) this shows up as the model replying "you haven't
asked me anything yet" right after a task is pasted, or re-running tool calls whose results are
already in the conversation. It happens whenever the new request continues an earlier exchange, so
long agent sessions hit it on most turns.

Expected: a cache hit changes latency, not the answer.

## Steps to reproduce

Against a node (or a client routing to nodes) serving a `qwen4exp` model through `/v1/chat/completions`:

1. Send turn 1: a system message, then a user message of ~15k tokens of log lines with
   `The secret word for this facility is OBSIDIAN-4.` in the middle, ending with a question that
   does not ask for it ("how many distinct facility types appear?"). Non-streaming,
   `temperature: 0`, `reasoning_effort: "none"`, `max_tokens: 80`, `prompt_cache_key: "K"`.
   With more than one replica, send it once per replica (identical requests round-robin) so every
   replica holds the snapshot.
2. Send turn 2 with the same key: turn 1's messages, the assistant reply verbatim, and
   `"New information: the code name for this week is COPPERLEAF. Answer two things on one line: the
   secret word from the logs, and the code name I just gave."`
3. Send the identical turn-2 body with a fresh random `prompt_cache_key`.

Observed, three trials with different needles, all identical in shape:

| Request | `cached_tokens` | Time | Answer |
|---|---|---|---|
| Turn 2, same key | 14,983 of 15,033 | ~2 s | "The logs do not contain a code word." |
| Turn 2, fresh key | 0 | ~44 s | "OBSIDIAN-4, COPPERLEAF" (correct) |

<details>
<summary>Script for the three steps above (stdlib only)</summary>

```python
import json, random, time, urllib.request, uuid
URL, MODEL = "http://127.0.0.1:9337/v1/chat/completions", "Qwen3.8-Flash-Next-UD-IQ4_XS"
random.seed(7)
logs = "\n".join(
    (f"Entry {i:04d}: Security note. The secret word for this facility is OBSIDIAN-4. Keep it on file."
     if i == 200 else
     f"Entry {i:04d}: crew {random.randint(100,999)} logged readings {random.randint(100,999)} and "
     f"{random.randint(100,999)} within tolerance.")
    for i in range(400))

def post(messages, key):
    body = {"model": MODEL, "messages": messages, "max_tokens": 80, "temperature": 0,
            "reasoning_effort": "none", "prompt_cache_key": key}
    req = urllib.request.Request(URL, json.dumps(body).encode(), {"content-type": "application/json"})
    t = time.time()
    with urllib.request.urlopen(req, timeout=900) as r:
        out = json.load(r)
    u = out["usage"]
    print(f"{round(time.time()-t,1)}s prompt={u['prompt_tokens']} "
          f"cached={u.get('prompt_tokens_details', {}).get('cached_tokens')} "
          f"-> {out['choices'][0]['message']['content'].strip()[:80]}")
    return out["choices"][0]["message"]["content"]

key = "probe-" + uuid.uuid4().hex
turn1 = [{"role": "system", "content": "You are a careful operations assistant. Answer briefly."},
         {"role": "user", "content": "Here are this week's facility logs.\n\n" + logs +
          "\n\nIn one short sentence, how many distinct facility types appear in the logs?"}]
reply = post(turn1, key)
post(turn1, key)  # once per replica, so every replica holds the snapshot
turn2 = turn1 + [{"role": "assistant", "content": reply},
                 {"role": "user", "content": "New information: the code name for this week is COPPERLEAF. "
                  "Answer two things on one line: the secret word from the logs, and the code name I just gave."}]
post(turn2, key)                          # cache hit -> wrong
post(turn2, "probe-" + uuid.uuid4().hex)  # fresh key -> correct
```

</details>

## Diagnostics

- mesh-llm 0.76.1 on the serving nodes (NVIDIA DGX Spark GB10, CUDA 13), 0.76.2 on the client
  node; private mesh. Model reports `metadata.architecture: "qwen4exp"`, served at ctx 32768.
- Detection is `usage.prompt_tokens_details.cached_tokens` on the response.
- No context shift is involved: this version rejects over-context prompts rather than shifting.
- Reproduced on 0.76.1. v0.76.2 and `main` were checked by reading the source, not by running.

**Cause (from reading the v0.76.1 tag and the pinned llama.cpp `b31b71f3`):**

- `qwen4exp` is a hybrid with three caches: attention KV, Gated DeltaNet recurrent state, and the
  sparse-attention (QSA) indexer keys. llama.cpp builds `llama_memory_hybrid_idx` for it
  (`src/llama-model.cpp`, `needs_mem_idx = (arch == LLM_ARCH_QWEN4EXP)`).
- Skippy's exact-state snapshot exports the attention KV page from
  `skippy_get_kv_cache`, which for a hybrid memory returns `hybrid->get_mem_attn()` only
  (`third_party/llama.cpp/patches/0004-skippy-implement-staged-execution-and-state-runtime.patch:5762`,
  used from `crates/skippy-server/src/kv_integration/exact_state.rs:280`), and the recurrent
  state with `LLAMA_STATE_SEQ_FLAGS_PARTIAL_ONLY` (same patch, `:296`, `:309`).
- llama.cpp writes the indexer cache only without that flag
  (`src/llama-memory-hybrid-idx.cpp:209-211`: "The indexer mirrors the attention cache, so it uses
  the same PARTIAL_ONLY gate"). Nothing in skippy exports it separately, so a restore brings back
  attention KV and recurrent state with an empty indexer.
- What happens next inside the graph I have not verified: with the indexer empty, the restored
  cells appear either tied or masked in the QSA top-k selection, so the sparse-attention layers
  see little of the restored prefix. The observable part is that no error is raised, the cell
  counts match, and `cached_tokens` reports a successful hit.
- By code reading the same import runs on downstream stages of a split, so splits should be
  affected too; I have not tested a split.

Closest existing report: issue #1434 (open) lists as finding B4 that `stage_export_kv_page`
"never exports `k_idx` indexer rows (GLM-DSA/MSA silently wrong…)", in the context of split
serving, and PR #1436 added `k_idx_row_bytes` to the KV page ABI. That covers indexer rows stored
inside the attention KV cache. `qwen4exp` keeps its indexer in a separate cache
(`llama_memory_hybrid_idx::get_mem_idx()`), which `skippy_get_kv_cache` never returns, and
`qwen4exp` support (PR #1509) landed after both. This report adds that architecture and the
single-node prefix-cache path, with a live reproduction. Other related work
that did not cover it: PRs #1253, #1342 and #1662 (recurrent prefix reuse; the smoke test
in #1662 uses a model without an indexer), PRs #1509 and #1513 (`qwen4exp` certification compared
one predicted token without a cache restore), and PR #1606 (replay certification). Unchanged in
v0.76.2, the latest release, and on `main` (`3d35e08b`, llama.cpp pin `661643e4`, same gate) as
of 2026-09-17.

## Extra context

**Workarounds found:**

- Per request: send a fresh `prompt_cache_key` on every request. It is hashed into the cache
  namespace (`crates/skippy-server/src/frontend/generation/cache_hints.rs:212`), so the lookup
  always misses. Costs a full prefill every turn. There is no `cache_prompt` field.
- Per node: `[models.model_fit] prompt_cache = false` (or `[models.model_fit.prefix_cache]
  enabled = false`), or `SKIPPY_PREFIX_CACHE=off` on the serving process.
- Untested: `prefix_cache.payload_mode = "full-state"` exports the whole native state without
  `PARTIAL_ONLY`, which should include the indexer and keep caching.

**Suggested fix:** export and import the indexer cache alongside the attention page for
`llama_memory_hybrid_idx` (or fall back to full-state payloads for any memory that has one), and add
a cache-restore equivalence check to model certification: the same continuation with and without a
restored snapshot must produce the same tokens.

i386:
## Problem

`qwen4exp` builds a separate QSA **indexer memory** on top of its hybrid attention/recurrent state (upstream `needs_mem_idx` allowlist, `llama-model.cpp`). The hybrid state serializer only carries that indexer when neither `PARTIAL_ONLY` nor KV-page extraction is involved, so server-side `KvRecurrent` snapshots restore attention + recurrent state while the indexer resets to its initial view. A prefix-cache hit then silently corrupts the conversation: the restored context and the indexer disagree from the first restored token (#1901).

## Fix

Resolve the payload at config level — no native exporter patches:

- **skippy-ffi**: bind `llama_model_meta_val_str` (static + dynamic symbol paths) and expose a safe wrapper that reads `general.architecture`, following the existing two-path probe pattern.
- **skippy-runtime**: `LoadedModelCapability.has_indexer_memory`, classified from an `INDEXER_MEMORY_ARCHITECTURES` allowlist that mirrors upstream `needs_mem_idx`. Missing metadata accessor fails closed (flag false → current behavior, never claims safety it can't verify).
- **skippy-server**: `effective_cache_payload` resolves `KvRecurrent` → `FullState` for indexer-capable models, so every cache-hit restore round-trips the complete session state via the existing full-state import path. Explicit `ResidentKv` stays untouched so the existing recurrent-mismatch rejection still fires. All four production `from_loaded_model` call sites pass the new flag; `RECURRENT_CACHE_MAX_ENTRIES` continues to clamp these entries in the exact-state catalog.

## Tests

- `effective_cache_payload` downgrade matrix: Auto + explicit `KvRecurrent` → `FullState` when indexed; explicit `ResidentKv` untouched; dense/flag-absent behavior unchanged.
- `from_loaded_model` integration: hybrid + indexer flag serves `StagePrefixCachePayload::FullState`.
- Allowlist classification unit in `skippy-runtime` (`qwen4exp` true; other archs and probe-absent false).

## Validation

- `cargo test -p skippy-runtime -p skippy-server -p skippy-ffi` (per-package): **164 + 753 + 11 passed, 0 failed**. One pre-existing quirk: invoking all three packages in a single `cargo test` command fails 3 `skippy-runtime` ngram/device tests with "native runtime library has not been loaded" — reproduced identically on clean `62a55f24d` without this diff, and everything passes per-package, so it is test-harness-level (concurrent test binaries contending for native runtime materialization), not from this change.
- `cargo clippy --all-targets -- -D warnings` clean on the three touched crates; `cargo fmt --check` clean; `mesh-llm-host-runtime` compiles against the widened API.

Closes #1901

## Summary by CodeRabbit

- **New Features**
  - Added support for reading string-valued model metadata.
  - Model capability detection now recognizes indexer memory for supported `qwen4exp` models.
  - Exposed indexer-memory capability information for runtime integrations.

- **Bug Fixes**
  - Cache snapshots for models with indexer memory now use full-state snapshots when needed, preventing inconsistent session restoration.
  - Added handling for missing or unsupported architecture metadata, defaulting safely when capabilities cannot be confirmed.


ndizazzo (approved):
Deep-verified against the vendored llama.cpp: the architecture allowlist matches `needs_mem_idx` 1:1, the `meta_val_str` wrapper is correct against the upstream snprintf contract, and it fails closed when the accessor is missing. The downgrade semantics and the test matrix check out. One thing to keep in mind: the allowlist is a cross-language duplicate of upstream `needs_mem_idx`, so it needs extending in lockstep if upstream adds an indexer arch (the comment already says so). Approving.

PR #1913, opened 2026-09-18
fix(responses): normalize flat function tools before chat validation

### LINKED ISSUE #1904: `/v1/responses` rejects Responses-API function tools: the endpoint validates them against the Chat Completions schema
## Problem

`/v1/responses` accepts a Responses-API request body, but tools are validated against the *Chat
Completions* schema. The flat function-tool shape the Responses API defines is rejected:

```
tools[0].function must be an object
```

The same tool wrapped in the Chat Completions shape — which is not the Responses shape — is
accepted and answered. The endpoint therefore requires the wrong schema for the API it implements.

Any client or bridge that speaks the Responses API cannot use tools against a mesh endpoint.

Expected: `/v1/responses` accepts Responses-API tools.

## Steps to reproduce

One node serving one model, no mesh required. Use a real model id (not the `mesh` virtual id — see
the note at the end).

```bash
# 1. Flat Responses-API tool — the shape this endpoint's own API defines
curl -s -w '\nHTTP %{http_code}\n' -X POST localhost:9337/v1/responses \
  -H 'content-type: application/json' -d '{
  "model":"<id>","input":"hi","max_output_tokens":4,
  "tools":[{"type":"function","name":"f","description":"d",
            "parameters":{"type":"object","properties":{}}}]}'
```

```json
{"error":{"message":"tools[0].function must be an object","type":"invalid_request_error",
          "param":"tools","code":"invalid_value"}}
HTTP 400
```

```bash
# 2. The same flat tool against /v1/chat/completions, where rejecting it is correct
curl -s -w '\nHTTP %{http_code}\n' -X POST localhost:9337/v1/chat/completions \
  -H 'content-type: application/json' -d '{
  "model":"<id>","messages":[{"role":"user","content":"hi"}],"max_tokens":4,
  "tools":[{"type":"function","name":"f","description":"d",
            "parameters":{"type":"object","properties":{}}}]}'
```

→ **byte-identical error, also HTTP 400**. That identity is the finding: `/v1/responses` is
applying the Chat Completions contract unchanged.

```bash
# 3. Control: the same tool nested Chat-style on /v1/responses succeeds
#    "tools":[{"type":"function","function":{"name":"f", … }}]   -> normal "object":"response"
```

## Diagnostics

- mesh-llm v0.76.2 (released binary), reproduced 2026-09-17 on a single node with no peers.
- Not model- or backend-specific: the request is rejected at the HTTP frontend before routing.

**Cause (all citations at the v0.76.2 tag):**

- `POST /v1/responses` is registered at
  [`crates/openai-frontend/src/router.rs:260`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/openai-frontend/src/router.rs#L260)
  and handled by `async fn responses` at
  [`:404`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/openai-frontend/src/router.rs#L404).
  That handler normalises the body, deserialises the result into `ChatCompletionRequest`, and calls
  `request.validate()` at
  [`:422`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/openai-frontend/src/router.rs#L422)
  — the same validation `/v1/chat/completions` runs at `:325`.
- The normaliser is `translate_openai_responses_input`
  ([`crates/openai-frontend/src/responses.rs:322`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/openai-frontend/src/responses.rs#L322),
  called at `:453`). It converts `instructions` and `input` into `messages`, derives
  `prompt_cache_key`, and removes eight Responses-only keys (`conversation`, `include`, `output`,
  `output_text`, `previous_response_id`, `store`, `text`, `truncation`). **`tools` is never read,
  converted or removed** — it is forwarded verbatim.
- `validate_tools_value`
  ([`crates/openai-frontend/src/chat.rs:130`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/openai-frontend/src/chat.rs#L130),
  reached from `:100`) then requires `tools[i].function` to be an object and emits the message at
  [`:142`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/openai-frontend/src/chat.rs#L142).

So this is not a leak into the wrong validator — the reuse is deliberate. The defect is that the
Responses endpoint inherits the Chat tool contract while nothing converts the tool shape on the way
in, even though the same function already converts the rest of the envelope.

**Two things suggest this is an oversight rather than a decision:**

- **The runtime behind the validator already accepts the flat shape.** Skippy reads a tool's object
  as either shape —
  [`tool_emulation.rs:121-124`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/skippy-server/src/frontend/tool_emulation.rs#L121-L124)
  (`tool.get("function")…or_else(|| tool.as_object())`) — and its name likewise
  ([`generation/parsing.rs:365-369`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/skippy-server/src/frontend/generation/parsing.rs#L365-L369),
  `…or_else(|| tool.get("name"))`). The frontend is stricter than the engine it guards.
- **Nothing tests it.** The only tools case for this endpoint,
  [`responses.rs:1182`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/openai-frontend/src/responses.rs#L1182),
  passes `{"type":"function","function":{"name":"lookup"}}` and asserts
  `body["tools"][0]["function"]["name"]` — the Chat shape is encoded as the expected input. The
  `/v1/responses` certification smoke sends no tools at all
  ([`certification.rs:429-433`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-host-runtime/src/inference/skippy/certification.rs#L429-L433):
  `model`, `input`, `max_output_tokens` only). There is no test asserting the flat shape is
  rejected, and no comment saying it is out of scope.

I am aware of `docs/AGENTS.md:11` ("Prefer chat-completions mode unless the client explicitly
documents Responses API support"). That is good advice for choosing a mode, but it does not settle
which schema `/v1/responses` should accept once a client does use it — and
`docs/EXO_COMPARISON.md:207` advertises the endpoint as "OpenAI Responses API" without qualification
while the handler already implements five Responses-only fields.

## Suggested fix

Convert tools inside `translate_openai_responses_input`, beside the conversions it already does:
lift `name`, `description`, `parameters` and `strict` into a nested `function` object, and leave
already-nested tools untouched so both shapes work. A test with the flat shape next to the existing
one at `responses.rs:1182` would pin it.

`tool_choice` and `function_call` also differ between the two APIs and are likewise untouched by
the normaliser; I have not tested whether they are handled elsewhere.

## Note on the `mesh` virtual model

If you reproduce this with `model: "mesh"`, the failure is wrapped by the mixture-of-agents
gateway: HTTP 502, `"code":"all_reducers_failed"`, and the 400 quoted inside a message string. The
"tried 3" in that text is the number of candidate models tried, not retries. Use a real model id to
see the plain 400. (That wrapped response also comes back shaped as a `chat.completion` rather than
a Responses object, which I am reporting separately.)

i386:
Responses clients sending flat function tools currently receive `tools[0].function must be an object`. Normalize flat tools and named function choices to the shared Chat representation before validation, retaining descriptions, schemas and strictness. Existing nested tools remain accepted; Chat requests still require the Chat schema.

Regression coverage exercises streaming and non-streaming Responses through the router and records the backend request. Malformed functions and unsupported tool types remain errors.

Validation: full frontend package 234 passed; full host package 3,597 passed, 11 ignored; strict all-target Clippy and cargo check for frontend/shipped mesh-llm, formatting and no-console-print passed. Validated working tree committed unchanged as b37612fed.

Closes #1904.

Originating Buzz channel: a9ea5982-8d74-407f-b82e-6b55c69d411f.

## Summary by CodeRabbit

* **New Features**
  * Added support for flat function-tool definitions and function tool choices in Responses API requests.
  * Converted supported tool formats for compatibility with Chat Completions processing.
* **Bug Fixes**
  * Function names are now trimmed before processing.
  * Added validation that rejects malformed tools, unsupported tool types, ambiguous definitions, and missing or whitespace-only function names with clear client errors.
* **Tests**
  * Added coverage for streaming and non-streaming requests, successful tool forwarding, and invalid request handling.


crates/openai-frontend/src/responses/tools.rs
```diff
@@ -0,0 +1,57 @@
...
+                OpenAiError::invalid_request("ambiguous flat and nested function tool")
+                    .with_param(param),
+            );
+        }
+        return Ok(false);
+    }
+    if !object
+        .get("name")
+        .and_then(Value::as_str)
+        .is_some_and(|name| !name.trim().is_empty())
```
ndizazzo, line 45: This check trims the name, but the stored name is untrimmed, so ` lookup ` passes validation and reaches the backend with the padding intact. Let's trim the name we store (or reject whitespace-padded names).

crates/openai-frontend/src/responses/tools.rs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        return Err(
+            OpenAiError::invalid_request("function name must be a non-empty string")
```
ndizazzo, line 48: This error can also fire for a tool_choice object, where it's a bit imprecise. Minor.

ndizazzo (changes requested):
The normalization is wired correctly end-to-end (I traced the fast path and the changed propagation, and the router tests prove the backend gets the nested shape in both modes). The direction is right. One real bug and one wording nit, inline.

crates/openai-frontend/src/responses/tools.rs
i386: Fixed in 2e980cc39 — the validated name is now stored trimmed, so whitespace padding can no longer pass validation and reach the Chat backend. Added a contract test round-tripping "  lookup  " to the backend as "lookup" in both streaming and non-streaming modes.

crates/openai-frontend/src/responses/tools.rs
i386: Fixed in 2e980cc39 — missing-name errors are param-aware now: tool_choice gets "tool_choice must reference a function by name" while tools keep the original message. Covered by responses_missing_function_names_error_per_parameter.

PR #1914, opened 2026-09-18
fix(agent-cli): use served model context limits in launchers

### LINKED ISSUE #1903: Agent launchers discard `metadata.context_length` from `/v1/models`: pi writes 262144 and goose 65536 for a model served at 16384
## Problem

`check_mesh` fetches `/v1/models` to build the model list it hands the agent, keeps only
`data[].id`, and throws the rest of the response away — including `metadata.context_length`, the
window the model is actually being served with, keyed by exactly the id the launcher just took.
The launchers then reconstruct that number from two other endpoints, and get it wrong.

On a node serving one model at `ctx_size 16384` (GGUF native 262144), on v0.76.2:

| Launcher | What it writes for that model | Source |
|---|---|---|
| `mesh-llm pi` | `contextWindow` / `maxTokens` **262144** — 16× the served window | `merge_context_lengths` |
| `mesh-llm opencode` | 262144, or `OPENCODE_DEFAULT_CONTEXT_LIMIT` 32768 when unknown | same |
| `mesh-llm goose` | **65536** for every model, always | hardcoded literal |
| `mesh-llm claude` | nothing — no context variable at all | — |

The agent then plans against a window the mesh will not serve.

Expected: each launcher sizes the agent from the window the mesh reports for the chosen id.

## Steps to reproduce

One node, one model, no mesh required. Serve a model below its native window (here `ctx_size =
16384`, native 262144), then:

```bash
curl -s localhost:9337/v1/models | jq '.data[] | {id, metadata}'
```

```jsonc
{ "id": "ornith-ai/Ornith-1.5-9B-GGUF:Q4_K_M",
  "metadata": { "context_length": 16384, "native_context_length": 262144, … } }
```

Both numbers are present and correctly distinguished. Now run `mesh-llm pi --write` and read
`~/.pi/agent/models.json`:

```json
{ "contextWindow": 262144,
  "id": "ornith-ai/Ornith-1.5-9B-GGUF:Q4_K_M",
  "maxTokens": 262144,
  "name": "ornith-ai/Ornith-1.5-9B-GGUF:Q4_K_M" }
```

and `~/.config/goose/custom_providers/mesh.json`:

```json
"models": [ { "context_limit": 65536, "name": "ornith-ai/Ornith-1.5-9B-GGUF:Q4_K_M" }, … ]
```

`models.json` also shows a second failure mode: a model this node is not running gets no
`contextWindow` key at all, so the agent silently falls back to its own default. On a four-model
mesh, three of the four entries had no `contextWindow` — the three served by peers — and the `mesh`
virtual id has none either, on any configuration I have seen.

## Diagnostics

- mesh-llm v0.76.2 (released binary), observed 2026-09-17.
- Reproduces on a single node for a model running locally — it is not a "remote model" edge case.

**Cause — three separate defects with one outcome (citations at the v0.76.2 tag):**

**1. The right answer is fetched and discarded.** `check_mesh`
([`agent_cli.rs:398`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L398))
GETs `/v1/models` and keeps ids only at
[`:425-430`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L425-L430):

```rust
models = body["data"]
    .as_array()
    .unwrap_or(&vec![])
    .iter()
    .filter_map(|model| model["id"].as_str().map(String::from))
    .collect();
```

`fetch_mesh_models` does the same at [`:495`/`:513`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L513).
All four launchers take their model list from one of these two.

**2. The replacement lookup resolves to the native window.** `fetch_model_context_lengths`
([`:1026`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L1026))
issues two more requests and merges them in `merge_context_lengths`
([`:1046`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L1046))
into one flat map fed by two **disjoint key spaces**:

- `/api/runtime/processes` contributes `processes[].name` — the *runtime process* name, e.g. `Ornith-1.5-9B-Q4_K_M`, carrying the correct `16384`;
- `/api/models` contributes `mesh_models[].name` — the *served id*, e.g. `ornith-ai/Ornith-1.5-9B-GGUF:Q4_K_M`,
  carrying the catalog's native `262144`.

The comment at [`:1052-1055`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L1052-L1055)
says the process value is "the actual context_length the model is running with", and the
`entry(n).or_insert(ctx_len)` at
[`:1072`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L1072)
is meant to give it precedence. But precedence only applies when both sources produce the *same*
key, and in practice they never do. The runtime entry lands under a name nothing ever looks up, so
it is inert, and the served-id lookup resolves to the catalog value — which the fallback's own
comment scopes to "remote models or stopped instances"
([`:1066`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L1066)),
neither of which describes a model this node is running.

To be precise about the mechanism: the lookup *succeeds*, on the wrong entry. It does not miss. A
miss is visible and different — `build_pi_provider_config_with_limits`
([`:710`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L710))
omits both keys entirely when the value is absent
([`:722-725`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L722-L725)),
which is what produces the keyless entries for peers' models.

**Why CI does not catch it:** every fixture in `agent_cli/tests.rs` uses the *same* name on both
sides (`ModelA`, `Qwen3-8B`), so the `or_insert` precedence works in tests and never does in
production. `merge_context_lengths_runtime_process_trumps_api_models` asserts exactly the behaviour
that is dead at runtime.

**3. Two launchers never consult any of this.** goose writes a literal at
[`:563`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L563)
(`json!({"name": name, "context_limit": 65536})`) for every id, and `run_claude` sets no context
variable at all in its injected settings — while setting
`"CLAUDE_CODE_MAX_OUTPUT_TOKENS": "128000"` at
[`:643`](https://github.com/Mesh-LLM/mesh-llm/blob/v0.76.2/crates/mesh-llm-commands/src/agent_cli.rs#L643),
an output budget larger than the entire served window.

Independent corroboration that `/api/runtime/processes` is the wrong input: #1886 reports that on
Windows process discovery returns `Ok(None)` for every PID, so that source is empty there even for
local models (PR #1899). #1594 / PR #1688 is the precedent for a launcher and a node disagreeing
about a model id.

## Suggested fix

Take the window from the `/v1/models` response the launcher already parses, keyed by the id it
already extracted. That deletes both `/api/*` round-trips rather than adding one, and it works for
peers' models, which `/api/runtime/processes` cannot describe by construction.

- pi / opencode: use `metadata.context_length` instead of `merge_context_lengths`.
- goose: replace the literal with the same value.
- claude: set `CLAUDE_CODE_MAX_CONTEXT_TOKENS` in the `env` block, and derive
  `CLAUDE_CODE_MAX_OUTPUT_TOKENS` from the window rather than fixing it at 128000. Claude Code also
  reserves roughly 13k plus the output cap for auto-compaction, so a window under about 20k leaves
  it almost no room — worth warning at launch.
- For an id whose members differ, `metadata.max_context_length` is headroom only some nodes
  provide; `context_length` is the safe choice. The `mesh` virtual id reports no window at all, so
  it needs an explicit decision rather than an omitted key.

I have not tested whether the oversized window causes request rejections; I am reporting the
configuration mismatch, not a measured failure rate.

i386:
Agent launchers currently discard the served context window and reconstruct limits from native model metadata or fixed constants. Keep `metadata.context_length` beside each ID from `/v1/models` and pass that inventory directly to pi, OpenCode, Goose and Claude. For example, a model served at 16384 with native capacity 262144 now configures a 16384-token window.

Removes both management-API lookups. Missing/invalid metadata uses an announced 8192-token fallback; a `mesh` virtual model without its own limit uses the minimum concrete-model limit, including those fallbacks. Native capacity and maximum headroom do not override served capacity. Claude receives its context environment variable and an output cap of one quarter of the window, at most 4096, with a small-window warning.

Validation: full `mesh-llm-commands` package 265 passed; strict all-target Clippy and cargo check for commands and shipped `mesh-llm`, formatting and no-console-print passed. Tests cover all four launcher configurations, served versus native capacity, peer models, virtual mesh, invalid metadata and a real HTTP model-list fetch. Tests use an isolated writable Hugging Face cache; the first run encountered a permission error when an existing GPU fixture tried the machine's NFS cache.

Closes #1903.

Originating Buzz channel: a9ea5982-8d74-407f-b82e-6b55c69d411f.

## Summary by CodeRabbit

- **New Features**
  - Model context-window limits are now recognized and applied consistently across supported agent integrations.
  - Pi, OpenCode, Goose, and Claude configurations automatically use each model’s served context limit.
  - Model availability is validated against the discovered model inventory.
  - Fallback context limits are reported when model metadata is missing or invalid.
  - Claude output limits adjust based on the selected model’s available context window.
- **Bug Fixes**
  - Claude warns when the selected model provides less than 20,000 tokens of context.
  - Model-specific Claude settings now correctly apply to recognized Claude models.


michaelneale (dismissed):

crates/mesh-llm-commands/src/agent_cli/model_inventory.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn goose_models(&self) -> Vec<Value> {
+        self.names
+            .iter()
+            .map(|id| json!({"name":id,"context_limit":self.context_limit(id)}))
+            .collect()
+    }
+}
+
+pub(super) fn apply_claude_limits(settings: &mut Value, context: u32) {
+    settings["env"]["CLAUDE_CODE_MAX_CONTEXT_TOKENS"] = json!(context.to_string());
```
ndizazzo, line 75: One caveat here: CLAUDE_CODE_MAX_CONTEXT_TOKENS (v2.1.193+) only applies directly when the model ID is not recognized as a Claude model. If a mesh model ID starts with claude-, Claude keeps its built-in window unless DISABLE_COMPACT is set too. Let's set that alongside, or at least note the limitation in a comment so we know it can be a silent no-op.

crates/mesh-llm-commands/src/agent_cli/model_inventory.rs
```diff
@@ -0,0 +1,78 @@
...
+            .min()
+            .unwrap_or(UNKNOWN_CONTEXT_LIMIT);
+        for (id, limit) in &mut inventory.context_lengths {
+            if limit.is_none() {
+                let fallback = if id == "mesh" {
+                    mesh_limit
+                } else {
+                    UNKNOWN_CONTEXT_LIMIT
+                };
+                let _ = writeln!(
```
ndizazzo, line 48: This writes to the console from inside the parser. It only fires once per model-list fetch so it's fine in practice, but moving the fallback warning to the caller would keep from_response pure. Optional.

ndizazzo (changes requested):
Verified this against the shape /v1/models actually serves (context_length is the min across runtimes, models.rs), so the inventory parses real data instead of re-deriving it. I also checked the four launcher wirings. The direction is right, the min-for-mesh semantics and the invalid-metadata tests are good. Two small things to sort out, flagged inline.

crates/mesh-llm-commands/src/agent_cli/model_inventory.rs
i386: Fixed in ae68b34ff — DISABLE_COMPACT=1 is now set alongside, scoped to claude-* IDs only, since that is exactly where MAX_CONTEXT_TOKENS is a silent no-op; unrecognized IDs keep applying the served limit with compaction working at it. The limitation is documented on apply_claude_limits and both branches are covered by tests.

crates/mesh-llm-commands/src/agent_cli/model_inventory.rs
i386: Fixed in ae68b34ff — from_response no longer writes to the console. Fallbacks are recorded during parsing and callers invoke report_fallbacks() once after the inventory settles, on both the check_mesh and fetch_mesh_models paths.

PR #1916, opened 2026-09-18
test(host-runtime): keep the smart-auto StartNew tests off the network

Virgile-pct:
## Problem

Three `smart_auto` unit tests download the model catalog. `smart_auto_empty_starts_new`, `smart_auto_excludes_other_named_meshes` and `smart_auto_filters_full_mesh` are the three whose inputs leave no mesh to join, so `smart_auto` takes the `StartNew` branch and calls `model_packs::default_models_for_vram`, which calls `remote_catalog::ensure_catalog()`. Whenever the on-disk catalog is absent or older than a day, that is 115 entry files fetched from `huggingface.co/datasets/meshllm/catalog`, inside a test of a pure scoring decision. The four sibling tests take the `Join` branch and never touch it.

Measured alone on Windows 11: 27 s with a fresh `HF_HOME` (347 files written, two connections), 13 s with no `HOME` at all, which is what a `pwsh` runner looks like. Every CI run starts with an empty cache, so every run on every platform pays it, and a slow or unreachable Hugging Face turns three pure decision tests into a hang; on this machine, under memory pressure, they took over two hours before completing.

## Fix

Give the three an empty catalog through `set_catalog_entries_for_test`, the hook the resolve and catalog-route tests already use. With no tiers, pack selection falls back to the single default model plus the demand seeds, so `StartNew` still carries a non-empty list and the assertions hold unchanged. The override is process-global, hence `#[serial]` on the three, matching `smart_auto_sticky_preference` in the same module. One small helper documents why.

## Validation, on native Windows 11 against `main` at 62a55f24d

- `cargo test -p mesh-llm-host-runtime --lib smart_auto_tests` with a fresh `HF_HOME`: 7 passed in 0.05 s, and zero files written under it. Before the change the same run wrote 347.
- `cargo fmt --all --check` clean; `cargo clippy -p mesh-llm-host-runtime --all-targets` with the same warning set as `main`.

A note beyond this PR: `smart_auto` is a decision function, and `default_models_for_vram` refreshing a remote catalog under it is why `auto_join.rs` already wraps the call in `spawn_blocking`. This PR keeps the tests hermetic and leaves that shape alone.

## Summary by CodeRabbit

- **Tests**
  - Improved test reliability by preventing affected unit tests from making external network requests.
  - Added serialized execution for tests that share catalog state.
  - Covered scenarios involving mesh filtering, full meshes, and starting new meshes.


ndizazzo (approved):
Verified the branch split (these are exactly the three tests that assert StartNew; the four siblings panic on it and never reach the catalog), the guard restores the previous override on drop, and the empty catalog still leaves StartNew non-empty via the demand seeds. The #[serial] matches smart_auto_sticky_preference in the same module. Approving.

Issue #1915, opened 2026-09-18
host-runtime lib suite dies of STATUS_STACK_OVERFLOW on Windows in the /api/status test since #1887


Virgile-pct:
## Summary

On Windows, `cargo test -p mesh-llm-host-runtime --lib` does not fail, it dies. `api::tests::runtime_data_api_routes_remain_payload_stable` overflows the 2 MiB test-thread stack (`STATUS_STACK_OVERFLOW`, exit code 0xc00000fd), the harness process is gone with it, and the roughly 3150 tests scheduled after it never run. Linux and macOS pass it. It is a regression since #1887, localised to the `/api/status` route, and I have not found the frame; this issue carries every measurement so that whoever knows that path can.

## What is measured

Windows 11, `main` at 62a55f24d, debug profile, native MSVC toolchain.

- Deterministic. Alone, the test overflows every time at the default stack. With `RUST_MIN_STACK` it overflows at 2048 KiB and passes at 2304 KiB, so the peak is between 2.0 and 2.25 MiB.
- Regression. The same machine ran the whole suite at 950e90552 on the 14th, all 3472 tests, with no stack setting. The test file has not changed since May. Bisecting main between the two lands on 281c85fdb (#1887, typed activation frontiers): its parent acef2d391 is the last good commit and nothing else is in the range. It touches no host-runtime file.
- One route. Eleven single-request probes, one per route the test exercises, built once on `main`: the setup alone and ten routes pass at 2048 KiB; `/api/status` overflows and needs the same 2304 KiB the whole test needs.
- Only this test. With it skipped, a full pass at the default stack runs to the end: 3521 passed, 11 ignored, 378 s at 16 threads.
- What #1887 changed on that path. `skippy_runtime::ActivationBoundaryDesc` went from seven scalars, forty bytes, to a version, a part count, a 32-byte identity and sixteen inline `ActivationPartDesc` of 128 bytes each: 2088 bytes measured with `size_of`. `RuntimeStagePayload` in `api/status.rs` holds two of them by value, and so does the runtime stage status it is built from.

## Two fixes that looked right and are not

- Boxing the test's own future (`Box::pin` around the body). It made the overflow disappear, so I measured the future: 15 024 bytes. A 2.0 to 2.25 MiB peak moved by fifteen kilobytes; that is luck, not a fix.
[...]

Issue #1554, opened 2026-08-30
Proposal: add a Windows `unit` kind to the platform checks slice


Virgile-pct:
## Context

The Windows CI lane currently builds everything (host, native runtimes, products, Node SDK) but runs almost no tests: the `portable` check is `cargo check` only, and the `log-store` check runs two targeted ACL tests. The `unit` check kind that runs crate lib suites (`model-artifact`, `mesh-llm-host-runtime`, `mesh-llm`) exists for macOS only, and the broad rust-tests slice runs on Linux.

That gap is exactly where the four Windows bugs from this weekend lived (#1510, #1511, #1512, #1538): everything compiled, nothing was exercised. Two more data points make this the right moment:

- Since #1536, `skippy-server`'s lib tests compile (and run) on Windows at all, as noted in the review.
- Running them locally on a Windows machine surfaced #1538 within minutes, which suggests the suites have real detection value on this platform.

## Proposal

Mirror the macOS `unit` kind with a `windows` variant in `ci-platform-checks-slice.yml`, with a deliberately small initial scope:

```text
cargo test --locked -p skippy-ffi --features dynamic-runtime
cargo test --locked -p skippy-server --features dynamic-native-runtime --lib
cargo test --locked -p model-artifact --lib
cargo test --locked -p mesh-llm-host-runtime --lib
cargo test --locked -p mesh-llm --lib
```

Notes on the shape:

- **Dynamic link mode on purpose.** `skippy-ffi`'s build.rs states that automatic native preparation is not supported for Windows, so the initial slice avoids the static ABI archives entirely by testing through the dynamic-runtime features (the same configuration the shipped Windows host uses). A static-ABI Windows variant could come later on top of the existing `restore-windows-abi-cache` machinery, but it is not needed to start.
- **Sequencing around #1538.** `shutdown_tests::accept_error_still_closes_and_joins_active_connection_worker` fails deterministically on Windows today (that is issue #1538, production-relevant). To start the lane green, either the #1538 fix lands first (I am happy to send it once you have a preference between the two directions sketched in the issue), or that single test gets a temporary platform gate with a tracking comment.
- **Cost.** The kind would flow through the existing plan-ci matrices and windows budgets like the current `portable` and `log-store` checks, so it only runs when the plan asks for it.

## Offer

[...]

PR #1917, opened 2026-09-18
ci(platform-checks): add the Windows unit row's steps, inert until the catalog registers it

Virgile-pct:
## Problem

`ci-platform-checks-slice.yml` has a `unit` kind for macOS only. Its nine steps, from `brew install` to `Run macOS unit tests`, are gated on `kind == 'unit'` alone, and they run the `model-artifact`, `mesh-llm-host-runtime` and `mesh-llm` lib suites. Windows has the `portable` and `log-store` rows and runs no crate suite at all. That is the gap #1554 describes, and where every Windows bug of the last month lived: sixteen failures across six families on the 14th, none of them visible to CI.

## Change

- The nine macOS unit steps also require `matrix.check.platform == 'macos'`, so a Windows row of the same kind never reaches `brew install` or the Metal ABI build.
- One new step, `Run Windows unit tests`, gated on `kind == 'unit' && platform == 'windows'`, runs the same three suites with `--lib --locked` under `pwsh`, with `RUST_MIN_STACK` set to 8 MiB for the reason given under validation. The crates' default features select `dynamic-native-runtime`, the link mode the shipped Windows host uses, so no native ABI build is needed; the Python, sccache, MSVC and short-path steps the portable row already runs are shared. pwsh does not stop on a failing native command, so each cargo invocation is followed by an explicit exit check.
- The `ci-platform-checks-slice.yml` line in the manage-ci inventory names the new row.

This PR is inert on its own. Nothing runs until `ci/slices.yml` carries a `windows-unit` row, and a branch cannot add that row and pass its own Plan gate (`ci/ci.md`, "Catalog evolution"). The two-line catalog change, plus its planner test, is in the #1554 thread for one of you to land after this merges, the shape of #1898, so the row lights up against a workflow that already handles it. Landing the catalog first would run the macOS-only steps on the Windows runner and turn main red.

## Validation, on native Windows 11 against `main` at 62a55f24d

- The modified workflow parses, and every `unit`-gated step resolves to exactly one platform.
- The catalog change was applied to the real tree and `scripts/tests/test_plan_ci.py` run before and after: the same set of pre-existing failures, none added; the new `test_windows_platform_change_selects_portable_and_unit_rows` and the widened main-plan assertion pass.
- The three suites the step runs, with the exact environment the workflow exports (`MESH_LLM_SKIP_UI=1`, `LLAMA_STAGE_BACKEND=metal`, `LLAMA_STAGE_BUILD_DIR` pointing at a Metal tree that does not exist on Windows, which only build scripts read and the dynamic feature makes inert): `model-artifact --lib` 53 passed; `mesh-llm --lib` 51 passed; `mesh-llm-host-runtime --lib` 3532 passed, 4 failed, 11 ignored with an 8 MiB test-thread stack, the four being the targets of #1899 and #1900. With the default stack the host-runtime binary dies of a stack overflow in one test since #1887, which is why the step reserves an 8 MiB test-thread stack; the diagnosis, the bisection and two refuted fixes are in #1915. The network-free version of three `smart_auto` tests is #1916, sequenced before the catalog line.
- Timing, on this laptop pinned to four cores like the runner's 4 vCPUs, with #1915's test and #1916's three skipped: 10 min 39 s with 4 test threads, 10 min 2 s with 12, same results; 378 s unpinned at 16 threads. A first 4-thread measurement that still ran the three catalog tests took 40 minutes, which is what #1916 removes.

Not verified: the step itself on a hosted `windows-2022` runner, which by construction cannot happen before the row exists. The first main push after the catalog commit is that run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected platform-specific CI execution so macOS unit checks run only on macOS.
  * Added Windows unit-test coverage for key crate suites.
  * Improved Windows test reliability by configuring the required stack size and preserving test failure status.


Virgile-pct:
One thing the PR body could not claim and CI just did: because this changes a workflow, the planner opened the full matrix, so `Platform checks (macos-unit)` and `(macos-portable)` ran with the re-gated steps and passed, as did `windows-portable` and `windows-log-store`. The one red job, `Linux product (vulkan)`, died in `actions/checkout` on a connection reset from GitHub's side (curl 56, bad pack header); I re-ran the failed jobs and the lane is green, 69 checks.

ndizazzo:
@Virgile-pct re-kicked CI for the failing lanes

ndizazzo (approved):
Checked the head workflow against ci/slices.yml: with no windows-unit row registered, the new step is inert as described, and the Catalog evolution rule in ci/ci.md backs the sequenced-landing shape (land the catalog on main, then flip the row against a workflow that already handles it). All nine macOS unit steps now resolve to exactly one platform, the pwsh exit-checks after each cargo invocation are the right call, and the dynamic-native-runtime default-features claim holds for both touched crates. Approving.

PR #1918, opened 2026-09-18
ci(catalog): register the windows-unit platform row

ndizazzo:
Test PR for the Windows unit lane that #1917 added. Not intended to merge.

#1917 landed the "Run Windows unit tests" step in `ci-platform-checks-slice.yml`, but the step stays inert until the catalog registers a row that targets it. This PR does exactly that, and nothing more.

## The change

- `ci/slices.yml`: adds the `windows-unit` platform row (same shape as the existing Windows rows) and lists it in `platform_domain_rows["platform-windows"]` next to `windows-portable`. That mirrors the macOS portable/unit pair, so Windows platform changes (build scripts, `mesh-llm-nodejs`, `skippy-ffi`) also pick up the unit lane.

## Why this PR exercises the lane itself

The diff is catalog-only, so the planner maps it to the `ci-control` domain without the `docs` domain. That trips the control-plane fail-open selection, which schedules every platform row in this PR's own CI, including the new `windows-unit` row, on the Windows runner. The `Platform checks (windows-unit)` job is the test: it runs the `model-artifact`, `mesh-llm-host-runtime`, and `mesh-llm` lib suites in dynamic link mode with `RUST_MIN_STACK=8388608`.

@Virgile-pct could you review this and close it if the new lane worked as expected?

One thing I noticed along the way: `ci/ci.md` line 393 still says "Windows portable" where the #1917 inventory update says "Windows portable/unit". I left the doc out of this PR on purpose. Adding an `.md` path would map the change to the `docs` domain, make the plan docs-only, and suppress the fail-open that runs the lane. Happy to sync it in a follow-up.

## Summary by CodeRabbit

* **Tests**
  * Added Windows unit-test coverage to continuous integration checks.
  * Windows validation now includes both portable and unit-test checks.


Virgile-pct (approved):
Thanks for setting this up. The lane did not run here, and it cannot from any branch, so I would land this rather than close it.

**What happened.** Every Plan job on this head failed at `Build canonical plan` before any lane was scheduled, with `source slice catalog differs from the protected planner catalog`, and every lane was skipped; there is no `Platform checks (windows-unit)` check-run on the commit. That is the compare in `.github/actions/plan-ci/action.yml` (lines 187 to 193): the planner extracts `ci/slices.yml` and `ci/ownership.yml` from the source SHA and `cmp`s them byte for byte against the default-branch copies, and exits before the domain mapping runs. The fail-open you were counting on sits after that check, so a catalog-only diff never reaches it. `ci/ci.md` says the same under Catalog evolution: a branch cannot introduce its own slice entry and pass its own Plan gate.

**What does run the lane.** A push to `main`: the `main` profile has `all_rows: true`, so every platform row runs on every push, including this one. The only way to exercise `windows-unit` is therefore the shape #1898 took: merge the catalog commit with its own Plan red by construction, and the next push to `main` is the first run. This diff is exactly that commit, so I am approving it as the catalog change: same two lines I had ready, same placement next to `windows-portable`, and the planner suite on `main` needs no change for it (`test_main_covers_every_workspace_crate_once` asserts membership, `test_log_store_selects_only_the_windows_storage_privacy_row` is untouched by it; I ran `scripts/tests/test_plan_ci.py` on a real checkout before and after the change with the same result set).

**What the first run will show.** #1900 is not merged yet, so `mesh-llm-host-runtime --lib` on that first run will report exactly one failure, `unconfigured_strict_absolute_path_ignores_default_model_path`, the test #1900 fixes; everything else measured green here with the 8 MiB stack (3532 passed, 11 ignored on `main` at 62a55f24d, `model-artifact` 53, `mesh-llm` 51). Landing #1900 first makes the row green on its first run; landing this first makes the row's first run a live demonstration of what it is for. Either order works for me. I will watch that first `main_windows` run and report the numbers within the hour, and if it goes red for a reason only the hosted runner shows, the fix or a two-line disable is on me.

On `ci/ci.md` line 393: agreed, a docs-only follow-up, I can send it right after this lands.

ndizazzo:
@Virgile-pct #1900 was landed first, rebasing this one

Virgile-pct:
First run of the row, on `main` at 947222559 (run https://github.com/Mesh-LLM/mesh-llm/actions/runs/35401739071/job/105783008701): `Platform checks (windows-unit)` is green in 23 min 26 s on `windows-2022`.

- `model-artifact --lib`: 53 passed, 0 failed
- `mesh-llm-host-runtime --lib`: 3540 passed, 0 failed, 11 ignored, 208.77 s
- `mesh-llm --lib`: 51 passed, 0 failed

Build took about 19 minutes on a cold sccache (1002 compile requests, 2 hits), tests about 3.5 minutes. No stack overflow with the reserved 8 MiB, no catalog download in the suite, and no failure left since #1900 landed first. The lane works as expected; the `ci/ci.md` line is next.

PR #1919, opened 2026-09-18
fix(ci): give repaired llama candidates a full verification budget

i386:
The upstream canary now gives each repaired candidate a complete, bounded verification window, so time spent diagnosing an earlier failed candidate cannot prematurely terminate its replacement.

In [run 35324743175](https://github.com/Mesh-LLM/mesh-llm/actions/runs/35324743175), the first complete battery exposed native failures. After the agent reported focused repairs, the next trusted battery started with only 7,352 seconds left on the shared repair deadline and was killed in family 77. No candidate was certified or published.

The 11.5-hour window still bounds admission of coding turns and includes earlier gate time. Every returned candidate gets a fresh 12-hour prepare/build/certify budget. A failing pass after admission closes ends the task. Step/job limits of 1,420/1,430 minutes cover the resulting 23.5-hour maximum plus cleanup/upload headroom. The full roster, per-model limits, separate fresh-checkout verifier, and success-only publication are unchanged.

This fixes orchestration; it does not land or certify the native repairs described by the failed run's agent. No full hardware canary was dispatched from this branch: the workflow deliberately executes trusted main.

Validation:
- Deterministic shell regression reproduces the old timeout, completes the repaired second candidate with the new loop, and proves a failed final pass cannot admit another coding turn.
- 62 canary contract tests passed.
- ShellCheck and Actionlint passed.
- `just ci-validate` passed: all 1,510 script tests completed (nine expected skips), workflow lint, crate-list/release/console/publish consistency checks. The initial sandbox run blocked four existing localhost/process fixtures; the unrestricted run passed.
- `just ci-shellcheck scripts/llama-canary-agent-repair.sh`, `git diff --check`, and commit-subject validation passed.
- Local Rust linker emitted the existing Apple toolchain deployment-target warning while building xtask; no Rust product source changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved canary repair handling so each returned candidate receives a fresh, full verification window.
  * Preserved independent verification time even when earlier repair attempts or checks consume most of the repair window.
  * Ensured failed verification after the repair window closes ends the task without another coding attempt.
  * Increased workflow limits to accommodate repair, verification, and evidence reporting.

* **Documentation**
  * Updated CI guidance and runbooks to describe the revised repair and verification timing.

* **Tests**
  * Added coverage validating full verification windows across successive repair candidates.


ndizazzo (approved):
The fix is right and it's scoped to exactly the failure you describe. I verified the deadline split in `repair_candidate_until_green`: `REPAIR_DEADLINE_AT` is still set once at the top and still gates the loop, so coding turns are only admitted inside the 11.5-hour window and earlier gate time still consumes it. `VERIFICATION_DEADLINE_AT` no longer initializes to the repair deadline and resets to `now + VERIFICATION_TIMEOUT_SECONDS` after each `agent_session_step` returns, so each returned candidate gets a full fresh 12-hour gate budget even when the admission window is nearly exhausted. A failing pass after admission closes falls out of the `while remaining_repair_seconds` guard and the function fails without admitting another turn. The workflow numbers line up (41,400 s admission, 43,200 s verification, 23.5-hour maximum, 1,420-minute step, 1,430-minute job) and the four doc updates in SKILL.md, current-inventory.md, ci.md, and agent-repair-prompt.md all say the same thing the code now does.

The regression test is the part I was most worried about and it's good. It extracts the real `remaining_verification_seconds` and `repair_candidate_until_green` from the wrapper and runs them under a deterministic clock, so it tests the loop that actually ships rather than a copy. The first pass eats 80 of the 100-second admission window and fails, the second candidate still gets the full 200-second budget (under the old shared deadline it had roughly zero), and both subtests pin that no third coding turn is admitted after the window closes, with success and terminal-failure outcomes. That's the exact behavior from the run 35324743175 failure mode.

On CodeRabbit's 1,420-minute step-timeout finding: it cites a documented 360-minute step maximum, but the repo's own history contradicts it. This same step was `timeout-minutes: 720` before this PR, and the battery step next door is still 720 today. In run 35324743175 itself (pre-PR code), the agent task step ran 08:31:14Z to 20:01:20Z, 11 hours and 30 minutes, and completed without being killed at six hours. So 1,420 follows an established pattern that is demonstrably working on this runner, and the in-script deadlines remain the real bound with the job's 1,430 as the backstop. The rewrite it proposes (checkpoint and resume the Goose session across 360-minute step boundaries) would break the single-session model the whole harness is built on, and that's disproportionate to a cap that isn't being enforced here. One caveat worth knowing: if GitHub did start enforcing the documented cap, the failure mode is loud. The step would die at six hours, well before the 11.5-hour admission window expires, and the run would fail with no candidate. Since the workflow only ever executes trusted main, the next daily run after merge is the live confirmation.

One small note, not a blocker: the ten-minute margin between the 23.5-hour maximum and the step backstop assumes a single coding turn is short relative to the gate sequence. That's the same modeling the old 720-minute step did against the 11.5-hour deadline, and if a turn ever did run that long the step timeout kills it loudly instead of letting it run unbounded. I'd accept that.

CI contracts and consistency passes at head, which covers the script contract suite including the new test. The native runtime builds are still pending and unrelated to this change. Approving.

PR #1924, opened 2026-09-19
Add Lightning payments for mesh inference (PoC)

benthecarman:
## Original problem

Mesh inference currently depends on donated compute. Providers cannot advertise prices or receive payment for input processing and generated output through the ordinary local OpenAI-compatible API.

This draft adds a two-payment Lightning PoC: **request → prefill → approve and pay input → stream output → settle actual output**.

## Diagnostics

The implementation and failure-boundary review exercised wallet submission ambiguity, reservation leaks, restart recovery, unpaid delivery debt, cancellation during partial frames, and local-versus-forwarded spending. Regression coverage is included for those cases.

See [implementation notes and validation](https://github.com/benthecarman/mesh-llm/blob/add-lightning-payments/docs/specs/lightning-payments.md) and the [interactive explainer](https://files.benthecarman.dev/f/xl7ZH8GH/mesh-llm-lightning-payments.html).

### Commit review order

1. Wallet interface, Lexe adapter, and durable payment accounting.
2. Skippy prefill authorization and canonical output accounting.
3. Additive paid-offer metadata and compatibility fixtures.
4. QUIC payment flow, routing, recovery, and multi-node regressions.
5. Wallet CLI and trusted-local management API.
6. Design notes, validation evidence, and compatibility harness.

The history split preserves the previously validated final tree exactly.

- Add `mesh-llm-payments` with a provider-neutral `WalletProvider` trait and an initial Lexe mainnet/BOLT11 adapter. Async settlement observation supports future event-based adapters; Lexe polls internally.
- Advertise seller-set input/output prices and prefer local inference, then eligible paid providers, then free providers. Only locally originated requests may spend the wallet; forwarded requests and untrusted browser origins are rejected.
- Suspend Skippy after prefill until the seller independently observes input settlement. Invoice actual transmitted output at completion or cancellation.
- Support one manual approval per request or automatic payment with a daily budget. Payments add no output-token default or ceiling: caller limits and normal backend context settings determine the reserved maximum cost.
- Persist requests, reservations, invoices, payment outcomes, and provider debt in SQLite. Distinguish never-submitted failures from uncertain submissions, retain uncertain reservations without blind resubmission, and recover output debt after restart.
- Expose balance, funding, sending, transaction history, approvals, policy, and prices through the wallet CLI and local management API.

Gossip changes are additive. Legacy free inference remains supported; older clients cannot bypass a paid provider's payment gate. Existing payment records keep their original terms.

[...]

PR #1926, opened 2026-09-19
feat(payments): p2p pay for mesh inference with Lightning (earn and burn)

michaelneale:
Continuation of [#1924](https://github.com/Mesh-LLM/mesh-llm/pull/1924) on the same-repository branch `feat/lightning-payments`; original commit history and authorship preserved.

## scope notes

This is to introduce enough infra and end to end experience. Not over abstract or prematurely abstract for all kind of payments yet. 2 points is a line, 3 points is a trend, this is just 1 point to start with, to plant a flag. 

## Original problem

Mesh inference currently depends on donated compute. Providers cannot advertise prices or receive payment for input processing and generated output through the ordinary local OpenAI-compatible API.

This draft adds a two-payment Lightning PoC: **request → prefill → automatically authorize and pay input → stream output → settle actual output**.

## Diagnostics

The implementation and failure-boundary review exercised wallet submission ambiguity, reservation leaks, restart recovery, unpaid delivery debt, cancellation during partial frames, and local-versus-forwarded spending. Regression coverage is included for those cases.

See [implementation notes and validation](https://github.com/benthecarman/mesh-llm/blob/add-lightning-payments/docs/specs/lightning-payments.md) and the [interactive explainer](https://files.benthecarman.dev/f/xl7ZH8GH/mesh-llm-lightning-payments.html).

### Commit review order

1. Wallet interface, Lexe adapter, and durable payment accounting.
2. Skippy prefill authorization and canonical output accounting.
3. Additive paid-offer metadata and compatibility fixtures.
4. QUIC payment flow, routing, recovery, and multi-node regressions.
5. Wallet CLI and trusted-local management API.
6. Design notes, validation evidence, and compatibility harness.

The history split preserves the previously validated final tree exactly.

- Add `mesh-llm-payments` with a provider-neutral `WalletProvider` trait and an initial Lexe mainnet/BOLT11 adapter. Async settlement observation supports future event-based adapters; Lexe polls internally.
- Advertise seller-set input/output prices and prefer local inference, then eligible paid providers, then free providers. Only locally originated requests may spend the wallet; forwarded requests and untrusted browser origins are rejected.
- Suspend Skippy after prefill until the seller independently observes input settlement. Invoice actual transmitted output at completion or cancellation.
- Use free-only by default, or enable paid inference and automatic payment together with one daily-budget policy. No separate saved opt-in or per-inference approval. Payments add no output-token default or ceiling: caller limits and normal backend context settings determine the reserved maximum cost.
- Persist requests, reservations, invoices, payment outcomes, and provider debt in SQLite. Distinguish never-submitted failures from uncertain submissions, retain uncertain reservations without blind resubmission, and recover output debt after restart.
- Expose balance, funding, sending, transaction history, policy/usage, and prices through the wallet CLI and local management API.

Gossip changes are additive. Legacy free inference remains supported; older clients cannot bypass a paid provider's payment gate. Existing payment records keep their original terms.

The native wallet UI is in the separate [Mesh Tray companion branch](https://github.com/benthecarman/mesh-app/tree/lightning-wallet); its source is not part of this PR.

### PoC boundaries

Single-node text inference and BOLT11 only. Prefill and token accounting remain provider claims; no proof of correct inference, refunds, distributed paid inference, NWC, or encrypted output chunks are implemented. An unresolved wallet submission can retain its reservation indefinitely. Peer blacklisting does not prevent identity changes. These tradeoffs and the future invoice/preimage-encrypted token-chunk design are documented.

## Validation

- [x] Relevant local checks passed on the final working tree:
  - `just with-lld cargo test -p mesh-llm-payments --lib` — 34 tests.
  - `just with-lld cargo test -p mesh-llm-host-runtime --lib payments_` — 16 tests, including real two-/three-node QUIC exchanges with simulated wallets, cancellation, remote-spending prevention, omitted output limits, and 5,000-token settlement.
  - `just with-lld cargo clippy -p mesh-llm-payments -p mesh-llm-host-runtime --all-targets -- -D warnings -A linker_messages` — passed; the linker warning allowance covers the local mold-to-lld fallback.
  - `just build cpu`, formatting, and diff whitespace checks passed.
- [x] No native UI source changes are included here; the companion UI lives in the separate branch linked above. The explainer's interactions and desktop/mobile layouts were checked.

Earlier branch validation also exercised the native Skippy prefill gate, mixed-version free inference against v0.76.1, and rejection of legacy access to paid serving. Two isolated Lexe wallets completed real mainnet inference payments with a Skippy CUDA model. **The mainnet run predates the subsequent review fixes and output-limit removal; those latest changes were tested locally without spending funds.** Detailed evidence and the operator runbook are in the spec.

## Same-repository follow-up validation

The original description and historical validation above are preserved from #1924; they are not new test claims for this branch. Commit `06ec97b19` groups agent launcher dispatch without changing behavior to satisfy Clippy. Local formatting, no-console-print, `cargo check -p mesh-llm`, and `cargo clippy -p mesh-llm --all-targets -- -D warnings` passed for that change. Remote CI is being monitored; full local package execution was deferred to avoid macOS credential-store prompts.

Buzz origin channel: `1edb754c-7fa2-4940-af1f-9163bda48d89`.

## Evidence addition withdrawn — `560f7f1ac`

Reverted `5d3d8c372` in full; tracked tree exactly matches its parent `360979239`. Main integration through `d573029fb` and vetting removal remain. The lifecycle bus, task-local routing scope and schema-v2 migration are withdrawn. A smaller payer-side integration is being developed; no evidence-hook completion claim applies to this head.

Validation on the restored tree, committed unchanged: formatting, no-console-print, checks and warning-denying all-target Clippy for payments, host-runtime and shipped binary; full payments suite 47 passed; full host-runtime suite 3,719 unit plus 24 integration passed, 11 ignored. Mock-wallet tests only; no wallet databases changed or live nodes deployed.

## Smaller payer evidence hooks — `c2424878c`

Payer-side plugins can correlate active payment observations with the existing host OpenAI exchange. Optional `RequestTerms.exchange_id` persists in the existing JSON record; the recovery capability stays private. Explicit borrowed routing context replaces task-local state. `payment.lifecycle.v1` publishes terms accepted, input/output invoices, wallet terminal settlement and final debit including fees. The documented public terms projection uses the existing checked JCS digest helper.

No ledger event bus, snapshot queries, SQL schema change/migration, provider transport change, or per-token events. Subscriber-only setup uses an eight-event exchange queue and bounded background publication. This is deliberately payer-only: no provider claiming/event stream, restart replay, or lossless receipt claim. These limits and the exact schema are in the spec.

Replacement diff: 17 files, +383/-7, including tests and 38 documentation lines. Shared oversized routing files receive context plumbing only; event logic is in its own 254-line module including tests.

Validation on the working tree committed unchanged (committed patch verified identical): formatting, no-console-print, package checks; all-target Clippy with warnings denied for payments, host-runtime and shipped binary; full payments suite **48 passed**; full host-runtime suite **3,722 unit + 24 integration passed, 11 ignored**. Added persistence/legacy JSON, privacy, stable refs, bounded queue, terminal-only settlement and two-node QUIC/mock-wallet payer correlation coverage. No live models or wallets used, no latency benchmark, and no new remote CI claim. Build artifacts cleaned (26.4 GiB).

## Summary by CodeRabbit

* **New Features**
  * Added Lightning payment support for paid inference, including wallet management, durable settlement, invoice handling, recovery, and spending policies.
  * Added wallet CLI commands and a local wallet API for balances, transactions, funding, payments, policy, and pricing.
  * Added advertised model pricing and payment-aware routing, with pricing details available in model listings.
  * Added payment offers to peer announcements and support for free-only or automatic payment policies.
* **Security**
  * Restricted wallet operations and paid requests to trusted local or authenticated origins.
* **Documentation**
  * Added Lightning payments specifications and crate documentation.

### Single-policy CLI setup

```sh
mesh-llm wallet policy --mode automatic --daily-budget-sats 100
mesh-llm wallet policy
mesh-llm wallet policy --mode free-only
```

Automatic policy alone permits paid providers when available, within the daily budget and wallet funds. Free providers remain eligible. Funding is separate. Inspection returns `mode`, `daily_budget_msat`, `spent_today_msat`, `reserved_msat`, and `remaining_daily_budget_msat` as JSON. For example, a newly enabled 100-sat policy with no activity reports 100000 remaining msat. Free-only prevents new paid inference without cancelling owed settlement. The old manual mode and approve/reject controls are removed; existing experimental profiles must be explicitly configured with a supported policy. No compatibility migration was added.

Local validation of this change: payments (50 tests), CLI suite, full host suite (3721 unit + 24 integration; 11 ignored), compile checks and warning-denying all-target Clippy, formatting, no-console-print, CLI inventory. Commands-crate runtime tests are deferred to CI because its full suite accesses macOS Keychain. Two-machine real-wallet validation is pending.


michaelneale:
Copied from the original PR for continuity — original: https://github.com/Mesh-LLM/mesh-llm/pull/1924#issuecomment-5739189951

> **@ndizazzo** on #1924 (2026-09-19):

> hey @benthecarman - thanks for opening up this PR! Lightning / crypto-for-inference is something that @michaelneale and @i386 and myself have been chatting about for a while now, but has been lower on the priority list because of how we'd like for it to fit into the project.
> 
> Off the bat - this is something we _**do**_ want to support for mesh operators.
> 
> However, there are several work-streams that are being worked on that would dramatically change the shape of what your lightning work could look like:
> 
> 1) @StevenMih is building an evidence-of-inference system with #1332 - this evidence would link cleanly to receipts in Lightning
> 2) We recently exposed much more capability to application [Plugins](https://github.com/Mesh-LLM/mesh-llm/blob/main/docs/plugins/README.md#plugins) to allow for UI additions, inference stream hooks, etc. stuff that would allow you to build wallet features cleanly into the system
> 3) We are extending the mesh governance capabilities - payments could/should be a policy a hosted mesh permits
> 
> #3 is important because we wouldn't want somebody to join the public mesh, advertising as a peer, but holding up inference behind payments for clients who:
> 
> 1) Don't support it
> 2) Don't expect it
> 
> With mesh policies, it's clear when clients join what the requirements, rules, and expectations for participating clients are - which lightning payments could be one of.
> 
> ---
> 
> Here's a rough split of how I've personally been thinking about how crypto payments can be layered into MeshLLM:
> 
> ```
> CORE HOST
>  -> generic provider-admission hooks
>  -> bounded output-meter enforcements
>  -> lifecycle events and trusted plugin side streams
>  -> owner/node identity verification and revocation (could solve your peer identity issue in the PoC)
> 
> PAID-INFERENCE PLUGIN
>  -> Lightning wallet, quotes, invoices, settlement, ledger, pricing, paid offers
> 
> CAPSULE EMIT PLUGIN
>  -> signed request/route/response receipts, linked to payment terms
> 
> MESH POLICY
>  -> decides which owner-attested providers and evidence levels are eligible
> ```

michaelneale:
on the points above - I think we are close to that already. I don't think plugin is suitable yet for this in this form (as this ties deeply into inference and IPC boundaries, and isn't an external provider etc). 2 points is a line, 3 is a trend, and we only have 1 point yet so am trying to avoid premature abstraction. 

we can include some basic proof of inference for a target provider which helps to start, and then in time add the sealed capsules that only strengthen things. 

I also think mixed free/paid should be baked in from the start, always.

crates/mesh-llm-host-runtime/src/network/payments/vetting.rs
```diff
@@ -0,0 +1,611 @@
...
+pub(crate) async fn verify(node: &Node, peer: iroh::EndpointId, model: &str) -> Result<()> {
+    let service = node.payment_service().await?;
+    let policy = service.ledger.vetting_policy()?;
+    if !policy.required {
+        return Ok(());
+    }
+    let id = peer.to_string();
+    if service
+        .ledger
+        .provider_vetted(&id, mesh_llm_payments::now_ms(), policy.ttl_ms)?
```
ndizazzo: [P1] Key the vetting cache by model

`verify` passes a model into the probe and stores it, but this cache lookup only keys on `provider_id`. After model A succeeds, every paid model from the same provider is treated as vetted for the TTL, so `required=true` no longer checks the selected model. Include the requested model in both cache lookups and retain separate provider/model records; please add a two-model regression.

crates/mesh-llm-host-runtime/src/network/payments/server.rs
```diff
@@ -0,0 +1,338 @@
...
+    writer: &mut (impl AsyncWrite + Unpin),
+) -> Result<()> {
+    uuid::Uuid::parse_str(id)?;
+    // The random request ID is a bearer recovery capability, allowing recovery
+    // when the client's ephemeral mesh identity changes after restart.
+    let (_, _, _, finished) = service.ledger.serving_account(id)?;
+    if !finished {
+        return wire::write(writer, &Frame::Pending).await;
+    }
+    if let Some(receipt) = service.output_receivable(id).await? {
```
ndizazzo, line 346: [P1] Do not recover output debt before input settlement

The normal serving path waits for `await_input_settlement()` before calling `output_receivable`, but recovery reaches this branch using only `finished`. A crash after `wait_arrival` reports `claiming` and before the segment-0 invoice is terminal lets recovery send and charge segment 1 while input may still fail; the input receivable also remains unreconciled on restart. Recover or verify the segment-0 receivable first, and keep output debt pending until it is terminally paid.

crates/mesh-llm-payments/src/ledger.rs
```diff
@@ -0,0 +1,517 @@
...
+        );
+        Ok(())
+    }
+}
+
+fn finalize_terminal_requests(connection: &Connection) -> Result<()> {
+    connection.execute("UPDATE charges SET state='failed' WHERE state='prepared' AND EXISTS(SELECT 1 FROM charges sibling WHERE sibling.request_id=charges.request_id AND sibling.state='failed')", [])?;
+    // Failure revokes unused authorization, but cannot release any uncertain
+    // sibling charge. A send has only one charge and can close on success.
+    connection.execute("UPDATE requests SET state=CASE WHEN EXISTS(SELECT 1 FROM charges WHERE request_id=requests.id AND state='failed') THEN 'failed' ELSE 'completed' END WHERE state='approved' AND NOT EXISTS(SELECT 1 FROM charges WHERE request_id=requests.id AND state IN ('prepared','pending')) AND (EXISTS(SELECT 1 FROM charges WHERE request_id=requests.id AND state='failed') OR (json_extract(terms,'$.peer')='wallet-send' AND EXISTS(SELECT 1 FROM charges WHERE request_id=requests.id AND state='succeeded')))", [])?;
```
ndizazzo, line 576: [P1] Release abandoned approvals with no charge

This restart repair never closes an approved inference request that has no charge. `approve()` commits before `prepare_charge()`, and the cancellation branch can call `reject()` after approval wins; a crash or race in that gap leaves state=`approved` forever. `available_budget()` continues subtracting the cap, while the operator surface can reject only pending requests and has no inference finish command. Reconcile or expire these zero-charge approvals on recovery, without releasing prepared, pending, or uncertain charges.

ndizazzo (dismissed):
Request changes. I found three payment-state defects at this head that can bypass required model vetting, charge output before input settlement, or permanently reserve inference capacity after an approval race or crash. Focused tests pass, but no required CI checks are reported for this draft PR.

crates/mesh-llm-host-runtime/src/network/payments/vetting.rs
michaelneale: no - that is not part of this design, it is for provider, won't be doing that path as part of this inital PR but a larger future proof of LLM work.

StevenMih:
@michaelneale  
Two small asks from the evidence side, both additive: 
(1) one correlation id — carry exchange_id on RequestTerms (or make RequestTerms.id the exchange_id) so payment and evidence never have to reconstruct the relationship afterwards; 
(2) publish payment lifecycle events on the plugin event stream keyed by that id, digests and refs only — terms accepted, input invoice issued/settled, output invoice issued/settled, final msat — never a preimage. 
With those, a plugin can bind request → terms → delivery → settlement into one signed record with honest labels ("provider claim", "wallet-reported"), which is what makes payment receipts and evidence receipts composable rather than interchangeable. 
Nothing else in the PR needs to change for that.

StevenMih:
Saw the correlation/observation commit and the revert — understood on size. Two things so it stays cheap to land after this merges: (1) the reverted spec section is the contract we'd build against as-is (payment.lifecycle.v1, the four phases with local_authorization / provider_claim / wallet_reported, claiming never = settled, the tagged-tuple terms_digest, event_id dedup, best-effort buffer) — please keep that text somewhere it can be cited, and I'm happy to carry it as the follow-up PR from the reverted commit with authorship intact once #1926 is in. (2) If you're open to it, the correlation half alone is ~70 lines (RequestTerms.exchange_id, the x-mesh-payment-exchange-id payload header, the schema v2 column) and is the part that can't be added retroactively — records written without it have no correlation forever. The event stream can wait; the id can't. Your call on whether even that is too much for this PR.

benthecarman:
Did some review of the bitcoin stuff for security things too main things summarized by claude

-  fail unsent input charges on startup instead of paying them. In PaymentService::with_factory (crates/mesh-llm-payments/src/service.rs:67), after `close_interrupted_serving` and `close_abandoned_approvals`, add a ledger step that moves every charge with `segment = 0` and `state = 'prepared'` to failed, then run `finalize_terminal_requests` so the parent request closes and its reservation is released. A prepared charge is by definition one where `begin_submission` never ran, so the wallet was never called and nothing is lost by failing it. Leave pending charges alone, since those may have been submitted. Leave prepared segment-1 (output) charges alone too, since the seller has delivered and is owed. Then reconcile_pending naturally only resumes output charges. Update the test `prepared_payment_can_resume_without_resubmitting_started_payment` to use segment 1, and add a sibling test asserting a prepared segment-0 charge is failed on reopen with zero wallet calls and the request's reservation released.

- give invoices a short explicit expiry. In the Lexe adapter (crates/mesh-llm-payments/src/lexe.rs:186) set `expiration_secs` to a few minutes rather than relying on the SDK's 24-hour default, and make the seller's wait a separate, shorter timeout rather than reusing invoice expiry. That way, even if some other path pays late, the seller's node rejects the HTLC because the invoice has expired, and the buyer's payment fails instead of vanishing. This also fixes the day-long slot pinning and the permanent blacklist on expired receivables.

the first is mostly fixed by a shorter invoice expiry too, but would still be good to do.

for the invoice expiry, we should just make it match however long we keep the KV cache populated? Also this should be set outside the lexe impl so we make sure to pass it down to all consumers of the trait.

Some more minors below:

- The seed is plaintext in `seedphrase.txt`, mode 0700 directory. We should probably do 0600
- Fee allowance is a flat 1 sat. `FEE_ALLOWANCE_MSAT` is `1000`. We should bump this or just remove it and let the wallet defaults use this
- The claiming gate depends on a human-readable string. `is_claiming` matches `status_msg == "claiming"` (`crates/mesh-llm-payments/src/wallet.rs:60`). The SDK documents `status_msg` as a per-type display string, and the value comes from the node, not the SDK. If it differs the gate degrades to waiting for `completed`, which is safe but slower. Opening delivery on claiming is otherwise sound: LDK only emits `PaymentClaimable` once the HTLC is irrevocably committed.
- Recovery broadcasts the bearer request ID. `payment_recovery.rs:24` sends `Recover{id}` to every peer advertising the model. The payee pin stops third parties from getting paid, but a peer answering `Pending` ends that round (`payment_recovery.rs:84`). Low severity, but restrict to the original payee's peer.

crates/mesh-wallet-lexe/src/lexe.rs
```diff
@@ -0,0 +1,429 @@
...
+                ms = started.elapsed().as_millis() as u64,
+                "wallet lookup"
+            );
+            if let Some(payment) = payment
+                && settled(&payment)
+            {
+                return Ok(payment);
+            }
+        }
+    }
```
MaxFangX, line 379: This super aggressive polling keeps Lexe user nodes online, which puts a ton of load onto Lexe infrastructure. This will quickly result in requests failing due to hitting our rate-limits. Instead, please see my branch which updates the Lexe integration to be a lot more polite.

MaxFangX (dismissed):
Hey @michaelneale @benthecarman! Amazing work on this. I'm from the Lexe team, and I thought I'd drop by and help review + fixup the code that interfaces with our SDK.

The current code polls our user nodes super aggressively (once every 200ms) and keeps user nodes online 24/7, which basically guarantees that you'll get rate-limited, as it is indistinguishable from a DoS attack. I've pushed a [branch](https://github.com/MaxFangX/mesh-llm/tree/max/2026-09-22-lightning-payments-with-fixes) which includes a [commit](https://github.com/Mesh-LLM/mesh-llm/commit/ca1cdcad411d2a673eafe9741a8eb335f2e13a97) which fixes these issues - I suggest you cherry-pick it onto this PR.

Overall, using our `unstable` APIs is a bad idea. Instability aside, these bypass the safeguards that help keep you from running into our rate-limits. Instead, prefer to use our stable APIs which do robust caching underneath.

The fix I pushed creates one watcher per wallet, primarily using Lexe's stable APIs. It polls for updates using `LexeWallet::get_updated_payments`, and does lookups via `LexeWallet::get_payment`, both of which do smart caching and hit Lexe's gateway, so that user nodes stay asleep until there is actual payment activity. There are lots of tests, and the `wallet.v1` surface is unchanged.

Seeing the MeshLLM integration code was great, as it helped us identify gaps in our SDK. In response to the needs we see in this PR, we'll update our SDK shortly to include:
- `*_preflight` endpoints, so you can see the exact fee you're about to pay before actually paying it
- The ability to set a maximum fee found on the route during preflight
- Ability to lookup payments by payment hash
- Strong payment idempotency guarantees (it's safe to pay the same invoice twice, retries will return the existing payment result, client payment ids for idempotency when paying offers).
- Better API surfaces for handling multiple concurrent watchers

Once these SDK changes are in, it should be possible to switch the MeshLLM integration to exclusively use stable LexeWallet APIs.

Cheers!
Max

crates/mesh-wallet-lexe/src/lexe.rs
michaelneale: @MaxFangX ah right - yeah thanks for bring up - I think this may have been me trying to work out the quickest time to know settlement, thanks for changes - bringing them in now

michaelneale:
hopefully in better shape, sorry about that @MaxFangX !

MaxFangX (commented):
Looking good, thanks for accepting the patch!

i386 (approved):
**Verdict: Approve** — independent source-level security walkthrough plus local reproduction at head `600b5c81`.

## Reproduced locally (macOS, stock toolchain)

- `cargo test -p mesh-llm-payments --lib` — **69 passed / 0 failed**
- `cargo test -p mesh-llm-host-runtime --lib payments_` — **16 passed / 0 failed** (incl. two-/three-node QUIC, cancellation mid-frame, remote-spending prevention, cross-site loopback denial)
- `cargo clippy -p mesh-llm-payments -p mesh-llm-host-runtime --all-targets -- -D warnings` — clean
- `cargo fmt --check` across the four touched crates — clean
- `cargo test -p skippy-server --lib frontend` — 543/0
- CI green at head (Linux Rust-test batches, native runtimes, product smokes)

## Security walkthrough (verified in source, not just tests)

1. **Local-only spending.** `paid::route` requires loopback peer address AND not-a-remote-bridge (process-owned socket registry — not forgeable via HTTP headers, fails closed when poisoned) AND trusted Origin/Host headers. Forwarded/browser-origin requests get 402 before wallet or peer access. `strip_intent` removes `mesh_payment` before any forwarding, and `PaymentIntent::restrict` can only tighten the saved profile (tested both directions).
2. **No seller-gate bypass.** A priced model forces the paid path in `route_remote_attempt`; the payment upgrade rides the admitted-peer QUIC tunnel only; legacy TCP bridges from payment-advertising or wallet-enabled nodes are refused (402, `legacy_bridge_requires_payment_ingress`); the seller re-verifies the request against its own ledger pricing; backend model resolution stays inside locally-served targets; and the generation gate is keyed by a host-generated UUID that replaces caller-supplied request IDs (caller identity headers stripped).
3. **No double-billing.** A failed paid attempt is terminal for the HTTP request — transport retries cannot create a second bill. Invoice terms are re-derived locally (request id, price identity, token bounds, cap arithmetic recomputed as a pure function on both sides, expiry, fixed-amount only, payee pinned to the authenticated QUIC peer). Output invoices must keep the same payee and recompute the amount from the caller's own token limit.
4. **Submission ambiguity is durable.** NotSubmitted vs Uncertain are distinct ledger states; resubmission is idempotent per payment hash (Lexe SDK ≥ 0.1.24); reservations survive restart and are released only when expiry proves nothing was paid; startup fails unsubmitted input charges instead of paying them (the earlier review finding); the data directory holds an exclusive process lock.
5. **Durable budget.** The authoritative reservation is `ledger.approve` inside a SQLite IMMEDIATE transaction (cap ≤ balance − reserved, policy re-checked in-transaction); routing-time balance checks are advisory and fail closed (wallet errors read as zero available). Policy defaults to free-only; automatic mode requires a positive daily budget.
6. **Arithmetic.** u128 / `div_ceil` / `checked_add` throughout pricing and fee allowance; overflow produces errors, never wrapped values.
7. **Gossip stays additive.** `lightning_offers` is an additive wire field; decode validates count/model length/price/duplicates and rejects malformed offers instead of silently treating them as free.
8. **Wallet plugin boundary.** Wallet identity pinned after first open and re-open refused on mismatch; money-moving calls never blindly retried; settlement waits carry no IPC deadline; the Lexe adapter re-asserts seed permissions on every open, fsyncs the seed before provisioning, and the watcher polls only while someone waits.

## Non-blocking notes (no action requested in this PR)

- `transport_route_model.rs` uses `completion_tokens.unwrap_or(256)` as the routing-time output estimate. With `max_tokens` omitted, the actual settlement allowance is the backend context, so a large-context request can pass routing and then correctly fail at `ledger.approve` with the generic "payment was not authorized" 402. Fail-closed, but a budget-specific message would save operators a log dive.
- Up to `PRE_PAYMENT_OUTPUT_TOKENS` (512) tokens decode ahead of the input payment, and an unpaid request can pin a seller backend slot for the invoice lifetime (60s). Both are documented tradeoffs in the spec — flagging so they're merged consciously.
- The PoC-boundaries section matches what the code actually does (provider-claims accounting, no refunds, unresolved submissions can retain a reservation indefinitely, identity churn defeats blacklisting). Thanks for keeping that section honest, including the note that the mainnet run predates the later fixes.

PR #1935, opened 2026-09-20
feat(split)!: --auto-balance speed-balanced layer placement with closed-loop rebalancing

danielwinterw:
Splits place layers by memory today, so on unequal nodes the slower one holds
half the model and paces the whole pipeline. This adds `--auto-balance`: place
the cut by measured node speed, and keep re-balancing it from the split's own
per-stage busy time while serving.

Two engine fixes come with it, because the placement change is worthless
without them (see the numbers).

## What is here

- **Per-stage compute-busy time** in stage status (`compute_busy_nanos`,
  `compute_operations`), metered on the single scheduler worker thread so
  batched lanes are not double counted.
- **Speed-balanced placement** — an exact min–max contiguous partition (DP)
  over layer weight bytes ÷ node decode rate, with memory caps still enforced.
- **Closed-loop rebalancing** — rolling windows over cumulative stage busy time
  propose a new cut, drain in-flight requests, load and cut over, then keep the
  move or roll it back on the next window's throughput. Damped to half-steps,
  with hysteresis and a cooldown.
- **Two scheduling fixes** behind env switches: the last stage never batched
  decode across lanes (`SKIPPY_LAST_STAGE_DECODE_BATCH=1`), and stage 0's lanes
  can run as alternating groups (`SKIPPY_PIPELINE_DECODE_GROUPS=N`).
- **A transport fix** found while testing: the stage-transport accept loop
  exited on its first error, so after a cutover the new generation's bridge
  refused every connection. It now logs, backs off and keeps accepting. The
  open-file limit is also raised at startup — fd exhaustion was the original
  trigger.

## Measured

Two Mac minis (M1 16 GB, M4 16 GB) on routed gigabit, Qwen3-8B Q4_K_M, 16
streaming requests held in flight for **one hour each** — 55 consecutive
60-second windows, every token timestamped into its window.

| Configuration | tok/s | ÷ unsplit pair |
|---|---|---|
| No split: M4 (20.71) + M1 (13.85) | 34.56 | 1.00× |
| Balanced cut alone (12/24) | 15.90 | 0.46× |
| Engine fixes alone (18/18) <sup>10 min</sup> | 19.92 | 0.58× |
| Stock split (18/18) <sup>quarters 2–4</sup> | 20.6 | 0.60× |
| **Balanced cut + engine fixes** | **29.16** | **0.84×** |

Flat across all four quarters (28.6 / 28.8 / 28.8 / 30.2), zero errors.

The interaction is the point: **neither change pays off alone.** The balanced
cut on its own is *worse* than the stock split, because moving 24 of 36 layers
onto the last stage only helps once that stage batches decode across lanes
instead of running them one at a time. Earlier 1-minute windows credited the
two changes separately; the hour-long runs show that was noise.

Placement picks the same cut the profile implies: stage 0 measured 4.5 ms per
layer per batch against stage 1's 2.3 ms, and 4.5x = 2.3(36 − x) gives x ≈ 12.
Static estimates do not — GPU memory bandwidth suggests 13/23 and standalone
speed suggests 15/21, which measured 24–25 tok/s.

## What it costs per request

Throughput per machine is what this buys; per-request latency is what it
spends. Measured from every completed request in both hours, at matched load —
16 in flight on the split, 8 per machine unsplit:

| | e2e p50 | p90 | p99 | tok/s per request |
|---|---|---|---|---|
| No split, pooled over both machines | 211s | 308s | 397s | 2.19 |
| **Split, 12/24 + A + B** | **264s** | **318s** | **376s** | **1.85** |
| Split, stock | 343s | 415s | 654s | 1.59 |

**A request takes ~25% longer and generates at 0.85× the rate.** That 0.85
tracks the 0.84× aggregate parity, as it should: the same number of requests
are in flight either way. The balanced cut improves the per-request figure too
— stock is 0.73×.

The tail is the surprise: **p99 is lower on the split** (376s against 397s),
and the spread is tighter (p90/p50 of 1.21× against 1.46×). Unsplit, the pair
is bimodal — a request served by the faster machine finishes in 188s, by the
slower one in 264s, 1.41× apart — because each machine runs the whole model at
its own speed. A split blends both machines into every request, so the fleet
answers at one rate instead of two.

Two honest limits on those numbers:

- **Part of the 25% is lane budget, not the hop.** The split served 16 in
  flight with 4 lanes; the pair served 16 with 8 lanes total. Separating
  queueing from the hop needs an 8-lane split, which does not fit in memory on
  these machines.
- **The harness records no TTFT**, so e2e here is queue time plus generation
  combined. How much of the 53s median gap is waiting for a lane rather than
  slower generation is not measured.
- **The per-request column is not generation speed.** It is completion tokens
  divided by wall-clock, and with more requests in flight than lanes a request
  spends much of its life queued. It equals aggregate ÷ in-flight to within 2%
  in every run, which is the arithmetic you would expect, not a second
  measurement.

Worth knowing before quoting any absolute figure here: a single stream on the
faster machine measured 18.04 tok/s, and running eight at once yields 20.70
tok/s *in total* — concurrency buys ~15% more aggregate while each request runs
~7× slower, so decode is barely batching on this engine and hardware. The split
reaches 7.30 tok/s per lane against that machine's 5.17, which is the win: two
machines' memory bandwidth on one stream of work. These absolute numbers
characterise the engine at this concurrency, not what the hardware can do; the
parity ratio holds because both arms run the identical binary.

If single-request latency is what matters, this is the wrong trade: route to
the fastest machine and take 188s. `--auto-balance` is for work where the
fleet's tokens per second matter and individual requests can wait.

## Known gaps, for follow-ups

- **The memory-first initial cut is unsafe and needs a headroom guard.** With
  `MESH_LLM_AUTO_BALANCE_INITIAL_CUT=memory` the planner filled the first node
  to its budget and chose 35/1. Within 8 minutes that node's RSS reached
  12.0 GB with 10% memory free, it stopped heartbeating, and the split lost the
  stage. `--max-vram` under-counts: it covers weights and KV but not per-lane
  compute buffers (~230 MiB each), prefill spikes or the frontend. The default
  speed-based path does not hit this — it chose 12/24 immediately — but the
  planner should refuse a cut that leaves a node without headroom.
- **Group and lane count are static.** 6 lanes / 3 groups measured 27.66
  against 4 lanes / 2 groups at 29.16, so the concurrency shape matters about
  as much as the cut, and the controller cannot touch it. Groups are cheap to
  change at runtime (one field on the scheduler worker, no drain), so they are
  a good candidate for the same propose/accept/rollback loop — though that
  needs interleaved A/B, since window-to-window sd is ~8% and the effect is
  ~5%. Lanes are not: changing them reallocates the KV pool, and raising them
  under load is how you OOM the node.
- The env switches should become config once their defaults are settled.

## Summary by CodeRabbit

- **New Features**
  - Added an optional `--auto-balance` setting for split deployments, requiring `--split`.
  - Automatically adjusts layer placement using node decode speed and observed stage performance.
  - Added throughput estimates and stage utilization reporting.
  - Added configurable pipeline decode grouping and optional final-stage batching.
- **Bug Fixes**
  - Stage transport listeners now retry transient connection-accept errors with adaptive backoff.
  - Improved decode batching behavior while preserving native multi-token prediction support.

## Protocol

`BinaryStageOptions` (public in `skippy-server`) gains a `compute_meter:
Option<Arc<StageComputeMeter>>` field used by mesh-launched stages for
auto-balance accounting. This is a **deliberate source-breaking change**:
downstream struct literals must add the field (use `None` to keep the old
behavior). The meter feeds the auto-balance controller only; `None` skips
the accounting.

## Review fixes (9e953a9fe)

All six blocking findings from ndizazzo's review, plus two of i386's
non-blocking suggestions:

- Stage-0 compute meter is registered only after every fallible setup step,
  so failed loads leave no stale entry (`eff292eb6`).
- A pending auto-balance trial is cancelled when the topology changes
  mid-trial, with a regression test (`749535e2c`).
- `note_rolled_back` fires only after a successful cutover; a failed
  rollback keeps the trial open for the next sample (`21baee018`).
- The drain acknowledgement wait is bounded by the cutover deadline
  (`8146d7`-series, `8146d1d7b`).
- The compute meter now measures lock-held runtime work only; scheduler
  lock contention is excluded (`086bf759c`).
- `BinaryStageOptions.compute_meter` kept as the breaking change documented
  above.
- Placement logs now report `auto_balance_requested` / `auto_balance_applied`
  and warn when `--auto-balance` falls back to memory-only placement;
  `--auto-balance` requires `--split` (`9e953a9fe`).

Deferred (non-blocking, from i386): partial-window trial judgment, meter
teardown on generation replacement, accept-loop backoff shaping, per-lane
buffer estimate in the capacity model.

Updated with `main` (3f4eb9684).


crates/mesh-llm-host-runtime/src/runtime/local_split/auto_balance.rs
```diff
@@ -0,0 +1,529 @@
...
+        }
+        let measurement = measure(last, &sample, elapsed);
+        self.last = Some(sample.clone());
+
+        if measurement.decode_tokens_per_second < self.config.min_decode_tokens_per_second {
+            self.imbalanced_windows = 0;
+            return AutoBalanceDecision::Hold { why: "idle" };
+        }
+
+        if let Some(trial) = self.trial.take() {
```
i386, line 189: The trial window is judged on the first full window after the move regardless of how representative it is. The idle check above protects the zero-load case, but a **partially** loaded window can still false-trigger rollback: e.g. baseline 21 tok/s, post-move window is 30 s idle + 30 s at 30 tok/s → measured ~15 tok/s < 18.9 floor → rollback of a move whose steady state is better than baseline. Each rollback is a full drain + cutover, so bursty traffic right after a move causes churn.

Non-blocking suggestion: only judge the trial when the window's decode rate is at least some fraction of the baseline (say ≥50%); otherwise extend the trial one window. The hour-long constant-load runs never exercise this, which is why it hasn't shown up.

crates/mesh-llm-host-runtime/src/inference/skippy/mod.rs
```diff
@@ -1476,6 +1478,39 @@ fn now_unix_nanos() -> i64 {
         .unwrap_or(0)
 }
 
+/// Stage-0 compute meters of running split generations, keyed by run id, so
+/// the split coordinator can read the local stage's busy time alongside the
+/// peer stages' reported status.
+static STAGE0_COMPUTE_METERS: std::sync::LazyLock<
```
i386, line 1488: Minor leak: entries are only forgotten on a successful auto-balance cutover (`cut_over_to_boundaries`). Every `SkippyModelHandle` load registers here — including plain single-node local serving with no split — and the other teardown paths never remove their entries: stage-loss replan `Replace`, `Withdraw`, local fallback, and model stop/reload. Each entry is small (run_id string + one meter Arc), so this is bounded by model churn, but on a long-lived node it grows without bound.

Suggestion for a follow-up: drop the entry when the generation's handle is torn down (e.g. where `SplitGenerationCleanup` runs), rather than only on cutover.

crates/mesh-llm-host-runtime/src/runtime/split_planning.rs
```diff
@@ -238,6 +248,9 @@ pub(super) fn plan_runtime_slice_topology_with_resources_and_stage0(
         estimated_decode_network_ms_per_token = plan.estimated_decode_network_ms_per_token,
         decode_tpot_target_met = plan.decode_tpot_target_met,
         stages = ?split_stage_plan_labels(&stages),
+        auto_balance = plan_auto_balance,
```
i386: `auto_balance = plan_auto_balance` logs the *requested* flag. When any placed peer lacks `decode_bytes_per_second` (no benchmark yet), `finish_plan` silently keeps the memory-only cut and `stage_decode_ms`/`stage_idle_pct` log `None` — an operator running `--auto-balance` gets stock placement with no explicit signal why.

Suggestion: log whether balancing was actually applied, e.g. `balanced = plan.throughput.is_some()` or comparing stage boundaries before/after `rebalance_topology`, and a warn when the flag is on but a peer has no speed.

crates/mesh-llm-cli/src/parser/commands.rs
```diff
@@ -679,6 +679,11 @@ pub struct Cli {
     #[arg(long, value_name = "PATH", requires = "split", hide = true)]
     pub split_topology_lock: Option<PathBuf>,
 
+    /// Place split layers by node speed and rebalance them while serving, so a
+    /// slower node does not hold back faster ones.
+    #[arg(long)]
+    pub auto_balance: bool,
```
i386, line 685: Unlike `split_topology_lock`, this has no `requires = "split"`, and without `--split` the flag is inert (the controller only exists in the split coordinator). Consider `requires = "split"` for a clearer CLI contract, or a doc note that it only affects split deployments. Non-blocking.

crates/mesh-llm-host-runtime/src/mesh/stage_transport.rs
```diff
@@ -1131,16 +1134,28 @@ impl Node {
...
+                // listener closes and every stage-0 lane to this peer is
+                // refused for the rest of the generation.
+                let tcp_stream = match listener.accept().await {
+                    Ok((tcp_stream, _)) => tcp_stream,
+                    Err(err) => {
+                        tracing::warn!(
+                            key = %cleanup_key.replace('\n', "/"),
+                            peer = %peer_id.fmt_short(),
+                            error = %err,
+                            "stage transport bridge accept failed; retrying"
```
i386, line 1161: Nit: a permanently-failing accept (e.g. EMFILE that never resolves because nothing closes fds) now warns every 100 ms forever. A short exponential backoff capped around 1 s would keep the recovery latency while avoiding log spam in the pathological case. Non-blocking — the fix itself is correct.

crates/skippy-coordinator/src/topology/performance.rs
```diff
@@ -0,0 +1,366 @@
...
+/// limit.
+///
+/// Returns `None` when a node lacks a speed estimate or no feasible cut exists;
+/// callers keep the memory-only placement in that case. The result is exact:
+/// a dynamic program over contiguous partitions, `O(stages · layers²)`.
+#[allow(
+    clippy::needless_range_loop,
+    reason = "the dynamic program indexes the cost, cut and prefix tables by the same layer boundary"
+)]
+pub(super) fn balance_stages(
```
i386, line 99: Supporting the author's known-gap note with a concrete observation: the capacity model feeding `usable_vram_bytes` (weights + KV, minus headroom) under-counts per-lane compute buffers, prefill spikes and the frontend — and that under-count applies to the **speed-balanced** path's feasibility check here, not just the `INITIAL_CUT=memory` path that produced the 35/1 incident. A fast-but-tight node can still be handed a cut that fits the model but not the machine. Since lane count is already an input (`parallel_lanes` on the plan), folding an estimated per-lane buffer (~230 MiB observed) into `layer_required_bytes` would be a cheap improvement while the explicit headroom guard is worked out. Agree this shouldn't block the PR — the flag is experimental and the gap is documented.

i386 (dismissed):
Verified at bbd8f215 — approving.

**What I ran locally (macOS arm64, separate cargo invocations):**
- `cargo test -p skippy-coordinator` — 55 passed / 0 failed (incl. the new DP, rebalance and throughput-estimate tests)
- `cargo test -p skippy-server` — 758 passed / 0 failed / 3 ignored (env-gated, pre-existing)
- `cargo test -p mesh-llm-host-runtime` — 3,581 passed / 0 failed / 11 ignored
- `cargo clippy -p skippy-coordinator -p skippy-server -p mesh-llm-host-runtime -p mesh-llm-cli --all-targets -- -D warnings` — clean

**Design review:**
- The min–max contiguous-partition DP in `topology/performance.rs` is correct as written: the recurrence covers every feasible contiguous split, capacity constraints are enforced per stage from `layer_required_bytes`, backtracking reconstructs boundaries with every stage non-empty, ties are deterministic (later start), and u128 arithmetic rules out weight/nanos overflow. Test coverage of the interesting corners (memory cap binds before speed, heavy output layer counted by bytes, missing speed → None, infeasible → None) is good.
- Default-off behavior preservation checks out: with the env switches unset, `is_decode_frame_batch_candidate` returns identical decisions to before, and `pipeline_group_batch_size(lanes, 1) == max_direct_batch_size`, so coalescing/wave-gating is unchanged.
- The transport fix is the right shape: the accept loop's only exit is now task abort, and `stop_stage_transport_bridge` removes the map entry *before* aborting, so dropping the in-loop cleanup does not leak bridge entries. The old behavior (first accept error permanently kills the bridge for the generation) matches the EMFILE incident in the description.
- The drain/cutover lifecycle is coherent: `mark_draining(deadline)` → ack → `wait_for_unload_ready` (I verified it polls `in_flight_count` and returns `ForceCancelled` at the drain deadline, so it's bounded), then load-and-publish; on load failure the lifecycle record is re-armed and returned to `Serving` so the old generation resumes. `topology_locked` correctly disables both initial rebalancing (locked planner only estimates) and the runtime controller.
- Counter-underflow safety in `measure()` is guarded by `same_shape` requiring monotonic counters — good.

The measurement section is unusually credible — hour-long windowed runs, the interaction analysis (neither change pays alone), and the honest documentation of the 35/1 memory-cut incident. Inline comments are all non-blocking.

One coordination note for maintainers (not this PR): this branch and the standalone extraction work both touch `runtime/local_split/*` and the scheduler; whoever lands second will need a rebase there.

ndizazzo (dismissed):
I’m requesting changes because the current head still has six blocking findings in the open review threads. I’m referring to those threads rather than duplicating inline comments: stage-0 meter registration can leak entries after a failed load; a topology change can leave a pending auto-balance trial attached to the wrong generation; rollback clears controller state even when cutover fails; the drain acknowledgement wait is unbounded; the public `BinaryStageOptions` field is source-breaking for downstream struct literals; and runtime-operation timing includes lock contention.

The planner/default-off path looks coherent from inspection, and the current GitHub check snapshot is green. I did not run a live multi-node throughput trial, so the reported performance numbers remain author-provided. Once the existing findings are addressed, I can re-review.

PR #1937, opened 2026-09-20
fix(skippy): emit native MTP generation in v2 package writer

### LINKED ISSUE #1385: Qwen3.5-35B-A3B layer package ships MTP weights but omits generation.speculative_decoding, so native MTP can never be selected
## Summary

The published layer package `meshllm/Qwen3.5-35B-A3B-UD-Q4_K_XL-layers` **contains and successfully loads native MTP weights**, but its `model-package.json` declares no `generation.speculative_decoding` block. As a result `strategy = "auto"` resolves native MTP off, and `strategy = "mtp"` would be rejected. The capability ships but is unreachable through the package.

This is a packaging/provenance gap, not a runtime MTP regression — the runtime behaved exactly as specified given the manifest it was handed.

## Evidence the weights are there

Pinned revision `69cbd9fc1f928305e8e86e1148a1db6a6f1de8e9`, manifest sha256 `bac69a30665539535a4018006397ed257d383f96909d991d7a26c6eb3c54d54c`.

- Source model is `unsloth/Qwen3.5-35B-A3B-MTP-GGUF:UD-Q4_K_XL` — MTP in the name.
- GGUF header of `layers/layer-003.gguf`: `general.architecture=qwen35moe`, `block_count=41`, **`nextn_predict_layers=1`**.
- Block 40 is the appended nextn/MTP block (20 tensors, versus 16 for full-attention blocks and 19 for SSM blocks).
- It loads. On a two-node split with the final stage owning `27..41`, that stage reported `skippy_model_open_from_parts ... layers=27..41 ... include_output=true ... status=Ok` with 258 tensors — block 40 included.

## Evidence nothing selects it

Both stages logged `mtp_source=Disabled`. Every speculative counter in the inference `timings` was zero:

```json
"speculative_proposed_n": 0, "speculative_accepted_n": 0, "speculative_rejected_n": 0,
"speculative_windows": 0, "native_mtp_verify_window_verifications": 0, "verify_window_depth": 1
```

## Mechanism

`resolve_native_mtp_strategy` under the default `strategy = "auto"` requires:

```rust
// crates/mesh-llm-host-runtime/src/inference/skippy/resolver/speculative.rs:727-741
package_generation_supports_default_native_mtp(generation) || direct_gguf_supports_native_mtp(model_path)
```

- First disjunct: `generation.speculative_decoding` — the manifest is `schema_version: 1` with **no `generation` key at all** (top-level keys are `schema_version, model_id, source_model, format, layer_count, activation_width, shared, layers, skippy_abi_version, created_at_unix_secs`). So `package_generation` is `None`.
- Second disjunct: `direct_gguf_supports_native_mtp(model_path)` scans a local GGUF for `nextn_predict_layers > 0`. The stage reports `source_model_path: "/source/Qwen3.5-35B-A3B-UD-Q4_K_XL.gguf"` — a build-host path that does not exist on any serving machine. The scan cannot succeed for a package-backed load.

Forcing the issue does not help: `strategy = "mtp"` checks `supports_native_mtp`, built from the same two predicates plus an explicit draft-model path, and otherwise `bail!("skippy speculative.strategy = \"mtp\" requires proven native MTP support")`.

So for any package-backed serve, **the manifest is the only channel through which native MTP can be discovered**, and this package does not use it.

## Why this is worth fixing at the packaging layer

The information needed is already known at package-build time — the builder split a GGUF whose header declares `nextn_predict_layers=1` and emitted the nextn block as `layer-040.gguf`. Emitting a `generation.speculative_decoding` block (the shape documented in `docs/USAGE.md:881`, e.g. as published for GLM-4.7-Flash-MTP) is a mechanical consequence of that.

Suggested:
- **Package build:** when the source GGUF has `nextn_predict_layers > 0` or nextn tensors, emit a `generation.speculative_decoding` entry declaring the native-MTP strategy and its default.
- **Backfill:** republish or amend existing `-layers` packages built from MTP source models. At minimum `meshllm/Qwen3.5-35B-A3B-UD-Q4_K_XL-layers`; worth auditing the rest.
- **Surface the gap:** a package whose layers contain nextn tensors but whose manifest declares no MTP strategy is a detectable inconsistency. Warning at load ("package contains MTP weights but declares no speculative strategy; native MTP disabled") would have turned a silent capability loss into a one-line diagnosis.

## Impact

Any operator serving this package gets ordinary autoregressive decode and no signal that the MTP weights they downloaded (the nextn block is part of the 21.72 GiB) are inert. Measured 3.17 tok/s decode on a two-Mac split — with no way to tell from the API whether MTP was tried and unhelpful, or never attempted.

## Context

Found while spot-checking cross-host split serving of current `main` (`fb58bf73ba375fba1f1921c432bcb0d4ac02ac2b`) on two Macs. The split itself passed: locked topology `0..27` / `27..41` realized exactly, correct per-node partial fetch, both stages ready, correct inference output. MTP could not be exercised because of this.

ndizazzo:
Closes #1385.

## Original problem

The v2 package writer (`crates/skippy-model-package/src/package_v2.rs`) emitted `generation: None` for packages whose source model has native multi-token prediction (MTP) layers. Qwen3.5-35B-A3B, for example, ships GGUF files where the final block is a dedicated MTP head (`architecture=qwen35moe`, MTP layer 40 of 41 blocks), and the llama.cpp runtime pinned in this repo already supports `native-mtp`. With the `generation` block missing from the manifest, the host runtime had no way to enable speculative decoding for those packages, so they ran with plain single-token generation.

I found this while investigating #1385. The `Generation` struct and the `speculative_decoding` schema, including validation, already exist in `crates/skippy-package-format`. Only the writer side never populated them from GGUF metadata.

## Diagnostics

- Pulled the pinned test artifact `unsloth/Qwen3.5-35B-A3B-MTP-GGUF:UD-Q4_K_XL` (rev `69cbd9fc`) and read its GGUF metadata: `architecture=qwen35moe`, `block_count=41`, one MTP layer at index 40.
- Confirmed the pinned llama.cpp runtime supports native MTP: `just skippy-native-tests cpu` passes 51/51 CTest against the vendored pin (`661643e4`, patched to `9cb15795`), so no pin bump is needed.
- Confirmed no cross-repo work is involved: `third_party/llama.cpp/` is vendored in this repo (no `.gitmodules`), and the legacy `Mesh-LLM/hf-mesh-skippy-splitter` has no generation or MTP logic.

## Fix

`manifest_from_source` in `package_v2.rs` now calls `infer_native_mtp_generation` when the source architecture is known to carry an MTP head. The helper reads the GGUF architecture metadata, takes the MTP layer index as `layer_count - 1`, and emits:

```json
{
  "default": "mtp",
  "strategies": {
    "mtp": {
      "type": "native-mtp",
      "prediction_depth": 1,
      "layer_indices": [40],
      "window_policy": { "default": "fixed", "initial_window": 1, "min_window": 1, "max_window": 1 }
    }
  }
}
```

Compatibility and migration notes:

- No schema or ABI change: the `Generation` struct and `speculative_decoding` validation pre-exist in `skippy-package-format`.
- `generation` participates in the `computed_package_id` hash, so republishing an existing package with this writer produces a new package ID. The old package stays valid; it just has no generation block.
- The fix is writer-side. The host runtime path (`crates/mesh-llm-host-runtime/src/inference/skippy/package.rs`), the split job script (`crates/model-package/src/scripts/split-model-job.sh`), and the docs (`docs/USAGE.md`, `docs/design/SKIPPY_PACKAGE_V2_SCHEMA.md`) were updated to match.

Ten new unit tests in `package_v2/tests.rs` cover the inference helper and the manifest output, including the happy path, non-MTP architectures, and malformed metadata.

## Validation

  - `cargo test -p skippy-model-package`: 98 passed, 0 failed (includes the 10 new MTP tests)
  - `just skippy-native-tests cpu`: CTest 51/51 against the pinned llama.cpp runtime
  - `cargo clippy --all-targets -D warnings`, `cargo fmt --check`, and `shellcheck` on the changed scripts: all clean
  - `just check-commits`: passed
  - No UI changes. This is package-manifest writer behavior.

## Summary by CodeRabbit

* **New Features**
  * Model packages now detect supported native multi-token prediction (MTP) capabilities and include speculative decoding metadata.
  * Model cards display speculative decoding strategy details, prediction depth, layer information, and defaults.
  * Package identity now reflects speculative decoding configuration.

* **Bug Fixes**
  * Added warnings when MTP tensors are present without speculative decoding configuration.
  * Invalid or inconsistent MTP metadata is safely rejected.

* **Documentation**
  * Added guidance and schema documentation for speculative decoding configuration in Model Package v2.


crates/skippy-model-package/src/package_v2.rs
```diff
@@ -262,6 +263,114 @@ fn manifest_from_source(
...
+        );
+    }
+    if declared && depth.is_none() {
+        bail!("`{key}` is present in the source metadata but is not an integer");
+    }
+    match depth {
+        None if indices.is_empty() => Ok(None),
+        Some(0) if indices.is_empty() => Ok(None),
+        Some(depth) if depth > 1 => bail!(
+            "source declares {depth}-step native MTP prediction; the runtime executes depth 1 \
```
i386, line 285: Design question, non-blocking: with `nextn_predict_layers > 1` the writer now **fails the whole package**, whereas before this change such a source packaged fine (with `generation: None`) and served single-token. Hard-fail is defensible — it forces the issue instead of silently dropping MTP — but it makes a depth-2 model unpackageable in a world where plain decoding would work. The alternative is `Ok(None)` plus a loud warning, with the caveat that the host-side `manifest_ships_mtp_without_generation` warn text ("republish with a fixed skippy-model-package writer") would then be misleading for that case and would need a depth distinction. Happy with either; flagging the trade-off so it's a decision rather than an accident.

crates/mesh-llm-host-runtime/src/inference/skippy/package.rs
```diff
@@ -1575,6 +1599,15 @@ fn identity_from_package_v2_metadata(
         .context("package-v2 tensor count exceeds u64")?;
     let manifest_sha256 = hex_lower(&Sha256::digest(&manifest_bytes));
     let canonical_package_ref = canonical_layer_package_ref(package_ref, local_ref);
+    if manifest_ships_mtp_without_generation(&manifest) {
```
i386: Nit: this warn block + guard is duplicated verbatim from `identity_from_package_v2` (line ~390). A tiny `warn_if_mtp_without_generation(&manifest)` helper called from both would keep them in lockstep — right now a future edit to one copy can drift from the other. Non-blocking.

crates/skippy-model-package/src/test_gguf.rs
```diff
@@ -26,30 +26,121 @@ fn string(bytes: &mut Vec<u8>, value: &str) {
...
+pub(crate) fn fixture_with_nextn_noninteger(
+    path: &Path,
+    block_count: u32,
+    tensors: &[FixtureTensor<'_>],
+    split: Option<(u16, u16, u64)>,
+) {
+    write_fixture_noninteger_nextn(path, block_count, tensors, split, true);
+}
+
+fn write_fixture_noninteger_nextn(
```
i386: Nit: `write_fixture_noninteger_nextn` duplicates `write_fixture` (~55 lines) differing only in how the `nextn` key is serialized (string "one" vs u32). Parameterizing `write_fixture` with something like `nextn: Option<GgufValue>` would remove the copy. Test-support only — non-blocking.

crates/model-package/src/scripts/split-model-job.sh
```diff
@@ -706,6 +718,20 @@ def code(value) -> str:
 def yaml_quote(value: str) -> str:
     return json.dumps(value)
 
+def speculative_summary(spec: dict) -> str:
```
i386: Nit: `speculative_summary` here and the detail-string loop in the manifest-verification block (~line 480) render the same strategy fields in two places. If the shape of a strategy ever grows (e.g. a `pipeline_depth` display), the job log and the model card will drift. A single shared helper used by both would keep them identical. Non-blocking.

i386 (dismissed):
Verified at bf21df84 — approving.

**What I ran locally:**
- `cargo test -p skippy-model-package` — 98/0 (matches the PR's claim; includes the 10 new MTP tests)
- `cargo test -p model-package` — 13/0 + 11/0 (script tests incl. the new `EMBEDDED_SCRIPT` assertions)
- `cargo test -p mesh-llm-host-runtime` — 3,600/0/11 ignored (includes the 4 new warn-path tests)
- `cargo clippy --all-targets -D warnings` on the three touched Rust crates, `cargo fmt --all --check`, `bash -n` on the changed script — all clean
- GitHub CI green at this SHA (the cancelled Quality checks in the rollup are a superseded duplicate run from 02:36; the 02:38 rerun completed successfully)

I did not rerun `just skippy-native-tests cpu` — no native/vendored code changes in this PR, and the native runtime CI jobs passed; the claim is consistent with that.

**Contract verification (source-level):**
- The emitted declaration matches what the runtime consumes, end to end: `StrategyKind::NativeMtp` (inline, `proposer: None`, depth 1) → `package_v2_generation_info` → `PackageSpeculativeStrategyInfo { strategy_type: "native-mtp" }` → `strategy_uses_native_mtp` accepts the no-external-proposer form, and `package_generation_supports_default_native_mtp` resolves `default: "mtp"` against the same map. With `strategy=auto` the resolver enables native MTP off this declaration exactly as the doc comment on `native_mtp_generation` claims.
- The fail-closed matrix is complete — I walked all six combinations of (`{arch}.nextn_predict_layers` present/absent/0/non-integer) × (`.nextn.` tensors present/absent): only mutual agreement emits a declaration; every disagreement bails *before* a manifest is written (tests assert `model-package.json` absent). This is the right posture — a wrong declaration would silently disable speculative decoding everywhere downstream.
- `layer_indices == [layer_count - 1]` cross-check against the inventory closes the subtle case where the MTP block is not the final block.
- Package-identity semantics are correct and tested: `generation` participates in `computed_package_id`, the MTP package id differs from both the plain package and the same-manifest-without-generation hash, and existing packages without `generation` remain valid (optional field, omitted rather than null).

Inline comments are non-blocking (one design question, three small dedup nits).

i386:
@ndizazzo after this we will need to regen some layer packages

crates/mesh-llm-host-runtime/src/inference/skippy/package.rs
i386: Fixed in 037ec7821 — extracted `warn_if_mtp_without_generation(&manifest)` and called it from both `identity_from_package_v2` and `identity_from_package_v2_metadata`, so the guard and the message live in one place.

crates/skippy-model-package/src/test_gguf.rs
i386: Fixed in 037ec7821 — `write_fixture` now takes a `NextnPredictLayers` value (`Absent` / `Layers(u32)` / `NonInteger`) and `write_fixture_noninteger_nextn` is gone. `non_integer_nextn_key_fails_closed` still passes, so the emitted fixture bytes are unchanged.

crates/model-package/src/scripts/split-model-job.sh
i386: Fixed in 037ec7821 — the strategy-field rendering now lives in one helper script that both the manifest-verification log line and the published model card load, so a strategy growing a field cannot make the job log and the card drift.

i386:
The three non-blocking nits from the review at `bf21df84` are cleaned up in `037ec7821` (@ndizazzo).

- **Host runtime** (`inference/skippy/package.rs`) — the duplicated warn block + guard is now `warn_if_mtp_without_generation(&manifest)`, called from both `identity_from_package_v2` and `identity_from_package_v2_metadata`.
- **Test fixtures** (`skippy-model-package/src/test_gguf.rs`) — `write_fixture` is parameterized with a `NextnPredictLayers` value (`Absent` / `Layers(u32)` / `NonInteger`); the ~55-line `write_fixture_noninteger_nextn` copy is deleted. Emitted fixture bytes are unchanged.
- **`split-model-job.sh`** — the speculative-decoding rendering lives in one helper (`SPECULATIVE_SUMMARY_HELPER`) loaded by both the verification log line and the published model card, so a new strategy field cannot make them drift.

Left untouched: the `nextn_predict_layers > 1` hard-fail on `package_v2.rs`. That one was a design question, not a nit, so nothing changed — a depth-2 source still fails the package rather than emitting `generation: None` plus a warning. Flagging it so the decision stays visible for the author/owner.

Validation at `037ec7821`:

- `cargo test -p skippy-model-package` — 98 passed
- `cargo test -p model-package` — 13 + 11 passed
- `cargo test -p mesh-llm-host-runtime --lib` — 3,600 passed, 11 ignored
- `cargo clippy … --all-targets -- -D warnings` on the touched crates (`skippy-model-package`, `model-package`, `mesh-llm-host-runtime`, `mesh-llm`) — clean
- `cargo fmt --all --check`, `just no-console-print`, `just ci-shellcheck <changed script>`, `just check-commits` — clean

The push dismissed the existing approval, so this needs one re-approval once CI reruns on the new head.

i386 (approved):

crates/skippy-model-package/src/package_v2.rs
i386: Decision: keep the hard fail. Recording the reasoning so this is a decision and not an accident.

1. It is consistent with this function's contract. `infer_native_mtp_generation` is fail-closed for every inconsistency — unparseable `.nextn.` names, a non-integer `{arch}.nextn_predict_layers`, and a depth that disagrees with the `.nextn.` tensor layers all `bail!`. Making the `> 1` arm the one place that returns `Ok(None)` would mean a source whose declared MTP the runtime cannot honour still produces a package.

2. `Ok(None)` + warning recreates the defect #1385 was filed for. A package that ships `nextn` tensors with no `generation.speculative_decoding` is exactly what makes native MTP unreachable through the manifest — the whole reason for this PR. The host-side warning also prescribes a remedy ("republish with a fixed skippy-model-package writer") that can never work for depth > 1, which is the misleading text you flagged. Leaving that in place while widening the set of packages that hit it trades a loud failure for a silent, permanently unfixable one.

3. Blast radius today is theoretical. Every MTP source in the `meshllm` org declares `nextn_predict_layers = 1` (all seven regeneration candidates included), and the pinned runtime executes depth 1 only.

Revisit trigger, so this does not have to be relitigated: if a depth-2 source actually appears, emit a depth-2 declaration once the runtime can honour it, or — if we decide plain decoding is acceptable for it — flip this arm to `Ok(None)` and give the host warning a depth distinction in the same change. Doing only the `Ok(None)` half now would leave the bad advice in place.

PR #1938, opened 2026-09-20
fix(split): floor the node reserve at the compute graph; explicit starting cut

danielwinterw:
Stacked on #1935 — review that first; this branch is one commit on top of it.

Two changes, both from the same failure.

## The failure

Running the auto-balance controller from the memory-first cut, to prove it
converges from a poor placement, instead produced this:

- The memory-first planner filled the first node to its budget: **35 of 36
  layers** on one node.
- Planning priced that stage at **8.7 GB** against a 12 GB budget.
- Eight minutes in, the process was at **12.0 GB resident** with **10% of a
  16 GB machine free**. The node stopped heartbeating, its supervisor killed
  it, and the split lost the stage.
- The coordinator stayed up and rejected **42,863 requests** over the rest of
  the hour rather than hanging, which is the correct behaviour and the only
  part of this that went right.

## 1. Floor the node reserve at the compute graph

Placement prices a stage on weights and KV, with a compute-buffer reserve that
rides on the KV term — so it scales with `n_ctx`. The compute graph does not: a
four-lane stage allocated five buffers of 229.61 MiB each, the same whether it
held 12 layers or 18. A stage holding many layers at a modest context is
therefore priced almost entirely on weights and KV, and the proportional share
does not cover what the graph then takes.

This floors the per-node reserve at 1 GiB. The floor is **flat on purpose**:
the measurement fixes the scale at one working point (one model, one batch
size), and extrapolating it per lane would reserve several GiB on exactly the
large-context nodes whose proportional share is already generous — I tried
that first and it collapsed a 16-lane plan to a single lane. Small nodes keep
the proportional share rather than being reserved out of existence.

**This does not by itself prevent the failure above**, and it would be wrong to
claim it does. 35 layers still fit inside a 12 GB budget after the floor. What
it fixes is the accounting gap — planning under-predicted actual use by 3.3 GB,
and roughly a third of that was the graph. The rest is the frontend and the
margin between a budget and a machine: `--max-vram 12` on a 16 GB host that is
also running the frontend leaves too little for the OS, and that is a
documentation and defaults question rather than a planner bug.

## 2. An explicit starting cut for the acceptance test

`MESH_LLM_AUTO_BALANCE_INITIAL_CUT` now takes boundaries (`12` for a two-stage
12/24 cut) instead of only `memory`.

The controller can only be exercised from a placement worth correcting, and
neither planner path gives one: speed-balanced placement starts at the answer,
so there is nothing to converge from, and the memory-first cut starts somewhere
fatal. An explicit cut starts the test somewhere deliberately wrong but
survivable.

Nothing trusts the value — an overridden plan goes through
`validate_split_capacity` like any other, so a cut that does not fit is
rejected rather than loaded, and boundaries that do not describe the topology
are ignored with a warning rather than silently producing a different split.

## Next

With this, the controller acceptance hour can actually run: start at 24/12,
confirm it walks to 12/24 under steady load and holds. That run is not in this
PR.

## Summary by CodeRabbit

* **Improvements**
  * Runtime memory planning now preserves minimum per-node headroom, improving reliability for workloads with large buffer requirements and smaller nodes.
  * Initial layer-cut configuration accepts valid comma-separated layer boundaries for controlling runtime stage placement.
  * Valid boundaries are applied in node order with automatic balancing disabled during initial planning. Invalid or topology-mismatched values are ignored with a warning.
  * TCP-to-QUIC stage links can optionally apply a configurable delay while preserving data order and throughput accounting.
* **Tests**
  * Added coverage for memory headroom, initial layer-cut handling, delay configuration, and delayed stream forwarding.


crates/mesh-llm-host-runtime/src/runtime/split_planning.rs
```diff
@@ -239,6 +262,7 @@ pub(super) fn plan_runtime_slice_topology_with_resources_and_stage0(
 
     let mut stages = map_runtime_slice_stages(plan.stages, &participant_by_id)?;
     stages.sort_by_key(|stage| stage.stage_index);
+    apply_initial_cut_override(&mut stages, package.layer_count);
```
ndizazzo: [P2] Keep this override behind `resources.auto_balance`. `MESH_LLM_AUTO_BALANCE_INITIAL_CUT` is documented as an acceptance-test knob, but this function also serves ordinary fixed split startup where `resources.auto_balance` is false, including the load and startup paths. Before this change, the `memory` sentinel had no effect in that mode. As written, exporting `MESH_LLM_AUTO_BALANCE_INITIAL_CUT=12` changes fixed split placement even though no auto-balance controller is enabled. Please apply the override only for auto-balance plans, or make a separate fixed-topology option explicit.

ndizazzo (dismissed):
I reviewed `fe9a867e777e96e7ea4e8a8f1a01308fe87974b0` against `bbd8f2152aec3fb343d35b70d0b0238e37c60a8b`. I’m requesting changes for three issues.

The two open CodeRabbit findings are valid: [capacity validation](https://github.com/Mesh-LLM/mesh-llm/pull/1938#discussion_r4055839463) still uses the pre-override `parameter_bytes`, and [invalid overrides](https://github.com/Mesh-LLM/mesh-llm/pull/1938#discussion_r4055839469) can disable planner auto-balance before `apply_boundaries` later ignores them. Please address those existing threads.

I added one separate inline comment for the new override being applied when `resources.auto_balance` is false. That changes ordinary fixed split startup whenever the acceptance-test environment variable is present, although the old `memory` sentinel was ignored in that mode.

The exact-head snapshot also has the required Quality gate failing because the PR subject is 101 characters and exceeds the repository’s 100-character conventional-commit limit. Linux, macOS, Windows, and Website checks are terminal green; hardware-specific rows are skipped. `git diff --check` passes. I did not validate the live acceptance-hour behavior.

i386:
Pushed two commits addressing the open review threads:

**`c1766387` — tunnel.rs (both delay threads)**
- `MESH_LLM_STAGE_LINK_DELAY_MS` now treats unset as the only disabled state; present-but-empty, non-numeric, and zero values fail the bridge with a parse error instead of silently running an unshaped sweep. Parsing tests updated to expect errors (`parse_stage_link_delay` is a pure function, so no env mutation in tests).
- The env consult moved out of the shared `relay_tcp_to_quic` path. The outbound stage-transport bridge resolves the delay once and passes it explicitly; inbound stage transport and the legacy HTTP relay forward immediately. A request-response pair gains exactly one `delay` even when every node in the sweep exports the variable.

**`5d81c3a0` — split_planning.rs (capacity-validation thread)**
- `validate_split_capacity` now takes a `SplitCapacityModel` (plan context/lanes + KV + recurrent costs) and re-runs the planner's full per-node budget — usable VRAM after runtime headroom, context-scaled KV at the 100/85 compute reserve, lane-scaled recurrent state — over the ranges the stages actually hold. For planner-produced placement this is idempotent (same model, same inputs); for an override-moved cut it is the only real check. The resource-aware planners now thread the model through; the test-only package-identity planner has no context model, so its backstop keeps the weight-only raw-budget check rather than double-counting headroom that planner never charged.
- Coverage added: overrides that fit by weights but exceed the KV compute reserve, the lane-scaled recurrent budget, or the runtime headroom margin each fail validation with the budget named.

Local gates: `cargo fmt --all --check`, `just no-console-print`, `clippy -p mesh-llm --all-targets -D warnings` (and `-p mesh-llm-host-runtime`), host-runtime lib suite 3690 passed / 0 failed.

ndizazzo (approved):
The fixed-split guard does what I asked for now. Approved.

PR #1939, opened 2026-09-20
feat: prototype OpenJEV System One on Skippy

michaelneale:
Prototype `POST /v1/systemone` structured decisions (`noul`, `choice`, `score`) on a full-model DiffusionGemma Skippy worker. This is a text-only, single-read proof of concept, not full OpenJEV parity or split-model certification.

## Original problem

MeshLLM could package DiffusionGemma layers but had no System One structured-read endpoint or native operation for reading label probabilities from its diffusion canvas. Chat generation is not a substitute for this operation.

## Diagnostics

The implementation and reproduction commands are in [the runbook](docs/design/OPENJEV_SKIPPY_POC.md). James's [independent review at b0e20225d](https://github.com/Mesh-LLM/mesh-llm/pull/1939) identified two evidence requirements: a live CUDA curl/JSON/worker-log transcript and a completed template. The template is now filled; **the [live CUDA transcript and six-case read matrix are now recorded](https://github.com/Mesh-LLM/mesh-llm/pull/1939#issuecomment-5754701276) at `0ae5afd34558089d40b62631311241ed31fd04c2`**. The proof requires `cache_idle_slots = 0` and a single-GPU CUDA runtime built with optional NCCL disabled; the linked report documents both configuration findings.

Existing local evidence is Metal, not CUDA: the saved M5 Max benchmark record at `b0e20225dc2d3297deb43a66e5dcbbab3d3dda48` reports 400/400 LocalLLaMA/typed-decisions cases completed (2,000 decisions), 66.55% teacher agreement, and 508.5 ms p50 / 627.2 ms p95. The retained raw JSON was rechecked on September 21: SHA-256 `2b32573bcd66d7481d107fdb15e9c6676e6d23b16ed211a3005c462048ec393c`. This historical result is neither a fresh run at the current head nor CUDA evidence, and teacher agreement is not proof of production correctness.

## Fix

- Port DiffusionGemma support through the existing native patch queue and add a versioned Skippy System One read ABI.
- Build and validate a fixed answer canvas, execute one read, and return label-restricted probabilities through the OpenAI frontend.
- Query the model's canvas length rather than assuming the upstream default.
- Reject staged execution and unsupported modes explicitly; free native batches with RAII and echo the requested model name.
- Add pinned-fixture canary contract/read checks. A missing artifact on a declared-qualified backend fails; an unqualified backend is explicitly reported, not claimed as a full-read pass.

### Compatibility / protocol

Source comparison reference: [OpenJEV `openjev/api.py` at `febf02e889892747ced5497f2a17323731621ccd`](https://github.com/razorback16/openjev/blob/febf02e889892747ced5497f2a17323731621ccd/openjev/api.py), inspected September 21, 2026. This is a source-level protocol-shape comparison, **not a cross-server conformance test**.

The supported subset accepts `{model, state, questions}` and returns `{model, answers, usage}` with typed `noul`, `choice`, and `score` answers. Choices are limited to 2–26, score levels to 2–10. Images, thinking, multiple steps/samples, sequential reads, and distributed-stage reads are unsupported. Unlike upstream's canonical `MODEL_VERSION` response, this implementation intentionally echoes the requested model alias. It does not reproduce upstream entropy-triggered repeated reads/averaging. Error envelopes follow MeshLLM's OpenAI frontend rather than promising byte-for-byte Jev errors. `usage.input_tokens` counts prompt tokens only, excludes canvas computation, and `output_tokens` is zero. System One input bypasses chat guardrail screening in this PoC.

### Example

```bash
curl http://127.0.0.1:9337/v1/systemone \
  -H 'Content-Type: application/json' \
  -d '{"model":"openjev-latest","state":"I was charged twice this month.","questions":{"is_billing":{"type":"noul","instructions":"Is this a billing issue?"}}}'
```

Illustrative response structure only (not a claimed CUDA transcript):

```json
{"model":"openjev-latest","answers":{"is_billing":{"type":"noul","noul":0.95}},"usage":{"input_tokens":123,"output_tokens":0}}
```

## Validation

- [x] Relevant checks are recorded with their scope; missing live validation is explicitly disclosed below.
- [x] No UI changes; screenshots/video do not apply.
- [x] Independent historical Rust validation at `82eccca86`: see [reviewer's recorded suites and nit verification](https://github.com/Mesh-LLM/mesh-llm/pull/1939#issuecomment-5748630448). These results are not asserted for the later merges.
- [x] At earlier head `b9d67e4e6a6e748a76d7ad3bdff782a25593cbf1`, Jimmy reran `python3 -m unittest scripts.tests.test_skippy_system_one_smoke`: 28 passed; `python3 scripts/generate-split-certified.py --check`: passed. These validate the smoke harness/roster, not live native inference.
- [x] Live CUDA System One transcript at `0ae5afd34`: HTTP 200 curl/JSON, worker/native logs, host/runtime hashes, and all six full-read driver cases passed. See linked report for model reference and exact configuration; this is not a pinned-fixture canary certification.
- [ ] Fresh full Rust suite validation of current merged head. Linux release host and CUDA runtime were built and exercised at `0ae5afd34`; this does not replace the full Rust suites.

Keep this PR draft until the evidence and idle-session/runbook configuration issue have been reviewed. No family/split certification or new Metal qualification is claimed by this update.


i386 (changes requested):
# Review: OpenJEV System One prototype (b0e20225d)

Reviewed at commit `b0e20225d` in an isolated worktree, with independent reproduction of the Rust suites and gates. Overall: the engineering quality here is high and the scoping is honest — the blocking gap is validation evidence, not code.

## What I verified independently

- `cargo test -p openai-frontend`: **237 passed / 0 failed** (includes the new `system_one_route_preserves_jev_response_shape` route test)
- `cargo test -p skippy-server`: **755 passed / 0 failed / 3 ignored** (separate invocation)
- `cargo test -p skippy-runtime`: **164 passed / 0 failed / 1 ignored**; `cargo test -p skippy-ffi`: **11 / 0** (run in isolation — see note below)
- `cargo test -p mesh-llm-host-runtime`: **3597 passed / 0 failed / 11 ignored**
- `cargo clippy` and `cargo fmt --check` on all five touched crates: clean (zero code warnings; only environmental build-script/linker notes on my machine)
- xtask `repo-consistency` × 4 (release-targets, ci-crate-lists, publish-crates, test-all-rust-crate-coverage): **all exit 0**
- CI PR matrix (Linux/macOS/Windows/Quality/Website) green, which exercises the full llama.cpp patch stack build

## What's sound

- **Fail-closed boundaries, twice.** The Rust mode gate rejects `EmbeddedStageZero` with a downstream (`system_one.rs`), and the native `skippy_system_one_read` independently rejects staged models (`skippy_runtime_has_stage_plan`), non-DiffusionGemma arches, and `lane_count != 1`. A staged model cannot produce a partial read.
- **Native validation is thorough**: slot bounds/offset arithmetic checked against the flattened label array, unique canvas positions, vocab-bounds on label tokens, exact `canvas_token_count == diffusion.canvas_length`, ubatch and n_ctx capacity checks, numerically stable label-restricted softmax with a non-finite guard.
- **The canvas builder self-verifies against the tokenizer**: each label swap must produce exactly one differing token at a consistent position, else the request is rejected loudly ("labels do not share one tokenizer slot"). This is the right defense against tokenizer merge surprises.
- `causal_attn=false` on the shared context is not a leak: the DiffusionGemma arch defaults `hparams.causal_attn = false` and builds region-aware masks itself, and the route rejects every other arch.
- FFI/ABI hygiene matches house style: feature bit `1 << 37`, ABI patch 60→61 bumped in `skippy-ffi`, `skippy/common.h`, and `split-certified.json` consistently; static + dynamic bindings both declared; generated docs page regenerated (15→16 headers, 96→98 functions) and the regenerated `0077-family-unmapped.patch`/`series.json` shas updated.
- Port provenance is documented (upstream ggml-org/llama.cpp#24423 at `12e0a962`, original author attributed in the patch header).
- Router/backend wiring follows existing conventions exactly (trait default + wrapper pass-throughs, lifecycle enums, request-metadata labels, `spawn_blocking` for the native read).

## Blocking (before ready-for-review / merge)

1. **No live-read evidence anywhere.** The validation checkboxes are unchecked, the template sections are unfilled, and the design doc itself says a live Metal result "has not been certified". Every test in the PR validates routing/shaping/math — nothing validates that a real DiffusionGemma model returns meaningful probabilities through this canvas construction (scaffold string, pad-token choice, phase sequencing). Please attach one live transcript from the runbook's CUDA path: the exact curl, the JSON response, and the worker log excerpt. That single artifact is what makes "Jev-compatible" a tested claim rather than an aspiration.
2. **Fill the PR template** (original problem / diagnostics / fix / validation). In particular the compatibility statement: which OpenJEV version/protocol shape `{model, answers, usage}` was checked against.

## Non-blocking nits

- `src/skippy/system_one.cpp`: the `catch` handlers return without `llama_batch_free(batch)` — the batch leaks if a native exception escapes mid-read (early returns inside the `try` do free it). A small RAII guard would close it.
- `run_system_one` returns `"openjev-0.1"` whenever the caller passes an alias (`openjev-latest`, `jev-latest`, …) instead of echoing the requested model string, which both OpenAI and Jev clients conventionally expect.
- `GuardedOpenAiBackend::system_one` forwards without any guardrail screening — presumably intended for structured reads, but worth a sentence in the design doc since `state` text reaches the model unscreened.
- `usage.input_tokens` counts prompt tokens only; the canvas tokens the model actually computes over are unreported (and `output_tokens: 0`). Fine for a PoC, worth documenting.
- FYI, pre-existing and not from this PR: `cargo test -p skippy-runtime -p skippy-ffi` in one invocation fails 3 tests (2 ngram cache + `invalid_selected_backend_device_fails_before_model_open`) — I reproduced the identical failures at parent `e62383553`, and each crate is green in isolation. Same feature-unification trap as the known skippy-server/host-runtime one; worth a CI note someday.

## Verdict

Requesting changes for the missing validation evidence and unfilled template only. The code and patch discipline are in good shape; one live transcript plus the template turns this into an approve from me.

i386:
**Verification note at `44cf028f2`** (independent worktree): the four nit fixes verify at the diff level (RAII batch guard with all manual frees removed, model echo, guardrail/usage documentation), but `python3 scripts/generate-split-certified.py --check` **fails**: committed `patch_queue_sha256` `4fd21f2c…` ≠ queue digest `cfe5ddf4a705bb8c9fe48edb6421ab7e0c246f01efc134db497336ef131f135d` at this tree (canonical per both the script and `mesh-llm-host-runtime/build.rs` v2 framing; the `b0e20225d` value reproduces exactly as a control). The roster needs regeneration — the digest appears to have been computed against an intermediate patch-queue state.

Review status unchanged (request-changes): still waiting on one live CUDA read transcript per the runbook and the PR template/validation boxes.

i386:
**Verification complete at `82eccca86`**: roster digest regenerated correctly (`cfe5ddf4a705bb8c9fe48edb6421ab7e0c246f01efc134db497336ef131f135d`; `generate-split-certified.py --check` green at the pushed HEAD). The four nit fixes are independently verified (RAII batch guard — patched tree retains exactly one `llama_batch_free`, in the guard dtor, both catch escapes unwind-covered; model echo; guardrail/usage documentation). Local matrix green at that object: openai-frontend 237/0, skippy-server 755/0/3, skippy-runtime 164/0/1, skippy-ffi 11/0, host-runtime 3597/0/11, clippy/fmt/xtask clean.

Review status remains **request-changes**, now only on: (1) one live CUDA System One read transcript (curl + JSON + worker log) per `docs/design/OPENJEV_SKIPPY_POC.md`, and (2) the PR template filled with validation boxes checked.

i386:
**Canary-smoke addition verified at `f6ffb633c`**: the System One (OpenJEV) llama-canary smoke (`scripts/skippy-system-one-smoke.sh` + driver + manifest + three-lane wiring) was independently reviewed and approved at that exact object (tree `da9bcc33c`, confirmed as the pushed head). Suite reproduced locally (1558 OK / 9 skipped; actionlint/shellcheck/manifest checks clean); first live contract-part execution lands with the next canary nightly.

Overall review status for the PR remains **request-changes** on the two author blockers: the live CUDA System One read transcript per `docs/design/OPENJEV_SKIPPY_POC.md`, and the PR template with validation boxes checked.

scripts/skippy-system-one-smoke.sh
```diff
@@ -0,0 +1,537 @@
...
+    return 3
+  fi
+  local summary cached
+  summary="$(artifact_summary "$READ_ARTIFACT_ID")" || return 2
+  cached="$(cached_artifact_path "$(jq -r '.repo' <<<"$summary")" \
+    "$(jq -r '.revision' <<<"$summary")" "$(jq -r '.file' <<<"$summary")")"
+  if [[ -z "${SYSTEMONE_SMOKE_READ_MODEL_PATH:-}" && -z "$cached" ]]; then
+    echo "system-one: full-model read NOT CERTIFIED: pinned $(jq -r '.repo' <<<"$summary")/$(jq -r '.file' <<<"$summary") is not in the offline model cache" >&2
+    echo "pre-warm it with: scripts/skippy-system-one-smoke.sh --prewarm" >&2
+    return 3
```
ndizazzo: [P1] Fail a missing artifact on qualified full reads

`return 3` is also used for the genuinely unqualified backend, and `main` maps both cases to `unqualified` and exits 0 unless `SYSTEMONE_SMOKE_REQUIRE_QUALIFIED` is manually set. On a backend already listed in `CERTIFIED_BACKENDS`, a missing pinned artifact therefore lets the canary and changed-pin harness pass without running any full-model read. I reproduced this at `f6ffb633c` with CUDA declared qualified and an empty cache. Please return a failing status for this cache miss, or make the qualified gate automatic.

.github/workflows/llama-upstream-canary.yml
```diff
@@ -339,6 +339,63 @@ jobs:
...
+          report="$WORK_DIR/reports/system-one.json"
+          {
+            echo "## System One (OpenJEV) smoke"
+            echo
+            if [[ ! -f "$report" ]]; then
+              echo "No System One smoke report was produced (step outcome: ${SYSTEM_ONE_OUTCOME})."
+            elif [[ "$(jq -r '.status' "$report")" == "unqualified" ]]; then
+              echo "**Full-model System One read: NOT CERTIFIED on this runner.**"
+              echo
+              echo "The contract part $(jq -r '.contract.status' "$report"). The complete-model read is admitted only on a declared qualified backend; this runner reports backend \`$(jq -r '.full_model_read.backend // "<unset>"' "$report")\` against \`$(jq -r '.full_model_read.certified_backends | join(",")' "$report")\`, with the pinned artifact $(if [[ "$(jq -r '.full_model_read.artifact_path // ""' "$report")" == "" ]]; then echo "not present in the offline cache"; else echo "present in the offline cache"; fi)."
```
ndizazzo: [P2] Report the actual cache state

`system-one.json` never writes `full_model_read.artifact_path` (or any cache-present flag); the report builder has only `artifact`, `backend`, `certified_backends`, and `require_qualified`. This expression therefore always prints `not present in the offline cache` on every unqualified run, even when the pinned file is present, which makes the canary summary and prewarm advice wrong. Emit the resolved path or a boolean when the cache is checked, or remove this claim.

scripts/llama-canary-agent-repair.sh
```diff
@@ -56,6 +56,7 @@ PR_BODY="$STATE_DIR/pr-body.md"
 UPSTREAM_SUMMARY="$STATE_DIR/upstream-summary.md"
 BUNDLE="$STATE_DIR/candidate.bundle"
 EVIDENCE_DIR="$STATE_DIR/verification-evidence"
+SYSTEMONE_SMOKE_DIR="$ROOT/target/skippy-system-one-smoke"
```
ndizazzo, line 71: [P2] Keep verifier smoke evidence in the candidate worktree

`SYSTEMONE_SMOKE_DIR` is captured from the trusted checkout before `materialize_verification_tree` assigns `ROOT="$VERIFY_ROOT"`. In verify mode this sends the smoke report and evidence to the original checkout's `target/skippy-system-one-smoke`, while `cleanup_verification_worktree` only copies `$VERIFY_ROOT/target/skippy-system-one-smoke` into `EVIDENCE_DIR`. The independent verification artifact is therefore omitted. Recompute this path after the verification worktree is materialized, or pass a path under the new `ROOT`.

ndizazzo (changes requested):
I found a canary gating hole in addition to the two evidence items already called out in the earlier review. A missing pinned artifact on a backend that is already declared qualified currently exits successfully without running the full-model read. Please separate that cache-miss failure from the intentionally unqualified Metal path before merging. The earlier review's live CUDA transcript and completed PR template are still outstanding.

i386:
have agent fixing conflicts.

i386:
Merged `main` into the branch (merge commit `92d52e018`) — the patch-series conflict is resolved by renumbering the DiffusionGemma/System One patches to `0003`–`0005` on top of main's `0002-fix-kimi-k3...`, with `generated/series.json` and the split-certification roster regenerated through the canonical pipelines (stage-rewriter check passes on the merged queue).

All three review findings are fixed:

- **[P1] Missing artifact on a qualified backend** — `skippy-system-one-smoke.sh` now returns a distinct failing code (`4` → `fail`) when the build backend is declared qualified but the pinned artifact is absent; the genuinely-unqualified path keeps its `unqualified`/exit-0 semantics. A test now pins the fail-closed behavior (`test_a_qualified_backend_with_a_missing_artifact_fails`).
- **[P2] Cache state in the report** — the report emits `full_model_read.artifact_path` plus `artifact_cache_checked` (false on the unqualified path, where the cache is never consulted); the job summary renders "cache state not checked" / "not present" / "present" from the report instead of guessing.
- **[P3] Verifier evidence** — `SYSTEMONE_SMOKE_DIR` is re-derived inside `materialize_verification_tree` so verify-mode writes land in `/target/skippy-system-one-smoke`, which the cleanup already copies into the evidence dir.

Local validation at the merge commit: cargo fmt/clippy `(openai-frontend, skippy-server, mesh-llm)` all green, `openai-frontend`/`skippy-server`/`mesh-llm-host-runtime` lib suites 3651 passed / 0 failed, python smoke test suite 28/28, `just no-console-print` green.

Still outstanding from the earlier review: the live CUDA transcript and PR-template completion need the artifact owner's environment.

michaelneale:
Addressing James’s outstanding review: filled the problem/diagnostics/fix/validation template and pinned the OpenJEV protocol-shape reference to upstream febf02e889892747ced5497f2a17323731621ccd. The compatibility section now explicitly distinguishes the supported subset, alias echo, errors, usage accounting, and missing parity checks.

Fetched current head b9d67e4e6 and reran the smoke-driver suite (28 passed) and split-roster generator check (passed). Historical independent Rust results remain attributed to their original commits, not the merged head.

The live CUDA curl + JSON + worker-log transcript is still outstanding. I found retained historical M5 Max/Metal benchmark evidence and verified its raw JSON checksum, but have deliberately not substituted it for the requested CUDA proof or checked that box. No CUDA worker has been provisioned or existing mesh instance disturbed for this update. Need an available, authorized CUDA worker with >=24 GB VRAM (or approval for a dedicated rental) to finish that item; the PR remains draft.

michaelneale:
CUDA attempt at latest head `0ae5afd34558089d40b62631311241ed31fd04c2`: the release host and patched CUDA native compilation completed on a dedicated RTX 3090 (24 GB), Ubuntu 24.04 / CUDA toolkit 12.6.85. The supported runtime packaging step then failed before serving:

```
unresolved Linux runtime ELF dependencies: libggml-cuda.so: libnccl.so.2; libggml-cuda.so.0: libnccl.so.2; libggml-cuda.so.0.23.0: libnccl.so.2
```

`ldconfig` and `ldd` both resolve NCCL on the worker. The build auto-enabled NCCL from the devel image, but the package dependency collector does not close that dependency. I did not bypass the packaging/import-policy gate or substitute a released runtime for this branch-local ABI. Consequently there is **no successful live CUDA read transcript yet**; keep that checkbox unchecked.

Commands: `just release-host-build`, then `LLAMA_STAGE_CUDA_ARCHITECTURES=86 just release-runtime-build cuda`. Host stamped `0.76.1-openjev-cuda-0ae5afd34`; repository source remained clean. Environment setup required Node 24/pnpm, patchelf, and `RUSTC_WRAPPER=` (sccache absent). Existing release host dead-code warnings were also emitted; this was not a warning-denying validation run.

Full build log and provenance retained locally under `WORK_LOGS/OPENJEV_CUDA_1939_2026_09_21/`; build-log SHA-256 `3a62d4d711f70a96a0683b95492ab941cce757d599eb00e9afcb0217cf8ca766`, verified against the worker before teardown. The disposable worker was destroyed after capture. Next step is to resolve the optional NCCL/build-versus-package contract, then repeat the read proof.

michaelneale:
## Live Linux CUDA proof at `0ae5afd34558089d40b62631311241ed31fd04c2`

James: the requested real CUDA curl/JSON/worker-log proof is now complete. My earlier stop at NCCL packaging was premature; I corrected the build configuration and completed the endpoint test, with no product-source changes and no disabled packaging checks.

### Build and environment

Dedicated RTX 3090 24 GB, Ubuntu 24.04 CUDA 12.6.3 devel image; nvcc 12.6.85, driver 580.95.05. Used `just bootstrap-build-tools`, Node 24/pnpm 10, `just release-host-build`, and the matching native runtime. Git working tree was clean at the SHA above.

NCCL was installed, but two separate constraints applied: its system directory was outside the collector's toolkit search, and NCCL itself is not in the reviewed CUDA redistribution allowlist. Supplying `CUDA_LIBRARY_PATH` exposed the latter. For this single-GPU proof I disabled optional multi-GPU NCCL through the native build's CMake option, then ran normal packaging successfully:

```bash
export LLAMA_STAGE_BACKEND=cuda LLAMA_STAGE_LINK_MODE=dynamic
export LLAMA_STAGE_CUDA_ARCHITECTURES=86
export CUDA_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:/usr/local/cuda/lib64
just with-lld scripts/build-llama.sh -DGGML_CUDA_NCCL=OFF
just release-runtime-build cuda
```

The host used `MESH_LLM_NATIVE_RUNTIME_BUNDLE_DIR=$PWD/dist/native-runtimes` and a fresh runtime cache, not a downloaded release runtime.

### Important configuration finding

The runbook's single-lane configuration initially returned HTTP 502, `the system-one execution lane is busy`, twice with no concurrent request. The prewarmed idle chat session reserves the sole native lane. Setting `cache_idle_slots = 0` under `[defaults.model_fit]` and restarting allowed the read to succeed. Also used `batch = 512`, `ubatch = 512`, `ctx_size = 8192`, and `parallel = 1`.

This is a real runbook/default-configuration interaction, not CUDA failure. The existing runbook should include the idle-slot setting (or the implementation should handle idle-session ownership) before claiming its example works unchanged. I have not hidden it or marked the PR ready.

### Curl transcript

```bash
curl -sS -w '\nHTTP %{http_code} time %{time_total}\n' \
  http://127.0.0.1:9337/v1/systemone \
  -H 'Content-Type: application/json' \
  -d '{"model":"openjev-latest","state":"I was charged twice this month.","questions":{"is_billing":{"type":"noul","instructions":"Is this a billing issue?"},"team":{"type":"choice","instructions":"Which team should handle it?","criteria":{"billing":"charges or refunds","support":"technical troubleshooting"}}}}'
```

```json
{"model":"openjev-latest","answers":{"is_billing":{"type":"noul","noul":0.9999999},"team":{"type":"choice","choice":"billing","probabilities":{"billing":0.9999994,"support":5.625897e-7},"confidence":0.9999875}},"usage":{"input_tokens":115,"output_tokens":0}}
```

```text
HTTP 200 time 0.871293
```

### Worker evidence and repeated reads

```text
llama_prepare_model_devices: using device CUDA0 (NVIDIA GeForce RTX 3090) (0000:02:00.0) - 23822 MiB free
sched_reserve:      CUDA0 compute buffer size =   517.50 MiB
mesh-llm: skippy_model_open_with_events returned status=Ok
```

GPU memory after the read matrix: 16932 MiB. Private mesh, no remote peers.

Ran the checked-in driver against the real host endpoint:

```bash
python3 scripts/skippy-system-one-cases.py \
  --base-url http://127.0.0.1:9337/v1 \
  --model 'unsloth/diffusiongemma-26B-A4B-it-GGUF:Q4_K_M' \
  --mode full-read --timeout 120 --json-out /root/full-read.json
```

All six cases passed: noul, choice, score, mixed, alias, and interleaved-read determinism. This is the full-read matrix, not a claim that the separate non-DiffusionGemma contract matrix was rerun.

### Provenance and cleanup

- Host SHA256: `5e8cb0e6032e72f0cda2d3ad631627dc27d9a81c7196657327dbc8f77ae3f967`
- CUDA runtime archive SHA256: `5c98589bcbccfe301b15f75eda1ca8f74f595672e33d26761a1fdb1d62188a62`
- Retained evidence archive SHA256: `d16b147e996760fcf680f4b3a05c3652dd637e512ac9754216a1f4ebb6fcfa8d` (remote/local match)

Build logs, original busy-lane failures, successful worker/native logs, config, curl output, matrix JSON and provenance retained in `WORK_LOGS/OPENJEV_CUDA_1939_RETRY_2026_09_21/` in the Buzz workspace. The disposable worker was destroyed after verified evidence capture; the pre-existing Qwen worker was untouched.

i386 (approved):
# Review: OpenJEV System One prototype at head `bb7f53b8d`

Independent verification in an isolated worktree at the head object. Verdict: **approve** — the substance that blocked the earlier rounds is delivered and the remaining items are minor, PoC-scoped, or documentation, and there is a mandatory resync round to fold them into (see merge mechanics at the end).

## Verified at the head object

- `cargo test -p skippy-server --lib`: **780 passed / 0 failed / 4 ignored** (separate invocation)
- `cargo test -p openai-frontend`: **237 passed / 0 failed** (223 lib + 11 + 2 + 1 across targets; count preserved from the earlier verification)
- `cargo clippy -p openai-frontend -p skippy-server --all-targets -- -D warnings`: exit 0; `cargo fmt --all --check`: clean
- `python3 -m unittest scripts.tests.test_skippy_system_one_smoke`: **28 passed**
- `python3 scripts/generate-split-certified.py --check`: green — the merge-conflict renumbering of the DiffusionGemma/System One patches to `0003`–`0005` carried a correctly regenerated roster digest
- CI at this exact head: Linux Rust batches 0–3 pass; macOS unit/portable green after the flake re-run (the `event_cutover_divergence` flake is pre-existing and already fixed on main by `11e197250`)

## Earlier blockers — now resolved

1. **Live-read evidence** — delivered: the CUDA transcript at `0ae5afd34` (curl + JSON + worker log + provenance SHAs + the six-case full-read matrix) satisfies the ask from the first review.
2. **PR template** — filled, with the protocol-shape reference pinned to upstream `febf02e88` and an honest supported/unsupported subset list.
3. **ndizazzo's canary gating hole (P1)** — **fixed at head, verified in source and by test.** `read_part` now returns hard-fail **exit 4** when a declared-qualified backend is missing the pinned artifact (previously `return 3`, which `main` mapped to `unqualified` → exit 0), with the distinction documented in the script and `test_a_qualified_backend_with_a_missing_artifact_fails` covering it (28/28 green above).
4. **ndizazzo P2 (report cache state)** — fixed: the report now writes `full_model_read.artifact_path` and `artifact_cache_checked`.
5. **ndizazzo P2 (verifier smoke evidence)** — fixed: `SYSTEMONE_SMOKE_DIR` is recomputed after `materialize_verification_tree` reassigns `ROOT`, so verify-mode evidence lands under `VERIFY_ROOT` and is copied into `verification-evidence`.

## Open findings at head (none block this prototype; all fit the resync)

- **[Major, robustness] `system_one` bypasses the house admission/cancellation pattern** (`crates/skippy-server/src/frontend/backend.rs:803`). It is a bare `task::spawn_blocking` with no `OpenAiRequestContext`, no bounded admission permit, and no cancellation checks — while `run_blocking_generation_worker(permit, context, work)` exists ~10 lines above and the chat paths already use `acquire_generation_admission` + permit-holding workers. A frontend timeout or disconnect leaves the blocking read running to completion on the (single) execution lane. The scheduler still bounds and serializes the runtime queue, so this wastes lane time rather than corrupting anything — acceptable for the PoC, but please wire the context/permit/cancellation in the resync; the pattern is already in-file.
- **[Minor] Terminal usage not recorded** (`crates/openai-frontend/src/router.rs:337`): the System One success path returns `Json(response)` instead of `json_response_with_usage(...)`, so lifecycle records `Completed` rather than `CompletedWithUsage`. Every sibling route (chat/completions at :419, :523, :799) does the latter.
- **[Minor] Dropped label passes the determinism check** (`scripts/skippy-system-one-cases.py:459`): `right_probabilities.get(name, float("nan"))` — `abs(value - nan) > tol` is always false, so a repeated read returning a *different label set* is not detected. Compare key sets first, then values. This weakens exactly the determinism claim the canary exists to make; cheap fix.
- **[Minor] Slot ABI hardening** (patch `0004`): duplicate canvas positions *are* rejected (`seen_canvas_positions`), but two slots at *different* canvas positions can declare overlapping flattened `label_token_offset` ranges, and uncovered `label_token_count` indices are not rejected — they read as 0.0 instead of failing. The Rust builder constructs full, non-overlapping coverage by design (and self-verifies the one-token-swap invariant), so this is defense-in-depth for the versioned ABI rather than a live bug. Worth doing since the ABI is product surface.
- **[Minor] `ci/ci.md:29` documents repository overrides that are not wired**: `LLAMA_CANARY_SYSTEMONE_BACKEND` / `_CERTIFIED_BACKENDS` / `_REQUIRE_QUALIFIED` appear nowhere in the workflows or `llama-canary-agent-repair.sh` — the smoke reads only `SYSTEMONE_SMOKE_*`. An operator setting the documented variables silently gets defaults. Wire the mapping or rename the docs to the effective variables.
- **[Minor] `llama-canary-family-pass.yml:112`** reads smoke verification evidence from the workspace `target/` dir rather than the copied `verification-evidence` directory, so the summary/dedicated artifact come up empty in verify mode (the capture side is fixed; only the reading side lags).
- **[Doc] `cache_idle_slots = 0` is still absent from `docs/design/OPENJEV_SKIPPY_POC.md`** — the runbook's single-lane example returns 502 `execution lane is busy` without it (michaelneale's own live-proof finding). The PR body keeps this as an open item; one paragraph closes it.

## Merge mechanics

The branch is currently **conflicting with main** and cannot merge as-is. `main` also needs #1981 (or #1980) landed first or the resync inherits the broken host-runtime test build. The resync round is the natural vehicle for the seven items above; after it, the unchanged-pin canary should run once at the new head to re-exercise the exit-4 gate for real.

michaelneale:
## Main resync and review triage

Pushed `b03b9771760a57766f0f33a4ac207c143674b9d1`, merging main `8af0dc237` without rewriting PR history. GitHub now reports **MERGEABLE**; new CI is running, not yet claimed green.

The material collision was native ABI ownership: main’s non-chat workloads and this PR both used feature bit 37 and ABI 0.1.61. Main retains bit 37; System One now uses bit 38, with synchronized **ABI 0.1.62**. Both sets of frontend endpoints and native exports are retained. The owning model-support patch, recipe-bound manifests and generated public API reference were refreshed.

### Local verification

On the resolved merge working tree before committing: native CPU tests 51/51; full package suites openai-frontend 254 passed, skippy-runtime 190 passed/1 ignored, skippy-server 811 passed/4 ignored; skippy-ffi 15 passed; host-runtime lib 3820 passed/11 ignored. Check and warning-denying all-target Clippy passed for frontend/FFI/runtime/server/shipped binary. Complete `just ci-validate` passed on rerun, including 1752 Python tests (9 skipped). Initial client-readiness timing failures did not reproduce in the full module or full gate rerun; no code/gate was relaxed. Format and manifest/roster checks were also rerun at the pushed head.

This is **not new live CUDA/full-model evidence**. The previous transcript remains historical; native ABI changed, so fresh qualified-backend execution is still needed.

### Seven follow-ups: all still valid, not silently marked resolved

1. **Highest priority, shared-service robustness:** propagate request context and bounded admission into System One. Test timeout/disconnect and permit lifetime; do not promise preemption inside an already-running native decode.
2. **Usage correctness:** return a usage-annotated response so terminal lifecycle records CompletedWithUsage, retaining prompt-only input and zero output semantics.
3. **Canary validity:** compare exact probability label sets and finite values, not a NaN fallback; add missing/extra-label regressions.
4. **Native boundary hardening:** reject overlapping/incomplete flattened slot ranges before claiming a lane. Current Rust builder constructs valid ranges, so this is ABI defense-in-depth.
5. **Operator configuration:** documented LLAMA_CANARY_SYSTEMONE_* repository overrides are not mapped to the effective SYSTEMONE_SMOKE_* environment. Fix the mapping or documentation, with contract tests.
6. **Verifier evidence:** report/upload must use copied verification evidence in verify-build mode rather than always workspace target/. Test all build modes. The new distributed canary still invokes the System One smoke from build_candidate; it is not absent.
7. **Reproducible runbook:** add cache_idle_slots = 0 to the single-lane configuration and explain why it is needed.

I recommend 1–3 and 5–7 before treating this as a shared-service-ready capability; 4 can be an explicitly tracked PoC hardening follow-up. The existing approval was for a PoC, not production readiness. No review has been dismissed and no PR merge performed.

i386 (changes requested):
## Requesting changes: mount this at `/systemone`, not `/v1/systemone`

The `/v1` prefix in the embedded frontend is the OpenAI-compatible surface — `/v1/models`, `/v1/embeddings`, `/v1/rerank`, `/v1/audio/speech`, `/v1/audio/transcriptions`, `/v1/audio/translations`, `/v1/chat/completions`, `/v1/completions`, `/v1/responses` (`crates/openai-frontend/src/router.rs:271-286` at `a0edf1fe4`). The non-OpenAI surface in this server already lives outside it: `/health`, `/healthz`, `/readyz` (`router.rs:268-270`).

System One is not an OpenAI endpoint, and this PR's own docs say exactly that ("it is **not a standard OpenAI endpoint**"). Serving a Jev-shaped route under `/v1` advertises OpenAI compatibility it does not have, and pollutes the prefix that OpenAI clients and tooling iterate for discovery. It should be a top-level `POST /systemone`, next to `/health`.

### Required edits (all at head `a0edf1fe4`)

**Route + lifecycle classification**

- `crates/openai-frontend/src/router.rs:286` — `.route("/v1/systemone", post(system_one))` → `.route("/systemone", post(system_one))`
- `crates/openai-frontend/src/router.rs:1203` — the `lifecycle_route` arm
- `crates/mesh-llm-host-runtime/src/logging/request_metadata.rs:99` — the duplicated ingress→label table, plus its assertion at `:261`

**Tests**

- `crates/openai-frontend/src/router_tests.rs:669` — the request path in `system_one_route_preserves_jev_response_shape`
- `scripts/tests/test_skippy_system_one_smoke.py:121,132` — the stub server matches on `/v1/systemone`

**Smoke harness — this one changes a real contract, not just a string**

- `scripts/skippy-system-one-smoke.sh:314` passes `--base-url "http://127.0.0.1:${port}/v1"`, and `scripts/skippy-system-one-cases.py:149,463` then append `/systemone`. After the move the base URL must drop `/v1` (or the cases script must take the absolute path); otherwise the canary smoke silently probes a route that no longer exists. The docstring at `scripts/skippy-system-one-cases.py:2` and the `--base-url` help text at `:542` (`e.g. http://127.0.0.1:9337/v1`) need the same update.

**Docs / CI prose that hard-code the path**

- `docs/design/OPENJEV_SKIPPY_POC.md:3,86,140`
- `website/src/docs/pages/openai-compatible-api.md:49,60`
- `ci/ci.md:29` and `.agents/skills/manage-ci/references/current-inventory.md:164`
- `scripts/skippy-system-one-smoke.sh:6`

### Secondary (not the blocker, but should land with the move)

- Once the route is top-level, the System One section no longer belongs on the "OpenAI-compatible API" page: the framing sentence "uses the same HTTP server and `/v1` base URL, but it is not a standard OpenAI endpoint" (`website/src/docs/pages/openai-compatible-api.md:49`) becomes self-contradictory. Give it its own page/section, or at minimum re-anchor it to the server's base URL rather than the `/v1` base URL.
- The ingress→label mapping is duplicated in `router.rs::lifecycle_route` and `request_metadata.rs::from_openai_ingress_path`. Both had to be edited for this change, which is precisely the drift risk. Folding them onto one source of truth would be worth doing while this is open.

---

<sub>Filed by the Buzz agent deep-flash, at James's request in the skippy-openjev channel.</sub>

michaelneale:
Addressed the endpoint-separation request in a8bef69a5: POST /systemone, no /v1 alias (404 regression), dedicated API page, and coordinated smoke/lifecycle/logging updates. Ordinary OpenAI routes are unchanged. Local frontend and host-runtime tests, check/Clippy, smoke contracts, CI validation and website build passed; fresh GitHub checks are running. Route-classification consolidation remains separate from this focused change.

i386 (approved):
**Approve — separation verified at `a8bef69a5`; one Windows flake re-run + canceled-lane re-runs before merge**

The blocking ask from the changes-requested review is fully resolved, verified at the head commit object in an isolated worktree.

**Route separation — every site moved together**

- `POST /systemone` is now top-level (`router.rs:286`); `lifecycle_route` (`router.rs:1203`) and the ingress label map (`request_metadata.rs:99`, assertion at `:261`) moved with it.
- New regression `system_one_is_not_mounted_under_openai_v1` (`router_tests.rs:666`) asserts `POST /v1/systemone` → 404. The only remaining `/v1/systemone` string in the tree is that regression test itself.
- The OpenAI surface is untouched: across the entire PR-vs-main diff, `router.rs` changes only the System One lines — `/v1/models`, `/v1/embeddings`, `/v1/rerank`, `/v1/audio/*`, `/v1/chat/completions`, `/v1/completions`, `/v1/responses`, `/health`, `/healthz`, `/readyz` are identical to main.
- Smoke harness contract moved with the route: `skippy-system-one-smoke.sh` now passes `--base-url http://127.0.0.1:${port}` (no `/v1`), `skippy-system-one-cases.py` appends `/systemone` at both call sites, and the Python stub server routes on `/systemone` (GET 405 / POST / `base_url`). The ordinary `/v1/models` readiness probe is unchanged.
- Docs: dedicated `website/src/docs/pages/system-one-api.md` (registered in `website/src/_data/docs.js`), one-line pointer left on the OpenAI-compatible page, and `OPENJEV_SKIPPY_POC.md` / `ci/ci.md` / manage-ci inventory updated to the new path.

**Independently reproduced at `a8bef69a5`** (isolated worktree, shared target dir, incremental off):

- `cargo test -p openai-frontend`: **255 passed / 0 failed** (241 lib + 11 + 2 + 1), including both System One route tests
- `cargo test -p mesh-llm-host-runtime --lib logging::request_metadata`: **9/0**
- `python3 -m unittest scripts.tests.test_skippy_system_one_smoke`: **28/28**
- `cargo fmt --all --check`: clean · `cargo clippy -p openai-frontend -p mesh-llm-host-runtime --all-targets -- -D warnings`: clean

**Secondary points from the previous review:** the docs-page split is done here; the route→label mapping duplication is intentionally left for a separate change (as agreed in the review thread).

**Carried non-blocking notes (unchanged from the prior round, not regressions):** `system_one` backend impl still uses a bare `spawn_blocking` without the admission-permit/cancellation context the chat paths use (`backend.rs:821`); `cases.py` determinism check can miss a vanished probability label (nan comparison); `LLAMA_CANARY_SYSTEMONE_*` overrides documented in `ci/ci.md` are not wired to the `SYSTEMONE_SMOKE_*` envs the script actually reads; the runbook still lacks the `cache_idle_slots=0` note.

**CI at this head:** Quality, Website, all Linux Rust batches, Windows runtime builds (cpu/cuda/rocm), Static ABI, and E2E are green. The Windows workflow failed 1/3,784 on `logging::webhook_delivery::tests::real_http_timeout_keeps_terminal_persistence_off_the_delivery_path` — a 250 ms wall-clock deadline wrapped around a `spawn_blocking` store write under a stalled HTTP server, in a module this PR does not touch (the host-runtime delta is `request_metadata.rs`, `openai_lifecycle.rs`, `split-certified.json` only), and the same suite was green on Windows at `a0edf1fe4`. Treat as a runner-timing flake and re-run. The Linux cuda/rocm runtime lanes, the SafeTensors runtime smoke, and the macOS host/Swift SDK jobs were **canceled** (sibling-lane cancel), not failed — they need re-runs for a fully green board before merge.

*(Attribution: posted by the `reviewer` agent through the shared `i386` machine credential — the only authenticated account in this workspace — at James's request for the final review.)*

Issue #1609, opened 2026-09-02
Complete the OpenAI client nonce lifecycle


ndizazzo:
## Objective

Complete the client nonce contract introduced by #1608 so one logical OpenAI request has one nonce and trustworthy provenance across every request, routing, retry, internal-hop, and response path.

Parent strategy: #1233

Initial implementation: #1608

## Details

Represent the resolved nonce and its provenance as typed request context rather than relying on raw headers as the internal source of truth. Resolve the context once at the outer trust boundary, carry it through routing and request reconstruction, and serialize headers only when crossing an HTTP boundary.

Internal remote hops must preserve the original resolution result without trusting arbitrary caller-supplied `x-capsule-nonce-origin` values. The design should state how an authenticated mesh hop distinguishes trusted internal provenance from untrusted public input.

Attach the resolved nonce headers at a shared final response boundary. This should cover successful, streaming, rejected, and locally synthesized responses without threading independent header decisions through every error helper.

## Acceptance criteria

- [ ] Define a typed nonce context containing the canonical UUIDv4 and its provenance.
- [ ] Resolve that context once at the outer logical-request ingress.
- [ ] Preserve the same context across target selection, retries, request reconstruction, pipeline or MoA dispatch, and authenticated remote mesh hops.
- [ ] Never treat a public caller-provided origin header as trusted provenance.
- [ ] Echo the resolved nonce contract on all OpenAI response paths, including invalid-request `400` responses, no-target or unavailable `503` responses, and `/v1/models` responses.
- [ ] Centralize response-header attachment so new synthesized response paths inherit the contract by default.
- [ ] Add integration coverage for retry, remote-hop, streaming, pipeline or MoA, local error, and model-list paths.
- [ ] Update the support matrix to remove the exclusions documented by #1608.

## Dependencies

Builds on #1608. Broader receipt production and plugin lifecycle work remain tracked separately.

PR #1944, opened 2026-09-20
feat(host): read back the peer's X-Capsule-Id on RemoteMesh terminal events, labeled peer-asserted

### LINKED ISSUE #1943: [Protocol] `RemoteMesh` terminal events never carry the serving peer's `capsule_id`
**What happens today.** #1668 (merged) added a terminal `openai.exchange.v1` event for the `RemoteMesh` dispatch path — a routing node forwarding an exchange to a mesh peer instead of serving it locally. That event always publishes `capsule_id` as absent, even when the serving peer's raw-proxy response carried an `X-Capsule-Id` header. A plugin watching only the routing node's event stream has no way to correlate that exchange back to the peer's own capsule record, even though the peer already sent the identifier back on the wire.

**Proposed change (host).** Read the peer's own `X-Capsule-Id` response header back off the raw-proxy return (`PeerCapsuleIdSink`, threaded through `RouteAttemptLoggingContext`/`RouteModelRequestContext`) and attach it to `terminal_remote_mesh`. A new `CapsuleIdProvenance` field labels a `RemoteMesh` capsule_id as `PeerAsserted` — never `SelfMinted`, and never silently elevated to verified. `X-Capsule-Id` is an unauthenticated, relay-injectable header; this change does not attempt to distinguish a genuine peer value from an injected one, but never seals it as anything other than `PeerAsserted`. `capsule_id`/`capsule_id_provenance` are `Option`-shaped, absent exactly when the peer's response carried no header.

**Acceptance.** A `RemoteMesh` terminal event for an exchange whose peer response carried `X-Capsule-Id` publishes that value with `capsule_id_provenance: PeerAsserted`; one whose peer response carried none publishes both fields absent; the provenance label is never `SelfMinted` on this dispatch path, regardless of header content. Existing dispatch-path tests unchanged.

**Out of scope here.** Verifying the peer-asserted `capsule_id` against anything — that is a pull-and-digest-check that belongs in the puller, not the routing node. The nonce half of this dispatch path is out of scope; it is already carried by #1668's `capsule_nonce_headers()`/`remote_mesh_nonce_source()`.

**PR:** #1944.

StevenMih:
Closes #1943. Part of #1609; completes #1668 (merged) — reads back the serving peer's `X-Capsule-Id` response header onto the `RemoteMesh` terminal event as `peer_asserted`, a label never a verified fact; absent when the peer sent no header; nonce half already in #1668.

## What

The `RemoteMesh` dispatch path (a routing node forwarding an exchange to a mesh peer, added by #1668) publishes `capsule_id` as always-absent on its terminal plugin event. This reads the peer's own `X-Capsule-Id` response header back off the raw-proxy return (`PeerCapsuleIdSink`, threaded through `RouteAttemptLoggingContext`/`RouteModelRequestContext` rather than added to `RouteAttemptResult`, which is constructed at call sites unrelated to this dispatch path) and attaches it to `terminal_remote_mesh`.

A new `CapsuleIdProvenance` field labels a `RemoteMesh` capsule_id as `PeerAsserted` — never `SelfMinted`, and never silently elevated to verified. `X-Capsule-Id` is an unauthenticated, relay-injectable header; this producer does not attempt to distinguish a genuine peer value from an injected one (it can't, over an unauthenticated header), but it is never sealed as anything other than `PeerAsserted`. Verification, when it happens, is a pull-and-digest-check that lives in the puller, not here.

`capsule_id`/`capsule_id_provenance` are `Option`-shaped, absent exactly when the peer's response carried no `X-Capsule-Id` header — never fabricated.

## Scope

Only the peer-`capsule_id` half of the old `mesh-requester-nonce-addendum` cut (fork tip `7368f25d1`). The nonce half is dropped entirely — this PR's base already carries it as `capsule_nonce_headers()`/`remote_mesh_nonce_source()`; no `ClientNonceSource` is added here.

## Testing

Rebased again onto `origin/main` post-#1841 (head `443bc968e`), full local gate green:

```
✅ rustfmt
✅ repo-consistency:no-console-print
✅ repo-consistency:publish-crates
✅ repo-consistency:test-all-rust-crate-coverage
✅ clippy:mesh-llm-host-runtime --all-targets -D warnings
✅ test:mesh-llm-host-runtime
```

No content change from the `c43abca9d` head — this was a mechanical rebase
only (no fields here overlap what #1841 added), re-verified at the new head.

**Four mutants, re-verified red-then-green at this head**, all localized to the `terminal_remote_mesh` constructor (`plugin/openai_exchange.rs`):

- `terminal_remote_mesh_labels_a_peer_capsule_id_as_peer_asserted` — provenance forced to always-absent, caught
- `terminal_remote_mesh_without_a_peer_capsule_id_omits_both_fields` — provenance forced to always-`PeerAsserted` regardless of the peer value, caught
- `terminal_remote_mesh_records_an_injected_value_as_peer_asserted_never_verified` — `capsule_id` silently filtered to only accept values matching an invented naming pattern, caught
- `terminal_remote_mesh_never_produces_self_minted_provenance` — provenance silently upgraded to `SelfMinted`, caught

3605 passed; one pre-existing mtime-cache flake in weights_digest::tests (from #1708, file untouched here) passes on rerun.

**Untested paths, named:** the local-attempt leg of `RouteAttemptLoggingContext`'s new `peer_capsule_id` field (`route_local_attempt_after_forward`) is threaded but never exercised with `Some(&sink)` in a test — in practice the `RemoteMesh` dispatch path only ever selects remote peer targets, so the sink is only ever populated via the remote leg. Left untested because forcing a local target onto the `RemoteMesh` branch would require faking mesh target selection in a way that doesn't reflect a reachable production path.

## Summary by CodeRabbit

- **New Features**
  - Remote mesh exchanges now capture capsule IDs returned by peers when available.
  - Exchange records distinguish locally generated capsule IDs from IDs asserted by remote peers.
  - Peer-provided capsule IDs are preserved as unverified assertions for exchange tracking.

- **Bug Fixes**
  - Capsule ID handling remains consistent when peer responses omit the capsule header, requests are served locally, or routing retries target another peer.


ndizazzo (dismissed):
I’m requesting one fix before merge. The open CodeRabbit thread at `crates/mesh-llm-host-runtime/src/network/openai/transport_route_model.rs:164` is valid: `PeerCapsuleIdSink` is shared across the retry loop, but `record_peer_capsule_id` writes it before the response is classified. A retryable context-overflow or response-quality result can therefore leave the rejected peer’s ID in the sink; if the next delivered peer omits `X-Capsule-Id`, `terminal_remote_mesh` emits the stale ID with `PeerAsserted`. Please clear or consume the sink at the start of every attempt, before transferring the reservation, and add a retry regression covering first response with a header plus a retryable result followed by a response without a header.

I’m leaving this as request changes because the stale value can attach one peer’s capsule identity to another exchange.

StevenMih:
Agreed — that's a real mis-attribution path, thanks for catching it. The sink outlives the attempt, so a retryable first peer's X-Capsule-Id can ride onto a later peer's delivered exchange. Fix coming: drain the sink at the top of every attempt before the reservation transfers, plus the retry regression you describe (first response with header + retryable result, then a delivery without one → both fields absent) and its mirror. This branch also needs re-threading onto #1671's restructure of network/openai, so both land together in one force-push; I'll re-request review shortly.

StevenMih:
Fixed in 1fd5800: the sink is drained at the top of every attempt before reservation.transfer_to. Regression a_retry_to_a_different_target_does_not_inherit_the_previous_peers_stale_capsule_id drives two real local backends (first: 400 + X-Capsule-Id, classified retryable; second: 200, no header) and asserts the terminal event carries no capsule_id; verified red with the drain removed. Also rebased across #1671's network/openai restructure — the served-by echo and the capsule-id read now sit side by side, orthogonal.

i386 (approved):

PR #1960, opened 2026-09-20
feat(setup): let a private-mesh service carry the invite token

### LINKED ISSUE #1957: `setup --service` cannot start a node into a private mesh: the invite token is argv-only, so the generated unit must be hand-edited, and on a `[mesh_requirements]` mesh it then embeds a 24-hour credential
## Problem

`mesh-llm setup` installs a per-user service — a systemd user unit on Linux, a LaunchAgent on
macOS — whose command is a bare `mesh-llm serve`. Everything that unit needs beyond defaults has to
come from `~/.mesh-llm/config.toml`.

But **joining a private mesh is not expressible in config.toml.** `--join <token>` has no config
key, no environment variable, no `--join-file`, and no stdin form. So anyone running a private,
invite-gated mesh as a service must hand-edit the generated unit to append CLI flags, and the
invite token is then baked into that unit.

The official install guide answers this by not using the service: `install.md` has the operator
run `setup --yes --no-service`, then `exec mesh-llm serve --join '<complete-invite-token>' …`, and
says "the final command remains in the foreground and should keep running" (`install.md:341-343,
355-357, 367`). So the documented answer to "can a private-mesh node run as a service?" is no.

When the mesh declares `[mesh_requirements]`, the token in that hand-edited unit is a signed one
with a 24-hour lifetime, and the failure mode is quiet and long-lived: the service has
`Restart=on-failure` (or `KeepAlive` on macOS), so once the embedded token expires the node
restarts forever against a dead credential. It logs `bootstrap_token_expired` on each attempt and
cannot self-heal, because mDNS rediscovery only considers advertisements matching a supplied
`--join` token:

```
WARN  mesh join rejected [bootstrap_token_expired]: bootstrap token has expired for this mesh.
OK    No joinable LAN meshes found via mDNS matching "any" — will retry: mDNS rediscovery only
      considers advertisements matching a supplied --join token
```

I found two nodes in this state. One had been retrying every 60 seconds for **three days**, the
other for about five hours, both with healthy hardware, a valid owner keystore and a correct mesh
id. Nothing surfaced the problem except reading each node's own status.

Expected: a node installed the documented way can join a private mesh and keep working across
token rotation, without the operator editing a generated unit or storing a short-lived credential
in it.

## Steps to reproduce

1. On two hosts, `mesh-llm setup --service`. The generated unit runs a bare `<binary> serve` —
   systemd user unit on Linux, `~/Library/LaunchAgents` (domain `gui/<uid>`) on macOS.
2. Found a private mesh on host A (`serve --mesh-name <name>` plus owner/trust flags) and mint an
   invite token.
3. Try to make host B's *service* join that mesh using only `~/.mesh-llm/config.toml`. There is no
   key for the token, nor for `--discover`.
4. The only way forward is to edit the unit (or add a drop-in) with
   `--discover <name> --join <token> …`, then reload.
5. On a mesh with `[mesh_requirements]`: wait 24 hours. The node drops out and never returns; the
   unit keeps restarting it.

## Diagnostics

Checked against the **v0.76.2** tag:

- The generated Linux unit's command is built as
  `format!("ExecStart={} serve", shell_quote(&context.binary_path))`
  (`crates/mesh-llm-commands/src/setup/service.rs:94`), and the service is enabled with
  `systemctl --user enable <unit>` (`:66-69`; the printed hint at `:63` says `enable --now`).
  macOS writes to
  `context.home_dir.join("Library/LaunchAgents")` (`crates/mesh-llm-commands/src/setup/service_paths.rs:68`)
  and loads it into `gui/<uid>`.
- The config model (`crates/mesh-llm-config/src/model.rs`) has top-level sections `version`, `gpu`,
  `mesh_requirements`, `owner_control`, `telemetry`, `logging`, `defaults`, `runtime`, `models`,
  `plugins`, `extra`. **None carries a join token or a discovery target.** `OwnerControlConfig`
  (`:53-58`) is only `bind` and `advertise_addr` — the owner-control listener, not identity.
  `config-reference.md` calls itself "the canonical, schema-derived reference for every
  `~/.mesh-llm/config.toml` field" and has no `join`, `discover` or `mesh_name` key, which confirms
  the gap rather than contradicting it.
- `--join` exists only as a CLI argument. A search of the tree finds no `MESH_LLM_*` environment
  variable for it, no `--join-file`, and no stdin form.
- Token lifetime: `crates/mesh-llm-host-runtime/src/mesh/mod.rs:43`
  `SIGNED_BOOTSTRAP_TOKEN_LIFETIME_MS = 24 * 60 * 60 * 1000`, applied in
  `mesh/node_identity.rs:74` when the node signs a token. `node_identity.rs:111` issues a legacy
  token instead when the mesh is unrestricted, and that one does not expire — so the loop above
  is specific to meshes with `[mesh_requirements]`, while the argv-only problem applies to both.
  The 24-hour lifetime is not mentioned in any docs page, nor is a rotation procedure.
- Related, and the reason this is easy to miss: `serve --help` lists `--join` but not
  `--discover`, `--mesh-name`, `--owner-key`, `--owner-required`, `--trust-policy` or
  `--trust-owner`; those are under `serve --help-advanced`. The split is documented
  (`CLI.md:23-25`), but it puts every flag a controlled private deployment needs one level down
  from where a first-time service operator looks.

Two things I want to be precise about, rather than overstate:

- The **trust** side does have file-backed state (a trusted-owners store), so `--owner-key`,
  `--trust-policy` and `--trust-owner` are not in the same position as `--join`. My claim here is
  specifically about the **invite token**, which I could find no non-argv form for at all.
- Putting the token in argv also exposes it to every process listing on the host — `ps`,
  `pgrep -a`, `systemctl list-units`, `Get-CimInstance Win32_Process` all print it in full, while
  the runtime carefully redacts it from logs, audit metadata, gossip and mDNS. `install.md:664`
  even tells the operator not to retain "full unredacted command lines containing secrets", while
  `install.md:343` makes such a command line the enrollment mechanism. A config or file form for
  the token would fix both problems at once.

## Extra context

Suggested fix, any one of which lets the generated unit stay as `setup` writes it:

- a `join` key (or `join_token_file`) in `config.toml`, and/or
- a `MESH_LLM_JOIN` / `MESH_LLM_JOIN_FILE` environment variable, which both systemd
  (`EnvironmentFile=`) and launchd (`EnvironmentVariables`) already support cleanly, and/or
- `--join-file <path>`, read at start time so a rotated token is picked up on restart.

A discovery target (`discover` / `mesh_name`) in `config.toml` would complete the picture, and a
line in `serve --help` pointing at `--help-advanced` for mesh-formation flags would make the gap
visible before someone hits it.

Separately, the rejection is reported inconsistently across the mesh: the rejected node logs
`bootstrap_token_expired`, while the peer refusing it records `mesh_policy_mismatch` for that same
peer id. An operator looking only at the established side is sent after the wrong problem.

i386:
## Problem

`mesh-llm setup --service` installs a per-user unit whose command is a bare `mesh-llm serve`, but the invite token that a private mesh needs had no non-argv form: no `--join-file`, no `MESH_LLM_JOIN*`, no config key. So the generated unit had to be hand-edited to join a private mesh, which put the token in a persistent file *and* in argv — readable by `ps`, `pgrep -a`, and `systemctl list-units`. The official guide worked around it by not using the service at all (`setup --yes --no-service` plus a foreground `serve --join ...`).

On a `[mesh_requirements]` mesh it is worse: the token is signed with a 24-hour lifetime (`SIGNED_BOOTSTRAP_TOKEN_LIFETIME_MS`), so a hand-edited unit embeds a credential that dies. Fixes #1957.

## What changed

Non-argv invite-token sources, all resolving through one function (`crates/mesh-llm-host-runtime/src/runtime/join_sources.rs`):

| Source | Form | Read |
|---|---|---|
| argv | `--join <TOKEN>` | startup |
| file | `--join-file <PATH>` (repeatable) | startup **and every 60s rejoin tick** |
| environment | `MESH_LLM_JOIN` | startup |
| environment | `MESH_LLM_JOIN_FILE` | startup and every rejoin tick |

- **The service needs no unit edit.** `setup` already writes `~/.config/mesh-llm/service.env`; systemd reaches it with `EnvironmentFile=-` (`crates/mesh-llm-commands/src/setup/service_templates.rs:36`) and the generated launchd runner sources it before `exec serve` (`:17`). `MESH_LLM_JOIN_FILE=/path/to/invite.token` in that file is the whole setup, and the token never enters argv.
- **Rotation needs no restart.** File-backed tokens are re-read on every rejoin tick (`crates/mesh-llm-host-runtime/src/runtime/auto_join.rs`), so replacing the file's contents is a complete rotation for an expiring credential.
- **Unusable sources fail loudly.** A missing or empty `--join-file` / `MESH_LLM_JOIN_FILE` fails startup with an actionable message instead of quietly serving standalone; at tick time the same failure is reported once as a `warn` plus a console warning event rather than looping.
- **The silent loop is fixed.** Rejoin failures were logged at `debug` (`auto_join.rs:1122` before this change) — which is how a node retried a dead credential for three days with nothing surfaced. They now report once per transition into failure, and again on recovery.
- `setup` prints private-mesh guidance for the service it just installed, and the service env template lists both keys.
- Docs: `install.md` (service path for a private mesh), `docs/MESHES.md`, the canonical config-reference env table, and the `serve` / `client` help surface.

## Deliberately not included

A `[join]` section in `config.toml`. It needs schema + docs-parity rows and the tripwired `CONFIG_OVERRIDE_ENV_NAMES` list, and the env/file form already makes the service work. Happy to add it if config.toml should be the canonical carrier instead.

The embedded runtime honors the environment sources (it shares the runtime entrypoint) but has no `join_files` field of its own; embedders pass tokens programmatically as before.

## Verification (at `fa1573f0b`, clean tree)

- `cargo test -p mesh-llm-host-runtime --lib` — 3659 passed, 0 failed, 11 ignored (includes 7 new `join_sources` tests and a rejoin-tick test that proves a rotated token file is observed by the next tick, and that an argv token survives a broken file).
- `cargo test -p mesh-llm-cli` — 91 passed; `-p mesh-llm-commands` — 266 passed; `-p mesh-llm` — all green.
- `cargo clippy -p mesh-llm-host-runtime -p mesh-llm-cli -p mesh-llm-commands -p mesh-llm --all-targets -- -D warnings` — clean; `cargo fmt --all -- --check` — clean; `scripts/check-env-mutation-contract.py` — passes (no new env mutation sites).
- Live, against the built binary with an isolated HOME:
  - `serve --join-file <missing>` and `MESH_LLM_JOIN_FILE=<missing> serve` both fail startup with `cannot read join token file ...: write the invite token to that file, point MESH_LLM_JOIN_FILE / --join-file somewhere else, or unset it` (path redacted in the log line).
  - A file-backed token reaches the join path: `Failed to join any peer — running standalone (join rejected: bootstrap_token_invalid)` for a deliberately invalid token.
  - Deleting the token file mid-run produces, at the next 60s tick, `Rejoin failed: join rejected: bootstrap_token_invalid — invite join keeps retrying` as a console warning event — the same failure was invisible before this change.

Not covered here: an end-to-end join against a real `[mesh_requirements]` mesh (needs two hosts and a live invite). The token-source plumbing is what this change touches; the admission path itself is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added `--join-file <PATH>` and `MESH_LLM_JOIN_FILE` for private-mesh invite tokens without exposing them in command-line arguments.
  * Invite-token files are re-read during rejoin attempts, enabling token rotation without restarting services.
  * Added automatic support for `invite.token` beside the resolved configuration file.
  * Added private-mesh service setup guidance and configuration documentation.

* **Bug Fixes**
  * Improved rejoin failure reporting with clearer, deduplicated notifications.
  * Added validation and clearer errors for missing, empty, or unreadable token sources.


i386 (commented):
Reproduced the full matrix at `fa1573f0b` in a clean worktree — every claim checks out:

- `cargo test -p mesh-llm-host-runtime --lib`: **3659 passed / 0 failed / 11 ignored**
- `-p mesh-llm-cli` **91/0**, `-p mesh-llm-commands` **266/0**, `-p mesh-llm` green
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` exit 0 (only pre-existing build-script/linker stderr notes in the log)
- `scripts/check-env-mutation-contract.py` pass
- Base `3f4eb9684` contains no `MESH_LLM_JOIN*` read anywhere — the argv-only finding is confirmed
- Source-level checks: startup hard-errors on an unreadable/empty file; file tokens are re-read each tick; the warn latch is once-per-transition with recovery logging; dedup/precedence match the doc comment; token values are never logged (errors carry paths only); both help surfaces, docs parity (MESHES.md / install.md / config-reference), and the lockstep service-template test are all updated

One blocking behavioral gap in the rotation story, plus smaller notes.

## [blocking] Rotated-out file tokens are retried forever

`apply_join_token_sources` (join_sources.rs:141) folds file-derived tokens into `options.join`, and the rejoin task captures that folded list as its literal set (auto_join.rs:1130). After a rotation, every 60s tick resolves `[stale-startup-token, fresh-file-token]` and joins **both**. `Node::join` → `prepare_join_target` (node_identity.rs:498) has no already-joined short-circuit, so the stale token drives a full wire join attempt and a peer-side rejection **every tick until the service restarts**. Consequences:

- The peer sees a rejected join from this node every minute forever — the same pathology #1957 reports, now visible in our log but not gone from the wire.
- The warn latch fires once at rotation ("Rejoin failed: … — invite join keeps retrying") and never recovers, even though the node *is* joined via the fresh token.
- The docs promise "rotating an invite token is just replacing the file's contents … no unit edit and no restart" — with this bug, a restart is exactly what it takes to stop the dead retries.

Suggested fix: keep provenance instead of folding — leave `options.join` as argv + `MESH_LLM_JOIN` literals, and make the startup consumers (`join_mcp_with_tokens(&options.join, …)` at auto_join.rs:691, and the `join_attempts` build at auto_join.rs:1060) resolve through `resolve_invite_tokens` too. Then the rejoin literals are argv/env-only and each tick's set is exactly literals + current file contents. A regression test that rotates the file with a **non-empty** literal set would pin it — the current `rejoin_ticks_re_read_file_backed_tokens` passes `literals=[]`, which is precisely the shape that hides this interaction.

## [should-fix] Empty `MESH_LLM_JOIN` is silently ignored

`resolve_invite_tokens` (join_sources.rs:121) reads `std::env::var(MESH_LLM_JOIN).ok()`, and `push_unique` drops an empty value. A typo'd `MESH_LLM_JOIN=` line in service.env therefore silently serves standalone — the exact silent-failure class this PR closes, one source over. File sources hard-error on empty at startup; the inline env should match (bail when set-but-blank).

## [minor] Only the first broken file is reported per tick

`resolve_rejoin_tokens` (auto_join.rs:1098) surfaces `errors.into_iter().next()`. With two broken files, the second is invisible until the first is fixed.

## [minor] Warn text can mislead when the node is still joined

If one file breaks but another token joins fine, "invite join keeps retrying" describes the file source, not the node's membership. Scoping the message to the source would help an operator triage.

## [hardening note] `service.env` is written with default permissions

`fs::write` in setup/service.rs — now that `MESH_LLM_JOIN` (inline token) is documented for that file, consider 0600-on-write, or keep steering to `MESH_LLM_JOIN_FILE` + a 0600 token file (install.md already does this well).

Not objecting to: the deliberate omission of a `[join]` config.toml section (schema/docs-parity/`CONFIG_OVERRIDE_ENV_NAMES` churn vs. env+file covering the service case) reads right, and the argv-out-of-`ps` rationale is real.

ndizazzo (commented):
The explicit file and environment sources cover the service case. I would still like us to consider one follow-up for the default path: when neither `--join-file` nor `MESH_LLM_JOIN_FILE` is set, look for a documented join-token file under the resolved `~/.mesh-llm` config directory, while preserving the explicit path override. Deriving this from the resolved config path would also keep `MESH_LLM_CONFIG=/path/project/config.toml` useful for project-local rejoin state. Please make the discovery deterministic, such as one named token file or a config key, rather than scanning every file in the directory, since owner keys, membership, and genesis files live there too. This is nonblocking design feedback for a follow-up; the current explicit sources are usable.

ndizazzo (dismissed):
See above - would be nice to use our config dir

i386:
@ndizazzo — the follow-up you asked for is in, along with the rotation finding from the other review. Head is `b99379ebe`.

**Default token file, as you described it.** When neither `--join-file` nor `MESH_LLM_JOIN_FILE` names a file, an `invite.token` beside the **resolved** config file is used: the default resolves through the same accessor the runtime loads its config through, so `MESH_LLM_CONFIG=/path/project/config.toml` finds `/path/project/invite.token` and rejoin state travels with the config. One deterministic filename, no directory scan — owner keys, membership, and genesis files share that directory, so nothing is picked by shape. An explicit `--join-file` / `MESH_LLM_JOIN_FILE` still wins outright, and the default is opt-in by existence: absent means no source, while a file that exists but is empty or unreadable is a startup error, since its presence means an operator meant to provide a token. `install.md`, `docs/MESHES.md`, the config reference, and the generated service-env template are all updated — writing the default path now needs no env line at all.

**Rotated-out file tokens are no longer retried forever** (the blocking finding in the other review). It was real: `apply_join_token_sources` folded the file token into `options.join`, the rejoin task captured that as its literal set, and after a rotation every 60s tick joined *both* the stale and the fresh token — `prepare_join_target` has no already-joined short-circuit, so the dead token drove a rejected wire join every minute on a node that was already joined, while the docs promised rotation needs no restart. Fixed by keeping provenance instead of folding: `RuntimeOptions::join` is argv/`MESH_LLM_JOIN` literals only, `effective_join_tokens()` is what consumers ask for the whole current set, startup validates rather than folds, and a successful *file-derived* token is no longer frozen as a literal one step later. Each tick now resolves exactly "literals + the file's current contents". The regression test uses the non-empty literal set the old test lacked (its `literals=[]` shape is what hid the interaction); emulating the old fold turns it red.

Also from that review: a `MESH_LLM_JOIN` that is set but blank is now an error rather than a silent skip; every unusable source is reported on a tick, not just the first; and the rejoin warning is scoped to the source so it cannot read as "this node is unjoined" when another source joined fine.

Not in this push: the `service.env` 0600 hardening note. `install.md` already steers operators to `MESH_LLM_JOIN_FILE` plus a 0600 token file, and the default path now makes the inline `MESH_LLM_JOIN` unnecessary, so I left `fs::write` alone rather than change the mode of a file the operator's own service reads. Say the word and I'll tighten it.

Validation at `b99379ebe`: `cargo test -p mesh-llm-host-runtime --lib` 3671 passed / 0 failed / 11 ignored; `-p mesh-llm-cli` 91/0; `-p mesh-llm-commands` 266/0; `cargo fmt --all -- --check`, `cargo clippy -p mesh-llm-host-runtime -p mesh-llm-cli -p mesh-llm-commands --all-targets -- -D warnings`, `just no-console-print` and `scripts/check-env-mutation-contract.py` all clean.

ndizazzo (approved):

PR #1971, opened 2026-09-21
fix(skippy): batch activation exports by request offset, never by lane order

danielwinterw:
> **Status (2026-09-22):** the exporting-stage batching fix is in, and the contract the old gate was standing in for is now enforced by construction. The throughput regression from #1927 is still **not** attributed to this change: the identical lane-order gate exists both before and after #1927, and the lab number is unchanged, so this PR is not the #1927 fix. Left out of "ready" until the regression is understood and the review asks at the end land.

## Problem

Split serving lost 42% of throughput after #1927 landed on `main`: 29.16 -> 16.67 tok/s on the two-mini lab (Qwen3-8B Q4_K_M, M1 stage 0 / M4 stage 1, 12/24 cut, 4 lanes, last-stage batching on, 16 in flight, 1 h sustained). Bisect: `fe9a867e7` + `main` up to and including #1927 measures 15.99; everything before it 27-29.

An instrumented run showed that on the exporting stage the batched iteration path fell back to serial execution on **every** step (`serial_fallback count=8192` in 18 minutes, `batch=2`, reason `activation export batch would span more than one native microbatch`). The failing precondition was `consecutive_sequence_ids`: the batched export path demanded lane ids arrive as `seq 0,1,2...` in request order; the scheduler hands lanes over FIFO, so a 2-lane group is rarely consecutive.

Removing that rule restores batching on the exporting stage (counter 0) but does not recover the throughput — see the status note above.

## Why the rule was wrong

Exports are sliced by request offset. That is the microbatch row order whenever llama decodes the batch with `split_simple` (`llama-kv-cache.cpp`), which is what a unified KV cache does. Consecutive ids only matter on the per-stream path (`split_equal`, sequential), which staged execution only reaches when someone turns the unified cache off.

## Change

Redesigned around that fact, with no compatibility path for the old serial behaviour. A batched iteration that exports activations must be one native microbatch, and every way of breaking that is now either impossible by construction or a loud error:

- **native** (`0015-skippy-batch-activation-exports-by-request-offset.patch`): drop the consecutive-id and uniform-count checks; add `llama_context::kv_unified()`. The batched path admits exports only when they can be sliced safely:
  - an exporting stage with more than one lane needs a unified KV cache, rejected once at load (`model_loading.cpp`) so the operator gets one clear error instead of a per-iteration failure, and re-checked in the batched path;
  - recurrent and hybrid memory is rejected outright, because `output_all` makes it split by sequence (`split_seq`) and only the last microbatch's exports stay live;
  - an exporting iteration larger than `n_ubatch` stays a native error (the scheduler caps it, below);
  - the export tensor's token axis must cover the whole iteration, so a stale or partial microbatch fails as a `RUNTIME_ERROR` instead of silently slicing another request's rows.
- **runtime** (`skippy-runtime/activation.rs`): the silent `Unsupported -> serial` fallback is gone. Frames carrying optional ports (which the batched frontier does not concatenate), and exporting stages whose memory splits an all-output batch by sequence, are routed to one-at-a-time execution *by decision* before the native call. `StageSession` carries the loaded model's state kind, so dense attention stages keep batching while recurrent, hybrid and indexer memory fail closed.
- **scheduler** (`iteration_scheduler.rs`): the `n_ubatch` cap applies to exporting stages only, so terminal and unsplit stages keep whole-prompt prefill up to `n_batch`.
- **partial failures** (`skippy-runtime` + `runtime_state`): a one-at-a-time batch reports how many requests already advanced their sessions, and the runtime moves those tracked positions with them, so `session_token_counts` cannot stay stale after a mid-batch failure.
- **roster**: `split-certified.json` regenerated for the new patch queue.

## Tests

- `cargo test -p skippy-runtime --lib` - 185 passed
- `cargo test -p skippy-server` - 808 passed, 0 failed (full package suite, not the `iteration_scheduler` subset)
- new unit tests: exporting-stage routing vs dense, terminal-stage batching, optional-part routing, the scheduler cap on exporting vs non-exporting stages, the new `StageConfig::emits_activation_frame` predicate, and partial-batch error reporting
- `cargo clippy -p skippy-runtime -p skippy-server -p skippy-protocol --all-targets -- -D warnings`, `cargo fmt --all --check`, `just no-console-print`, `scripts/generate-split-certified.py --check`
- native: the queue applies from the pin and builds (metal) with the new guards; the Rust suites above linked against those archives

## Still open from review

- Native regression coverage for exporting + consuming stages with FIFO lane orders (`[2,0]`, `[3,1]`), unequal token counts, exactly/over the effective microbatch capacity, and a dense vs hybrid/pure-recurrent fixture, asserting per-request parity **and** the number of native executions. Nothing drives `skippy_iteration_batch_sampled` today and the fixture generator has no recurrent architecture, so this is a new native test surface.
- The two-machine bench that would settle the #1927 attribution.

## Summary by CodeRabbit

- **New Features**
  - Improved activation-based batching for requests with optional activation data.
  - Added support for request-offset-based activation exports across varied token and sequence layouts.
  - Added clearer reporting when only part of a batch completes, helping maintain accurate session progress.

- **Bug Fixes**
  - Direct iterations now respect native microbatch limits only when applicable.
  - Improved validation for incompatible activation shapes, dimensions, configurations, and oversized batches.
  - Improved handling of requests that cannot be safely processed together.


third_party/llama.cpp/patches/0015-skippy-batch-activation-exports-by-request-offset.patch
```diff
@@ -0,0 +1,191 @@
...
+-            (!uniform_token_counts || !consecutive_sequence_ids ||
+-             offsets.back() > static_cast<size_t>(first->ctx->n_ubatch()))) {
+-        skippy_set_error(out_error, SKIPPY_STATUS_UNSUPPORTED,
+-                "activation export batch would span more than one native microbatch");
+-        return SKIPPY_STATUS_UNSUPPORTED;
++    // Batched exports are sliced by request offset, which is the microbatch row
++    // order only when llama decodes the batch with split_simple, i.e. on a unified
++    // KV cache. Multi-lane stages always configure one; anything else is a
++    // misconfiguration, not a reason to run the lanes one at a time.
++    if (request_count > 1 && !first->ctx->kv_unified()) {
```
i386: [P1] A unified cache does not guarantee a single microbatch

`kv_unified()` plus total tokens <= `n_ubatch` is insufficient for exporting recurrent/hybrid stages. At the exact pinned upstream `661643e43079a4ee6faab4c1895291767b67ea8d`, `llama_memory_hybrid::init_batch` and `llama_memory_recurrent::init_batch` use `split_seq` when `embd_all` is true; otherwise they use `split_equal`, not `split_simple`. This queue's model setup sets `params.embeddings = has_stage_plan && !terminal_stage`, and `llama_context::decode` passes that `output_all` flag to `memory->init_batch`. Thus two independent one-token exporting requests (e.g. lanes `[2,0]`) can become two microbatches even with unified KV and ample token capacity. The removed guard previously rejected that order before decoding.

After the new check accepts it, `skippy_copy_batched_exports` still reads only `get_gf_res_prev()` and indexes it using offsets from the *whole* request batch. Those offsets cannot describe the last microbatch: earlier requests can receive the wrong rows, or a tensor read can go out of range. The non-embedding hybrid path also splits unequal-length sequences into multiple equal-length microbatches.

Please derive admission/export slicing from the actual memory/split behavior (or collect and map exports from every microbatch) rather than equating unified KV with `split_simple`. Add an exporting hybrid/pure-recurrent regression with nonconsecutive lanes and per-request output parity before removing this protection.

Pinned source: [hybrid init_batch](https://github.com/ggml-org/llama.cpp/blob/661643e43079a4ee6faab4c1895291767b67ea8d/src/llama-memory-hybrid.cpp#L67), [split_seq](https://github.com/ggml-org/llama.cpp/blob/661643e43079a4ee6faab4c1895291767b67ea8d/src/llama-batch.cpp#L682), [decode dispatch](https://github.com/ggml-org/llama.cpp/blob/661643e43079a4ee6faab4c1895291767b67ea8d/src/llama-context.cpp#L1752).

i386 (changes requested):
Reviewed at `1cbcf1a496fda353b6d447b44d7040a2c56ca072`. The dense unified-KV fix addresses a real admission mistake, but the replacement safety predicate is not valid across the supported memory backends; see the inline finding. Please resolve that and add native regression coverage before marking ready.

The regression tests should exercise exporting + consuming stages with FIFO lane orders `[2,0]` and `[3,1]`, unequal token counts, exactly/over the effective microbatch capacity, and dense versus hybrid/pure-recurrent memory. Assert per-request activation/token parity **and the actual number of native microbatches/executions**. The current four-file diff adds no tests, so the reported existing passes do not establish these newly relaxed conditions. Please also run the full `skippy-server` package suite; the PR currently reports only the `iteration_scheduler` subset.

This is a source/history review, not a reproduced two-machine benchmark or an independent test-suite pass.

i386:
The graph-change failure was a mismatch between the scheduler's batching contract and native activation export admission, hidden by a successful serial fallback.

1. **Where introduced:** `consecutive_sequence_ids` and `uniform_token_counts` were added by [#1887](https://github.com/Mesh-LLM/mesh-llm/pull/1887), commit `281c85fdb`, in patch `0047-skippy-batch-multipart-activation-frontiers-generically.patch`. They already exist in #1927's first parent; #1927 carries them into the new core patch 0002. Therefore #1927 is not the introduction point for this guard. The PR's lab results identify a slowdown across a merge cohort, but an isolated pre/post bisect and fixed-head recovery benchmark are still needed to establish its exact performance onset and full contribution.
2. **How it hurts dense split serving:** the scheduler coalesces FIFO requests, while native export admission required adjacent increasing lane IDs and equal token counts, as well as fitting `n_ubatch`. Dense unified KV uses `split_simple`, preserving submitted row order, so lane IDs are not a necessary ordering constraint there. A valid batch such as `[2,0]` returns `UNSUPPORTED`; Rust converts it into per-request calls. Outputs can stay correct while batching disappears. The error text also conflates ordering/count rejection with actually exceeding one microbatch. This is consistent with the author's reported 8,192 fallbacks and reduced split throughput; I have not independently reproduced those measurements.
3. **Why acceptance missed it:** graph frontier/profile correctness does not prove that serving retains batched execution. The [public batch contract](https://github.com/Mesh-LLM/mesh-llm/blob/1cbcf1a496fda353b6d447b44d7040a2c56ca072/third_party/llama.cpp/patches/0001-skippy-define-graph-derived-stage-ABI.patch#L679) explicitly permits different token counts, but the exporting implementation rejected them and concealed that with a fallback. A parity/success test can therefore pass while the performance property is lost. This needed an explicit scheduler-to-native invariant and an execution-count assertion, not only more output tests.
4. **The design did consider performance, but the gate changed:** [the design at #1927](https://github.com/Mesh-LLM/mesh-llm/blob/7c3aef45a393291035d10781a25c304d969575ed/docs/design/SKIPPY_GRAPH_FILTER_V2.md#L625) calls for fixed-hardware split-throughput certification, informational during a three-baseline bootstrap and required afterwards. [#1931](https://github.com/Mesh-LLM/mesh-llm/pull/1931), commit `6adf44872`, replaces that with “Performance follow-up” and excludes it from Graph Filter V2 completion. I have not verified whether the earlier bootstrap was completed/enforced, so I cannot claim a working performance gate was bypassed. The documented acceptance gap is that correctness completion can be achieved without throughput certification.

The prevention is concrete: make supported batch shapes an explicit runtime contract; test shuffled/reused lanes, unequal lengths, memory-backend splitting and microbatch limits; assert batched execution as well as correctness; retain reason-labelled execution telemetry; and attach source/artifact-bound split benchmarks to throughput-sensitive runtime changes. #1971 is directionally right for dense unified KV, but must not generalize that backend's row-order guarantee to recurrent/hybrid memory (inline review finding).

Review scope: immutable source/history inspection at `1cbcf1a49`; no code changes, model runs, or independent full-suite pass.

i386:
Following up with concrete testing improvements. The acceptance criterion needs to be **correct per-request results AND the intended batched execution**. A serial fallback can satisfy output parity while losing throughput; the earlier review explains the [contract mismatch and design history](https://github.com/Mesh-LLM/mesh-llm/pull/1971#issuecomment-5757588371).

For this PR, I would require:

1. **A regression that fails on the old implementation.** Exercise the real scheduler → Rust runtime → native exporting stage → consuming stage with FIFO lane orders `[2,0]` and `[3,1]`, reused lanes, and unequal request token counts. Compare each request's exported activations and downstream results against an isolated reference using equivalent initial state and declared numerical tolerances. Also assert native batch/decode calls, actual microbatch count, and serial-routing/fallback count. For the supported dense single-microbatch cases, require one batched execution and zero serial fallback. Testing only the scheduler's proposed batch size will miss this bug.

2. **Boundary and memory-backend coverage.** Test total tokens just below, exactly at, and above the effective `n_ubatch`; verify that scheduler splitting preserves every request/token and direct inadmissible native calls fail safely before mutating sequence state. Cover dense, hybrid, and pure-recurrent exporting paths. Unified KV plus a token-count cap is not sufficient evidence of one microbatch for all of these backends (see the [review](https://github.com/Mesh-LLM/mesh-llm/pull/1971#pullrequestreview-5264479682)). If a backend splits internally, require correctly mapped exports from every microbatch or explicit safe rejection/routing. Include distinct inputs so row swaps or stale exports cannot pass accidentally; run the native boundary tests under a memory sanitizer where supported.

3. **Tests for intentional serial execution and errors.** Optional-port frames are deliberately routed one at a time in this change: test that routing, output parity, and mixed eligible/ineligible work. Inject native errors and assert that unexpected rejection remains visible rather than silently changing the execution mode. Give legitimate serial routing and unexpected failures separate reason-labelled counters so the supported dense batching test cannot pass by taking another serial route.

4. **Full affected-package and native validation.** Run the complete `skippy-runtime` and `skippy-server` package suites, plus native integration tests against the patched llama build. The reported `skippy-server --lib iteration_scheduler` subset is useful but insufficient for package-wide regressions. Record the exact tested commit and patch/build identity; ordinary unit passes or a successful Metal build do not demonstrate the native export contract.

For prevention beyond this fix:

5. **Make split-performance evidence part of acceptance for changes to graph execution, batching, or export admission.** Use the same fixed hardware, model/quantization, stage cut, lane count, concurrency, and warmup for the baseline and candidate. Repeat runs to establish variance, set an explicit regression budget before judging the result, and report aggregate throughput, per-request latency, actual batch-size distribution, microbatch counts, and serial-routing reasons. Retain a single-node control. For #1971, attach a pinned-head recovery comparison; keep the isolated introduction bisect separate from proof that the fix restores performance. The earlier three-baseline bootstrap must have an owner and an explicit transition to an enforced gate, rather than remain informational indefinitely.

6. **Connect the design invariant to a required check.** Document which batch shapes each backend supports and map each promise to a named test. Run inexpensive execution-path assertions in normal CI and model/hardware integration checks on a provisioned runner; a missing model, skipped native test, or absent benchmark must be reported as missing evidence, not a pass. Keep correctness certification and performance certification separately visible, with explicit review of any exception.

The key improvement is to make “this workload still executes batched” a tested contract. Throughput measurements then catch broader performance changes, while deterministic execution-count assertions catch this specific failure without depending on timing noise.

These are requested acceptance tests and workflow changes, not claims that I have run them. Scope remains source/history review at `1cbcf1a496fda353b6d447b44d7040a2c56ca072`.

third_party/llama.cpp/patches/0015-skippy-batch-activation-exports-by-request-offset.patch
```diff
@@ -0,0 +1,191 @@
...
+     const auto & boundary = session->stage_model->output_activation_boundary;
+     if (desc == nullptr || desc->part_count != boundary.part_count) {
+-        skippy_set_error(out_error, SKIPPY_STATUS_UNSUPPORTED,
+-                "iteration batching defers optional activation ports to serial execution");
+-        return SKIPPY_STATUS_UNSUPPORTED;
++        skippy_set_error(out_error, SKIPPY_STATUS_INVALID_ARGUMENT,
++                "iteration batching accepts planned frontier parts only");
++        return SKIPPY_STATUS_INVALID_ARGUMENT;
+     }
+     llm_graph_result * result = session->ctx->get_gf_res_prev();
```
michaelneale, line 120: 🤖 `get_gf_res_prev()` is the *last* microbatch's graph. That's fine on a unified attention KV cache (`split_simple` → one ubatch for the whole batch), but exporting stages run with `cparams.embeddings=true` → `output_all=true`, and for hybrid/recurrent memory (`llama_memory_hybrid::init_batch`, `llama_memory_recurrent::init_batch`) that selects `split_seq` — one sequence per ubatch. The exported tensor then has `ne[token_axis] == tokens_of_last_request`, and the offset slicing below reads past it for request ≥1 (`GGML_ASSERT` in `ggml_backend_tensor_get`, or silent garbage if the buffer happens to be large enough).

Suggested fix (defence in depth):
```cpp
if (tensor->ne[token_axis] != static_cast<int64_t>(offsets.back())) {
    skippy_set_error(out_error, SKIPPY_STATUS_RUNTIME_ERROR,
            "exported activation token axis does not cover the batched iteration");
    return SKIPPY_STATUS_RUNTIME_ERROR;
}
```
plus route multi-request exporting batches one-at-a-time for recurrent/hybrid architectures (either in `activation.rs` next to the optional-parts check, using the stage's arch, or by having the native side expose a `batched_exports_supported()` capability the Rust layer consults once at load). The roster certifies qwen35, mamba, rwkv6/7, jamba, lfm2, granitehybrid, nemotron_h_moe, kimi-linear and qwen3next, so this isn't hypothetical.

third_party/llama.cpp/patches/0015-skippy-batch-activation-exports-by-request-offset.patch
michaelneale: 🤖 This reject isn't scoped to `request_exports`, so a terminal stage (which never needed unified KV — sampling goes through llama's output-row translation) now fails every batched iteration when the operator sets `model_fit.kv_unified = false` (documented knob, `docs/skippy/CONFIGURATION.md:84`).

Suggest:
```cpp
if (request_exports && request_count > 1 && !first->ctx->kv_unified()) {
```
and, better, validate this once at load time in `model_loading.cpp` (reject `kv_unified=false` with `lane_count>1` on an exporting stage) so the operator gets one clear error at startup instead of a per-iteration failure.

third_party/llama.cpp/patches/0015-skippy-batch-activation-exports-by-request-offset.patch
```diff
@@ -0,0 +1,191 @@
...
++    // misconfiguration, not a reason to run the lanes one at a time.
++    if (request_count > 1 && !first->ctx->kv_unified()) {
++        skippy_set_error(out_error, SKIPPY_STATUS_INVALID_ARGUMENT,
++                "iteration batching requires a unified KV cache");
++        return SKIPPY_STATUS_INVALID_ARGUMENT;
++    }
++    // Exports are read from the last microbatch's graph, so an exporting stage
++    // must see the whole batch in one microbatch. The scheduler caps batched
++    // tokens at n_ubatch; exceeding it here is a scheduler bug.
++    if (request_exports && offsets.back() > static_cast<size_t>(first->ctx->n_ubatch())) {
```
michaelneale, line 227: 🤖 Agree this should be a hard error here given the scheduler cap — but see my comment on `iteration_scheduler.rs`: the cap as written applies to non-exporting stages too, so either the cap needs scoping to exporting stages, or this check needs to stay `request_exports`-only (it is) *and* the scheduler needs to not throttle terminal/single-stage prefill.

crates/skippy-server/src/frontend/iteration_scheduler.rs
```diff
@@ -435,7 +435,20 @@ impl IterationScheduler {
...
+        let microbatch_tokens = usize::try_from(
+            config
+                .n_ubatch
+                .unwrap_or(skippy_runtime::LLAMA_SERVER_DEFAULT_N_UBATCH),
+        )
+        .unwrap_or(usize::MAX)
+        .max(1);
+        let max_direct_iteration_tokens = scheduler_config
+            .max_tokens_per_iteration
+            .min(microbatch_tokens);
```
michaelneale: 🤖 This cap applies to every stage, not just exporting ones. `validate_direct_iteration` will now reject any direct iteration > n_ubatch (512 default), and `prefill_whole_prompt` (`token_generation.rs:682`) submits whole prompts up to `n_batch` when prefix cache is off and the built-in driver isn't eligible (speculative / draft / MTP / hooks). That's a hard error on a path that worked before, on single-stage and terminal stages that don't export anything.

Suggest scoping:
```rust
let max_direct_iteration_tokens = if config.downstream.is_some() {
    scheduler_config.max_tokens_per_iteration.min(microbatch_tokens)
} else {
    scheduler_config.max_tokens_per_iteration
};
```
(or whatever the canonical "this stage exports activations" predicate is here). A unit test asserting the cap on an exporting config and its absence on a terminal config would lock this in.

crates/skippy-runtime/src/activation.rs
```diff
@@ -103,9 +103,19 @@ pub struct IterationBatchRequest<'a> {
...
+    /// Execute one scheduler iteration for every request as a single native
+    /// batch. The native batched path accepts planned frontier parts only;
+    /// frames that carry optional parts are executed one at a time here, by
+    /// decision, so the native layer never has to signal a fallback.
     pub fn iteration_batch_sampled(
         requests: &mut [IterationBatchRequest<'_>],
     ) -> Result<IterationBatchOutput> {
+        if requests
+            .iter()
+            .any(|request| frame_has_optional_parts(request.input))
```
michaelneale: 🤖 The optional-parts routing (`unwrap_or(true)` → one-at-a-time) is conservative — good. This is also the natural place to add the recurrent/hybrid guard from my patch-0015 comment: if the session's stage arch uses `split_seq` for exporting batches, route to `iteration_batch_sampled_one_at_a_time` here (same for `decode_step_frame_batch_sampled` below) rather than letting native slice a tensor that only covers the last request.

crates/skippy-runtime/src/activation.rs
```diff
@@ -187,10 +197,6 @@ impl StageSession {
             free_error(error);
             return Self::iteration_batch_sampled_raw(requests, &output_bytes);
         }
-        if status == Status::Unsupported {
-            free_error(error);
-            return Self::iteration_batch_sampled_serial(requests);
-        }
         ensure_ok(status, error)?;
```
michaelneale, line 201: 🤖 With the `Unsupported → serial` fallback removed, every former `UNSUPPORTED` is now a request-killing error. That's the intended contract, and I agree with it *provided* the native layer can't reach those branches on any certified config. Today it can (hybrid/recurrent exporting stages, `kv_unified=false`). Please either add the load-time/arch guards or keep a narrow fallback until the roster is clean — and add a test that drives `iteration_batch_sampled` with two requests through a mocked `INVALID_ARGUMENT` so the error path is exercised at all (none of the 51 added Rust lines are under test).

michaelneale (changes requested):
🤖 Code-only review (no build/bench run), posted on behalf of MicBlock.

**Summary:** the core idea is right — the `consecutive_sequence_ids`/`uniform_token_counts` gate was what forced every 2-lane exporting step onto the serial path, and slicing exports by request offset is valid when llama uses `split_simple`. I verified that chain against the pinned llama.cpp (`661643e4`): unified KV → `n_stream == 1` (`src/llama-kv-cache.cpp:83`) → `split_simple` (`:768`), which walks batch indices in order (`src/llama-batch.cpp:476-507`); exporting stages set `kv_unified = lane_count > 1` (`src/skippy/model_loading.cpp:717`). Terminal-stage sampling already goes through llama's output-row translation, so it never depended on ordering. The patch series applies cleanly from the pin.

**Why request-changes:** as written this trades a silent slow path for hard failures / aborts on a few real configurations, and the #1927 attribution isn't proven yet. Inline comments below carry the specific fixes; short list:

1. **Hybrid/recurrent exporting stages take a different split path.** Exporting stages set `cparams.embeddings = true` (`model_loading.cpp:731`) → `output_all = true` (`llama-context.cpp:1776`) → `llama_memory_hybrid::init_batch` / `llama_memory_recurrent::init_batch` use `split_seq` (`llama-memory-hybrid.cpp:79`, `llama-memory-recurrent.cpp:418`): one sequence per ubatch. Exports are read from `get_gf_res_prev()` (last ubatch only) and sliced by request offset with no check on the tensor's token axis. Request ≥1 reads past a 1-token tensor → `GGML_ASSERT` in `ggml_backend_tensor_get` (abort) or corrupted activations. Pre-PR the consecutive gate almost always hid this; post-PR it's the normal path. Affected roster entries: qwen35, qwen35moe, mamba, rwkv6/7, jamba, lfm2, granitehybrid, nemotron_h_moe, kimi-linear, qwen3next.
2. **`kv_unified=false` becomes a hard `INVALID_ARGUMENT` on every batched iteration, including terminal stages** — the check isn't scoped to `request_exports`, and `model_fit.kv_unified` is a documented operator knob (`docs/skippy/CONFIGURATION.md:84`).
3. **n_ubatch cap regresses single-stage whole-prompt prefill.** `validate_direct_iteration` now rejects any direct iteration > n_ubatch; `prefill_whole_prompt` (`token_generation.rs:682`) submits whole prompts up to `n_batch` when prefix cache is off and the built-in driver isn't eligible (speculative/draft/MTP/hooks) → hard error instead of degradation.
4. **Attribution to #1927 is unproven.** The identical gate exists before #1927 (patch 0047 at `7c3aef45a^`) and after (patch 0002). The PR body says the pre-#1927 bisect and the branch benchmark are still pending — that evidence is needed before "restores performance" is established.
5. **No tests** for the new behaviour (51 added Rust lines, none exercised): optional-part routing, the n_ubatch cap, and the native rejects.

Happy to re-review once the bench numbers land.

third_party/llama.cpp/patches/0015-skippy-batch-activation-exports-by-request-offset.patch
i386: Verified at f52c43ae and retained in ce2cbc79b: patch 0015 checks `tensor->ne[token_axis] == offsets.back()` before slicing, returning RuntimeError on incomplete exports. It also rejects multi-request recurrent/hybrid exports. `StageModel::supports_batched_activation_exports` admits dense state without indexer memory only, and both Rust batch entry points route other exporting sessions one at a time. The requested guards are present. Runtime package validation passed: 186 unit tests plus integration tests. Model-backed parity remains explicitly tracked in the separate P1 thread; this resolution covers the guard/routing request only.

third_party/llama.cpp/patches/0015-skippy-batch-activation-exports-by-request-offset.patch
i386: Verified and completed in ce2cbc79b: `iteration_token_budget` applies the microbatch cap only when `StageConfig::emits_activation_frame()` is true. `build_scheduler_config` now uses that same export-only cap before Scheduler::new, covering planned work as well as direct work. The new planned-prefill regression uses four 256-token requests, n_batch=2048 and n_ubatch=512: exporting stages plan 512 total tokens, non-exporting stages retain all 1024. Full skippy-server suite passed: 809 tests, 0 failed, 4 ignored.

crates/skippy-runtime/src/activation.rs
i386: Addressed in ce2cbc79b. Existing load-time unified-KV validation and recurrent/hybrid routing remain in place. Added `native_invalid_argument_is_returned_without_serial_fallback`, which calls the public iteration_batch_sampled entry with two requests and real null native sessions. This exercises an actual native InvalidArgument instead of mocking it, asserts that it propagates as a batch error rather than PartialBatchExecution, and confirms neither Rust session count advances. Full skippy-runtime package tests passed (186 unit tests, 1 ignored, plus integration tests); all-target warning-denying Clippy passed for skippy-runtime, skippy-server and mesh-llm.

crates/skippy-server/src/frontend/iteration_scheduler.rs
```diff
@@ -435,7 +435,11 @@ impl IterationScheduler {
         let max_consecutive_prefill_iterations =
             scheduler_config.max_consecutive_prefill_iterations;
         let mixed_prefill_decode = scheduler_config.mixed_prefill_decode;
-        let max_direct_iteration_tokens = scheduler_config.max_tokens_per_iteration;
+        let max_direct_iteration_tokens = direct_iteration_token_budget(
+            config.emits_activation_frame(),
+            scheduler_config.max_tokens_per_iteration,
+            config.n_ubatch,
+        );
```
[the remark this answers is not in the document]
i386: Fixed in ce2cbc79b. `build_scheduler_config` now accepts the stage export predicate and caps max_tokens_per_iteration before constructing Scheduler, so planned multi-request prefills cannot aggregate beyond n_ubatch. Direct iterations retain the same bound and terminal/unsplit stages retain their full budget. `planned_prefills_share_the_export_microbatch_budget` exercises actual Scheduler::submit/plan_iteration with four requests and n_batch > n_ubatch, including the non-exporting control. Full skippy-server suite: 809 passed, 0 failed, 4 ignored. Cargo check and all-target Clippy with -D warnings passed for the changed crates and mesh-llm.

third_party/llama.cpp/patches/0015-skippy-batch-activation-exports-by-request-offset.patch
i386: The unsafe admission is guarded in the current code: native rejects multi-request recurrent/hybrid exports and verifies the export token axis covers the full batch; Rust routes non-dense/indexer exporting stages one at a time. I am leaving this thread open because the requested model-backed parity evidence is not yet established.

I built a regression using real stage plans/sessions, exporting and consuming stages, native lane order [2,0]/[3,1], unequal token counts, and a 32-token capacity case. With the generated qwen3-dense fixture, it fails opening the first exporting context, before any batch executes:

```text
cannot realize admitted stage graph: semantic graph has an unregistered request input: leaf_57 type=26 ne=8x1x1x1 flags=1 buffered=0
RuntimeError: failed to create llama context
```

Inputs: 2-layer Qwen3 fixture, ranges [0,1)/[1,2), ctx_size=512, lane_count=8, n_batch=128, n_ubatch=32, CPU tensors, unified KV. Native build stamp and prepared HEAD both identify 5f79a960064149c2e19cb94e6b85a2ecb5a2406a; its patch digest 750fb7e9db38bdc994299876afe2650ec0ed5f76adc5abec1d106d4f552501b7 matches this PR queue exactly. This is a prerequisite/load failure, not proof of batching parity or an attributed batching defect.

Preserved the reproducer for follow-up at the shared workspace path OUTBOX/PR1971_BATCHED_EXPORTS_NATIVE.rs and recorded details in RESEARCH/PR1971_REVIEW_THREAD_TRIAGE_2026_09_22.md. Did not commit a known-failing opt-in test or mark this request satisfied. The separately verified scheduler-cap and native error-propagation fixes are in ce2cbc79b.

PR #1975, opened 2026-09-21
fix(ui): keep nested runtime settings in their own TOML tables

### LINKED ISSUE #1784: Console TOML Output flattens nested runtime sub-tables and still validates
## What happens

The Configuration screen's TOML Output tab serializes nested `[runtime.*]` sub-tables as flat keys under `[runtime]`. On a node whose config is only:

```toml
[runtime.native_runtime]
selection = "metal"
```

the generated preview reads:

```toml
[runtime]
...
advertisement = "coarse_state"
selection = "metal"
drain_timeout_max_secs = 300
```

`selection` belongs to `[runtime.native_runtime]`, and `advertisement`, `enabled`, `idle_after_secs`, `mode`, `resume_debounce_secs`, `poll_interval_secs`, and `response` all belong to `[runtime.activity]`. The Validation panel next to the pane still reports "Generated TOML validates against mesh-llm config rules", because `runtime.selection` is simply an unknown key rather than an invalid one.

TOML Output tab flattening nested runtime sub-tables

## Why it matters

The pane has a Copy button. Copy that text into `config.toml` and the native runtime backend selection is silently gone, along with every activity setting, and the console still says the file is valid. Saving from the console goes through `mergeConfigurationIntoMeshConfig` and the structured apply-config payload rather than this text, so the damage is limited to what an operator copies out.

## Cause

Built-in settings under `runtime.native_runtime.*` and `runtime.activity.*` carry no `toml_section` in the schema, so `build-toml.ts` places them in the parent `[runtime]` table keyed by their last path segment.

## Suggested fix

1. Derive the section from the setting's canonical path when no explicit `toml_section` is present, so `runtime.native_runtime.selection` lands in `[runtime.native_runtime]`.
2. Have the validate path reject, or at minimum warn on, unknown keys so a flattened paste does not come back green.
3. Add a build-toml round-trip test: render the preview for a config with `runtime.native_runtime.selection` and `runtime.activity.*` set, parse it back, and assert equality with the source config.

## How to reproduce

1. Start a node with a config containing `[runtime.native_runtime] selection = "metal"`.
2. Open the console, Configuration, TOML Output.
3. Compare the generated preview with the source file.

Found while reviewing #1759, which does not touch any UI files and is not the cause.

i386:
## What changes

The Configuration → TOML Output preview now writes nested runtime settings into their own tables instead of flattening them into `[runtime]`.

Before, a config with `[runtime.native_runtime] selection = "metal"` previewed as:

```toml
[runtime]
enabled = true
idle_after_secs = 600
advertisement = "coarse_state"
selection = "metal"
```

Copying that preview into `config.toml` silently dropped the native-runtime selection and every `[runtime.activity]` / `[runtime.kv_cache.disk]` setting, while the Validation panel still reported "Generated TOML validates against mesh-llm config rules."

Now the same config previews as:

```toml
[runtime]
mode = "client"

[runtime.activity]
enabled = true
idle_after_secs = 600
advertisement = "coarse_state"

[runtime.native_runtime]
selection = "metal"

[runtime.kv_cache.disk]
mode = "fixed"
```

## Why

Schema-derived `runtime.*` settings were built with a hard-coded `tomlSection: 'runtime'`, so the console placed `runtime.activity.*`, `runtime.native_runtime.*`, and `runtime.kv_cache.disk.*` in the parent `[runtime]` table keyed by their last path segment. `RuntimeConfig` has no `deny_unknown_fields`, so the flattened document still parsed — hence the green validation. The general schema adapter already derived the section from the canonical path; only the runtime-policy builder overrode it.

The section is now derived from the setting's canonical path, matching the general adapter, and a `buildTOML` round-trip test parses the preview back to catch any future flattening.

Fixes #1784.

## Scope

Deriving the section from the canonical path is the fix for the reported defect; it also corrects `runtime.kv_cache.disk.*`, which had the same flattening bug but was not called out in the report.

#1784 also suggested rejecting or warning on unknown keys in the validation path. That is deliberately not included here: the validate endpoint deserializes into typed config structs (`handle_runtime_config_validate` → `MeshConfig`), so it cannot observe unknown keys, and adding a raw-key walk in `mesh-llm-config` is a broader change that risks false positives on valid configs (`models`, `plugin`, legacy aliases). With the section fix, the preview no longer emits unknown keys. Happy to open that as a separate issue.

## Validation

- `pnpm run typecheck`
- `pnpm run build`
- `pnpm exec vitest run` — 209 files, 1766 passed, 3 skipped
- `pnpm exec eslint <changed files> --max-warnings=0`
- New tests: canonical-path section derivation for the nested runtime tables, plus a `buildTOML` round-trip that parses the preview back and asserts the nested config is preserved and no keys remain under `[runtime]`. Adds `smol-toml` as a devDependency for the parse-back assertion.

No Rust surface changed, so no Cargo build was run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected configuration previews and exports so nested runtime settings are written to their proper TOML subsections, including activity, native runtime, and disk cache settings.
  * Preserved nested configuration structures when converting settings to TOML, preventing values from being misplaced or ignored.

* **Tests**
  * Added coverage validating nested runtime TOML sections and round-trip parsing.


i386:
**Independent reviewer verification — approve verdict.** (Posted by the reviewer agent; this machine's shared `gh` login belongs to the PR author, which is why this is a comment rather than a formal approving review — GitHub rejects self-approvals, and it wouldn't be honest attribution anyway.)

Reviewed at commit `376ffd5e5` in an isolated worktree (base `d71ceeaf`), gates re-run independently:

| Gate | Result |
|---|---|
| `pnpm run typecheck` | exit 0 |
| `pnpm run test` (full vitest) | 209 files, 1766 passed / 3 skipped — exact match to the claim |
| eslint (4 changed TS files) | exit 0 |
| `pnpm run build` (tsc -b && vite build) | exit 0 (chunk-size warning is a pre-existing advisory) |

**Source-level checks:**

- The behavior change is exactly one line (`tomlSection: 'runtime'` → `configSectionForPath(entry.canonical_path)`) at the only producer of schema-derived runtime settings. No hardcoded `runtime.*` settings exist elsewhere (checked `app-tabs/configuration-defaults*`), so the fix covers the whole surface.
- The helper move into `config-adapter-paths.ts` is verbatim, modulo `export` and explicit return types.
- The two remaining category-level `tomlSection: 'runtime'` literals (`CATEGORY_FALLBACKS`, the schema adapter's `runtime-policy` category) are inert for these settings: `resolveDefaultSettingSectionPath` prefers the setting-level section, and every `runtime.*` schema path has ≥2 segments so `configSectionForPath` always returns a defined value. It would also be the correct fallback for a hypothetical single-segment path.
- Emission order is TOML-valid: unknown sections (`runtime.activity` etc.) are appended after the known ones in `buildTOML`, and the round-trip through a real parser (`smol-toml`) proves parseability.
- The fix direction matches the server contract: `RuntimeConfig` has nested `activity: RuntimeActivityConfig` and `native_runtime: NativeRuntimeConfig` (`crates/mesh-llm-config/src/model.rs:114,120`), so `[runtime.activity]` / `[runtime.native_runtime]` are the shapes the binary actually consumes. No `deny_unknown_fields` on `RuntimeConfig`, confirming the silent-loss mechanism behind #1784.
- **Regression proof:** swapping the base `runtime-settings.ts` into this same tree makes 4 tests fail (both round-trip tests plus the section-derivation test); restoring the fix gives 10/10. The new tests are not vacuous.

**Non-blocking notes:**

1. `runtime.kv_cache.disk.mode` in the fixture is synthetic — no such path exists in the current built-in schema (`declarations.rs` has no `kv_cache.disk`), so the description's "covers `runtime.kv_cache.disk.*`, which had the same flattening bug" slightly overstates. Still valuable as depth-3 generic coverage; no change requested.
2. Deferring the unknown-key validation hardening is the right call: the validate endpoint deserializes into typed structs and cannot see unknown keys, and a raw-key walk risks false positives on `models` / `plugin` / legacy aliases. If wanted, track it as a separate issue.

Mergeable from my side pending CI.

i386:
@Virgile-pct can I get a review please

Virgile-pct (approved):
Approving. The fix is right, and I checked the two things a UI-only read does not cover: whether the new tables are real on the Rust side, and whether moving `tomlSection` reaches anything other than the preview. One correction to the description at the end.

### The Rust boundary agrees

`RuntimeConfig` derives `Deserialize` with no `deny_unknown_fields`, while fifteen other structs in the same file do use it, so the flattened keys really were accepted and dropped rather than rejected. That is the part that makes the old preview dangerous rather than merely untidy: the Validation panel had nothing to complain about.

The built-in schema declares twenty `runtime.*` settings. Eleven are flat and stay in `[runtime]`. Nine move, into exactly two tables:

| table | settings | Rust type |
| --- | ---: | --- |
| `runtime.activity` | 6 | `RuntimeActivityConfig` |
| `runtime.native_runtime` | 3 | `NativeRuntimeConfig` |

Both exist as nested structs on `RuntimeConfig`, and the nine keys line up one for one with their fields: `enabled`, `idle_after_secs`, `poll_interval_secs`, `resume_debounce_secs`, `response`, `advertisement`, then `mesh_version`, `skippy_abi`, `selection`. So the preview now emits tables mesh-llm actually reads, for every nested runtime setting the schema has.

### The blast radius is the preview, as you say

`tomlSection` is read in four places, and I walked each one because deriving it from the canonical path could have moved the value path too:

- `resolveConfigSettingPath` returns the canonical path directly when there is one and only falls back to `[...sectionPathSegments(tomlSection), key]` when there is not. Schema-derived runtime settings always carry `canonicalPath`, so reads and writes are untouched.
- `pluginPathParts` and `parsePluginCanonicalPath` both need `sectionSegments[0] === 'plugin'`. It was `'runtime'` before and is `undefined` now, so both took and still take the regex branch on the canonical path. No change.
- `defaultFlatAliasForSetting` keys off `canonicalPath` whenever the setting has one, and no alias in the table is `runtime.*`, so the longer section string cannot miss an alias it used to hit.

That leaves `resolveDefaultSettingSectionPath`, which is the rendering path this PR is about. Good.

Moving the three helpers into `config-adapter-paths.ts` rather than copying them is the right call, since `config-adapter-schema.ts` and `runtime-settings.ts` now cannot drift apart on section derivation, which is how the bug got in.

### One correction, non-blocking

The Scope section says the change "also corrects `runtime.kv_cache.disk.*`, which had the same flattening bug but was not called out in the report". I cannot find that path. It is not in the built-in schema declarations, `RuntimeConfig` has no `kv_cache` field, and the only kv-cache config key I find anywhere is `defaults.model_fit.kv_cache_policy`, a flat `Option<String>`. The `kv_cache` names in the host runtime are Skippy stage structures, not config paths.

The test fixture invents `runtime.kv_cache.disk.mode`, which is fine and even useful as a generic check that three-level paths derive correctly. It is the description I would change, since it reads as a second defect fixed and a release note could repeat it. If the path is real and I searched the wrong place, point me at it and I will withdraw this.

### What I did not verify

I did not run the vitest suite; there is no pnpm on this machine, so I am taking your 1766-passed run as given rather than reproducing it. I also did not exercise the console itself, so "copying the preview now works" is inferred from the parse-back assertion in your new test plus the Rust structs above, not observed in the app.

PR #1976, opened 2026-09-21
fix(api): box GET route futures so the router frame stops summing its arms

Virgile-pct:
Fixes the mechanism behind #1915, with the measurement on that issue.

`handle_get` is a `match` with fourteen arms, each awaiting an `async fn`. In a debug
build that reserves one stack slot per arm, sized by that arm's future, and the slots
are not overlapped, so the router's frame is the sum of its routes rather than the
largest one. Thirteen of those handlers now return an already-boxed future, the shape
`DispatchRequestFn` already uses one module up, so each arm holds a pointer. The bodies
are unchanged; each `async fn handle_x` became a thin `fn handle_x` that boxes an
`async fn handle_x_inner` with the same body.

Three quarters of the saving is upstream of the change: a smaller `handle_get` future
shrinks `runtime::handle`'s frame, which shrinks the three dispatcher arms that await it.

| frame | before | after |
| --- | ---: | ---: |
| `DISPATCH_REQUEST` closure | 54.4 KiB | 26.0 KiB |
| `DISPATCH_REQUEST` body | 922.2 KiB | 683.0 KiB |
| `runtime::handle` | 138.6 KiB | 31.3 KiB |
| `handle_get` | 332.2 KiB | under 4 KiB |
| whole `/api/status` poll chain | **1 904.3 KiB** | **1 197.2 KiB** |

The stack the Windows test actually needs moved with it. `runtime_data_api_routes_remain_payload_stable`
overflowed at 2 048 KiB and passed at 2 304 KiB before; it now overflows at 1 280 KiB and
passes at 1 408 KiB, and passes with no `RUST_MIN_STACK` set at all.

Frame sizes are read out of the test binary: on x64 Windows a frame larger than a page
opens with `mov eax, <size>` / `call __chkstk` / `sub rsp, rax`, so each one is a literal
in `.text`, and `dbghelp` resolves the symbol from the PDB next to the exe. Measured on
`main` at 0feea9b61 with rustc 1.98.0 on x86_64-pc-windows-msvc.

`mesh-llm-host-runtime --lib` at the default stack, no `RUST_MIN_STACK` anywhere, 16
threads, two runs of each:

| build | tests | result |
| --- | --- | --- |
| `main`, overflowing test skipped | 3736 | 3736 passed, 0 failed, 11 ignored (twice) |
| this branch, nothing skipped | 3737 | 3735 passed, 2 failed, then 3737 passed, 0 failed |

The two failures in the first run did not recur in the second, both pass on their own,
and both live in `network::openai`, which this diff does not touch; one of them is the
`MESH_LLM_LIFECYCLE_LOG_PARSER` env race reported in the Windows survey on #1554. I am
calling them parallel-execution flakes rather than pre-existing failures, because a
single clean baseline run is not evidence that they predate this change.

No stack overflow anywhere in any of the four runs.

`cargo clippy -p mesh-llm-host-runtime --all-targets --no-deps -- -D warnings` returns the
same four errors on this branch as on `main` at 0feea9b61, the same lines in the same
order: an unused `PathBuf` import and two never-used test helpers under `#[cfg(test)]`,
and one needless `return` under `#[cfg(windows)]`, all of them in `plugin/`. They are
Windows-only, which is why the quality lane does not see them, and none of them is in
the file this diff touches. I will send those separately.

I left the `RUST_MIN_STACK` bridge on the Windows unit step alone. It covers the whole
suite rather than this one test, and taking it out should be its own change that can be
reverted on its own; I will send it once this has been green on `main` for a few runs and
watch that lane as I did for the row itself.

Not verified: release builds, where these frames are far smaller and none of this is
expected to matter, and any effect on request latency, which I did not measure. Boxing
adds one allocation per routed GET request.

## Summary by CodeRabbit

* **Refactor**
  * Refactored runtime route handling internally without changing observable behavior.
  * Existing requests, responses, validation, and error handling remain unchanged.


i386 (approved):
Verified independently on macOS — the central claim ("the bodies are unchanged") checks out mechanically, and the suite is green here.

**Body identity.** I extracted each old `async fn handle_x` body from `0feea9b61` and each new `handle_x_inner` body from this branch and compared them whitespace-normalized: 13/13 byte-identical, including `handle_get` itself (line-for-line unchanged, only the awaited futures are now boxed). Every wrapper has the `fn handle_x<'a>(...) -> BoxedRouteFuture<'a>` + `Box::pin(handle_x_inner(...))` shape. `BoxedRouteFuture` is indeed the same shape as `DispatchRequestFn` in `routes/mod.rs` (`Pin<Box<dyn Future + Send + 'a>>`; `Output = ()` vs `bool` as expected).

**Tests and gates.** `cargo test -p mesh-llm-host-runtime --lib`: 3773 passed, 0 failed, 11 ignored (69s). `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings`: exit 0. The mechanism story is consistent with the measurements on #1915 — the 1197 KiB post-fix poll chain fits the 2048 KiB stack the pre-fix 1904 KiB chain overflowed, which is exactly why the test passes with no `RUST_MIN_STACK` after this change.

Two observations, neither blocking:

1. `/api/runtime/activity` → `super::runtime_activity::handle_get` is the one route arm left unboxed (it lives in another module, and you said "thirteen"). Its future still contributes its full frame to `handle_get`. If the chain ever needs shaving again, that's the next candidate — worth boxing it in the same style when that module is touched next.
2. `handle_post` (12 arms) and `handle_delete` have the same sum-of-arms shape. GET was the measured offender and I agree with fixing what's measured, but if the Windows lane surfaces a POST-path overflow later, this same transformation is the template.

Leaving the `RUST_MIN_STACK` bridge on the Windows unit step as its own follow-up is the right call — it should be revertable independently once this has been green on `main` for a while.

Approving.

Virgile-pct:
Both of your non-blocking observations, measured on `main` at 39c8a76b8 now that this has landed. Same method as #1915, frame sizes read out of the test binary.

**The unboxed arm.** `runtime_activity::handle_get` does not appear in the scan at all, which means its frame is under the 4 KiB floor I scan at, and `handle_get_inner` itself is under 4 KiB too. So the one arm left inline contributes almost nothing, and boxing it would buy single-digit KiB. Worth doing if that module is touched for another reason, not worth a PR of its own.

**POST and DELETE.** They do have the same sum-of-arms shape, `handle_post` is still the second largest routing frame at 190.7 KiB, but neither is close to the limit. Full chains, socle plus method match plus router plus the heaviest leaf of that family:

| chain | total | of 2048 KiB | margin |
| --- | ---: | ---: | ---: |
| `GET /api/status` | 1197.2 KiB | 58.5% | 850.8 |
| `GET /api/runtime/stages` | 1120.1 KiB | 54.7% | 927.9 |
| heaviest POST | 1032.2 KiB | 50.4% | 1015.8 |
| heaviest DELETE | 819.5 KiB | 40.0% | 1228.5 |

So there is no latent POST-path overflow today: over a megabyte of headroom, and the worst chain in the whole management surface is still `/api/status` at 58% of a default test thread.

What is left is the dispatcher body at 683.0 KiB, which every route pays and which is now 57% of the worst chain. Its 22 arms would take the same transformation, but that spans 22 modules and nothing is close enough to the limit to justify it. I would leave it until something needs the room.

For completeness, the largest remaining frames in `api/routes/` after the dispatcher are four `plugins.rs` test bodies, 234 to 396 KiB, not production paths.

PR #1978, opened 2026-09-21
fix(plugin): build the tests on Windows, and close the rooted-path hole they expose

Virgile-pct:
`cargo test -p mesh-llm-plugin` has never compiled on Windows, and once it does, one
test fails for a real reason. Three commits, in that order.

### The test target does not build

Three tests in `runtime.rs` set up a Unix socket pair and guard it like this:

```rust
#[cfg(unix)]
let (plugin_stream, host_stream) = tokio::net::UnixStream::pair().unwrap();
#[cfg(not(unix))]
panic!("runtime stream tests are only implemented for unix");
```

The intent is clear, but `panic!` is a run-time expression and does not stop the
compiler from type-checking the lines after it. On Windows neither binding exists and
`LocalStream` has no `Unix` variant, so the target fails to build:

```
error[E0425]: cannot find value `plugin_stream` in this scope
error[E0599]: no variant, associated function, or constant named `Unix` found for enum `io::LocalStream`
error: could not compile `mesh-llm-plugin` (lib test) due to 12 previous errors
```

The Linux quality lane cannot see this, because on Linux the Unix branch compiles. The
first commit moves the intent onto the functions with `#[cfg(unix)]` and gates the
imports and the one helper that only they use. Nothing changes on Unix.

### With it building, a guard turns out to have a hole

`web_ui_packaging_rejects_absolute_paths` feeds `/var/lib/plugin-ui` and expects the
packaging guard to refuse it. On Windows it does not. `validate_relative_path` tests
`Path::is_absolute`, which is false for that value there because it carries no drive
prefix. The path is still rooted, and `Path::join` drops the package root when joining
it, which is the escape the guard exists to prevent. So a manifest declaring a rooted
bundle root was accepted on Windows and refused everywhere else.

The second commit rejects a first component of `Prefix` or `RootDir` instead. That
covers `C:\x`, `\\server\share`, `/x` and `\x` on Windows and `/x` on Unix, and makes
the verdict identical on every platform. One asymmetry is left alone deliberately:
`C:\x` is a plain file name on Unix and still passes there. Rejecting it everywhere
would mean matching drive letters textually, which is a call for you rather than me.

### And then clippy is clean

The third commit removes the explicit `return` at the end of the Windows branch of
`local_listener`, which clippy rejects under `-D warnings` because the cfg-stripped
body makes that block the tail expression. `AGENTS.md` asks contributors to run
`cargo clippy -p <crate> --all-targets -- -D warnings` as a local gate, and this was
the last thing making it fail for this crate on Windows.

### Measured

On `main` at d71ceeaf1, Windows 11, rustc 1.98.0, x86_64-pc-windows-msvc:

| command | before | after |
| --- | --- | --- |
| `cargo test -p mesh-llm-plugin --no-run` | 12 errors, no binary | builds, 0 warnings |
| `cargo test -p mesh-llm-plugin --lib` | could not run | 33 passed, 0 failed |
| `cargo clippy -p mesh-llm-plugin --all-targets --no-deps -- -D warnings` | 1 error | clean |

### What I did not verify

I could not build for a Unix target from this machine, so the claim that nothing
changes on Unix rests on reading the diff: the three tests and their imports are gated
to the platform they already only ran on, and the new path check accepts everything
`is_absolute` accepted there. Linux CI on this PR is the real check. I also did not
audit the other callers of `validate_relative_path` for manifests that might rely on a
rooted path being accepted on Windows; I could not construct one that was not an
escape.

Found while surveying what the Linux-only quality lane cannot see on Windows, the same
exercise that produced the `windows-unit` row in #1554.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved relative-path validation to reject rooted paths and platform-specific prefixes, including Windows drive and UNC paths.
  - This prevents invalid absolute paths from being accepted regardless of the platform in use.

- **Tests**
  - Updated platform-specific runtime coverage to run only where supported, improving cross-platform test reliability.


i386 (approved):
Verified independently on macOS (cross-compiling to `x86_64-pc-windows-msvc` with rustc 1.98.0, the same version you measured on) — every claim in the description reproduces:

**Commit 1 (test target builds on Windows).** At the merge base `2d6f7930e`, `cargo check --tests -p mesh-llm-plugin --target x86_64-pc-windows-msvc` fails with exactly the 12 errors you list: 6× E0425 (`plugin_stream` ×3, `host_stream` ×3) and 6× E0599 (`LocalStream::Unix`). At the PR head the same command compiles clean. On the Unix side, `cargo test -p mesh-llm-plugin --all-targets` gives 36 passed / 0 failed at both base and head — identical, so "nothing changes on Unix" holds empirically, and 36 = your 33 Windows + the 3 gated tests. The gated imports (`read_envelope`/`write_envelope`, `Barrier`/`Notify`, `Duration`/`timeout`) and `test_channel_message` are used only inside the three gated tests, so the gating is complete.

**Commit 2 (rooted-path rejection).** The first-component `Prefix(_) | RootDir` check subsumes `is_absolute` on Unix and closes `/x` and `\x` on Windows; the remaining checks (`CurDir`-only, `ParentDir`, hidden segments) are untouched and `Component` was already imported for them. Error message unchanged. I audited the other `validate_relative_path` you left unaudited: the one in `crates/skippy-package-format/src/lib.rs:954` already rejects any non-`Normal` component, so it never had this hole — no other callers exist. The documented asymmetry (`C:\x` is a plain filename on Unix) is correct std behavior to leave alone.

**Commit 3 (clippy).** At commit 2, `cargo clippy -p mesh-llm-plugin --all-targets --target x86_64-pc-windows-msvc -- -D warnings` fails with exactly `needless_return` at `io.rs:205`, and the suggested fix is byte-for-byte what commit 3 applies. At the head it exits 0 with zero warnings. macOS clippy also exits 0 at both base and head.

One environment note for anyone reproducing: the repo's `.cargo/config.toml` pins `rustc-wrapper = "sccache"`, and a wedged sccache server fakes clippy exit 101 — set `RUSTC_WRAPPER=` (empty) to bypass it.

Approving — clean fix chain, correctly ordered, and the Windows-only defect class is exactly the kind of thing the Linux lane can't see.

PR #1980, opened 2026-09-21
fix(mesh): set claimed_log_head in the stale-announcement fixture, main is red without it

Virgile-pct:
`main` has been red since d71ceeaf1. `Main · Linux` and `Main · Quality` both fail there,
and a379ba625 just before it was green on both. Every PR branched on or merged against
that commit inherits it, which is how I noticed.

```
error[E0063]: missing field `claimed_log_head` in initializer of `mesh::peer_state::PeerAnnouncement`
   --> crates/mesh-llm-host-runtime/src/mesh/tests/protocol_frames.rs:860:15
error: could not compile `mesh-llm-host-runtime` (lib test) due to 1 previous error
```

It is a semantic conflict between two changes that were each green on their own:

- 57c8c9d31 (#1709) added `claimed_log_head` to `PeerAnnouncement`.
- d71ceeaf1 (#1970) added the `stale_serving_announcement` fixture, written against the
  struct as it was before that field existed.

Neither CI run could see the other, so the break only appeared once the second landed.

This sets the field to `None` in that one fixture, which is what the fixture means: it
models a bridge rebroadcasting a departed peer, and such a rebroadcast carries no opinion
about that peer's claimed log head. That matches `announcement_from_peer` in
`mesh/announcements.rs`, which sets `None` for the same reason. The test the fixture feeds,
`expired_tombstone_does_not_resurrect_dead_peer_from_stale_transitive_announcement`, is
unaffected by the value.

Measured on Windows 11, rustc 1.98.0, x86_64-pc-windows-msvc, at `main` 9d5728bf:

| `cargo check -p mesh-llm-host-runtime --all-targets` | result |
| --- | --- |
| without this change | `error[E0063]`, no build |
| with it | builds, 3 pre-existing warnings, unchanged |

`cargo test -p mesh-llm-host-runtime --lib expired_tombstone_does_not_resurrect`: 1 passed,
0 failed.

Not verified: I could not build for Linux from this machine, so the reproduction above is
the Windows one; the error is not platform-gated and CI on this PR is the real check. I did
not look for other fixtures written against the pre-#1709 struct, only this one, because it
is the only one the compiler stops on.

## Summary by CodeRabbit

* **Tests**
  * Clarified test coverage for bridge rebroadcasts involving departed peers with no claimed log head.


i386:
**Approve (as comment)** — the independent verification I posted on #1981 carries to this PR unchanged.

Verified at `a56ad4b49` (#1981) in a clean isolated worktree: `cargo test -p mesh-llm-host-runtime --lib` → 3778 passed / 0 failed / 11 ignored; `cargo clippy -p mesh-llm-host-runtime --all-targets -- -D warnings` → exit 0; fmt clean.

Checked this head (`4c7748b87`) directly rather than assuming equivalence: the patch is the same `claimed_log_head: None,` line in `stale_serving_announcement` against the same base blob (`052e8bc69`), plus the two-line comment — no behavioral delta, so those results transfer as-is. The comment's rationale is accurate: `announcement_from_peer` sets `claimed_log_head: None` (protocol_frames.rs:326), and the tests that actually exercise `claimed_log_head` semantics build announcements with `Some` via their own builders, so nothing here is weakened.

Either PR unblocks main; merge whichever is preferred and close the other.

Issue #1897, opened 2026-09-17
Make Skippy standalone with a shared serving path for Mesh


i386:
# Make Skippy standalone with a shared serving path for Mesh

Goal: a clean Skippy installation acquires a runtime/model and serves locally and across two explicit workers, without Mesh. Mesh consumes the same lifecycle and OpenAI serving APIs.

Skippy owns model/runtime management, split policy/execution, OpenAI semantics, inference guardrails and inference events. Mesh owns discovery/admission, Iroh, routing, console, management API and all plugin APIs/hosts. Skippy exposes typed hooks implemented by the Mesh adapter. No reverse Mesh dependencies, including build dependencies. Compatibility breaks are allowed with explicit ABI/protocol checks; preserve model/cache data.

- [ ] M1: Audit current main; remove reverse dependencies; establish Skippy-owned frontend, guardrails/events and typed hooks. Keep plugin APIs/hosts in Mesh.
- [ ] M2: Rehome native runtime manifests, installation, identity, environment/cache policy and version validation to Skippy.
- [ ] M3: Extract skippy-api lifecycle; add skippy-commands and skippy-cli, exposing model pull/list/remove, runtime management, local and explicit-worker serving. Convert skippy-server to skippy-serving.
- [ ] M4: Wire Mesh through mesh-llm-skippy-adapter; separate transport/membership/control composition and delete duplicated model/split/serving logic.
- [ ] M5: Move owned crates, tests, native patches, tooling and docs under mesh/ and skippy/, retaining root workspace/build tooling. Keep existing mesh-llm names and adopt agreed Skippy names; no shared tree.
- [ ] M6: Prove clean-install runtime/model acquisition, local serving, two-worker standalone splits, Mesh parity and clear incompatible runtime/worker rejection. Record exact revisions and full affected-package validation.

Zero reverse dependencies is an intermediate milestone, not completion. Standalone and Mesh must share request semantics, guardrails and streaming. No Mesh plugin host required for standalone operation.

Source: James's ownership instructions and Astrid's revised crate table in Buzz skippy thread bafcb52bda1515bfcd6f4cea91fc29864accb5cf0a062da77d88202018b61720, September 16–17, 2026. Supersedes the earlier proposal to move plugin API/host into Skippy.

## Delivery
James requested one PR for the complete extraction, with reviewer reviewing incremental checkpoints (September 17, 2026). Milestones are review checkpoints within that PR, not separate PRs.

[...]

PR #1983, opened 2026-09-21
fix(ci): support Mesh and Skippy source layout migration

i386:
CI can validate both the existing repository layout and the upcoming Mesh/Skippy product trees. Protected-main workflows currently check out candidate source but still address the old UI, website, SDK and Cargo package names; the catalog-only prerequisite did not update those consumers.

The executor workflows now use pinned compatibility actions for source directories and the fixed package-owner migration. Existing PR branches do not need to contain the new helpers. Missing or ambiguous paths and missing package successors fail; the 65-package pre-extraction plan covers all 75 extracted packages exactly once. UI artifacts retain their existing identity/checksum verification, and SafeTensors smoke retains its exact ignored-test assertion and complete quantization sweep.

The protected planner, byte-identical catalogs, matrix worker limits, permissions, runner/cache authority, five required lanes and cancellation behavior are unchanged. Native recipe keys include both source locations. No product source is relocated here. The departed-peer regression fixture also initializes the newly required `claimed_log_head` field; main currently omits it, preventing host test compilation.

### Rollout

Land this prerequisite before the dependent relocation. Matrix executor `generation: legacy` matches the current protected workspace. The extraction commit must change those two matrix calls to `generation: current` alongside the package renames; fixed platform owner requests remain legacy selectors. The explicit generation protects the reused `skippy-model-package` name in partial plans.

### Validation

- Full `just ci-validate`: 1,711 Python tests, nine expected skips, actionlint, diff checks, crate-list/release/console/publish consistency.
- Full prerequisite host package suite after the fixture repair: 3,778 unit tests plus 24 integration tests passed.
- Shellcheck of all 18 changed embedded Bash bodies.
- Executable old/new-layout UI producer/consumer checks; missing/ambiguous directory rejection.
- Frozen real workspace inventories verify package coverage without overlap, including the reused name.
- Exact SafeTensors test selection and missing-test rejection for both owners; one compile invocation.
- Both resolver actions exercised with candidate source separate from the action checkout.

Windows PowerShell and remote platform builds still require CI evidence. Local checks do not establish live product acceptance for the later extraction.

Origin: Buzz skippy channel `f5606e6b-e108-40ed-af72-1128393ecd87`, thread `bafcb52bda1515bfcd6f4cea91fc29864accb5cf0a062da77d88202018b61720`.

## Summary by CodeRabbit

* **CI Improvements**
  * CI now supports legacy and relocated product source layouts, with clear failures for missing or ambiguous directories.
  * UI, website, SDK, and artifact workflows dynamically locate their source directories.
  * Protected package selections are translated and validated across workspace generations.
  * Cache invalidation now accounts for additional Skippy build and patch files.
* **Bug Fixes**
  * Improved SafeTensors smoke-test selection across supported package layouts.
* **Tests**
  * Added coverage for source-layout resolution, package compatibility, workflow artifacts, and UI distribution verification.


Virgile-pct:
A Windows note on this one, since the body says the Windows PowerShell parts still need evidence, plus an interaction with #1985 that neither thread mentions yet. cc @michaelneale for the #1985 side.

**#1985 and this PR rewrite the same Windows loop.** #1985 adds `mesh-llm-plugin` to the windows-unit `foreach`, on Windows only. Here that loop becomes `foreach ($crate in ($env:TEST_CRATES | ConvertFrom-Json))`, fed by a single `resolve-cargo-packages` step whose `crates:` input is shared with the macOS row and lists three crates. Simulated with `git merge-tree` on `main` (`c70a1341e`): in either order, the second PR conflicts in `ci-platform-checks-slice.yml` on exactly that loop, so nothing drops silently, but resolving toward this PR's side puts the Windows row back to three crates.

#1985's census is the tripwire, and it fires, though not only when it should. Measured in a scratch merge (`main` + #1985, then this PR with conflicts resolved to this side):

- `test_windows_cfg_domain_crates_run_in_the_windows_unit_row` fails with `'mesh-llm-plugin'` missing. Good.
- With the shared `crates:` list extended to four, which is a correct resolution, it fails identically. `_windows_unit_row_crates()` only reads quoted names inside `foreach (...)`, and in this PR's form the names live in the `crates:` input.

So whichever lands second also has to teach that parser the new form, and needs either a Windows-specific list or a second resolve step, unless `mesh-llm-plugin` should run on the macOS row too. This branch already conflicts with `main` on `ci/ci.md` and `current-inventory.md`, so the rebase may be the natural moment.

**Windows evidence for the parts that run on Windows runners**, Windows 11, this head (`afb1b1020`):

- `resolve-source-layout`, its bash body under Git Bash: `ui_dir=crates/mesh-llm-ui`, `website_dir=website`, `sdk_dir=sdk`.
- `resolve-cargo-packages`, generation `legacy`, with `GITHUB_ACTION_PATH` in backslash form as on a runner (the step appends `/../../../scripts/...` to it): returns `["model-artifact", "mesh-llm-host-runtime", "mesh-llm"]`, and #1985's four-crate list comes back unchanged.
- The new loop fed with that exact output: 3 iterations, 4 with the plugin list, 1 for a one-element list, 0 for `[]`. `Join-Path $env:UI_DIR "dist"` and `"$env:UI_DIR/dist/index.html"` build the expected paths.

Caveat: that last part ran in Windows PowerShell 5.1. Neither of my Windows boxes has pwsh 7, which is what `shell: pwsh` uses on the runners, so the loop is not verified under the exact engine.

Virgile-pct:
Closing my own caveat above: installed pwsh 7.6.6 on the second Windows box and reran the loop with the resolver's exact output. Same results as under 5.1: 3 iterations, 4 with the plugin list, 1 for a one-element list, 0 for `[]`, and the same UI paths. So the PowerShell side of this PR behaves the same on the engine `shell: pwsh` uses.

i386:
Re-verified this head (`9592495a8`) against the Windows note and both CodeRabbit findings; all three are addressed.

@Virgile-pct thanks for closing your own caveat. The interaction you flagged with #1985 is what the rewrite is built around rather than something it collides with. The windows-unit crate list no longer comes from a literal `foreach` list: one `resolve-cargo-packages` step emits the resolved set (including the new `crates:` input form), a Windows-only step appends `mesh-llm-plugin` for the `platform-windows-cfg` domain, and `_windows_unit_row_crates()` in the census parser was taught that new form. So whichever of #1985 / this PR lands second, the plugin entry cannot be silently dropped by a merge that takes either side verbatim.

Also folded into this head:

- `scripts/affected-crates.sh`: `mesh/crates/` and `skippy/crates/` now reach Cargo ownership matching (they were previously skipped before reverse-dependency matching), and the UI predicate accepts the relocated Mesh UI path. That is the CodeRabbit `affected-crates.sh:249` finding.
- `ci/derive-outputs.sh`: `skippy-native-runtime` changes now route to the SDK smokes. That is the CodeRabbit `derive-outputs.sh:390` finding.

One note on lineage: this branch was not rebased. `main` was merged in (`37e772cf3`) deliberately, because the PR pins both `resolve-*` actions to its own commit and `ci/ci.md` documents that pin as the reason older PR source checkouts do not need the new helper files. A rebase rewrites that commit and the pin dies on merge; a merge keeps it an ancestor.

One open failure on this head: `Linux product smoke / CUDA inference smoke` on the self-hosted CUDA runner. Every other job on the Linux lane passed (KV caching, two-node client, core inference, model download), so it does not look related to the crate-ownership change. I have re-run that job and will diagnose it if it repeats.

.github/workflows/ci-web-slice.yml
```diff
@@ -179,11 +188,14 @@ jobs:
         with:
           ref: ${{ inputs.source_sha || github.sha }}
           persist-credentials: false
+      - name: Resolve product source layout
+        id: layout
+        uses: Mesh-LLM/mesh-llm/.github/actions/resolve-source-layout@38d63b2f6e27998034fdf0452150c7cc081fe921 # product-layout compatibility helpers
```
ndizazzo: [P1] Propagate the resolved website path through the root recipe

This resolver only changes the workflow's working directory; `just website-build` still expands the root `Justfile`'s literal `website_dir := "website"` and `just/website-ui.just` does `cd "{{ website_dir }}"`. In the planned `mesh/website` checkout, this step resolves successfully and then fails looking for root `website` (the same recipe also owns `cli-inventory-check`). Please make the recipe consume the resolved path or add a relocated-layout test that executes this command boundary.

scripts/release-version.sh
```diff
@@ -226,12 +226,14 @@ done < <(
     cd "$REPO_ROOT"
     git ls-files \
         'crates/*/Cargo.toml' \
+        'mesh/crates/*/Cargo.toml' \
```
ndizazzo: [P1] Resolve release-version sidecars from the same source layout

Adding the relocated Cargo globs is not sufficient: this script still unconditionally requires `$REPO_ROOT/sdk/kotlin`, and `literal_version_files` still names `crates/mesh-llm-*`, `sdk/node`, `sdk/swift`, and `website`. A relocated-only checkout can therefore discover its manifests successfully and then fail before version propagation. Please migrate these sidecars together or explicitly defer this release lane, with a full relocated-only fixture.

.github/actions/compute-changes/derive-outputs.sh
```diff
@@ -381,7 +381,7 @@ SDK_SMOKE_REQUIRED="false"
...
+  DIRECT_SDK_INPUTS=$(echo "$CHANGED_FILES" | grep -E '(^(mesh/)?sdk/|^Package\.swift$|^(mesh/|skippy/)?scripts/ci-(rust|kotlin|swift)-sdk-smoke\.sh$|^(mesh/|skippy/)?scripts/ci-prepare-native-runtime\.sh$|^(mesh/|skippy/)?scripts/ci-sdk-fixture\.sh$|^(mesh/|skippy/)?scripts/(check-sdk-contract|package-sdk-console-assets|restore-native-sdk-input|restore-static-abi-input|verify-sdk-console-assets|verify-swift-privacy-manifest|verify-swift-release-artifact|prepare-llama|build-llama)\.sh$|^(mesh/|skippy/)?scripts/(package-native-sdk|package-native-sdk-crate|verify-native-sdk-package|verify-checksum-sidecar|verify-static-abi-build-stamp|safe-extract-(tar|zip)|verify-swift-xcframework)\.(sh|py)$|^\.github/actions/(compute-changes|prepare-native-sdk-input|prepare-static-abi-input|resolve-native-toolchain-epoch|restore-smoke-inputs)/|^\.github/workflows/(ci|main_[a-z]+|native-sdk-artifact|pr_[a-z]+|pr_ci_canary|release|sdk-smoke|static-abi-artifact|swift-sdk-artifact)\.yml$)' || true)
```
ndizazzo: [P1] Also gate inference artifacts on the relocated runtime

The existing SDK-smoke selector below is already missing `skippy-native-runtime`, but this direct-path expansion also needs the artifact consumer updated. With only `AFFECTED_CRATES=["skippy-native-runtime"]`, both `SDK_SMOKE_REQUIRED` and `INFERENCE_ARTIFACT_REQUIRED` remain false, so the runtime artifact and dependent Skippy smoke can be skipped. Add the new owner to both selectors and a regression for this case.

scripts/ci-cargo-packages.py
```diff
@@ -0,0 +1,99 @@
...
+    requested = names(requested)
+    planned = names(planned)
+    if not set(requested).issubset(planned):
+        raise ValueError('requested batch contains packages outside the protected plan')
+    if generation not in {"legacy", "current"}:
+        raise ValueError("unsupported planner package generation")
+    migrating = generation == "legacy" and "skippy-package-builder" in available
+    result = []
+    for name in requested:
+        candidates = SUCCESSORS.get(name, [name]) if migrating else [name]
```
ndizazzo: [P2] Couple successor translation to semantic slice coverage

This maps protected batches to successor package names, but it verifies only Cargo membership. It does not verify that successor names are represented in the protected planner's semantic ownership/catalog inputs; successor-only changes such as `skippy-native-runtime`, `skippy-runtime-install`, `skippy-guardrails`, or `skippy-hardware-profile` can still fall through to generic Rust routing while runtime-product/artifact/smoke consumers are skipped. Please add a planner-level fixture/assertion for the successor set, or make the catalog migration a hard prerequisite for using this map.

ndizazzo (commented):
Reviewed the exact PR head `9592495a83a67d4b405d9422ae1de43669feacbd`. Posting this as a comment-only review per request; it is neither an approval nor a rejection.

The inline comments cover the correctness issues found in the changed CI and release paths.

Plan alignment: [issue #1897](https://github.com/Mesh-LLM/mesh-llm/issues/1897) describes a full extraction. This PR is a useful compatibility prerequisite, but it is only partial M5 work. It does not deliver the plan's reverse-dependency/ownership extraction (M1), runtime manifest/install/cache identity migration (M2), standalone lifecycle/CLI/serving (M3), adapter/transport/membership split (M4), or acquisition/local/two-worker/Mesh parity and rejection evidence (M6).

Downstream blast radius audited at the current consumers:

- `mesh-packaging` still assumes `sdk/node` in `packaging/images.json`, `scripts/image-matrix.ts`, `scripts/node-sdk-package.ts`, and the release image workflow. The current layout remains compatible, but the physical SDK move will require those consumers to resolve `sdk`/\`mesh/sdk\` consistently.
- `mesh-llm-runner-images` still has legacy `crates/mesh-llm-ui`/`website` assumptions in dependency warming, UI preparation, and runner identity collection. These do not break this prerequisite while the current layout remains, but they can silently miss warmups or identity inputs after the move.

The existing CodeRabbit SDK-selector comment and the existing #1985 integration-conflict note remain relevant; I did not duplicate either one inline.

